PR #50, opened 2026-05-25
Phase 5b/5c/5d: Active-set SQP solver with working-set warm start

jkitchin:
## Summary

This PR introduces a complete active-set sequential quadratic programming (SQP) solver alongside pounce's existing interior-point method. The new solver is designed for warm-started NLP sequences (MPC, parametric continuation, MINLP branch-and-bound) where the working set can be carried across solves to achieve O(1) QP iterations when the active set is unchanged.

### New `pounce-qp` Crate
- **Sparse parametric active-set QP solver** with Schur-complement updates for efficient working-set changes
- **Cold-start path** (Phase 5a): KKT assembly and single factor-and-solve for equality-only QPs
- **Parametric warm-start** (Phase 5b): reuse cached factorization across consecutive QPs via Sherman-Morrison-Woodbury formula
- **Inertia-controlled factorization** (§4.5): automatic diagonal shifting when the Hessian is indefinite
- **l1-elastic phase-1** (§4.3): infeasibility detection and minimal-infeasibility solutions
- **Working-set management**: add/drop constraints with EXPAND anti-cycling
- **QPS reader**: pure-Rust Maros-Mészáros format parser for benchmark problems

### SQP Algorithm (`pounce-algorithm::sqp`)
- **Outer-loop SQP driver** consuming the same `IpoptNlp` interface as the IPM
- **Two globalization strategies**: Fletcher-Leyffer filter (default) and Han-Powell l1-merit
- **Hessian approximation options**: exact Hessian, Powell-damped BFGS, or limited-memory BFGS
- **Parametric corrector** for sensitivity-based warm starting (Phase 5c §6/§7)
- **Working-set classification** from IPM-converged multipliers via `classify_working_set`

### API Integration
- **Algorithm selection**: new `algorithm = active-set-sqp` option in `AlgorithmBuilder`
- **Warm-start interface**: `set_sqp_warm_start()` / `last_sqp_working_set()` on `IpoptApplication`
- **Python bindings**: `Problem.solve(..., working_set=...)` and `classify_working_set()` function
- **C ABI**: `IpoptSetWarmStartWorkingSet`, `IpoptGetWorkingSet`, `IpoptSolveWarmStart`
- **GAMS link**: persistent working-set state files via `sqp_state_file` option

### Testing & Validation
- **Analytical correctness ladder** (§8.0): 6 closed-form QPs with hand-computable answers
  - Problems 1–2 (equality-only, unconstrained) in Phase 5a
  - Problems 3–6 (bounds, degeneracy, infeasibility, indefinite Hessian) in later phases
- **Hock-Schittkowski subset**: 10 published NLP benchmarks with closed-form optima
- **Maros-Mészáros framework**: QPS reader + published-optimum comparison harness
- **Unit tests**: KKT assembly, Schur updates, elastic reformulation, scaling sweeps
- **Integration tests**: parametric corrector handoff, filter vs. l1-elastic comparison

### Documentation
- **Design note** (`dev-notes/research/active-set-sqp-warm-start.md`): 1435 lines covering algorithm, architecture, warm-start contract, and all design choices pinned to literature
- **User tutorial** (`docs/tutorials/active-set-sqp.md`): when/how to use the SQP path
- **Python notebook**: parametric continuation demo with iteration-count comparison
- **C example**: working-set warm start across a sequence of perturbed QPs
- **GAMS example**: parametric sweep with persistent state files

## Notable Implementation Details

- **No architectural changes to IPM**: SQP is a new `AlgorithmStrategy` end-to-end; the existing

https://claude.ai/code/session_011A8LRAEsmyAmf2kHsEtpVQ


jkitchin:
## Review: Phase 5b/5c/5d — Active-set SQP with Working-set Warm Start

### Summary

Well-structured addition grounded in the literature, with unusually thorough inline documentation. The algorithm design is sound. However, there is one critical correctness bug that should block merge, one subtle API/interface bug that will silently misclassify working-set variables, and a non-standard Armijo condition that degrades convergence on nonconvex problems. Several testing gaps are significant given the algorithmic complexity.

---

### Correctness / Numerical Issues

**C1 — Critical: `cold_general_initial` drops the Hessian term from the RHS** (`crates/pounce-qp/src/solver.rs:931`)

```rust
for (rhs_i, &g_i) in rhs[..n].iter_mut().zip(qp.g.iter()) {
    *rhs_i = -g_i;   // missing -Hx term
}
```

Every other path uses `−(Hx + g)`. This is harmless when `x = 0` exactly, but when the equality-relaxed KKT solve returns a nonzero `x` (i.e. essentially every non-trivial problem), the RHS is wrong and the resulting primal warm-start is incorrect. Fix: add the `h_times_x` contribution before this loop.

**C2 — Moderate: Schur `v_cols` layout has no round-trip correctness test** (`crates/pounce-qp/src/schur.rs:371–374`)

The cross-indexing of `u_cols`, `v_cols`, and `kinv_u_cols` across rank-2 updates (`u vᵀ + v uᵀ`) is correct on close reading but sufficiently opaque that a future edit could break the sign silently. No test verifies `K_W * K_W⁻¹ * b == b` for a multi-step update sequence.

**C3 — Moderate: Armijo fallback clause nullifies the sufficient-decrease condition** (`crates/pounce-algorithm/src/sqp/line_search.rs:139`)

```rust
let armijo_ok = phi_trial <= target || phi_trial < phi_curr;
```

The `|| phi_trial < phi_curr` arm accepts *any* descent, even by 1e-300, which bypasses the Armijo condition entirely on nonconvex problems where `predicted < 0`. Standard fix: check `predicted >= 0` separately; don't bypass sufficient-decrease as the fallback.

**C4 — Minor: Filter gives no protection on the first step** (`crates/pounce-algorithm/src/sqp/filter.rs`)

An empty filter accepts any trial iterate unconditionally. The first full step is always accepted regardless of constraint violation increase. Should at minimum be documented in the module-level caveat alongside the existing f-mode/h-mode note.

**C5 — Minor: `expand_tol` grows on every iteration, including unblocked steps** (`crates/pounce-qp/src/solver.rs:878`)

In classical EXPAND (GMSW 1989), τ grows only when a constraint blocks (α < 1). Growing unconditionally means the hard-reset snap happens on non-degenerate problems, discarding numerical precision unnecessarily.

---

### Safety / Rust Issues

**S1 — Minor: `worst.unwrap()` inside a match arm that already destructures `worst`** (`crates/pounce-qp/src/solver.rs:724`)

```rust
Some((_, prev_viol)) => { let prev_key = drop_target_key(worst.unwrap().0); }
```

Cannot actually panic (match guarantees `Some`), but the `unwrap()` is needless. Change the pattern to `Some((prev_target, prev_viol))` and use `prev_target` directly.

**S2 — Minor: `IpoptSolveWarmStart` silently discards invalid warm-start status** (`crates/pounce-cinterface/src/lib.rs:1151–1155`)

If `IpoptSetWarmStartWorkingSet` returns `FALSE` (invalid status codes), the error is silently dropped and the solve proceeds cold. C callers have no way to detect the silent fallback. Surface the return value or document the behavior.

**S3 — Minor: `i64 as Index` (`i32`) cast can silently truncate** (`crates/pounce-py/src/problem.rs:83–84`)

The `n < 0` check is there but there's no upper-bound check. Add `if n > i32::MAX as i64 { return Err(...) }`.

**S4 — Minor: `#![allow(unsafe_op_in_unsafe_fn)]` blanket suppression** (`crates/pounce-cinterface/src/lib.rs:31`)

Masks the per-block audit trail for pointer dereferences. Consider removing and making each `unsafe {}` block explicit.

**S5 — Minor: `debug_assert!` for dimension validation in release hot paths** (`crates/pounce-algorithm/src/sqp/bfgs.rs:93`, `crates/pounce-algorithm/src/sqp/lbfgs.rs:86`)

Compiled away in release; a mismatched-length array silently produces wrong results. Use `assert!` or return `Result`.

---

### API / Interface Issues

**A1 — `IpoptSolveWarmStart` has no null-check on output pointers** (`crates/pounce-cinterface/src/lib.rs:1157`)

`x`, `g`, `obj_val`, etc. are passed straight through from the caller without null validation. Document whether NULL is accepted or add early-return on required-pointer NULL.

**A2 — GAMS state-file corruption falls back silently** (`gams/gams_pounce.c:260–280`)

A corrupted or dimension-mismatched `sqp_state_file` triggers a silent cold start. Should emit at least one `gevLog` diagnostic so operators know warm-start was lost.

**A3 — `Problem.solve(working_set=...)` silently ignores the argument on the IPM path** (`crates/pounce-py/src/problem.rs:180–192`)

If the user passes `working_set=` but `algorithm="interior-point"`, the value is silently discarded. Should issue a `warnings.warn()` so users catch the mistake.

**A4 — Moderate: `classify_working_set` uses `is_finite()` instead of `NLP_*_BOUND_INF` sentinels** (`crates/pounce-algorithm/src/sqp/warm_start.rs:72`)

```rust
if x_l[i].is_finite() { ... }
```

`NLP_LOWER_BOUND_INF = −1e19` is finite by IEEE 754. Variables whose bound is exactly the pounce sentinel will be incorrectly classified as `AtLower`. Use the same sentinel comparison (`x_l[i] > NLP_LOWER_BOUND_INF`) used everywhere else in the codebase.

---

### Testing Gaps

**T1** — No test drives an indefinite Hessian end-to-end through inertia correction and verifies the shift magnitude and solution correctness. Analytical ladder Problem 6 is listed as a future commit.

**T2** — EXPAND anti-cycling machinery (`expand_tol` growth, τ-reset) has no test that would detect silent breakage. Problem 4 is deferred.

**T3** — Parametric warm-start tested only when the active set is *unchanged* across the perturbation. Need a test where one constraint drops and another enters.

**T4** — No multi-step Schur test cross-checks `K_W⁻¹ b` against a fresh factorization after 3–4 working-set changes (directly relevant to C2 above).

**T5** — `solve_parametric` is a stub that ignores `qp_prev`/`sol_prev` and cold-starts. Tests calling it are testing the cold-start path under a misleading name.

---

### Minor / Style

- `crates/pounce-qp/src/solver.rs:1413`: `r > alpha_min + 0.0` — no-op `+ 0.0`; remove or replace with `f64::EPSILON`.
- `crates/pounce-qp/src/elastic.rs:169`: `original_inertia()` always returns `Psd`, making the `Indefinite` branch in `as_qp()` dead code. Should propagate the original QP's inertia hint.
- `crates/pounce-qp/src/factor.rs:164`: `expect("factored ⇒ cache present")` could just be an `Option`-free `Vec` with `factored` as the guard.
- ~40 lines of duplicated H-block normalization across four KKT assembly functions. A `push_h_lower_triangle` helper would eliminate the duplication and a future sign-bug vector.
- `crates/pounce-algorithm/src/sqp/lbfgs.rs::materialize()` allocates an n×n dense matrix every SQP outer iteration. Acknowledged in the doc but the call site in `sqp_alg.rs` should have a `// PERF:` comment flagging the current overhead.

---

### Verdict

**Request changes.** Two bugs need to be fixed before merge:

1. **C1** — `cold_general_initial` RHS missing `−Hx` (incorrect warm-start primal for any non-trivial problem)
2. **A4** — `is_finite()` vs `NLP_LOWER_BOUND_INF` sentinel in `classify_working_set` (silent misclassification of unbounded variables)

**C3** (Armijo bypass) should also be fixed to prevent degraded convergence on nonconvex problems. The remaining findings are improvements rather than blockers, but T1–T4 are worth addressing before this path is used in production.

---

*Review generated with assistance from Claude Code.*

jkitchin:
Thanks for the thorough review. Pushed `b3480f7` addressing the findings. Triage and disposition:

### Blockers — fixed

- **A4** (`classify_working_set` sentinel): switched from `is_finite()` to `> NLP_LOWER_BOUND_INF` / `< NLP_UPPER_BOUND_INF` in `crates/pounce-algorithm/src/sqp/warm_start.rs`. Added regression `classify_treats_nlp_bound_inf_sentinel_as_unbounded` that pins the −1e19 sentinel case to `Inactive`.
- **C3** (Armijo bypass): the `|| phi_trial < phi_curr` fallback now gates on `predicted >= 0` — strict Armijo when `predicted < 0` (the descent-direction case where the test has teeth), relaxed acceptance only when the predicted derivative isn't a descent direction. Cites Wächter-Biegler 2006 §3.3 in the comment.

### Confirmed non-bug

- **C1** (`cold_general_initial` RHS missing `−Hx`): respectfully pushing back here. `cold_general_initial` solves for the *primal x directly*, not a step from a current x. The KKT system `[H Aᵀ; A 0][x; λ] = [−g; b_eq]` returns `x* = argmin ½xᵀHx + gᵀx s.t. Ax = b_eq` — the H term is in the matrix factor, not the RHS. There's no current x to subtract from at the cold start. Compare to `solve_general`'s step-based inner loop (line 684–689), which solves for a step `p` from a known `x` and correctly uses RHS = `−(Hx + g)`. Two different (and mathematically equivalent) formulations. Added explicit regression `cold_general_initial_solves_qp_with_nonzero_hessian_at_solution` with `H = 4I, g = −2·1, A x = 1` ⇒ `x* = (0.5, 0.5)` — exercises a non-zero H contribution at the optimum; pre- and post-c37 it returns the same correct primal. The existing HS76 fixture (also non-trivial H) similarly passes through this code path.

### Other improvements landed

- **C2** — added `schur_multi_step_add_drop_add_matches_fresh_factor`, a 3-change cross-check (add → drop → add) against a fresh factor.
- **C4** — documented the first-step filter caveat in `sqp/filter.rs` module doc; pointed users at `L1Elastic` for hard first-step constraint-violation bounding.
- **C5** — gated EXPAND τ growth on `blocker.is_some()` per GMSW 1989 §3 (both `solve_general` and `solve_general_schur`); non-degenerate problems no longer trip the hard reset.
- **S1** — replaced `worst.unwrap()` with destructuring `Some((prev_target, prev_viol))`.
- **S2** — documented `IpoptSolveWarmStart`'s silent-fallback semantics for invalid warm-start input in the header.
- **S3** — added i64 → i32 truncation guards on `n`, `m` in `PyProblem::new`.
- **S5** — promoted `debug_assert_eq!` → `assert_eq!` for dimension checks in `BFGS::update` and `LBFGS::update`.
- **A2** — GAMS solver link now logs both the state-file hit and miss outcomes via `gevLogStat`.
- **A3** — `Problem.solve(working_set=…)` issues a `warnings.warn()` if `algorithm` is not `active-set-sqp`.
- **T3** — added `sqp_corrector_handles_active_set_flip_between_perturbations` that releases an active bound when one parameter coordinate sign-flips.
- Minor: removed `+ 0.0` no-op in the blocker selector.

### Acknowledged, deferred

- **S4** (`#[allow(unsafe_op_in_unsafe_fn)]`) — kept as-is for parity with the rest of `pounce-cinterface`; per-block audit-trail enablement is a separate refactor across the crate.
- **A1** (`IpoptSolveWarmStart` output NULL checks) — `IpoptSolve` (which `IpoptSolveWarmStart` delegates to) already handles NULL for every output pointer; the new entry inherits that behaviour. Happy to add an explicit note to the header doc if you'd like.
- **T1, T2, T4** — focused additional tests for end-to-end inertia-correction, EXPAND τ-reset, and the multi-step Schur back-solve. T1 and T2 are tracked in §10 of the design note as Phase 5a+ items. T4 is now partially covered by the new `schur_multi_step_add_drop_add_matches_fresh_factor`.
- **T5** (`solve_parametric` stub) — documented in `solver.rs` as falling through to cold-start; the parametric integration that would replace it is Phase 5c+ proper.
- L-BFGS `n×n` materialization — already documented in the design note §4.6; not adding a `// PERF:` to the call site because the design choice (PSD dense triplet feeding the QP) is intentional.
- ~40 lines of duplicated H-block KKT assembly — agreed, but the refactor crosses four KKT assembly functions and warrants its own focused PR.

**Test counts**: 922 workspace tests passing (up from 918) including 4 new regressions targeting the C1, A4, C2, and T3 findings.

Re-review when you're able. Happy to spin any of the deferred items into a follow-up PR if you'd like them in this one.

---

jkitchin:
Follow-up to the earlier review (commit b3480f7):

- **A4** and **C3** fixed with regression tests — looks correct.
- **C1** was a misread on my part. `cold_general_initial` solves the equality-relaxed KKT system *directly for the primal x* (not a Newton step from a current iterate), so RHS = `[-g; b_eq]` is correct — the H term belongs in the matrix factor, not the RHS. The `solve_general` inner loop solves for a step `p`, which is why that RHS is `-(Hx + g)`. The new `cold_general_initial_solves_qp_with_nonzero_hessian_at_solution` regression pins this. Apologies for the noise.
- C2/C4/C5, S1/S2/S3/S5, A2/A3, T3 all addressed.
- Deferred items (T1, T2, T5, L-BFGS PERF) have clear reasoning.

918 → 922 tests passing. Merging.

PR #60, opened 2026-05-27
Auxiliary-equality preprocessing (closes #53)

### LINKED ISSUE #53: Auxiliary equality preprocessing (port of ripopt #32)
## Summary

Port ripopt PR #32 (https://github.com/jkitchin/ripopt/pull/32) — auxiliary equality-subsystem preprocessing — into `crates/pounce-presolve` as a Phase-0 pre-pass on `PresolveTnlp`. Adds structural NLP preprocessing (incidence → Hopcroft-Karp matching → Dulmage-Mendelsohn → BTF → block solve → reduced problem → postsolve with multiplier recovery), gated behind `presolve_auxiliary` (default off).

## Motivation

- ripopt PR #32 demonstrates substantial reductions on real problems (e.g. `gaslib11_steady`: 204/200 → 140/136 vars/cons; `tutorial_flow_density`: 6–7 IPM iterations → 0).
- `pounce-presolve` currently has bound tightening, redundant-row removal, LICQ, warm starts, and sensitivity passthrough — but no structural preprocessing. This issue fills that gap.

## Design

See `dev-notes/auxiliary-equality-preprocessing.md` for the full design.

## Phased PRs

- [x] PR 1 — scaffolding, options, no-op orchestrator
- [x] PR 2 — incidence + Hopcroft-Karp matching
- [x] PR 3 — Dulmage-Mendelsohn + connected components
- [x] PR 4 — Tarjan SCC + block-triangular form
- [x] PR 5 — coupling classification + rejection diagnostics
- [x] PR 6 — block solve (lightweight Newton ≤ 8-dim)
- [x] PR 7 — reduction frame + postsolve multiplier recovery
- [x] PR 8 — orchestrator wiring into `PresolveTnlp` + full-space residual check (linear blocks)
- [x] PR 8b — nonlinear block support via TNLP-backed Newton callbacks
- [x] PR 9 — diagnostics + user docs
- [x] PR 10 — CLI integration test + benchmark harness (ripopt fixture vendoring deferred)
- [x] PR 11 — large-block solver fallback (RelaxedNewtonSolver default; pounce-algorithm IPM impl deferred)
- [x] PR 13 — trivial-elimination pre-pass (fixed vars, free rows, trivially-slack inequalities) before Hopcroft-Karp; shrinks the incidence graph
- [x] PR 14 — InequalityCoupled block admission via inequality projection; widens the accepted class for linear blocks coupled to linear inequalities implied by the variable box
- [ ] PR 12 — default flip (after benchmark evidence)

Each PR lands behind the master switch, so the feature stays a no-op for users until PR 12.

Validation work:

- [x] König-theorem fuzz cross-check for Hopcroft-Karp
- [x] DM partition formal-invariant fuzz test
- [x] Components vs independent BFS-reference fuzz test
- [x] BTF vs Floyd-Warshall transitive-closure SCC fuzz test
- [x] Newton vs direct-LU fuzz on random linear systems
- [x] Reduction-frame multiplier-recovery fuzz on synthetic KKT problems
- [x] End-to-end pipeline fuzz (incidence → ... → multiplier recovery) on 20 random lower-triangular problems
- [x] Hand-crafted pipeline worked example with full KKT residual verification
- [x] CLI smoke test against `parametric.nl`
- [x] e2e CLI tests on the vendored ripopt fixtures (tutorial_flow_density 0 IPM iters; gaslib11_steady regression-tests the aux+bound_tighten interaction)

Follow-ups gated on vendoring the ripopt `.nl` fixtures (vendored in `f4baa65`):

- [x] Vendor `tutorial_flow_density{,_perturbed}.nl` from ripopt with EPL-2.0 attribution
- [x] Vendor `gaslib11_steady.nl` from ripopt (GasLib CC-BY-SA 3.0 documented)
- [x] Run `benchmarks/preprocessing/run_preprocessing_benchmark.py` and confirm 0 IPM iterations on the tutorial fixtures
- [ ] Confirm `gaslib11_steady.nl` reduces to ~140/136 vars/cons (currently lands at a different local optimum than the un-presolved path; tracked as a separate concern before flipping the default to on)

Architectural follow-ups:

- [ ] `pounce-algorithm`-side `LargeBlockSolver` impl that re-enters the IPM (trait + RelaxedNewtonSolver fallback already in place via PR 11)
- [ ] Postsolve recovery for fixed variables at original bounds (v1 assumes interior)
- [ ] Iterative re-elimination (v1 runs Phase 0 once)
- [ ] General Fourier-Motzkin projection that synthesizes new rows on the IPM-visible problem (v1 PR 14 only admits when projected inequalities are implied by the variable box)

Review follow-ups (from #60 review):

- [ ] Bound-aware `BlockSolveOptions` + `BlockSolveError::OutOfBounds` (defence in depth #3)
- [ ] `rcond_threshold` in dense LU
- [ ] Hopcroft-Karp `dist[NIL]` sentinel for proper √V phase bound
- [ ] Per-block CSR index in orchestrator inner loop
- [ ] Reduction-frame sparse RHS for recovery solve
- [ ] `project_lambda` / `lift_lambda` paired round-trip test + multi-frame `ReductionStack` round-trip

## Non-goals

- discopt presolve orchestrator / pass-delta protocol / Python handshake
- tree-persistent bound tightening, MINLP reformulation
- linear-solver-level structural reordering

## Acceptance

- `tutorial_flow_density.nl` and `tutorial_flow_density_perturbed.nl` solve in 0 IPM iterations with `presolve_auxiliary=yes`.
- `gaslib11_steady.nl` reduces to ~140/136 vars/cons and converges to the same objective within tolerance.
- No regressions on existing `pounce-presolve` tests.

## References

- ripopt PR #32: https://github.com/jkitchin/ripopt/pull/32
- Fork tracker (bernalde/ripopt#11): https://github.com/bernalde/ripopt/issues/11
- ripopt source: `src/auxiliary_preprocessing.rs`, `src/reduction_frame.rs`, `src/preprocessing.rs`, `src/ipm.rs` (`try_auxiliary_preprocessed_solve`).
- Related discopt issue: jkitchin/discopt#53.

jkitchin:
Implements the auxiliary-equality preprocessing pass from
[pounce#53](https://github.com/jkitchin/pounce/issues/53) — a port of
[ripopt PR #32](https://github.com/jkitchin/ripopt/pull/32) into
`crates/pounce-presolve`. The pass identifies small square equality
sub-systems in an NLP, solves them structurally, and pins their
variables / drops their rows before the IPM runs. Gated behind
`presolve_auxiliary` (default off); the wrapper is byte-identical to
today when the master switch is off.

## Pipeline

For every TNLP wrapped by `PresolveTnlp` with `presolve_auxiliary=yes`:

```
incidence  → Hopcroft-Karp matching  → Dulmage-Mendelsohn partition
  → connected components  → Tarjan SCC / block-triangular form
  → coupling classification  → block solve (lightweight or large)
  → full-space residual check  → reduction frame
```

After the IPM finishes, `finalize_solution` walks the reduction
stack: full-space `∇f` and Jacobian at `sol.x`, dense-LU
stationarity solve per frame to recover the dropped rows'
multipliers, splice into the full λ before forwarding to the inner
TNLP.

## Commits (15, one per PR-sized change)

| # | Commit | Title |
|---|--------|-------|
| PR 1 | `9bf4472` | scaffold auxiliary-equality preprocessing |
| PR 2 | `535b6a5` | incidence + Hopcroft-Karp matching |
| PR 3 | `68047bb` | Dulmage-Mendelsohn + connected components |
| PR 4 | `7df6d96` | Tarjan SCC + block-triangular form |
| | `a3dc14a` | fuzz invariants for DM/components/BTF |
| PR 5 | `8219d98` | coupling classification |
| PR 6 | `192225d` | damped Newton block solver |
| PR 7 | `9fd038b` | reduction frame + multiplier recovery |
| | `cb115a3` | fuzz tests + end-to-end pipeline validation |
| PR 8 | `5e5f308` | orchestrator wired into PresolveTnlp |
| PR 9 | `23290c4` | expanded diagnostics + user docs |
| PR 8b | `89c5c88` | nonlinear block elimination |
| | `b98c431` | wire presolve_auxiliary_diagnostics flag |
| PR 10 | `19b6fb6` | CLI integration test + benchmark harness |
| PR 11 | `c6e040e` | large-block solver fallback |

The "Commits" tab is the natural review order — each commit carries
a per-PR plan in its body explaining what landed and why.

## What's in

- 11 new modules under `crates/pounce-presolve/src/`: `incidence`,
  `matching`, `dulmage_mendelsohn`, `components`, `btf`, `coupling`,
  `block_solve`, `reduction_frame`, `auxiliary`, `diagnostics`
  (plus extensions to `lib.rs` and `options.rs`).
- Six new options: `presolve_auxiliary`, `_tol`,
  `_max_block_dim`, `_wall_time_fraction`, `_coupling`,
  `_diagnostics`. Documented in `docs/src/options.md`.
- New user-facing page `docs/src/auxiliary-presolve.md`, registered
  in `SUMMARY.md`.
- 9 worked examples under `crates/pounce-presolve/examples/`.
- 119 unit + 1 integration (20 random seeds) + 9 doctests, all
  green on `cargo test -p pounce-presolve`.
- CLI smoke test against `parametric.nl` under
  `crates/pounce-cli/tests/auxiliary_presolve.rs`.
- Benchmark script `benchmarks/preprocessing/run_preprocessing_benchmark.py`.

## What's deferred (tracked in #53)

- Vendoring the ripopt `.nl` fixtures (`tutorial_flow_density.nl`,
  `gaslib11_steady.nl`) — they aren't accessible from the sandbox
  and need their MIT headers carried. Drop into
  `crates/pounce-cli/tests/fixtures/aux_presolve/` to unlock the
  issue's headline acceptance criteria.
- `pounce-algorithm`-side `LargeBlockSolver` impl that re-enters
  the IPM. PR 11 ships the trait + a `RelaxedNewtonSolver` default;
  the IPM-backed impl is a drop-in.
- Flipping `presolve_auxiliary` default to `yes` (the master plan's
  PR 12) — held until benchmark evidence on the vendored fixtures.
- Postsolve recovery for fixed variables at original bounds (v1
  assumes interior).

## How to review

```sh
git fetch origin claude/sleepy-shannon-Jvste
git checkout claude/sleepy-shannon-Jvste
cargo test -p pounce-presolve
cargo run -p pounce-presolve --example pipeline_demo
cargo run -p pounce-presolve --example phase0_via_tnlp
```

For a real-world check, run the binary on any `.nl` with and without
the flag:

```sh
./target/release/pounce path/to/problem.nl \
    presolve=yes \
    presolve_auxiliary=yes \
    presolve_auxiliary_diagnostics=yes
```

Stderr should carry a one-line diagnostics summary; the solve
should converge to the same objective as without the flag.

## Known v1 limitations

Documented at `docs/src/auxiliary-presolve.md`:

- Fixed variables are assumed interior to their original bounds at
  the optimum.
- The pass runs once at start; no iterative re-elimination.
- Inequality-coupled blocks are never eliminated under any policy.
- Linear blocks always use the pre-fetched Jacobian; nonlinear
  blocks use TNLP callbacks.

Closes #53.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X3DVnKNH8emugvFuDZ7eFR)_


jkitchin (commented):
## Review summary

Functionality lands the design and most acceptance criteria for #53, but there is **one merge-blocking correctness bug** that crashes the headline benchmark fixture (`gaslib11_steady.nl`) on the PR's own defaults. The bug is interaction between the new auxiliary phase and the pre-existing `bound_tighten` phase; both endpoints need a small fix. Several lower-priority issues are noted below.

Fixtures from the issue acceptance list (`tutorial_flow_density.nl`, `tutorial_flow_density_perturbed.nl`, `gaslib11_steady.nl`) were not bundled with the PR. Vendored locally from upstream `jkitchin/ripopt` (EPL-2.0, same as pounce) for testing — pushed in a follow-up if useful.

### Acceptance-criteria validation (with vendored fixtures, release build)

| Fixture | `presolve_auxiliary=no` | `presolve_auxiliary=yes` (defaults) | Verdict |
|---|---|---|---|
| `tutorial_flow_density.nl` | 6 IPM iters, obj 0.0 | **0 IPM iters**, obj 0.0 | ✅ |
| `tutorial_flow_density_perturbed.nl` | 7 IPM iters, obj 0.0 | **0 IPM iters**, obj 0.0 | ✅ |
| `gaslib11_steady.nl` | 18 iters, obj 3.286e-02 | **`Invalid Problem Definition`**, 0 iters | ❌ |

With `presolve_bound_tightening=no` added to the aux-on gaslib11 run, the IPM solves cleanly (17 iters, obj 1.825e-02 — a different local optimum than the un-presolved path, which is also worth investigating). This isolates the bug to the bound-tighten interaction described below.

---

## 🔴 Blocker: aux + bound_tighten produce `x_l > x_u` and crash the IPM

**Path to repro** (release binary):
```
pounce gaslib11_steady.nl presolve=yes presolve_auxiliary=yes
→ EXIT: Invalid Problem Definition

pounce gaslib11_steady.nl presolve=yes presolve_auxiliary=yes presolve_bound_tightening=no
→ EXIT: Optimal Solution Found.
```

**Root cause** (two cooperating defects, both introduced by this PR):

1. **`crates/pounce-presolve/src/block_solve.rs`** — the new Newton block solver has no notion of variable bounds. `BlockEquations` (line 71) carries only `dim`/`eval`/`jacobian`; the solver converges to a Newton root and returns it regardless of `[x_l, x_u]`. On gaslib11, this lands `fixed_values[k]` slightly outside (or numerically inconsistent with) other rows' bounds.

2. **`crates/pounce-presolve/src/lib.rs:438-470`** — the orchestrator clamps
   ```rust
   x_l[i] = frame.fixed_values[k];
   x_u[i] = frame.fixed_values[k];
   ```
   then immediately calls `tighten_bounds` over **all** original `linear_rows`, including rows aux just marked dropped in `row_kept_inner`. Tighten propagates the now-clamped values through those linear rows, derives `x_l[j] = v + ε` for some aux-fixed `x_j` whose `x_u[j] = v`, sets `report.infeasible = true` (`bound_tighten.rs:128`), and returns. **That flag is never inspected** — `grep '.infeasible' crates/pounce-presolve/src/lib.rs` returns nothing. The corrupted `x_l > x_u` is handed to the IPM, `TNLPAdapter::new_with_options` returns `INCONSISTENT_BOUNDS` (`crates/pounce-nlp/src/tnlp_adapter.rs:169-285`), and the application surfaces it as `ApplicationReturnStatus::InvalidProblemDefinition` (`crates/pounce-algorithm/src/application.rs:980`).

**Suggested fix** (any of these alone unblocks; doing 1 + 3 is the principled set):

1. Filter `linear_rows` by `row_kept_inner` before passing to `tighten_bounds`, so aux-absorbed rows don't propagate anymore.
2. Check `tighten_report.infeasible` after the call; on `true`, log and disable presolve for this solve rather than handing corrupted bounds to the IPM.
3. Thread `(x_l, x_u)` into `BlockSolveOptions` and reject any solution outside the box with a new `BlockSolveError::OutOfBounds` variant; convert that into an `AuxiliaryRejectionReason` in the orchestrator so the block falls back instead of fixing an out-of-bounds value.

I have a local reproducer ready (toggle `presolve_bound_tightening` on gaslib11_steady.nl) — happy to push a regression test under the new fixtures directory once one of these fixes lands.

---

## 🟡 Secondary: integration tests are silent no-ops

`crates/pounce-cli/tests/auxiliary_presolve.rs` runs against `tests/fixtures/parametric.nl`, which carries sensitivity suffixes. `crates/pounce-cli/src/main.rs:306-312` **silently disables presolve** in that case:
```rust
if (sens_active || args.compute_red_hessian) && presolve_opts.enabled {
    eprintln!("pounce: disabling presolve — ...");
    presolve_opts.enabled = false;
}
```
Consequences:
- `presolve_auxiliary_yes_does_not_panic` passes trivially — the code path is never executed.
- `presolve_auxiliary_diagnostics_yes_emits_summary` **should fail** because the diagnostics `eprintln!` at `crates/pounce-presolve/src/lib.rs:452-454` never runs. Worth confirming with `cargo test -p pounce-cli auxiliary_presolve -- --nocapture` why CI is green here.
- `presolve_auxiliary_no_unchanged_from_baseline` reduces to "the solver is deterministic with identical CLI args" — does not validate the aux-no-op claim.

Re-targeting the integration tests at one of the three vendored aux fixtures fixes all three at once and exercises the headline acceptance criteria from #53.

---

## 🟡 Numerical: `block_solve.rs` doesn't honor the issue's acceptance criteria

The issue's acceptance list calls out `rcond_threshold`, `residual_tol`, `max_iter`. Of those:
- `max_iter` — respected (`block_solve.rs:198`).
- `residual_tol` — respected (`block_solve.rs:199`).
- **`rcond_threshold` — not implemented.** The LU only has an **absolute** pivot floor of `1e-14` (`block_solve.rs:278`); no `rcond` is computed (no 1-norm of `A`, no `A^{-1}` probe). A block scaled by `1e-10` will spuriously be called singular; a true singularity at scale `1e+10` passes through.
- Newton line search uses bare residual reduction (`res_new < res`, line 228), not Armijo / sufficient decrease — fine for the small blocks here but worth a follow-up.

---

## 🟡 Numerical: Hopcroft-Karp BFS missing the `dist[NIL]` sentinel

`crates/pounce-presolve/src/matching.rs:109-162` runs the BFS/DFS but never records `dist[NIL]` and never has DFS bail past the first augmenting layer. **Correctness is preserved** (each round still augments to maximum and the König-fuzz test confirms it), but the per-phase work loses the `O(√V)` round bound — on a pathological 100k-vertex bipartite graph you can do `O(V)` BFS phases instead of `O(√V)`. Fix is small (track `dist_nil`, gate `next == NIL || dist[next] == dist[r]+1`). Defer if you'd like, but worth a tracking issue.

---

## 🟡 Orchestrator perf — O(nnz × blocks) inner loops

`crates/pounce-presolve/src/auxiliary.rs`:
- `solve_linear_block` (lines 400-429) and `residual_norm_linear` (line 590) walk the **entire global** `jac_irow/jac_jcol` arrays inside the per-block loop. For N nnz, B blocks of size k: O(B·k·N). Build a per-row CSR index once.
- `block_cols.iter().position(...)` at line 421 (and 540, 548) is a linear scan over `block_cols` for every Jacobian nonzero touching the row. `HashMap<col, jj>` built once per block.

Acceptable for the gaslib-sized fixtures, problematic on 100k-var problems.

---

## 🟡 Orchestrator correctness — `x_running` not threaded into nonlinear Newton's starting point

`crates/pounce-presolve/src/auxiliary.rs:280` and the nonlinear-block path at `:574` seed the Newton iterate from `probe.x_probe[c]` only. `x_running` updates from prior accepted blocks in the same pass are not seen by the starting iterate (though they *are* seen by the residual evaluator). For a chain of dependent blocks this can converge Newton to a fixed point inconsistent with previously-accepted blocks.

---

## 🟢 Minor (non-blocking)

- **`reduction_frame.rs:88`** — `let _ = fixed_set.split_off(&0);` discards its result then drops `fixed_set` immediately. Dead code, likely leftover.
- **`reduction_frame.rs:237-243`** — recovery RHS materializes a dense `n_rows × n_vars` Jacobian. On a 10k-var problem this is 100M entries. Accept a sparse triplet or document the contract.
- **`reduction_frame.rs:67, 78, 224`** — `BTreeSet::contains` inside `n_vars` loop is O(log k) where `var_map[i].is_none()` would be O(1) and uses an array already in scope.
- **`reduction_frame.rs`** — `project_lambda`/`lift_lambda` round-trip is not tested as a pair; only one-direction tests at lines 336-347. No multi-frame `ReductionStack` round-trip test either.
- **`btf.rs:99, 107, 137`** — three `.expect("matched")` panics guard a cross-module invariant. The public `SquareComponent` ctor (`components.rs:14`) allows constructing one with unmatched rows. Either make the ctor private or document the invariant on the struct.
- **`matching.rs:14-15`** — `NIL` and `INF` both alias `usize::MAX` for different semantic roles; readability nit.
- **`incidence.rs:106-119`** — `(jac_irow[k] as isize - 1) as usize` is hardened by a downstream bound check (`i >= p.m_rows`). Worth a `debug_assert!(jac_irow[k] != 0)` for the 1-based path so corruption fails loud during dev.
- **`auxiliary.rs:170`** — `x_running` silently accepts NaN/inf from the probe. Newton seeded with NaN returns NaN, residual NaN > tol is `false`, block "succeeds." Add a finite check.
- **`diagnostics.rs`** — `candidate_blocks == blocks_eliminated + rejection_reasons.len()` is an invariant that's not asserted; `Display` prints "pure=N" for *candidates*, not eliminated. Easy to misread. Either rename the field or split into `_candidates`/`_eliminated`.
- **Docs gap** — `docs/src/auxiliary-presolve.md` doesn't mention (a) that sensitivity / reduced-Hessian post-processing silently disables presolve, or (b) the auxiliary → bound_tighten ordering and its current interaction footgun.

---

## What I checked locally

- `cargo test -p pounce-presolve` — green
- `cargo test -p pounce-cli auxiliary_presolve` — green (but see the no-op caveat above)
- Manual runs of all three #53 acceptance fixtures, both with and without aux, with and without bound_tighten
- Read auxiliary.rs, reduction_frame.rs, block_solve.rs, coupling.rs, diagnostics.rs, the graph layer (incidence/matching/dulmage_mendelsohn/components/btf), bound_tighten.rs, the presolve orchestrator (lib.rs phases 0–4), and the docs/options.md presolve_auxiliary_* entries.

Strong work on a sizeable refactor. The graph and frame layers in particular look thoughtful and well-tested. Once the bound-tighten interaction is resolved, this is ready to land.

jkitchin:
Thanks for the thorough review — that's a real bug. Pushed `fe818f6` addressing the blocker and the secondary issues I could land cheaply. The deferred items are spelled out in the commit message and below.

## Blocker — fixed in `fe818f6`

Three-pronged fix as you suggested, with #1 + #2 in this commit (defence in depth):

**(1) Filter `linear_rows` by `row_kept_inner` before `tighten_bounds`.** The root cause: `linear_rows` was materialised before Phase 0 ran, so it still contained the rows aux just dropped. Now it's built **after** Phase 0, filtered to kept rows only. Dropped rows can no longer propagate bounds.

**(2) Check `tighten_report.infeasible`.** If it still trips after the filter (genuine inconsistency between aux's solution and a kept row's bounds), the orchestrator rolls back the aux pass: restores `x_l`/`x_u` from `inner_x_l`/`inner_x_u`, resets `row_kept_inner`, clears the reduction stack, re-runs `tighten_bounds` on the unfiltered rows. A one-line warning lands on stderr.

**(3) Bound-aware `BlockSolveOptions` + `BlockSolveError::OutOfBounds`** — **deferred** as a follow-up. The structural fix in #1 + the rollback in #2 already prevents the crash; bound-aware Newton would be defence-in-depth that catches the same case one step earlier in the pipeline.

New regression test `phase0_via_tnlp_no_infeasible_with_default_bound_tightening` confirms the wrapper init no longer trips `tighten_report.infeasible` on the configuration that crashed. Happy to take your local `gaslib11_steady.nl` reproducer once the fixtures land — the test under `tests/fixtures/aux_presolve/` is the natural home.

## Secondary issues — also addressed

- **CLI integration tests were silent no-ops.** Rewritten with explicit acknowledgement. The first test now asserts the documented `"disabling presolve"` warning lands on stderr — proves the sens-disable code path is genuinely executed. The diagnostics-emit test was conceptually impossible on `parametric.nl` (presolve never runs); replaced with a baseline-objective check.
- **Docs gap.** `docs/src/auxiliary-presolve.md` gains two new sections: (a) interaction with `bound_tighten` (ordering, filter, rollback) and (b) the silent disable when sensitivity / reduced-Hessian is active.
- **`reduction_frame.rs` BTreeSet-in-loop + dead split_off.** Replaced with flat `bool` arrays for O(1); the dead `let _ = fixed_set.split_off(&0);` is gone. `recover_dropped_multipliers` now uses `self.row_map[r].is_none()` directly.
- **`auxiliary.rs` NaN/inf gate.** Added a finite-check on `probe.x_probe` at orchestrator entry. Newton seeded with NaN returns NaN, and `NaN > tol` is `false` — would have let a block spuriously succeed.
- **Nonlinear Newton x0 not threaded from x_running.** Fixed — nonlinear blocks now seed from `x_running[c]` instead of `probe.x_probe[c]`, so a chain of dependent blocks within one pass picks up earlier fixed values.
- **Diagnostics `Display` ambiguity.** The line previously read `coupling: pure=N` but the counts are over candidates, not blocks eliminated. Renamed to `candidates by coupling class: pure=N` so it can't be misread.

## Deferred to follow-up issues (not in this commit, documented in the message)

- Bound-aware `BlockSolveOptions` + `BlockSolveError::OutOfBounds` (your defence-in-depth #3).
- `rcond_threshold` in dense LU.
- Hopcroft-Karp `dist[NIL]` sentinel.
- Per-block CSR index in the orchestrator inner loop.
- Reduction-frame sparse RHS for recovery.
- `project_lambda` / `lift_lambda` paired round-trip test + multi-frame `ReductionStack` round-trip.
- `btf.rs` `.expect("matched")` invariant: either tighten the `SquareComponent` ctor or document on the struct.
- `matching.rs` `NIL` vs `INF` naming.
- `incidence.rs` debug_assert for 1-based path.
- `diagnostics.rs` invariant assertion (`candidate_blocks == blocks_eliminated + rejection_reasons.len()`).

Happy to land any subset of those in this PR if you want them before merge — none are large. Otherwise I'll open follow-up issues against #53 for tracking.

Tests: 120 unit + 1 integration + 9 doctests, all green.

---

jkitchin:
Landed the cheap follow-ups in `21f38fe`. Five review nits + a docs note on the gaslib11 local-optimum question. Everything else from your review stays deferred — the remaining items genuinely need their own scoping (cross-crate IPM `LargeBlockSolver`, postsolve for fixed-at-bound vars, general FM with row synthesis, iterative re-elim, rcond, CSR perf).

## In this commit

1. **Hopcroft-Karp `dist[NIL]` sentinel.** `bfs` tracks `dist_nil` and prunes past the shortest aug-path layer. Recovers the O(√V) phase bound.

2. **Paired round-trip + multi-frame stack tests.** `frame_project_lift_lambda_roundtrip` and `reduction_stack_multi_frame_roundtrip`. No production code change; closes the test-coverage gap you noted.

3. **Bound-aware `BlockSolveOptions` + `BlockSolveError::OutOfBounds`.** Defence-in-depth #3. `BlockSolveOptions` gains `bounds: Option<(Vec<Number>, Vec<Number>)>`. Newton checks at both success-return paths with `tol` slop. New `AuxiliaryRejectionReason::OutOfBounds` for orchestrator-level diagnostics. Two new tests: `newton_out_of_bounds_rejects` and `newton_in_bounds_accepts`.

4. **`SquareComponent` matched-rows invariant.** Struct-level doc + `debug_assert!` in `BlockTriangularForm::of_component`. The downstream `.expect("matched")` panics are still there but the boundary check fires first with a clearer message.

5. **gaslib11 documented, not fixed.** Added a "Caveat: nonconvex problems" section to `docs/src/auxiliary-presolve.md`. The honest read is: aux pins variables before the IPM gets to pick a basin; on nonconvex problems with multiple feasible solutions to the equality system, aux can land in a different basin than where the un-presolved path settles. Both points are KKT-feasible. The e2e regression test already deliberately doesn't assert objective parity for this fixture; the docs now explain why. A multiple-basin-aware policy is a v2 item; flipping the default (PR 12) should wait for it.

## Test counts

138 unit + 1 integration + 9 doctests, all green (was 134 + 1 + 9).

## Still deferred (tracked on the issue)

- `pounce-algorithm`-side `LargeBlockSolver` IPM impl (cross-crate dep)
- Postsolve for fixed variables at original bounds (moderate scope, deserves its own commit + tests)
- Iterative re-elimination (significant feature)
- General Fourier-Motzkin with row synthesis on the IPM-visible problem (real architecture change)
- `rcond_threshold` in dense LU (algorithmic depth)
- Per-block CSR index in the orchestrator inner loop (modest refactor; would matter on 100k-var problems but doesn't move the current fixtures)
- Reduction-frame sparse RHS for `recover_dropped_multipliers`
- gaslib11 multiple-basin-aware policy

---

PR #94, opened 2026-06-02
Add Gaussian-hump basin-flooding multi-minima example

jkitchin:
Demonstrates the MD-style idea of escaping a found minimum by adding a
repulsive Gaussian bump to the objective and re-solving, so successive
solves discover distinct minima. Built on pounce.minimize with analytic
hump gradient/Hessian; bounds and constraints pass through untouched.
Includes saddle rejection via the Hessian and global box restarts, and a
six-hump camel demo that recovers all six minima.


jkitchin (commented):
## Review — Gaussian-hump basin-flooding multi-minima

**Overall: strong, well-tested, well-documented PR. Core math is verified correct. The remaining work is edge-case robustness polish, not redesign — no blockers.**

### What's solid ✅
- **Kernel math verified correct** — the Gaussian and pole repulsion value/gradient/Hessian match finite differences to ~1e-11.
- **Tests pass** — 26/26 across `test_minima`, `test_critical`, `test_minima_scaling`.
- **Examples work** — `gaussian_hump_minima.py` recovers all 6 six-hump-camel minima incl. the global `f = -1.031628`; all 3 examples run clean.
- Clean public API (`find_minima`, `find_critical_points`, `find_saddles`, `reaction_network` + result dataclasses), 3 notebooks, docs pages, research note.

### Should-fix (robustness)

**1. No NaN/Inf guards anywhere** (`_minima.py` `consider`, `_critical.py` throughout).
If the solver reports `success` at a point where `fun` returns NaN (realistic with restricted-domain functions — the FBBT scenario), it is accepted. Worse, the dedup distance to a NaN point is NaN, which never compares `<= dedup`, so it is never "known" — **duplicate NaN "minima" can fill the archive**. In `_critical.py`, NaN grad/Hess flows straight into `eigh`/`eigvalsh` → `LinAlgError` aborts the whole enumeration. Add `np.isfinite` guards on both the candidate and its function value before acceptance / before any eigendecomposition.

**2. Dedup metric is inconsistent between routes.**
`find_minima` uses a bounds-**scaled** distance (`norm((a-b)/L)`, documented "scale-free"); `find_saddles` / Route-A use **raw** Euclidean (`_critical.py:252`, `known = any(np.linalg.norm(x - p.x) <= dedup ...)`). With the same default `dedup=1e-3` these mean two different things, and the saddle route breaks on non-unit-scale problems (large variable ranges merge distinct saddles, or vice-versa). Unify on the scaled metric (thread `L` into `_critical`).

**3. `find_saddles` doesn't enforce bounds during the eigenvector-following walk** (`_eigvec_follow`).
It only discards out-of-bounds points *after* the walk, wasting solves and potentially missing near-boundary saddles. It also calls `_morse_index` (an eigendecomposition) *before* checking `gn <= grad_tol`, so a diverged/NaN iterate triggers an `eigh` that can raise. Clip inside the walk and short-circuit on grad-norm first.

**4. Degenerate-eigenvalue classification gap** (`_critical.py:90-93`, `:59`).
The Morse index counts `eig < -eig_tol`; eigenvalues in `[-eig_tol, +eig_tol]` (numerically zero — a degenerate/non-Morse point) are treated as positive, so a degenerate point or monkey-saddle is silently mislabeled a minimum/saddle. Add a "degenerate" flag/status.

### Nits
- **Dead code**: `grad = grad; hess = hess` (`_critical.py:132-133`) — no-op, looks like a refactor leftover. `sobol`/`_make_sobol` computed but unused in `_run_tunneling` (`_minima.py`).
- `reaction_network` floors saddle `grad_tol` at `1e-5` even if the caller passes a tighter value (`:433`), undocumented.
- MLSL radius is `0.0` when `samples_per_round=1` (`log(1)=0`); guard `N >= 2`.
- No `i == j` guard in `reaction_network` → a basin can self-connect.
- `psd_tol` saddle test is absolute (scale-sensitive) while dedup is scale-free — worth documenting the asymmetry.

### UX wart (pre-existing, not this PR)
Every unconstrained solve logs `ERROR pounce-py: jacobian(): AttributeError …`. This reproduces with a plain `pounce.minimize(lambda x: x@x, ...)`, so it's a `pounce.minimize` issue, not this PR — and the module docstring already acknowledges it. But this PR makes it *loud* (dozens of ERROR lines per run), which is alarming for an otherwise-successful global search. Worth a follow-up to downgrade that log to debug for the no-constraint case.

---
*Verified locally: 26 tests pass, all 3 examples run, six-hump camel recovers all 6 minima, kernel derivatives FD-checked to ~1e-11.*

PR #97, opened 2026-06-02
Increase scipy minimize compatibility

jduerholt:
Hi,

when starting to play around with the repo and trying to use it as drop-in replacement for `scipy.optimize.minimize`, I stumbled about two things:

1. `scipy.optimize.minimize` offer an `args` argument which are arguments that are passed to the objective function on optimization. The pounce interface is currently not supporting this. 
2. `scipy.optimize.minimize` offer the option to set `jac=True`, in this case it expects that `fun` returns both the function value and its gradient. This is especially useful when computing the gradient via autodiff as then only one forward pass is needed. 

This PR adds compatibility for the `args` argument and for the `jac` as boolean to incease scipy compatibility. This PR is kind of a draft and needs further polishing and testing etc. It should serve as a proposal, if you find it useful, I am happy to improve it so that one can get it merged into `pounce`.

Best,

Johannes


jkitchin:
This is a good idea. It looks like maybe args should go to the hessian, and maybe constraints?  I can merge this and add those easy enough I think if you want.

jduerholt:
Good point, I think args should also go to hessian. For constraints, I am not sure. I (or Claude) have to lookup the exact scipy api. I will also add the callback functionality from scipy. These are all features that are used in BoTorch and I want to test how POUNCE perform as drop-in replacement for SLSQP for ACQF optimization in the presence of linear constraints.

jduerholt:
I just converted to draft, I want to teach the scipy interface a few more things, I tell you when it is ready for review ;)

jkitchin:
I will probably work on a new release this weekend. Do you think you will finish it by then?

jduerholt:
> I will probably work on a new release this weekend. Do you think you will finish it by then?

Sorry, I totally overlooked this. I try to get it finished this week. Maybe we can get it then into the next release.

jduerholt:
Hi @jkitchin,

what do you think about these changes? 

I added the following things to the scipy interface:

- support for sparse linear constraints 
- support for callbacks
- support for scipy.optimze.Bounds
- support for scipy specifc keywords including translations

If this is in principle fine for you, I would solve the merge conflicts with main and implement the missing tests.

Best,

Johannes

jkitchin:
These look fine to me!

jduerholt:
@jkitchin I have some questions regarding the routing and probing introduced recently for the merge:

`caf6bcb` introduced `_route.classify_and_extract`, which probes `fun` / `jac` /
`hess` (and `jac_combined` for constraints) at 5+ points on every
`pounce.minimize` call to detect whether the problem is actually a convex LP/QP
that should be dispatched to `solve_qp`. The probe is run **by default**
(`solver_selection="auto"`), with the legacy NLP path as the fallback.

If you know that the objective is NLP,  this probing is pure overhead with zero possibility of
benefit as we introduced in this PR the option to ingest `scipy.optimize.LinearConstraint`
classes which declares a constraint as linear, so we do not need to probe. 

I propose to  **flip the routing default to opt-in.** 

Something like:

```python
# Current main: auto-route runs the probe on every call.
res = pounce.minimize(f, x0, jac=grad, constraints=lc)
#  → probes fun/jac/hess → falls through to NLP for non-QP problems
#  → ~5 extra evals per call wasted

# Proposed: explicit opt-in to the convex IPM, NLP otherwise.
res = pounce.minimize(f, x0, jac=grad, constraints=lc)
#  → straight to NLP, no probe

res = pounce.minimize(f, x0, jac=grad, constraints=lc,
                      solver_selection="qp-ipm")
#  → forces the convex IPM (current main: still validates via probe)

res = pounce.minimize(f, x0, jac=grad, constraints=lc,
                      solver_selection="auto")
#  → preserves today's UX for users who want it
```

What do you think?

jkitchin:
That sounds reasonable to me.

jduerholt:
Ok. Perfect. I will finish it tmr ;)

jkitchin:
Does that mean it is working as a drop-in for your application?

jduerholt:
Yes, works as drop-in  replacement.

Application is acquistion function optimization in BO using BoFire (https://github.com/experimental-design/bofire/tree/main) and BoTorch (https://github.com/meta-pytorch/botorch/tree/main). I have also a seperate PR open in BoTorch implementing sparse representations of the constraints: https://github.com/meta-pytorch/botorch/pull/3320

It is now ready for review ;)

jduerholt:
@jkitchin: should i resolve the new merge conflicts after your latest merge? btw, I see really nice results with pounce ;)

jkitchin:
I can take care of it if it's ready to review and merge.

jduerholt:
@jkitchin: it is now ready from my side. Merge conflicts are resolved and I added even a new feature namely hessian support in the presence of **only** linear constraints, as the lagrange hessian is then just the pure objective hessian. 

i am looking forward to your opinion on this PR.

jkitchin:
Let me know if the commits above look ok to you, then I think we can merge it.

jduerholt:
Hi @jkitchin, thanks, looks good to me. This ready to go from my side. I really like the new compatibility stuff here ;) Best, Johannes

PR #150, opened 2026-06-19
Add boundary value problem solver with differentiable JAX/PyTorch frontends

jkitchin:
## Summary

Adds a complete boundary value problem (BVP) solver suite to pounce:

- **`pounce.solve_bvp`**: A drop-in replacement for `scipy.integrate.solve_bvp` that discretises two-point BVPs with 4th-order Hermite–Simpson collocation on a fixed mesh and solves the resulting square root-find as a pounce feasibility NLP.
- **`pounce.solve_bvp_constrained`**: A pounce-unique capability for constrained BVPs with state bounds, inequality path constraints, and optional objectives.
- **`pounce.jax.solve_bvp` and `pounce.torch.solve_bvp`**: Differentiable frontends that make the converged solution `z*(θ)` differentiable with respect to any parameter `θ` baked into the ODE or boundary conditions, via the implicit-function theorem on the collocation KKT system.

### Core BVP infrastructure (`python/pounce/bvp/`)
- **`_core.py`**: Backend-agnostic collocation residual computation (Hermite–Simpson scheme) shared across NumPy, JAX, and PyTorch paths.
- **`_jac.py`**: Exact sparse Jacobian of the collocation residual. Per-node blocks are obtained from user-supplied `fun_jac`/`bc_jac` or via vectorised finite differences (`O(n)` evaluations, not `O(n·m)`).
- **`_newton.py`**: Damped-Newton root-finder on FERAL's unsymmetric sparse LU for the square collocation system (the fast path, no interior-point overhead).
- **`_solve.py`**: SciPy-compatible `solve_bvp` entry point; poses the root-find as a pounce feasibility NLP (`min 0` s.t. `R(z) = 0`).
- **`_constrained.py`**: Constrained BVP solver with state bounds and path constraints, routed through pounce's interior-point method.

### Differentiable frontends
- **`python/pounce/jax/_bvp.py`**: JAX frontend using `jax.custom_vjp` and `jax.pure_callback` for the Newton forward pass, with implicit-function-theorem backward via the collocation KKT system.
- **`python/pounce/torch/_bvp.py`**: PyTorch eager-mode mirror using `torch.autograd.Function` for the same implicit-differentiation strategy.

### Rust sparse linear algebra
- **`crates/pounce-py/src/sparse_lu.rs`**: PyO3 binding for FERAL's unsymmetric sparse LU (`SparseLU` class), enabling direct Newton iteration without the interior-point method's symmetric saddle system.

### Tests and examples
- **`python/tests/test_bvp.py`**: Comprehensive test suite covering SciPy compatibility, unknown parameters, solver method agreement, analytic vs. finite-difference Jacobians, constrained BVPs, and differentiability (JAX/PyTorch gradients vs. finite differences).
- **`python/examples/bvp_scipy_compare.py`**: Accuracy/speed benchmarks against SciPy and differentiability demonstrations (gradients, Jacobians, Hessians, second-order derivatives).
- **`docs/src/bvp.md`**: User-facing documentation with examples and API reference.

## Notable Implementation Details

1. **Fixed mesh by design**: Unlike SciPy's adaptive refinement, the mesh is fixed. This keeps the solution map `θ ↦ y` smooth, which is essential for differentiability via the implicit-function theorem.

2. **Sparse collocation Jacobian**: The global Jacobian is sparse (O(m) nonzeros for m mesh nodes) because each collocation block couples only two adjacent nodes. Per-node `∂f/∂y` blocks are computed via vectorised finite differences that perturb

https://claude.ai/code/session_01CkewNiehrg7D1sf5SPRfC3

### DIFF
--- a/crates/pounce-py/src/sparse_lu.rs
+++ b/crates/pounce-py/src/sparse_lu.rs
@@ -143,4 +143,68 @@ impl PySparseLu {
         })?;
         Ok(rhs.into_pyarray_bound(py))
     }
+
+    /// Solve `A X = B` for `n_rhs` right-hand sides against the single held
+    /// factor. ``b_flat`` is the row-major flatten of the ``(n_rhs, n)``
+    /// RHS block; the result is the matching flat ``(n_rhs, n)`` solution.
+    /// One factorization, many back-solves — the multi-RHS path the
+    /// implicit-diff Jacobian (``jax.jacobian`` / batched VJP) needs.
+    fn solve_many<'py>(
+        &mut self,
+        py: Python<'py>,
+        b_flat: PyReadonlyArray1<f64>,
+        n_rhs: usize,
+    ) -> PyResult<Bound<'py, numpy::PyArray1<f64>>> {
+        self.multi(py, b_flat, n_rhs, false)
+    }
+
+    /// Like :meth:`solve_many` but solves `Aᵀ X = B` (the VJP direction).
+    fn solve_transpose_many<'py>(
+        &mut self,
+        py: Python<'py>,
+        b_flat: PyReadonlyArray1<f64>,
+        n_rhs: usize,
+    ) -> PyResult<Bound<'py, numpy::PyArray1<f64>>> {
+        self.multi(py, b_flat, n_rhs, true)
+    }
+}
+
+impl PySparseLu {
+    fn multi<'py>(
+        &mut self,
+        py: Python<'py>,
+        b_flat: PyReadonlyArray1<f64>,
+        n_rhs: usize,
+        transpose: bool,
+    ) -> PyResult<Bound<'py, numpy::PyArray1<f64>>> {
+        let lu = self
+            .lu
+            .as_mut()
+            .ok_or_else(|| PyRuntimeError::new_err("SparseLU: call factor() first"))?;
+        let nn = lu.dim();
+        let b = b_flat.as_slice()?;
+        if b.len() != n_rhs * nn {
+            return Err(PyValueError::new_err(format!(
+                "SparseLU multi-solve: rhs length {} != n_rhs ({}) * n ({}) = {}",
+                b.len(),
+                n_rhs,
+                nn,
+                n_rhs * nn
+            )));
+        }
+        let mut out = vec![0.0_f64; b.len()];
+        for r in 0..n_rhs {
+            let mut rhs = b[r * nn..(r + 1) * nn].to_vec();
+            let res = if transpose {
+                lu.btran(&mut rhs)
+            } else {
+                lu.ftran(&mut rhs)
+            };
+            res.map_err(|e| {
+                PyRuntimeError::new_err(format!("SparseLU multi-solve failed: {e:?}"))
+            })?;
+            out[r * nn..(r + 1) * nn].copy_from_slice(&rhs);
+        }
+        Ok(out.into_pyarray_bound(py))
+    }
 }
--- a/crates/pounce-py/src/sparse_lu.rs
+++ b/crates/pounce-py/src/sparse_lu.rs
@@ -1,0 +1,146 @@
+//! PyO3 binding for FERAL's unsymmetric sparse LU (general `A x = b`).
+//!
+//! POUNCE's NLP solver factorizes *symmetric* KKT saddle systems. A square
+//! root-find (e.g. a BVP collocation system) is better served by a direct
+//! Newton iteration on the unsymmetric `N x N` Jacobian — exactly what
+//! SciPy's `solve_bvp` does. FERAL (POUNCE's pure-Rust linear backend)
+//! ships an unsymmetric sparse LU; this class exposes it so the Python
+//! side can run that Newton loop without going through the interior-point
+//! method (and its `2N` saddle system).
+//!
+//! Usage: build once over a fixed sparsity pattern (`rows`, `cols` in
+//! 0-based COO), then `factor(values)` / `solve(b)` repeatedly as the
+//! Newton iterates change the values but not the structure. `solve` does
+//! `A x = b` (FERAL's `ftran`); `solve_transpose` does `Aᵀ x = b`
+//! (`btran`), which the implicit-differentiation backward needs.
+
+use feral::{LuParams, SparseColMatrix, SparseLu, SparseLuSymbolic};
+use numpy::{IntoPyArray, PyReadonlyArray1};
+use pyo3::exceptions::{PyRuntimeError, PyValueError};
+use pyo3::prelude::*;
+
+#[pyclass(name = "SparseLU", module = "pounce._pounce", unsendable)]
+pub struct PySparseLu {
+    n: usize,
+    nnz: usize,
+    /// Per-column list of `(row, index-into-values)` so a fresh value
+    /// vector is scattered into FERAL's column-sparse format in O(nnz).
+    col_entries: Vec<Vec<(usize, usize)>>,
+    symbolic: Option<SparseLuSymbolic>,
+    lu: Option<SparseLu>,
+}
+
+#[pymethods]
+impl PySparseLu {
+    /// Build a reusable LU over the fixed sparsity pattern of an `n x n`
+    /// matrix. `rows` / `cols` are 0-based COO coordinates; the matching
+    /// values are supplied later by :meth:`factor`.
+    #[new]
+    fn new(n: usize, rows: Vec<i64>, cols: Vec<i64>) -> PyResult<Self> {
+        if rows.len() != cols.len() {
+            return Err(PyValueError::new_err(
+                "SparseLU: rows and cols must have equal length",
+            ));
+        }
+        let mut col_entries: Vec<Vec<(usize, usize)>> = vec![Vec::new(); n];
+        for (t, (&r, &c)) in rows.iter().zip(cols.iter()).enumerate() {
+            if r < 0 || c < 0 || (r as usize) >= n || (c as usize) >= n {
+                return Err(PyValueError::new_err(format!(
+                    "SparseLU: entry ({r}, {c}) out of range for n={n}"
+                )));
+            }
+            col_entries[c as usize].push((r as usize, t));
+        }
+        Ok(Self {
+            n,
+            nnz: rows.len(),
+            col_entries,
+            symbolic: None,
+            lu: None,
+        })
+    }
+
+    /// Numerically (re)factor `A` from `values` aligned to the
+    /// `(rows, cols)` pattern. The symbolic analysis is computed once and
+    /// reused across calls (the pattern is fixed), so repeated Newton
+    /// factorizations pay only the numeric cost. Raises ``RuntimeError``
+    /// if the matrix is singular.
+    fn factor(&mut self, values: PyReadonlyArray1<f64>) -> PyResult<()> {
+        let v = values.as_slice()?;
+        if v.len() != self.nnz {
+            return Err(PyValueError::new_err(format!(
+                "SparseLU.factor: values has length {} but the pattern has {} nonzeros",
+                v.len(),
+                self.nnz
+            )));
+        }
+        let mut cols: Vec<Vec<(usize, f64)>> = Vec::with_capacity(self.n);
+        for c in 0..self.n {
+            let entries = &self.col_entries[c];
+            let mut col = Vec::with_capacity(entries.len());
+            for &(r, t) in entries {
+                col.push((r, v[t]));
+            }
+            cols.push(col);
+        }
+        let m = SparseColMatrix::from_sparse_columns(self.n, &cols)
+            .map_err(|e| PyValueError::new_err(format!("SparseLU: bad matrix: {e:?}")))?;
+        if self.symbolic.is_none() {
+            let s = SparseLuSymbolic::analyze(&m)
+                .map_err(|e| PyValueError::new_err(format!("SparseLU: analyze failed: {e:?}")))?;
+            self.symbolic = Some(s);
+        }
+        let symbolic = self.symbolic.as_ref().unwrap();
+        let lu = SparseLu::factor(&m, symbolic, LuParams::default()).map_err(|e| {
+            PyRuntimeError::new_err(format!("SparseLU: factorization failed (singular?): {e:?}"))
+        })?;
+        self.lu = Some(lu);
+        Ok(())
+    }
+
+    /// Solve `A x = b`, returning `x`. Requires a prior :meth:`factor`.
+    fn solve<'py>(
+        &mut self,
+        py: Python<'py>,
+        b: PyReadonlyArray1<f64>,
+    ) -> PyResult<Bound<'py, numpy::PyArray1<f64>>> {
+        let lu = self
+            .lu
+            .as_mut()
+            .ok_or_else(|| PyRuntimeError::new_err("SparseLU.solve: call factor() first"))?;
+        let mut rhs = b.as_slice()?.to_vec();
+        if rhs.len() != lu.dim() {
+            return Err(PyValueError::new_err(format!(
+                "SparseLU.solve: rhs length {} != n {}",
+                rhs.len(),
+                lu.dim()
+            )));
+        }
+        lu.ftran(&mut rhs)
+            .map_err(|e| PyRuntimeError::new_err(format!("SparseLU.solve: ftran failed: {e:?}")))?;
+        Ok(rhs.into_pyarray_bound(py))
+    }
+
+    /// Solve `Aᵀ x = b`, returning `x`. Requires a prior :meth:`factor`.
+    fn solve_transpose<'py>(
+        &mut self,
+        py: Python<'py>,
+        b: PyReadonlyArray1<f64>,
+    ) -> PyResult<Bound<'py, numpy::PyArray1<f64>>> {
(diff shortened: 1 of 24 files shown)


python/pounce/bvp/_newton.py
```diff
@@ -0,0 +1,112 @@
...
+            if rnorm_trial < (1.0 - 1e-4 * alpha) * rnorm:
+                accepted = True
+                break
+            alpha *= 0.5
+
+        if not accepted:
+            if fresh:
+                # A fresh factor still can't reduce the residual — the iterate
+                # has converged to round-off (quadratic convergence stalls).
+                converged = rnorm < 1e-6
```
jkitchin: **Convergence declared at a hardcoded `1e-6`, ignoring the requested tolerance.**

When a fresh-factor step stalls, this sets `converged = rnorm < 1e-6`. But `_solve.py:277` calls `newton_solve` with `newton_tol = min(tol, 1e-10)`, and nothing downstream re-checks `rnorm` against `tol` (the post-loop check at line 110 only ever sets `converged=True`). A problem that stalls at `rnorm=1e-7` — four orders looser than requested — returns `success=True`/`status=0`. Tie the stall threshold to the caller's `tol`, or surface the achieved residual so `_solve.py` can judge.

crates/pounce-py/src/sparse_lu.rs
```diff
@@ -0,0 +1,210 @@
...
+        let m = SparseColMatrix::from_sparse_columns(self.n, &cols)
+            .map_err(|e| PyValueError::new_err(format!("SparseLU: bad matrix: {e:?}")))?;
+        if self.symbolic.is_none() {
+            let s = SparseLuSymbolic::analyze(&m)
+                .map_err(|e| PyValueError::new_err(format!("SparseLU: analyze failed: {e:?}")))?;
+            self.symbolic = Some(s);
+        }
+        let symbolic = self.symbolic.as_ref().unwrap();
+        let lu = SparseLu::factor(&m, symbolic, LuParams::default()).map_err(|e| {
+            PyRuntimeError::new_err(format!("SparseLU: factorization failed (singular?): {e:?}"))
```
jkitchin, line 95: **Singular Jacobian raises an uncaught `RuntimeError`, breaking the SciPy contract.**

`factor` raises `PyRuntimeError` on a singular matrix, and neither `_newton.py:73` nor `_solve.py` wraps the call. SciPy's `solve_bvp` never raises here — it returns a result with `success=False`, `status=2` ("singular Jacobian encountered"). Singular Jacobians are routine in BVPs (poor initial guess), so SciPy-ported code expecting a result bunch will crash instead. Catch this and map it to `status=2`.

python/pounce/bvp/_solve.py
```diff
@@ -0,0 +1,443 @@
...
+        cu = np.zeros(N, dtype=np.float64)
+        problem = Problem(n=N, m=N, problem_obj=obj, cl=cl, cu=cu)
+        problem.add_option("tol", float(tol))
+        # Collocation residuals are naturally well-scaled; skip the scaling
+        # pass (its setup cost buys nothing here).
+        problem.add_option("nlp_scaling_method", "none")
+        problem.add_option("print_level", 5 if verbose >= 2 else 0)
+        z_star, info = problem.solve(x0=z0)
+        niter = int(info.get("iter_count", 0))
+        status = 0 if info.get("status", 1) in (0, 1) else 1
```
jkitchin: **IPM "acceptable level" is reported as full success.**

This maps `info['status'] in (0, 1)` -> `status=0`/`success=True`. But pounce's return codes (`return_codes.rs:15-16`) define `1 = SolvedToAcceptableLevel` — the *looser* `acceptable_tol`, which `qp.rs:173-180` explicitly warns "callers that need full accuracy" must not treat as a clean solve. Compounding it, `rms_residuals` is computed only from the collocation block (`r_star[:n*(m-1)]`), so a violated boundary/path constraint never shows up in `success`. Same pattern at `_constrained.py:332`.

python/pounce/bvp/_solve.py
```diff
@@ -0,0 +1,443 @@
...
+        # mesh ``tol``: the residual estimate that drives adaptive refinement
+        # divides the collocation residual by the interval width, so a
+        # loosely-solved system reads as a huge residual on fine meshes.
+        # ``tol`` controls refinement, not the Newton stop.
+        newton_tol = min(float(tol), 1e-10)
+        z_star, niter, converged, _rn = newton_solve(
+            residual_fn, jac, z0, n, m, k, tol=newton_tol,
+        )
+        info = {}
+        status = 0 if converged else 1
```
jkitchin: **`status=1` on Newton non-convergence collides with SciPy's "max mesh nodes exceeded".**

SciPy reserves `status=1` for the node-budget overflow, `status=2` for a singular Jacobian, and `0` for converged. Since this module is documented as a drop-in replacement, SciPy-ported code branching `if res.status == 1: refine_more()` will misread a convergence failure as a node-limit failure. Use a distinct status code for Newton divergence.

python/pounce/bvp/_solve.py
```diff
@@ -0,0 +1,443 @@
...
+    x,
+    y,
+    p=None,
+    S=None,
+    fun_jac=None,
+    bc_jac=None,
+    tol=1e-3,
+    max_nodes=1000,
+    verbose=0,
+    bc_tol=None,
```
jkitchin, line 276: **`bc_tol` is accepted but never used.**

`grep` finds exactly one hit: this signature default. It is not forwarded to `_solve_bvp_adaptive`, never checked, and not in the docstring. A user porting SciPy code that passes `bc_tol=1e-8` to tighten boundary accuracy silently gets the default behavior. Either honor it (and emit `status=3` when unmet, per SciPy) or drop it and document the divergence.

python/pounce/jax/_bvp.py
```diff
@@ -0,0 +1,340 @@
...
+        ``fun(x, y, p, theta) -> (n, m)`` and
+        ``bc(ya, yb, p, theta) -> (n + k,)`` (drop ``p`` if there are no
+        unknown parameters). JAX-traceable.
+    x : array (m,)
+        Fixed mesh.
+    y : array (n, m)
+        Initial guess for the node states (not differentiated through).
+    p : array (k,) or None
+        Initial guess for unknown parameters, or ``None``.
+    theta : pytree
```
jkitchin: **Docstring advertises `theta` as a pytree, but the default Newton path only handles arrays.**

`_make_newton_vjp` computes `int(jnp.ndim(theta))` (line 255) and threads `theta` as a single positional callback arg. A dict `theta` is silently treated as rank-0 by `jnp.ndim`; a mixed tuple crashes. Only the non-default `ipm` path actually tolerates pytrees. Either narrow this docstring or flatten the pytree on the Newton path.

python/pounce/jax/_bvp.py
```diff
@@ -0,0 +1,340 @@
...
+        def residual_fn(z):
+            return _core.residual_of_z(z, nfun, nbc, x_np, n, m, k, np.concatenate)
+
+        jac = CollocationJacobian(nfun, nbc, x_np, n, m, k)
+        z_star, _it, _ok, _rn = newton_solve(
+            residual_fn, jac, z0_np, n, m, k, tol=float(tol),
+        )
+        return np.asarray(z_star, dtype=np.float64)
+
+    def _host_btran(z_star_np, theta_np, v_np):
```
jkitchin, line 103: **(efficiency) The IFT backward re-factorizes `R_z` from scratch, duplicating the forward solve.**

Each backward rebuilds `CollocationJacobian`, allocates a fresh `SparseLU`, and re-runs symbolic `analyze()` + numeric `factor()` at `z*` — the dominant per-solve cost — even though the forward Newton solve just factored the identical matrix at the same `z*` and discarded it. Thread the factored handle out of `newton_solve` and let the backward only `solve_transpose`. Same issue in `torch/_bvp.py` backward (line 101).

python/pounce/jax/_bvp.py
```diff
@@ -0,0 +1,340 @@
...
+        # R_z (N x N) and the directional derivative R_theta . theta_dot.
+        Rz = jax.jacfwd(lambda zz: g(zz, theta))(z)
+        _, R_theta_dot = jax.jvp(lambda th: g(z, th), (theta,), (theta_dot,))
+        z_dot = -jnp.linalg.solve(Rz, R_theta_dot)
+        return z, z_dot
+
+    return solve_root
+
+
+def _make_spline(x, y, yp):
```
jkitchin: **(reuse) `_make_spline` is copy-pasted three times, and the IFT backward is re-derived per frontend.**

`_make_spline` is near-verbatim here, in `torch/_bvp.py:200`, and in `_solve.py`. The whole Newton-IFT backward (factor `R_z` -> `solve_transpose` -> residual VJP) is also duplicated across `jax/_make_newton_vjp` and `torch/_newton_autograd_fn` — and they've already diverged (only JAX has the batched `solve_transpose_many` and a second-order route). The PR bills `_core.py` as the backend-agnostic core; the spline factory and IFT math belong there, leaving each frontend only its thin `custom_vjp`/`autograd.Function` shell.

python/pounce/jax/_bvp.py
```diff
@@ -0,0 +1,340 @@
...
+        return _pounce_solve(
+            theta, f=f, g=g, x0=z0, n=N, m=N, cl=cl, cu=cu, options=opts,
+        )
+
+    @solve_root.defjvp
+    def solve_root_jvp(primals, tangents):
+        (theta,), (theta_dot,) = primals, tangents
+        z = solve_root(theta)
+        # R_z (N x N) and the directional derivative R_theta . theta_dot.
+        Rz = jax.jacfwd(lambda zz: g(zz, theta))(z)
```
jkitchin, line 305: **(efficiency) Second-order path forms a dense `N x N` `R_z`.**

`jax.jacfwd` + `jnp.linalg.solve` here is `O(N^2)` memory / `O(N^3)` factor on what is an `O(m)`-nonzero banded collocation Jacobian — the dense cost the rest of the suite avoids. Reuse the sparse `CollocationJacobian` + FERAL `SparseLU` back-solve.

python/pounce/bvp/_constrained.py
```diff
@@ -0,0 +1,345 @@
...
+        # d c / d y_b at every node: (q, m) per perturbed state b.
+        dcdy = np.empty((q, m, n), dtype=np.float64)
+        for b in range(n):
+            step = _FD_EPS * np.maximum(1.0, np.abs(Y[b]))
+            Yb = Y.copy()
+            Yb[b] += step
+            cb = np.asarray(self._npath(x, Yb, p), dtype=np.float64)
+            dcdy[:, :, b] = (cb - c0) / step
+        # Structure order is (j, i, b): for each component j, node i, state b.
+        vals = [dcdy.transpose(0, 1, 2).reshape(-1)]
```
jkitchin: **(simplify) `dcdy.transpose(0, 1, 2)` is the identity permutation — a no-op.**

Just `vals = [dcdy.reshape(-1)]`.

python/pounce/jax/_bvp.py
```diff
@@ -0,0 +1,340 @@
...
+    ``(dR/dtheta)^T`` product is taken by JAX VJP over the traceable
+    residual at the converged ``z*``. Both directions stay on the ``N``
+    system — no ``2N`` saddle — so it is the fast path. First-order only
+    (the forward is an opaque callback); use ``method="ipm",
+    second_order=True`` for higher-order derivatives.
+    """
+    import numpy as np
+    from ..bvp._jac import CollocationJacobian
+    from ..bvp._newton import newton_solve
+    from ..bvp._solve import _BvpNlp  # noqa: F401  (kept for parity / reuse)
```
jkitchin: **(simplify) Dead import.**

`_BvpNlp` is never referenced on the Newton path; the `# noqa: F401 (kept for parity / reuse)` only exists to silence the linter for an unused symbol. Delete the line.

python/pounce/bvp/_newton.py
```diff
@@ -0,0 +1,112 @@
...
+        ratio = rnorm_trial / rnorm
+        z, R, rnorm = z_trial, R_trial, rnorm_trial
+        fresh = False
+        # Refactor next step if the frozen factor is going stale — slow
+        # reduction or a heavily damped step both signal the linearisation
+        # no longer matches; a fresh factor restores fast convergence.
+        if ratio > _REFACTOR_RATIO or alpha < 1.0:
+            need_factor = True
+
+    if rnorm < tol:
```
jkitchin: **(simplify) Redundant post-loop convergence check.**

This can only fire after `max_iter` exhaustion, where the top-of-loop test (line 67) would already have broken with `converged=True` on the next iteration. It never changes the result.

jkitchin (commented):
Automated high-effort review (8 finder angles + verification). The numerics (Jacobian assembly, sparse layout, IFT backward) check out; findings cluster around **convergence/status reporting** (success reported too eagerly) and **SciPy-compatibility gaps**, plus some frontend duplication. Three candidates were refuted and dropped during verification.

The top three to block on: the `1e-6` convergence floor, the uncaught singular-Jacobian `RuntimeError`, and IPM "acceptable" being reported as full success — each lets a wrong or crashing answer be presented as a successful solve.

crates/pounce-py/src/sparse_lu.rs
jkitchin: Fixed, but in Python rather than Rust. `newton_solve` now wraps the `lu.factor` call in `try/except RuntimeError` and returns `status=2` (singular), which `solve_bvp` maps to a `BVPResult` with `success=False`, `status=2`, message "A singular Jacobian was encountered" — no exception propagates, matching SciPy. Left the binding raising (a singular factor is a genuine error for a general `SparseLU` consumer); the BVP layer is responsible for the SciPy contract. Added `test_solve_bvp_singular_does_not_raise` (resonant `y''+π²y=0`, `y(0)=y(1)=0`).

---

python/pounce/jax/_bvp.py
jkitchin: Partly. I extracted the host-side back-solve into a single shared `ift_solve_transpose` (in `_solve.py`) now used by both the jax and torch backwards, so the factor-at-`z*` + `solve_transpose` half is no longer duplicated/diverged.

But I did **not** thread the forward factor out, because the premise doesn't hold with the new frozen-Jacobian Newton: the forward reuses a single factor across damped steps and converges by breaking at the top-of-loop `‖R‖<tol` test *before* refactoring — so the held LU is generally at an earlier iterate, **not** at `z*`. Reusing it would give an inexact `R_zᵀ` and a wrong gradient. To reuse correctly we'd have to add a factorization at `z*` at the end of the forward (the very work we just removed), so factoring once at `z*` in the backward is the natural, cheapest correct place. I did document this in the helper. The genuinely-avoidable waste you flagged — re-running `analyze()` — is the remaining cost; happy to add a thread-pinned cached symbolic if this shows up in a training-loop profile.

---

python/pounce/jax/_bvp.py
jkitchin: Kept dense here on purpose, and added a comment saying so. This is the `second_order=True` path, whose whole point is that the tangent rule stays in pure JAX so `jax.grad`/`jax.hessian` can recurse through it. A FERAL `SparseLU` back-solve is a `pure_callback`, which has no JVP rule — using it would re-introduce exactly the "pure callbacks do not support JVP" failure that blocks second order on the plain `pounce.jax.solve` path. So the dense `O(N^3)` `jnp.linalg.solve` is the deliberate cost of arbitrary-order autodiff on this opt-in path; the fast first-order default (`method="newton"`) already uses the sparse FERAL back-solve. A sparse-but-differentiable custom JVP/transpose rule for the linear solve would get the best of both, but that's a larger follow-up.

---

python/pounce/bvp/_newton.py
jkitchin: Fixed. `newton_solve` no longer uses the `1e-6` floor: on a fresh-factor stall it now sets the outcome by `rnorm < tol` (the requested tolerance, passed through as `newton_tol = min(tol, 1e-10)`), and returns a status code (`0` converged / `2` singular / `4` not converged) instead of a bool. `solve_bvp` maps that straight through, so a solve that stalls at e.g. `1e-7` when `1e-10` was requested now reports `status=4`, `success=False` rather than a false `status=0`. (commit on the branch)

---

python/pounce/bvp/_solve.py
jkitchin: Fixed. Added `_ipm_status()`: IPM code `0` → status `0`; code `1` (`SolvedToAcceptableLevel`) → a distinct status `5` with `success=False` (message "the IPM reached only its acceptable tolerance"); anything else → `4`. Applied in both `_solve.py` and `_constrained.py:332`. So the looser `acceptable_tol` outcome is no longer reported as a clean solve.

On the `rms_residuals`-only point: `success` is now driven by the status code, not by `rms_residuals`. For the Newton path `rnorm` is the inf-norm of the **full** residual (collocation + boundary), and a separate `bc_tol` check (status `3`) guards the boundary block; for the IPM/constrained path the IPM status reflects the full KKT (bounds + path rows included), so a violated boundary/path constraint now shows up in `success`. `rms_residuals` stays collocation-only to match SciPy's field semantics.

---

PR #153, opened 2026-06-19
Add pounce.ode: stiff Radau IIA(5) solve_ivp, index-1 DAEs, differentiable odeint

jkitchin:
The ODE/IVP analog of the BVP work: a `scipy.integrate.solve_ivp`-compatible
stiff solver, index-1 DAE support, and differentiable JAX/PyTorch frontends.

## `pounce.ode.solve_ivp`

Drop-in for `scipy.integrate.solve_ivp(method="Radau")` — the 3-stage Radau
IIA collocation method (order 5, L-stable; the RADAU5 of Hairer–Wanner that
SciPy implements). Adaptive step control with the embedded order-3 error
estimate; each step's stage system is solved by a simplified Newton whose
Jacobian is factored with FERAL's sparse LU.

- **Tracks SciPy step-for-step on stiff problems.** Van der Pol μ=1000 over
  `[0, 3000]`: **1082 steps vs SciPy's 1188, agreeing to ~7e-7** (0.53s vs
  0.42s).
- Getting there required fixing the adaptive controller: the simplified-Newton
  stage solve now uses the Hairer–Wanner tolerance-aware, rate-predictive
  convergence test (not a fixed `1e-10`), and the Jacobian is marked stale
  after each accepted step and refreshed on a convergence failure (rather than
  frozen while `rate<0.3`). This removed a stall where a stale Jacobian capped
  the achievable step. Also added `_select_initial_step` (SciPy/Hairer
  heuristic).
- Returns a SciPy-shaped bunch (`t`, `y`, `sol`, `nfev`, `njev`, `nlu`,
  `status`, `message`, `success`). Supports `t_eval`, `dense_output`, `args`,
  `jac`, `first_step`, `max_step`, `rtol`/`atol`.
- Only `method="Radau"` is implemented (the stiff/DAE niche); other methods
  raise rather than silently substitute, and `events=` is not yet supported.

## Index-1 DAEs (beyond SciPy)

Pass `mass=M` to integrate `M y' = f`. A **singular** `M` makes it an index-1
differential-algebraic equation — something `scipy.integrate.solve_ivp` cannot
do. Validated on Robertson kinetics (conservation constraint held to
round-off; `y(1e4)` matches the literature value).

## Differentiable integration

`pounce.jax.odeint` / `pounce.torch.odeint` integrate on a fixed mesh and
return the trajectory differentiably w.r.t. the ODE parameters `theta` **and**
the initial condition `y0`. An IVP on a fixed mesh is a BVP with
`bc(ya, yb) = ya - y0`, so these reuse the Hermite–Simpson collocation core
and the same FERAL sparse-LU implicit-diff back-solve as `solve_bvp`.
Gradients/Jacobians validated against analytic and finite differences (~1e-8)
in both backends.

## Results

`OdeResult` and `BVPResult` are now dict-subscriptable (SciPy-`Bunch` style:
`res["y"]`, `"success" in res`, `res.keys()`, `res.get(...)`) via a shared
`ResultMixin`, alongside attribute access. `pounce.solve_ivp` /
`pounce.OdeResult` are exposed at the top level.

## Docs / tests / example

- `docs/src/ode.md` (+ SUMMARY entry), CHANGELOG section.
- `python/tests/test_ode.py` — 14 tests (stiff vs scipy, Van der Pol, DAEs,
  `t_eval`/`args`, dense output, dict access, jax + torch gradients). All
  green; the existing 17 BVP tests still pass.
- `python/examples/ode_scipy_compare.py` — runs end-to-end (stiff vs scipy,
  Robertson DAE, differentiability).

## Scope note

This is a faithful scipy-class Radau5 (same algorithm) — "competitive with a
production stiff solver", not a numerically-SOTA stepper. The genuine new
value is **DAE support** and **differentiability**, which SciPy's `solve_ivp`
lacks.

https://claude.ai/code/session_01CkewNiehrg7D1sf5SPRfC3

---
_Generated by [Claude Code](https://claude.ai/code/session_01CkewNiehrg7D1sf5SPRfC3)_

### DIFF
--- a/python/pounce/_result.py
+++ b/python/pounce/_result.py
@@ -18,13 +18,21 @@ class ResultMixin:
     ``res.keys()`` / ``res.get(...)`` and iteration over field names on top
     of normal attribute access, so a pounce result is interchangeable with a
     SciPy result in downstream code.
+
+    The dict view is restricted to the **public** fields (those that appear
+    in ``repr`` — i.e. fields declared ``repr=False``, such as internal
+    solver state, are excluded), and ``__getitem__`` honours that same set,
+    so ``"k" in res`` and ``res["k"]`` agree and private state is never
+    exposed through the SciPy-Bunch surface.
     """
 
+    def keys(self):
+        return [f.name for f in dataclasses.fields(self) if f.repr]
+
     def __getitem__(self, key):
-        try:
-            return getattr(self, key)
-        except AttributeError:
+        if key not in self.keys():
             raise KeyError(key)
+        return getattr(self, key)
 
     def __setitem__(self, key, value):
         setattr(self, key, value)
@@ -32,11 +40,8 @@ class ResultMixin:
     def __contains__(self, key):
         return key in self.keys()
 
-    def keys(self):
-        return [f.name for f in dataclasses.fields(self)]
-
     def __iter__(self):
         return iter(self.keys())
 
     def get(self, key, default=None):
-        return getattr(self, key, default)
+        return getattr(self, key, default) if key in self.keys() else default
--- a/python/pounce/_result.py
+++ b/python/pounce/_result.py
@@ -1,0 +1,42 @@
+"""Shared dict/attribute access for SciPy-style result objects.
+
+SciPy's ``solve_ivp`` / ``solve_bvp`` return a ``Bunch`` whose fields are
+reachable both as attributes (``res.y``) and as dict items (``res["y"]``).
+:class:`ResultMixin` gives pounce's dataclass results the same dual access
+so they are drop-in for code written against SciPy.
+"""
+
+from __future__ import annotations
+
+import dataclasses
+
+
+class ResultMixin:
+    """Dict-style access for dataclass results, mirroring SciPy's ``Bunch``.
+
+    Adds ``res["field"]`` indexing, ``"field" in res`` membership,
+    ``res.keys()`` / ``res.get(...)`` and iteration over field names on top
+    of normal attribute access, so a pounce result is interchangeable with a
+    SciPy result in downstream code.
+    """
+
+    def __getitem__(self, key):
+        try:
+            return getattr(self, key)
+        except AttributeError:
+            raise KeyError(key)
+
+    def __setitem__(self, key, value):
+        setattr(self, key, value)
+
+    def __contains__(self, key):
+        return key in self.keys()
+
+    def keys(self):
+        return [f.name for f in dataclasses.fields(self)]
+
+    def __iter__(self):
+        return iter(self.keys())
+
+    def get(self, key, default=None):
+        return getattr(self, key, default)
--- a/python/pounce/jax/_ode.py
+++ b/python/pounce/jax/_ode.py
@@ -30,6 +30,7 @@ import numpy as np
 
 from ..bvp import _core
 from ..bvp._solve import _make_spline, ift_solve_transpose
+from ..ode._solve import mesh_initial_guess
 
 
 @dataclass
@@ -38,36 +39,15 @@ class JaxODESolution:
 
     ``t`` ``(m,)`` and ``y`` ``(n, m)`` (SciPy ``solve_ivp`` layout). ``y``
     carries the custom-VJP back to ``theta`` and ``y0``; differentiate
-    through it with ``jax.grad`` / ``jax.jacobian``. ``yp`` is ``dy/dt`` at
-    the nodes and ``sol`` a (non-differentiable) cubic-Hermite interpolant.
+    through it with ``jax.grad`` / ``jax.jacobian``. ``yp`` (``dy/dt`` at the
+    nodes) and ``sol`` (a cubic-Hermite interpolant) are **non-differentiable**
+    diagnostics — both are detached, so only ``y`` should appear in a loss.
     """
 
     t: jnp.ndarray
     y: jnp.ndarray
     yp: jnp.ndarray
     sol: Callable
-
-
-def _initial_guess(fun_np, t_np, y0_np, n, m):
-    """Cheap explicit trajectory on the mesh, used only to seed Newton.
-
-    Runs pounce's adaptive Radau on the concrete RHS sampled at the mesh
-    nodes. Init-guess quality only affects convergence, never the solution
-    or its gradient, so any reasonable trajectory will do.
-    """
-    from ..ode import solve_ivp as _solve_ivp
-    try:
-        res = _solve_ivp(
-            fun_np, (float(t_np[0]), float(t_np[-1])), y0_np,
-            method="Radau", t_eval=t_np, rtol=1e-3, atol=1e-6,
-        )
-        Y = np.asarray(res.y, dtype=np.float64)
-        if Y.shape == (n, m) and np.all(np.isfinite(Y)):
-            return Y
-    except Exception:
-        pass
-    # Fallback: hold the initial state across the mesh.
-    return np.broadcast_to(y0_np[:, None], (n, m)).copy()
 
 
 def odeint(fun, y0, t, theta=None, *, tol=1e-8):
@@ -145,23 +125,32 @@ def odeint(fun, y0, t, theta=None, *, tol=1e-8):
 
     def _host_solve(p_np):
         from ..bvp._jac import CollocationJacobian
-        from ..bvp._newton import newton_solve
+        from ..bvp._newton import newton_solve, STATUS_CONVERGED
         p_np = np.asarray(p_np, dtype=np.float64)
         nfun, nbc, y0v = _np_callables(p_np)
 
         def fun_np(ti, yi):
             return nfun(np.array([ti]), np.asarray(yi)[:, None], None)[:, 0]
 
-        Yg = _initial_guess(fun_np, t_np, y0v, n, m)
+        Yg = mesh_initial_guess(fun_np, t_np, y0v, n, m)
         z0 = Yg.reshape(-1)
 
         def residual_fn(z):
             return _core.residual_of_z(z, nfun, nbc, t_np, n, m, 0, np.concatenate)
 
         jac = CollocationJacobian(nfun, nbc, t_np, n, m, 0)
-        z_star, _it, _ok, _rn = newton_solve(
+        z_star, _it, status, rnorm = newton_solve(
             residual_fn, jac, z0, n, m, 0, tol=float(tol),
         )
+        # Returning a non-converged z* would yield a wrong trajectory *and*
+        # IFT gradients about a point where R(z) != 0. There is no status
+        # surface on the solution object, so fail loudly instead.
+        if status != STATUS_CONVERGED:
+            raise RuntimeError(
+                "pounce.jax.odeint: collocation Newton did not converge "
+                f"(status={status}, ||R||={rnorm:.3e}). The fixed mesh `t` is "
+                "likely too coarse to resolve the dynamics — refine it."
+            )
         return np.asarray(z_star, dtype=np.float64)
 
     def _host_btran(z_star_np, p_np, v_np):
@@ -206,7 +195,11 @@ def odeint(fun, y0, t, theta=None, *, tol=1e-8):
     z_star = solve_fn(combined)
     Y_star = z_star.reshape(n, m)
     th, _ = _split(combined)
-    yp = _vec_rhs(tj, Y_star, th)
-    sol = _make_spline(tj, Y_star, yp)
+    # yp / sol are non-differentiable diagnostics. yp is recomputed outside
+    # the custom-VJP boundary, so left attached it would carry a spurious
+    # direct df/dtheta term inconsistent with the true converged sensitivity
+    # (which flows only through Y_star); stop_gradient makes that explicit.
+    yp = jax.lax.stop_gradient(_vec_rhs(tj, Y_star, th))
+    sol = _make_spline(tj, jax.lax.stop_gradient(Y_star), yp)
 
     return JaxODESolution(t=tj, y=Y_star, yp=yp, sol=sol)
--- a/python/pounce/jax/_ode.py
+++ b/python/pounce/jax/_ode.py
@@ -1,0 +1,212 @@
+"""Differentiable ODE integration on a fixed mesh (JAX frontend).
+
+``pounce.jax.odeint`` integrates ``dy/dt = f(t, y, theta)`` and returns the
+trajectory on a user-supplied mesh ``t`` differentiably w.r.t. the ODE
+parameters ``theta`` **and** the initial condition ``y0``.
+
+An initial value problem on a fixed mesh is just a boundary value problem
+whose boundary condition pins the left end, ``bc(ya, yb) = ya - y0``, so this
+reuses pounce's 4th-order Hermite--Simpson collocation and its
+implicit-function-theorem differentiation verbatim (see
+:mod:`pounce.bvp._core`). The converged node states solve the *square*
+collocation root-find ``R(z, p) = 0`` with ``p = [theta; y0]``; gradients
+fall out of ``dz*/dp = -(dR/dz)^{-1}(dR/dp)`` with the same FERAL sparse-LU
+back-solve used by the BVP layer — no per-step adjoint, no unrolled tape.
+
+This is the *differentiable* counterpart to the adaptive, stiff/DAE
+:func:`pounce.ode.solve_ivp`: the result here is the collocation solution on
+the mesh you pass (make it fine enough to resolve the dynamics), and its
+gradients are exact for that discretisation. ``f`` must be JAX-traceable.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Callable
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+
+from ..bvp import _core
+from ..bvp._solve import _make_spline, ift_solve_transpose
+
+
+@dataclass
+class JaxODESolution:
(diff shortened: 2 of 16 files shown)


jkitchin (commented):
Automated high-effort review (8 finder angles + verification). The core Radau stepper is numerically sound — I checked the embedded error estimate and it correctly matches SciPy's `Z`-increment + `E`-weight convention (refuted a tempting false positive). Findings cluster in **SciPy-contract failure reporting**, **backward/decreasing-mesh edge cases**, and one significant **efficiency regression** (the RADAU5 factor-reuse trick is defeated).

Blockers are the first four: two break SciPy's never-raise/report-via-`success` contract, one silently corrupts backward dense output, and one lets `odeint` return wrong gradients with no warning surface.

python/pounce/ode/_solve.py
```diff
@@ -0,0 +1,145 @@
...
+            jac = lambda t, y: _jac(t, y, *args)
+
+    try:
+        res = _radau.integrate(
+            fun, t0, t1, y0, rtol=rtol, atol=atol, first_step=first_step,
+            max_step=max_step, mass=mass, jac=jac, t_eval=t_eval,
+            dense_output=dense_output or t_eval is not None,
+        )
+        status, message, success = 0, "The solver successfully reached the end of the integration interval.", True
+    except RuntimeError as e:
```
jkitchin: **Integration failure raises instead of returning `success=False`.**

This `try: … except RuntimeError as e: raise` is a no-op handler, and `status/message/success` are only ever assigned the success triple (line 130). When `_radau.integrate` raises `RuntimeError("Radau: step underflow at t=…")` (`_radau.py:222`), it propagates as an exception. SciPy's `solve_ivp` **never** raises on numerical failure — it returns `success=False, status=-1` with a message. Drop-in code doing `if res.success:` crashes against pounce. Map the failure to a SciPy-style failed result (and ideally return the partial trajectory accumulated so far).

python/pounce/ode/_radau.py
```diff
@@ -0,0 +1,315 @@
...
+
+    # Step records for dense output: (t_start, h, y_start, K).
+    records = [] if (dense_output or t_eval is not None) else None
+    ts = [t]
+    ys = [y.copy()]
+    nstep = nrej = 0
+    jac_current = True
+    rejected = False
+
+    while (t - t1) * s < -1e-12 and nstep < max_steps:
```
jkitchin, line 227: **Hitting the `max_steps` cap is reported as success.**

`while (t - t1)*s < -1e-12 and nstep < max_steps:` — if the step cap (`max_steps=10**6`) trips first, the loop exits, returns its dict, and `_solve.py:130` unconditionally sets `status=0, success=True`, with no check that `t` actually reached `t1`. A step-rejection storm yields a **truncated trajectory** (`res.t[-1] != t_span[1]`) reported as fully successful. SciPy returns `status=-1` with a "maximum number of steps" message. Check `t` reached `t1` on loop exit and set a failure status otherwise.

python/pounce/ode/_radau.py
```diff
@@ -0,0 +1,315 @@
...
+    # Fit u'(τ) = Σ_i K_i ℓ_i(τ) (Lagrange on nodes c), integrate.
+    C = RADAU_C
+    # Vandermonde for u'(τ)=Σ a_j τ^j (degree 2) matching u'(c_i)=K_i.
+    V = np.vander(C, 3, increasing=True)         # (3,3)
+    Vinv = np.linalg.inv(V)
+
+    def sol(tq):
+        tq = np.atleast_1d(np.asarray(tq, dtype=float))
+        # locate step index for each query point
+        idx = np.searchsorted(ends, tq, side="left")
```
jkitchin: **Dense output / `t_eval` is silently wrong for backward integration.**

`np.searchsorted(ends, tq)` assumes an ascending array, but for backward integration (`t_span=(10, 0)`, `s=-1`, `hs<0`) `ends = starts + hs` is strictly **descending** — so this returns wrong indices and interpolates from the wrong step's polynomial (the `np.clip` only bounds the range, it doesn't fix the ordering). Backward integration is advertised and `dense_output`/`t_eval` are supported, but no test covers backward + dense, so it's uncaught. Make `_make_dense` direction-aware (e.g. search on `np.minimum(starts, ends)` ascending, or reverse the records).

python/pounce/jax/_ode.py
```diff
@@ -0,0 +1,212 @@
...
+            return nfun(np.array([ti]), np.asarray(yi)[:, None], None)[:, 0]
+
+        Yg = _initial_guess(fun_np, t_np, y0v, n, m)
+        z0 = Yg.reshape(-1)
+
+        def residual_fn(z):
+            return _core.residual_of_z(z, nfun, nbc, t_np, n, m, 0, np.concatenate)
+
+        jac = CollocationJacobian(nfun, nbc, t_np, n, m, 0)
+        z_star, _it, _ok, _rn = newton_solve(
```
jkitchin: **`odeint` discards the Newton convergence flag → silently wrong solution and gradients.**

`z_star, _it, _ok, _rn = newton_solve(...)` — `_ok` (a real `STATUS_CONVERGED`/`STATUS_NOT_CONVERGED` code) is never checked. On a too-coarse fixed mesh the collocation root-find can stop at `max_iter` without converging; `odeint` then returns the non-converged node states as `y` **and** computes IFT gradients about a point that does not satisfy `R(z)=0`. `JaxODESolution`/`TorchODESolution` expose only `t,y,yp,sol` — no `success`/`status` field — so there's no surface to warn the user. Propagate `_ok` and attach a status / raise / NaN-poison on non-convergence. Same issue in `torch/_ode.py:126`.

python/pounce/ode/_radau.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+class _RadauProblem:
+    """Holds the RHS, mass matrix, Jacobian, and counters for one solve."""
+
+    def __init__(self, fun, n, mass=None, jac=None):
+        self.fun = fun
+        self.n = n
+        self.M = np.eye(n) if mass is None else np.asarray(mass, dtype=float)
+        self.is_dae = mass is not None and np.linalg.matrix_rank(self.M) < n
```
jkitchin: **A full-rank non-identity mass matrix is routed to the wrong error-estimate branch.**

`is_dae = mass is not None and matrix_rank(M) < n`, so a legitimate non-singular `M y' = f` with e.g. `M=diag(2,1)` gets `is_dae=False`. The branches at line 140 are `prob.M @ (Z.T @ (E/h)) + f0` (DAE) vs `f0 + Z.T @ (E/h)` (non-DAE) — the non-DAE branch omits the `M @`, while the smoothing operator `lu_real = MU_REAL/hs*M - J` (line 206) uses the true `M`. For `M≠I` the error norm is mis-scaled and step control degrades. (Identity `M` is fine.) Apply the `M @` factor regardless of `is_dae`, or key the branch on `M is not None` rather than singularity.

python/pounce/ode/_radau.py
```diff
@@ -0,0 +1,315 @@
...
+    ys = [y.copy()]
+    nstep = nrej = 0
+    jac_current = True
+    rejected = False
+
+    while (t - t1) * s < -1e-12 and nstep < max_steps:
+        h = min(h, abs(t1 - t))
+        hs = s * h
+        # Factor the stage system and the error operator for this (h, J).
+        big = np.kron(np.eye(3), prob.M) - hs * np.kron(RADAU_A, J)
```
jkitchin: **(efficiency) The stage Jacobian and its LU are re-factored every step, defeating the core RADAU5 reuse trick.**

Lines 204–206 rebuild `np.kron(eye(3), M) - hs*np.kron(A, J)` and call `_dense_lu` **twice unconditionally** every iteration (`nlu += 2`), with no guard on whether `h` or `J` changed. The module docstring claims "The Jacobian and its factorisation are reused across steps … the standard Hairer–Wanner efficiency trick" — but the factor is only reused across Newton iterations *within* a step, never across steps (even after a rejected step with unchanged `h`, or a deliberately-frozen `J`). Cache `(h_factored, lu3, lu_real)` and refactor only when `J` is refreshed or `|h/h_factored − 1|` leaves a band. Worth re-measuring the "typically faster than SciPy" benchmark after fixing.

python/pounce/jax/_ode.py
```diff
@@ -0,0 +1,212 @@
...
+        (dp,) = vjp_p(-u)
+        return (dp,)
+
+    solve_fn.defvjp(fwd, bwd)
+
+    z_star = solve_fn(combined)
+    Y_star = z_star.reshape(n, m)
+    th, _ = _split(combined)
+    yp = _vec_rhs(tj, Y_star, th)
+    sol = _make_spline(tj, Y_star, yp)
```
jkitchin: **`sol` breaks on a decreasing mesh that the docstring permits.**

The docstring (line 85) says the mesh `t` need only be "monotonic", but `sol` is built via `_make_spline` → `CubicHermiteSpline(x, …)`, which requires strictly increasing `x`. A decreasing mesh (`linspace(2.0, 0.0, 81)`) computes `.y` fine but raises `ValueError: x must be strictly increasing` on the first `.sol(tq)` call (lazy, so it surfaces at call time). Either sort inside `_make_spline` or document the mesh as increasing-only. Same in `torch/_ode.py`.

python/pounce/jax/_ode.py
```diff
@@ -0,0 +1,212 @@
...
+        _, vjp_p = jax.vjp(lambda pp: residual_jax(z_star, pp), p)
+        (dp,) = vjp_p(-u)
+        return (dp,)
+
+    solve_fn.defvjp(fwd, bwd)
+
+    z_star = solve_fn(combined)
+    Y_star = z_star.reshape(n, m)
+    th, _ = _split(combined)
+    yp = _vec_rhs(tj, Y_star, th)
```
jkitchin: **`yp` differentiates inconsistently.**

`yp = _vec_rhs(tj, Y_star, th)` is recomputed *outside* the custom-VJP boundary from graph-connected inputs, so it carries the IFT gradient through `Y_star` **plus** a spurious direct `∂f/∂θ` term. Differentiating a loss built on `res.yp` silently yields a gradient inconsistent with the true converged sensitivity. (`sol` is documented non-differentiable and detached via `_to_numpy`, so it's safe — only `yp` is exposed and not documented as non-diff.) Either detach `yp` or document it as non-differentiable. Same in `torch/_ode.py:159`.

python/pounce/_result.py
```diff
@@ -0,0 +1,42 @@
...
+        except AttributeError:
+            raise KeyError(key)
+
+    def __setitem__(self, key, value):
+        setattr(self, key, value)
+
+    def __contains__(self, key):
+        return key in self.keys()
+
+    def keys(self):
```
jkitchin: **`ResultMixin` leaks the internal `info` field and has an inconsistent dict contract.**

`keys()` returns *all* dataclass fields including `info` (internal IPM state), so `dict(res)`, iteration, and `'info' in res` expose private state SciPy's bunch never has — undercutting the advertised "SciPy-Bunch drop-in." Separately, `__getitem__` (line 23) delegates to `getattr` for any attribute while `__contains__` (line 32) only checks field names, so `res['get']` returns a bound method while `'get' in res` is `False`. Restrict `keys()`/`__getitem__` to the public field set (exclude `repr=False` fields).

python/pounce/torch/_ode.py
```diff
@@ -0,0 +1,178 @@
...
+    z_star = _Solve.apply(combined)
+    Y_star = z_star.reshape(n, m)
+    th, _ = _split(combined)
+    yp = _vec_rhs(fun, tt, Y_star, th, has_theta)
+    sol = _make_spline(tt, Y_star, yp)
+
+    return TorchODESolution(t=tt, y=Y_star, yp=yp, sol=sol)
+
+
+def _initial_guess(fun_np, t_np, y0_np, n, m):
```
jkitchin: **(cleanup) `_initial_guess` is duplicated verbatim, and `_radau` carries dead `J` params.**

This `_initial_guess` is byte-identical to `jax/_ode.py:51` — it's pure NumPy with no backend dependency, so it should live once in a shared module and be imported by both frontends (mirroring how `_make_spline`/`ift_solve_transpose` are already shared). Separately, in `ode/_radau.py` both `_solve_stages` and `_error_estimate` take a `J` parameter that neither body uses (the Jacobian is already baked into the LU factors) — drop it.

jkitchin:
Thanks for the thorough review — all ten addressed in d0fcfc0. Mapping:

**Correctness**
- *Failure raises instead of `success=False`* — `_radau.integrate` no longer raises. Step underflow and hitting the step cap return the partial trajectory with `status<0` / `success=False` and a message; `solve_ivp` propagates it. (`_radau.py`, `_solve.py`)
- *`max_steps` cap reported as success* — on loop exit I check whether `t` actually reached `t1`; if not, `status=-2` ("maximum number of internal steps"), `success=False`. New test `test_step_cap_reports_failure_not_success`.
- *Dense output wrong for backward integration* — `_make_dense` now locates query points via an `argsort`-sorted lower-bound key, correct for a descending mesh. New test `test_backward_dense_output`.
- *`odeint` discards Newton convergence flag* — both frontends now check the status and raise a clear "mesh too coarse" error rather than returning a non-converged `z*` with IFT gradients about a point where `R(z)≠0`.
- *Full-rank non-identity `M` mis-scaled* — `_error_estimate` applies the `M @` factor unconditionally (it reduces to the plain form when `M=I`). New test `test_full_rank_nonidentity_mass`.

**Differentiability / API**
- *`yp` differentiates inconsistently* — `yp` (and `sol`) are now detached and documented non-differentiable on both `JaxODESolution`/`TorchODESolution`.
- *`sol` breaks on decreasing mesh* — `_make_spline` reverses a decreasing mesh before the `CubicHermiteSpline` build.
- *`ResultMixin` leaks `info` / inconsistent contract* — `keys()`/`__getitem__`/`__contains__`/`get` are all restricted to the public (`repr=True`) fields, so `info` is gone from the Bunch surface and `k in res` ⇔ `res[k]` agree. New test `test_info_excluded_from_dict_surface`.

**Efficiency / cleanup**
- *LU re-factored every step* — the stage/error-operator LU is now cached and refactored only when `J` is refreshed or `h` changes; `h` is held on mild growth (band `[1, 1.2)`) to keep reusing it. VdP μ=1000 `nlu` 2528→2090. (The CHANGELOG already states pounce is ~slightly slower than SciPy here, 0.5s vs 0.44s, so no "faster" claim to retract.)
- *Dead `J` params / duplicated `_initial_guess`* — dropped `J` from `_solve_stages`/`_error_estimate`; moved the guess into shared `pounce.ode._solve.mesh_initial_guess`, imported by both frontends.

18 ODE tests pass, the 17 BVP tests still pass.

---

python/pounce/ode/_radau.py
```diff
@@ -0,0 +1,375 @@
...
+        out["sol"] = sol
+        if t_eval is not None:
+            te = np.asarray(t_eval, dtype=float)
+            # Only evaluate t_eval points covered by the (possibly partial)
+            # trajectory; clip keeps a failed solve from extrapolating wildly.
+            out["t"] = te
+            out["y"] = sol(te)
+    elif t_eval is not None:
+        out["t"] = np.asarray(t_eval, dtype=float)
+        out["y"] = np.empty((n, np.asarray(t_eval).size))
```
jkitchin: **(minor, follow-up) Uninitialized memory on the no-records + `t_eval` failure path.**

`out["y"] = np.empty((n, np.asarray(t_eval).size))` returns garbage values. This is reachable when the very first step underflows before any step is recorded but `t_eval`/`dense_output` was requested: `y` comes back as uninitialized memory. `success=False` does flag the failure, but returning garbage numbers is a footgun — prefer `np.full((n, size), np.nan)` so a caller that ignores `success` gets obvious NaNs rather than plausible-looking noise.

python/pounce/ode/_radau.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+class _RadauProblem:
+    """Holds the RHS, mass matrix, Jacobian, and counters for one solve."""
+
+    def __init__(self, fun, n, mass=None, jac=None):
+        self.fun = fun
+        self.n = n
+        self.M = np.eye(n) if mass is None else np.asarray(mass, dtype=float)
+        self.is_dae = mass is not None and np.linalg.matrix_rank(self.M) < n
```
jkitchin: **(minor, follow-up) `is_dae` is now dead.**

With the `_error_estimate` branch removed (good fix — `M @` is now applied unconditionally), `self.is_dae` is computed but never read anywhere. It's an unused `np.linalg.matrix_rank(M)` call on every solve. Drop it, or — if DAE introspection is worth keeping — surface it on the result/info instead of computing it for nothing.

jkitchin (commented):
Re-reviewed the fixes in d0fcfc0 — all 10 prior findings are correctly addressed (status reporting, backward dense output, unconditional mass-matrix term, LU factor reuse with an h-hysteresis band, ResultMixin info exclusion, yp/sol detachment, shared mesh_initial_guess), with good regression tests. Two minor, non-blocking follow-ups below.

jkitchin:
Both follow-ups addressed in 9589116:

- **Uninitialized memory on the no-records + `t_eval` failure path** — now `np.full((n, size), np.nan)` so a caller ignoring `success=False` gets obvious NaNs. New test `test_failure_with_t_eval_returns_nans_not_garbage`.
- **Dead `is_dae`** — dropped; it was an unused `matrix_rank()` per solve once `M @` became unconditional in the error estimate.

19 ODE tests pass.

---

PR #164, opened 2026-06-23
feat(minimize): honor sparse Jacobians from dict constraints

jduerholt:
## Summary

`pounce.minimize`'s **dict-constraint** path declared a fully dense
`m_rows × n` Jacobian sparsity pattern regardless of what the user's `jac`
returned. A genuinely sparse constraint Jacobian — the common case for
block-structured problems like D-optimal experimental design, where each
constraint touches only one experiment's variables — was therefore densified
into the KKT system, which is exactly the cost the sparse `Problem` API avoids.

This PR teaches the dict path to honor a sparse Jacobian, for **linear and
nonlinear** constraints alike.

## Why this design — and how cyipopt does it

cyipopt's scipy interface (`minimize_ipopt`) already infers constraint
sparsity **by inspecting the `jac` return type**, not via any flag on the
constraint. In `cyipopt/scipy_interface.py::_get_sparse_jacobian_structure`, it
evaluates each constraint's jacobian at `x0` and, if the result is a
`coo_array`, marks that constraint sparse and reads its nnz structure:

```python
jac_val = con_jac(x0, ...)
if isinstance(jac_val, coo_array):
    jacobians.append(jac_val)
    con_jac_is_sparse.append(True)
else:
    ...  # dense
```

(One subtlety we hit while reconciling the two: cyipopt only sees this when the
constraint is given as a **dict** with a coo-returning `jac`. Scipy's
`standardize_constraints(..., 'old')` — which cyipopt runs internally, and which
`new_constraint_to_old` implements with `A = A.toarray()` / `dy.toarray()` —
**densifies** `LinearConstraint`/`NonlinearConstraint` *objects*. So the dict
form is the representation that carries sparsity through to cyipopt.)

pounce already does the equivalent detect-and-use-sparse for `LinearConstraint`
blocks (`_block_from_linear_constraint` coalesces `A` to COO). This PR extends
the **same** treatment to dict constraints — mirroring cyipopt's
detect-by-return-type — so that a single
`{"type": ..., "fun": ..., "jac": -> coo_array}` dict is a sparse constraint for
**both** cyipopt's and pounce's scipy interfaces. That common representation is
what lets a downstream caller (e.g. BoFire's DoE module) feed one constraint
list to either solver and keep sparsity.

## What changed

`python/pounce/_minimize.py`:
- `_block_from_dict` probes `jac(x0)`; if it returns any scipy-sparse matrix, it
  declares that matrix's COO structure (canonicalized via `.tocsr().tocoo()` for
  a deterministic row-major order, so the build-time structure and the
  per-iteration values stay aligned) instead of the dense `m_rows × n` grid.
- `jac_values` streams the COO `.data` for such blocks instead of a dense
  `ravel()`, with a fixed-pattern guard that raises a clear error if the nnz
  count changes between the probe and a solve (Ipopt requires a fixed sparsity
  pattern).
- A dense return — or no `jac` (finite differences) — keeps the original dense
  pattern, so existing behavior is unchanged. `LinearConstraint` handling is
  untouched. Any scipy sparse format is accepted (slightly more lenient than
  cyipopt's coo-only check).

No Rust changes — this is entirely in the Python facade; the native `Problem`
already consumes the `(rows, cols)` structure + values.

## Tests

`python/tests/test_minimize.py` — 5 new tests:
- structure check (block-diagonal: declares 4 nonzeros, not the dense 2×6 = 12);
- linear-as-sparse-dict reaches the same optimum as the `LinearConstraint` form;
- nonlinear sparse `jac` reaches the same point as the dense-`jac` form;
- fixed-pattern-change raises a clear error;
- dense-`jac` dict regression stays dense.

Full suite: **533 passed, 3 skipped**.

### DIFF
--- a/python/pounce/_minimize.py
+++ b/python/pounce/_minimize.py
@@ -349,6 +349,12 @@ class _ConstraintBlock:
     lb: np.ndarray  # length n_rows
     ub: np.ndarray  # length n_rows
     n_rows: int
+    # Dict block whose ``jac`` returns a *scipy-sparse* matrix: ``(rows, cols)``
+    # above are that matrix's COO structure (canonicalized) and ``jac_values``
+    # streams the per-iteration COO data instead of a dense ``ravel()``. Lets a
+    # nonlinear (or linear) dict constraint declare a sparse Jacobian, mirroring
+    # cyipopt's detect-by-return-type behavior.
+    sparse_jac: bool = False
     # Pre-assembled ``sparse.coo_array`` for linear blocks. ``g_combined``
     # would otherwise rebuild this from ``(rows, cols, constant_vals)`` on
     # every Ipopt iteration (per-iter waste in the inner loop). ``None`` for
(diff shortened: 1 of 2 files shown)


jkitchin:
Two points from review for consideration:

**1. Fixed-pattern guard checks only nnz *count*, not nonzero *positions* (`_minimize.py`, `jac_values`)**

The guard compares `J.data.size != (end - start)` but never checks `J.row`/`J.col` against the structure declared at probe time. A nonlinear constraint whose sparse `jac` keeps the same nnz count but moves a nonzero to a different column between iterations (e.g. a jac built by dropping numerical zeros, where one structural entry vanishes as another appears) passes the guard, and the value for the new column is fed to Ipopt as the derivative at the *old* column — a silently-wrong Jacobian, no error.

Reproduced concretely: declared cols `[0, 2]`, solve returns cols `[0, 1]` with values `[5, 7]` → guard passes and `7.0` (the col-1 derivative) is reported at the declared col 2. This defeats the guard's stated promise of "a clear error if the pattern changes."

Cheap fix: since both sides are already canonicalized via `.tocsr().tocoo()`, also assert `J.row`/`J.col` equal the declared `blk.rows`/`blk.cols`, not just the count.

**2. Stale module docstring (`_minimize.py`, header, ~lines 21–23)**

The module header still states: *"Constraint Jacobian is dense by design — sparse Jacobians belong on the `Problem` API."* This PR makes the dict path honor sparse Jacobians, so that documented contract is now false. Worth updating to describe the new detect-by-return-type sparse behavior.

(For context: I confirmed the `.tocsr().tocoo()` canonical ordering is genuinely order-deterministic regardless of the input COO ordering, so the core alignment design is sound — these are the two gaps around it.)

---

jkitchin:
Superseded by #170 — I rebased this change onto current `main` (it was cut from an older base and would otherwise re-add files `main` has since removed) and added a follow-up commit addressing the review feedback above: the per-iteration guard now verifies the full COO row/col pattern, not just the nnz count, so a same-count, shifted-position Jacobian raises instead of silently feeding Ipopt misaligned derivatives. Your original commit is carried forward with authorship intact.

Thanks @jduerholt! Closing in favor of #170.

---

PR #170, opened 2026-06-27
feat(minimize): honor sparse Jacobians from dict constraints (supersedes #164)

jkitchin:
## Summary

Supersedes #164 (from @jduerholt), rebased onto current `main`, with a follow-up fix for the review feedback on that PR.

`pounce.minimize`'s **dict-constraint** path declared a fully dense `m_rows × n` Jacobian sparsity pattern regardless of what the user's `jac` returned, densifying a genuinely sparse constraint Jacobian into the KKT system. This teaches the dict path to honor a sparse Jacobian (linear and nonlinear alike), mirroring how cyipopt's scipy interface infers sparsity by inspecting the `jac` return type.

## Commits

1. **feat(minimize): honor sparse Jacobians from dict constraints** — @jduerholt's original #164 change. `_block_from_dict` probes `jac(x0)`; if it returns a scipy-sparse matrix it declares that matrix's COO structure (canonicalized via `.tocsr().tocoo()` for a deterministic row-major order) instead of the dense grid, and `jac_values` streams the COO `.data` per iteration. A dense return — or no `jac` — keeps the original dense pattern, so existing behavior is unchanged.

2. **fix(minimize): verify full sparse-Jacobian pattern, not just nnz count** — the per-iteration guard originally checked only that the nonzero *count* matched the build-time structure. A `jac` that kept the same count but moved a nonzero to a different column passed the guard, and its values were silently misaligned against the declared `(row, col)` pattern and fed to Ipopt as wrong derivatives. The guard now verifies the full COO structure (row/col positions) — both sides are already canonicalized via `.tocsr().tocoo()`, so they compare directly. Also adds a position-change regression test and refreshes the stale module docstring that still claimed constraint Jacobians are dense by design.

## Tests

`python/tests/test_minimize.py` — 6 sparse/dense Jacobian tests (structure check, linear-as-sparse-dict, nonlinear sparse `jac`, count-change raises, **position-change raises**, dense-`jac` regression). Full `test_minimize.py`: **51 passed**.

https://claude.ai/code/session_011kxMxJGmxu6aLuJDbuZvL1

---
_Generated by [Claude Code](https://claude.ai/code/session_011kxMxJGmxu6aLuJDbuZvL1)_

### DIFF
--- a/python/pounce/_minimize.py
+++ b/python/pounce/_minimize.py
@@ -369,6 +369,12 @@ class _ConstraintBlock:
     lb: np.ndarray  # length n_rows
     ub: np.ndarray  # length n_rows
     n_rows: int
+    # Dict block whose ``jac`` returns a *scipy-sparse* matrix: ``(rows, cols)``
+    # above are that matrix's COO structure (canonicalized) and ``jac_values``
+    # streams the per-iteration COO data instead of a dense ``ravel()``. Lets a
+    # nonlinear (or linear) dict constraint declare a sparse Jacobian, mirroring
+    # cyipopt's detect-by-return-type behavior.
+    sparse_jac: bool = False
     # Pre-assembled ``sparse.coo_array`` for linear blocks. ``g_combined``
     # would otherwise rebuild this from ``(rows, cols, constant_vals)`` on
     # every Ipopt iteration (per-iter waste in the inner loop). ``None`` for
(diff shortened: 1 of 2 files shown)


jduerholt:
Thanks for fixing and merging, I totally forgot to move on here. Sorry!

PR #185, opened 2026-07-07
Initialization tooling: docs, check-x0 preflight, WarmStart, point generation, Pyomo helpers, skill/MCP

jkitchin:
Implements the full initialization-tooling roadmap (phases 0-5) on one branch, one commit per phase.

## What's here

**Phase 0 — docs.** `docs/src/initialization.md`: one reference page consolidating cold-start behavior (the bound_push interior clamp), the per-algorithm warm-start recipes and their traps, batch chaining, and bad-start diagnostics. Registered in SUMMARY.md.

**Phase 1 — x0 preflight.** One contract (`pounce.check-x0/v1`), three surfaces: `pounce check-x0 <model.nl> [--builtin N] [--x0-file F] [--json]` (exit 0 = evaluates cleanly, 21 = NaN/inf at x0), `pounce.preflight(problem_obj, x0, ...)` in Python, and a `check_x0` MCP tool. Reports non-finite evaluations, bound violations, interior-clamp displacement, initial constraint violation, and derivative scale spread. *Deviation from the original plan:* the core lives in `pounce-cli` (mirroring `pounce verify`), not `pounce-studio-core`, which is deliberately WASM-clean/serde-only and cannot evaluate NLPs.

**Phase 2 — `pounce.WarmStart`.** `WarmStart.from_info(x, info)` captures primal, duals, mu, and SQP working set; `solve(warm_start=ws)` / `minimize(..., warm_start=ws)` apply the whole enabling-option recipe; `save()/load()` (npz) carry it across processes. HS071: 11 cold iterations → 4 warm. Also wires the previously parsed-but-ignored `warm_start_mult_bound_push` in the Rust warm-start initializer (bound multipliers floored, matching upstream); `entire_iterate`/`same_structure` stay parse-for-compat, now documented as such.

**Phase 3 — generation/repair.** `pounce.generate_starts` (the find_minima sampler refactored public, plus a bounds-midpoint strategy), `pounce.project_to_feasible` (min-norm linearized repair via one convex QP), `pounce.race_starts` (truncated-run tournament returning OptimizeResults, composable with WarmStart). *Deviation:* the bounds-aware cold start is a Python sampling strategy, not a Rust `initializer=midpoint` solver option; FBBT integration stays future work.

**Phase 4 — Pyomo.** `pyomo_pounce.preflight(model)` (evaluates the model exactly as the NL writer sends it — unset values as 0 — and restores it), `initialize_missing_values`, and experimental `block_initialize` (Dulmage-Mendelsohn-ordered equality-block solves via `pyomo.contrib.incidence_analysis`; 1x1 blocks by Newton, larger blocks as square POUNCE subsystem solves).

**Phase 5 — skill/MCP.** The studio skill gains a check-x0 section and an initialization playbook recipe; the MCP server gains `suggest_initialization` (deterministic advisory suggestions from the preflight, sharpened by an optional prior solve report).

## Verification

- Rust: full workspace `cargo test` (excluding the HSL-linked `pounce-hsl`, no MA57 locally): **1888 passed, 0 failed**, including 7 new check-x0 unit tests + 4 binary integration tests.
- Python: **553 passed, 3 skipped** (full `python/tests`, native extension rebuilt), including new `test_preflight`, `test_warm_start_object`, `test_starts` suites; `hs071_warm_start.py` runs and shows 11 → 4.
- pyomo-pounce: 13 new tests pass (subsystem-solve case exercises the real binary).
- studio/mcp: 83 passed; both new tools smoke-tested end-to-end.
- `mdbook build docs` and `scripts/check-docs-consistency.sh` pass (38 pages, no dead links).

## Pre-existing failures surfaced (not caused by this branch)

- `pyomo-pounce/tests/test_pounce.py::test_options_forwarded`: the convex QP routing path ignores `max_iter=0` (`pounce quad.nl -AMPL max_iter=0` reports optimal via the convex IPM; `solver_selection=nlp` honors the limit). These tests don't run in CI, so this rotted silently. Deserves its own issue.
- `studio/mcp/tests/test_gams_tools.py::test_run_gams_problem_solves_hs071`: needs a registered GAMS link in the environment (gams exit 6 here).

### DIFF
--- a/crates/pounce-algorithm/src/init/warm_start.rs
+++ b/crates/pounce-algorithm/src/init/warm_start.rs
@@ -18,10 +18,14 @@
 //!   apply the same multiplier clamps.
 //!
 //! Wired options today: `bound_push`, `bound_frac`,
-//! `slack_bound_push`, `slack_bound_frac`, `mult_init_max`,
-//! `target_mu`. The remaining knobs (`mult_bound_push`,
-//! `entire_iterate`, `same_structure`) are stored but not yet
-//! consumed.
+//! `slack_bound_push`, `slack_bound_frac`, `mult_bound_push`,
+//! `mult_init_max`, `target_mu`. `mult_bound_push` floors the four
+//! bound-multiplier blocks (mirroring upstream's `ElementWiseMax`
+//! with `warm_start_mult_bound_push`): a user-seeded `z = 0` would
+//! otherwise start the barrier on its boundary. The remaining knobs
+//! (`entire_iterate`, `same_structure`) are parsed for Ipopt option
+//! compatibility but not yet consumed — they require the
+//! `GetWarmStartIterate` TNLP surface, which pounce does not expose.
 
 use crate::alg_builder::WarmStartOptions;
 use crate::init::default::push_x_into_interior;
@@ -84,23 +88,32 @@ impl IterateInitializer for WarmStartIterateInitializer {
             seed_from_nlp(data, nlp, &self.opts);
         }
 
-        if self.opts.mult_init_max > 0.0 {
+        if self.opts.mult_init_max > 0.0 || self.opts.mult_bound_push > 0.0 {
             // Rebuild `curr` with clamped multipliers. Components are
             // shared via `Rc` with previous solves, so we make fresh
             // copies before mutating to avoid clobbering downstream
-            // borrowers.
+            // borrowers. Bound multipliers are additionally floored at
+            // `mult_bound_push` (upstream `warm_start_mult_bound_push`):
+            // the barrier needs them strictly positive, and a carried-in
+            // 0 (e.g. an inactive bound in the previous solution) would
+            // otherwise start on the boundary.
             let mut borrow = data.borrow_mut();
             let curr = borrow.curr.as_ref().unwrap();
-            let cap = self.opts.mult_init_max;
+            let cap = if self.opts.mult_init_max > 0.0 {
+                self.opts.mult_init_max
+            } else {
+                f64::INFINITY
+            };
+            let z_floor = self.opts.mult_bound_push.max(0.0);
             let new_curr = IteratesVector::new(
                 Rc::clone(&curr.x),
                 Rc::clone(&curr.s),
                 clone_clamped(&curr.y_c, -cap, cap),
                 clone_clamped(&curr.y_d, -cap, cap),
-                clone_clamped(&curr.z_l, 0.0, cap),
-                clone_clamped(&curr.z_u, 0.0, cap),
-                clone_clamped(&curr.v_l, 0.0, cap),
-                clone_clamped(&curr.v_u, 0.0, cap),
+                clone_clamped(&curr.z_l, z_floor, cap),
+                clone_clamped(&curr.z_u, z_floor, cap),
+                clone_clamped(&curr.v_l, z_floor, cap),
+                clone_clamped(&curr.v_u, z_floor, cap),
             );
             borrow.set_curr(new_curr);
         }
@@ -280,6 +293,20 @@ mod tests {
         let out = clone_clamped(&v, 0.0, 1.0);
         assert!((out.max() - 0.5).abs() < 1e-15);
         assert!((out.min() - 0.5).abs() < 1e-15);
+    }
+
+    #[test]
+    fn mult_bound_push_floors_zero_bound_multipliers() {
+        // A carried-in z = 0 (inactive bound in the previous solution)
+        // must be floored at warm_start_mult_bound_push, matching
+        // upstream's ElementWiseMax — the barrier needs z > 0.
+        let v = dense(3, 0.0);
+        let out = clone_clamped(&v, 1e-3, 1e6);
+        assert!((out.min() - 1e-3).abs() < 1e-18);
+        // Values already above the floor pass through.
+        let v2 = dense(3, 0.7);
+        let out2 = clone_clamped(&v2, 1e-3, 1e6);
+        assert!((out2.max() - 0.7).abs() < 1e-15);
     }
 
     #[test]
--- a/crates/pounce-cli/src/check_x0.rs
+++ b/crates/pounce-cli/src/check_x0.rs
@@ -312,8 +312,7 @@ fn load_model(args: &CheckX0Args) -> Result<LoadedModel, String> {
         .nl
         .as_ref()
         .ok_or("expected a <problem.nl> argument or --builtin <name>")?;
-    let bytes =
-        std::fs::read(path).map_err(|e| format!("cannot read {}: {e}", path.display()))?;
+    let bytes = std::fs::read(path).map_err(|e| format!("cannot read {}: {e}", path.display()))?;
     let sha = sha256::hex(&bytes);
     let prob = nl_reader::read_nl_file(path)?;
     let var_names = prob.var_names.clone();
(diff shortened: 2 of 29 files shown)


jkitchin:
## What this PR adds, and how to use it

A user-facing tour of the new initialization tooling. Everything below ships in this PR; the reference chapter is [`docs/src/initialization.md`](https://github.com/jkitchin/pounce/blob/init/docs-initialization-page/docs/src/initialization.md).

### 1. Check a starting point before solving (`check-x0`)

```sh
pounce check-x0 model.nl                 # human-readable report
pounce check-x0 model.nl --json          # machine-readable (pounce.check-x0/v1)
pounce check-x0 model.nl --x0-file cand.txt   # check a candidate point instead
pounce check-x0 --builtin rosenbrock
```

One evaluation of each callback, no solve. It reports: NaN/inf in f, gradient, constraints, Jacobian, or Hessian at x0 (**exit 21**: a solve would abort with `Invalid_Number_Detected`); bound violations of x0 and components sitting exactly on a bound; how far the `bound_push` interior clamp will silently move the point; initial constraint violation per row; and gradient/Jacobian magnitude spread (the early scaling signal). Exit 0 means the model evaluates cleanly (warnings allowed).

The same check from Python, for callback-defined problems:

```python
report = pounce.preflight(problem_obj, x0, lb=lb, ub=ub, cl=cl, cu=cu)
print(report)          # report.fatal, report.warnings, report.to_dict()
```

### 2. Warm starts as one object (`pounce.WarmStart`)

Passing a previous solution as `x0` is not a warm start: the default `mu_init` re-walks the barrier and the `warm_start_*` pushes shove an at-the-bound solution off its bounds. `WarmStart` packages the whole working recipe:

```python
x, info = prob.solve(x0=x0)                  # cold solve
ws = pounce.WarmStart.from_info(x, info)     # captures x, duals, mu (and the SQP working set)
x2, info2 = prob.solve(warm_start=ws)        # warm re-solve
ws.save('state.npz'); ws = pounce.WarmStart.load('state.npz')   # across processes
```

Also accepted by `pounce.minimize(..., warm_start=ws)`. On HS071 the re-solve drops from 11 iterations to 4 (`python/examples/hs071_warm_start.py` shows the one-liner and what it does under the hood). Explicit `x0`/seed arguments override the captured ones. Bonus fix: the Rust warm-start initializer now honors `warm_start_mult_bound_push` (previously parsed but ignored), flooring seeded bound multipliers so a carried-in `z = 0` cannot wedge the barrier.

### 3. Generate or repair starting points

```python
starts = pounce.generate_starts(16, bounds=bounds, seed=0)      # sobol / uniform / jitter / midpoint
x0 = pounce.project_to_feasible(problem_obj, x0, lb=lb, ub=ub, cl=cl, cu=cu)  # min-norm linearized repair (one convex QP)
best = pounce.race_starts(fun, starts, bounds=bounds, iters=10)[0]            # cheap tournament
res = pounce.minimize(fun, best.x, warm_start=pounce.WarmStart.from_info(best.x, best.info))
```

`generate_starts` is the sampler that already powered `find_minima`, now public and composable (feed it to `solve_nlp_batch`, `race_starts`, or a loop of `minimize` calls).

### 4. Pyomo helpers (`pyomo-pounce`)

The classic trap: a `Var` with no `.value` is written as **0** into the `.nl` file, so an uninitialized model starts at the origin.

```python
import pyomo_pounce

report = pyomo_pounce.preflight(model)          # evaluates the model exactly as the
if report.fatal: ...                            # NL writer will send it (unset = 0),
                                                # restores the model untouched
pyomo_pounce.initialize_missing_values(model)   # bounds-aware fill (midpoint / zero)
rep = pyomo_pounce.block_initialize(model)      # experimental: Dulmage-Mendelsohn-ordered
                                                # equality-block solves fill Var.value
                                                # (1x1 by Newton, kxk via POUNCE)
```

`block_initialize` needs `pyomo.contrib.incidence_analysis` (`pip install networkx scipy`); fix a variable to treat it as a known input, and follow with `initialize_missing_values` for the under-determined remainder.

### 5. Agent surfaces (studio skill + MCP)

The studio skill gained an initialization playbook (preflight, interpret, fix with the matching surface, re-solve, confirm via `diagnose`), and the MCP server two tools: `check_x0` (the preflight over MCP) and `suggest_initialization` (runs the preflight and returns ordered advisory suggestions: options to set, Python/Pyomo calls to make, doc pointers, sharpened by an optional prior solve report). In a Claude Code session with the skill installed (`make install-skill`), "this model will not converge, check my initialization" now has a mechanical path.

### A 60-second workflow

```sh
pounce check-x0 model.nl        # fatal? fix values/domain. warnings? read them
pounce model.nl --json-output r.json --json-detail full
pounce-studio diagnose r.json   # restoration findings gone after re-init = confirmed
```

and for every re-solve of a related problem, capture a `WarmStart` and pass it back.

devin-griff:
This is useful. The warm-start and preflight pieces look complete; these comments are all about the Pyomo layer. Using a distillation column as the running example, by hand I initialize one like this: set the feed, reflux, and boilup, guess holdups and compositions on the trays, solve for a physical column profile with those decisions held constant, then let the optimizer move the decisions.

1. `block_initialize` supports that workflow but doesn't express it. Consider a `decisions=` argument that fixes the flagged variables at their current values, verifies the remaining system is square, solves it, and unfixes afterward. When the flagged set does not square the system, it would help to report which variables are unmatched rather than just counts. The Dulmage-Mendelsohn decomposition you already compute contains those names, and that diagnostic is most of the value for figuring out what I forgot to specify.

2. How does this relate to `solve_strongly_connected_components` in `pyomo.contrib.incidence_analysis`? That function already solves a square system block by block, with small blocks by Newton and larger ones by a solver. If `block_initialize` built on it, this code would reduce to the Dulmage-Mendelsohn partitioning plus the seeding.

3. `initialize_missing_values` can produce internally inconsistent fills, for example mole fractions that do not sum to one. `project_to_feasible` looks like the intended fix, but it is not reachable from the Pyomo layer. A fill, then repair, then block-solve pipeline would complete the story.

jkitchin:
@devin-griff Thanks — all three points are addressed in 1fa4a9b, and your distillation workflow is now the documented shape of the API.

**1. `decisions=` + names.** `block_initialize(model, decisions=[m.feed, m.reflux, m.boilup])` fixes the flagged variables at their current values (each must have one — `ValueError` otherwise), solves, and releases them in a `finally`; already-fixed variables stay fixed. Squareness is now first-class: `report.square`, with `report.underconstrained_variables` and `report.overconstrained_constraints` carrying the **names** from the Dulmage-Mendelsohn partition (previously discarded down to counts). The square part is still solved best-effort when the system isn't square, so you get the profile *and* the list of what you forgot to specify.

**2. `solve_strongly_connected_components`.** You're right, and the module now reduces to exactly what you described: DM extraction of the square part, decision handling, seeding, and diagnostics, with the block-by-block solve delegated to Pyomo's own machinery (`create_subsystem_block` + `TemporarySubsystemManager` around `solve_strongly_connected_components`). One behavioral trade-off, noted in the docstring: the SCC solve fails fast, so a failed block leaves downstream blocks at their seeds instead of the old keep-going-per-block collection.

**3. Fill → repair → block-solve.** `pyomo_pounce.project_to_feasible(model)` is the Pyomo-reachable repair: it swaps in `min Σ(v − v0)²` over the anchored (valued, unfixed) variables, solves against the model's own constraints and bounds with POUNCE (the full nonlinear projection, not the NumPy layer's linearized one), and restores the original objective. `pyomo_pounce.initialize(model, decisions=...)` chains the pipeline. Implementing it caught a subtlety worth flagging: the decisions must be held across **all** stages, not just the block solve — otherwise the min-norm projection drifts the feed/reflux you just specified and the block solve then holds them at the drifted values. The pipeline fixes decisions up front and releases them at the end.

Your mole-fraction example is a test verbatim (`test_repair.py::test_project_repairs_inconsistent_fill`: midpoint fill gives Σx = 1.5, projection lands each xᵢ at 1/3), and the split-plus-composition pipeline test asserts the decisions come back unfixed at their original values. Docs: the Pyomo page's initialization section is rewritten around the pipeline.

devin-griff:
Awesome. Pumped to try it.

jkitchin:
> Awesome. Pumped to try it.

Thanks for the suggestion, let me know if there are any issues with it!

PR #197, opened 2026-07-11
Fix silent drop of sensitivity/reduced-Hessian on specialized solve paths (#196)

### LINKED ISSUE #196: sIPOPT suffixes silently ignored when the convex-QP fast path triggers
When an .nl declares the sIPOPT suffixes (`sens_state_1`, `sens_state_value_1`, `sens_init_constr`), pounce runs the post-optimal sensitivity step and returns `sens_sol_state_1` as documented — but only on the general IPM path. If presolve routes the problem to the convex-QP path (banner: `POUNCE convex QP IPM (pounce-convex)`), the suffixes are silently ignored and no `sens_sol_state_1` is written to the .sol.

**Reproduction**

`min (x - p)^2 + y^2` with `p` pinned by an equality constraint and the sIPOPT suffixes declared in the .nl:

- As stated (pure convex QP) -> routed to the QP path, .sol contains no `sens_sol_state_1`.
- Same model plus `1e-6 * exp(y)` in the objective (forces the general IPM path) -> `sens_sol_state_1` returned, and the perturbed primal is correct to the analytical sensitivity (dx*/dp = 1: p 1.0 -> 1.5 gives x* 1.0 -> 1.5).

Environment: pounce 0.7.0 (pounce-solver wheel), Windows, .nl written by Pyomo with symbolic labels; suffixes verified present in the .nl (`S0/S4/S1` blocks).

**Expected**: either the QP path also runs the post-optimal sensitivity step, or pounce warns that declared sens suffixes were ignored. The silent drop is the painful part: it hits exactly the problems where the sensitivity step is cheapest.

jkitchin:
Fixes #196.

## Problem

The sIPOPT parametric-sensitivity step (`sens_state_1` / `sens_state_value_1` / `sens_init_constr` suffixes) and the reduced-Hessian computation (`--compute-red-hessian`) were honored **only** on the general NLP filter-IPM path. Several fast paths early-return before that post-optimal wiring and dropped the request **silently**. The reported case: a pure convex QP declaring the sens suffixes routes to the pounce-convex solver under `solver_selection=auto`, and no `sens_sol_state_1` is written to the `.sol`.

An audit for the same bug-class (early-return dispatch skipping a general-path feature) turned up three more instances, all fixed here.

## Fixes

| Path | Before | After |
|---|---|---|
| **Convex LP/QP/QCQP routing** (`main.rs`) | `auto` sent sens/red-Hessian solves to pounce-convex, silently dropping them | `auto` declines the fast path → NLP path honors the request; explicit convex `solver_selection` warns instead of silently skipping; banner names the path that actually runs |
| **`--minima` multistart** (`main.rs`) | early-return skipped sens/red-Hessian silently | warns it is not computed in a multistart search |
| **`--minima` `.sol` duals** (`minima/`) | wrote a zero placeholder for constraint multipliers | recovers the real base-problem duals per minimum via a clean, budget-exempt re-solve (avoids the augmented penalty/tunnel-objective multiplier trap); verified to match the NLP-path dual |
| **Python `pounce.minimize` convex route** (`_minimize.py`) | a user `callback` was silently dropped (it is a named arg, not in `options`, so the drop-warning missed it) | `callback` now listed in the convex-route dropped-options warning |

## Design notes

- **Convex reroute vs. warn.** Under `auto` we prefer correctness (route to the path that honors the request); under an *explicit* convex force we respect the user's choice but stop being silent. Mirrors how the CLI and Python already surface dropped options.
- **`--minima` duals correctness.** The multistart archive can accept a point found by a penalty/tunnel solve on an *augmented* objective, whose multipliers are **not** the base problem's duals. Naively capturing them would write wrong duals — worse than the honest zeros. So the fix recovers duals via a clean re-solve seeded at each reported minimum (already optimal → converges immediately), replicating the NLP path's `finalize_solution_lambda` convention.

## Tests

- `qp_sens_dispatch.rs` — `auto` reroutes and writes `sens_sol_state_1 = 1.5` (matches the analytical `dx*/dp = 1`); explicit `qp-ipm` warns and skips; `nlp` control.
- `minima_sol_duals.rs` — `--minima` best-minimum dual (−2) matches the NLP-path dual, not the old zero.
- Python autoroute callback-warning test.
- Fixture `convex_qp_sens.nl` is the issue's exact model (`min (x−p)² + y²`, p pinned, perturbed to 1.5), generated with pyomo.

Full `pounce-cli` suite: 278 passed. Python autoroute suite: green. rustfmt + clippy clean on changed files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,30 @@ changes.
 
 ### Fixed
 
+- **Post-optimal requests are no longer silently dropped on the specialized
+  solve paths (#196).** When an `.nl` declared the sIPOPT sensitivity suffixes
+  (`sens_state_1` / `sens_state_value_1` / `sens_init_constr`) or the solve
+  asked for a reduced Hessian (`--compute-red-hessian`), the request was
+  honored only on the general NLP filter-IPM path. Three fast paths bypassed it
+  without a word:
+  - **Convex LP/QP/QCQP routing.** Under `solver_selection=auto`, a problem
+    that classifies as LP / convex-QP was sent to the pounce-convex solver,
+    which has no sensitivity / reduced-Hessian machinery, so no
+    `sens_sol_state_1` was written. `auto` now declines the fast path and
+    routes such a solve to the NLP path (which honors the request); an
+    *explicit* convex `solver_selection` still runs convex but now warns that
+    the request is skipped.
+  - **`--minima` multistart.** The `--minima` early-return skipped the
+    post-optimal step entirely; it now warns that sensitivity / reduced-Hessian
+    is not available in a multistart search.
+  - **`--minima` `.sol` duals.** The multistart `.sol` wrote a zero placeholder
+    for the constraint multipliers; it now recovers the real base-problem duals
+    at each reported minimum (via a clean re-solve, so a point accepted from an
+    augmented penalty/tunnel solve still gets the base problem's multipliers).
+  - **Python `pounce.minimize` convex route.** A user `callback` cannot fire on
+    the convex/SOCP route (the solver consumes the extracted quadratic form and
+    never calls back into Python); it is now surfaced in the dropped-options
+    warning rather than silently ignored.
 - **Registered-but-unread algorithmic tuning options are now honored (#191).**
   A range of options were registered but never read, so the solver always ran
   with the hard-coded defaults and any user-set value was silently dropped.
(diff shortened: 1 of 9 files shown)


devin-griff:
Thanks!

jkitchin:
Thanks for reporting this. I found a few similar issues because of it! New version of pounce will be landing in the next day or two with this!

PR #199, opened 2026-07-11
Declared-parameter sensitivity for Pyomo (declare_sens_param / gradient / estimate)

devin-griff:
This provides a fix for longstanding issues with sIPOPT and the pyomo sensitivity toolbox. The main themes are 1) a better pyomo user interface for sensitivity calculations and 2) a sensitivity solve step that is separate from the full NLP solve, thereby truly enabling things such as advanced step NMPC.

## What it adds

A declare-then-query sensitivity interface in `pyomo_pounce`:

```python
from pyomo_pounce import declare_sens_param, gradient, estimate

declare_sens_param(m.p)                # a flag while building the model:
                                       # no perturbed values required
SolverFactory("pounce").solve(m)       # ordinary solve, keeps the KKT factor

gradient(m.x, wrt=m.p)                 # exact dx*/dp (float)
gradient(m.con, wrt=m.p)               # d(multiplier of con)/dp
G = gradient(m.z, wrt=m.r)             # containers -> Gradient object
G[m.z[1], m.r[2]]; G.to_dataframe()    # element access / full Jacobian
estimate(m, [(m.p, 2.5)])              # first-order estimate at new values,
                                       # bound-clamped, warns on active-set risk
```

When declarations are present, the solve runs in-process (Pyomo writes the .nl, `read_nl` evaluates it through a cyipopt-style bridge into the `Solver` session) and the converged factorization answers every later query as a single backsolve. Perturbation values are never needed at solve time: gradients come from unit-delta backsolves, and `estimate` combines the stored columns for arbitrary values afterwards, matching the separation that advanced-step NMPC needs. Models without declarations use the AMPL/CLI path unchanged.

Parameters ride as pinned variables via `pyomo.contrib.sensitivity_toolbox`'s model surgery; the per-model registry is deepcopy-aware so re-solves and `model.clone()` stay clean.

## Rust/bindings additions

`Solver::parametric_step_full` (full compound-vector step, exposing multiplier sensitivities) and `Solver::g_multiplier_rows` in `pounce-sensitivity`, with Python bindings `parametric_step_full` and `multiplier_rows`.

## Definition of done

- 7 pytest tests in `pyomo-pounce/tests/test_sens.py`: primal and multiplier gradients validated against finite-differenced re-solves, estimate accuracy vs a true re-solve, bound-clamp warning, container/Jacobian forms, plain-solve passthrough, re-solve/clone cleanliness.
- CHANGELOG entry under Unreleased (Python / Pyomo surfaces).
- Book section: `docs/src/sensitivity.md` gains a Pyomo entry point (intro updated to four entry points).
- README: fourth entry-point bullet in the sensitivity section.
- Executed walkthrough notebook `python/notebooks/25_pyomo_sensitivity.ipynb` (quad-tank optimal control; the first-move gradients are the NMPC feedback gains; needs `pyomo-cvp`), indexed in the notebooks README.

## Notes for review

- Developed and tested on Windows. Pre-existing local failures are unchanged from main and unrelated: the two `test_bvp` failures, the hs71 timing-report test, and `pounce-hsl` not linking without HSL libraries.
- clippy adds nothing beyond the existing pymethods `useless_conversion` baseline; `cargo fmt` clean.
- Related: #196 (QP fast path skips the sIPOPT suffixes; independent of this PR since this path always uses the general IPM).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,29 +198,47 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - run: python -m pip install --upgrade pip build
-      - name: Build wheel + sdist
+      - name: Build pyomo-pounce wheel + sdist
         working-directory: pyomo-pounce
         run: |
           python -m build
           ls -l dist/
-      - name: Build pounce CLI (stand-in for pounce-solver dep)
+      # Build the pounce-solver wheel from THIS branch and install it, so the
+      # in-process sensitivity session (`pyomo_pounce.sens` does `import pounce`
+      # and calls the new solver methods) has a real Python solver — not just
+      # the CLI. The wheel bundles the CLI, so its console script also satisfies
+      # the SolverFactory plugin's `.solve()` path. The pyomo-pounce sensitivity
+      # tests exercise both.
+      - name: Build pounce CLI binary
         run: cargo build --release --bin pounce
+      - name: Stage CLI binary into pounce-solver wheel
+        run: |
+          mkdir -p python/pounce/bin
+          cp target/release/pounce python/pounce/bin/pounce
+      - name: Build pounce-solver wheel
+        uses: PyO3/maturin-action@v1
+        with:
+          working-directory: python
+          target: x86_64
+          manylinux: manylinux2014
+          args: --release --out dist
+          sccache: "true"
       - name: Install + smoke
         run: |
-          python -m pip install pyomo
-          # Install without deps because `pounce-solver` may not yet be
-          # published; the locally-built CLI on PATH satisfies the
-          # runtime requirement.
+          # pyomo + numpy (sensitivity needs numpy); the locally-built
+          # pounce-solver wheel provides the `pounce` Python package + CLI.
+          python -m pip install pyomo numpy
+          python -m pip install python/dist/*.whl
+          # pyomo-pounce itself installed without deps: its declared
+          # pounce-solver dep is satisfied by the wheel installed just above.
           python -m pip install --no-deps pyomo-pounce/dist/*.whl
-          export PATH="$PWD/target/release:$PATH"
           which pounce
           python -c "import pyomo_pounce; from pyomo.opt import SolverFactory; s = SolverFactory('pounce'); print('exe:', s.executable())"
       - name: Run pyomo-pounce tests
         run: |
           python -m pip install pytest
-          # The plugin resolves `pounce` from PATH, so re-export it here
-          # (each `run:` is a fresh shell). These tests drive real solves
-          # through the AMPL/`.nl` layer — option forwarding, solver routing —
-          # which is what catches regressions the import smoke test misses.
-          export PATH="$PWD/target/release:$PATH"
+          # Sensitivity tests (test_sens.py) drive real in-process solves +
+          # parametric backsolves through the installed pounce-solver; the
+          # AMPL/`.nl` tests exercise the CLI plugin path. Both catch
+          # regressions the import smoke test misses.
           python -m pytest pyomo-pounce/tests -q
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,7 +227,7 @@ jobs:
         run: |
           # pyomo + numpy (sensitivity needs numpy); the locally-built
           # pounce-solver wheel provides the `pounce` Python package + CLI.
-          python -m pip install pyomo numpy
+          python -m pip install pyomo numpy pandas
           python -m pip install python/dist/*.whl
           # pyomo-pounce itself installed without deps: its declared
           # pounce-solver dep is satisfied by the wheel installed just above.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,25 @@ changes.
 
 ## [Unreleased]
 
+### Added — declared-parameter sensitivity for Pyomo (Python / Pyomo)
+
+- **`pyomo_pounce` sensitivity interface.** Declare parameters while
+  building the model (`declare_sens_param(m.p)` — a flag, no perturbed
+  values), solve normally with `SolverFactory('pounce')`, then query:
+  `gradient(m.x, wrt=m.p)` (exact dx*/dp; equality constraints give
+  their multiplier's derivative), container/Jacobian access via the
+  `Gradient` object, and `estimate(m, [(m.p, value)])` for
+  first-order perturbed-solution estimates with bound clamping and an
+  active-set warning. When declarations are present the solve runs
+  in-process through the `pounce.Solver` session (`read_nl` + callback
+  bridge) and keeps the converged KKT factorization; models without
+  declarations use the CLI path unchanged.
+- **Python `Solver` session additions**: `parametric_step_full`
+  (full KKT-space step, exposing multiplier sensitivities) and
+  `multiplier_rows` (map constraint indices to their `y_c` rows), with
+  matching Rust methods `Solver::parametric_step_full` /
+  `Solver::g_multiplier_rows` in `pounce-sensitivity`.
+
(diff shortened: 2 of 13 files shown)


jkitchin (commented):
## Review — Declared-parameter sensitivity for Pyomo

### Overview
Adds a declare-then-query sensitivity interface to `pyomo_pounce` (`declare_sens_param` / `gradient` / `estimate`). When a model has declared params, `SolverFactory("pounce").solve()` diverts to an in-process path: the sensitivity toolbox pins the params as variables, the clone is written to `.nl`, evaluated through `read_nl` + a cyipopt-style bridge into a `pounce.Solver`, and the converged KKT factorization answers gradient/estimate queries as backsolves. Backed by two new Rust methods (`parametric_step_full`, `g_multiplier_rows`) and their Python bindings.

Overall this is a well-structured, well-documented PR that mirrors existing conventions closely. The Rust additions are clean and consistent with the existing `parametric_step` binding (same validation, error mapping, `into_pyarray_bound`). The deepcopy-aware registry with an explicit clone/re-solve test is a nice touch. My concerns are concentrated in the Python in-process path — which is also, per the PR notes, the surface CI can't exercise and that was developed/tested only on Windows.

### Issues & risks

**1. Temp directory leak — every solve (medium).** `sens_solve` does `tmp = tempfile.mkdtemp(prefix="pounce_sens_")` and never removes it. Each solve leaves behind a dir with `model.nl/.col/.row`. The headline use case is advanced-step NMPC — *repeated* solves in a loop — so this accumulates unboundedly. After the solve returns, the `.nl` file isn't needed (bounds are already materialized as arrays in `session.nl.x_l/x_u`), so cleanup is safe. Recommend `shutil.rmtree(tmp, ignore_errors=True)` at the end, or an `atexit`/finalizer.

**2. `solve()` return type diverges by path (medium).** `POUNCE.solve` returns the plain `info` dict from `sens_solve` when declarations are present, but `super().solve()` returns a Pyomo `SolverResults`. Any caller doing `res = solver.solve(m); res.solver.termination_condition` works on the plain path and silently breaks on the declared path. No test checks the declared-path return value, so this is untested. Consider returning a `SolverResults` (or at minimum documenting the difference prominently).

**3. Unclosed file handles (low, but Windows-relevant).** In `sens_solve`:
```python
var_names = open(nl_path[:-3] + ".col").read().splitlines()
con_names = open(nl_path[:-3] + ".row").read().splitlines()
```
These rely on CPython refcounting to close. Use `Path(...).read_text()` or a `with` block. On Windows (the stated dev/test platform) a lingering handle can also block the tempdir cleanup from #1.

**4. Keyword-model call misses the fast path (low).** In `pounce_solver.py`, `if args and has_declarations(args[0]):` — a `solve(model=m)` keyword form has empty `args` and would silently fall through to the CLI path, ignoring declarations. Pyomo convention is positional, so low risk — worth a `kwds.get("model")` fallback or a comment.

**5. Fixed absolute bound tolerance in `estimate` (low).** The clamp check uses a hardcoded `1e-9` regardless of variable scale. Fine for the tested models; could misfire on very large/small-scaled variables. The docs claim it's "the same single-pass projection as `--sens-boundcheck`" — worth confirming that matches.

### Style / conventions
- **Notebook README ordering:** entry `25` is inserted between `04` and `09`, breaking the otherwise-numeric table order (03, 04, **25**, 09, 13, …). Minor.

### Test coverage
Good: FD-validated primal *and* multiplier gradients, `estimate` vs true re-solve, the clamp warning, container/`to_dataframe` forms, plain-solve passthrough, and clone/re-solve cleanliness. The multiplier test explicitly asserts the sign convention. The main gaps are the two untested behaviors above (return type; and no assertion that temp artifacts are cleaned).

### Merge-readiness
`mergeable_state` is `unstable` and the status API reports `total_count: 0` — no CI checks appear to have run on the head SHA. Given the in-process path is CI-untestable and Windows-only-tested, I'd want the standard pytest/clippy/fmt suite confirmed green on a runner before merge, plus a manual check on Linux/macOS since the tempfile + file-handle code is exactly where cross-platform issues surface.

**Bottom line:** Solid, cohesive feature with good tests and docs. Nothing blocks correctness of the core math, but I'd fix the temp-dir leak (#1) and the `solve()` return-type inconsistency (#2) before merge, and tidy the file handles (#3). The rest are nits.

---

devin-griff:
Thanks for the review. All five items addressed in 5f60c07:

1. Temp-dir leak: the .nl temp directory is now removed as soon as read_nl has parsed it (everything needed later lives in memory), with a repeated-solve test asserting no pounce_sens_* dirs accumulate.
2. Return type: the declared path now returns a Pyomo SolverResults (termination_condition, status, message, objective bounds) matching the plain path's shape, with a test.
3. File handles: .col/.row reads use Path.read_text.
4. solve(model=m) keyword form now reaches the declared path, with a test.
5. Clamp tolerance is scale-aware (1e-9 relative to variable magnitude), and the docs now say the clamp is analogous to --sens-boundcheck rather than the same as it (the CLI default eps is 1e-3 absolute, so equating them was wrong).

Also restored numeric order in the notebooks README table. Suite is 41 passing locally (10 in test_sens.py); happy to have CI approved for a runner pass whenever convenient.

jkitchin:
Confirmed — all five are resolved in `5f60c07`, each with a backing test:

1. **Temp-dir leak** — `read_nl` now runs inside the `try` and `shutil.rmtree(tmp, ignore_errors=True)` in `finally`; `test_no_temp_dir_leak` asserts nothing accumulates over repeated solves. The delete-before-`solver.solve()` ordering is safe because `read_nl` parses eagerly (the passing solve tests exercise the evaluators after the dir is gone). 
2. **Return type** — proper `SolverResults`; `optimal`/`ok` is correct since the non-converged case raises earlier.
3. **File handles** — `Path.read_text()`. 
4. **Keyword `model=`** — `args[0] if args else kwds.get("model")`, tested.
5. **Clamp tolerance** — scale-aware `1e-9 * max(1, |x|)`, and the "analogous to `--sens-boundcheck`" wording is the accurate framing.

Notebook README order restored too. Nothing further from me on the code — only outstanding item is a CI runner pass on the head SHA (fork PR shows no checks yet). Thanks for the quick turnaround.

---

jkitchin:
CI ran on `5f60c07` and one job fails: **`wheel smoke (pyomo-pounce)`**. Everything else is green (Rust build/clippy/fmt/test, torch tests, jax tests, `pounce-solver` wheel smoke).

**Failure:**
```
File ".../pyomo_pounce/__init__.py", line 24, in <module>
    from pyomo_pounce.sens import (
File ".../pyomo_pounce/sens.py", line 34, in <module>
    import numpy as np
ModuleNotFoundError: No module named 'numpy'
```

**Cause:** `sens.py` adds a top-level `import numpy as np`, and `__init__.py` now imports from `sens` unconditionally, so `import pyomo_pounce` hard-requires numpy at import time. But `numpy` isn't a declared dependency of `pyomo-pounce` (`dependencies = ["pyomo>=6.0", "pounce-solver>=0.1"]`) — it was only ever present transitively via `pounce-solver`. The smoke job installs the wheel with `--no-deps` and only adds `pyomo` explicitly, so numpy is absent and the bare import crashes. (In a real `pip install pyomo-pounce` numpy comes in via `pounce-solver`, so this is masked outside the `--no-deps` smoke — but the package is now importing numpy directly without declaring it.)

**Fix — two parts:**

1. Declare the direct dependency in `pyomo-pounce/pyproject.toml`:
   ```toml
   dependencies = [
       "pyomo>=6.0",
       "pounce-solver>=0.1",
       "numpy>=1.20",
   ]
   ```
2. Make the `--no-deps` smoke green by installing numpy explicitly (`.github/workflows/ci.yml`, the "Install + smoke" step):
   ```yaml
   python -m pip install pyomo numpy
   ```

Part 2 is what actually turns the job green (the smoke deliberately skips declared deps because `pounce-solver` may be unpublished); part 1 is the correctness fix so a direct `import numpy` isn't relying on a transitive dependency.

---

PR #203, opened 2026-07-12
One-solve parameter covariance and identifiability for Pyomo (declare_estimated / declare_residual / covariance)

devin-griff:
Follow-on to #199: parameter estimation gets the same declare-then-query treatment. Fit a least-squares model with ONE ordinary solve, then read the asymptotic parameter covariance and identifiability diagnostics off the held KKT factorization, with no perturbed values, no re-solving, and no information the model does not already contain.

## What it adds

```python
from pyomo_pounce import covariance, declare_estimated, declare_residual

m.A = pyo.Var(); m.k = pyo.Var()          # the fitted parameters, FREE
declare_estimated(m.A); declare_estimated(m.k)

m.r = pyo.Var(m.I)                        # residuals, one per data point
m.res = pyo.Constraint(m.I, rule=...)     # r[i] == y[i] - model(A, k, t[i])
declare_residual(m.r)

m.obj = pyo.Objective(expr=sum(m.r[i]**2 for i in m.I))
pyo.SolverFactory("pounce").solve(m)      # one solve

cov = covariance(m)                       # zero arguments
cov.std_err[m.A]; cov.correlation[m.A, m.k]; cov.matrix; cov.eigen()
```

- The noise variance is derived from the declared residuals (count and SSR both come from the container); `sigma_sq=` (known variance, scalar or per-group) and `n_data=` remain as alternatives.
- `declare_residual(..., group="conc")` partitions residuals into noise groups by arbitrary user strings: pooled variance by default, per-group variances and the heteroscedastic sandwich covariance when groups are labeled.
- All declarations also gain explicit call-time forms on `solve()` (`sens_params=`, `estimated=`, `residuals=`), mirroring the declared/explicit duality in pyomo-cvp.
- Guardrails: the solve warns when the declared residuals do not reproduce the objective (weights or regularization would corrupt the variance estimate); covariance() warns on an active-bound parameter and on a regularized (inertia-corrected) factor, and errors clearly on missing declarations or structurally dependent parameters.

## Mechanics and validation

The parameter block of the inverse KKT matrix, one `kkt_solve` backsolve per estimated parameter against the held factor, equals the inverse reduced Hessian of the eliminated problem; for a plain sum-of-squares objective, cov = 2 sigma^2 (K^-1)_pp. The constant is pinned empirically, not assumed:

- linear-regression fixture vs the analytical sigma^2 inv(X^T X): agreement to 5e-16 relative;
- estimated sigma_hat^2 from declared residuals vs the classical SSR/(n-p): machine precision;
- nonlinear (exponential-decay) case vs a finite-difference Hessian of the reduced objective: 1e-4;
- two-group sandwich vs its independent closed form (X^T X)^-1 (sum_g sigma_g^2 Xg^T Xg) (X^T X)^-1: 1e-8, which also validates recovering the per-group residual Jacobians from the same backsolves.

No Rust changes: everything rides the existing `kkt_solve`/`kkt_dim` session bindings, so this is a pure-Python review.

## Relation to pyomo.contrib.parmest

parmest is an estimation workflow (multi-experiment management, bootstrap, likelihood-ratio regions) that restructures the problem into its experiment framework and computes covariance by finite differences or an ipopt re-solve. This is a post-solve primitive: the model as written, one declaration per component, covariance and identifiability from the factorization the solve already produced. The book section and notebook state the positioning explicitly.

## Definition of done

- 12 tests in `pyomo-pounce/tests/test_covariance.py` (all of the validations above, plus explicit-form equivalence, error paths, warning paths, clone survival); full pyomo-pounce suite 53 passing.
- CHANGELOG entry under Unreleased (Pyomo surface).
- Book: `docs/src/sensitivity.md` gains "Parameter covariance and identifiability" including the parmest relation.
- README: covariance line added to the Pyomo entry-point bullet.
- Executed walkthrough notebook `python/notebooks/26_parameter_covariance.ipynb`: declare, one solve, `covariance(m)`; a 95% confidence ellipse validated by 1000 Monte Carlo refits (95.8% coverage); a late-window practical-unidentifiability diagnosis. Indexed in the notebooks README.

Developed and tested on Windows as before; CI approval appreciated whenever convenient.


jkitchin:
@devin-griff — reviewed this and pushed a follow-up commit (7a77b1f). Summary of what's in it, plus two things I'd like your call on.

## Fixed in the pushed commit

**1. Explicit-form registry accumulation (the one real bug).** The call-time form — `solve(m, estimated=..., residuals=..., sens_params=...)` — appended its components to the persistent per-model registry on *every* solve. Re-solving one model (the NMPC use case the module docstring already calls out as supported) double-counted the residuals: SSR and `n` both doubled, `sigma^2`/covariance drifted a few % per solve, and it fired the spurious "declared residuals give SSR=... but the objective value is ..." warning. `sens_solve` now builds solve-local `eff_params`/`eff_estimated`/`eff_residuals` (declared + explicit) and never writes the explicit ones back into `reg`. The declared form is unchanged. Regression test: `test_explicit_form_repeated_solve_is_stable`.

**2. Minor / test-quality**
- `test_n_data_fallback` didn't actually exercise the `n_data` branch — the `linear` fixture declares residuals, which take precedence, so `n_data` was silently ignored and the test passed trivially. Rewritten to use a model with `declare_estimated` but no `declare_residual`, validated against `sigma^2 (X^T X)^-1`.
- `covariance()` now **warns** when `n_data=` is passed but a higher-precedence source (`sigma_sq` or declared residuals) is present, instead of silently dropping it.
- A per-group dict `sigma_sq` with no *named* residual groups now raises a clear `ValueError` instead of a bare `KeyError: None`.

## Relation to `pounce.curve_fit` — needs a decision

While reviewing I compared this against `pounce.curve_fit`, which computes the same thing. They're the same scale-and-invert-the-reduced-Hessian recipe read from the held KKT factor, and they agree to ~1e-16 on a **linear** fit. But for **nonlinear** models they factor different Hessians and give genuinely different answers (~4% on a strongly-curved exponential fit I tried):

- `curve_fit` passes `gn_hessian` as the solver's Hessian callback → its factor is the **Gauss-Newton** Hessian → `pcov = 2 s^2 (J^T J)^-1`. This is the expected-information / scipy / `pycse.nlinfit` convention and is always PSD.
- `covariance()` here feeds the **exact Lagrangian Hessian** through the `.nl` bridge → the reduced block is the **observed-information** Hessian (includes the `-2 sum r_i grad^2 f` residual-curvature term GN drops). This is exactly what your `test_nonlinear_against_fd_hessian` pins against a full FD Hessian, so the test is (correctly) asserting the observed-information form.

Neither is "wrong" — it's the classic observed-vs-expected (Efron & Hinkley 1978) information choice — but it's worth being deliberate about, because:
- observed information can go **indefinite** (large/structured residuals) → negative variances, which is why this PR needs its negative-diagonal warning; GN structurally cannot;
- most users comparing against scipy/`nls`/`curve_fit` will expect the Gauss-Newton number.

I documented the distinction and cross-referenced the two tools in both directions (`sensitivity.md`, `curve-fitting.md`, and both docstrings). **Question for you:** do we want to leave them semantically different (observed here, GN in `curve_fit`) as a documented, intentional choice — or make `covariance()` match `curve_fit` and report Gauss-Newton for consistency (e.g. an optional `hessian="gauss-newton"|"observed"` knob)? My lean is to keep observed as the default here (it's the exact curvature of the model as posed, and it's what your validation test already encodes) but say so loudly, which is what the pushed docs do. Happy to add the GN option if you'd prefer parity.

## Also open (not touched in the commit)

**Active-bound handling.** `curve_fit` projects the covariance onto the active-constraint nullspace (`_projected_covariance`); `covariance()` only *warns* on a bound-active parameter and returns the raw inverse-KKT block. In practice the barrier makes the pinned direction come out ~0 anyway, but it isn't the guaranteed structural projection. Want me to port the projection over, or is warn-only fine for now?

Full pyomo-pounce covariance suite is green (16 passing). The 6 `test_sens.py` failures on my machine are a pre-existing local issue (installed `pounce` wheel predates the `parametric_step_full`/`multiplier_rows` bindings) — unrelated to this change and not seen in CI.

jkitchin:
@devin-griff Thanks for submitting this. It is moving towards completion. There is one detail I thought is worth discussing first. There is some overlap with curve_fit (which is fine, this PR is aimed at pyomo users, and curve_fit is a different convenience function) where there is not fully consistent behavior:

Active-bound handling. curve_fit projects the covariance onto the active-constraint nullspace (_projected_covariance); covariance() only warns on a bound-active parameter and returns the raw inverse-KKT block. In practice the barrier makes the pinned direction come out ~0 anyway, but it isn't the guaranteed structural projection. Want me to port the projection over, or is warn-only fine for now?

WDYT?

devin-griff:
Thanks for the review and the fixes. The registry accumulation was a real bug, and solve-local effective sets are exactly the semantics the explicit form was meant to have. Good catch on the n_data test as well.

On observed vs Gauss-Newton: agreed with your lean, and I implemented the option on the branch rather than leaving it open (4e7c93d). The default is unchanged; `covariance(m, hessian="gauss-newton")` rebuilds J from the residual rows of the same backsolves (`J = Z_r inv(M)`), so it costs no extra solve, and it raises a clear error without declared residuals. One naming deviation from your suggestion: the strings are `"lagrangian"` and `"gauss-newton"` rather than `"observed"`, so both values name Hessians, matching the kwarg; the observed/expected translation stays in the docs. Beyond scipy parity there is a concrete use for the GN form downstream: it is structurally PSD, which an MHE arrival-cost update needs, exactly where the exact Hessian can go indefinite under model mismatch.

On active bounds: ported, not warn-only. `covariance()` now projects a bound-active fitted parameter the same way `_projected_covariance` does (invert the free block of the information matrix, embed zeros), keeps the warning, and reports correlation entries involving the pinned parameter as 0 so nothing downstream divides by a zero standard error. Docs note the scope: only variable bounds on the fitted parameters themselves are detected.

Also on the branch: `declare_estimated` is renamed to `declare_fitted` (explicit kwarg `estimated=` to `fitted=`). The old name read as if it paired with `estimate()`, which is a sensitivity verb, and the fitting vocabulary lines up with `curve_fit`. Unreleased API, so no deprecation shim.

Suite is at 61 passing; notebook 26 re-executed with a lagrangian vs gauss-newton comparison section.

PR #206, opened 2026-07-13
pyomo_pounce: show the solver log under tee=True, including in Jupyter

devin-griff:
Two commits.

**1. The solver core emits its own console log.** The problem-statistics and end-of-run summary blocks move out of the CLI into the shared `pounce-solve-report::console` crate, and `pounce-algorithm` emits them itself (before and after the iteration table), gated on `print_level`. Every frontend (CLI, Python bindings, C interface) now gets the identical Ipopt-style log at `print_level >= 1`; `print_level 0` is silent. Eval tallies come from the solver's own counters via a new `IpoptNlp::eval_counts()` drained into `SolveStatistics`. Both the IPM (`optimize_tnlp`) and active-set SQP (`optimize_sqp_tnlp`) paths emit through two shared helpers (`emit_problem_stats`, `emit_end_summary`), so no algorithm or frontend carries duplicated presentation code. `pounce.print_banner()` is exposed for in-process frontends that print the up-front banner themselves. CLI stdout is byte-identical except the eval counts now report the solver's true count (one lower than before, which had included a frontend-side evaluation).

**2. pyomo-pounce streams the engine's log under `tee=True`.** On the declared/in-process solve path, `SolverFactory('pounce').solve(m, tee=True)` now streams the engine's own log (banner, problem statistics, iteration table, end-of-run summary) live to `sys.stdout`, including in Jupyter, instead of reproducing ~300 lines of CLI formatting in Python and buffering it to the end. The solve runs on the main thread (releasing the GIL) while a worker thread tails a redirect tempfile; a regular file, not a pipe, is tailed so a slow reader can never block the solver, and a raising solve still flushes its partial log before re-raising. `tee=False` stays silent. The results object keeps `solver.name` and the objective bounds, and `solver.time` measures the solve alone (excluding stream and decode).

Minor bump to `0.9.0`: the public surface genuinely changed (new `print_banner` pyfunction, new default core stdout, new `eval_counts()` trait method). All three version surfaces (`Cargo.toml`, `python/pyproject.toml`, `pyomo-pounce/pyproject.toml`) agree, and `scripts/check-release-consistency.sh` passes.


jkitchin (commented):
## Summary

On the declared/in-process solve path (`declare_sens_param` / `declare_fitted` / `declare_residual`), `SolverFactory("pounce").solve(m, tee=...)` now behaves like an ordinary Pyomo solver: `tee=True` reproduces the full Ipopt/CLI-style log (banner, problem stats, iteration table, end-of-run summary) including in Jupyter, `tee=False` is silent, the returned `SolverResults` is shaped to resemble the subprocess path, and a non-converged declared solve now *returns* a mapped results object (and clears the stale sensitivity session) instead of raising.

This is a genuinely useful UX fix and the tests are thoughtful (tee output, silence, results shape, non-convergence mapping, stale-session clearing, three-declaration coexistence). No blocking correctness bug. My recommendations are two design points plus a handful of smaller items; none are hard blockers, but the first two materially affect maintainability and UX.

---

## 1. Source the wrapper blocks from Rust instead of re-implementing the CLI

The ~300 new lines in `sens.py` (`_LOGO`, `_logo_rows`, `_tee_banner`, `_tee_problem_stats`, `_tee_summary`, `_fmt_ipopt`, `_STATUS_MESSAGES`) are a near-line-for-line transliteration of `crates/pounce-cli/src/print.rs` (`print_banner`, `print_problem_stats`, `print_summary`, `logo_rows`, `fmt_ipopt`, `status_message`) and its `ProblemStats` computation at `print.rs:58`. The banner/stats/summary are printed only by the CLI (`main.rs:484/1097/1290`); the algorithm crate emits just the iteration table, which is why the Python `Solver` binding never sees them.

The duplication is deliberate ("no Rust, CLI, or wheel change") but it creates a second copy of the CLI's presentation with no shared source of truth, and it already carries two concrete casualties:

- **`"linear solver FERAL"` is hardcoded** in the Python banner, whereas Rust passes the actual `backend_tag` (`main.rs:484`). With several linear-solver crates in the tree (`pounce-feral`, `pounce-hsl`, `pounce-linsol`), this line will silently misreport if the in-process build uses a different backend.
- **`_tee_problem_stats` recomputes the reduced-problem counts in Python** (assuming `fixed_variable_treatment=make_parameter`) rather than reading the `ProblemStats` the CLI already builds, so the printed variable/constraint counts can disagree with the engine's actual iteration table.

**Recommendation:** move the `print.rs` formatting + `ProblemStats` computation into a shared library crate — `pounce-solve-report` already exists for exactly this kind of thing — and have both the CLI and a PyO3-exposed function call it, so Python renders identical text with zero duplicated formatting. The cleanest variant is to push `print_problem_stats`/`print_summary` down into the algorithm's `output` layer gated on `print_level` (where Ipopt itself emits them), so the Python `Solver` with `print_level=5` gets them natively and the C interface benefits too. Note the `info` dict already exposes `status_msg` via the same Rust `status_message()` (`pounce-py/nlp_batch.rs:234`), so the re-hardcoded `_STATUS_MESSAGES` table is already redundant.

If merging as-is for expediency, at minimum drop the hardcoded `"FERAL"` and add a prominent comment in `sens.py` naming `print.rs`/`print_summary` as the source of truth these blocks must track.

## 2. Stream the log instead of buffering it to the end

`_teed_solve` redirects fd 1 to a tempfile and replays it only *after* `solve()` returns, so in Jupyter the user sees nothing during the solve and then a wall of text — which defeats the point of watching an iteration table on a long solve. This is fixable entirely on the Python side; two source facts confirm it:

- **The GIL is released during the solve** — both `pounce-py/solver.rs:107` and `problem.rs:332` wrap it in `py.allow_threads(...)`, so a concurrent Python thread can run.
- **The engine's stdout is line-buffered** — iteration rows are emitted via `print!` in `pounce-algorithm/application.rs:1181/1215`, i.e. Rust's `LineWriter`, which flushes on each `\n` even when fd 1 is redirected to a pipe/file. So each row is available at fd 1 the moment it's produced.

**Recommendation:** run `solver.solve(x0)` in a worker thread and *tail the tempfile* from the main thread (seek/read new bytes → `sys.stdout.write` + flush → ~50 ms sleep → repeat until the worker finishes, then a final drain). Prefer the tempfile-tail over an `os.pipe()` reader: a regular file never applies write backpressure, so the engine can never block on a slow/dead reader, whereas a pipe's ~64 KB kernel buffer means a stalled reader thread would **hang `solve()` forever** — a far worse failure than losing tee output. Two details either way: decode with `codecs.getincrementaldecoder("utf-8")(errors="replace")` (fixed-size reads split multibyte chars — the ANSI wordmark will corrupt otherwise), and note in the docstring that ipykernel's `OutStream` coalesces on its own ~30–200 ms timer, so updates arrive in bursts rather than frame-perfect. Bonus: streaming closes the "output lost if `solve()` raises" gap below for free.

---

## Correctness

- **`_teed_solve` discards captured output if `solve()` raises.** The replay sits *after* the outer `try/finally`, so an exception skips it and the Jupyter user loses the log up to the failure. Move the replay into the `finally`, or let the streaming rework (§2) resolve it.
- **`Feasible_Point_Found` asymmetry (correct, but document it).** The `on_converged` callback fires only for `SolveSucceeded | SolvedToAcceptableLevel` (`application.rs:1694`), so a `Feasible_Point_Found` solve has `converged == False` → session dropped, yet `_STATUS_RESULT` reports `termination_condition = feasible`. That's actually right (no KKT factor is retained, so sensitivity can't be computed), but a caller seeing `feasible` may not expect the session to be gone — worth one sentence in the docstring.
- **Verified non-issue:** `Solved_To_Acceptable_Level` returns `converged == True` (same callback guard), so the session *is* built, consistent with its `optimal/warning` mapping. No bug there.

## Results-object shape (silent changes beyond the PR summary)

- **`results.solver.name` dropped** (was `"pounce (in-process sensitivity session)"`) — confirm ASL parity or restore.
- **`results.problem.upper_bound` / `lower_bound` dropped** (were `obj_val`) — any caller reading the objective off the results object loses it.
- **`results.solver.time` includes replay/decode time**, since `solve_secs` spans all of `_teed_solve`. Measure strictly around the solve so the reported "Total seconds in POUNCE" is engine time.

## Tests

- **The non-live (`live == False`) branch is untested** — the most important gap. Every tee test uses `capfd`, which captures fd 1 directly, i.e. the *live* path (`sys.stdout.fileno() == 1`). The capture-and-replay branch that exists specifically to fix Jupyter is never exercised. Add a test wrapping the solve in `contextlib.redirect_stdout(io.StringIO())` and asserting the banner/summary land in the buffer.
- Otherwise coverage is strong and the assertions are meaningful.

## Minor / withdrawn

- Failure branch loads the iterate with `float(val)` + `skip_validation=True` while the success path uses raw `val` — trivial inconsistency.
- **Not bugs (noted for completeness):** the eq/ineq eval-count reuse (`n_con`/`n_jac` printed for both), the hardcoded `Variable bound violation: 0.0`, and scaled==unscaled objective all faithfully mirror `print_summary` (`print.rs:476,498–511`). They're artifacts of the transliteration, not defects — which is itself an argument for §1.
- `.gitignore` and executed-notebook refreshes are fine.

---

## Verdict

Approve in principle — solid, well-tested, real UX improvement. Before merge I'd want, in priority order: **(1)** at minimum drop the hardcoded `"FERAL"` and pin the duplicated blocks to `print.rs` with a comment (ideally do the shared-crate/`print_level` refactor so the blocks come from Rust); **(2)** stream rather than buffer-to-end; **(3)** add the redirected-stdout test; **(4)** confirm/restore the `solver.name` and `problem` bounds and fix the timing to exclude replay.

---

devin-griff:
Reworked this in the direction you called the **cleanest variant**: rather than reproducing the CLI's blocks in Python, the solver core now emits them itself, gated on `print_level`. That turns the PR into two commits.

**`b49e2f4` — the solver core emits its own console log**
- Moved the Ipopt-style console printers from `pounce-cli` into the shared `pounce-solve-report::console` crate (one source of truth).
- `pounce-algorithm::solve()` emits the problem-statistics block (before the iteration table) and the end-of-run summary (after), gated on `print_level`. The CLI, the Python bindings, and the C interface all get the identical log at `print_level >= 1`; `print_level 0` is silent. JSON mode already forces `print_level 0`, so structured output stays clean.
- Added `IpoptNlp::eval_counts()` (default zeros; `OrigIpoptNlp` reports its live counters) and drained it into `SolveStatistics`, finally populating those `num_*_evals` fields.
- Exposed `pounce.print_banner()` (extension + package) for in-process frontends.

**`7dcb502` — pyomo-pounce streams the engine's log**
- Deleted the ~300-line transliteration; `sens_solve` now sets `print_level`, calls `print_banner()`, and tails fd 1.

### Point by point

**§1 — source from Rust, not re-implement the CLI.** Done via the engine emission; `sens.py` reproduces nothing. Both casualties are fixed structurally: the banner reports the real backend tag (derived in the engine, no hardcoded `FERAL`), and the statistics come from the engine's own reduced problem rather than a Python recompute. `_STATUS_MESSAGES` is gone — the engine's `status_message()` is the source. Your three "not bugs" (eq/ineq eval-count reuse, hardcoded `Variable bound violation: 0.0`, scaled==unscaled) are now the engine's real `print_summary`, i.e. correct rather than transliteration artifacts.

**§2 — stream, don't buffer.** The solve runs on the **main** thread (`pounce.Solver` is a pyo3 `unsendable`, so it can't cross threads); it releases the GIL, so a **worker** thread tails the redirect tempfile to `sys.stdout` live. Tempfile, not `os.pipe`, per your note (a stalled reader can't block the solver). Incremental UTF-8 decoder; the docstring notes ipykernel's ~30–200 ms coalescing.

**Correctness.**
- Output-lost-on-raise: the tail is drained in a `finally`, so a raising solve streams its partial log before re-raising (there's a test).
- `Feasible_Point_Found` asymmetry: documented — `on_converged` doesn't fire, so it reports `feasible` yet has `converged=False`, lands in the no-session path, and its session is dropped. `Solved_To_Acceptable_Level` is unchanged (converges, session built), as you noted.

**§4 — results object.** Restored `solver.name` and the objective bounds; `solver.time` now measures the solve alone (excludes stream/decode).

**§3 + tests.** Added a `contextlib.redirect_stdout(StringIO)` test that exercises the non-live capture/stream branch every `capfd` test skipped. Also fixed the `float(val)` inconsistency between the failed- and success-path iterate loads.

### Two deliberate behavior changes worth flagging

- The summary's obj/grad/constr/jac **eval counts drop by one** — they now report the solver's true count instead of the old CLI wrapper's, which counted one frontend-side evaluation. CLI stdout is otherwise byte-identical on the NLP fixtures.
- `print_level 0` now also suppresses the stats/summary blocks (previously it silenced only the per-iteration table) — needed so `tee=False` is fully quiet, and closer to Ipopt.

### Verification
Crate suite passes except the pre-existing `hs071_output_file_captures_timing_report` (fails identically on a clean base — a `/tmp` output path that doesn't exist on Windows). 19/19 pyomo-pounce tests pass against a locally rebuilt extension, and notebooks 25/26 run clean.

For release: `pounce-solver` needs the minor bump to `0.9.0` (pyomo-pounce's floor is already `>=0.9.0`).

jkitchin (commented):
## Re-review of the revised PR (head `7dcb5023`)

Big step up — the revision implements all four points from the previous review at the source rather than papering over them. One likely-blocking release-coordination bug remains, which is almost certainly why CI is `unstable`.

### What got fixed (verified in the diff)

- **Wrapper blocks sourced from Rust, properly.** Printers moved from `pounce-cli` into the shared `pounce-solve-report::console` crate; `pounce-algorithm`'s `solve()` now emits the problem-stats block (before the iteration table) and the summary (after), gated on `print_level` (`application.rs`: `console_output = v >= 1`; default/unset shows, only `≤0` suppresses — matching the old iteration-table gate). Eval tallies come from the solver's own counters via a new `IpoptNlp::eval_counts()` (default zeros; `OrigIpoptNlp` reports live counts) drained into `SolveStatistics`. The ~300 lines of Python reproduction are gone. Exactly the right shape.
- **Streaming, deadlock-safe.** `_stream_solve` tails a **regular tempfile** (explicitly not an `os.pipe`, with the reasoning inline) from a worker thread while the GIL-released solve runs on the main thread; incremental UTF-8 decoder; `final=True` drain; partial log flushed in the `finally` before a raising solve re-raises. Solve time measured strictly around `solver.solve`.
- **Results object restored.** `solver.name` and objective bounds are back; timing excludes stream/decode; both iterate-load sites use `float(val)`.
- **Test gap closed.** New `test_tee_streams_under_redirect_stdout` drives the non-live branch via `redirect_stdout(StringIO())` — the exact path that was untested.
- **`Feasible_Point_Found` asymmetry** is now documented in a code comment.

### Blocking: version wasn't bumped, but the floor was

`pyomo-pounce/pyproject.toml` now requires `pounce-solver>=0.9.0`, but the PR doesn't bump the actual version:

- root `Cargo.toml` `[workspace.package]` → still `0.8.0`
- `python/pyproject.toml` (the `pounce-solver` wheel) → still `0.8.0`
- the diff touches neither file.

So the wheel this branch builds is `0.8.0`, which does **not** satisfy `>=0.9.0` — `pip install pyomo-pounce` can't resolve against the project's own artifacts, and the pyomo tests (which now hard-depend on the new engine behavior: banner via `print_banner`, stats/summary from the core) can only pass against a `≥0.9.0` wheel that doesn't exist yet. Per `CLAUDE.md`, `scripts/check-release-consistency.sh` also fails unless the three surfaces agree on X.Y.Z. This is almost certainly the cause of the `unstable` state.

**Fix:** bump root `Cargo.toml` and `python/pyproject.toml` to `0.9.0` and run `scripts/check-release-consistency.sh`. The public surface genuinely changed (new `print_banner` pyfunction, new default core stdout, new trait method), so a minor bump is correct — the `0.9.0` floor shows this was intended and just missed on the version fields.

### Should verify: broadened stdout for non-CLI embedders

The core now prints problem-stats + summary at *default* `print_level`, so every frontend that calls `solve()` without setting `print_level=0` — the C interface, the native GAMS link, any library consumer — now emits blocks it never did before. CLI stdout is byte-identical (except the eval counts, now the solver's true count, one lower), per the commit message. Worth confirming no Rust golden-stdout test or downstream parser regresses; this is the other plausible contributor to CI instability. (The Python paths are safe: `nlp_batch` and non-tee `sens_solve` both force `print_level=0`.)

### Minor / cleanup

- **Dead code:** `_NlBridge` still increments `n_obj/n_grad/n_con/n_jac/n_hess`, but nothing reads them anymore — the Python summary that consumed them was deleted and the engine now reports its own counts via `eval_counts()`. Remove the counting and the docstring line about it.
- **`print_banner` still hardcodes `"FERAL"`** (now at `pounce-py/src/lib.rs`, documented as the wheel default). Better than before — single site, in Rust — but still a literal that misreports if a wheel is ever built on a different linsol backend. Low priority given FERAL is the wheel default.
- **Tiny leak edge in `_stream_solve`:** `os.dup(1)`, `mkstemp`, and `open(path)` run before the `try:`; if `open` raised, `saved`/`fd_w`/the temp file leak. Extremely unlikely; optional hardening.
- **PR body is stale** — it still says "reproduced in Python… no Rust, CLI, or wheel change," which is now the opposite of what the PR does.

### Verdict

Much improved — the substantive design is right and the correctness details hold up. The one thing between this and green is the **version bump** (blocking, and the likely CI failure); confirm the embedder-stdout broadening is clean; the cleanup items are optional. Approve once the version is bumped and CI is green.

---

devin-griff:
## Revision addressing the re-review

Thanks for the re-review. All items are handled; accounting below in your priority order. Force-pushed and the PR body is rewritten to match.

### 1. Version bump (blocking) — done

Bumped to `0.9.0` across every surface, not just the floor:

- root `Cargo.toml` `[workspace.package]` -> `0.9.0`, plus all 17 internal `[workspace.dependencies]` path pins
- `python/pyproject.toml` (`pounce-solver`) -> `0.9.0`
- `pyomo-pounce/pyproject.toml` -> `0.9.0` (the `pounce-solver>=0.9.0` floor now resolves against the wheel this branch builds)
- `Cargo.lock` regenerated

`scripts/check-release-consistency.sh` reports the three surfaces agree at `0.9.0`, and there are no stray `0.8.0` pins left in any toml. This was the likely cause of the `unstable` CI, together with item 2.

### 2. Broadened embedder stdout (verify) — clean, after fixing one real regression

You were right to flag this as the other plausible CI contributor. Running the full workspace suite surfaced exactly one golden-stdout regression, now fixed:

The core emitted the stats and summary from the IPM path (`optimize_tnlp`) only. The active-set QP route (`solver_selection=qp-active-set`) dispatches through `optimize_sqp_tnlp`, which early-returns at `application.rs:505` before that emission, so it lost the `Optimal Solution Found` summary the CLI's now-removed `print_summary` used to supply. `forced_qp_active_set_solves_convex_qp` (`qp_dispatch_end_to_end.rs`) caught it.

Fix: the emission is extracted into two shared methods on `IpoptApplication`, `emit_problem_stats` and `emit_end_summary`, gated on the same `print_level` (`console_output = v >= 1`; default/unset shows, `<=0` suppresses). Both `optimize_tnlp` (IPM) and `optimize_sqp_tnlp` (SQP) call them, so every algorithm and frontend gets the identical block with no duplicated presentation code. `emit_end_summary` drains the NLP's counters via the new `IpoptNlp::eval_counts()` into `SolveStatistics` in one place, shared by both paths.

Local results: `pounce-cli` full suite green, `qp_dispatch_end_to_end` 18/18, `pyomo-pounce` 69/69. (One pre-existing failure, `hs071_output_file_captures_timing_report`, is Windows-only and unrelated: it hardcodes a `/tmp` path. It passes on Linux CI.)

### 3. Cleanup

- **Dead `_NlBridge` counting removed.** The `n_obj/n_grad/n_con/n_jac/n_hess` fields, their init, and the increments are gone, along with the docstring line. The engine reports its own counts via `eval_counts()` now, so nothing read them.
- **`print_banner` FERAL tag derived, not hardcoded.** It now reads `linear_solver` and its explicit-set flag exactly as the CLI's `backend_tag` does (`main.rs`), so the two frontends cannot drift. Since the registered default `ma57` is never explicitly set on a fresh application, the default wheel correctly banners `FERAL` rather than the registered-default `ma57`.
- **`_stream_solve` leak edge closed.** The `mkstemp` fd and its reader are acquired inside the `try`, and the `finally` `None`-guards each handle. `os.dup(1)` of the saved fd is taken immediately before the `try` with no failing statement in between, so the scenario you described (a later acquisition raising and leaking the temp file or fds) cannot happen.
- **PR body rewritten** to match what the PR now does (core emits its own log, streamed under `tee`), replacing the stale "reproduced in Python, no Rust/CLI/wheel change" description.

Version bump plus the SQP-route fix should take CI green.

jkitchin:
The only remaining red check here was `check-release-consistency.sh`: moving the console printers into `pounce-solve-report` made `pounce-algorithm` depend on it, but `scripts/publish-crates.sh` still listed `pounce-algorithm` ahead of `pounce-solve-report`, breaking the topological publish order. Since this PR is from a fork, I couldn't push the one-line reorder here directly.

I've opened #208, which carries both of your commits verbatim (unchanged, your authorship) plus that one-line fix on top. `check-release-consistency.sh` passes there. Once #208 goes green and merges, I'll close this as superseded. Thanks @devin-griff!

---

PR #207, opened 2026-07-13
Let user select algorithm from `pounce-rs`

GermanHeim:
Hello @jkitchin,

The current state of pounce lets users manually select the algorithm being used by passing CLI options. `pounce-rs` users don't have an option to set this.

In this PR, I add support for it by:
1. Moving the registration into the shared library so any pounce-rs/library consumer can set and read them
2. I deduplicate the parameters from the CLI library
3. Made it so that the library honors `solver_selection=qp-active-set`
4. Builder defaults `sqp_hessian=lbfgs` (overridable), since the builder supplies no analytic Hessian and the SQP engine's exact default would otherwise fail on the new route

I have also added some tests.
Let me know if you find this useful or if you would like some changes.


jkitchin (commented):
Thanks @GermanHeim — this is a useful contribution, and it checks out mechanically. I pulled the branch and ran the new and surrounding tests locally: `pounce-algorithm --test solver_selection_library_option` (3/3), `pounce-rs` (5 unit + 3 doc tests), and the CLI's `dispatch_routing` + `qp_dispatch_end_to_end` suites (23/23) all pass, so the registration move doesn't regress the CLI path. The builder's `sqp_hessian=lbfgs` default is correctly placed before the user-option loop with `allow_clobber=true`, so `.option_str("sqp_hessian", ...)` still overrides it — same pattern as the existing `hessian_approximation` default.

Three requests before merge:

**1. Error on forcing values the library can't honor (main ask).** Moving the registration into the library means every `IpoptApplication` consumer now *accepts* all six `solver_selection` values, but only `qp-active-set` (plus the trivial `auto`/`nlp`) is actually implemented there — `lp-ipm`, `qp-ipm`, and `socp` routing lives in the CLI's `.nl` dispatch, which needs the parsed `NlProblem`. Before this PR a library caller setting `solver_selection=lp-ipm` got a hard "Unknown option" error; after it, they silently get the NLP filter-IPM even though the registered doc promises "errors if the problem is not an LP". Suggest `optimize_tnlp` return an error for those three values with a message like: *"solver_selection=lp-ipm routing is only available through the pounce CLI (.nl input); library consumers can use qp-active-set, nlp, or auto."* That preserves the pre-PR "this doesn't work here" signal and keeps the option's documented contract honest. (`qp_presolve` is just a tuning knob, so a doc note is enough there — no error.)

**2. Document per-surface behavior in the registered long-descriptions.** State which values are honored by the library (`auto`/`nlp`/`qp-active-set`) vs. the CLI (all six + `qp_presolve`), that the library's `qp-active-set` path does not class-validate (it runs the active-set SQP engine on any TNLP, unlike the CLI which rejects non-LP/convex-QP inputs), and that `qp-active-set` takes precedence over an explicit `algorithm=interior-point` (the `||` in `is_sqp_algorithm_selected`).

**3. `cargo fmt`.** Removing the CLI registrations left a double blank line after `let mut app = IpoptApplication::new();` in `crates/pounce-cli/src/main.rs`, and `cargo fmt --check` fails on it.

Nice touches worth keeping as-is: the HS35 test exercises the exact-Hessian SQP default on the library path while the builder test exercises the `lbfgs` default (complementary coverage), and asserting on `last_sqp_working_set()` is a good observable proxy for "the SQP engine actually ran".

---

GermanHeim:
Hi @jkitchin, all requests added.

jkitchin (approved):
The follow-up commits resolve all three review requests — verified locally on the new head:

- `optimize_tnlp` now rejects an explicitly-set `lp-ipm` / `qp-ipm` / `socp` with `Invalid_Option` and an instructive message, correctly gated on the "explicitly set" flag so defaults are untouched; covered by new tests on both the `pounce-algorithm` and `pounce-rs` surfaces.
- Both option long-descriptions now document the per-surface behavior, the library path's lack of class validation on `qp-active-set`, and its precedence over `algorithm=interior-point`.
- `cargo fmt --check` is clean.

All relevant suites pass here: `solver_selection_library_option` (4/4), `pounce-rs` (6 unit + 3 doc), and the CLI `dispatch_routing` / `qp_dispatch_end_to_end` / `minima_sol_duals` tests.

One optional cosmetic note, fine to leave for a follow-up: the CLI's `--minima` mode calls `optimize_tnlp` before the dispatch block, so `--minima` plus a forced convex selection now surfaces the new error — an improvement over the previous silent ignore, but the "library consumers can use..." wording reads oddly from the CLI. A surface-neutral phrasing ("this solve path supports qp-active-set, nlp, or auto") would cover both.

Thanks for the quick, thorough turnaround!

---

PR #211, opened 2026-07-18
Thread-scoped iteration capture + richer Nlp::solve() result

GermanHeim:
A library that uses `pounce-rs` as a solver backend cannot claim the global tracing subscriber, so capturing a solve's iteration trajectory requires direct `tracing` and `tracing-subscriber` dependencies and additional code.
IterCaptureGuard::start() without that installs silently and captures nothing. Also, `builder::Solution` exposes only status/success/x/objective/multipliers, discarding statistics the driver already computes (wall time, iteration count, evaluation counts, final infeasibilities) and the g/z_L/z_U values available at finalize_solution.

In this PR, I added scoped capture helpers to `pounce-observability` and improved results and exposure in `pounce-rs`.

Some of the code and the changelog were generated using GitHub Copilot. See commit trailer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,8 +45,11 @@ changes.
   the per-iteration trajectory (`stats.iterations`, one `IterRecord` per
   Newton iteration) by installing the thread-scoped collector around the
   solve — note the collector shadows a global subscriber's console output on
-  that thread for the solve's duration. The field additions are breaking
-  only for code that exhaustively destructures `Solution` (pre-1.0).
+  that thread for the solve's duration, and that only the interior-point
+  engine emits the per-iteration event (an active-set SQP solve leaves
+  `stats.iterations` empty while `iteration_count` still counts). The field
+  additions are breaking only for code that exhaustively destructures
+  `Solution` (pre-1.0).
 
 ### Changed — pyomo-pounce streams the engine's log under `tee=True`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,45 @@ changes.
 
 ## [Unreleased]
 
+### Added — thread-scoped iteration capture from Rust (pounce-rs)
+
+- **A Rust consumer embedding POUNCE can now record a solve's iteration
+  trajectory with no direct `tracing`/`tracing-subscriber` dependencies and
+  without touching the global subscriber.** `pounce-observability` gains
+  thread-scoped helpers that bundle the collector-layer install:
+  `with_iter_capture(|| nlp.solve())` runs a closure with capture active and
+  returns its result alongside the recorded `IterRecord`s;
+  `ScopedIterCapture::start()`/`.finish()` is the guard-shaped equivalent for
+  solves that don't fit in a closure; and `collector_scope()` installs just
+  the collector for the `IpoptApplication` path
+  (`enable_iter_history()` + `statistics().iterations`), where the driver
+  manages its own capture guard. All three install the collector as the
+  *thread-default* subscriber (never global) and uninstall on drop; while a
+  scope is active it shadows any global subscriber on that thread, so scope
+  it tightly around the solve. Restoration sub-solve exclusion and nested /
+  sequential capture semantics are unchanged. `pounce-rs` re-exports the
+  helpers plus `IterRecord`, `SolveStatistics`, `IterCaptureGuard`, and
+  `init_subscriber` (and the whole `pounce_observability` crate), so the
+  facade alone covers both iteration capture and console logging;
+  `with_iter_capture`, `collector_scope`, `IterRecord`, and
+  `SolveStatistics` join the prelude.
+
+### Changed — richer `Nlp::solve()` result (pounce-rs)
+
+- **`builder::Solution` now carries the full per-solve picture.** New fields:
+  `g` (constraint values at the solution), `z_l`/`z_u` (bound multipliers),
+  and `stats: SolveStatistics` — wall time
+  (`stats.total_wallclock_time_secs`), `iteration_count`, evaluation counts,
+  final scaled and unscaled infeasibilities, final barrier `mu`, and
+  restoration counters, all filled on every solve with no new bookkeeping
+  (the driver already computed them; the builder now reads
+  `app.statistics()`). A new `.capture_iterations()` builder flag opts into
+  the per-iteration trajectory (`stats.iterations`, one `IterRecord` per
+  Newton iteration) by installing the thread-scoped collector around the
+  solve — note the collector shadows a global subscriber's console output on
+  that thread for the solve's duration. The field additions are breaking
+  only for code that exhaustively destructures `Solution` (pre-1.0).
+
 ### Changed — pyomo-pounce streams the engine's log under `tee=True`
 
 - **`SolverFactory('pounce').solve(m, tee=True)` now streams the engine's own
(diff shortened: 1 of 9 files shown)


jkitchin:
Nice work — the scoped-capture design is sound, and I like that `pounce-observability` gains this purely additively (201 insertions, 0 deletions: `IterCaptureGuard`, `IterCollectorLayer`, `iteration_event_wanted`, and `init_subscriber` are all untouched). I confirmed the blast radius is contained: nothing in the workspace depends on `pounce-rs`, and `pounce-algorithm` has zero changed lines, so the solver loop and the Python/CLI/GAMS paths are unaffected. Branch builds clean and `cargo test -p pounce-observability -p pounce-rs` passes. Publish ordering is fine too — `pounce-observability` already precedes `pounce-rs` in `scripts/publish-crates.sh`, so the new dependency doesn't trip the release guard.

Two real bugs, both of which surface as a *silently empty* trajectory rather than an error.

### 1. `.capture_iterations()` silently no-ops on the active-set SQP engine

`grep -rn ITER_TARGET crates/pounce-algorithm/src/` finds exactly one emitter — `ipopt_alg.rs:538`. No SQP path emits the event. But the builder explicitly supports that engine: it defaults `sqp_hessian` for it (`builder.rs:243`) and tests it (`builder.rs:491`). So

```rust
Nlp::new(p)
    .option_str("solver_selection", "qp-active-set")
    .capture_iterations()
    .solve()
```

returns `stats.iteration_count > 0` alongside `stats.iterations == []`. Since the doc on `capture_iterations` promises "one `IterRecord` per Newton iteration … (empty without this call)", a caller reads that empty vector as "no iterations ran". Either document the IPM-only limitation on `capture_iterations`, or make the unsupported case distinguishable from a genuinely empty one.

### 2. Nesting `with_iter_capture` around a `.capture_iterations()` solve steals the records

```rust
let (sol, iters) = with_iter_capture(|| Nlp::new(p).capture_iterations().solve());
// iters is empty; sol.stats.iterations is full
```

The outer guard sets `CAPTURE = Some(vec![])`; the driver's own guard (`application.rs:1748`) then replaces it with a fresh `Vec`, saving the outer one as `prev`; every event lands in the inner `Vec`; `finish()` restores the still-empty outer buffer.

`ScopedIterCapture` documents this hazard ("Do not combine with the driver's own iteration-history capture"), but `with_iter_capture` and `Nlp::capture_iterations` don't — and those are the two APIs a user actually reaches for. Worth noting that `tests/iteration_capture.rs:23` carefully omits `.capture_iterations()` inside the closure, so the failing combination is never exercised. Minimum fix is a doc note on both; better would be having the inner guard append into an enclosing buffer instead of shadowing it.

### 3. Worth a deliberate decision: `collector_scope` discards the host's tracing output

`collector_scope()` installs a registry containing *only* the collector as the thread-default, so for the solve's duration every tracing event on that thread is dropped — the host's own application logs, POUNCE's warnings, and the `POUNCE_LOG_FORMAT=json` iteration sink. It's documented in three places, so I assume this is known, but "capture and console logging are mutually exclusive" is a sharp constraint for precisely the library-embedder case the PR exists to serve, and `init_subscriber` already shows the composing alternative (collector layered *alongside* the output layers). Detecting an already-installed global collector and skipping the shadowing install would make the two compose. Fine to ship as-is if it's a deliberate call.

### Smaller items

- **`g` / `z_l` / `z_u` doc lengths** (`builder.rs:93-97`): documented as length `n_constraints` / `n` unconditionally, but all three are only assigned in `finalize_solution`, which the driver skips on its early-return arms (the success path is `application.rs:1851`). A caller who follows the doc and writes `sol.z_l[i]` panics. Suggest "empty unless `finalize_solution` ran". (`x` has the same gap today, but the new fields state a hard guarantee the code doesn't provide.)
- **`#[non_exhaustive]` on `Solution`** (`builder.rs:81`): it's only ever constructed inside `solve()`, so this is free, and it makes this the last field addition that breaks downstream exhaustive destructuring. `SolveStatistics` will keep growing.
- **Duplicated condition** (`builder.rs:261-263`): `self.capture_iterations` is branched on twice in consecutive statements. Folding them keeps the two halves of one decision together, so a future edit can't update the `.then()` without the `enable_iter_history()`:
  ```rust
  let scope = self.capture_iterations.then(|| {
      app.enable_iter_history();
      crate::collector_scope()
  });
  ```
- **Thread affinity**: `CollectorScope` / `ScopedIterCapture` are `!Send` (they hold a `DefaultGuard`) and the capture slot is thread-local. The `!Send` bound blocks the obvious misuse, but `with_iter_capture(|| block_on(…))` or holding the guard across an `.await` that migrates threads returns an empty `Vec` with no error. A one-line "the closure must run to completion on the calling thread" would close it.

jkitchin:
Let me know when this is ready for another review. Thanks!

GermanHeim:
This should be ready for review!

After this, I would like to implement a way to get the raw logs back since I am recreating it in oximo by doing the following:
https://github.com/oximo-rs/oximo/pull/40/changes#diff-d478295dc518f89ced7ce4c035a16582dfefafb9fd70838fcf73deb0705e6a04R118

jkitchin:
On getting the raw logs back — happy to support this properly so you can stop reconstructing it in oximo. One thing that shapes the design: POUNCE emits **two separate log streams, and only one goes through `tracing`**:

1. **The iteration table** (`iter  objective  inf_pr  inf_du  lg(mu) …`) is written straight to `anstream::stdout()` at `ipopt_alg.rs:494-516` — it never touches `tracing`. The `pounce::iteration` event right below it is a structured duplicate, which is what `with_iter_capture` already hands you as `IterRecord`s. So this is the stream you're re-rendering in oximo.
2. **Everything else** — banners, warnings, the `EXIT:` summary — goes through `tracing` to stderr via the `TigerFormat`/JSON fmt layer.

Because of that split, I don't think one hook can cover both:

**For the iteration table (the part you're hand-rolling):** `IterRecord`'s fields are exactly the table's columns, and `OrigIterationOutput::format_row` / `::HEADER` already produce the canonical line — they're just unreachable from an `IterRecord` today because `format_row` takes the internal `IpoptData`/`IpoptCq` handles. I'd refactor it to funnel through a `format_row_from_record(&IterRecord) -> String`, then expose something like `pounce_observability::format_iteration_table(&[IterRecord], ansi: bool) -> String` (header + one styled row each, reusing `pounce_common::style::iteration_row_style`). Then oximo gets byte-identical console output from `sol.stats.iterations` with no duplicated formatting — and it's testable in-crate against the live stdout path, so it can't drift.

**For the stderr side (warnings/summary), if you want it too:** a scoped `with_log_capture(|| solve()) -> (R, String)` in `pounce-observability`, mirroring the `collector_scope`/`with_iter_capture` pattern you just built — same `TigerFormat` + `console_filter`, but `.with_writer()` into a shared buffer and `.with_ansi(false)`, installed as the thread-default. It has to live inside the crate since `TigerFormat`/`console_filter` are private (which is exactly why you can't do it from oximo now). One wrinkle: unlike `collector_scope`, this one shouldn't skip when a global subscriber is installed — the point is to redirect into the buffer, so it should always shadow for the scope's duration.

I'd steer away from threading an injectable writer through the algorithm to redirect stdout — invasive, and stdout isn't thread-scoped so it'd break under concurrent solves. The formatter-reuse route gets the same result without touching the hot loop.

Let's do the `format_iteration_table` helper as its own follow-up after this lands — happy to sketch the `format_row_from_record` refactor if it helps.

---

Issue #222, opened 2026-07-19
[Adversary] SOUNDNESS: direct conic driver returns Optimal with a NaN solution on PSD programs


jkitchin:
> Found by the randomized SOC/PSD differential suite added in #221; reduced to a minimal standalone case by hand afterwards. Reproduces on `main` at f9feb74 — **not** introduced by #221, which touches only the HSDE driver.

## Family / surface

`convex` — `pounce_convex::solve_socp_ipm` with `use_hsde: false` (the **direct** symmetric driver in `ipm.rs`), on a mixed orthant + PSD program.

## Summary

The direct driver returns `QpStatus::Optimal` with `x = [NaN, NaN]` and `obj = NaN`. A caller that checks the status — which is the documented way to know the answer is usable — gets a garbage solution with no indication anything went wrong.

This is a **soundness** problem, not a limitation: an `Optimal` verdict is a claim, and here the claim is accompanied by no solution at all. It is strictly worse than the `nan`-on-failure guard of #117, which at least reports a failing status alongside the `nan`.

The HSDE driver solves the same instance correctly, so this is specific to the direct path.

## Reproduce

Self-contained — 2 variables, 4 orthant rows, one `Psd(3)` block, no equalities:

```rust
use pounce_convex::{ConeSpec, QpOptions, QpProblem, QpStatus, Triplet, solve_socp_ipm};
use pounce_feral::FeralSolverInterface;

let prob = QpProblem {
    n: 2,
    p_lower: vec![],
    c: vec![-0.4587520836280501, -0.28511097640465644],
    a: vec![],
    b: vec![],
    g: vec![
        Triplet::new(0, 0, 1.0),
        Triplet::new(1, 0, -1.0),
        Triplet::new(2, 1, 1.0),
        Triplet::new(3, 1, -1.0),
        Triplet::new(4, 1, 0.9144362258975529),
        Triplet::new(5, 0, 0.5271709449981303),
        Triplet::new(5, 1, -0.6408278339462257),
        Triplet::new(6, 0, -0.9444454430308236),
        Triplet::new(6, 1, -0.9246001786463665),
        Triplet::new(8, 0, -0.7591854473470727),
        Triplet::new(8, 1, -0.533518750473565),
        Triplet::new(9, 0, -0.7290269156600226),
        Triplet::new(9, 1, 0.6420378359547312),
    ],
    h: vec![
        10.0, 10.0, 10.0, 10.0,
[...]

PR #224, opened 2026-07-19
pyomo-pounce: add block_analyze, the analysis-only DM partition

devin-griff:
## What

`pyomo_pounce.block_analyze(model, decisions=[...])`: the analysis half of `block_initialize` on its own. Same decision handling, same Dulmage-Mendelsohn decomposition of the active equality system, but nothing is seeded, projected, or solved, and no values are read or written (so, unlike `block_initialize`, decisions do not need values). It returns a `BlockAnalysisReport` carrying the **full** partition as the component objects themselves, with nothing capped:

- the underconstrained subsystem (variables and constraints), with `n_extra_degrees_of_freedom`
- the overconstrained subsystem (constraints and variables), with `n_extra_specifications`
- the square part and its block-triangular calculation order (`variable_blocks` / `constraint_blocks`)

`str(report)` previews long name lists at 10 entries; the data is never capped.

## Why

`block_initialize` already computes this partition, but a caller can only see it through the initialization report: capped, name-only, and only after the fill and block solves have run under a specification already known to be bad. Downstream tooling (the steady-state initialization in [drto](https://github.com/devin-griff/drto) is the immediate consumer) wants to *diagnose first, decide what to specify, then solve once*. A badly specified distillation column is the motivating shape: the levels land underconstrained and the holdup balances overconstrained, and repairing that specification needs the full partition, not a display list.

## How

The partition step is extracted from `block_initialize`, which now delegates to `block_analyze` (its behavior and report are unchanged; the existing suite passes as-is). Decisions are fixed for the analysis and restored in a `finally`, same as before.

Per CONTRIBUTING: code + tests (6 new, hermetic, no solver binary needed), CHANGELOG entry under Unreleased, and the book's `docs/src/pyomo.md` section extended. `pyomo-pounce/tests` passes locally (75 passed). No solver routing touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,28 @@ changes.
 
 ## [Unreleased]
 
+### Added — `pyomo_pounce.block_analyze`, the analysis half of `block_initialize` on its own (Pyomo, #224)
+
+- **The Dulmage-Mendelsohn partition is now available without solving
+  anything.** `block_initialize` already computes the full partition of the
+  equality system, but a caller could only see it through the initialization
+  report, where the underconstrained/overconstrained lists are capped,
+  name-only, and bundled behind the fill and block solves. `block_analyze(model,
+  decisions=[...])` runs the same decision handling and the same decomposition
+  and returns a `BlockAnalysisReport` carrying the **full** partition as the
+  component objects themselves: the underconstrained and overconstrained
+  subsystems (variables and constraints on both sides), the square part, and
+  its block-triangular calculation order, with nothing capped. Convenience
+  counts (`n_extra_degrees_of_freedom`, `n_extra_specifications`) say how far
+  from square the system is.
+- Analysis is purely structural, so decisions do not need values here (they do
+  in `block_initialize`, whose solve must hold them at a concrete point), no
+  values are read or written, and fixed flags are restored on the way out.
+  This makes it a safe first pass on a badly specified model: diagnose, decide
+  what to specify, then call `initialize`/`block_initialize` to do the work.
+- `block_initialize` now delegates its partition step to `block_analyze`; its
+  behavior and report are unchanged.
+
 ### Fixed — a diverged conic solve could report `Optimal` with a `NaN` solution (#222)
 
 - **`solve_socp_ipm` could return `QpStatus::Optimal` alongside `x = [NaN, NaN]`.**
--- a/docs/src/pyomo.md
+++ b/docs/src/pyomo.md
@@ -98,3 +98,25 @@ variables and constraints are reported **by name** —
 specify or flag as decisions, `overconstrained_constraints` the
 redundant or conflicting specifications. Permanently-known inputs can
 simply be `fix()`ed instead of listed as decisions.
+
+The analysis half is also available on its own:
+
+```python
+rep = pyomo_pounce.block_analyze(               # the DM partition only:
+    model, decisions=[m.feed, m.reflux])        # nothing seeded or solved
+rep.underconstrained_variables                  # VarData objects, uncapped
+rep.n_extra_degrees_of_freedom                  # how many specs are missing
+rep.variable_blocks                             # the calculation order
+```
+
+`block_analyze` runs the same decision handling and the same
+Dulmage-Mendelsohn decomposition, but touches nothing: no values are
+read or written (so, unlike `block_initialize`, the decisions do not
+need values), and no solve happens. Where the initialization reports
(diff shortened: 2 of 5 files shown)


jkitchin:
## Review of PR #224 — `pyomo_pounce.block_analyze`

I read the full diff, verified the `block_initialize` delegation against the base implementation line by line, and ran the test suite locally on the PR head: all 16 tests in `pyomo-pounce/tests/test_block_init.py` pass (1 solver-dependent skip); the 6 new tests pass in ~1s with no solver binary.

### Overview

Extracts the Dulmage-Mendelsohn analysis step out of `block_initialize` into a standalone `block_analyze(model, decisions=[...])` returning a `BlockAnalysisReport` with the **uncapped** partition as live component objects (under/overconstrained subsystems, square part, block-triangular calculation order). `block_initialize` now delegates to it. Touches 5 files: implementation, `__init__.py` exports, 6 new tests, CHANGELOG, and `docs/src/pyomo.md`.

### Correctness — verified equivalent ✅

- All four former `igraph` uses (`constraints` check, `dulmage_mendelsohn`, `block_triangularize`, and the interface construction itself) are covered by the delegation; nothing downstream in `block_initialize` references `igraph` anymore.
- `square` semantics are identical: `analysis.square` is computed from the same `unmatched + underconstrained` / `unmatched + overconstrained` unions the old inline code used.
- The empty-equality-system path is equivalent: `block_analyze` returns early with defaults, and `block_initialize` then exits at `if not square_vars` with the same report contents as the old early return.
- The fix/unfix interplay is right: `block_initialize` fixes decisions first, so the inner `block_analyze(model)` sees them already fixed, fixes nothing, and its `finally` unfixes nothing — each layer restores only its own fixes.
- `_flatten_vars` builds its full list before any `fix()` call, so a bad `decisions` entry can't leave the model partially fixed.
- The value-free decision handling (no `ValueError` for valueless decisions, unlike `block_initialize`) is deliberate, documented, and tested.

### Code quality ✅

Follows the module's existing conventions closely: dataclass report with `__str__`, matching docstring style, same networkx probe-import pattern with a matching error message, CHANGELOG + docs per CONTRIBUTING. The `_preview` helper keeps display capping out of the data model, which is the whole point of the feature.

### Test coverage ✅

The 6 new tests are well-chosen: touches-nothing invariant, uncapped-vs-capped contrast (15 vars vs the display cap of 10), overconstrained part, valueless decisions with fixed-flag restore, calculation order, and the no-equalities edge case. All hermetic.

### Issues

1. **Merge conflict (blocking)** — the PR is based on `f10809a`, and `main` has since merged #225, whose #222 entry now sits at the top of CHANGELOG's `Unreleased` section — exactly where this PR inserts its entry. Needs a rebase onto `main`; the conflict is CHANGELOG-only and trivial.
2. **Bare `List` annotations (nit)** — `BlockAnalysisReport` uses unparameterized `List` for its component lists, where `BlockInitReport` uses `List[str]`. Since these hold `VarData`/`ConstraintData` (only documented in comments), `List["VarData"]`-style annotations would help IDEs and readers. Non-blocking.
3. **Untested restore-on-exception path (nit)** — the `finally` unfix is exercised only on the success path. A test that decisions are unfixed when `IncidenceGraphInterface` raises (e.g. via a monkeypatch) would pin the guarantee down. Non-blocking.

### Security / performance

No concerns. No new dependencies (reuses the existing optional `incidence_analysis` stack), no solver routing touched, and `block_initialize` still constructs exactly one `IncidenceGraphInterface` per call — no duplicated work from the delegation.

**Verdict:** Well-scoped, behavior-preserving, and thoroughly tested — good to merge once rebased to resolve the CHANGELOG conflict.

---

devin-griff:
Thanks for the review. All three items are addressed on the updated branch:

1. **Rebase (blocking)**: rebased onto current `main`. The CHANGELOG conflict was resolved by placing this PR's entry above the #222 entry in `Unreleased`, with the PR number added to its heading.
2. **Typed annotations (nit)**: `BlockAnalysisReport`'s component lists are now `List[VarData]` / `List[ConstraintData]` (`variable_blocks` as `List[List[VarData]]`), with the pyomo imports under a `TYPE_CHECKING` guard so nothing changes at runtime.
3. **Restore-on-exception (nit)**: added `test_block_analyze_unfixes_decisions_on_exception`, which monkeypatches `IncidenceGraphInterface` to raise during construction and asserts the decision's fixed flag is restored by the `finally`.

All 18 tests in `test_block_init.py` pass locally; CI is running on the new head with the ubuntu test job already green.

PR #228, opened 2026-07-19
pyomo-pounce: automatic specification repair (block_repair_plan)

devin-griff:
Builds on #224. Some specifications are structurally wrong, not just badly started: hold every flow control of a distillation column at steady state and the drum levels are undetermined while the holdup balances turn redundant — square by count, singular in structure, unsolvable from any starting point.

## What this adds

**`block_repair_plan(model, decision_candidates=...)`** — a planning sibling of `block_analyze`, same touch-nothing character. From the variables you would like held it plans a valid specification:

- `decisions` — the subset a square system can hold;
- `pruned` — candidates the equalities claim, solved for instead (tier-preference matching, provably the minimum number pruned);
- `pinned` — variables the equalities provably cannot determine, identified **automatically with no user input**. The rule is structural: an equation `0 == f/g` cannot determine a variable appearing only in the denominator `g` (its sensitivity there vanishes at every solution) — exactly the shape substituting `d/dt = 0` into a dynamic balance produces, which is how loose integrators hide in steady-state models. Dropping those unusable edges makes the identification canonical: on the 569-equation double-column example the four drum levels come out identically under every matching order, where raw matching gives an order-dependent (and numerically singular) choice;
- `loose_variables` / `redundant_constraints` — genuine model defects, uncapped component objects.

**Automatic repair in `initialize` and `block_initialize`** — no new arguments. The specification is checked first: square means the decisions are used exactly as shipped; broken means they become the candidate pool, the plan is applied call-scoped (fixed flags restored, values only), and `report.repair` records it. A pruned decision no longer needs a value; a valueless pinned variable gets a bounds-aware seed. Must-hold needs no API: a variable the user fixes is out of the plan by construction.

**Notebook 28** (`python/notebooks/28_pyomo_initialization_repair.ipynb`) — a self-contained 569-equation two-column distillation train authored at steady state with no starting values, showing both failure modes and their fixes: the singular all-flows-held specification (diagnose, plan, automatic repair, solve to reference purities) and the all-zeros starting point (dies in restoration, rescued by the block-triangular solve). Executed outputs saved.

Plus 13 new hermetic tests (87 pass), a CHANGELOG entry under Unreleased, and a "Repairing a bad specification" section in the book's pyomo page.


pyomo-pounce/pyomo_pounce/block_init.py
```diff
@@ -362,19 +646,29 @@ def block_initialize(
...
+                f"decision variable {vd.name!r} has no value: a decision "
+                "must be held at a concrete value during initialization"
+            )
+        vd.fix()
+        fixed_by_us.append(vd)
+    for vd in plan.pinned:
+        if vd.value is None:
+            _seed_var(vd)
+        vd.fix()
+        fixed_by_us.append(vd)
```
jkitchin: **A pin can seed to exactly 0.0 — the one value guaranteed to blow up.**

A pinned variable, by construction, appears *only in denominators* of `0 == f/g` rows (that's what makes every edge unusable). But `_seed_var` returns `0.0` for an unbounded variable — and also for symmetric bounds like `(-1, 1)` (midpoint) or a lower bound of exactly `-1` (`lo + 1.0`). Every equation the pin touches then evaluates `f/0` during the SCC solve. It fails soft (caught into `report.failures`), but the repair converts a cleanly diagnosable spec problem into an opaque evaluation failure. The tests only avoid this because `drum_model` uses bounds `(1, 3)`.

Suggest a pin-specific seed that avoids zero (e.g. nudge off 0, or fall back to 1.0), or at minimum a warning when a pin seeds to 0. Same applies to the pin loop in `repair.py::initialize`.

---

pyomo-pounce/pyomo_pounce/block_init.py
```diff
@@ -362,19 +646,29 @@ def block_initialize(
...
     report = BlockInitReport()
 
-    # --- hold the decisions -------------------------------------------
+    # --- check the specification, then hold it -------------------------
+    # On a well-specified system the plan is a no-op (every decision
+    # selected) and this is exactly the shipped behavior. A pruned
+    # decision needs no value (it gets solved for); a pinned variable
+    # without one gets a bounds-aware seed (the pin is the repair's
+    # choice, not the user's, so erroring on it would be hostile).
+    plan = block_repair_plan(model, decision_candidates=decisions)
```
jkitchin: **Silent repair with no strict escape hatch (design).**

Previously a non-square specification held the user's decisions at their values and reported `square=False`; now decisions can be pruned and *solved to different values* than the user set, and `block_initialize(m)` with **no** decisions at all will auto-pin-and-fix variables that previously surfaced as underconstrained. Everything is recorded in `report.repair` and the no-new-arguments choice is clearly deliberate — but there's no way to say "my spec should be square; report/error instead of repairing."

Consider a `repair="auto"|"off"` keyword (or explicitly documenting `block_analyze` as the strict path) so users who want spec errors surfaced loudly aren't silently "fixed."

---

pyomo-pounce/pyomo_pounce/block_init.py
```diff
@@ -226,6 +257,247 @@ def _flatten_vars(vars_like):
...
+    ]
+
+    tiers = (
+        [i for i, v in enumerate(gvars)
+         if id(v) not in candidate_ids and var_adj[i]],
+        [i for i, v in enumerate(gvars) if id(v) in candidate_ids],
+    )
+    eq_match, var_match = _tiered_matching(var_adj, tiers)
+
+    pruned_ids = {id(gvars[i]) for i in tiers[1] if i in var_match}
```
jkitchin, line 530: **Which candidate gets pruned is order-dependent w.r.t. an arbitrary internal order.**

When several candidates conflict, the one that gets pruned depends on `IncidenceGraphInterface` variable order, not the order the user listed candidates. The docs stress the pins are canonical but say nothing about pruning ties.

Cheap improvement: augment the candidate tier in **reverse user-list order** — the same lexicographic-matroid argument the `_tiered_matching` docstring already invokes then makes later-listed candidates preferentially *matched* (pruned), i.e. earlier-listed candidates preferentially kept. That gives users an implicit priority knob for free. Otherwise, worth documenting that the tie-break is arbitrary (though deterministic per model).

---

pyomo-pounce/pyomo_pounce/block_init.py
```diff
@@ -226,6 +257,247 @@ def _flatten_vars(vars_like):
...
+    plan.n_variables = len(gvars)
+    if not eqs:
+        # nothing to determine: every candidate is simply an input
+        plan.decisions = candidates
+        return plan
+
+    vindex = {id(v): i for i, v in enumerate(gvars)}
+    raw_degree = [0] * len(gvars)
+    var_adj = [[] for _ in gvars]
+    for e, con in enumerate(eqs):
```
jkitchin, line 502: **Repeated structural analysis (minor, perf).**

`block_repair_plan` builds an `IncidenceGraphInterface` *and* re-walks every constraint body with `get_incident_variables` (duplicating adjacency the igraph already computed); `block_initialize` then builds a second graph inside `block_analyze`, and a full `initialize` call performs the whole analysis a third time (its own plan + the inner `block_initialize`'s no-op plan + `block_analyze`). Irrelevant at 569 equations, noticeable on 10⁵-row flowsheets.

Not blocking — but reusing the igraph's own adjacency here would halve the walks, and a TODO would at least record the intent.

---

pyomo-pounce/pyomo_pounce/block_init.py
```diff
@@ -362,19 +646,29 @@ def block_initialize(
...
-            vd.fix()
-            fixed_by_us.append(vd)
+    for vd in plan.decisions:
+        if vd.value is None:
+            raise ValueError(
+                f"decision variable {vd.name!r} has no value: a decision "
+                "must be held at a concrete value during initialization"
+            )
+        vd.fix()
+        fixed_by_us.append(vd)
```
jkitchin: The fixing loops sit outside the `try/finally`, so a `ValueError` on the second valueless decision leaves the first one fixed on the model. This shape pre-exists the PR, but the change extends it (the pin loop too, and the same pattern in `repair.py::initialize`) — moving the fixing inside the `try` would close the leak in both places.

---

pyomo-pounce/pyomo_pounce/block_init.py
```diff
@@ -226,6 +257,247 @@ def _flatten_vars(vars_like):
...
+    if value(con.lower) != 0 or value(con.upper) != 0:
+        return incident
+    body = con.body
+    while isinstance(body, NegationExpression):
+        body = body.args[0]
+    if not isinstance(body, DivisionExpression):
+        return incident
+    from pyomo.contrib.incidence_analysis import get_incident_variables
+
+    numerator_ids = {id(v) for v in get_incident_variables(body.args[0])}
```
jkitchin, line 364: Two small things here:

- The three `import`s run on every call (once per constraint in the plan loop). Python caches them, but hoisting to the top of `block_repair_plan` (or module scope behind the existing probe) would be cleaner.
- The rule only inspects the top-level division, so a nested case like `0 == (a/b)/c` keeps the `b` edge even though `b` is equally undeterminable (`numerator_ids` includes it). That's the safe direction — false negatives only — but a sentence in the docstring noting the rule is deliberately shallow/conservative would prevent someone "fixing" it into a recursive version without thinking about false positives.

The math itself checks out: `d/dg (f/g) = -f/g²` vanishes at any solution of `f/g = 0`, and the `f/g == 5` scope test guards the boundary well.

---

python/notebooks/README.md
```diff
@@ -19,6 +19,7 @@ jupyter lab python/notebooks/01_getting_started.ipynb
 | 01 | [`01_getting_started.ipynb`](01_getting_started.ipynb) | First solve — define a problem, call `minimize`, read the result. |
 | 02 | [`02_jax_autodiff.ipynb`](02_jax_autodiff.ipynb) | Let JAX supply exact gradients/Hessians. |
 | 05 | [`05_pyomo.ipynb`](05_pyomo.ipynb) | Drive POUNCE from a Pyomo model. |
+| 28 | [`28_pyomo_initialization_repair.ipynb`](28_pyomo_initialization_repair.ipynb) | The two ways an initialization goes bad, on a 569-equation distillation train: a structurally singular specification (diagnosed by `block_analyze`, repaired by `block_repair_plan`, applied automatically by `initialize`) and a garbage starting point (rebuilt by the block-triangular solve, even from all zeros). |
```
jkitchin: Nit: this row lands between 05 and 07, breaking the numeric ordering every other table in this README follows. Move it after 08 (or note the thematic grouping next to the Pyomo notebook if that placement is deliberate).

---

jkitchin (commented):
**Overall: solid, well-tested feature with sound graph theory at its core.** Two design points deserve attention before merge — a zero-seed edge case that can turn pins into division-by-zero, and the silent-repair default with no strict escape hatch (both detailed inline). CI hasn't run on the head commit yet.

**What's good**

- `_tiered_matching` is a clean iterative Kuhn's algorithm; greedy augmentation in tier order does give the lexicographic-optimal transversal, so "provably the minimum number pruned" holds. The path-flip logic and the no-recursion claim both check out.
- The denominator rule is mathematically right and strictly conservative — edges are only *removed* when provably unusable, so false positives are impossible.
- Careful behavioral compatibility: fixed candidates stay inputs, pruned decisions correctly no longer need values, the repair is call-scoped, and on a square spec the plan is a no-op so shipped behavior is unchanged.
- The 13 hermetic tests are targeted, including the "plan, not action" invariant (nothing fixed, nothing moved).

**Smaller points not anchored inline**

- `report.n_decisions_fixed` now counts pins under the label "decisions fixed"; the repair line disambiguates, but a separate `n_pinned` field would be cleaner.
- Missing test cases worth adding: a pin with no/symmetric bounds (the zero-seed case), and an already-fixed variable passed as a `decision_candidate`.

No security surface (pure structural analysis), no release-surface changes, and CHANGELOG/docs/`__all__` all follow the pattern established by #224.

---

pyomo-pounce/pyomo_pounce/block_init.py
devin-griff: Fixed: pins now seed through a dedicated `_seed_pin` that never returns exactly zero (off-center fallback within finite bounds, shifted otherwise), used in both pin loops. Added both test cases you named: an unbounded pin seeds 1.0, symmetric bounds (-1, 1) seed 0.5.

pyomo-pounce/pyomo_pounce/block_init.py
devin-griff: Took the suggestion as-is: `repair="auto"|"off"` on both entrypoints, auto default. Off restores shipped behavior exactly: decisions held as given, values required for all, nothing pruned or pinned, non-square reported via `report.square`. `initialize` passes the mode down so the two layers cannot disagree. Tested on both entrypoints, plus rejection of invalid values.

pyomo-pounce/pyomo_pounce/block_init.py
devin-griff: Implemented the reverse-listing augmentation: earlier-listed candidates are preferentially kept, documented in the docstring, book page, and CHANGELOG, with a test asserting the pruned candidate flips when the list is reordered. Pleasant side effect: with the column's natural listing (reflux, boilup, then draws) the plan now keeps LT/VB and prunes the draws, the textbook LV configuration. Notebook 28 re-executed to match.

pyomo-pounce/pyomo_pounce/block_init.py
devin-griff: Halved the walks: raw incidence now comes from the igraph's own adjacency, and expression bodies are inspected only on `0 == f/g` shaped rows. The remaining duplication (plan, analyze, and pipeline each building a graph) is recorded as a TODO at the construction site.

pyomo-pounce/pyomo_pounce/block_init.py
devin-griff: Moved the fixing loops inside the try in both entrypoints, so a mid-loop ValueError unfixes anything already held. Pinned with a test: first decision valued, second valueless, neither left fixed.

pyomo-pounce/pyomo_pounce/block_init.py
devin-griff: Hoisted the expression-type imports to module scope; the numerator walk's import stays inside its rare branch. `_usable_incident` now documents that the shallow rule is deliberate: false negatives only, with a warning not to make it recursive without considering the false-positive direction.

python/notebooks/README.md
devin-griff: Moved the row after 08.

devin-griff:
All seven inline points and both summary items are addressed in a84734d: `repair="auto"|"off"`, ordered tie-breaks, nonzero pin seeds, fixing inside the try/finally, adjacency reuse plus a TODO for the shared graph, hoisted imports with the shallow-rule note, `n_pinned` as its own field, the two missing tests, and the README row. 96 tests pass; notebook 28 is re-executed under the new tie-break and now lands the textbook LV configuration on the column.

PR #230, opened 2026-07-20
pyomo-pounce: failed initialization solves no longer poison variable values

devin-griff:
A failed solve anywhere in the initialization pipeline now leaves variable values exactly as they were. Failures are reported, never written into the model.

**The bug.** `block_initialize` delegated its block loop to Pyomo's `solve_strongly_connected_components`, which loads whatever the subsystem solver returns: a diverged block with a warning-status result (infeasible, iteration limit) was written into `Var.value` and the loop continued, with the report claiming every variable initialized. `project_to_feasible` had the same defect, so a diverged projection also loaded its iterate, making the pipeline's "continuing with the unrepaired point" warning untrue.

**The evidence.** On a 1525-equation air-separation model, the model's own build point solves the repaired steady-state specification to optimal, while the "initialized" point, containing the diverged iterate of a failed 1518-variable block solve, stalls for 3000 iterations. The poisoning was the entire difference.

**The fix.** The block loop moves into this module so each solve's verdict is checked before its values are kept: 1x1 blocks by Pyomo's single-constraint Newton, larger blocks by the solver, each block snapshotted first. A failed block restores its seeds, the loop stops there (downstream blocks feed on the failed one by construction), and the failure names the block and its size. The projection restores the pre-projection point on any non-optimal termination. Same restore-don't-report-through philosophy as the refused-stop mechanism in #229, one layer up.

Arguably the underlying sharp edge belongs to the Pyomo utility, which offers no per-block status check for a tool whose stated purpose is initialization; owning the loop fixes it here regardless and buys the named-block reporting.

Four new tests (failed coupled block keeps seeds, failed Newton restores its seed, downstream blocks keep seeds after a failure, failed projection restores the point); 100 pass. CHANGELOG under Unreleased and the no-poisoning guarantee stated in the book's pyomo page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,26 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — failed initialization solves no longer poison variable values (Pyomo)
+
+- **A failed solve anywhere in the initialization pipeline now leaves
+  variable values exactly as they were.** Previously `block_initialize`
+  delegated its block loop to Pyomo's `solve_strongly_connected_components`,
+  which loads whatever the subsystem solver returns: a diverged block with a
+  warning-status result (infeasible, iteration limit) was written into
+  `Var.value` and the loop continued, reporting success. On a 1525-equation
+  air-separation model this turned a solvable steady state into a
+  3000-iteration stall — the model's own build point solved fine, but the
+  "initialized" point, containing a diverged 1500-variable block iterate,
+  did not. The block loop is now in this module: each block's verdict is
+  checked before its values are kept, a failed block restores its seed
+  values, the loop stops there (downstream blocks feed on the failed one by
+  construction), and the failure names the block. `project_to_feasible` had
+  the same defect — a diverged projection loaded its iterate, making the
+  pipeline's "continuing with the unrepaired point" warning untrue — and now
+  restores the pre-projection point on any non-optimal termination.
+
+
 ### Added — automatic specification repair: `block_repair_plan`, and repair inside `initialize`/`block_initialize` (Pyomo, #228)
 
 - **A structurally broken specification now initializes anyway, and the report
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ changes.
 
 ## [Unreleased]
 
-### Fixed — failed initialization solves no longer poison variable values (Pyomo)
+### Fixed — failed initialization solves no longer poison variable values (Pyomo, #230)
 
 - **A failed solve anywhere in the initialization pipeline now leaves
   variable values exactly as they were.** Previously `block_initialize`
--- a/docs/src/pyomo.md
+++ b/docs/src/pyomo.md
@@ -84,6 +84,12 @@ one); `project_to_feasible` repairs that by minimizing
 bounds — the full nonlinear projection, solved with POUNCE, with the
 original objective restored afterwards.
 
+Both stages guarantee that a failed solve leaves variable values
+exactly as they were: a diverged projection restores the
+pre-projection point, and a failed block solve restores that block's
+seeds and stops, so initialization can never make your starting point
+worse than it found it.
+
 `block_initialize` is IDAES-flavored initialization without
 hand-written routines. `decisions=` holds the listed variables at
 their current values for the solve and releases them afterwards (each
--- a/pyomo-pounce/pyomo_pounce/block_init.py
+++ b/pyomo-pounce/pyomo_pounce/block_init.py
@@ -6,10 +6,11 @@ active **equality** constraints, extract the square (well-determined)
 part of the variable/constraint incidence graph (Dulmage-Mendelsohn,
 via ``pyomo.contrib.incidence_analysis``), and solve it block by block
 in topological order, writing the solution into ``Var.value``. The
-block-by-block solve itself is delegated to Pyomo's
-``solve_strongly_connected_components`` (1x1 blocks by Newton, larger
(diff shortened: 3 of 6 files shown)


jkitchin (commented):
## Review — failed initialization solves no longer poison variable values

### Overview

`block_initialize` previously delegated its block-by-block solve to Pyomo's `solve_strongly_connected_components`, which loads whatever the subsystem solver returns — so a diverged block solve with a warning status wrote its junk iterate into `Var.value` and kept going. This PR moves the block loop into `block_init.py` (1x1 blocks via `calculate_variable_from_constraint`, larger blocks via the solver), snapshots each block before solving, restores the snapshot on any failure, and stops the loop at the first failed block. `project_to_feasible` gets the same treatment: snapshot before the projection solve, restore on any non-optimal termination. Docs, CHANGELOG, and four new tests round it out.

### Verification performed

I checked the load-bearing assumptions against real Pyomo and ran the PR branch:

- **Topological order**: `igraph.block_triangularize` returns a block *lower triangular* partition, so each block depends only on earlier blocks — iterating `analysis.variable_blocks`/`constraint_blocks` in order is a faithful replacement for Pyomo's SCC loop.
- **Nested `TemporarySubsystemManager`s**: the manager records `(var, var.fixed)` on entry and restores it on exit, so the per-block manager re-fixing variables already fixed by the outer one is safe.
- **Live probe of mid-sequence failure** (block 1 solvable, block 2 not): the earlier block keeps its solved value, the failed block restores its seed, `n_vars_initialized == 1`, and the failure message names the block correctly — the semantics the docstring promises.
- The solver-free new test (`test_failed_newton_block_restores_seed`) passes on the PR branch; `set_value(None, skip_validation=True)` round-trips cleanly, so the projection restore of originally-valueless vars works.

### Strengths

- The fix is the right shape: restore-then-report, never report-through, consistent with #229. Snapshot/restore also covers the case where `solver.solve` *raises* after partially loading, and the 1x1 Newton case where `calculate_variable_from_constraint` leaves a garbage iterate before raising.
- `n_vars_initialized`/`n_subsystem_solves` are now incremental and honest on partial failure (the old code reported 0 for both even when earlier blocks had solved).
- Taking the block-loop snapshot *after* seeding (so failures restore seeds, not `None`) while `project_to_feasible` snapshots *before* seeding (so failures restore the exact prior point, including `None`) — both are the right choice for their respective contracts, and both are what the docs now say.

### Issues and suggestions

1. **Inconsistent "ok termination" sets (minor bug, worth fixing in this PR).** The two new sets include `"globallyOptimal"`, but two pre-existing consumers in `repair.py` don't: `InitializeReport.ok` and the `initialize` pipeline's warning check. A `globallyOptimal` projection would now be *kept* (correctly, per the new restore logic) yet `initialize` would warn "continuing with the unrepaired point" and `report.ok` would read False — the exact untruth this PR sets out to kill, from the other direction. There are now four hand-rolled copies of this tuple across two files; suggest one module-level constant (e.g. `_OK_TERMINATIONS`) used everywhere.

2. **The "downstream by construction" claim slightly overreaches.** The comment and failure message justify stopping the loop because "downstream blocks feed on the failed one by construction." Block triangularization linearizes a DAG, not a chain — a later block can be entirely independent of the failed one (parallel branches of a flowsheet). Stopping is still a defensible, conservative choice (and matches the old exception behavior), but the comment states something stronger than what's true. Suggest softening the comment, and possibly a follow-up that continues with blocks not reachable from the failed one — on a big flowsheet, one bad side-chain block currently forfeits initialization of everything after it in the linear order.

3. **Test gap: mid-sequence failure.** All three new block tests fail at the *first* block. The most subtle promise — earlier blocks keep their *solved* values while the failed block restores its *seeds* — is untested. A ready-made test: `a == 4` (solvable 1x1) feeding `x**2 == -a` (unsolvable); assert `a == 4.0`, `x == 3.0` (seed restored), `n_vars_initialized == 1`.

4. **Docstring nit in `project_to_feasible`.** The return doc still says success is `"optimal"` / `"locallyOptimal"`, immediately followed by the new "On any other termination the pre-projection values are restored" — but the restore actually also spares `globallyOptimal`/`feasible`. Once the shared constant from (1) exists, the docstring can reference the same set.

5. **Trivia**: the CHANGELOG hunk adds a double blank line before the `### Added` section; the termination check reads `results.solver.termination_condition` but not `results.solver.status` (string comparison is brittle but consistent with existing codebase style, fine to leave).

### Risk assessment

Low. The refactor reproduces Pyomo's own SCC-solve mechanics (same subsystem-block construction, same input-var fixing, same 1x1/Newton dispatch) while adding the verdict check, and the two non-obvious runtime assumptions plus the partial-failure semantics were verified against the real thing. The one behavior change beyond the fix — warning-status large-block solves now stop the loop instead of continuing on a poisoned point — is the point of the PR. Item 1 is the only thing I'd ask to change before merge; 2–4 are cheap and worth doing while the file is open.

---

jkitchin:
The review's suggested changes are implemented on [`claude/pounce-pr-230-review-149thk`](https://github.com/jkitchin/pounce/tree/claude/pounce-pr-230-review-149thk) (one commit on top of this PR's head): a shared `OK_TERMINATIONS` constant replacing the four hand-rolled tuples (fixes the `globallyOptimal` disagreement in `InitializeReport.ok` and the pipeline warning), a mid-sequence-failure test (solved block keeps its value, failed block restores its seed), softened "by construction" wording, and the docstring/CHANGELOG nits. Merge it into `no-poison` or cherry-pick `8f26eac` to bring this PR up to date.

---

Issue #239, opened 2026-07-20
Release v0.9.0


jkitchin:
Tracking issue for the 0.9.0 release. Raise anything here that should block it.

**Target release date:** TBD

## Where things stand

Work is on the local `release/0.9.0` branch, which is **not yet pushed**. The
version bump and changelog cut are already done:

- All three version surfaces agree at `0.9.0` (root `Cargo.toml`,
  `python/pyproject.toml`, `pyomo-pounce/pyproject.toml`).
- `CHANGELOG.md` has a cut `## [0.9.0]` section; `[Unreleased]` is empty.
- `scripts/check-release-consistency.sh` passes locally: versions agree, the
  publish list is 20 crates matching `cargo metadata` in topological order,
  and every publishable crate's deps carry a crates.io version requirement.
- Full workspace test suite passes locally (exit 0).

## Pre-tag checklist

- [ ] Push `release/0.9.0` and open a PR into `main`
- [ ] CI green on the PR (includes `check-release-consistency.sh`)
- [ ] Re-read the `## [0.9.0]` changelog section — headline entries accurate,
      issue refs correct, nothing landed since the cut that needs an entry
- [ ] Confirm nothing merges into `main` after the cut without a changelog entry
- [ ] Docs build clean; `docs/src/gams.md` still accurate if the GAMS link moved
- [ ] Decide whether anything currently open must land in 0.9.0 (only #149 is
      open — Lean-verified certificates — presumably 0.10.0, confirm)
- [ ] Merge the PR into `main`

## Release checklist (in order — tag prefixes are independent)

- [ ] `git tag v0.9.0 && git push origin v0.9.0` → fires `release-crates.yml`,
      publishes 20 crates in topological order (idempotent; resume a mid-batch
      failure with `--start-from <crate>`)
- [ ] Verify crates.io landed at 0.9.0 (remember the `User-Agent` header, and
      sanity-check against `serde` first)
- [ ] `git tag python-v0.9.0 && git push origin python-v0.9.0` → fires
      `release-pounce.yml` (wheels incl. Windows + sdist → PyPI `pounce-solver`)
- [ ] `git tag pyomo-pounce-v0.9.0 && git push origin pyomo-pounce-v0.9.0` →
      fires `release-pyomo-pounce.yml` (PyPI `pyomo-pounce`)
- [ ] Verify both PyPI projects show 0.9.0
- [ ] Create the GitHub Release **by hand** — no workflow does this, and a tag
      alone does not create one:
[...]

PR #241, opened 2026-07-21
Improve `pounce-rs` README

GermanHeim:
Thanks @GermanHeim! Your README rewrite was merged into `main` via #247, which includes your original commit (dc14c08, authorship preserved) plus two small polish items from the review — switching the Quick start to the documented `prelude::*` import and documenting the `stats` field in the Result section.

Closing as superseded by #247. 🎉

### DIFF
--- a/crates/pounce-rs/README.md
+++ b/crates/pounce-rs/README.md
@@ -1,13 +1,87 @@
 # pounce-rs
 
-Single-crate facade for solving nonlinear programs with [POUNCE](https://github.com/jkitchin/pounce)
-from Rust. Re-exports the `TNLP` problem trait (`pounce-nlp`), the
-`IpoptApplication` driver (`pounce-algorithm`), and the supporting scalar
-types (`pounce-common`) in one place, plus a `prelude`:
+[![crates.io](https://img.shields.io/crates/v/pounce-rs.svg)](https://crates.io/crates/pounce-rs) [![CI](https://github.com/jkitchin/pounce/actions/workflows/ci.yml/badge.svg)](https://github.com/jkitchin/pounce/actions/workflows/ci.yml) [![docs.rs](https://img.shields.io/docsrs/pounce-rs)](https://docs.rs/pounce-rs)
 
-```rust
-use pounce_rs::prelude::*;
+A single-crate entry point for solving nonlinear programs with
+[POUNCE](https://github.com/jkitchin/pounce) in Rust. It provides two APIs:
+
+- a high-level builder API (`Problem` + `Nlp`) for the common case, where only the objective is required and everything else is optional; and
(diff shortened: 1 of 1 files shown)


jkitchin (commented):
## Review — Improve `pounce-rs` README

Docs-only change (1 file, +81/−7). I verified every factual claim in the new README against the crate source.

### Overview
Replaces the terse "re-exports only" README with a proper crate landing page: badges (crates.io / CI / docs.rs), an Install section, a runnable Quick Start on the high-level `builder::{Problem, Nlp}` API, a `Result` section documenting `Solution`, a "Full control" pointer to the low-level `TNLP` trait, and a License line. Good, well-targeted improvement for crates.io discoverability.

### Correctness — all claims verified ✅
- **API surface** — `pounce_rs::builder::{Problem, Nlp}` both exist and are re-exported. `Problem` requires only `objective`; `n_constraints`/`constraints` have defaults (`builder.rs:54-64`). Builder methods `new`/`var_bounds`/`constraint_bounds`/`x0`/`option_num`/`option_int`/`option_str`/`solve` all match.
- **"variable count inferred"** — accurate; `n` is inferred from `var_bounds`/`x0` (`builder.rs:116`, `128-149`).
- **`Solution` fields** — `success`, `status`, `x`, `objective`, `multipliers`, `g`, `z_l`, `z_u` all match (`builder.rs:87-103`), including the "empty if solve aborts before finalization" caveat.
- **FD / L-BFGS defaults** — matches the trait docs (`builder.rs:49-53`).
- **License `EPL-2.0`** — matches `license.workspace = true` → root `Cargo.toml`.
- **Badges** — `ci.yml` exists; crates.io / docs.rs slugs correct.
- **Version pin `pounce-rs = "0.8"`** — currently correct: crates.io max_version is **0.8.0**. Already flagged in the PR description for bump under #239.

### Minor suggestions (optional, non-blocking)
- **Quick Start is not a doctest**, so CI won't compile-check it (unlike the `lib.rs` examples). It reads correctly as-is; consider promoting it to a doctest or keeping it in sync with the verified `lib.rs` example. Minor: the crate documents `use pounce_rs::prelude::*;` as the "common case" entry point, while the README uses the explicit `builder::{...}` path — either is fine, just worth being deliberate.
- **Version-pin durability** — `"0.8"` resolves to `>=0.8.0, <0.9.0`, so it silently stops tracking latest once 0.9.0 ships. #239 already covers this; folding the README bump into the release checklist would keep it from being missed.
- **`Result` section omits `stats`** (`SolveStatistics`) — a public `Solution` field that's arguably the most useful (wall time, iteration count, trajectory). A one-line mention with a pointer to `.capture_iterations()` would round it out.

### Risks
None — docs-only, no code paths touched. The only time-sensitivity is the version pin, already tracked.

**Recommendation:** Approve. Clean and accurate; the notes above are polish, not blockers.

---

Issue #242, opened 2026-07-21
Enhancement: honor max_wall_time with bounded overshoot (currently overruns up to ~7x on large problems)


jkitchin:
## Summary

`Problem.solve` honors the `max_wall_time` option only **coarsely** — the wall
deadline appears to be checked between (outer) iterations, so on a large problem
where a single iteration/KKT factorization is slow, the solve overshoots the
budget by up to a full iteration. Requested budgets are exceeded by up to ~7x.

This is an **enhancement** (not a correctness bug — the result is still valid), but
it makes POUNCE unusable as a time-limited subsolver: a caller that hands POUNCE a
1-second budget cannot rely on getting control back near 1 second.

## Evidence

Measured on a 1611-variable NLP (MINLPLib `emfl050_3_3`, warm-started via
`_solve_with_warm_start`):

```
options={"max_wall_time": 2.0} -> ACTUAL wall 13.2s  (~7x over),  status=TIME_LIMIT
options={"max_wall_time": 5.0} -> ACTUAL wall  0.2s  (converged), status=OPTIMAL
```

So the deadline IS eventually respected (status=TIME_LIMIT), but only after a large
overshoot; when the solve converges quickly there is no overrun (the check simply
never trips). The overrun scales with per-iteration cost, so it is worst on exactly
the large/ill-conditioned problems where a tight budget matters most.

Callers observe the same end-to-end: a downstream solver requesting a 5s solve
sees ~29s wall, ~23s of it inside a single `Problem.solve`.

## Request

Check the wall-clock deadline **more frequently** — ideally at the granularity of
the expensive inner steps (each linear-algebra / line-search step), not only at
outer-iteration boundaries — and return the current best iterate with
`status=TIME_LIMIT` as soon as the deadline is crossed. A cheap monotonic-clock
comparison inside the iteration loop is enough; the goal is bounded overshoot
(e.g. ≤ one cheap step, not ≤ one full iteration).

If a within-iteration check is impractical, a fallback would be to expose a hard
wall cap that aborts the current factorization/step, but the finer-grained check
is preferable (it can return a usable iterate).

## Context

Found while root-causing a wall-budget-overrun issue in a downstream solver
(discopt) that uses POUNCE as its per-node / single-NLP backend. Honoring
`max_wall_time` tightly would let the caller enforce its own time limits reliably.

PR #245, opened 2026-07-21
Check the time deadline between KKT factorizations (bound overshoot to a sub-step)

### LINKED ISSUE #244: Enhancement: bound max_wall_time overshoot to a sub-step (intra-iteration deadline checks) — follow-up to #242
Follow-up to #242 (which added a shared `Deadline` and, importantly, bounded the
previously-unbounded restoration phase). #242 checks the deadline in the
**convergence check**, which runs only *between* outer iterations, so overshoot is
bounded to **≤ one full outer iteration**. On a large NLP a single iteration is
dominated by one expensive step (KKT factorization / linear solve), so a tight
budget still overshoots by that whole step.

## Evidence (post-#242, rebuilt)

Requested `max_wall_time` / `time_limit` = 5 s, measured actual wall:

| problem | vars | actual | overshoot |
|---|---|---|---|
| emfl050_3_3 | 1611 | 16.9 s | **3.4x** |
| emfl100_3_3 | 2961 | 29.6 s | **5.9x** |

Isolated single-NLP probe (same as #242's), unchanged by #242:

```
max_wall_time=2.0s -> ACTUAL 12.7s  status=TIME_LIMIT  (6.4x)
max_wall_time=5.0s -> ACTUAL  0.2s  status=OPTIMAL     (converges, no overshoot)
```

So when the solve converges quickly there is no overshoot; when it doesn't, a
single ~12 s iteration blows a 2 s budget. #242 fixed the restoration grind (a
real win — emfl050 end-to-end dropped 29.5 s -> 16.9 s), but the per-iteration
check granularity leaves the single-expensive-iteration case unbounded to within
one iteration.

## Request

Check the deadline **within** an outer iteration — before/after the expensive
inner steps (linear-algebra factorization / solve, line search, restoration inner
iterations), not only in the between-iteration convergence check. Ideally with
**cooperative cancellation** of the current factorization/solve so the solver can
return the best iterate promptly with `status=TIME_LIMIT`. Goal: bound overshoot
to a cheap sub-step (or a small fixed slack), not a full iteration.

If aborting mid-factorization is impractical, a pragmatic middle ground is to
check the `Deadline` at every inner-iteration boundary of the nested IPMs
(restoration, feasibility) and at each major KKT step, which would cap the worst
case well below a full outer iteration on these large problems.

## Why it matters

The downstream solver (discopt) uses POUNCE as its per-node / single-NLP backend
with **tight** budgets (often < 5 s per node). A 3–6x overshoot on the large-
instance family means the caller still cannot honor its own time limit there,
which caps a time-limited primal heuristic that depends on many short, bounded NLP
solves. The convergent/fast case is already perfect; this is specifically about
the large/ill-conditioned solves where a tight budget matters most.

## Context

Found while re-validating #242 downstream. #242 is necessary and shipped; this is
the remaining increment to make `max_wall_time` tight (bounded overshoot) rather
than "eventually respected within one iteration."

jkitchin:
Fixes #244. Follow-up to #242.

## Problem

#242 bounded `max_wall_time`/`max_cpu_time` overshoot to **one outer iteration** by checking the shared `Deadline` in the between-iteration convergence check, and fixed the unbounded restoration grind. But it checked the deadline only *after* `compute_search_direction` returned. On a large, ill-conditioned NLP a single outer iteration is dominated by the KKT solve, whose inertia-correction / iterative-refinement sweep can run many refactorizations under one search-direction computation — so that whole sweep overshot the budget. This is the "unchanged by #242" case in the issue (2 s budget → 12.7 s actual, 6.4x).

## Change

Thread the same `Deadline` into `PdFullSpaceSolver` and check it **between the major KKT steps**, giving the cooperative cancellation the issue asked for:

- top of the escalation loop in `solve_once` (before each factorization);
- top of the outer solve loop and the iterative-refinement loop in `solve`.

On a crossing the solve aborts by returning `false`, leaving `data.curr` / `data.delta` untouched (they are only promoted on a `true` return), so the algorithm keeps its last accepted iterate. This bounds the overshoot to roughly **one factorization** instead of the whole sweep.

The abort is routed to a clean time-limit termination rather than a step failure:

- the main loop checks `deadline_status()` right after `compute_search_direction` returns — before the `!ok` branch that would otherwise send a deadline-aborted solve into restoration — and terminates with the matching `Wall/CpuTimeExceeded` status;
- `invoke_restoration` checks the deadline right after the inner IPM returns, so a budget crossing inside restoration (its KKT solves now share the deadline too) surfaces as the time limit instead of being mapped onto `RestorationFailure` / `StopAtAcceptablePoint`.

## Scope / what's not done

This implements the issue's **pragmatic middle ground** (check at each major KKT step + at the nested-IPM inner boundaries), which bounds worst-case overshoot to a single factorization. It does **not** implement cancellation *inside* a single factorization/back-solve — if one individual factorization of a huge system runs longer than the budget, that one factorization still completes. Aborting mid-factorization would require cancellation hooks in the FERAL/MUMPS backend and is out of scope here.

## Tests

- New unit tests for the `deadline_exceeded` helper (no deadline / unbounded budget / zero-budget-crossed).
- Existing HS71 wall/CPU integration tests now flow through the KKT-level abort path and pass unchanged; full `pounce-algorithm` (276) and `pounce-restoration` (110) lib suites pass. No new clippy warnings.

https://claude.ai/code/session_01Sbp5ftAv68sr4QdnXBWMDw

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sbp5ftAv68sr4QdnXBWMDw)_

### DIFF
--- a/crates/pounce-algorithm/src/ipopt_alg.rs
+++ b/crates/pounce-algorithm/src/ipopt_alg.rs
@@ -1043,6 +1043,21 @@ impl IpoptAlgorithm {
             tracing::debug!(target: "pounce::linsol", "kkt solve complete");
             drop(ls_enter);
             timing.compute_search_direction.end();
+            // Fine-grained time-budget gate (pounce#244). The KKT solve now
+            // checks the shared deadline *between* its major factorization
+            // steps (inertia correction / iterative refinement) and aborts
+            // cooperatively when the budget is crossed — bounding the
+            // overshoot to roughly one factorization instead of the whole
+            // multi-factorization sweep that #242's post-solve check let run
+            // to completion. Whether the solve returned a completed step or
+            // bailed mid-escalation, if the deadline tripped, stop here with
+            // the time-limit status *before* the `!ok` branch below would
+            // otherwise route a deadline-aborted solve into restoration.
+            // `data.curr` is untouched by the step computation, so it still
+            // holds the last accepted iterate.
+            if let Some(ret) = self.deadline_status() {
+                return IterateOutcome::Terminate(ret);
+            }
             if !ok {
                 // Mirror upstream `IpIpoptAlg.cpp:417-430`: a failed
                 // step computation puts the algorithm in emergency
@@ -1702,6 +1717,18 @@ impl IpoptAlgorithm {
         self.resto_inner_iters = self
             .resto_inner_iters
             .saturating_add(resto.last_inner_iter_count());
+        // pounce#244: the restoration inner IPM shares the outer solve's
+        // `Deadline` (both its convergence check and — post-#244 — its KKT
+        // solves consult it), so a budget crossing inside restoration
+        // terminates the inner solve with a time-limit status. Surface that
+        // as the time limit directly instead of letting the `Failed` arm map
+        // it onto `RestorationFailure` / `StopAtAcceptablePoint`. `data.curr`
+        // is the last accepted outer iterate — restoration stages its
+        // recovered point onto `trial`, not `curr`, and we return before
+        // promoting it — so this hands back a valid iterate.
+        if let Some(ret) = self.deadline_status() {
+            return IterateOutcome::Terminate(ret);
+        }
         match outcome {
             RestorationOutcome::Recovered => {
                 // The driver has staged the recovered point on
--- a/crates/pounce-algorithm/src/kkt/pd_full_space_solver.rs
+++ b/crates/pounce-algorithm/src/kkt/pd_full_space_solver.rs
@@ -254,6 +254,14 @@ impl PdFullSpaceSolver {
         let mut improve = improve_solution;
 
         while !done {
+            // pounce#244: bail between major KKT steps when the shared time
+            // budget is crossed (see `deadline_exceeded`). Returning `false`
+            // routes through the caller's post-KKT deadline check, which
+            // terminates the solve with the time-limit status rather than
+            // treating the abort as a step-computation failure.
+            if deadline_exceeded(data) {
+                return false;
+            }
             let solve_ok = if improve {
                 true
             } else {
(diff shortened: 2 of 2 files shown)


jkitchin:
## Empirical confirmation (emfl family, #244 vs #242 baseline)

The exact `emfl050_3_3`/`emfl100_3_3` fixtures aren't in this repo (they come from the downstream discopt/GLOBALLib bench-data tree, and this environment has no `pyomo`/`ampl`/`ipopt`), so I generated **`emfl`-family instances** — Euclidean multifacility location, sum of `sqrt` distances → indefinite Hessians → the inertia-correction refactorization thrash this issue is about — and ran this branch head-to-head against its parent `bb67314` (has #242, not #244). "solver time" is the internal `Total seconds in POUNCE` the deadline governs (excludes `.nl` parse/setup).

**Headline — n=1600 (≈ emfl050's 1611 vars), all-pairs indefinite, `max_wall_time=2 s`:**

| build | solver time | overshoot | status |
|---|---|---|---|
| #242 baseline | 10.20 s | **5.10x** | Maximum wallclock time exceeded |
| #244 (this PR) | 3.78 s | **1.89x** | Maximum wallclock time exceeded |

`iters=0` in both — the entire cost is inside iteration 0's single search-direction computation (the dense inertia-correction escalation loop), exactly the case #242's post-solve check could not interrupt.

**Across regimes:**

| instance | per-iter cost | budget | #242 | #244 |
|---|---|---|---|---|
| emfl800 (sparse, converges slowly) | ~8 ms | 2 s | ~1.0x | ~1.0x |
| emfl_hard (n=2800) | ~226 ms | 2 s | 1.07x | 1.02x |
| emfl_vhard (n=1000, dense) | ~1.3 s | 1 s | **2.67x** | **1.09x** |
| emfl_vhard800 (n=1600, dense) | iter-0 ~10 s | 2 s | **5.10x** | **1.89x** |

**Takeaways**
- Reproduces the reported 3.4–6.4x overshoot on the single-expensive-iteration case and cuts it to ~1.1–1.9x.
- On sparse/cheap-iteration solves #242 was already tight and #244 does not regress it.
- Every run returns the correct `Maximum wallclock time exceeded` status (not `RestorationFailure`/step-error) — the abort routing works.

**Measured caveat:** #244's residual overshoot equals **one factorization** (the 1.89x above ≈ one dense ~1.8 s LDLᵀ on the 1600-var system). That's the documented limit — the deadline is checked *between* factorizations but not *inside* one; cancelling mid-factorization would need FERAL/MUMPS interrupt hooks (out of scope). So overshoot is bounded to a KKT sub-step, matching the issue's "pragmatic middle ground," rather than to a fixed small slack.

---

PR #255, opened 2026-07-21
dev-notes: active-set-aware parametric sensitivity design record

devin-griff:
One document, `dev-notes/sensitivity-design.md`, recording the active-set-aware parametric sensitivity system as shipped: the three modes and their costs, the report, the interior point background and the sensitivity system in Biegler's notation, the degeneracy decision with all three options, and the corrector's predicted-point operator. It describes the surface as merged, `sens_solution` and its siblings after the naming pass, `release_all` included, with the barrier-correction back-solve and the measured `path` behavior for an undecided weak bound stated rather than the advertised versions. Nothing in it proposes anything.

Rendered: https://github.com/devin-griff/pounce/blob/sensitivity-design-wip/dev-notes/sensitivity-design.md


devin-griff:
Hi @jkitchin , I find it useful to work in this mode of spec driven development. Posting this ahead of development for review and brainstorming.

jkitchin:
## Review — sensitivity roadmap (docs-only)

Reviewed as a design note: one new file (`dev-notes/sensitivity-roadmap.md`, +288), no code touched. Verified every concrete code reference against `main`.

### Accuracy — strong
All the load-bearing references check out:
- `estimate()` at `sens.py:598`; `x_new = base_x + dx` at `:623`; the clamp warns before clipping. ✓
- `parametric_step` at `solver.rs:360` via `IndexSchurData`. ✓
- `boundcheck.rs` gated by `sens_boundcheck`; the quoted "simpler single-pass clamp rather than upstream's iterative Schur-refinement loop" is verbatim its module doc. ✓
- `sens_boundcheck` registered default `false` with the exact "Mirrors upstream `sens_boundcheck` (SensApplication.cpp:63)" comment. ✓
- `pounce-convex/src/ipm.rs` exists (the item-3 QP target). ✓
- **Central premise confirmed:** `boundcheck.rs:31` points at #7, and #7 is **closed (completed)** — so the note is right that the fix-relax follow-up has no open tracking issue. That's the most actionable takeaway; worth actually filing that issue so the claim stays true.

The technical framing (sIPOPT = fix-relax + eq. 10 µ-correction; eq. 14 / path-following / correctors left unimplemented) matches both the 2012 paper and the original #7 scope. No overclaiming found.

### Validation plan — good coverage, under-specified thresholds
It does propose known-answer / easily-derived cases: the paper's §2.8 parametric QP (documented active-set change), a constructed weakly-active case + finite differences, and brute-force ratio-test as breakpoint ground truth. Three items (fix-relax, path, corrector) use a full `solve()` re-solve as the exact oracle. The cstr demo check is qualitative ("flattens today, should bend under `fix_relax`").

Gap vs. its own parent issue: **#7 pinned numeric acceptance targets (`dx*/dp` to ~1e-8 on named `cstr1`/`cstr2`)**; this roadmap names *what* to compare against but no tolerance or named test problem per item. Recommend pinning an ε + a named case to each validation bullet so the eventual implementation PRs are gateable.

### Minor
- **Snippet line drift:** the clamp is annotated `# sens.py:636 (warns)`, but the clip is at `:638` and the warning is a separate `clipped.any()` branch (~631–637), not the clip line. Hard-coded `:NNN` in prose rots — prefer symbol names.
- **External links to the author's own projects:** `github.com/devin-griff/drto` and the `cstr-sensitivity.griffith-pse.com` demo are load-bearing references here. Reasonable as downstream-consumer citations, but flagging so embedding a contributor's personal repo/domain into project dev-notes is a deliberate call.
- **Unverified citation:** "the paper's §2.8 parametric QP" — plausible but I couldn't confirm the section number against the actual Pirnay/López-Negrete/Biegler paper. Worth a check before it becomes *the* fix-relax oracle.

### Recommendation
Approve as a design note — it's well-grounded in the real code and honest about what's proposed vs. done. Two small asks before merge: fix the `:636`→`:638` / "(warns)" annotation, and make the `drto` / external-domain links a conscious decision. Pinning validation tolerances can follow in the implementation PRs.

---

devin-griff:
@jkitchin addessed the mentioned issues and added a note on the relation to pounce.jax.

devin-griff:
Minor reference corrections. Item 3 now states what `estimate()` and `gradient()` return at a degenerate base point, and item 0's classifier cites item 0 of the covariance roadmap, which #262 now carries.

The rest has been broadly revised for clarity.

jkitchin:
## Status check: the roadmap has been overtaken by its own implementation

Picking this up after a long gap. Nothing here is blocked — CI is green (12/12),
the branch is `MERGEABLE`/`CLEAN`, and there are no unresolved review threads. But
the PR is still a **draft** and still reads as a proposal for work that has since
entirely shipped on `main`:

| item | shipped as | merged |
|---|---|---|
| 0 diagnostics / `estimate_report()` | #584 | 2026-08-13 |
| 1 fix-relax + barrier (μ) correction | #587 | 2026-08-14 |
| 2 `mode="path"` | #631 | 2026-08-16 |
| 3 `degeneracy="directional"` | #672 | 2026-08-18 |
| 4 `corrector_iter` | #733 | 2026-08-21 |

plus later hardening (#800, #829, #830, #832, #837). All of it landed *after*
0.10.0 (released 2026-08-11), so it is in Unreleased/0.11 — "targeting v0.10" is
no longer right either.

The document now contradicts itself: the header calls items 1–4 proposals while
the body measures "the shipped search" and quotes fix_relax/path timings.

### Closing the one open item from my review

I flagged "the paper's §2.8 parametric QP" as an unverified citation. **It checks
out.** Against the [optimization-online
preprint](https://optimization-online.org/wp-content/uploads/2011/04/3008.pdf),
§2.8 is "Example Problem" — `min x1^2+x2^2+x3^2` subject to two equalities and
`x >= 0` — introduced as covering "a change in the active set" and illustrating
"the capabilities of the fix-relax strategy". Eq. (10) is the mu-corrected update
and eq. (14) is the directional QP, with §2.4 stating verbatim that "the current
version of sIPOPT does not include an implementation of (14)". Every citation in
the note lands.

Two things came out of checking, though:

1. The paper calls it "a small parametric optimization problem", not a QP.
   Constraint (26c), `p2*x1 + x2 - x3 - 1 = 0`, is bilinear in `(p2, x1)`, so it
   is a QP in `x` only at fixed `p`. Worth dropping the "QP".
2. **The repo now carries two incompatible equation numberings.**
   `solver.rs:1137` and `docs/src/sensitivity.md:410` call the directional QP
   eq. **14**, which matches the preprint. But `boundcheck.rs:21-36`,
   `sens_resolve_oracle.rs:67` and `test_fix_relax.py:269` put fix-relax at
   **§2.5** with pin/release at eqs. **17/18** and the Schur form at **19-22** —
   exactly +1 on the preprint's §2.4, (16)/(17), (18-21). One of the two is keyed
   to a different edition. The published version is closed access so I could not
   settle which from here. Whichever way it goes, the citations should say which
   edition they are keyed to, and ideally name the object ("the fix-relax pin
   equation") rather than a bare number — the same lesson as 71fc606b about line
   numbers.

### What I think finishes this

Reframe it as a design record and merge it, the way #262 did for the covariance
roadmap (see af5dc2db, "reframe the roadmap as the final design record"). The
measured studies in here — the corrector refactorization policy table and the
item 3 findings from the double column — are the most valuable content in the
repo on this subsystem and should not stay unmerged.

Stale passages, as I read them:

1. **Status header** — "Item 0 is implemented and merged. The rest is a
   proposal", and "targeting v0.10".
2. **"Where we are"** — claims `estimate()` is "just the linear predictor today"
   and shows the `np.clip` snippet as current behavior. Shipped signature:
   `estimate(model, perturb, clamp=True, mode="linear", predictor_iter=16,
   degeneracy="directional", degeneracy_iter=16, corrector_iter=0,
   bound_eps=None, max_pdpert=None)`.
3. **"The `boundcheck.rs` pointer to pounce#7 ... is stale; no open issue tracks
   this"** and "it should get its own issue" — both moot. `boundcheck.rs` no
   longer references #7; its module doc documents `refine_step_onto_bounds` and
   both halves of fix-relax.
4. **API surface table** is the proposed surface, not the shipped one. It says
   item 3 is "not on the surface ... not a user knob" (shipped as `degeneracy=`
   plus `degeneracy_iter`), and describes the corrector as "a convenience wrapper
   that loops to a residual tolerance with an iteration cap and a stagnation
   guard" (shipped as a plain `corrector_iter=n` count). `bound_eps`,
   `predictor_iter` and `max_pdpert` are absent.
5. **"Design consequence for item 3"** (single-violator pivoting plus a regime
   guard) has been overtaken by what shipped: `parametric_step_directional`
   (`solver.rs:1183`) solves the eq. 14 dual through `pounce-qp` over the weak
   rows the direction engages, gated on `kappa = sigma * S_kk`. That gate
   subsumes the "equality-pinned weak rows must start released" recommendation —
   such a coordinate has `S_kk` exactly zero and is dropped by the same test.
6. **Validation section** is still future-tense with no tolerances. The tests
   exist now and can be named instead: `pyomo-pounce/tests/test_fix_relax.py`,
   `test_estimate_report.py`, `crates/pounce-sensitivity/tests/sens_invariance_legs.rs`,
   `sens_resolve_oracle.rs`.

Worth pointing at `docs/src/sensitivity.md` for the user-facing surface, as the
merged covariance record does.

@devin-griff — happy for you to take the rewrite, or say the word and I will push
the reframe to the branch.

---

devin-griff:
@jkitchin The dev note has been updated to reflect the functionality as implemented and should be ready to merge.

jkitchin:
Woohoo!

Issue #200, opened 2026-07-11
False Solve_Succeeded on flat quartic objectives (quartc, dqrtic): gradient-based scaling trips convergence far from the minimum


jkitchin:
## Summary

On the convex, unconstrained quartic problems **`quartc`** (n=10000) and **`dqrtic`** (n=5000)
from the Vanderbei/CUTE suite, POUNCE reports **`Solve_Succeeded` at a point that is not the
minimum**. Each has a unique minimum with objective ≈ 0, but POUNCE stops at obj ≈ 248.88
(quartc) / ≈ 39.36 (dqrtic) and declares optimality.

The minimum *is* findable — disabling objective scaling reaches it:

| Problem | default (`nlp_scaling_method=gradient-based`) | `nlp_scaling_method=none` |
|---------|-----------------------------------------------|----------------------------|
| quartc  | obj **248.88**, 24 iters, "Optimal"           | obj **6.8e-09**, 39 iters, "Optimal" |
| dqrtic  | obj **39.36**, 23 iters, "Optimal"            | obj **1.1e-09**, 38 iters, "Optimal" |

## Root cause

The default gradient-based objective scaling picks `obj_scaling_factor ≈ 1e-8` from the large
initial gradient (`∇f = 4(x−1)^3`, which is O(1) at the start `x0=2` but summed over 10^4 vars).
The convergence test then runs on the **scaled** gradient. POUNCE's own final report shows it:

```
Objective........:  2.4888e-06 (scaled)   2.4888e+02 (unscaled)
Dual infeasibility: 8.38e-09              <-- scaled; below tol 1e-8
```

The scaled dual infeasibility (8.4e-9) passes `tol=1e-8`, but the **true unscaled gradient is
~0.84** — nowhere near a KKT point. Because the objective gradient of a quartic vanishes cubically
near the minimum while the scaling factor stays fixed at its initial value, the scaled convergence
test trips ~30% of the way to the optimum and POUNCE reports success at a clearly non-optimal
point. This is a genuine convergence-quality bug (a false "Optimal"), not merely a
reference-annotation mismatch.

(ipopt-ma57 exhibits the same behavior — it also defaults to gradient-based scaling and stops at
the same value — so this is a shared gradient-scaling failure mode, not a POUNCE-vs-ipopt
discrepancy. But POUNCE declaring `Solve_Succeeded` at a non-minimum is still wrong.)

## Suggested direction

The termination test should not certify optimality when the **unscaled** dual infeasibility is far
[...]

PR #259, opened 2026-07-22
Make the dual-divergence guard opt-in, and its diversion non-destructive (#250 follow-up)

jkitchin:
Found by adversarially diffing the last five merged PRs over 400 MINLPLib models (baseline `f5aea43` vs `main`), then bisecting to `ba29b53`.

Two changes to the `dual_diverging_streak` guard added in #250: it becomes **opt-in**, and when opted into, its diversion can no longer leave the solve worse off than a point it already had.

## 1. The guard is now off by default

It shipped default-on at `15` to bound a reported emfl050 bad-warm-start grind. That justification did not survive reproduction: the reported `11.7 s / iterations=0` measurement is caller-side JAX compilation (it follows call order, not the guard, in both orders and under both settings), and the build predating the guard solves both emfl050 instances to the same optimum in the same time.

What remained is an effect on four of 1284 MINLPLib models, knife-edge and non-monotone in its own threshold:

| model | behaviour |
|---|---|
| `deb7` / `deb9` | better local optimum (104.95 → 97.56) at **exactly** 15, at no other value tried (0, 5, 25, 32, 40, 60) |
| `pooling_rt2stp` | `Solve_Succeeded` → `Maximum_Iterations_Exceeded` at 10 and 15 only; solves cleanly at 0, 5, 25, 40 |

And the decisive measurement — **the effect differs by host in sign**:

| host | guard off | guard on (15) |
|---|---|---|
| macOS / FERAL | 104.95 | **97.56** |
| Linux CI runner | 97.56 | **127.87** |

Same source, same fixture, opposite sign. A heuristic whose benefit depends on the host is basin luck on nonconvex problems, not a property of the algorithm. The two sides are also not commensurate: the upside is a better local optimum on an already-solved problem, the downside is a clean solve becoming a failure. So it stays available but is no longer imposed — set `dual_diverging_streak=15` to restore the old behaviour.

`pooling_rt2stp_solves_at_default_settings` fails if the default is flipped back.

## 2. When enabled, a lost bet is no longer destructive

On `autocorr_bern55-06` at streak 15:

| | status | objective | NLP error | iters |
|---|---|---|---|---|
| before | Solved To Acceptable Level | −2263.46 | **1.0** | 111 |
| after | Solved To Acceptable Level | −2304.00 | 3.7e−9 | — |
| Ipopt (oracle) | Optimal | −2304.00 | 3.7e−9 | 72 |

The guard fires at iteration 23. The diverted run then **reaches the true optimum and holds it from iteration 57 to 86** — but the dual residual sawtooths between 1e−8 and 2e−1 there, so it never strings together the `acceptable_iter` consecutive qualifying iterates that would stop the solve. It enters restoration a second time, wanders into a worse basin, and returns a point that is feasible (constraint violation 3.3e−12) but nowhere near KKT.

The better point had already passed the acceptable test. It was overwritten only because `store_acceptable_point` keeps the **latest** rather than the **best** — the exact hazard the gh #200 comment already flags for the veto path.

So POUNCE records the best acceptable-quality iterate and hands it back if the diverted run ends worse. Same "never worse off" bargain the #200 veto makes, at the same single finalization point so every `return` in the loop flows through it by construction.

Details worth review:
- Recording is **not** gated on the guard having fired; only its *use* is. Gating the recording was the first attempt and left a hole — the guard returns to the driver before the recording site on the iteration it fires, so nothing at or before the diversion was captured. Solves the guard never touches are still bit-identical, since the record is never read there.
- A strict `Success` is never overridden; a lower objective at a merely-acceptable point must not displace a real certificate.
- Statuses carrying a fact of their own are preserved — a restored point never relabels `MaxiterExceeded`/`CpuTimeExceeded`/`WallTimeExceeded`/`UserRequestedStop`, so a caller polling "did I run out of time" isn't told "solved to acceptable level" because a better point happened to be recoverable.
- Scaled objectives compared only under an unchanged `obj_scaling_factor`; comparison is a negated `<=` so a non-finite objective loses rather than silently winning.

### What this does not promise

It guarantees a diverted run never returns worse than the best acceptable point *that run visited*. It cannot make the diversion no worse than not diverting — that counterfactual solve never happened. On Linux the guard costs `deb7` 30 % of its objective and this fallback cannot recover it, because 127.87 *is* the best acceptable point the diverted run reached. That gap is a large part of why the guard is off by default rather than fixed with more machinery.

`deb7` therefore carries **no absolute assertion** in the tests. Two earlier versions tried to pin one and CI falsified both. It is exercised instead by the relative hair-trigger comparison, which is host-independent by construction.

## Verification

- `autocorr_bern55-06` → −2304.0000277 (**3.6e−11** from Ipopt), stationary KKT error
- `pooling_rt2stp` solves cleanly at the new default
- 400-model corpus re-sweep vs baseline: no objective regressions
- Full-corpus effect of this release's changes vs `f5aea43` over all 1284 models: 0 regressions, 4 status improvements (`jit1`, `nvs04`, `heatexch_spec2`, `supplychainp1_030510`), one objective improvement (`st_e35`)
- Full workspace suite: 2052 tests, 0 failures
- All four new tests fail on the parent commit

## Note on the guard's own coverage

No test pins the guard's *firing* behaviour on its original motivating scenario; `emfl050_5_5` solves in 91 iterations with the guard disabled, so that scenario isn't regression-protected. Out of scope here, but worth a follow-up.


jkitchin:
## Scope has grown — now includes a default change

Widening the guard on/off sweep from 500 to **all 1284** MINLPLib models surfaced a second harm the smaller sample missed, and it changed the conclusion.

### The new harm

`pooling_rt2stp`:

| | status | objective | NLP error (unscaled) |
|---|---|---|---|
| guard off | Solve_Succeeded | −3273.955 | 1e−8 |
| guard on (default 15) | **Maximum_Iterations_Exceeded** | −3265.566 | **18.0** |
| Ipopt | Optimal | −3273.955 | — |

The best-acceptable fallback **cannot** rescue this one: the diverted solve never reaches an acceptable point at all, so there is nothing to hand back. The fallback bounds the objective downside but cannot manufacture a point the solve never visited. That makes it a defaults question, not more machinery.

### The effect is noise, not a property

Sweeping the guard's own threshold shows a knife-edge, non-monotone response:

| streak | deb7 / deb9 | pooling_rt2stp |
|---|---|---|
| 0 (off) | 104.95 | ✅ Solved |
| 5 | 104.95 | ✅ Solved |
| **10** | 104.95 | ❌ MaxIter |
| **15 (old default)** | ✅ 97.56 | ❌ MaxIter |
| 25 / 32 / 40 / 60 | 104.95 | ✅ Solved |

`deb7`/`deb9` improve at *exactly* 15 and nowhere else; `pooling_rt2stp` breaks at 10 and 15 only. No threshold separates them.

### Why default-off rather than counting 3 helps vs 1 harm

The two sides aren't commensurate. The upside is a **better local optimum on an already-solved nonconvex problem**; the downside is a **clean solve becoming a failure**. POUNCE makes no global-optimality claim, so trading a status for a basin isn't a trade to impose on every user — especially when the mechanism is demonstrably arbitrary and the original justification is refuted (the emfl050 grind is caller-side JAX compilation; the pre-guard build solves both emfl050 instances to the same optimum in the same time — see #246 and #250).

`dual_diverging_streak=15` restores the old behaviour, and the never-worse-off fallback applies when enabled.

## Also in this branch

**Pre-diversion gap closed.** Recording was gated on `dual_guard_fired`, but the guard returns to the driver *before* the recording site on the iteration it fires — so nothing at or before the diversion was captured. `autocorr_bern55-06` hid it (its better point arrives at iteration 86, long after the guard fires at 23). Recording now happens on every acceptable iterate; still behaviour-neutral, since the record is only *read* under `dual_guard_fired`, and the iterate is cloned only on an actual improvement.

## Tests

Every guard test now enables the guard explicitly via a `GUARD_ON` constant — otherwise they'd exercise nothing and pass vacuously once the default flipped. Added:
- `pooling_rt2stp_solves_at_default_settings` — pins the decision, fails if the default is turned back on
- `hair_trigger_guard_never_degrades_the_answer` — `dual_diverging_streak=1` moves the diversion as early as it can go, the sharper probe for the gap autocorr couldn't expose
- options-wiring test asserts the registered default is now 0

## Full-corpus verification

All five recent PRs plus these follow-ups, against the commit preceding them (`f5aea43`), over **all 1284 models**:

**0 regressions.** 4 improvements — `jit1`, `nvs04`, `heatexch_spec2`, `supplychainp1_030510` all move from a limit/failure status to solved. Plus one objective improvement (`st_e35` 21357.40 → 21355.24, feasible to 1.8e−12).

Workspace suite: 2054 tests, 0 failures.

Issue #257, opened 2026-07-22
Spurious UNBOUNDED persists on jit1's real B&B node subproblems despite #252 (scaled + ub=+inf mix; 25 cyipopt overturns/solve)


jkitchin:
## Summary

Follow-up to **#252** (closed by ecb6f82, "Fix spurious UNBOUNDED on unbounded-box
subproblems with a finite optimum"). That fix passes its own `repro_issue252.rs`, but
it does **not** cover the subproblems discopt's spatial B&B actually generates for
**jit1** — POUNCE still returns spurious `UNBOUNDED` on them. Measured against the
current `main` (ecb6f82) rebuilt locally:

- jit1 solves end-to-end **only** because discopt's cyipopt-retry-on-UNBOUNDED
  workaround overturns the false verdicts — the retry **fires 25 times** in one solve.
- With that workaround disabled, **59/59** node NLP solves return `UNBOUNDED`, jit1
  returns `status=unknown` / no incumbent (SCIP solves jit1 in <0.1 s).

So #252 fixed a synthetic proxy but is a no-op on the real class; the discopt-side
workaround remains load-bearing and cannot be removed.

## The exact failing subproblem

jit1.nl is **25 vars / 32 constraints** (already a vendored fixture), so each failing
node is jit1's **continuous relaxation on a tightened B&B box**. The first
POUNCE-UNBOUNDED node, captured *at the instant POUNCE reports it* (same evaluator /
x0 / bounds handed to cyipopt):

```
n_vars = 25   n_constraints = 32
variables with ub = +inf: 9   (lb finite, some negative)
finite bound magnitude range: |lb,ub| in [2.2e-4, 70]
x0: the 12 box-bounded vars ~ 4e-3 .. 5e-2 ;  the 9 ub=+inf vars started at x0 ≈ 50
POUNCE  -> UNBOUNDED
cyipopt -> OPTIMAL   (obj = 173345.377)
```

The distinguishing feature vs the #252 repro: an **extreme scale mix** — 12 tightly
box-bounded tiny variables (the reciprocal-objective `Σ cᵢ/xᵢ` terms, ~1e-4) sitting
next to 9 `ub=+inf` variables initialized far out at ~50, with a badly scaled linear
tail (jit1's coefficients span ~10 to ~1e7). POUNCE's unboundedness test fires on a
large step along an `ub=+inf` direction even though the objective is bounded below;
#252's "finite optimum exists" logic doesn't trigger for this scaled/mixed case.

[...]

Issue #173, opened 2026-06-30
IPM reports `OPTIMAL` on stalled/ill-conditioned solves → drifted objective & untrustworthy duals (blocks HiGHS-free dual consumers in discopt)


jkitchin:
## Context

discopt is removing its HiGHS dependency to become pure-Rust. The blocker is dual information: `get_exact_dual_lp_solver` is HiGHS-only because POUNCE's duals can't currently be trusted for sound bound tightening (discopt #145, #356, #359).

## The core problem

On ill-conditioned problems POUNCE returns Ipopt status `OPTIMAL` while the iterates have stalled short of tolerance, so the objective has drifted and the multipliers (`mult_g`, `mult_x_L`/`mult_x_U`) are not at a converged KKT point. A consumer that trusts the `OPTIMAL` status then reads inconsistent duals.

## Why it matters for the consumer

discopt's safe bound (Neumaier–Shcherbina) is valid for **any feasible** dual `y ≥ 0`, so it does **not** need vertex duals — interior-point duals are fine. It only needs to trust that the solve actually converged.

Critically, discopt already follows a **"verify the certificate yourself, don't trust the solve"** pattern (the whole #356/#357 design — N–S safe bound, Farkas re-verification — does exactly this). So POUNCE does **not** have to change its status semantics. If POUNCE simply *exposes the raw KKT residuals*, discopt can apply its own threshold and degrade gracefully, the same way it already does for the Rust-simplex certificates. That makes residual exposure the minimal, backward-compatible unblock.

### Primary — expose final KKT residuals behind an optional flag

An opt-in flag (e.g. `report_residuals=True`) that attaches the final KKT residuals to the `info` dict. This alone unblocks discopt, because we threshold and verify on our side. Backward-compatible: flag off = today's behavior, zero risk to other POUNCE users.

For the residuals to actually unblock us they need three properties:

1. **The right components** — stationarity, complementarity, and primal/dual feasibility residuals (enough to threshold a true KKT point), not just the objective.
2. **Original / unscaled space** — unscaled back the same way the duals are, so the threshold is meaningful on exactly the ill-conditioned systems where this matters (mirrors the `unscale_dual`/`unscale_ray` work in discopt #357).
3. **Consistent with the returned duals** — residuals and `mult_g`/`mult_x_*` evaluated at the same point.

A convenience boolean (`converged_within_tol`) is fine to include but is derivable from the residuals, so it's secondary.

[...]

PR #261, opened 2026-07-22
Fix false Infeasible_Problem_Detected on a feasible, aggressively scaled NLP

jkitchin:
Surfaced by an adversary agent while testing #250 and set aside as out-of-scope. It predates all of the recent PRs — reproduces identically with `dual_diverging_streak=0`.

## A feasible problem reported infeasible

Hock–Schittkowski 13, `f* = 1`, started from `x₀ = (1e4, 1e4)`:

| | status | objective | iters |
|---|---|---|---|
| POUNCE (before) | **Infeasible_Problem_Detected** | 0.28583751 | 12 |
| Ipopt | Optimal | 0.98492872 | 29 |
| POUNCE (after) | Optimal | 0.98492872 | 29 |

## Why the verdict was false on its own terms

`LocalInfeasibility` asserts the iterate has converged to a **stationary point of the constraint violation**. That's checkable, and it fails at the point POUNCE returned, `x = (1.5698, 0.31744)`:

```
constraint violation   0.51     (user units; 1.7e-7 scaled)
‖∇θ‖                   1.40     (nowhere near stationary)
active bounds          none     (nothing blocks descent)
```

A single step downhill from that point reaches a feasible point.

## Cause: the two halves of the test live in different spaces

Detection fires when the violation is bounded away from zero **and** the stationarity `‖Jᵀc‖ / max(1, ‖c‖)` is ~zero. But:

- violation half → **unscaled** residual (where `constr_viol_tol` is defined, #173)
- stationarity half → **scaled** residual

The scaled measure carries a factor `dc²`. Here the starting Jacobian is ~3e8, so gradient-based scaling picks `dc ≈ 3.3e-7`, and the stationarity reads ~5e-14 — comfortably under `infeas_stationarity_tol=1e-8` — independently of where the iterate is. Both gates pass at a non-stationary point.

**Same class of bug as #257**: a two-part test where one half is scale-invariant by accident and the other isn't.

## Fix

Measure the stationarity unscaled. Only rows are scaled and POUNCE applies no variable scaling, so no unscaled Jacobian is needed:

```
J_user = diag(dc)⁻¹ J_scaled,   c_user = dc⁻¹ ⊙ c_scaled
  ⇒  J_userᵀ c_user = J_scaledᵀ (c_scaled ⊘ dc²)
```

Divide the residual by `dc²` before the existing transpose-product. With no scaling active the measure is unchanged.

## Blast radius: zero on the corpus

**Bit-identical across all 1284 MINLPLib models.** The failure needs an extreme constraint scaling *and* a remote starting point together, which that corpus doesn't exercise. A separate scan confirms it: of the models the default calls infeasible, every one is still called infeasible with scaling disabled — none was a false verdict of this kind.

Worth fixing anyway because of the asymmetry: a false *unbounded* verdict is loud and a driver can retry it; a false *infeasible* silently prunes a branch-and-bound node that may contain the optimum.

## Tests

Four, all failing on the parent commit: the verdict, the objective, agreement across both scaling paths, and — assumption-free — feasibility of the returned point by evaluating HS13's constraint closed-form on the returned `x`.

That last one earned its keep. An earlier version derived the unscaled violation from the report by dividing by the **objective** scale factor, which is a different factor from the constraint row scale; it reported 3.4e-5 where the true violation was 0.51, and passed pre-fix. Evaluating the constraint directly avoids the whole class of mistake.

Full workspace suite: 2053 tests, 0 failures. fmt and the CI clippy invocation clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,35 +11,41 @@ changes.
 
 ### Fixed — false `Infeasible_Problem_Detected` on a feasible, aggressively scaled NLP
 
-- **Rapid infeasibility detection now measures stationarity in the same space as
-  the violation it is paired with.** The detector fires only when the constraint
-  violation is bounded away from zero *and* the infeasibility stationarity
-  `‖Jᵀc‖ / max(1, ‖c‖)` is ~zero — no local move reduces the violation. Those two
-  halves were evaluated in different spaces: the violation on the **unscaled**
-  residual (where `constr_viol_tol` is defined, pounce#173), the stationarity on
-  the **scaled** one. The scaled measure carries a factor `dc²`, so an aggressive
-  constraint scaling collapses it toward zero on its own, independently of where
-  the iterate actually is, and both gates pass at a point that is nowhere near
-  stationary.
+- **Rapid infeasibility detection now confirms its own claim before issuing it.**
+  The detector fired on a violation gate plus a stationarity *surrogate*,
+  `‖Jᵀc‖ / max(1, ‖c‖)`, against an absolute tolerance. That surrogate is not
+  scale-invariant: under a row scaling `dc` the numerator carries `dc²` while the
+  denominator clamps at 1, so an aggressive scaling drives it toward zero
+  regardless of where the iterate actually is.
 - **Symptom it fixes.** On Hock–Schittkowski 13 from `x₀ = (1e4, 1e4)` the
-  starting Jacobian is ~3e8, gradient-based scaling picks `dc ≈ 3.3e-7`, and
-  POUNCE stopped after 12 iterations reporting
-  `Infeasible_Problem_Detected` — on a *feasible* problem with `f* = 1`. The
-  point it returned had a constraint violation of **0.51** in the user's units
-  (a reassuring `1.7e-7` scaled), `‖∇θ‖ = 1.40`, and no active bound blocking
-  descent; one step downhill from it reaches a feasible point. POUNCE now
+  starting Jacobian is ~3e8, gradient-based scaling picks `dc ≈ 3.3e-7`, and the
+  surrogate read `5e-14` — far under the `1e-8` tolerance — at a point whose
+  constraint violation was **0.51**, whose `‖∇θ‖` was 1.40, and where neither
+  bound was active to block descent. POUNCE reported
+  `Infeasible_Problem_Detected` on a *feasible* problem with `f* = 1`. It now
   converges to `0.98492872` in 29 iterations, matching Ipopt from the same start
-  to 9 significant figures, iteration for iteration.
+  to 9 significant figures.
+- **No tolerance fixes this, so the fix is not a retune.** Measured over 800
+  corpus models plus targeted infeasible problems: the scaled surrogate is not
+  separable on the targeted cases; measuring it *unscaled* needs a tolerance
+  ≥ 1e-2 to fire at all, which introduces new false infeasibility on 3+ corpus
+  models while still losing 2 correct detections; and a scale-invariant
+  `‖Jᵀc‖ / ‖c‖²` is not separable either. A single absolute threshold on a
+  surrogate cannot separate these cases.
+- **What it does instead.** The surrogate is kept as a cheap pre-filter, and
+  before the verdict is issued POUNCE probes for a materially less-violating
+  point nearby — a few bound-clamped steps along `−∇θ`, comparing `θ` to `θ`.
+  That is scale-free by construction and needs no calibration. The two regimes
+  are far apart: near a genuine infeasible stationary point only ~0.07 % further
+  descent exists, whereas at HS13's false verdict one step takes `θ` from 0.51 to
+  zero.
 - **Why it matters more than a wrong number.** A false *unbounded* verdict is
   loud and a driver can retry it; a false *infeasible* silently prunes a
   branch-and-bound node that may contain the optimum.
-- Only rows are scaled and POUNCE applies no variable scaling, so the unscaled
-  gradient needs no unscaled Jacobian: `J_userᵀ c_user = J_scaledᵀ (c_scaled ⊘ dc²)`.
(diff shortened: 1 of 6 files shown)


jkitchin:
## Do not merge — this fix is wrong in approach. Converting to draft.

A third verification pass found that it trades the false positive for a **false negative**, on a case the earlier checks did not cover.

### The regression

Two contradictory equalities, `x³ + y³ == 1` and `x³ + y³ == 2`, from `x₀ = (1e4, 1e4)` — genuinely infeasible, steep Jacobian:

| | verdict | iters |
|---|---|---|
| before this PR | **Converged to a point of local infeasibility** (correct) | 23 |
| with this PR | Maximum Iterations Exceeded | 3000 |
| Ipopt | "Feasible point for square problem found" | 35 |

The detector stops firing, and the solve grinds to `max_iter` instead of returning a correct verdict in 23 iterations.

### Why the approach is wrong, not just the details

`infeas_stationarity_tol` is an **absolute** tolerance (default `1e-8`) on `‖Jᵀc‖ / max(1, ‖c‖)`. That is only meaningful because gradient-based scaling normalizes `J` to O(1) — the tolerance and the scaled measure were co-designed. Measuring the numerator unscaled removes exactly the normalization the tolerance depends on, so on a problem with a large Jacobian the measure can no longer get under `1e-8` at all, however stationary the iterate is.

So the earlier diagnosis was right about the *symptom* — the two halves of the test live in different spaces — but unscaling the stationarity half is the wrong way to reconcile them.

### What the real problem is

The measure is not **scale-invariant**, because of the constant in `max(1, ‖c‖)`. Under a row scale `dc`, the numerator carries `dc²` while the denominator is clamped at 1, so the whole thing collapses as `dc → 0` regardless of the iterate. That is what produced the HS13 false positive (`dc ≈ 3.3e-7`, measure `5e-14` at a point with violation 0.51 and `‖∇θ‖ = 1.4`).

A scale-invariant normalization fixes both directions at once. The cheapest candidate is

```
‖Jᵀc‖ / ‖c‖²
```

numerator and denominator both carrying `dc²`, so it is invariant, needs no matrix norm, and is well-defined precisely because the detector already requires the violation to be bounded away from zero. Sanity checks on the two cases: HS13's bad point gives ~2.0 (correctly not stationary), and a genuine infeasible-stationary point gives 0 (correctly detected).

That changes the meaning and calibration of a user-facing option, so the default `1e-8` would have to be re-derived and validated across the corpus — a real piece of work rather than a tweak, and not something to bolt on in the same pass that got it wrong once.

### What stands

The **bug report is solid** and unaffected: HS13 from `x₀ = (1e4, 1e4)` is feasible, POUNCE reports `Infeasible_Problem_Detected`, and the returned point is provably not stationary for the infeasibility (`‖∇θ‖ = 1.40`, no active bound, one step downhill reaches feasibility). The regression tests here are also fine — they pin the right property. Only the mechanism of the fix is wrong.

Also worth keeping from this pass: the unit tests pinning the unscaling algebra against hand-computed values, and the `debug_assert` on the downcast, both of which would be reusable if the derivation is ever needed elsewhere.

Reproduce: `adversary/runs/2026-07-22_nlp_infeasibility_still_detected.py` (compares pre-fix and post-fix binaries side by side).

jkitchin:
## Reworked — the fix is now a probe, not a threshold. Ready for review again.

The previous approach (measure the stationarity surrogate unscaled) was wrong and is removed. Calibration showed the problem is deeper than the space the surrogate is measured in.

### No threshold works, on any of the three candidate measures

Measured over 800 corpus models plus targeted infeasible problems:

| measure | result |
|---|---|
| scaled `‖Jᵀc‖ / max(1,‖c‖)` (shipped) | not separable on the targeted cases — this is the bug |
| unscaled | needs tol ≥ 1e-2 to fire at all; at 1e-2 that is **3 new false infeasibility verdicts and 2 lost detections**; at 1e-1, 27 new false positives |
| scale-invariant `‖Jᵀc‖ / ‖c‖²` | not separable even on the targeted set |

The absolute `1e-8` tolerance is only meaningful because gradient-based scaling normalizes `J` to O(1). Measuring the numerator unscaled removes exactly the normalization the tolerance depends on — which is why the first attempt turned a genuinely infeasible problem from "local infeasibility in 23 iterations" into "Maximum_Iterations_Exceeded at 3000".

### What it does now

`LocalInfeasibility` asserts that no local move reduces the violation. That is checkable, so POUNCE checks it: the surrogate stays as a cheap pre-filter, and before the verdict is issued a few bound-clamped steps along `−∇θ` probe for a materially less-violating point. Comparing `θ` to `θ` is scale-free by construction, so there is nothing to calibrate — which is the point, since calibration is what failed.

The material-descent margin is deliberately coarse (10%). An interior-point iterate converging toward a stationary point always has some infinitesimal descent left, so a tight margin vetoes forever and no infeasibility is ever declared — the first version used `1e-6` and did exactly that. The regimes are far apart: near a genuine infeasible stationary point ~0.07% descent remains, while at HS13's false verdict one step takes `θ` from 0.51 to **zero**.

Steps are clamped to the variable bounds, so descent existing only outside the box does not count — that direction would suppress a *correct* verdict.

### Results

| | |
|---|---|
| HS13 from `(1e4,1e4)` | `Infeasible_Problem_Detected` → **Optimal**, 0.98492872 in 29 iters, matching Ipopt to 9 s.f. |
| 4 genuinely infeasible problems | all still detected |
| corpus, 1284 models | infeasible verdicts **37 → 37**, 0 new false positives, 0 lost, otherwise bit-identical |

### Tests

Five, and critically including `genuinely_infeasible_problem_is_still_detected` — the test whose absence let two wrong fixes look correct. It passes on the original code, **fails on both abandoned attempts**, and passes here.

The abandoned unscaled measure, its helper and its unit tests are removed rather than left as dead code documenting a rejected approach.

Full workspace suite: 2054 tests, 0 failures. fmt and the CI clippy invocation clean.

PR #262, opened 2026-07-22
dev-notes: covariance and information design record

devin-griff:
**Rendered (formatted) view:** https://github.com/devin-griff/pounce/blob/covariance-information-roadmap/dev-notes/covariance-information-roadmap.md

A dev-notes design proposal for the pyomo-pounce post-solve second-order surface, targeting v0.10. Companion to #255 (the active-set sensitivity roadmap): that note extends `estimate()`, this one extends `covariance()`.

## What it proposes

All additive to the v0.9 `covariance()` surface, nothing changes an existing signature:

- **`information()`**, the un-inverted sibling of `covariance()`: the reduced Hessian formed directly from the held factor, avoiding the invert-then-reinvert round trip, in the same natural-units convention as the core `solve_with_sens`.
- **`wrt=` block selection** on `covariance()` and `information()`: reduce onto any free-variable block post-solve off one retained factor, so a single solve serves multiple blocks. `declare_fitted` becomes the default block when `wrt=` is omitted, preserving v0.9 behavior.
- **`retain_kkt()`**, a factor-retention switch for the declaration-free case.

## Motivation

The moving-horizon-estimation information-form arrival cost needs the Lagrangian reduced Hessian marginalized onto the arrival state, plus the parameter covariance from the same solve. The note works through the marginal-versus-conditional semantics and the one-solve-two-blocks flow.

It also acknowledges that pounce already exposes the reduced Hessian in its core (`solve_with_sens`), convex-QP (`QpSensitivity`), and `curve_fit` interfaces; this brings that to the pyomo-pounce interface.

Docs-only (one dev note). Opening for design discussion before any implementation.


devin-griff:
Two new items at the front. The previous version delegated bound handling to `covariance()`'s convention in a single line. That convention classifies on slack alone, and embeds a pinned parameter as a zero row, which is right for a variance and wrong for an information matrix, where the same zeros claim the data says nothing.

Item 0 is an activity classifier keyed on barrier curvature over objective curvature, written as pseudocode, and is Rust core work rather than pyomo surface. Item 1 is the dispositions table.

Items 1 to 3 are now 2 to 4, largely intact.

The rest has been broadly revised for clarity, with a new Preconditions section and constraint rows for #362.

224 to 354 lines.

devin-griff:
Refreshed. A read-only pass over the solver internals ahead of implementing item 0 found the converged state already retains everything the classifier needs — bound multipliers, slacks, `curr_mu`, `curr_sigma_x`, and the exact Lagrangian Hessian — so `H` is never recovered by subtraction and the earlier claim that the pinned block loses precision as `μ` tightens is retired: `S` forms densely from the exact Hessian at full precision. The `unidentified` floor is pinned, and a review pass tightened items 0 and 1 (warning scope, row handling in the guard, notation stated forward).

Implementation of item 0 starts now against this text; the PR will reference back here, and anything implementation teaches lands as one more refresh.

devin-griff:
@jkitchin The roadmap's items are all implemented (items 0 through 3 merged; item 4 is #468, in review), so this is reframed as a record of the final design, now covering the whole covariance and information story including the original `covariance()` design that predated the roadmap. Marked ready: it can either be merged to main as that record, or closed if it is redundant with the docs and the CHANGELOG. Either is fine with me. 

new rendered view: https://github.com/devin-griff/pounce/blob/covariance-information-roadmap/dev-notes/covariance-information-design.md

jkitchin:
Merging it as the design record — it is not redundant with the docs. Seven of its ten sections cover constructions `docs/src/sensitivity.md` deliberately does not carry: the full-x / var-x split, tangent recovery, the dispositions table, the rank and singularity guards, Gauss-Newton, the reporting surface, and validation. The two-spaces section in particular is the kind of thing whose absence costs someone a day.

I pushed one correction to the branch first (`0a347468`). The sigma bullets said the estimation routes divide by `n - n_fit`, and the `retain_kkt()` table says nothing need be declared; read together they implied that retain-only plus declared residuals estimates `SSR / n`. That is exactly what #468 fixed a few hours ago — both estimation routes now raise when nothing is declared fitted, because dividing by `n` biases every variance low by `n/(n-p)`. The note now states that where sigma is defined.

Also retitled from "v0.10 roadmap" to match the content, since the PR title becomes the merge commit subject.

The band terminology you already propagated is correct as written — "the confidence band on the fitted trajectory (observation noise added on top makes it a prediction band)".

Thanks for the roadmap and the four implementation PRs behind it.

---

PR #287, opened 2026-07-22
Emit AMPL-convention duals in .sol; fix GAMS max-model marginals (#271, #272)

### LINKED ISSUE #271: [Adversary] SOLVER_BUG: constraint duals globally sign-flipped vs AMPL/Pyomo convention

## Family / solver

Cross-cutting — multiplier sign convention. Affects `.sol` writer, `pyomo-pounce`, `pounce.minimize`, `solve_qp`, JSON solve reports.

## Problem

Wyndor Glass Co. LP (Hillier & Lieberman, *Introduction to Operations Research*, Sec. 3.1; shadow prices Sec. 4.7), written as a minimization:

- optimum `z* = -36`
- analytic shadow prices `dz/db = (0, -1.5, -1.0)`

## Expected vs actual

| Source | Duals |
|---|---|
| analytic `dz/db` | `(0, -1.5, -1.0)` |
| pounce-only finite-difference re-solve | `(0, -1.5, -1.0)` |
| Ipopt 3.14.19 | `(0, -1.5, -1.0)` |
| glpk, cbc, CONOPT | `(0, -1.5, -1.0)` |
| **pounce** | **`(0, +1.5, +1.0)`** |

Magnitudes are always correct. Primal solutions and objectives are unaffected. The flip is uniform across equalities and inequalities and is sign-symmetric (a `<=` row reports `+1.5` where `-1.5` is right; a `>=` row reports `-2` where `+2` is right).

## Root cause / scope

This is **not** a `.sol`-writer defect. Solving the identical `.nl` through the Python library gives `info["mult_g"]` bit-identical to the `.sol` duals, so the writer applies no transformation — the flip is pounce's internal Lagrangian convention (`lambda_pounce = -lambda_ipopt`).

Affected surfaces:

- `.sol` writer, three sites: `crates/pounce-cli/src/main.rs:1379` (NLP), `:1750` (QP), `:1923` (conic, via `recover_socp_duals`)
- `crates/pounce-convex/src/qp_extract.rs:221` `recover_duals` and `:512` `recover_socp_duals`
- `Problem.solve` `info["mult_g"]`; `pounce.minimize` `res.mult_g`
- JSON solve report `solution.lambda`
- `pyomo-pounce` `model.dual`

Not affected: primal `x` and objective; bound multipliers `mult_x_L` / `mult_x_U` (these match Ipopt); `sens_sol_state_1` (primal-only payload, verified to 1.1e-16 rather than assumed).

## Two things that make the fix non-trivial

**1. GAMS currently compensates — do not fix one without the other.** Both links apply an explicit `pi = -mult_g` negation:

- `gams/gams_pounce.c:1096-1099`
- `python/pounce/gams/link.py:361-365`

Verified live against CONOPT on both `=l=` and `=e=` models: GAMS marginals are currently **correct** for `minimizing` models. Correcting the internal sign without removing these two negations will break the one surface that works. (GAMS is separately wrong for `maximizing` models — filed as its own issue.)

**2. `pounce verify` is structurally blind to this.** `crates/pounce-cli/src/verify.rs:393-408` evaluates stationarity for both `+lambda` and `-lambda` and keeps the better residual. It returns `VERIFIED` (rc=0) on pounce's `.sol`, on Ipopt's `.sol`, and on a hand-negated `.sol`, with a bit-identical residual. It therefore cannot serve as the regression guard — a dedicated sign test is required.

Relatedly, no existing test compares a `.sol` dual against Ipopt's, which is why this survived.

## Cross-check

- Analytic shadow prices (primary)
- Finite-difference re-solve **using pounce itself**, so the conclusion does not depend on any other solver
- Ipopt 3.14.19, glpk, cbc via the identical Pyomo model
- CONOPT via GAMS 53.2.0
- Confirmed at the raw `.sol` level: pounce `+2.0` vs Ipopt `-1.99999997` on the same `.nl`
- Reproduced on LP, QP and true NLP; `<=`, `>=` and `==`; under both `minimize` and `maximize`

Note that the **Python API's `solve_qp` duals are self-consistent** under pounce's own documented convention `Px + c + A'y + G'z - z_lb + z_ub = 0` (`crates/pounce-convex/src/crossover.rs:29-30`) and match cvxpy/CLARABEL sign-for-sign — a separate 8-case battery confirmed stationarity, dual feasibility, complementary slackness and zero duality gap to ~1e-8. The question is which convention the AMPL/Pyomo-facing surfaces should expose.

## Impact

Silent corruption of sensitivity analysis / shadow prices for every Pyomo and AMPL user. Objectives and primal solutions are correct, so this fails quietly.

## Reproduce

```
source .venv-qa/bin/activate
python adversary/runs/2026-07-22_pyomo_dual_signs.py
python adversary/runs/2026-07-22_duals_blast_radius.py
```

Reports: `adversary/runs/2026-07-22_pyomo_dual_signs.org`, `adversary/runs/2026-07-22_duals_blast_radius.org`

### LINKED ISSUE #272: [Adversary] SOLVER_BUG: GAMS link equation marginals sign-inverted for maximizing models (both links)

## Family / solver

GAMS link — both the pip/pure-Python link (`python/pounce/gams/link.py`) and the native C link (`gams/gams_pounce.c`).

## Problem

Both GAMS links negate pounce's `lambda` **unconditionally** to form GAMS's `pi`:

- `python/pounce/gams/link.py:361-365`
- `gams/gams_pounce.c:1096-1103`

That negation is correct only for `minimizing` models. The same code paths do apply the `obj_sign` flip to the objective (`link.py:353-354`, `gams_pounce.c:1081`) and to the variable marginals (`link.py:370-377`, `gams_pounce.c:1105-1118`) — but **not** to the constraint duals. So on a `maximizing` model every equation `.M` comes back with the wrong sign.

Test LP with analytic shadow prices, solved in both senses (`obj = 25.5` either way).

## Expected vs actual

Verified live against GAMS 53.2.0 with its default LP solver (CPLEX):

| Model sense | Marginal | GAMS/CPLEX (correct) | pounce |
|---|---|---|---|
| `maximizing` | `r1.M` | `+2.25` | **`-2.25`** |
| `maximizing` | `r2.M` | `+0.25` | **`-0.25`** |
| `maximizing` | `e1.M` (equality control) | `+2.000` | **`-2.000`** |
| `minimizing` | `r1.M` | `+2.25` | `+2.25` ✓ |
| `minimizing` | `r2.M` | `+0.25` | `+0.25` ✓ |

Correct in both senses: primal values, objective (`25.5`), variable marginals / reduced costs (`x3.M = -1.5`), and the status map (`Solve_Succeeded` → solver status 1 Normal / model status 2 Locally Optimal).

## Cross-check

- Live GAMS 53.2.0 install (present and licensed on this machine), default LP solver as oracle
- Analytic `dObj/d(rhs)` shadow prices
- A license-free fake-`GmoView` harness reproducing the defect offline in <0.01 s/solve, written under `adversary/` without modifying `python/tests/test_gams_link.py`

**The two links do not diverge from each other** — the arithmetic is byte-for-byte equivalent, so install method makes no difference to users. (The C link could not be exercised live because it is shadowed by the local `gamsconfig.yaml` registration of the pip link; forcing it would have required writing outside `adversary/`.)

## Secondary consequence

The C link's §7.4(a) warm-start reconstruction (`gams_pounce.c:973-1050`) re-reads `gmoGetEquM` assuming its own convention. On a `maximizing` model whose marginals came from another GAMS solver, the reconstructed active-set working set is inverted.

## Relationship to the internal dual-sign issue

This is entangled with the global dual-sign convention issue filed separately. Those two `pi = -mult_g` negations are the compensation that makes GAMS correct for `minimizing` models today. Any fix must treat the internal convention and these two call sites as a matched pair, and add the missing `obj_sign` factor here.

## Reproduce

```
source .venv-qa/bin/activate
python adversary/runs/2026-07-22_gams_dual_sign.py
```

Live GAMS evidence (`.gms` / `.lst`): `adversary/runs/gams_scratch/`
Report: `adversary/runs/2026-07-22_gams_dual_sign.org`

Note: `gamsapi==53.2.0` was installed into `.venv-qa` as oracle tooling.

jkitchin:
Fixes #271. Fixes #272.

## The defect

Constraint duals reached every AMPL-facing surface negated relative to the sign Ipopt, glpk, CBC and CONOPT report. Objectives and primal solutions were never affected — which is exactly why it went unnoticed.

Two different quantities were being conflated:

| | quantity | meaning |
|---|---|---|
| `.sol` dual block, Pyomo `model.dual` | **marginal** | `d obj / d b` — the shadow price |
| `mult_g`, JSON `solution.lambda` | **Lagrange multiplier** | the `λ` of `L = f + λ'g − z_L(x−x_L) + z_U(x−x_U)` |

They differ by a sign: `marginal = −λ`. The `.sol` writer emitted `λ` without converting it.

## #271 — `.sol` writer

`format_sol` now performs the conversion. `SolutionFile.lambda` is renamed to `mult_g` and documented as taking pounce's *internal* multipliers.

The rename is deliberate. It makes the compiler enumerate every construction site rather than trusting a grep, and it stops the field name from implying the file carries a Lagrange multiplier when it carries a shadow price.

**The Python API and JSON report are unchanged.** `mult_g` keeps the Lagrange-multiplier convention — it matches cyipopt, satisfies the stationarity identity directly, and `lambda` is the accurate name for it. Both conventions are now documented side by side in `docs/src/cli.md`, since users comparing a `mult_g` against a `model.dual` will legitimately see opposite signs.

*(If you'd rather the JSON report also flip for cross-surface consistency, say so — I kept it as-is because the field name is semantically correct and the benchmark harness reads that file. It's a one-line change.)*

## #272 — GAMS marginals on `maximizing` models

Both links negated `λ` unconditionally to form `pi`. That is correct only for `minimizing` models: pounce always minimizes, so a `maximizing` model is solved as `min(−f)` and its multipliers refer to the negated objective. The `obj_sign` factor already applied to the objective value (`gams_pounce.c:1081`) and to the variable marginals was simply missing from the multiplier conversion.

Both links now use `pi = −obj_sign · λ`, extracted into `gams_pi()` in `link.py` so it is unit-testable without a GAMS install, and mirrored in `gams_pounce.c`.

## pyomo-pounce coupling

`Gradient` returned `d(λ)/d(param)` from `parametric_step_full`'s y_c block, while the user compares against `model.dual`. Once duals are converted, that block needs the same conversion — added as `_convention_sign`, applied to `Constraint` targets only.

I did not anticipate this; `test_sens.py`'s two sign-convention tests caught it. They're good tests.

## Why this survived to a release candidate

Worth recording, because the fix alone doesn't address it:

- **The benchmark suite compares objectives, statuses, iteration counts and wall times — never duals.** A defect that leaves primals exact is invisible to it by construction.

- **`qp_dispatch_end_to_end.rs` asserted the wrong sign as expected behavior.** It pinned the equality dual of `min x₀²+x₁² s.t. x₀+x₁=2` at `−2`, when the analytic `d obj / d b = b = +2`. Its sibling test checked only that the QP and NLP paths *agreed with each other* — which a uniform sign error satisfies trivially. Both now assert `+2`, and the comments explain why cross-path agreement can't detect a shared convention defect.

- **`pounce verify` evaluates stationarity for both `+λ` and `−λ` and keeps the better residual**, so it certifies either sign. It could not have caught this and still cannot. The `.sol` tests are the guard; `verify`'s leniency is left alone here since tightening it is a separate change with its own blast radius.

## Validation

- Pyomo duals now match Ipopt/glpk and an analytic finite-difference on all 7 cases of the reproducer (LP/QP/NLP × `<=`,`>=`,`==` × min/max)
- Raw `.sol` duals match Ipopt to 1e-9 (`[-0, -1.5, -1.0]` both)
- 142 Rust suites, 685 Python tests, 101 pyomo-pounce tests pass
- The native C link compiles against GAMS 53.2.0 headers

New regression tests: 4 in `test_gams_link.py` covering both senses and an analytic shadow price; the two corrected `.sol` assertions; a corrected writer↔parser round-trip that documents why the dual block is *not* an identity round-trip.

## Compatibility

**This flips the sign of duals read through Pyomo and AMPL.** Anyone who compensated for the old behavior downstream should remove that workaround. Flagged in the changelog.

The GAMS `minimizing` path is unchanged — it was already correct, via the compensating negation. Only `maximizing` models change there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,30 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — constraint dual sign convention (#271, #272)
+
+- **`.sol` / Pyomo `model.dual`** carried pounce's internal Lagrange
+  multiplier instead of the AMPL *marginal* `d obj / d b`, so every dual came
+  back negated relative to Ipopt, glpk, CBC and CONOPT (#271). Objectives and
+  primal solutions were never affected. The `.sol` writer now performs the
+  conversion. **This flips the sign of duals read through Pyomo and AMPL** —
+  if you compensated for the old behavior in downstream code, remove that
+  workaround.
+- The Python API's `mult_g` and the JSON report's `solution.lambda` are
+  unchanged: they keep the Lagrange-multiplier convention, which matches
+  cyipopt. `marginal = −λ`; both conventions are now documented side by side
+  in [Running Solves](docs/src/cli.md).
+- **GAMS equation marginals on `maximizing` models** were inverted in both the
+  pip link and the native C link (#272): the `obj_sign` factor already applied
+  to the objective value and the variable marginals was missing from the
+  multiplier conversion, which is now `pi = −obj_sign · λ`. `minimizing`
+  models, objective values, variable marginals and status mapping were never
+  affected, and the two links behaved identically, so install method made no
+  difference.
+- `pyomo_pounce.gradient()` on a `Constraint` target returned
+  `d(λ)/d(param)`, which no longer matched `model.dual` once duals were
+  converted; it now reports `d(dual)/d(param)`.
+
 ### Fixed — ambiguous 2-parameter tuple bounds and NaN bounds (#265)
 
 - `curve_fit` / `curve_fit_minima` / `curve_fit_streaming`: at `n == 2`, a
--- a/crates/pounce-cli/src/main.rs
+++ b/crates/pounce-cli/src/main.rs
@@ -1376,7 +1376,7 @@ pub fn main() -> ExitCode {
         let payload = nl_writer::SolutionFile {
             message: &message,
             x: &x,
-            lambda: &lambda,
+            mult_g: &lambda,
             solve_result_num: status_to_solve_result_num(status),
             suffixes: &sol_suffixes,
         };
@@ -1747,7 +1747,7 @@ fn run_convex_qp(
         let payload = nl_writer::SolutionFile {
             message: &format!("POUNCE {} IPM (pounce-convex): {msg}", class.name()),
             x: &sol.x,
-            lambda: &lambda,
+            mult_g: &lambda,
             solve_result_num: srn,
             suffixes: &[],
         };
@@ -1920,7 +1920,7 @@ fn run_convex_socp(
         let payload = nl_writer::SolutionFile {
             message: &format!("POUNCE {} conic IPM (pounce-convex): {msg}", class.name()),
             x: &sol.x,
-            lambda: &lambda,
+            mult_g: &lambda,
             solve_result_num: srn,
             suffixes: &[],
         };
--- a/crates/pounce-cli/src/minima/mod.rs
+++ b/crates/pounce-cli/src/minima/mod.rs
@@ -898,7 +898,7 @@ fn write_sol_files(sol_path: &Path, minima: &[Minimum], m: usize) {
         let payload = crate::nl_writer::SolutionFile {
             message: &message,
             x: &mn.x,
-            lambda,
+            mult_g: lambda,
             solve_result_num: status_to_solve_result_num(ApplicationReturnStatus::SolveSucceeded),
             suffixes: &[],
         };
--- a/crates/pounce-cli/src/nl_writer.rs
+++ b/crates/pounce-cli/src/nl_writer.rs
@@ -101,8 +101,19 @@ pub struct SolutionFile<'a> {
     pub message: &'a str,
     /// Primal variable values, length `n`.
     pub x: &'a [Number],
-    /// Constraint dual values, length `m`.
-    pub lambda: &'a [Number],
+    /// Constraint multipliers in **pounce's internal (cyipopt)
+    /// convention**, length `m`: the `lambda` of
+    /// `L = f + lambda' g - z_L (x - x_L) + z_U (x - x_U)`, so
+    /// `d obj / d b = -lambda`.
+    ///
+    /// [`format_sol`] negates these on the way out, because the AMPL
+    /// `.sol` dual block carries *marginal values* (`d obj / d b`),
+    /// not Lagrange multipliers. Pass the internal multipliers here
+    /// and let the writer do the translation — do not pre-negate at
+    /// the call site.
+    ///
+    /// See [Gay, "Hooking Your Solver to AMPL" §5](https://ampl.com/REFS/hooking2.pdf).
+    pub mult_g: &'a [Number],
     /// AMPL solver return code. Convention: 0 = solved, 100..199 =
     /// "solved with warning", 200..299 = "infeasible", 300..399 =
     /// "unbounded", 400..499 = "limit reached", 500..599 = "failure".
@@ -132,7 +143,7 @@ pub fn format_sol(payload: &SolutionFile<'_>) -> String {
     // `.nl`. We write every dual and primal, so the pairs collapse to
     // (m, m) and (n, n). Emitting only `m` and `n` (the two-integer
     // short form) makes AMPL's and Pyomo's `.sol` readers fail.
-    let m = payload.lambda.len();
+    let m = payload.mult_g.len();
     let n = payload.x.len();
     let _ = writeln!(out, "{m}");
     let _ = writeln!(out, "{m}");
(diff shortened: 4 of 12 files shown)


jkitchin:
Follow-ups filed:
- #294 — hardening: neither the benchmark suite (objectives only) nor `pounce verify` (accepts either sign) can catch a uniform dual-sign flip; adds an external/analytic regression guard so this can't recur silently.
- #296 — parity: the `.sol` writer still omits bound multipliers (`ipopt_zL_out`/`zU_out`), so reduced costs are unavailable to AMPL/Pyomo/GAMS users.

PR #290, opened 2026-07-22
Scale the recession residual by ||P|| in the unboundedness certificate (#273)

### LINKED ISSUE #273: [Adversary] SOLVER_BUG: false 'unbounded' on a provably bounded QP via auto-routing (|P| < 1e-10)

## Family / solver

`nlp` / auto-routing — default `solver_selection="auto"` on `.nl` input.

## Problem

    min  -x + x^2/(2M)    s.t.  x >= 0

Strictly convex and coercive, with unique global minimum `x* = M`, `f* = -M/2`. Provably bounded for every finite `M > 0`. Tested at `M = 1e12`.

## Expected vs actual

| Solver | Objective | Status |
|---|---|---|
| analytic | `-5.0000e+11` | optimal at `x* = 1e12` |
| Ipopt 3.14.19 | `-5.000000e+11` | optimal |
| pounce, `solver_selection="nlp"` (same binary, same `.nl`) | `-5.000000e+11` | optimal |
| scipy `minimize_scalar` | `-5.0000e+11` | optimal |
| cvxpy Clarabel / SCS / OSQP | `-5.0000e+11` | optimal |
| **pounce, default auto route** | **`-3.990950e+02`** | **`objno 0 300` — "Problem is unbounded (dual infeasible)"** |

It fires after **2 iterations at `x = 399`** — twelve orders of magnitude short of the optimum — on a problem that is bounded below.

## Localization

- Threshold is a hard cut at `|P| < 1e-10`, and it is **not scale-invariant**: multiplying the objective by `1e3` flips the verdict back to correct.
- The QP core itself matches Clarabel/SCS/OSQP on this input, so the defect is at the **router/system level** — auto-routing to a solver whose unboundedness certificate is unreliable in this regime, with no fallback — while pounce's own NLP solver in the same binary solves it exactly.
- Auto disagrees with forced `solver_selection="nlp"` on the same `.nl`, so this is also a routing-transparency violation.

## Cross-check

Part of a 20-instance battery with analytic ground truth established per instance and re-verified with `scipy.minimize_scalar`:

- **Group A** — 3 genuine recession rays (linear, nonlinear, quadratic): 3/3 correctly detected.
- **Group B** — 3 slow rays (`-x^0.1`, `-log log x`, `-x/(1+log x)`): 1/3 detected. The 2 misses are shared *identically* with Ipopt (gradient → 0), so not a pounce finding.
- **Group C** — 14 provably bounded far-optimum controls: **2 false `unbounded`**.

No regression from #250/#253 in the false-negative direction; this is the false-positive direction, which those PRs did not cover.

## Reproduce

```
source .venv-qa/bin/activate
python adversary/runs/2026-07-22_nlp_unbounded_lookalike_battery.py
```

Report: `adversary/runs/2026-07-22_nlp_unbounded_lookalike_battery.org`

jkitchin:
Fixes #273.

## The defect

The convex IPM's dual-infeasibility (unboundedness) certificate tested whether the candidate recession direction `d` satisfies

```rust
inf_norm(&pd) <= rtol * x_norm     // ‖Pd‖ ≤ rtol·‖d‖
```

Because `‖Pd‖` is itself proportional to `‖P‖·‖d‖`, the `‖d‖` **cancels** and the test collapses to `‖P‖ ≤ rtol` — a bare comparison of the Hessian's magnitude against the absolute constant `FARKAS_RESID_TOL = 1e-10`, with no reference to `d` at all.

So **any** strictly convex QP whose Hessian is smaller than that constant read as unbounded, regardless of having a finite minimizer:

`min -x + x²/(2M) s.t. x ≥ 0` has the unique minimum `x* = M`, `f* = -M/2` for every finite `M`.

| M | P = 1/M | before | after |
|---|---|---|---|
| 1e8 | 1e-8 | `-5.0e7` ✓ | `-5.0e7` ✓ |
| 1e10 | 1e-10 | **unbounded** (obj `-399`, 2 iters) | `-5.0e9` ✓ |
| 1e12 | 1e-12 | **unbounded** (obj `-399`, 2 iters) | `-5.0e11` ✓ |
| 1e14 | 1e-14 | **unbounded** | `-5.0e13` ✓ |

The threshold sits exactly where `P` crosses `1e-10`, as the cancellation predicts. Ipopt, pounce's own NLP path, and Clarabel/SCS/OSQP all solve these exactly.

Note this is **not** a routing bug — the problem is correctly classified as a convex QP. The false certificate comes from the IPM itself.

## The fix

`d` is a recession direction of the quadratic only if `Pd = 0`, i.e. `d ∈ null(P)`. Expressing that requires a *relative* residual, so the bound is scaled by `‖P‖`:

```rust
inf_norm(&pd) <= rtol * x_norm * p_scale
```

Both sides now carry the same `‖P‖·‖d‖` units and nothing cancels.

Unaffected in both limiting directions (each pinned by a new test):

- **LP** (`P` empty): `p_scale = 0` and `Pd` is exactly `0`, so the test reads `0 ≤ 0` — genuine LP unboundedness still certified.
- **Genuinely singular `P`** with `d ∈ null(P)`: `Pd` is zero to round-off, `~ε·‖P‖·‖d‖`, comfortably inside `rtol·‖P‖·‖d‖`.

## Known remaining limitation (not introduced here)

`p_scale` is a single global magnitude, so a **mixed-scale** Hessian can still misreport. `P = diag(1e6, 1e-12)` with descent along the tiny block returns `dual_infeasible` even though the problem is bounded (`x₁* = 1e12`).

I verified this case explicitly rather than assuming: it fails identically **before and after** this change, so the fix is a strict improvement and not a regression. Distinguishing "bounded with an optimum at 1e12" from "unbounded" when the curvature in the descent direction is 18 orders below the largest is genuinely ambiguous under any relative criterion — it's a tolerance question, not a bug with a clean answer. Happy to file it as a separate issue if you'd like it tracked.

## Regression evidence

This changes a termination certificate in the convex IPM, so I built before/after binaries and ran **every `.nl` in `benchmarks/` (44 problems)** through both: **all 44 return a byte-identical `solve_result_num`.**

Also: 142 Rust suites pass, including the full `pounce-convex` suite.

## Tests

Three in `ipm::detect_infeasibility_tests`:

- `tiny_hessian_is_not_a_recession_direction` — `P ∈ {1e-10, 1e-12, 1e-16}` must all return `None` (bounded)
- `singular_hessian_nullspace_direction_is_still_dual_infeasible` — no false negative
- `empty_hessian_lp_unboundedness_unaffected` — the `p_scale = 0` edge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,24 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — strictly convex QP falsely reported unbounded (#273)
+
+- The convex IPM's dual-infeasibility (unboundedness) certificate tested
+  `‖Pd‖ ≤ rtol·‖d‖` for the candidate recession direction `d`. Because `‖Pd‖`
+  is itself proportional to `‖P‖·‖d‖`, the `‖d‖` cancelled and the test
+  collapsed to `‖P‖ ≤ rtol` — a bare comparison of the Hessian's magnitude
+  against the absolute constant `1e-10`, with no reference to `d` at all.
+- Consequence: **any** strictly convex QP with a small enough Hessian was
+  certified unbounded despite having a finite minimizer. `min -x + x²/(2M)
+  s.t. x >= 0` (unique minimum `x* = M`) was reported unbounded for every
+  `M >= 1e10`, terminating after 2 iterations twelve orders of magnitude short
+  of the optimum, on a problem Ipopt and pounce's own NLP path both solve
+  exactly.
+- The residual bound is now scaled by `‖P‖`, restoring the intended meaning:
+  a relative test for `d ∈ null(P)`. LPs (`P` empty, `Pd` exactly zero) and
+  genuinely singular Hessians with `d` in the nullspace are unaffected, so
+  real unboundedness is still detected.
+
 ### Fixed — constraint dual sign convention (#271, #272)
 
 - **`.sol` / Pyomo `model.dual`** carried pounce's internal Lagrange
--- a/crates/pounce-convex/src/ipm.rs
+++ b/crates/pounce-convex/src/ipm.rs
@@ -2537,8 +2537,32 @@ pub(crate) fn detect_infeasibility_with(
         // SOC/PSD ⇒ true cone membership). Checked, not componentwise, so a
         // direction that merely has `Gd ≤ 0` but `−Gd ∉ K` is rejected.
         let gd_ok = primal_recession_ok(&gd, ctol * x_norm);
+        // `d` is a recession direction of the *quadratic* only if `Pd = 0`,
+        // i.e. `d ∈ null(P)`. Testing `‖Pd‖ ≤ rtol·‖d‖` cannot express that:
+        // `‖Pd‖` is itself proportional to `‖P‖·‖d‖`, so `‖d‖` cancels and the
+        // test collapses to `‖P‖ ≤ rtol` — a bare comparison of the Hessian's
+        // magnitude against an absolute constant, with no reference to `d` at
+        // all. Every strictly convex QP whose Hessian happens to be smaller
+        // than `rtol` then reads as unbounded regardless of having a finite
+        // minimizer: `min -x + x²/(2M)  s.t.  x ≥ 0` (unique minimum at
+        // `x* = M`) was reported unbounded for `M ≥ 1e10`, i.e. exactly when
+        // `P = 1/M` crossed `FARKAS_RESID_TOL`. See gh #273.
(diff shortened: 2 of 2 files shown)


jkitchin:
Follow-up filed: #293 — the `‖P‖`-scaling here is a single global scalar, so a **mixed-scale** Hessian (`diag(1e6, 1e-12)`) still false-reports `dual_infeasible`, and a **uniformly tiny** 2-var Hessian now hits `iteration_limit` instead of converging. Both have repros in the issue.

PR #291, opened 2026-07-22
Reject non-finite inputs that silently reported success (#275)

### LINKED ISSUE #275: [Adversary] SOLVER_BUG: infinite bounds dropped sign-agnostically, and NaN x0 reports Solve_Succeeded

## Family / solver

`solve_qp`, `solve_qp_batch`, `pounce.minimize` — input validation.

## Problem

Two independent silent-wrong-answer paths, found by a 47-case sweep of non-finite and malformed numeric input (each case in an isolated subprocess). Base problem: a strictly convex 4-var QP with `obj* = -3.126865671642e+00`, cross-checked against cvxpy 1.9.1/CLARABEL.

---

### Finding 1 — `+/-Inf` bounds are dropped sign-agnostically

`solve_qp` / `solve_qp_batch` treat any infinite bound as "absent" without checking the sign. So `lb = +inf` or `ub = -inf` — which is an **infeasible** constraint — is silently discarded, and the solver returns `status="optimal"` at a point violating the stated bound by `+inf`.

pounce contradicts itself two ways here:

- the finite analogue `lb = 1e12` is correctly `primal_infeasible`
- finite inverted bounds `lb=1 > ub=0` are correctly `primal_infeasible`
- `pounce.minimize` raises a clear `ValueError` on the identical bound

**Suggested fix:** sign-aware absence test (`lb` absent only at `-inf`, `ub` only at `+inf`), plus the `lb <= ub` check `minimize` already performs.

---

### Finding 2 — NaN `x0` yields `Solve_Succeeded` with `fun=nan`

`pounce.minimize` with NaN in `x0` returns:

    success=True, status=0, message="Solve_Succeeded", fun=nan, x=[nan...], nit=0

NaN comparisons against the convergence tolerances all evaluate false, so convergence is declared at **iteration zero**.

Reproduces with analytic `jac`, with FD gradients, with equality constraints, and on forced `solver_selection="nlp"`. Masked when `bounds` are supplied (clipping repairs `x0`).

Same hole, two more shapes:

- an objective returning NaN reports `Solve_Succeeded`
- a **gradient** returning NaN reports `Solve_Succeeded` with a *finite* `fun` (the value at `x0`) — the caller gets no signal at all

**Suggested fix:** finiteness check on `x0` at the API boundary, plus a NaN guard in the convergence test.

---

## What passed

- Legitimate `+/-Inf` bounds are handled **correctly**: rel err ≤ 2.8e-16 vs cvxpy, and `0.0` relative difference vs a finite `1e12`-bound reference
- NaN/Inf in `P`, `c`, `A`, `b`, `G`, `h` all raise clear `ValueError`s
- 7 shape-mismatch cases all raise
- float32 / int64 / Fortran-order / transposed view / non-contiguous slice / reversed stride / read-only arrays: all bit-identical to baseline
- `x0 = +/-Inf` returns `success=False` (no false claim); `x0 = 1e18` recovers the optimum
- No crashes, panics or hangs in 47 cases

Cosmetic: NaN on the `P` diagonal escapes the named-input guard and surfaces as `LinAlgError: Eigenvalues did not converge` instead of the usual "`P` contains NaN or Inf" message.

## Reproduce

```
source .venv-qa/bin/activate
python adversary/runs/2026-07-22_api_nonfinite_input.py
```

Report: `adversary/runs/2026-07-22_api_nonfinite_input.org`

jkitchin:
Fixes #275.

Two input paths returned a confident wrong answer on malformed input.

## 1. Sign-agnostic infinite bounds

The solver's bound-presence test is `lb > -BOUND_INF` / `ub < BOUND_INF` — it doesn't look at the sign. So `lb = +inf` and `ub = -inf`, which are constraints *no finite value can meet*, were dropped as if the bound were absent, and `solve_qp` returned `status="optimal"` at a point violating the stated bound by an infinite margin.

pounce contradicted itself two ways:

| case | `solve_qp` before | `minimize` before |
|---|---|---|
| `lb=1 > ub=0` (finite) | `primal_infeasible` ✓ | `ValueError` ✓ |
| `lb=+inf, ub=5` | **`optimal`** ✗ | `ValueError` ✓ |
| `lb=ub=+inf` | **`optimal`** ✗ | **`success=True`** ✗ |
| `lb=ub=-inf` | **`optimal`** ✗ | **`success=True`** ✗ |
| `lb=-5, ub=-inf` | **`optimal`** ✗ | `ValueError` ✓ |

`minimize`'s `lb > ub` test catches the *mixed* spellings but not `lb == ub == ±inf`, where the comparison is False.

Both surfaces now reject any bound admitting no finite point, naming the index. **Deliberately unchanged:** `±inf` on the *absent* side stays legal (it's the one-sided encoding), and a finite reversed box keeps returning `primal_infeasible` rather than raising — only the degenerate spellings that previously produced a wrong `"optimal"` are rejected. `solve_qp_batch` and `solve_qp_multi_rhs` inherit the guard.

## 2. Non-finite `x0`

```
minimize(f, np.array([np.nan]), jac=g)
  -> success=True, status=0, message="Solve_Succeeded", fun=nan, nit=0
```

Every convergence test compares against a tolerance, and comparisons against NaN are False — including the ones that would have *rejected* the iterate. So the loop fell straight through to "converged" at iteration zero.

Reproduced with analytic `jac`, with finite-difference gradients, with equality constraints, and on forced `solver_selection="nlp"`; masked only when `bounds` were supplied, because clipping repaired `x0`. `minimize` now requires a finite `x0`.

## Not fixed here — filed separately

A NaN **gradient** still reports `Solve_Succeeded`, and with a *finite* `fun` (the value at `x0`), so the caller gets no signal at all.

I traced it, and the root cause is different from either fix above: the max-reduction behind the dual-infeasibility norm **drops NaN**. `compound_vector.rs:235` uses `if v > m`, which is False for NaN, and the dense path uses BLAS `iamax`. Confirmed end-to-end — with a NaN gradient the solver prints:

```
Dual infeasibility......:   0.0000000000000000e+00
Overall NLP error.......:   0.0000000000000000e+00
EXIT: Optimal Solution Found.
```

So the existing `!nlp_err.is_finite()` guard never fires, because the value it checks was laundered to `0.0` upstream. Correcting the reduction touches every consumer of `amax` — step sizes, convergence, line search — and is not a change I'd make during release prep without a dedicated regression pass. Filed as its own issue with the root cause recorded.

## Tests

Seven new tests. Beyond the rejections, they pin the things that must **not** change: legitimate one-sided `±inf` bounds still solve, finite reversed boxes still report `primal_infeasible` rather than raising, and the batch entry points inherit the guard.

695 Python tests and 101 pyomo-pounce tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,23 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — non-finite inputs silently reported success (#275)
+
+- **`lb = +inf` / `ub = -inf` were dropped as if the bound were absent.** The
+  solver's presence test (`lb > -BOUND_INF`, `ub < BOUND_INF`) is
+  sign-agnostic, so a bound no finite value can satisfy was discarded and
+  `solve_qp` returned `status="optimal"` at a point violating it by an
+  infinite margin. `solve_qp` / `solve_qp_batch` / `solve_qp_multi_rhs` and
+  `minimize` now reject these spellings with a message naming the index.
+  `±inf` on the *absent* side (lower `-inf`, upper `+inf`) remains the
+  documented one-sided encoding, and a finite reversed box (`lb=1 > ub=0`) is
+  still reported as `primal_infeasible` rather than raising.
+- **A NaN or infinite `x0` reported `Solve_Succeeded` with `fun=nan` at
+  iteration 0.** Every convergence test is a comparison against a tolerance
+  and comparisons against NaN are False — including the ones that would have
+  rejected the iterate — so the loop fell through to "converged". `minimize`
+  now rejects a non-finite `x0`.
+
 ### Fixed — `check_psd` validated a different matrix than the solver used (#279)
 
 - With a `scipy.sparse` COO `P` containing **duplicate `(row, col)` entries**,
--- a/python/pounce/_minimize.py
+++ b/python/pounce/_minimize.py
@@ -383,6 +383,25 @@ def _normalize_bounds(bounds, n: int):
             f"bounds[{i}] is reversed: lower {lb[i]} > upper {ub[i]}; "
             f"each bound must be (low, high) with low <= high"
         )
+    # ±inf is legal as the *absent* side of a bound (lower = -inf, upper =
+    # +inf), but not as the *present* side: no finite x satisfies `x >= +inf`
+    # or `x <= -inf`. The `lb > ub` test above catches the mixed spellings
+    # (e.g. lower=+inf with a finite upper) but not `lb == ub == ±inf`, where
+    # the comparison is False and the box silently admitted "any x". See
+    # gh #275.
+    for side, arr, bad_val, cmp_txt in (
+        ("lower", lb, np.inf, ">= +inf"),
+        ("upper", ub, -np.inf, "<= -inf"),
+    ):
+        bad = np.where(arr == bad_val)[0]
+        if bad.size:
+            i = int(bad[0])
+            raise ValueError(
+                f"bounds[{i}] has {side} bound {arr[i]}, which no finite value "
+                f"can satisfy (it requires x[{i}] {cmp_txt}). Use "
+                f"{'-inf' if side == 'lower' else '+inf'} (or None in the "
+                f"pair-list form) to leave the {side} side unbounded"
+            )
     return lb, ub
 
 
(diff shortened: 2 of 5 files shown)


jkitchin:
Follow-up filed: #295 — this guard lives in the Python layer only; the `pounce-convex` core's bound-presence test is still sign-agnostic, so the raw `_pounce` bindings and Rust crate callers remain exposed. Moves the invariant into the core as defense-in-depth.

PR #357, opened 2026-07-25
Rewrite declared-Param variable bounds as constraints

### LINKED ISSUE #356: estimate() and gradient() report zero sensitivity to a Param used in a variable bound
A `Param` declared with `declare_sens_param()` that appears in a `Var`'s bounds is invisible to the sensitivity path. `SensitivityInterface` substitutes declared Params only in constraint expressions (`_replace_parameters_in_constraints`), never in bounds, so after the surgery the Param appears nowhere in the model that is solved. The bound is written to the NL file as a constant at its pre-perturbation value, and the reported sensitivity to it is zero.

```python
import pyomo.environ as pyo
import pyomo_pounce as pp

m = pyo.ConcreteModel()
m.p = pyo.Param(initialize=2.0, mutable=True)
m.x = pyo.Var(bounds=(0, m.p), initialize=1.0)
m.obj = pyo.Objective(expr=-m.x)
pp.declare_sens_param(m.p)
pyo.SolverFactory("pounce").solve(m)          # x* = 2

pp.gradient(m.x, wrt=m.p)                     # 0.0; exact answer is 1.0
pp.estimate(m, [(m.p, 3.0)])[m.x]             # 2.0; exact answer is 3.0
```

Writing the same restriction as `m.cap = Constraint(expr=m.x <= m.p)` with `bounds=(0, None)` gives 1.0 and 3.0. The bound form returns the unperturbed value for any perturbation in either direction, so the estimate does not respond to the parameter at all.

`gradient()` is the quiet case: a derivative of zero is a plausible answer and nothing is emitted. `estimate()` does warn, but it attributes the problem to an active-set change rather than to a bound that never moved.

This is inherited from Pyomo's toolbox rather than specific to pounce: the same model through `sensitivity_calculation('sipopt', ...)` also returns 2.0 for the bound form and 3.0 for the constraint form.

## Proposed fix

In `sens_solve`, in the gap between `clone = si.model_instance` and the NL write: for each `Var` on the clone, test `_lb` and `_ub` with `identify_mutable_parameters` against the declared set, clear the matching bound, and add it back as a constraint on the sensitivity block written against the corresponding block Var, which `setup_sensitivity` has already created and recorded in `block._sens_data_list`. Doing it after the surgery rather than before avoids taking over the cloning and re-running the substitution.

This makes pounce answer differently from `sensitivity_calculation` on the same model, deliberately: writing a limit as a bound is the natural way to write it, and requiring a rewrite is the usability problem `declare_sens_param` exists to avoid.

devin-griff:
Closes #356.

`SensitivityInterface` substitutes declared Params in constraint expressions only, so a Param left in a `Var` bound was written to the NL file as a constant at its pre-perturbation value. The bound never moved and the reported sensitivity to that Param was exactly zero.

`sens_solve` now rewrites such a bound as a constraint over the substituted Var. It runs after `setup_sensitivity`, so the Param-to-Var map it needs (`block._sens_data_list`) already exists and the substitution never has to run twice. Going through `replace_expressions` rather than swapping a bare Param means an expression bound such as `bounds=(0, 2*p + 1)` works too.

The rewrite drops the bound. On the clone that is solved, `m.x.ub` reads `None` and the NL row carries the no-bound sentinel `1e19`, so anything downstream reading NL bounds sees no bound there. Two consequences are handled explicitly: `covariance()`'s active-bound projection reads the recorded pre-rewrite value rather than the sentinel, and `estimate()` no longer clamps or warns for those variables, which is correct because the bound now moves with the perturbation and the linear step respects it to first order.

Fixed Vars and Vars on deactivated Blocks are skipped. A fixed Var's bounds are not enforced, so rewriting one imposes a restriction on the pinned Param that the original model never had; a deactivated Block's Var would be pulled into the NL as a free column restricted only by the new row.

## Effect

Minimum-time racing problem with an acceleration cap `u_up`, discretized at `nfe=100`. The cap binds over the acceleration phase, so more allowed acceleration should mean a faster lap.

```
exact re-solve at u_up=1.1:  tf = 15.890520 s

limit written as        tf(1.0)   d tf/d u_up   estimate(1.1)      error
------------------------------------------------------------------------
bound  (before)       16.471138      0.000000       16.471138   5.81e-01
bound  (after)        16.471138     -6.323665       15.838771   5.17e-02
constraint            16.471138     -6.323665       15.838771   5.17e-02
```

The two spellings of the same limit now agree. The residual is linearization error on a 10% perturbation, and the constraint form carries it identically.

## Cost

A simple bound is handled directly in the barrier; a general inequality costs a slack and a Jacobian row. The racing example above trades roughly 100 bounds for roughly 100 rows. That is the price of the limit being differentiable with respect to the Param that sets it, and it is paid only by models that write a bound in terms of a declared Param.

## Tests

Eight cases in `tests/test_sens.py`: bare Param in an upper bound, an expression bound, a lower bound, an indexed Var, a fixed Var, a Var on a deactivated Block, the recorded pre-rewrite bound, and a `declare_fitted` Var whose rewritten bound must still trigger `covariance()`'s projection. Two controls: the constraint spelling still works and agrees, and an undeclared mutable Param in a bound is left as a bound, asserted against the NL bounds of the clone that was solved rather than against the original model.

Full suite: 120 passed, 5 failed. The same 5 fail on `main` without this change (three in `test_binary_check.py`, plus `test_bound_multipliers_populate_ipopt_zL_zU` and `test_multiplier_gradient_matches_finite_difference`).

## Note on divergence

This makes pyomo-pounce answer differently from `sensitivity_calculation` on the same model. Writing a limit as a bound is the natural way to write it, and requiring a rewrite to a constraint is the kind of usability problem `declare_sens_param` exists to avoid. Kept ungated per review, and recorded in the module docstring so it is discoverable without being opt-in-able.

### DIFF
--- a/pyomo-pounce/pyomo_pounce/sens.py
+++ b/pyomo-pounce/pyomo_pounce/sens.py
@@ -50,6 +50,8 @@ import pyomo.environ as pyo
 from pyomo.common.collections import ComponentMap
 from pyomo.contrib.sensitivity_toolbox.sens import SensitivityInterface
 from pyomo.core.base.constraint import Constraint
+from pyomo.core.expr import identify_mutable_parameters
+from pyomo.core.expr.visitor import replace_expressions
 from pyomo.opt import SolverResults, SolverStatus, TerminationCondition
 
 _REG = "_pounce_sens"
@@ -115,6 +117,51 @@ def declare_residual(*containers, group=None):
     to the heteroscedastic sandwich form."""
     for container in containers:
         _registry(container.model()).residuals.append((container, group))
+
+
+def _reformulate_param_bounds(clone):
+    """Move Var bounds that reference a declared Param into constraints.
+
+    `SensitivityInterface` substitutes declared Params in constraint
+    expressions only. A Param left in a bound is therefore written to the
+    NL file as a constant at its pre-perturbation value, so the bound never
+    moves and the reported sensitivity to that Param reads as exactly zero
+    (jkitchin/pounce#356). Rewriting the bound as a constraint over the
+    substituted Var puts it where the perturbation already reaches, which
+    makes the answer exact in the bound rather than approximating it.
+
+    Runs after `setup_sensitivity`, so the Param-to-Var map it needs is the
+    one the surgery has already built.
+    """
+    block = clone.component(SensitivityInterface.get_default_block_name())
+    if block is None:
+        return
+    sub = {id(param): var for var, param, _, _ in block._sens_data_list}
+    if not sub:
+        return
+
+    moved = []
+    for v in clone.component_data_objects(pyo.Var, descend_into=True):
+        for attr in ("_lb", "_ub"):
+            expr = getattr(v, attr, None)
+            if expr is None or isinstance(expr, (int, float)):
+                continue
+            if not any(id(p) in sub
+                       for p in identify_mutable_parameters(expr)):
+                continue
+            moved.append((v, attr, replace_expressions(expr, sub)))
+
+    if not moved:
+        return
+
+    block.boundConst = pyo.ConstraintList()
+    for v, attr, expr in moved:
+        if attr == "_lb":
+            v.setlb(None)
+            block.boundConst.add(expr <= v)
+        else:
+            v.setub(None)
+            block.boundConst.add(v <= expr)
 
 
 def has_declarations(model):
@@ -352,6 +399,7 @@ def sens_solve(model, tee=False, sens_params=None, fitted=None,
         si = SensitivityInterface(model, clone_model=True)
         si.setup_sensitivity(eff_params)
         clone = si.model_instance
+        _reformulate_param_bounds(clone)
     else:
         # estimation-only: nothing to pin, solve the model as written
         si = None
--- a/pyomo-pounce/pyomo_pounce/sens.py
+++ b/pyomo-pounce/pyomo_pounce/sens.py
@@ -34,6 +34,14 @@ is written to .nl and evaluated in-process via pounce.read_nl, and the
 pounce.Solver session's parametric_step answers gradient()/estimate()
 queries from the stored factorization -- the sIPOPT computation, with no
 suffixes and no upfront perturbation values.
+
+One deliberate divergence from pyomo.contrib.sensitivity_toolbox: a
+declared Param appearing in a Var's BOUND is rewritten as a constraint
+before the solve, so its sensitivity is real rather than zero. The
+toolbox substitutes declared Params in constraint expressions only, and
+leaves such a bound frozen at its pre-perturbation value. On the clone
+that is solved the bound is dropped, so m.x.ub reads None there and the
+NL carries the no-bound sentinel for that row.
 """
 import codecs
 import os
@@ -131,17 +139,29 @@ def _reformulate_param_bounds(clone):
     makes the answer exact in the bound rather than approximating it.
 
     Runs after `setup_sensitivity`, so the Param-to-Var map it needs is the
-    one the surgery has already built.
+    one the surgery has already built. Returns {var name: (lb, ub)} of the
+    numeric values the moved bounds had at the solve point, with None on the
+    side that was not moved; covariance()'s active-bound projection reads
+    NL bounds, which now say +/-inf for these rows.
     """
     block = clone.component(SensitivityInterface.get_default_block_name())
     if block is None:
-        return
+        return {}
     sub = {id(param): var for var, param, _, _ in block._sens_data_list}
     if not sub:
-        return
+        return {}
 
     moved = []
-    for v in clone.component_data_objects(pyo.Var, descend_into=True):
+    # active=True so deactivated Blocks are skipped: stripping a bound there
+    # and adding an active constraint would pull the Var into the NL as a
+    # free column. Fixed Vars are skipped because their bounds are not
+    # enforced -- Pyomo substitutes them out as constants -- so turning one
+    # into a constraint would impose a restriction on the pinned Param that
+    # the original model never had.
+    for v in clone.component_data_objects(pyo.Var, active=True,
+                                          descend_into=True):
+        if v.fixed:
+            continue
         for attr in ("_lb", "_ub"):
             expr = getattr(v, attr, None)
             if expr is None or isinstance(expr, (int, float)):
@@ -149,19 +169,27 @@ def _reformulate_param_bounds(clone):
             if not any(id(p) in sub
                        for p in identify_mutable_parameters(expr)):
                 continue
-            moved.append((v, attr, replace_expressions(expr, sub)))
+            # the numeric value the NL would have carried, kept so
+            # covariance() can still see where the bound was
+            moved.append((v, attr, float(pyo.value(expr)),
+                          replace_expressions(expr, sub)))
 
     if not moved:
-        return
+        return {}
 
+    recorded = {}
     block.boundConst = pyo.ConstraintList()
-    for v, attr, expr in moved:
+    for v, attr, val, expr in moved:
+        lo, hi = recorded.get(v.name, (None, None))
         if attr == "_lb":
             v.setlb(None)
             block.boundConst.add(expr <= v)
+            recorded[v.name] = (val, hi)
         else:
             v.setub(None)
             block.boundConst.add(v <= expr)
+            recorded[v.name] = (lo, val)
+    return recorded
 
 
 def has_declarations(model):
@@ -212,6 +240,7 @@ class _Session:
         self.pins = pins              # ComponentMap: param data -> pin row
         self.con_alias = con_alias    # original con name -> clone row name
         self.base_x = None
+        self.moved_bounds = {}        # var name -> (lb, ub) moved to rows
         self._columns = {}            # pin row -> full KKT-space column
 
     def orig_var(self, name):
@@ -399,11 +428,12 @@ def sens_solve(model, tee=False, sens_params=None, fitted=None,
         si = SensitivityInterface(model, clone_model=True)
         si.setup_sensitivity(eff_params)
         clone = si.model_instance
-        _reformulate_param_bounds(clone)
+        moved_bounds = _reformulate_param_bounds(clone)
     else:
         # estimation-only: nothing to pin, solve the model as written
         si = None
         clone = model
+        moved_bounds = {}
 
     # The .nl/.col/.row files exist only to hand the model to read_nl;
     # everything needed later (evaluators, bounds, names) lives in memory,
@@ -513,6 +543,7 @@ def sens_solve(model, tee=False, sens_params=None, fitted=None,
     session = _Session(model, nl, solver, var_names, con_names, pins,
                        con_alias)
     session.base_x = np.asarray(x)
+    session.moved_bounds = moved_bounds
 
     # fitted parameters: their rows in the primal vector
     session.fit_rows = ComponentMap()
@@ -668,6 +699,12 @@ def estimate(model, perturb, clamp=True):
     ComponentMap (plain dicts don't work: Pyomo components are unhashable).
     Returns a ComponentMap {original var data: estimated value}. Values are
     clamped to variable bounds (with a warning) unless clamp=False.
+
+    A bound written in terms of a declared Param is a constraint by the
+    time the model is solved, so it is not clamped against here and no
+    clamp warning is raised for it. That is deliberate: the bound moves
+    with the perturbation, so the linear step already respects it to
+    first order.
     """
     reg = model.__dict__.get(_REG)
     session = reg.session if reg else None
@@ -885,6 +922,15 @@ def covariance(model, sigma_sq=None, n_data=None, hessian="lagrangian"):
             "regularized rather than exact. Linearly dependent (structurally"
             " unidentifiable) parameters are the usual cause.")
     lo, hi = np.asarray(session.nl.x_l), np.asarray(session.nl.x_u)
+    # a bound moved to a constraint row reads as +/-inf here, which would
+    # silently skip the projection for that parameter; put the value the
+    # bound had at the solve point back for the test only
+    for name, (mlo, mhi) in session.moved_bounds.items():
+        r = session.var_names.index(name)
+        if mlo is not None:
+            lo[r] = mlo
+        if mhi is not None:
+            hi[r] = mhi
     active = []
     for i, p in enumerate(params):
         r = session.fit_rows[p]
--- a/pyomo-pounce/pyomo_pounce/sens.py
+++ b/pyomo-pounce/pyomo_pounce/sens.py
@@ -142,7 +142,8 @@ def _reformulate_param_bounds(clone):
     one the surgery has already built. Returns {var name: (lb, ub)} of the
     numeric values the moved bounds had at the solve point, with None on the
     side that was not moved; covariance()'s active-bound projection reads
-    NL bounds, which now say +/-inf for these rows.
+    NL bounds, which for these rows now hold the reader's no-bound sentinel
+    of +/-1e19. That value is finite, so an isinf() test would never fire.
     """
     block = clone.component(SensitivityInterface.get_default_block_name())
     if block is None:
@@ -922,7 +923,8 @@ def covariance(model, sigma_sq=None, n_data=None, hessian="lagrangian"):
             "regularized rather than exact. Linearly dependent (structurally"
             " unidentifiable) parameters are the usual cause.")
     lo, hi = np.asarray(session.nl.x_l), np.asarray(session.nl.x_u)
-    # a bound moved to a constraint row reads as +/-inf here, which would
+    # a bound moved to a constraint row reads as the no-bound sentinel
+    # +/-1e19 here (finite, so isinf() would never catch it), which would
     # silently skip the projection for that parameter; put the value the
     # bound had at the solve point back for the test only
     for name, (mlo, mhi) in session.moved_bounds.items():
--- a/pyomo-pounce/tests/test_sens.py
+++ b/pyomo-pounce/tests/test_sens.py
@@ -381,9 +381,10 @@ def test_undeclared_param_in_bound_is_left_alone():
     declare_sens_param(m.p)
     pyo.SolverFactory("pounce").solve(m)
 
+    assert m.x.ub == 5.0                              # original untouched
     session = m.__dict__["_pounce_sens"].session
-    r = session.var_names.index("x")
-    assert session.nl.x_u[r] == pytest.approx(5.0)    # still an NL bound
+    r = session.var_entry(m.x.name)
+    assert session.nl.x_u[r] == pytest.approx(5.0)    # survived into the solve
     assert session.moved_bounds == {}                 # nothing was rewritten
     assert pyo.value(m.x) == pytest.approx(5.0, abs=1e-6)
 
@@ -399,7 +400,7 @@ def test_moved_bound_is_recorded_for_covariance():
     pyo.SolverFactory("pounce").solve(m)
 
     session = m.__dict__["_pounce_sens"].session
-    r = session.var_names.index("x")
+    r = session.var_entry(m.x.name)
     assert session.nl.x_u[r] >= 1e19                  # NL no-bound sentinel
     assert session.moved_bounds["x"] == (None, pytest.approx(2.0))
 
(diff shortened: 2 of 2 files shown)


pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -117,6 +119,51 @@ def declare_residual(*containers, group=None):
...
+    """
+    block = clone.component(SensitivityInterface.get_default_block_name())
+    if block is None:
+        return
+    sub = {id(param): var for var, param, _, _ in block._sens_data_list}
+    if not sub:
+        return
+
+    moved = []
+    for v in clone.component_data_objects(pyo.Var, descend_into=True):
```
jkitchin: `component_data_objects` defaults to `active=None`, which descends into **deactivated** Blocks — so a Var the user deliberately deactivated has its bound stripped and an *active* constraint added for it on the sens block, which pulls it back into the NL.

Reproduced on pyomo 6.10.1:

```python
m.keep = pyo.Var(bounds=(0, m.p), initialize=1.0)
m.dead = pyo.Block()
m.dead.z = pyo.Var(bounds=(0, m.p), initialize=1.0)
m.dead.c = pyo.Constraint(expr=m.dead.z <= 1)
m.dead.deactivate()
m.obj = pyo.Objective(expr=-m.keep)
```

```
default active=None sees: ['keep', 'dead.z', '_SENSITIVITY_TOOLBOX_DATA.p']
active=True     sees: ['keep', '_SENSITIVITY_TOOLBOX_DATA.p']

moved: [('keep', '_ub'), ('dead.z', '_ub')]
NL cols: ['keep', '_SENSITIVITY_TOOLBOX_DATA.p', 'dead.z']
NL rows: [... 'boundConst[1]', 'boundConst[2]', ...]
```

Without this change `dead.z` is not in the NL at all. With it, `dead.z` is a free column (its bound is gone, and it is not in the objective) restricted only by `boundConst[2]` — a degenerate direction for the interior-point solve, on top of silently changing `n`/`var_names`.

Passing `active=True` fixes it:

```python
for v in clone.component_data_objects(pyo.Var, active=True, descend_into=True):
```

---

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -117,6 +119,51 @@ def declare_residual(*containers, group=None):
...
+        for attr in ("_lb", "_ub"):
+            expr = getattr(v, attr, None)
+            if expr is None or isinstance(expr, (int, float)):
+                continue
+            if not any(id(p) in sub
+                       for p in identify_mutable_parameters(expr)):
+                continue
+            moved.append((v, attr, replace_expressions(expr, sub)))
```
jkitchin: Fixed Vars need to be skipped here. A fixed Var's bounds are never enforced by the solver — Pyomo substitutes it out as a constant — but this rewrite still strips the bound and emits `f <= p`, which with `f` constant becomes a real restriction **on the pinned Param variable**.

Reproduced on pyomo 6.10.1 with `m.f = Var(bounds=(0, m.p)); m.f.fix(5.0)` and `p = 2.0`:

```
[before] cols=['x', '_SENSITIVITY_TOOLBOX_DATA.p']
[before] rows=['constList[1]', 'paramConst[1]', 'obj']

moved: [('f', '_ub')]
[after]  cols=['x', '_SENSITIVITY_TOOLBOX_DATA.p']
[after]  rows=['constList[1]', 'paramConst[1]', 'boundConst[1]', 'obj']
```

`f` is correctly absent from the columns both times, so `boundConst[1]` is `5.0 <= p` while `paramConst[1]` pins `p == 2.0` — the model is now infeasible. Before this PR the same model only drew Pyomo's `W1002` "outside the bounds" warning and solved fine. Even when the fixed value is *within* its bounds the row is pure redundancy against the pin.

```python
    for v in clone.component_data_objects(pyo.Var, active=True, descend_into=True):
        if v.fixed:
            continue
```

---

pyomo-pounce/tests/test_sens.py
```diff
@@ -315,3 +315,77 @@ def test_all_three_declarations_coexist():
...
+def test_undeclared_param_in_bound_is_left_alone():
+    """Only declared Params are rewritten; an undeclared one stays a bound."""
+    m = pyo.ConcreteModel()
+    m.p = pyo.Param(initialize=2.0, mutable=True)
+    m.cap = pyo.Param(initialize=5.0, mutable=True)   # never declared
+    m.x = pyo.Var(bounds=(0, m.cap), initialize=1.0)
+    m.obj = pyo.Objective(expr=(m.x - m.p) ** 2)
+    declare_sens_param(m.p)
+    pyo.SolverFactory("pounce").solve(m)
+    assert m.x.ub == 5.0                              # bound survived
```
jkitchin: This assertion can't fail, so the control doesn't test its stated claim. `_reformulate_param_bounds` only ever mutates the **clone** (`si.model_instance`); `m` is never touched, so `m.x.ub` reads `5.0` regardless.

Ran the same model both ways — with the bound's Param undeclared (rewrite must not fire) and declared (rewrite does fire):

```
declared bound param=False: rewrite happened=False | ORIGINAL m.x.ub=5.0 | CLONE x.ub=5.0
declared bound param=True:  rewrite happened=True  | ORIGINAL m.x.ub=5.0 | CLONE x.ub=None
```

The `gradient` assertion below doesn't discriminate either: `cap=5.0` is inactive at `x* = p = 2`, so `dx/dp == 1.0` holds whether or not `cap` was wrongly rewritten.

The behavior itself is correct — `sub` holds only declared Params, so `identify_mutable_parameters(m.cap)` misses and the bound is skipped. It just needs an assertion that would notice if that broke. The bound reaches the solver through the NL, so checking there is both faithful and clone-agnostic:

```python
assert m.x.ub == 5.0                              # original model untouched
sess = m.__dict__["_pounce_sens"].session
assert sess.nl.x_u[sess.var_entry(m.x.name)] == 5.0   # bound survived into the solve
```

Making `cap` bind (so a wrongly-rewritten bound would change the answer) would be a good complement.

---

jkitchin (changes requested):
## Summary

The diagnosis and the core mechanism are right, and the racing-problem table is convincing. I verified the crux empirically on pyomo 6.10.1: `block._sens_data_list` carries the **clone's** Param object, and its `id()` does match the Param inside the clone's `_lb`/`_ub`, so the substitution actually fires:

```
sens_data_list: [('_SENSITIVITY_TOOLBOX_DATA.p', 'p', 'ScalarParam')]
clone x._ub param id matches sens_data param id: True | is it the ORIGINAL m.p? False
```

Also confirmed: routing through `replace_expressions` really does handle `bounds=(0, 2*p + 1)` (and a bare-leaf root, which is the less obvious case); a Param appearing *only* in a bound and in no constraint still pins cleanly; and `sensitivity_toolbox` touches no bound machinery of its own, so there's nothing to collide with. All five checks are green.

Three things to fix before merge — two are behavior bugs in the new function, one is a test that can't fail. Details inline.

1. **Vars on deactivated Blocks are resurrected into the solve.** `component_data_objects` defaults to `active=None`, so it descends into deactivated Blocks; the bound is stripped and an *active* constraint is added, pulling the Var into the NL as a free column. Needs `active=True`.
2. **Fixed Vars produce a spurious constraint on the pinned Param.** A fixed Var's bounds aren't enforced (it's substituted out as a constant), but `f <= p` becomes a real restriction on the pin — I got a straightforwardly infeasible model out of one that previously solved with only a `W1002` warning. Needs `if v.fixed: continue`.
3. **`test_undeclared_param_in_bound_is_left_alone` is vacuous.** It asserts on the original model, which the rewrite never touches.

## Notes, not blockers

**`covariance()` loses its active-bound projection for rewritten bounds.** Once a bound becomes a constraint, `nl.x_u` for that row is `+inf`. The projection at `sens.py:887-899` detects fitted parameters sitting on a bound by comparing against `nl.x_l`/`x_u`, so a `declare_fitted` Var carrying a declared-Param bound would silently stop triggering both the projection and its nonstandard-asymptotics warning — reporting an unconditional covariance where it used to report the conditional one. Narrow, but `test_all_three_declarations_coexist` establishes the declarations as composable, so it's reachable. Worth either skipping Vars in `session.fit_rows`, or recording the moved bounds so `covariance()` can still see them.

The same `inf` disables `estimate()`'s clamp warning for those variables. That one looks correct to me — the bound now moves with the perturbation, so the linear step respects it to first order — but it's a deliberate change in behavior and worth a line in the docstring.

**Interior-point cost.** A simple bound is handled directly in the barrier; a general inequality costs a slack plus a Jacobian row. The racing example at `nfe=100` trades ~100 bounds for ~100 rows. Not a defect, but worth stating in the PR body so the tradeoff is on the record.

## On the divergence question

I'd keep it ungated. A limit written as a bound silently reporting exactly zero sensitivity is a wrong answer, not a differing convention, and sparing users the manual rewrite is precisely what `declare_sens_param` is for. A flag would mostly preserve the bug and give callers a way to opt into it unknowingly. If the divergence from `sensitivity_calculation` needs to be discoverable, the module docstring seems like the better place for it than an option.

One request either way: the PR body should mention that the rewrite drops the bound, since `m.x.ub` on the clone becomes `None` and anything downstream reading NL bounds sees `inf`.

---

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Confirmed on my side too, same shape:

```
before: ['x', '_SENSITIVITY_TOOLBOX_DATA.p', 'dead.y']
after:  ['x', '_SENSITIVITY_TOOLBOX_DATA.p']
```

The degenerate-direction point is the part I had not thought through. It isn't only that `n` and `var_names` change: the Var arrives with its bound gone and no objective term, so the only thing touching it is the `boundConst` row the rewrite just added. That is a free column pinned by one inequality, handed to an interior-point solve.

`active=True` applied. Added `test_deactivated_block_var_is_not_rewritten`, which builds your model and asserts nothing matching `dead.y` reaches `var_names`.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Reproduced exactly: `optimal` before this PR with only the `W1002`, `infeasible` after, because `boundConst[1]` is `5.0 <= p` against `paramConst[1]` pinning `p == 2.0`.

Your second point is the one that generalizes, and it is why the skip is unconditional rather than a check on whether the fixed value violates its bounds: even in range, the row is redundancy against the pin, so it buys nothing and costs a Jacobian row. The code comment says the restriction is one the original model never had, which covers both.

`if v.fixed: continue` applied, with `test_fixed_var_bound_is_not_rewritten` asserting the model comes back optimal and `dx/dp` is still `1.0`.

pyomo-pounce/tests/test_sens.py
devin-griff: Right on both counts, including that the `gradient` assertion doesn't discriminate either. `cap = 5.0` is slack at `x* = p = 2`, so that line passes whether or not the rewrite wrongly fired.

Took both suggestions. The objective is now `-x` so the cap binds, and the assertions are:

```python
assert m.x.ub == 5.0                              # original untouched
session = m.__dict__["_pounce_sens"].session
r = session.var_entry(m.x.name)
assert session.nl.x_u[r] == pytest.approx(5.0)    # survived into the solve
assert session.moved_bounds == {}                 # nothing was rewritten
assert pyo.value(m.x) == pytest.approx(5.0, abs=1e-6)
```

Kept your `m.x.ub` line, relabelled to say what it actually verifies rather than standing in for the bound surviving. `var_entry` instead of indexing `var_names` directly, in the other two new tests as well.

devin-griff:
All three reproduced exactly as written, and the two behavior bugs are fixed with tests that fail without the fix.

The `covariance()` note I fixed rather than deferred, since your reachability argument via `test_all_three_declarations_coexist` is right. `_reformulate_param_bounds` now returns `{var name: (lb, ub)}` of the numeric values the moved bounds held at the solve point, the session carries them, and `covariance()` overlays them onto `lo`/`hi` for the activity test alone. A `declare_fitted` Var capped by a declared Param sits on its bound at the optimum and still raises the nonstandard-asymptotics warning.

One correction: the row reads as the NL no-bound sentinel `1e19`, not `inf`. Identical effect on the comparison, but the assertion had to test for the sentinel.

The clamp change is documented in `estimate()`'s docstring as deliberate, using your reasoning. Divergence kept ungated and recorded in the module docstring rather than as an option, including that the rewrite drops the bound so `m.x.ub` on the clone is `None`. PR body updated with that and with the interior-point tradeoff, roughly 100 bounds becoming 100 rows on the racing example at `nfe=100`.

Suite is 120 passed, 5 failed, the same 5 that fail on `main`: three in `test_binary_check.py`, plus `test_bound_multipliers_populate_ipopt_zL_zU` and `test_multiplier_gradient_matches_finite_difference`.

devin-griff:
Version check, since you verified the crux on 6.10.1 and I had done the fix work on 6.10.0. Bumped locally and re-ran everything on 6.10.1:

- Suite: 120 passed, 5 failed, the same 5 as before and the same 5 that fail on `main`.
- The racing table is unchanged to every digit: `0.000000` before, `-6.323665` after, matching the constraint spelling.

So both halves are now confirmed on 6.10.1 rather than one on each.

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -117,6 +127,71 @@ def declare_residual(*containers, group=None):
...
+    moves and the reported sensitivity to that Param reads as exactly zero
+    (jkitchin/pounce#356). Rewriting the bound as a constraint over the
+    substituted Var puts it where the perturbation already reaches, which
+    makes the answer exact in the bound rather than approximating it.
+
+    Runs after `setup_sensitivity`, so the Param-to-Var map it needs is the
+    one the surgery has already built. Returns {var name: (lb, ub)} of the
+    numeric values the moved bounds had at the solve point, with None on the
+    side that was not moved; covariance()'s active-bound projection reads
+    NL bounds, which now say +/-inf for these rows.
```
jkitchin: Your correction stands — the sentinel is `1e19`, not `inf`. `crates/pounce-nl/src/nl_reader.rs:334-337`:

```rust
let mut x_l = vec[-1e19; n];
let mut x_u = vec[1e19; n];
```

with `const INF: Number = 1.0e19;` at `nl_reader.rs:1858` and the comment above it noting that bounds outside ±1e19 are treated as infinite. So my original review was wrong on this.

Which is why these two comments should say so too — they still read `+/-inf`:

- here: *"NL bounds, which now say +/-inf for these rows"*
- `sens.py:925`: *"a bound moved to a constraint row reads as +/-inf here"*

The PR body, the module docstring ("no-bound sentinel"), and all the new assertions (`>= 1e19`) have it right; it's only these two.

Worth fixing rather than leaving as a wording slip, because `1e19` is *finite*. Anyone extending the projection later and reasoning from these comments would naturally reach for

```python
if np.isinf(hi[r]):        # never true — silently skips every moved bound
```

which fails closed in the least visible way possible: the projection stops firing and the nonstandard-asymptotics warning goes quiet, which is exactly the bug the overlay was added to fix.

---

pyomo-pounce/tests/test_sens.py
```diff
@@ -315,3 +315,163 @@ def test_all_three_declarations_coexist():
...
+    declare_fitted(m.A)
+    declare_residual(m.r)
+    pyo.SolverFactory("pounce").solve(m)
+
+    session = m.__dict__["_pounce_sens"].session
+    assert session.nl.x_u[session.var_entry(m.A.name)] >= 1e19
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
+        covariance(m)
+    assert any("sits on its bound" in str(x.message) for x in w)
```
jkitchin, line 477: Suggestion, not a gate — the overlay is correct as written and I verified the recording logic handles both sides.

This asserts the warning fires but not what the covariance comes out as. The PR's headline claim is that the two spellings of the same limit agree, and the racing table establishes that for `gradient`/`estimate`. `covariance()` is the one place where the two spellings reach the answer by genuinely different routes: with a plain bound the variable is pinned by the barrier diagonal `Z_U/(u-x)`, whereas after the rewrite it is pinned through the `boundConst` Jacobian row and its multiplier. `M` is the same block of `inv(KKT)` either way, but it is assembled differently.

I expect them to agree — both routes leave that direction degenerate, so the projection is the right treatment in both — but that's inference on my part, not something the suite currently checks. The direct version:

```python
def test_covariance_agrees_bound_vs_constraint():
    """The two spellings of the same cap give the same covariance."""
    def build(as_bound):
        m = pyo.ConcreteModel()
        m.I = pyo.RangeSet(5)
        m.cap = pyo.Param(initialize=1.0, mutable=True)
        m.A = pyo.Var(bounds=(0, m.cap) if as_bound else (0, None),
                      initialize=0.5)
        if not as_bound:
            m.capcon = pyo.Constraint(expr=m.A <= m.cap)
        m.r = pyo.Var(m.I, initialize=0.0)
        m.res = pyo.Constraint(m.I, rule=lambda mm, i: mm.r[i] == 3.0 - mm.A)
        m.obj = pyo.Objective(expr=sum(m.r[i] ** 2 for i in m.I))
        declare_sens_param(m.cap); declare_fitted(m.A); declare_residual(m.r)
        with warnings.catch_warnings():
            warnings.simplefilter("ignore")
            pyo.SolverFactory("pounce").solve(m)
            return covariance(m).std_err[m.A]

    assert build(True) == pytest.approx(build(False), rel=1e-6)
```

If they turn out *not* to agree, that's worth knowing before merge — it would mean the projection is doing something different in the rewritten case, and the current test would not have caught it.

---

jkitchin (approved):
All three blocking items are fixed, and I checked the code rather than going off the replies. CI is green on `8f22eca`.

- `active=True` applied with a comment explaining why, and `test_deactivated_block_var_is_not_rewritten` asserts `dead.y` never reaches `var_names`.
- `if v.fixed: continue` applied. Your reasoning for making the skip unconditional rather than testing whether the fixed value violates its bound is better than what I suggested — an in-range fixed value still produces a row that is pure redundancy against the pin, so there is nothing to condition on.
- The control test now asserts against `session.nl.x_u` on the clone that was solved, plus `session.moved_bounds == {}`, and the objective change to `-x` makes the cap actually bind so a wrongly-fired rewrite would move the answer.

On the `covariance()` overlay: I traced the recording path and the accumulation is right — `moved` is built in `("_lb", "_ub")` order and the `recorded.get(v.name, (None, None))` merge preserves both sides when a Var has both bounds rewritten, which the single-sided tests would not have caught on their own.

One hazard I went looking for and did **not** find, recording it so nobody has to re-check: `lo[r] = mlo` mutating the session's stored NL bounds and corrupting `estimate()`'s clamp on later calls. It can't, because the getter builds a fresh array per access — `nl_problem.rs:137`, `self.x_l.clone().into_pyarray_bound(py)`. The overlay is safely local to `covariance()`.

Approving. Two non-blocking notes inline: the `inf`-vs-`1e19` wording in two comments (worth fixing precisely because `1e19` is finite, so reasoning from those comments leads to an `np.isinf` check that silently never fires), and a suggested bound-vs-constraint agreement test for `covariance()` to close the PR's own "two spellings agree" claim on the one path where the two routes differ. Neither needs to hold the merge — take them or leave them.

---

devin-griff:
Both taken.

The `1e19` correction is in, on both comments. Your framing of why it matters is the right one: the sentinel is finite, so the natural reading of those comments leads to an `isinf` test that never fires and quietly disables the projection.

On the agreement test, I ran it and it fails, so I have not added it as written:

```
bound, rewrite off (pre-PR)   std_err = 0.0
bound, rewrite on  (this PR)  std_err = 0.0
cap as a Constraint instead   std_err = 7.915413472554487e-06
```

The two spellings genuinely disagree, but not because of this branch. The bound route gives `0.0` with the rewrite on and off, so the overlay reproduces the old answer exactly, which is all it was for. The gap is that the activity test reads variable bounds only and never constraint rows, so a variable pinned by an active constraint has always come back as free. Filed as #362; it wants a classifier covering rows rather than a local patch, and I will propose that in a coming roadmap update.

So the assertion I added is the one this PR is actually responsible for: the rewritten bound still produces the projected `0.0`, not just the warning. I also added the both-bounds recording test, since you are right that the single-sided ones pass even if the merge drops a side.

Suite is 121 passed, same 5 pre-existing failures.

PR #363, opened 2026-07-25
Use presolve in pounce-rs

GermanHeim:
Hello!

Currently, `presolve=yes` only takes effect when callers manually wrap their TNLP. Library solves through `IpoptApplication::optimize_tnlp` (including `pounce-rs`) bypass generic NLP presolve. I found this while updating the oximo-pounce API.

This change wires presolve into `optimize_tnlp` so every library solve actually honors `presolve=yes`. The wrapper is created once per public solve, before algorithm dispatch, and is reused across IPM, SQP, and retry paths. Its existing postsolve mapping continues to return original-space primal values, constraint values, and multipliers to user callbacks.

I keep FBBT separate from this PR, since it needs an `ExpressionProvider`, which a generic callback TNLP does not provide. A follow-up PR could provide an `optimize_tnlp_with_expression_provider(...)` to enable it.

Also, for the algorithms lp-ipm, qp-ipm, and socp, deliberately return `InvalidOption` in `IpoptApplication` since they require Pounce’s CLI .nl parser to extract a convex LP/QP/QCQP and route it to `pounce-convex`. Maybe there are other options that aren't being correctly parsed from `pounce-rs`? I allow edits by maintainers in this PR in case you want to take a look at the parameter parsing/use, but we could probably open another PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ changes.
 
 ## [Unreleased]
 
+### Changed
+
+- **`pounce-nlp` warm starts now fetch one coherent TNLP starting-point
+  snapshot.** With `warm_start_init_point=yes`, the adapter calls
+  `TNLP::get_starting_point` once with the primal, bound-multiplier, and
+  constraint-multiplier flags enabled, then projects that single payload into
+  the IPM's `x`, `y`, and `z` blocks. This replaces three independent callback
+  invocations, matching Ipopt's `GetStartingPoint` contract and avoiding
+  inconsistent or side-effect-dependent warm starts in existing library,
+  Python, and C-bridge users.
+
 ## [0.9.0] - 2026-07-24
 
 ### Fixed — active-set-SQP stalled on curved-constraint NLPs via the Maratos effect (#349)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ changes.
 
 ## [Unreleased]
 
+### Added
+
+- **Generic presolve for library TNLP solves.** `presolve=yes` now wraps every
+  `IpoptApplication::optimize_tnlp` call once, before algorithm dispatch, so
+  IPM, SQP, and retry paths all use the same reduced TNLP while
+  `finalize_solution` continues to receive original-space values and
+  multipliers. Library code no longer needs to call `wrap_with_presolve` for
+  the ordinary callback-TNLP case.
+  - `IpoptApplication::set_presolve_already_applied` declares an explicit
+    caller-owned presolve wrapper; `optimize_tnlp_without_presolve` is the
+    scoped bypass for consumers, such as sensitivity analysis, that require
+    the original KKT coordinate system; and `TNLP::is_presolve_wrapper` lets
+    generic TNLP decorators preserve the no-double-wrap marker.
+
 ### Changed
 
 - **`pounce-nlp` warm starts now fetch one coherent TNLP starting-point
@@ -18,7 +32,9 @@ changes.
   the IPM's `x`, `y`, and `z` blocks. This replaces three independent callback
   invocations, matching Ipopt's `GetStartingPoint` contract and avoiding
   inconsistent or side-effect-dependent warm starts in existing library,
-  Python, and C-bridge users.
+  Python, and C-bridge users. A TNLP that refuses that requested warm-start
+  payload now fails explicitly with `Invalid_Problem_Definition` rather than
+  silently using default iterates.
 
 ## [0.9.0] - 2026-07-24
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1107,6 +1107,7 @@ dependencies = [
  "pounce-common",
  "pounce-linalg",
  "pounce-nlp",
+ "tracing",
 ]
 
 [[package]]
--- a/crates/pounce-algorithm/src/application.rs
+++ b/crates/pounce-algorithm/src/application.rs
@@ -515,6 +515,24 @@ impl IpoptApplication {
         // `algorithm` option resolves to "active-set-sqp", route
         // to the Phase 5b SQP path; otherwise fall through to the
         // existing IPM flow unchanged.
+        // Materialize generic TNLP presolve once at the public entry point.
+        // The wrapper owns the submitted callback TNLP, so every algorithm
+        // path below (including retry paths) continues to postsolve into
+        // the original user-facing space. With `presolve=no`, this returns
+        // the exact same Rc unchanged.
+        let tnlp = match pounce_presolve::wrap_from_options(tnlp, &self.options) {
+            Ok(tnlp) => tnlp,
+            Err(err) => {
+                use pounce_common::journalist::JournalCategory;
+                self.journalist.print(
+                    JournalLevel::J_ERROR,
+                    JournalCategory::J_MAIN,
+                    &format!("pounce: could not materialize presolve options: {err}\n"),
+                );
+                return ApplicationReturnStatus::InvalidOption;
+            }
+        };
+
         if self.is_sqp_algorithm_selected() {
             return self.optimize_sqp_tnlp(tnlp);
         }
--- a/crates/pounce-algorithm/src/application.rs
+++ b/crates/pounce-algorithm/src/application.rs
@@ -284,6 +284,23 @@ impl IpoptApplication {
         self.presolve_already_applied = applied;
     }
 
+    /// Solve without materializing the generic presolve wrapper.
+    ///
+    /// This is for consumers that require the original TNLP coordinate system
+    /// for the solve's KKT matrix, such as sensitivity and reduced-Hessian
+    /// drivers. It is scoped to this invocation and does not change the
+    /// application's `presolve` option or persistent explicit-wrapper setting.
+    pub fn optimize_tnlp_without_presolve(
+        &mut self,
+        tnlp: Rc<RefCell<dyn TNLP>>,
+    ) -> ApplicationReturnStatus {
+        let explicit_wrapper = self.presolve_already_applied;
+        self.presolve_already_applied = true;
+        let status = self.optimize_tnlp(tnlp);
+        self.presolve_already_applied = explicit_wrapper;
+        status
+    }
+
     pub fn registered_options(&self) -> &Rc<RegisteredOptions> {
         &self.reg_options
     }
--- a/crates/pounce-algorithm/src/application.rs
+++ b/crates/pounce-algorithm/src/application.rs
@@ -101,6 +101,9 @@ use std::time::Instant;
 
 pub struct IpoptApplication {
     options: OptionsList,
+    /// Whether the submitted TNLP has already been explicitly wrapped by the
+    /// caller's presolve layer.
+    presolve_already_applied: bool,
     reg_options: Rc<RegisteredOptions>,
     journalist: Rc<Journalist>,
     statistics: RefCell<SolveStatistics>,
@@ -240,6 +243,7 @@ impl IpoptApplication {
         let reg = Rc::new(reg);
         Self {
             options: OptionsList::with_registered(Rc::clone(&reg)),
+            presolve_already_applied: false,
             reg_options: reg,
             journalist: Rc::new(Journalist::new()),
             statistics: RefCell::new(SolveStatistics::new()),
@@ -266,6 +270,18 @@ impl IpoptApplication {
 
     pub fn options_mut(&mut self) -> &mut OptionsList {
         &mut self.options
+    }
+
+    /// Declare whether callers have already applied an explicit presolve
+    /// wrapper to the TNLPs submitted to [`Self::optimize_tnlp`].
+    ///
+    /// When set, `optimize_tnlp` leaves its input TNLP unchanged even if the
+    /// `presolve` option is enabled. This preserves the option table for
+    /// reporting and debugger use while allowing specialized frontends to
+    /// supply a wrapper with capabilities unavailable to generic callback
+    /// TNLPs, such as an expression provider for FBBT.
+    pub fn set_presolve_already_applied(&mut self, applied: bool) {
+        self.presolve_already_applied = applied;
     }
 
     pub fn registered_options(&self) -> &Rc<RegisteredOptions> {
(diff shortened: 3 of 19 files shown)


jkitchin (commented):
Thanks for this — the core idea is right, and the placement is right: one wrap point, before algorithm dispatch, with the same `Rc` reused across the SQP path and both the ℓ₁ and μ-strategy retry loops, so there's no double presolve within a solve. I checked out the head, traced the affected paths, and ran the new suite locally (`cargo test -p pounce-algorithm --test presolve_application` → 3 passed).

## Issues

**1. `--minima` bypasses the CLI guard — behavior regression (highest-priority).** `main.rs` returns into `pounce_cli::minima::run(...)` at ~line 590, well before the `presolve=false` clobber at line 1029. `minima/mod.rs:134,173` call `app.optimize_tnlp` directly, so with `presolve yes` every multistart local solve now gets a presolve wrapper — directly contradicting the comment at `main.rs:573-576` ("presolve / counting wrappers are intentionally bypassed so coordinates match the original problem"). Concrete fallout: the `on_converged` capture yields kept-row-space `lambda`, so `recover_duals`'s `lambda.len() == self.m` check (`minima/mod.rs:180`) fails and **every minimum's duals silently become zeros** in the `.sol` output. Plus presolve analysis re-runs on each of potentially hundreds of solves, and it presolves the *penalty-augmented* TNLP rather than the base problem. Fix: move the opt-out above the `--minima` early return, or set it inside `minima::run`.

**2. The opt-out mechanism itself is the wrong shape.** Issue 1 is a direct consequence of "every caller must remember to clobber a user-visible option." Consider a non-option opt-out on the application instead — e.g. a private `presolve_already_applied` flag set by `app.set_presolve_applied(true)`, or an `optimize_tnlp_presolved(...)` entry point that skips the wrap. Clobbering the option has three side effects beyond the guard:

- `print_user_options yes` now reports `presolve = no` for a user who asked for `yes` (both calls pass `dont_print: false`; `dont_print: true` would at least hide it).
- The debugger's `resolve presolve …` staged override is discarded on every loop iteration, silently. A one-line `eprintln!` warning when `req.options` contains `presolve` would be honest about it.
- Any future reader of the option post-solve sees the wrong value.

**3. Library sensitivity / reduced-Hessian path has no guard.** `main.rs:969-975` disables presolve because "sensitivity / reduced-Hessian post-processing operates on the original (un-presolved) KKT system." `pounce-sensitivity` (`solver.rs:198`, `convenience.rs:454`) calls `optimize_tnlp` directly with no such guard — so `presolve=yes` + library sIPOPT now silently computes sensitivities against the reduced KKT system while the user's parameter indices are original-space. Mirror the CLI's warn-and-disable inside those drivers, or reject the combination.

**4. Silent double-wrap for existing library callers.** `wrap_with_presolve` / `wrap_with_presolve_provider` are public and documented as the way library callers get presolve. Anyone following that pattern who also has `presolve=yes` in their options list now presolves twice (nested postsolve, doubled eval indirection, `presolve_licq_action=auto_l1` potentially firing twice). Worth making `PresolveTnlp` detectable (marker method on the TNLP trait, or an `Any` downcast) and skipping the wrap when already wrapped — that would also solve issues 1 and 2 structurally. At minimum it needs a CHANGELOG note.

**5. The `pounce-nlp` change is an unrelated bug fix riding along.** It's a *good* fix — `warm_start.rs`'s module doc already promises to "forward user-supplied primal/dual seeds via `TNLP::get_starting_point`", and the missing impls meant the trait defaults dropped them on the floor. But it changes numerics for every existing `warm_start_init_point=yes` user, and for `bound_mult_init_method != "constant"` it changes cold-start `y_c`/`y_d` too (`init/default.rs:344`). It deserves its own CHANGELOG bullet and its own unit test in `pounce-nlp`, and arguably its own PR.

**6. `library_presolve_projects_warm_start_and_restores_original_payload` doesn't test what it claims.** `warm_z_seen` / `warm_lambda_seen` are recorded by the TNLP *itself*, immediately after it writes those very values, so `assert_eq!(problem.warm_z_seen, Some((vec[3.0], vec[0.0])))` holds whether or not the new `get_starting_y/z` forwarding works — it only proves `init_z` was true on some call. To actually cover 1591c3e, assert on the solver's iteration-0 duals (intermediate callback / `IterateSnapshot`), or unit-test `OrigIpoptNlp::get_starting_z` directly.

**7. `get_starting_z` ignores `v_l` / `v_u` but returns `true`.** Benign today (callers pre-zero them and `mult_bound_push` floors them off the boundary), but the asymmetry deserves a comment: slack bound multipliers aren't forwarded because there's no user-facing TNLP surface for them.

**8. Three `get_starting_point` calls per warm-started solve.** `get_starting_x`, `get_starting_y`, `get_starting_z` each invoke the user callback and each allocate four full-length scratch vectors; upstream Ipopt makes one call with all flags. Noticeable for Python/C bridges with expensive starting-point hooks.

## Smaller notes

- `application.rs:515-533`: the new block was inserted *between* the pre-existing "Top-level algorithm dispatch (Phase 5b §7.1)…" comment and the `if self.is_sqp_algorithm_selected()` it describes. Move one or the other.
- `get_starting_y` allocates `x`, `z_l`, `z_u` scratch that `initialize_starting_point` never writes (all `init_*` false). Harmless, but a comment saying they're required-but-unused would help.
- No CHANGELOG entry; the repo maintains one and this is user-visible on three fronts (library presolve, warm-start duals, CLI guard).
- `docs/src/options.md` §"NLP Presolve" should say presolve now applies to library solves, not just the CLI.
- Option help wording: "while its dual attribution is carried by the corresponding tightened variable-bound multiplier" reads awkwardly — suggest "…is returned with a zero row multiplier; its dual contribution appears in the corresponding tightened variable-bound multiplier (`z_L`/`z_U`) instead."
- C API: with rows dropped, `GetIpoptCurrentIterate` / `GetIpoptCurrentViolations` hit their `full.len() != m_us` guards and return `FALSE` mid-solve. Not UB (the guards are correct), but it's a silent capability loss under `presolve=yes` worth documenting.

## On the open questions in the description

- **FBBT deferred** — agreed, `optimize_tnlp_with_expression_provider(...)` is the right follow-up shape.
- **`lp-ipm` / `qp-ipm` / `socp` returning `InvalidOption`** — that's `unsupported_library_solver_selection`, and it's deliberate rather than an option-parsing gap: the convex route needs the CLI's `.nl` classification to extract the LP/QP/QCQP before handing it to `pounce-convex` (`main.rs:700-760`). The fix is to expose convex routing as a library API taking a built QP, not to change parsing — worth a separate issue.

## Verdict

Sound direction, and the postsolve mapping does return original-space `g`/`lambda` to callbacks as claimed (verified by the new tests). I'd want **#1** fixed before merge (silent dual loss on `--minima`), **#3** at least warned about, and **#2** reconsidered since it's the root cause of #1. The rest are follow-ups.

---

jkitchin (commented):
Re-reviewed `4320d0f` + `869cc19`. These land #1, #2, #3 and #4 from my earlier review, in the shape I'd hoped for: `set_presolve_already_applied` / `optimize_tnlp_without_presolve` replace the option clobber, the flag is set in the `--minima` branch before `minima::run`, and the sensitivity drivers route through `optimize_tnlp_for_sensitivity`. The user's `presolve` option is now left intact, so `print_user_options` and the debugger both see the truth. `explicit_presolve_bypass_preserves_user_option_and_unwrapped_tnlp` is a real test — asserting `lambda[0].abs() > 1.0` proves the row survived. All 4 + 6 tests pass locally for me.

Two new small things, then the leftovers.

**`solver_keeps_sensitivity_kkt_in_original_space_when_presolve_is_enabled` is vacuous.** `ParametricTNLP` is 4 equality rows (`g_l == g_u`), so there is nothing for redundant-row removal to drop and presolve is a near-no-op on it. I verified this empirically: swapping the guard back to a plain `app.optimize_tnlp(tnlp)` in `optimize_tnlp_for_sensitivity` leaves the test passing. The surviving assertion (option not overwritten) is trivially true once the clobber is gone. To make it bite, give the fixture a droppable linear inequality row and assert the sensitivity KKT / `converged()` dims still match the original `m`.

**`eprintln!` in a library crate.** `optimize_tnlp_for_sensitivity` prints to stderr on every `Solver::solve`, which for a repeated-solve session is noisy and unswitchable. `application.rs` uses `tracing::warn!` for the analogous warm-start-dims warning — `tracing::warn!(target: "pounce::sensitivity", …)` would fit the codebase better, and the CLI already has its own message for this case.

Still open from the first pass:

- **#5** — the `pounce-nlp` `get_starting_y` / `get_starting_z` fix still has no dedicated test in `pounce-nlp` and no CHANGELOG bullet. It changes numerics for every existing `warm_start_init_point=yes` user, so it shouldn't land unannounced.
- **#6** — `library_presolve_projects_warm_start_and_restores_original_payload` still asserts on values the TNLP recorded about itself, so it passes whether or not the forwarding works.
- **#7 / #8** — `v_l`/`v_u` comment; three `get_starting_point` calls per warm-started solve.
- CHANGELOG entry and the `docs/src/options.md` §"NLP Presolve" note that presolve now applies to library solves.

---

jkitchin (commented):
Re-reviewed through `f52b0f9`. This addresses everything outstanding, and a couple of the fixes are better than what I suggested. Ran it locally on the new head: `pounce-nlp` lib 41 + 51, `presolve_application` 5/5, `solver_session` 6/6, all green.

- **`eebb79f`** — `prepare_warm_start()` + a cached `StartingPointSnapshot` is the right answer to the triple-callback problem, and better than the "cache it somewhere" hand-wave in my note: one `get_starting_point` with all three flags set is exactly the upstream `GetStartingPoint` contract, and it also makes the presolve wrapper's z-hint overlay coherent with its lambda projection instead of split across three independent calls. The two new `pounce-nlp` tests are real ones — `warm_start_duals_are_forwarded_into_algorithm_vectors` pins the projected values, `warm_start_prefetches_one_tnlp_snapshot_for_x_y_and_z` pins the call count at 1 — and dropping the self-reported `warm_z_seen` / `warm_lambda_seen` assertions is the right call.
- **`f20c7fb`** — `TNLP::is_presolve_wrapper()` as a defaulted trait method is a cleaner fix for the double-wrap case than the `Any` downcast I floated, and the `Rc::ptr_eq` test is unambiguous.
- **`313479a`** — the sensitivity fixture now has a genuinely droppable linear row and asserts `converged().block_dims() == [5,1,4,1,3,0,1,0]`. I re-ran the experiment from my last review: swapping the guard back to a plain `app.optimize_tnlp(tnlp)` now **fails** this test. It bites.
- **`c2e6000`**, **`f52b0f9`** — `tracing::warn!` and the docs/CHANGELOG updates, as discussed.

Remaining items, all minor:

**1. `Internal_Error` is the wrong status for a refused starting point.** `seed_from_nlp` now returns `bool`, `set_initial_iterates` propagates it, and `ipopt_alg.rs:2882` turns that into `SolverReturn::InternalError`. So a TNLP whose `get_starting_point` returns `false` under `warm_start_init_point=yes` now fails the solve, where previously it proceeded with defaults. Failing is upstream-correct — but `Invalid_Problem_Definition` is the honest status here; `Internal_Error` points the user at pounce rather than at their own callback. This is the one genuinely new behavioral edge in `eebb79f` and it isn't in the CHANGELOG either way.

**2. `warm_start_snapshot` is never cleared.** `OrigIpoptNlp` is rebuilt per solve so the blast radius is a single solve, but any later `get_starting_x` within that solve keeps serving the cache rather than re-querying. A `take()` at the end of seeding (or clearing once the initializer is done) would close it.

**3. `wrap_with_presolve_provider` doesn't check `is_presolve_wrapper`** — only `wrap_with_presolve` does. Worth mirroring for symmetry.

**4. Decorators don't forward `is_presolve_wrapper`.** `CountingTnlp`, `SeededTnlp`, and the ℓ₁ wrapper all take the default `false`, so "presolve wrapper with a decorator on top" still double-wraps. Not reachable today (the CLI uses the explicit flag), but the trait method's doc comment should tell decorator authors to forward it to their inner TNLP.

**5. CHANGELOG covers the warm-start change but not the headline feature.** Nothing under `Added` for "presolve now applies to every library solve" or for the new public API — `set_presolve_already_applied`, `optimize_tnlp_without_presolve`, `TNLP::is_presolve_wrapper`.

**6.** A one-line comment noting that `v_l` / `v_u` are deliberately not forwarded (no user-facing TNLP surface for slack bound multipliers) is still worth adding to `get_starting_z`.

My read: mergeable once #1 is settled one way or the other; the rest is polish. Thanks for turning these around so quickly.

---

GermanHeim:
Will fix tomorrow!

jkitchin (commented):
All six items from my last pass are addressed — `finish_warm_start()` scoping the snapshot (with a test asserting the call count goes to 2 afterwards), `InvalidProblemDefinition` for a refused starting point, the `wrap_with_presolve_provider` symmetry plus its `Rc::ptr_eq` test, the decorator-forwarding doc on `is_presolve_wrapper`, the `v_l`/`v_u` comment, and the `### Added` CHANGELOG section. Guarding the stats capture and `finalize_via_orig_nlp` behind `solver_status != InvalidProblemDefinition` is a good catch that I hadn't flagged — reading `cq` residuals off a never-initialized iterate would have been reading garbage.

I ran `pounce-nlp`, `pounce-presolve`, `pounce-algorithm`, and `pounce-sensitivity` lib + integration tests locally on `0a74fcf`: all green.

**Two blockers before this can merge, neither visible on GitHub yet** (no check runs have registered for this head):

**1. `cargo fmt --all -- --check` fails.** `crates/pounce-nlp/src/orig_ipopt_nlp.rs:2058` — the second line of the new `get_starting_z` comment is indented with two tab characters instead of spaces:

```
        // TNLP exposes only variable-bound multipliers.
^I^I// Slack-bound v_l/v_u have no user-facing warm-start payload to forward.
```

`c249633` (`cargo fmt`) landed before `2d13a96` added the comment, so it wasn't covered. `ci.yml:46` runs the check, so this will go red. `cargo fmt --all` fixes it.

**2. Merge conflict with `main`.** `main` is now at `ce0fa4a` (#367, #369, #370) and the PR is `dirty`. A trial merge conflicts in **`CHANGELOG.md` only** — both sides added entries under `## [Unreleased]`; `application.rs` and `pounce-cli/src/main.rs` auto-merge cleanly. Keeping both sets of entries is all that's needed.

Two smaller notes, neither blocking:

- **`SolverReturn` isn't `#[non_exhaustive]`**, so adding `InvalidProblemDefinition` is a breaking change for any external matcher on it (e.g. oximo-pounce). In-workspace is fine — `solver_return_to_app_status` got its arm and `resto_inner_solver.rs` only uses `matches!`. Under Cargo's 0.x rules that makes the next release `0.10.0` rather than `0.9.x`, so it's worth a deliberate call rather than a surprise at tag time.
- `InvalidProblemDefinition` now covers *every* `set_initial_iterates` failure, not just a refused starting point — `init/default.rs:241` (`curr` is `None`) reaches it too. That one is structural so the status still reads correctly, but if a numerical failure is ever added to that path it would get mislabeled as an invalid problem.

---

Issue #366, opened 2026-07-25
Reconcile: 5 pyomo-pounce tests reported failing on main, but CI runs the same suite green


jkitchin:
Two signals about `main` disagree, and one of them is wrong. Worth settling either way, because one of the five is a correctness test.

**I have not reproduced this** — the Rust extension is not built in the environment I reviewed from, so everything below is from the reports plus reading CI config. Filing to get it resolved, not asserting the tests are broken.

## The disagreement

Across two separate comments on #357, the contributor reported the suite as `120 passed, 5 failed` and then `121 passed, 5 failed`, describing the five as pre-existing on `main`:

- three in `pyomo-pounce/tests/test_binary_check.py`
- `test_bound_multipliers_populate_ipopt_zL_zU` (`tests/test_pounce.py:118`)
- `test_multiplier_gradient_matches_finite_difference` (`tests/test_sens.py:57`)

Meanwhile the `wheel smoke (pyomo-pounce)` job runs the whole suite and passes. `ci.yml:244`:

```yaml
python -m pytest pyomo-pounce/tests -q
```

No `continue-on-error`, no `|| true`, no `--deselect`, no `-k` filter. It went green on every push to #357 and on #364.

## Hypothesis (unverified)

A bundled-vs-PATH binary mismatch. CI stages the freshly built CLI into the wheel before installing:

```yaml
- name: Stage CLI binary into pounce-solver wheel
  run: |
    mkdir -p python/pounce/bin
    cp target/release/pounce python/pounce/bin/pounce
```

so `bundled` is true there. A local dev setup — `cargo build` plus the binary on `PATH`, no staging — has it resolvable but **not** bundled. That fits the three `test_binary_check.py` failures being the resolution/shadowing/warning tests, and it fits `test_binary_check.py` already carrying `pytest.skip("no bundled pounce binary in this environment")` guards at lines 36 and 181 — the pattern exists but evidently does not cover every case.

That hypothesis does not obviously explain the other two, which is part of why this needs a real diagnosis rather than a guess.

## Why it is worth resolving

`test_multiplier_gradient_matches_finite_difference` is a **correctness** test in `sens.py` — the module #357 just changed. If it genuinely fails on `main`, that matters on its own. If it is environment noise, it should skip cleanly with a reason.

The broader cost is that a suite which is normally red locally stops being read. Five expected failures is exactly the number at which people stop checking whether it is still five.

## Suggested resolution

1. Get the actual failure output from a local run (contributor has it).
[...]

PR #368, opened 2026-07-25
Fix test_binary_check.py on Windows

devin-griff:
Closes the Windows half of #366, built on top of ff21a07 (cherry-picked, so the two land together without a merge).

Three tests in `test_binary_check.py` were POSIX-only and fail on every Windows checkout regardless of binary staleness, which is the missing piece of the #366 puzzle: they are green in CI (Linux) forever and red on Windows forever, so they read as "pre-existing failures on main" from the one machine that ran them.

- `test_check_binary_flags_a_shadowing_build` joined `PATH` with a hardcoded `":"`; Windows needs `os.pathsep`.
- `test_build_id_keeps_dirty_suffix` (and the fake-binary helper generally) wrote extensionless `#!/bin/sh` scripts, which Windows cannot execute, so `_build_id` probed them to `None`.
- `test_all_path_pounce_finds_binary` planted a fake named `pounce`, which the scan never resolves on Windows since it looks for `pounce.exe` — as `test_all_path_pounce_resolves_via_which` itself documents.

Fixes keep POSIX behavior byte-identical:

- The fake helper writes `.bat` files on Windows where the path is probed directly; `subprocess` runs those without a shell.
- Where PATH resolution is exercised, the stand-in is a copy of the real binary, since a fabricated `.exe` cannot print a chosen `--about`. For the shadowing test, the fake's build id is injected at the seam `check_binary` reads through; the comparison logic under test is untouched, and the subprocess probe keeps its own coverage through the `.bat` fakes.

Verified on Windows 11 with the CLI and extension built from this branch: `test_binary_check.py` 10 passed, full pyomo-pounce suite **115 passed, 0 failed** — the first fully green Windows run. The two multiplier failures reported in #357 were the stale binary exactly as diagnosed in #366; details going to that issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -227,6 +227,22 @@ integer-only seeds the solver is bit-for-bit reproducible, and the #358 work in
 fact improved that family substantially (92 → 38 failures/144) before #361 took
 it to 0.
 
+
+### ### Tests — pyomo-pounce: `test_binary_check.py` failed on any Windows checkout (#366)
+
+- Three tests were POSIX-only: the shadowing test joined `PATH` with a
+  hardcoded `":"` instead of `os.pathsep`, and the fake `pounce` binaries were
+  extensionless shell scripts, which Windows can neither execute (so
+  `_build_id` probed them to `None`) nor resolve (the scan looks for
+  `pounce.exe`). They failed on every Windows machine and passed on Linux,
+  which is why CI never saw them and why they were misread as pre-existing
+  failures on `main`.
+- Fakes are `.bat` files on Windows where the path is probed directly; where
+  PATH resolution is exercised, the stand-in is a copy of the real binary,
+  with the shadowing fake's build id injected at the seam `check_binary`
+  reads through, since a fabricated `.exe` cannot print a chosen `--about`.
+
+
 ## [0.9.0] - 2026-07-24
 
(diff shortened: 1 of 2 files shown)


devin-griff:
Extended with the two guards from the latest #366 reproduction, so the three known configurations land as one change:

- **Windows** (this PR originally): `os.pathsep`, `.bat` fakes, real-binary stand-ins where a fabricated `.exe` cannot exist.
- **Source checkout, no wheel**: `ff21a07`, carried here since #364 turned out not to include it.
- **Foreign `pounce` on PATH, no bundled binary**: the shadowing test now skips when the resolved binary has no queryable build id — the mechanism under test is build-id discrimination, so a binary without one can neither shadow nor be shadowed meaningfully. And `test_sens.py` gains a module-level guard requiring the build under test (bundled, or a PATH binary with a queryable id), instead of failing every solve with "Solver (pounce) did not exit normally".

On Windows with a current staged binary: 126 passed, 0 failed, and the guards trigger nothing. The `mv`-the-binary reproduction script from the issue should now come back all-skips on the affected tests rather than five failures.

Not included: the documentation item (that `pyomo-pounce/tests` expects the staged bundled binary) — that reads like it belongs near the Makefile's existing staleness commentary rather than in a test-file PR.

Issue #362, opened 2026-07-25
covariance() does not project a fitted variable pinned by an active constraint


devin-griff:
`covariance()`'s active-bound projection tests a fitted parameter against the NL variable bounds only:

```python
lo, hi = np.asarray(session.nl.x_l), np.asarray(session.nl.x_u)
...
if xv - lo[r] < tol or hi[r] - xv < tol:
```

A variable held at the same value by an active *constraint row* is never detected, so it is treated as free. The reported variance is then whatever the barrier leaves, small but nonzero, rather than the conditional zero plus the nonstandard-asymptotics warning.

The same limit written two ways, five residuals pulling `A` well above a cap of 1.0:

```
A <= cap as a variable bound      std_err = 0.0
A <= cap as a Constraint          std_err = 7.915413472554487e-06
```

`A*` is `1.0000000099` in both. Only the first is projected.

Neither number is unreasonable in isolation, but a user who writes a cap the second way gets no signal that the asymptotics at that point are nonstandard, which is the whole purpose of the existing warning.

This is not new behavior and is unrelated to #357, where it surfaced. Verified by disabling that PR's bound rewrite: the variable-bound spelling gives `0.0` with the rewrite on and off.

The fix wants an activity classifier that covers constraint rows as well as variable bounds, and then a decision about what a caller gets back for a row-pinned variable, which is the same question the bound-pinned case answers. `pounce.classify_working_set` already returns a `(bounds, constraints)` pair, so the primitive exists. A fix will be proposed in a coming update to the covariance and information roadmap rather than patched in place here.

jkitchin:
@devin-griff are you working on this? I can pick it up if you aren't but didn't want to duplicate your work if you are.

devin-griff:
@jkitchin yes, it is part of the updated roadmap proposed here: https://github.com/jkitchin/pounce/pull/262

I would appreciate your feedback on that when you have some time, in particular items 0 and 1. I think implementation (which I am happy to do) will be straightforward once the design is correct.

PR #371, opened 2026-07-26
Add post-solve activity classification to Solver (roadmap item 0, #362)

devin-griff:
_Description corrected after merge to match what actually landed: the row formula gained its fourth power, the `μ > 1e-4` wording was wrong about the geometry, and the contamination rule became μ-relative. The original text is preserved in the review thread below._

Implements item 0 of the covariance/information roadmap (#262): post-solve activity classification for bounded variables and inequality rows, closing the mechanism gap behind #362.

## What

`Solver.classify_activity()` (Rust core: new `pounce_sensitivity::activity` module) classifies every bounded variable and every finite-bounded inequality row of the held converged solve on the ratio of barrier curvature to the model's own curvature:

```
r = Σ / q,   Σ = z/s summed over the sides that exist
             q = |H_ii|                 (variable)
                 |∇dᵀH∇d| / ‖∇d‖⁴       (inequality row)
```

`r` is O(μ) when the bound is inactive, O(1) when weakly active, O(1/μ) when strongly active, so one ratio separates the regimes where no fixed threshold on a slack or multiplier alone can (both are O(√μ) at weak activity). Edges at √μ and 1/√μ, fixed inner band [1e-1, 1e1], `ambiguous` in the gaps; the μ-edges meet the band at μ = 1e-2 and a full decade separates them from it at μ = 1e-4, so above 1e-4 — where that margin is thinning — only the two calls that stay clear are made and the middle is honest refusal. Curvature below `√ε·max(1, max_j|H_jj|)` reports `unidentified` with the sign of the raw value.

The row denominator carries the **fourth** power so that `r` is invariant to rescaling the row: `d → c·d` sends `Σ → Σ/c²` while the curvature along the unit normal is unchanged. Equivalently, the geometric barrier weight `Σ‖∇d‖²` (the distance to the surface is `d/‖∇d‖`, its conjugate multiplier `v‖∇d‖`) is measured against the curvature along the unit normal. Variable bounds are invariant as written. This also absorbs the solver's own per-row `d_scale`, so `nlp_scaling_method` cannot move a status either.

Per-entry runtime flags: `off_central_path` (s·z differs from μ by more than 10x on some side) and `contaminated` (classified inactive yet `r > 100μ` — `inactive` MEANS `r = O(μ)`, so the threshold is μ-relative; a fixed constant could never sit below the inactive edge `√μ` at a converged μ). The call refuses to run with `bound_relax_factor != 0` (the Ipopt default is 1e-8): relaxed bounds shift the slacks the classifier reads.

## Why rows too

That is #362's shape: move a variable bound onto a linear inequality row and the activity disappears from the bound-multiplier view entirely, silently breaking any covariance/activity heuristic keyed on z alone. Rows classify through the same rule via the curvature along the constraint normal, and the tests require the two formulations of the same scalar geometry to classify identically — at any row coefficient.

## Implementation notes

- Everything the classifier reads was already retained at convergence: `Σ` (`curr_sigma_x`/`curr_sigma_s`), the solver's own slacks, the bound multipliers on the iterate, `μ`, and the exact Lagrangian Hessian (`curr_exact_hessian`, so `H` is never recovered from the barrier-augmented factor). Item 0 is exposure plus the rule, not new computation.
- `H_ii` comes from one pass over the Hessian triplet structure. Row curvature gathers each Jacobian row's support in one triplet pass and takes the product over that support against a Hessian adjacency view, so a row costs its own support times its neighbours rather than a full mat-vec pair; the mat-vec loop survives as the fallback for any future non-triplet matrix type. Compressed bound-side vectors scatter to full length through the existing `ExpansionMatrix` machinery.
- The report is indexed in **user space**: `var_*` follows the user's `n` variables and `row_*` the user's `m` constraints, in their order. A variable removed by `fixed_variable_treatment = make_parameter` (`lb == ub`) reports `fixed` at its own index and an equality row reports `equality`, so user indices never shift.
- `BoundClassification` gained a `full_to_c` inverse map, making `full_g_to_c_block` O(1) for every caller rather than a linear scan of `c_map`.
- One crate-internal accessor on `PdSensBacksolver` hands the classifier the retained handles; no public surface changes besides `Solver::classify_activity()` and a `SolverError::BadOptions` variant (the enum is now `#[non_exhaustive]`).
- Python: `Solver.classify_activity()` returns a dict (statuses as strings, ratios/signs as ndarrays); raises `ValueError` on nonzero `bound_relax_factor`, `RuntimeError` before a converged solve.

## Tests

`python/tests/test_activity.py`: the scalar study model min ½x² − px with x ≥ 0 walks all three regimes by moving the unconstrained minimizer (p>0 inactive, p<0 strongly active, p=0 weakly active, where slack and multiplier are both O(√μ)); the same geometry as an inequality row classifies identically (the #362 agreement case) and the now-unbounded variable reports `unbounded`; the row cases repeat at coefficients 1, 100 and 1000 plus a `nlp_scaling_method=none` case pinning r = 1.0 exactly, so scale invariance is enforced rather than assumed; a linear objective on a box reports `unidentified`; ratio magnitudes track μ, 1, and 1/μ on the central path; a near-bound inactive case drives `contaminated` true end to end; a mixed model pins the user-space indexing (two-sided bound, fixed variable, equality plus an active row in one report); the `bound_relax_factor` precondition and the not-converged path raise. Rust unit tests cover `classify()`, `off_path`, the contamination rule and the identification floor directly.

Refs #262 (roadmap, item 0), #362. Items 1-4 follow as separate PRs in roadmap order. No closing keyword by intent: item 0 classifies; the `covariance()` fix for #362 is item 1, so the issue stays open until item 1 lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,11 @@ changes.
   of the held converged solve as `inactive`, `weakly_active`,
   `strongly_active`, `ambiguous`, or `unidentified`, from the ratio `r = Σ/q`
   of barrier curvature (`Σ = z/s`, summed over the sides that exist) to the
-  model's own curvature (`q = |H_ii|` for a variable, curvature along the
-  constraint normal for a row) at the converged iterate. `r` is O(μ)
+  model's own curvature (`q = |H_ii|` for a variable; for a row,
+  `|∇dᵀH∇d|/‖∇d‖⁴`, whose fourth power makes the ratio invariant to
+  rescaling the row and to the solver's per-row scaling, so both
+  spellings of one limit classify identically at any row coefficient)
+  at the converged iterate. `r` is O(μ)
   inactive, O(1) weakly active, O(1/μ) strongly active, so one ratio
   separates the regimes where no fixed threshold on a slack or multiplier
   alone can: both are O(√μ) at weak activity. Edges sit at √μ and 1/√μ with
@@ -35,7 +38,8 @@ changes.
   reports `equality`, so user indices never shift.
 - Per-entry honesty flags on variables and rows alike: `off_central_path`
   (`s·z` differs from `μ` by more than 10× on some side) and `contaminated`
-  (classified inactive yet carrying non-negligible barrier curvature).
+  (classified inactive yet `r > 100μ`: `inactive` means `r = O(μ)`, so
+  the threshold is μ-relative and reachable).
   `classify_activity` refuses to run with `bound_relax_factor != 0` (the
   Ipopt default is `1e-8`): relaxed bounds shift the slacks the classifier
   reads.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,36 @@ changes.
 
 ## [Unreleased]
 
+### Added — post-solve activity classification on `Solver` (#362, covariance roadmap item 0)
+
+- `Solver.classify_activity()` (Rust: `pounce_sensitivity::activity`)
+  classifies every bounded variable and every finite-bounded inequality row
+  of the held converged solve as `inactive`, `weakly_active`,
+  `strongly_active`, `ambiguous`, or `unidentified`, from the ratio `r = Σ/q`
+  of barrier curvature (`Σ = z/s`, summed over the sides that exist) to the
+  model's own curvature (`q = |H_ii|` for a variable, curvature along the
+  constraint normal for a row) at the converged iterate. `r` is O(μ)
+  inactive, O(1) weakly active, O(1/μ) strongly active, so one ratio
+  separates the regimes where no fixed threshold on a slack or multiplier
+  alone can: both are O(√μ) at weak activity. Edges sit at √μ and 1/√μ with
+  a fixed inner band [1e-1, 1e1]; gaps report `ambiguous`; at μ > 1e-4 only
+  the two clear calls are made. Curvature below `√ε·max(1, max_j|H_jj|)`
+  reports `unidentified`, with the sign of the raw value.
+- Inequality rows classify through the same rule as variable bounds, which
+  is the gap behind #362: a bound moved onto a row disappears from the
+  bound-multiplier view but not from this one. The tests walk the same
+  scalar geometry through both formulations and require identical statuses.
+- Per-entry honesty flags: `off_central_path` (`s·z` differs from `μ` by
+  more than 10× on some side) and `contaminated` (classified inactive yet
+  carrying non-negligible barrier curvature). `classify_activity` refuses
+  to run with `bound_relax_factor != 0` (the Ipopt default is `1e-8`):
+  relaxed bounds shift the slacks the classifier reads.
+- Item 0 of `dev-notes/covariance-information-roadmap.md` (#262).
+  Everything the classifier reads was already retained at convergence
+  (`Σ`, the solver's slacks, the bound multipliers, `μ`, and the exact
+  Lagrangian Hessian), so the change is exposure plus the rule, not new
+  computation. Items 1-4 build on these statuses.
+
 ### Fixed — active-set SQP: a warm start was discarded whenever the active set moved (#428)
 
 - `solve_with_working_set` pins the hinted active rows to their new
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,11 +28,17 @@ changes.
   is the gap behind #362: a bound moved onto a row disappears from the
   bound-multiplier view but not from this one. The tests walk the same
   scalar geometry through both formulations and require identical statuses.
-- Per-entry honesty flags: `off_central_path` (`s·z` differs from `μ` by
-  more than 10× on some side) and `contaminated` (classified inactive yet
-  carrying non-negligible barrier curvature). `classify_activity` refuses
-  to run with `bound_relax_factor != 0` (the Ipopt default is `1e-8`):
-  relaxed bounds shift the slacks the classifier reads.
+- The report is indexed in user space: `var_*` follow the user's `n`
+  variables and `row_*` the user's `m` constraints, in their order. A
+  variable removed internally by `fixed_variable_treatment = make_parameter`
+  (`lb == ub`) reports `fixed` at its own index and an equality constraint
+  reports `equality`, so user indices never shift.
+- Per-entry honesty flags on variables and rows alike: `off_central_path`
+  (`s·z` differs from `μ` by more than 10× on some side) and `contaminated`
+  (classified inactive yet carrying non-negligible barrier curvature).
+  `classify_activity` refuses to run with `bound_relax_factor != 0` (the
+  Ipopt default is `1e-8`): relaxed bounds shift the slacks the classifier
+  reads.
 - Item 0 of `dev-notes/covariance-information-roadmap.md` (#262).
   Everything the classifier reads was already retained at convergence
   (`Σ`, the solver's slacks, the bound multipliers, `μ`, and the exact
--- a/crates/pounce-nlp/src/orig_ipopt_nlp.rs
+++ b/crates/pounce-nlp/src/orig_ipopt_nlp.rs
@@ -2271,10 +2271,12 @@ impl IpoptNlp for OrigIpoptNlp {
     fn full_g_to_c_block(&self, full_idx: Index) -> Option<Index> {
         let cls = self.adapter.borrow();
         let cls = cls.classification();
-        cls.c_map
-            .iter()
-            .position(|&g_idx| g_idx == full_idx)
-            .map(|p| p as Index)
+        let f = full_idx as usize;
+        if f >= cls.full_to_c.len() {
+            return None;
+        }
+        let c = cls.full_to_c[f];
+        if c < 0 { None } else { Some(c) }
     }
 
     fn var_x_to_full_x(&self, var_idx: Index) -> Index {
--- a/crates/pounce-nlp/src/tnlp_adapter.rs
+++ b/crates/pounce-nlp/src/tnlp_adapter.rs
@@ -88,6 +88,9 @@ pub struct BoundClassification {
     pub d_l_map: Vec<Index>,
     /// Subset of `[0, n_d)` with a finite upper bound.
     pub d_u_map: Vec<Index>,
+    /// Maps full-g index → c-block position, with `-1` for inequality
+    /// rows: the O(1) inverse of `c_map`, mirroring `full_to_var`.
+    pub full_to_c: Vec<Index>,
 }
 
 impl BoundClassification {
@@ -412,6 +415,7 @@ fn classify_bounds(
     let mut d_map: Vec<Index> = Vec::new();
     let mut d_l_map: Vec<Index> = Vec::new();
     let mut d_u_map: Vec<Index> = Vec::new();
+    let mut full_to_c: Vec<Index> = vec![-1; ng];
 
     for i in 0..ng {
         let lo = g_l[i];
@@ -435,6 +439,7 @@ fn classify_bounds(
                 ));
             }
             if lo == hi {
+                full_to_c[i] = c_map.len() as Index;
                 c_map.push(i as Index);
                 continue;
             }
@@ -468,6 +473,7 @@ fn classify_bounds(
         d_map,
         d_l_map,
         d_u_map,
+        full_to_c,
     })
 }
 
--- a/crates/pounce-py/src/solver.rs
+++ b/crates/pounce-py/src/solver.rs
@@ -382,26 +382,36 @@ impl PySolver {
 
     /// Classify every bounded variable and every finite-bounded
     /// inequality row of the held solve by activity, from the ratio of
-    /// barrier curvature to the objective's own curvature at the
+    /// barrier curvature to the model's own curvature (the exact
+    /// Lagrangian Hessian, so constraint curvature contributes) at the
     /// converged iterate (covariance-information-roadmap item 0,
     /// pounce#362).
+    ///
+    /// Indexing is **user space**: `var_*` entries follow your `n`
+    /// variables and `row_*` entries your `m` constraints, in your
+    /// order. A variable fixed by its bounds (`lb == ub`) reports
+    /// `"fixed"` (the solve removes it and classifies no barrier
+    /// geometry for it), and an equality constraint reports
+    /// `"equality"`, so indices never shift.
     ///
     /// Returns a dict:
     ///
     /// - `"mu"`: the converged barrier parameter.
     /// - `"var_status"`, `"row_status"`: list of str per variable /
-    ///   inequality row: `"inactive"`, `"weakly_active"`,
-    ///   `"strongly_active"`, `"ambiguous"`, `"unidentified"`, or
-    ///   `"unbounded"` where there is no finite bound.
+    ///   constraint: `"inactive"`, `"weakly_active"`,
+    ///   `"strongly_active"`, `"ambiguous"`, `"unidentified"`,
+    ///   `"unbounded"` where there is no finite bound, plus the
+    ///   `"fixed"` / `"equality"` placeholders above.
     /// - `"var_ratio"`, `"row_ratio"`: ndarray of the ratio `Σ/q`
-    ///   (NaN where unbounded).
+    ///   (NaN where nothing was classified).
     /// - `"var_q_sign"`, `"row_q_sign"`: ndarray of the sign of the
     ///   signed curvature, so an indefinite direction is visible.
     /// - `"var_off_central_path"`, `"row_off_central_path"`: list of
     ///   bool, true where `s·z` differs from `μ` by more than 10× on
     ///   some side.
-    /// - `"var_contaminated"`: list of bool, true where classified
-    ///   inactive yet carrying non-negligible barrier curvature.
+    /// - `"var_contaminated"`, `"row_contaminated"`: list of bool,
+    ///   true where classified inactive yet carrying non-negligible
+    ///   barrier curvature.
     ///
     /// Requires `bound_relax_factor=0` (raises `ValueError` otherwise;
     /// the Ipopt default is `1e-8`): relaxed bounds shift the slacks
@@ -415,12 +425,15 @@ impl PySolver {
             codes
                 .iter()
                 .map(|&c| match c {
+                    activity::UNBOUNDED => "unbounded",
                     activity::INACTIVE => "inactive",
                     activity::WEAKLY_ACTIVE => "weakly_active",
                     activity::STRONGLY_ACTIVE => "strongly_active",
                     activity::AMBIGUOUS => "ambiguous",
                     activity::UNIDENTIFIED => "unidentified",
-                    _ => "unbounded",
+                    activity::FIXED => "fixed",
+                    activity::EQUALITY => "equality",
+                    _ => unreachable!("unknown activity status code {c}"),
                 })
                 .collect()
         };
@@ -449,6 +462,7 @@ impl PySolver {
                 .into_pyarray_bound(py),
         )?;
         out.set_item("row_off_central_path", rep.row_off_central_path)?;
+        out.set_item("row_contaminated", rep.row_contaminated)?;
         Ok(out)
     }
 }
--- a/crates/pounce-py/src/solver.rs
+++ b/crates/pounce-py/src/solver.rs
@@ -22,6 +22,7 @@ use numpy::{IntoPyArray, PyArray1};
 use pounce_common::types::{Index, Number};
 use pounce_nlp::return_codes::ApplicationReturnStatus;
 use pounce_nlp::tnlp::TNLP;
+use pounce_sensitivity::activity;
 use pounce_sensitivity::{Solver as RustSolver, SolverError};
 use pyo3::exceptions::{PyRuntimeError, PyValueError};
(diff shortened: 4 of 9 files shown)


jkitchin:
is this a WIP? can you make it a draft if so, or let me know when it ready to consider for review?

jkitchin:
# Review: post-solve activity classification (roadmap item 0, #362)

## Overview

Adds `Solver::classify_activity()` (new `pounce_sensitivity::activity` module, ~316 lines) classifying each bounded variable and finite-bounded inequality row of a converged solve as inactive / weakly / strongly active / ambiguous / unidentified, via the ratio `r = Σ/q` of barrier curvature to model curvature. Exposed to Python as a dict; guarded by a `bound_relax_factor=0` precondition with a new `SolverError::BadOptions` variant. 673 additions, no deletions, well-scoped.

The core scheme is sound and the implementation is careful in the right places — I checked the details that typically go wrong here, and they're correct:

- `curr_sigma_x`/`curr_sigma_s` are full-length (x-/s-space) while slacks and bound multipliers are compressed — the module treats them exactly that way (`ipopt_cq.rs:675`, `ipopt_cq.rs:277`).
- The scoped-borrow discipline in `compute()` is right: `curr_jac_d`/`curr_exact_hessian` internally take `nlp.borrow_mut()` (`ipopt_cq.rs:484-494`), and the module drops its `data`/`nlp` borrows before any Cq call, holding only the (separate) `cq` RefCell.
- The `values_mut()`-to-materialize-homogeneous-output trick matches `DenseVector`'s actual semantics (`dense_vector.rs:131-144`).
- Pyo3 style (`new_bound`/`into_pyarray_bound`) matches the rest of `pounce-py`.

## Issues

**1. Report indexing is internal-space, not user-space — undocumented (most important).** `n = curr.x.dim()` is the *internal* x dimension. Pounce implements Ipopt's default `fixed_variable_treatment = make_parameter` (`tnlp_adapter.rs:58`), so any variable with `lb == ub` is removed from internal x. For such models, `var_status` is shorter than the user's `n` and every index after a fixed variable is shifted — silently. Similarly `row_status` is indexed by internal inequality-row order, not user constraint index. Either resort back to user space (the analogue of `IpTNLPAdapter::ResortX`) or, at minimum, document the indexing and the fixed-variable caveat in the Python docstring; right now a user with fixed variables gets misaligned answers with no warning.

**2. Hessian diagonal costs O(n·nnz).** `hessian_diag_entry` does one full sparse mat-vec per variable, over *all* `n` (the floor needs the global max, but that doesn't require n mat-vecs). For n = 10⁵ with a modest Hessian this is prohibitive for what is conceptually a single pass over the sparse structure. Suggest extracting the diagonal directly (downcast to the concrete sparse type, or add a `diag()`/`max_abs_diag()` to `SymMatrix`). Fine to defer with a TODO, but item 0 is the foundation for items 1–4, so the cost will be inherited.

**3. A bounded row with a zero Jacobian row reports `unbounded`.** In the row loop, `if norm2 <= 0.0 { continue; }` leaves `row_status[j] = UNBOUNDED` even though `rhas_l||rhas_u` is true. A constant (all-zero-gradient) row genuinely has finite bounds; `UNIDENTIFIED` is the honest label, `unbounded` is wrong.

**4. PR is not mergeable** against current main (almost certainly the CHANGELOG, which moves on every merge). Needs a rebase.

## Smaller points

- **Silent downcast fallbacks:** `expand()`/`present()` return all-zeros/all-false if the projection matrix isn't an `ExpansionMatrix`. Today it always is (`orig_ipopt_nlp.rs:316`), but if that ever changes, every entry silently reports `unbounded`. Prefer `expect()` — loud breakage over quiet lies. Same for the `k < compressed.len()` bounds guard, which would mask a size mismatch.
- **Hardcoded `1e-8` fallback duplicates the registered default.** `get_numeric_value` returns the registered default with `found=false` (`options_list.rs:293-296`), so `.map(|(v, _)| v)` would track the option registry instead of a second hardcoded copy that can drift.
- **Comment vs. math on the `μ > 1e-4` branch:** the edges √μ / 1/√μ close inside the band `[1e-1, 1e1]` at μ ≥ 1e-2, not 1e-4. For μ ∈ (1e-4, 1e-2) the code conservatively refuses to report `weakly_active` even though the regions are still disjoint. The behavior is defensible; the comment (and CHANGELOG wording) claiming the edges "have closed" in that range is not accurate.
- **Python status mapping catch-all:** `_ => "unbounded"` maps any unexpected code to a legitimate-looking status. Match `activity::UNBOUNDED` explicitly and make the wildcard unreachable.
- **Asymmetry:** variables get `contaminated`, rows don't. Add `row_contaminated` (the computation is identical) or document why it's variable-only.
- **Doc wording:** several docstrings say "the objective's own curvature," but `q` comes from the exact *Lagrangian* Hessian (`eval_h(x, 1.0, y_c, y_d)`), so constraint curvature contributes. The module doc and PR body say "model's own curvature" — align on that.

## Tests

Good: the three-regime walk, the bound-vs-row agreement case (the actual #362 shape), the zero-curvature floor, and both error paths. Gaps:

- `classify()` is a pure function but has no Rust unit tests — band edges, the ambiguous gaps, and the μ > 1e-4 branch are exactly the kind of boundary logic that deserves direct coverage rather than being reachable only through full solves.
- No test ever produces `ambiguous`, `off_central_path=True`, or `contaminated=True` — the honesty flags are only ever asserted false.
- All models are single-variable; nothing exercises mixed statuses in one report, two-sided bounds, or the fixed-variable case from issue 1 above (which would have caught it).
- Nit: `test_variable_bound_ratio_scales` runs `_solve_bound(1.0)` twice; `test_relaxed_bounds_are_refused` doesn't assert the solve succeeded.

## Conventions

CHANGELOG updated in-style. One repo-convention check: the PR body says "closing the mechanism gap behind #362" and "Refs #262/#362" but contains **no closing keyword** — if merging this should close #362, the body needs a literal `Closes #362`/`Fixes #362`, or the issue will dangle after merge (as happened with PR #342).

## Verdict

Solid, well-reasoned contribution — approve direction. Before merge: the rebase (4), the indexing caveat documented or resolved (1), and the zero-Jacobian row mislabel fixed (3). The performance issue (2) is acceptable to defer with a note since the roadmap builds on this surface.

---

devin-griff:
HI @jkitchin , the intent is to break down the larger covariance/information roadmap (https://github.com/jkitchin/pounce/pull/262) into some bite sized chunks, this PR just being item zero of that roadmap. I will address the issues above and update, but yes the intent is that this scope should be mergable on its own.

devin-griff:
All addressed, in 7dbba83. Point by point:

1. (indexing) Resorted to user space rather than documented: `var_*` follows the user's n and `row_*` the user's m, in user order. A make_parameter-removed variable reports `"fixed"` at its own index, an equality row `"equality"`, so indices never shift. The new mixed-model test pins a fixed variable mid-vector and would have caught the shift.
2. (diagonal cost) Extracted in one pass over the `SymTMatrix` triplet structure (1-based indices, duplicates accumulate to match `mult_vector`); the e_i mat-vec loop survives only as a fallback for any future non-triplet `SymMatrix`, so no deferred TODO.
3. (zero-gradient row) Reports `unidentified` with `q_sign` 0, same as below-floor curvature.
4. Rebased onto current main.

Smaller points all taken: loud `expect()`s on the expansion downcasts plus an exact length assert; `bound_relax_factor` reads the registry default instead of a second hardcoded copy; the mu > 1e-4 comment now states the real geometry (the edges meet the band at mu = 1e-2; the branch refuses the weak call above 1e-4, where the margin thins) - the CHANGELOG never carried the inaccurate phrasing, only the code comment did; explicit `"unbounded"` arm with an unreachable wildcard; `row_contaminated` added, the asymmetry had no justification; curvature wording aligned on the Lagrangian.

Tests: `classify()`, `off_path`, the contamination rule and the floor now have direct Rust unit tests (band edges inclusive, both gaps, the mu > 1e-4 branch; exactly-on-a-mu-edge is float-fragile, so the tests sit either side of it); a solve-based `ambiguous` case via `mu_target`; the mixed model above (two-sided bound, fixed variable, equality plus an active row in one user-space report); both nits fixed. One honest limit: `off_central_path=True` and `contaminated=True` from a genuinely converged solve could not be forced deterministically (a converged IPM sits on the central path by construction; the flags exist to catch broken assumptions), so those two are asserted true at the unit level only.

On the closing keyword: intentional that there is none. Item 0 classifies; the `covariance()`-side fix is item 1 of the roadmap, so #362 should stay open until item 1 lands. The PR body now says so.

The unit tests caught one real bug in the review response itself: the one-pass diagonal first read the triplet indices as 0-based, and pounce triplets are 1-based.

devin-griff:
Rebased onto current main (post-#421); all review points from your pass are addressed in the earlier response commit, and the suite is green. Ready for a second look whenever it fits, @jkitchin — items 1-4 of the roadmap queue behind this one.

jkitchin:
# Review: post-solve activity classification (roadmap item 0, #362) — second pass

## Overview

Adds `Solver::classify_activity()` (new `pounce_sensitivity::activity`, 495 lines) classifying every bounded variable and finite-bounded inequality row of a converged solve from `r = Σ/q`, plus a Python dict wrapper, a `SolverError::BadOptions` guard on `bound_relax_factor`, Rust unit tests for the pure rules, and a Python integration suite. 965 additions, no deletions.

The response to the first pass is genuinely thorough — I re-verified each fix against the branch rather than taking it on trust:

- **The user-space resort is correct.** `c_map`/`d_map` are built by a single ascending scan (`tnlp_adapter.rs:411-445`), so the "the j-th full-g index outside the c-block is internal row j" assumption in the scatter loop holds, and the `assert_eq!(d_pos, m_d)` pins it. The `FIXED` default for unmapped variables is right under both `fixed_variable_treatment` modes (`RelaxBounds` keeps the variable in `x_not_fixed_map`, so it never falls through to the default).
- **The triplet diagonal is correct.** Indices are 1-based (`triplet.rs:6`) and duplicates accumulate, matching `mult_vector`. The loud `expect()`s and the exact length assert on the expansion downcasts are the right call.
- Borrow discipline around the `Cq` getters, the `values_mut()` materialization, and the pyo3 style all check out.

---

## Blocking: the row ratio is not invariant to how a constraint row is scaled

`q_j = |∇dⱼᵀH∇dⱼ| / ‖∇dⱼ‖²` normalizes the curvature, but nothing normalizes `Σ_j`. Rescale a row `d → c·d` (identical geometry, different units): the slack goes `s → c·s` and its multiplier `v → v/c`, so `Σ = v/s → Σ/c²` while `q` is invariant. **The ratio moves by `1/c²` and the classification changes with it.**

Variable bounds *are* invariant (`Σ_i` and `H_ii` pick up the same factor), so this breaks exactly the bound-vs-row agreement that is the PR's central claim and #362's whole shape. The tests pass only because every test row has `‖∇d‖ = 1`.

Verified on the branch with this PR's own `ScalarRow` geometry, varying only the row coefficient `c` in `g(x) = c·x ≥ 0` (min ½x² − p·x, `bound_relax_factor=0`, `tol=1e-10`):

```
scaling=none            p=0   c=1      weakly_active    (r = 1.0)
scaling=none            p=0   c=100    ambiguous        (r = 1.0e-4)
scaling=none            p=0   c=1000   INACTIVE         (r = 1.0e-6)   <- exactly-active bound called inactive
scaling=none            p=-1  c=1000   ambiguous        (r = 1.0e5)    <- strongly_active at c=1
scaling=gradient-based  p=0   c=100    ambiguous        (r = 1.0e-4)
scaling=gradient-based  p=0   c=1000   ambiguous        (r = 1.0e-4)
```

`nlp_scaling_method=gradient-based` (the default) caps the damage at `nlp_scaling_max_gradient` but does not remove it — and it adds a second, independent source of the same error: pounce applies per-row `d_scale` (`orig_ipopt_nlp.rs:1044`, `apply_d_scale_to_bounds`), so `r̃_row = r_row / dg_j²` even for a row the user wrote in natural units.

**Fix:** normalize `Σ` by the same power. Geometrically, the distance to the surface is `d/‖∇d‖` and the conjugate multiplier is `v‖∇d‖`, so `Σ_geom = Σ‖∇d‖²`, giving

```
q_j = |∇dⱼᵀ H ∇dⱼ| / ‖∇dⱼ‖⁴        (equivalently  r_j = Σ_j‖∇dⱼ‖⁴ / |∇dⱼᵀH∇dⱼ|)
```

This reduces to the current formula whenever `‖∇d‖ = 1`, so every existing test still passes, and it returns exactly `r = 1.0` for the weakly-active case at every `c` under both scaling methods (check it against the table: `1.0e-4 × 1e4 = 1.0`). Worth a test that parameterizes the row coefficient — that is the assertion the current suite is missing.

## `contaminated` can never fire on a converged solve

`CONTAMINATION_FLOOR = 1e-3` is a fixed constant, but `INACTIVE` at `μ ≤ 1e-4` requires `r < √μ`. Contamination therefore needs `√μ > 1e-3`, i.e. `μ > 1e-6` — above any converged barrier parameter at default tolerances (the probe's inactive case sits at `r ≈ 1e-11`). That is why it "could not be forced deterministically": the flag is structurally dead, not merely awkward to arrange.

Since `inactive` means `r = O(μ)`, the honest test is μ-relative — something like `r > 100·μ` — which *is* reachable and testable end to end. `off_path` already gets this right by keying on `μ`.

## Performance: two quadratic loops

- The row loop does one `jac_d.trans_mult_vector` **and** one `hess.mult_vector` per bounded row → `O(m_d · (nnz(Jd) + nnz(H)))`. The first review's fix removed exactly this pattern for the Hessian diagonal, but the rows kept it. Both are extractable in one pass over the triplet structures (gather the `Jd` rows once; `∇dⱼᵀH∇dⱼ` still needs a product, but only over the row's support).
- `full_g_to_c_block` is a linear scan of `c_map` (`orig_ipopt_nlp.rs:2274`), called once per full-g index in the scatter → `O(m · n_c)`. Build the `d_map` membership once instead.

Neither matters at test scale; both bite on the models items 1–4 will be pointed at.

## Smaller points

- `q = |H_ii|` reads only the Hessian diagonal, so off-diagonal coupling is invisible: `f = x₁x₂` with bounds reports `unidentified` on every bound despite well-defined curvature in the bound directions. Defensible for item 0, but worth a sentence in the module doc since items 1–4 inherit the semantics.
- For `unidentified`, the reported ratio is `Σ/floor`, not `Σ/q` — a lower bound, not the ratio. The Python docstring says only "NaN where nothing was classified"; say what the value means.
- `SolverError` is not `#[non_exhaustive]`, so `BadOptions` is a breaking change for any exhaustive downstream match. Fine at 0.x, but adding the attribute now would save a future bump.
- The `bound_relax_factor` guard reads `app.options()` at classify time, not what the held solve actually ran with — changing the option after `solve()` flips the guard without changing the state it guards. Snapshotting the value into `ConvergedState` would make the precondition mean what it says.
- The guard rejects the *default* value, so `classify_activity()` raises out of the box for anyone who has not opted in. Since pounce already snapshots the declared (unrelaxed) `d_L`/`d_U`, un-relaxing the slacks is a plausible follow-up; at minimum the message could say "re-solve with `bound_relax_factor=0`".
- The PR body still says the μ-edges "have closed inside the band" at `μ > 1e-4` — the code comment was corrected to say they meet at `μ = 1e-2`; the body was not.
- No user-facing docs entry for the new Python method (there is no `.pyi` for `pounce` to update, so the docstring is the only surface).

## Tests

The Rust unit tests for `classify` / `off_path` / `contaminated` / the floor are the right addition, and the mixed model is a good user-space regression (fixed variable mid-vector, equality before inequality, two-sided bound). The gaps map onto the findings above: nothing varies row scaling, nothing varies `nlp_scaling_method`, and `contaminated=True` is asserted only at unit level because it cannot be reached from a solve.

## Verdict

Careful work, and the mechanism is right for variables. The row branch needs the `‖∇d‖⁴` normalization before merge — as it stands the row classification tracks the units a constraint was written in, which is the one property #362 needs it not to have. The contamination floor and the two quadratic loops are worth fixing in the same pass; everything else is polish.

---

devin-griff:
All taken, in cf96a31:

**Blocking.** Rows classify on Σ‖∇d‖⁴/|∇dᵀH∇d|. The module doc carries the geometric reading, Σ‖∇d‖² as the invariant barrier weight against curvature along the unit normal, and notes the same normalization absorbs the per-row `d_scale`. Tests: the three regimes × c ∈ {1, 100, 1000} under default scaling, plus a sharp variant with `nlp_scaling_method=none` at c = 1000 asserting r = 1.0 exactly, your table's `inactive` misfile.

**Contamination.** μ-relative, r > 100μ, inactive only; the fixed floor is gone and the doc says why it was structurally dead. A unit test pins that the threshold sits inside the inactive region whenever μ < 1e-4, and an end-to-end test (bound 0.01 from the optimum, r ≈ 10⁴μ) asserts the flag true from a genuinely converged solve.

**Performance.** The row loop gathers Jd's rows in one triplet pass (duplicates merged to match `mult_vector`) and computes ∇dᵀH∇d over each row's own support against a both-directions Hessian adjacency view; the mat-vec pair survives only as the fallback for non-triplet types, mirroring the diagonal fix. For the scatter, the defect was in `full_g_to_c_block` itself, so the fix lands there: a `full_to_c` inverse map in `BoundClassification`, built beside the existing `full_to_var`, makes the accessor O(1) for every caller rather than special-casing the classifier's loop.

**Smaller.** The module doc states the diagonal-only semantics with the `f = x1*x2` example and that items 1-4 inherit them (their reduced-block classification is where coupling becomes visible); the `unidentified` ratio is documented as Σ/floor, a lower bound rather than the ratio, in both the Rust docs and the Python docstring; `SolverError` is `#[non_exhaustive]`, with the wildcard arm pounce-py then needs.

Suites: workspace green, Python activity tests 23 (from 12), full python suite 604 passed / 14 skipped, pyomo-pounce 147 passed.

jkitchin:
# Verification of `cf96a31` — second-pass findings confirmed fixed

Checked each fix against the branch rather than the commit message.

## Blocking finding: fixed

The row denominator now carries `‖∇d‖⁴`. Re-ran the probe from the second pass — same geometry `x ≥ 0` written as the row `c·x ≥ 0` — extended past the added tests' range to `c = 10⁶`:

| case | before | after |
|---|---|---|
| `p=0,  c=1000,  scaling=none` | **`inactive`** (r = 1.0e-6) | `weakly_active` (r = 1.0000) |
| `p=0,  c=10⁶,   scaling=none` | — | `weakly_active` (r = 1.0000) |
| `p=-1, c=1000,  scaling=none` | **`ambiguous`** (r = 1.0e5) | `strongly_active` (r = 1.0e11) |
| `p=0,  c=100,   gradient-based` | **`ambiguous`** (r = 1.0e-4) | `weakly_active` (r = 1.0000) |

Every status holds across four decades of row scaling under both `nlp_scaling_method` settings, and the weakly-active ratio pins at 1.0 rather than drifting as `1/c²`. `test_row_classification_is_scale_invariant` plus the `scaling=none` sharp case cover this in-tree.

## Adversarial check on the new one-pass row loop

The triplet gather is fresh index arithmetic, so I ran it on a coupled model — `H = [[1,1],[1,2]]`, two multi-entry Jacobian rows. The row `x₀ − x₁ ≥ −100` has `∇dᵀH∇d = 1 − 2 + 2 = 1`, a *negative* off-diagonal contribution, so a missing or double-counted mirror moves the answer by 2×:

```
measured   row_ratio = 3.49375e-15
analytic   mu/107^2 * ||grad d||^4 / (grad d^T H grad d)
         = 8.7344e-16 * 4 / 1  =  3.4937e-15
```

Exact. The `if a != b` mirror condition in the adjacency build matches `SymTMatrix::mult_vector` (`triplet.rs:526`) precisely, and the 1-based decrement is right on both the Jacobian and Hessian passes.

## Remaining findings

- **Contamination floor** — `r > 100·μ` now sits below the inactive edge `√μ` at every converged μ, so the flag is reachable instead of structurally dead. `test_near_bound_inactive_flags_contamination` (bound 0.01 from the optimum, `r = μ/s² ≈ 10⁴μ`) exercises it end to end, which the old fixed `1e-3` made impossible.
- **Quadratic loops** — both gone. The Jacobian supports and the Hessian adjacency each build in one triplet pass, with the mat-vec pair correctly retained as the non-triplet fallback. `full_g_to_c_block` picking up a `full_to_c` inverse map in `BoundClassification` fixes the `O(m·n_c)` scan for every caller, not just this one — good generalization of the point.
- **Smaller points** — `#[non_exhaustive]` on `SolverError` with the wildcard arm it requires; the `Σ/floor` meaning of an `unidentified` ratio documented in both languages; diagonal-only semantics of `q = |H_ii|` stated in the module doc with a forward note on where items 1-4 recover coupling.

`cargo test -p pounce-sensitivity` is green (57 unit tests plus all integration binaries) and all five CI checks pass on `cf96a31`.

## Still open — all minor, none blocking

- The `bound_relax_factor` guard still reads `app.options()` at classify time rather than a value snapshotted into `ConvergedState`, so changing the option after `solve()` flips the guard without changing the state it guards.
- No user-facing docs entry; the docstring remains the only surface.
- The PR **body** still says the μ-edges "have closed inside the band" at `μ > 1e-4` — the code comment was corrected to `μ = 1e-2` two passes ago, but the description was not.

Nothing here blocks merge. Note the intentional absence of a closing keyword: #362 stays open until roadmap item 1 lands.

---

Issue #372, opened 2026-07-27
Trivially infeasible bounded Pyomo NLP returns RestorationFailed/internalSolverError


bernalde:
## Summary

A trivially infeasible bounded NLP terminates as `RestorationFailed` in POUNCE
0.8.0 and 0.9.0. Through the AMPL/Pyomo path this becomes
`SolverStatus.error` / `TerminationCondition.internalSolverError`, even though
the contradiction is visible directly in the model:

```text
0.0 <= x <= 0.6
x >= 0.7
```

IPOPT 3.14.16 reports the identical Pyomo model as `infeasible`
(`Converged to a locally infeasible point`). This looks like an infeasibility
detection/status-classification gap in POUNCE's restoration path rather than a
Pyomo solution-loading issue: POUNCE itself writes `RestorationFailed`, which is
then placed in the AMPL 500 failure range.

## Minimal reproducer

Save as `repro.py`:

```python
from importlib.metadata import version

import pyomo.environ as pyo
import pyomo_pounce  # noqa: F401 - registers SolverFactory("pounce")

model = pyo.ConcreteModel()
model.x = pyo.Var(bounds=(0.0, 0.6), initialize=0.5)
model.objective = pyo.Objective(expr=model.x**3 + model.x**2)
model.impossible = pyo.Constraint(expr=model.x >= 0.7)

solver = pyo.SolverFactory("pounce")
solver.options.update(
    {
        "print_level": 0,
        "tol": 1.0e-10,
        "bound_relax_factor": 0.0,
        "honor_original_bounds": "yes",
        "max_cpu_time": 30.0,
    }
)
result = solver.solve(model, load_solutions=False)

print(f"pyomo={version('pyomo')}")
print(f"pyomo-pounce={version('pyomo-pounce')}")
print(f"pounce-solver={version('pounce-solver')}")
print(f"executable-version={'.'.join(map(str, solver.version()))}")
print(f"solver-status={result.solver.status}")
print(f"termination-condition={result.solver.termination_condition}")
print(f"message={result.solver.message}")
print(f"solutions={len(result.solution)}")
```

Run against the published 0.9.0 wheels:

```bash
uv run --no-project --isolated --python 3.12 \
  --with pyomo==6.10.1 \
  --with pyomo-pounce==0.9.0 \
  --with pounce-solver==0.9.0 \
  --no-build-package pounce-solver \
  repro.py
```

## Observed

```text
pyomo=6.10.1
pyomo-pounce=0.9.0
[...]

PR #392, opened 2026-07-28
Never report infeasibility while holding a feasible point

### LINKED ISSUE #379: NLP path reports Infeasible_Problem_Detected on feasible models that POUNCE's own convex solver solves
## Summary

The NLP path reports `Infeasible_Problem_Detected` (AMPL `solve_result_num` 200) on
models that are feasible — and on the *same* models POUNCE's own convex QP solver
returns `Optimal Solution Found` (0). One solver contradicting another inside the
same binary is the clearest possible statement that the verdict is wrong.

This is the **numerical** restoration path. It is independent of the presolve
certification work in #377, which reports 201 on a different code path and is not
involved here (both routes below reproduce with `presolve=no`).

## How it was found

The property test added in #377,
`pyomo-pounce/tests/test_infeasibility_no_false_positives.py`, generates models
that are feasible **by construction**: a witness point `x*` is chosen strictly
inside the box first, then every constraint is built to hold at `x*`. Any verdict
in the AMPL `200..299` band is therefore a false positive.

Over 400 instances it reports 3. One (seed 223) is the presolve path; the two here
are the numerical path.

## Reproduction

Both models are 2–4 variables, all bounds finite.

```
seed 193                                    seed 294
  x[0] in [0, 1e6]                            x[0] in [0, 1e6]
  x[1] in [-9.716393895220474,                x[1] in [0, 1e6]
           -8.716393895220474]
  x[2] in [0, 1]                            constraint rhs:
  x[3] in [0, 1e6]                            c[1]: type 2, -1e-06
                                              c[2]: type 2, -1.01
constraint rhs:
  c[1]: type 1, -871639389.5220464
  c[2]: type 1, -999989.7836061048
  c[3]: type 1, 1e-12
```

| Route | seed 193 | seed 294 |
|---|---|---|
| `solver_selection=nlp presolve=no` | `InfeasibleProblemDetected` **[200]** | `InfeasibleProblemDetected` **[200]** |
| `solver_selection=nlp presolve=yes presolve_fbbt=yes` | `InfeasibleProblemDetected` **[200]** | `InfeasibleProblemDetected` **[200]** |
| default auto-route (convex QP IPM) | `Optimal Solution Found` **[0]** | `Optimal Solution Found` **[0]** |

### Witness points, from POUNCE's own convex route

Every value is inside its declared bounds:

```
seed 193:  x = (500015, -9.49158, 0.21285, 500015)
seed 294:  x = (500000, 500000)
```

So the feasible region is demonstrably non-empty, by POUNCE's own arithmetic.

## Why this matters beyond the fuzz

`Infeasible_Problem_Detected` maps to AMPL 200, which Pyomo reads as
`TerminationCondition.infeasible`. A caller told a feasible model is infeasible
has no signal that anything went wrong — this fails silently and confidently,
which is worse than an error.

## What did not catch it

Two existing guards target exactly this class and both let it through:

- `crates/pounce-cli/tests/false_local_infeasibility.rs` — the regression suite for
  "POUNCE must not report a feasible model infeasible" (the HS13 cases). It passes;
  these shapes are not covered.
- The CLI's `feral_infeasibility_scaling_retry` — the MC64 re-solve that exists to
  second-guess a local-infeasibility verdict before believing it. It did not
  overturn either verdict.

Both models carry a wide dynamic range (bounds spanning `1e6` alongside a
`1e-9`-wide interval; right-hand sides from `1e-12` to `8.7e8`), which is the
common thread with the other scale-sensitive defects fixed in #373 and #376.

## Suggested next steps

1. Add both shapes to `false_local_infeasibility.rs`.
2. Work out why the MC64 re-solve does not overturn them — that guard is the
   intended backstop and is silently ineffective here.
3. Consider cross-checking against the convex route when a model is convex and the
   NLP path reports infeasibility; the two disagreeing is a detectable condition.

## Notes

- Found by the property test in #377, but **not caused by it** — reproduces with
  `presolve=no`, i.e. with none of that PR's code active.
- The generator is adversarial by design (extreme coefficient scales, near-degenerate
  boxes), so these are not typical models. The contradiction between two of POUNCE's
  own solvers is what makes it worth fixing regardless.

jkitchin:
Fixes #379.

## Problem

POUNCE reported `Infeasible_Problem_Detected` — AMPL `solve_result_num` 200, Pyomo `TerminationCondition.infeasible` — on a model it was **handed a feasible starting point for**.

Only one of the four seeds in the issue still reproduces on `main`. Seeds 99 and 193 (the #376 regressions) are fixed and pinned by `scaled_feasible_not_infeasible.rs`; seed 102 was fixed by the scale-relative tolerance work (#382–#386). Seed 294 remained, and its mechanism is different from all of those.

The model is two variables, both in `[0, 1e6]`, two linear rows, sum-of-squares objective. Its declared starting point is `(5e5, 5e5)`, which satisfies both rows **exactly**. `pounce verify` reports `0.000e0` constraint and bound violation on the convex route's solution.

| route | status | `solve_result_num` |
|---|---|---|
| `solver_selection=nlp presolve=no`, before | `InfeasibleProblemDetected` | **200** |
| `solver_selection=nlp presolve=no`, after | `ErrorInStepComputation` | 500 |
| default auto-route (convex QP IPM) | `Optimal Solution Found` | 0 |

Two of POUNCE's own solvers contradicting each other about whether the feasible set is empty, on a model whose feasibility is witnessed by its own `x0`.

## Cause

Not a threshold. The first row carries `±1e30` coefficients. In the scaled space that row's slack initializes far from anything `x` can follow, so the outer line search walks the violation **up** from an exactly feasible start:

```
iter  0  inf_pr 1.00e-02      <- x0, feasible in user units
iter  1  inf_pr 1.10e-01
iter  3  inf_pr 1.00e+03
iter  9  inf_pr 9.98e+03      alpha_pr 4.88e-04, 12 backtracks
```

Restoration then burns its full 3000-iteration budget and exits `MaxiterExceeded`, and `cycle_locally_infeasible` in `resto_inner_solver.rs` converts that exhaustion into an infeasibility verdict:

```
[PN_RESTO_LOCINF] status=MaxiterExceeded iter=3000 inner_kkt_err=1.000000e5
  orig_inf_pr=1.000000e13 orig_inf_pr_scaled=1.000000e5 outer_tol=1.000000e-8
  strict=false alt=false cycle=true step_fail=false tiny_step=false -> loc_inf=true
```

`inner_kkt_err = 1e5`. The inner never approached stationarity at all — the gate's own comment says as much ("the inner's line search is cycling between basins rather than approaching a stationary point"). Every gate here argues from a *stalled feasibility sub-problem*: violation bounded away from zero, no local move reduces it. That is evidence, not proof, and this is what it looks like when the evidence is wrong.

Note the existing scale-relative guards were all working correctly and did not fire falsely; the defect is upstream of them, in what the `cycle` gate treats as evidence.

## Fix

A refutation rather than a retune. One concrete feasible point settles the question outright, whatever a local argument concluded — and the model's own starting point is right there. New `pounce_algorithm::infeasibility_refutation` evaluates it (clamped into the variable box) against every row and withdraws the verdict if it satisfies them all.

**Applied as one gate over all four claim sites**, not per-site: the IPM path's restoration/cycle gates, the SQP infeasible-subproblem exit, and the ℓ₁ wrapper's uncollapsed-slack certificate. The two preceding safeguards in this area (#376, #380) were each added to one path and not its twin, and a hole survived both times.

One-directional by construction: a model with no feasible point cannot produce a witness, so every correct verdict is untouched, and any failure to evaluate simply declines to refute.

The withdrawn verdict becomes `Error_In_Step_Computation` (AMPL 500) — the status this codebase already uses for "the solve broke down and we are *not* claiming infeasibility"; the `cycle_exit` fallback in `ipopt_alg.rs` picks between exactly these two on exactly this question. The `±1e30` model still fails, because that row is genuinely hostile. It now fails visibly instead of returning a confident wrong answer.

**`is_negligible` for the witness test — rejected, measured.** The first version used the clamped accepting form, matching presolve's `witness_refutes_infeasibility`. `test_scale_invariance.py` caught it before it could ship: it withdrew *correct* infeasibility verdicts on three models (`inf_372`, `inf_clear`, `inf_two`) at row scalings `1e-12 … 1e-8`, turning 0 wrong cells into 2–3 each. Mechanism: `x >= 2` over `x ∈ [0, 1]` multiplied through by `1e-12` has a violation of `2e-12` against a row magnitude of `2e-12` — a full unit violation — but `tol * max(scale, 1)` is `1e-8`, so `x0` read as a witness. The clamp reinstates an absolute floor exactly in the down-scaled direction this gate must not be fooled in. Shipped form is pure relative `tol * scale` (`is_significant`), threshold `2e-20`, verdict stands. Pinned by a 25-decade sweep in both directions.

**Sampling the box (midpoint, corners) — deliberately not done.** Presolve's refutation does this, correctly: it is countering a *proof over the whole box* from interval arithmetic. Here the claim is numerical and the gate runs on every solve ending in the infeasible band, so widening it would change verdicts on models this PR cannot validate against — the benchmark corpus is not in the tree. The starting point needs no such justification. This is a known, documented gap, not an oversight.

**Cross-checking against the convex route (issue's step 3) — subsumed.** The refutation is strictly cheaper (no second solve), more general (works on genuinely nonlinear models, where there is no convex route to disagree), and catches the same case.

**The MC64 re-solve not overturning these (issue's step 2) — not a defect.** That guard exists for *hypersensitivity*: two backward-stable scalings falling into different basins. This failure reproduces under both scalings, so the second opinion correctly agreed. Confirmed by running with `nlp_scaling_method=none` and `feral_scaling=mc64` — same verdict. The refutation now runs first, so the wasted re-solve is skipped entirely.

## Blast radius

Any solve that would have returned `Infeasible_Problem_Detected` from a numerical argument, and only those. A correct verdict cannot be affected: the withdrawal requires a point that satisfies every constraint, which an infeasible model does not have.

Presolve *certificates* (`TNLP::presolve_infeasibility_proof`) are exempt — a proof is not a numerical inference, and it carries its own, tighter refutation.

Cost is one `eval_g` on a path that has already terminated.

Corpus sweep: **not run** — the benchmark `.nl` inputs are gitignored and not present in this checkout, so I cannot make the "bit-identical except the targeted case" claim that a solver change normally warrants. What I can state is measured below. Flagging this explicitly rather than implying coverage I do not have; the property sweeps and the full test suite are the evidence this rests on.

Measured, over 400 feasible-by-construction instances from `test_infeasibility_no_false_positives.py`'s generator:

| configuration | before | after |
|---|---|---|
| `solver_selection=nlp presolve=no` | 1 (seed 294) | **0** |
| property test's full option set | 3 (seeds 223, 294, + 1) | **1** (seed 223) |

Seed 223 is `solve_result_num` 201 on the presolve certification path — #377's territory, untouched here.

## Tests

**`crates/pounce-cli/tests/false_local_infeasibility.rs`** gains both reported shapes as fixtures (`feasible_x0_extreme_row.nl` = seed 294, `feasible_x0_wide_scale.nl` = seed 102), plus a pin that the convex route still solves them, so the two routes cannot drift back into disagreeing.

Verified biting on the parent commit (`91804a4`) by copying the test and fixtures into a worktree at that commit:

```
test feasible_models_with_a_feasible_start_are_not_reported_infeasible ... FAILED
  feasible_x0_extreme_row.nl is feasible — its own starting point satisfies every
  constraint, and POUNCE's convex QP route solves it to optimality — but the NLP
  path reported the AMPL infeasible band (solve_result_num=200,
  status=InfeasibleProblemDetected)
test result: FAILED. 6 passed; 1 failed
```

Two honest qualifications about which halves bite:

- The `feasible_x0_wide_scale.nl` (seed 102) half **passes on the parent**. It was a false positive before #382–#386 and is pinned here as a ratchet, not as a reproducer of this bug.
- `the_convex_route_still_solves_both_shapes` also **passes on the parent**. It pins the contradiction that made the issue unambiguous; it is not a regression test for this fix.

**`crates/pounce-algorithm/src/infeasibility_refutation.rs`** — 9 unit tests. The load-bearing ones: `a_down_scaled_infeasible_row_is_never_refuted` and `a_scaled_feasible_row_is_refuted_at_every_scale` sweep 25 decades in both directions and pin the `is_negligible` mistake so nobody reintroduces it; `genuinely_infeasible_model_is_not_refuted` uses #372's reproducer; `clamping_does_not_manufacture_a_witness` and `infinite_bound_sentinel_does_not_inflate_the_scale` cover the two ways this could fabricate a withdrawal.

**`pyomo-pounce/tests/test_infeasibility_no_false_positives.py`** — `MAX_FALSE_POSITIVES` ratchets 4 → 1. Measured count at `N_INSTANCES = 200` is 0; the 1 is headroom for a machine-dependent flip on a knife-edge instance, not a known failure.

Full workspace suite (`--exclude pounce-hsl`, which needs `COINHSL_DIR` and does not link in this environment): **156 test binaries, exit 0, zero failures**. `cargo fmt --all -- --check` clean; `cargo clippy --all-targets` introduces no new warnings.

---

- [x] Book page under `docs/src/` updated (`solution-output.md`, existing page — no `SUMMARY.md` change needed)

https://claude.ai/code/session_01DyT66Zbb3fZEPQgtEK9tPp

---
_Generated by [Claude Code](https://claude.ai/code/session_01DyT66Zbb3fZEPQgtEK9tPp)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,64 @@ changes.
   - Docs: [Solution output](docs/src/solution-output.md) gains a
     `solve_result_num` band table and the proved-vs-local distinction.
 
+### Fixed — the NLP path reported `Infeasible_Problem_Detected` on models whose own starting point is feasible (#379)
+
+- **No numerical path may now claim infeasibility while holding a point that
+  satisfies every constraint.** The feasible-by-construction property sweep in
+  `pyomo-pounce/tests/test_infeasibility_no_false_positives.py` found a model
+  (seed 294) that POUNCE was *handed a feasible starting point for* — `(5e5, 5e5)`,
+  satisfying both rows exactly — and reported as infeasible: AMPL
+  `solve_result_num` 200, Pyomo `TerminationCondition.infeasible`. POUNCE's own
+  convex QP route solves the same model to optimality, and `pounce verify`
+  reports `0.000e0` constraint and bound violation on that solution.
+  - Root cause is not a threshold. The gates that produce this verdict — the
+    restoration gates, the outer cycle detector, the SQP infeasible-subproblem
+    exit, the ℓ₁ wrapper's uncollapsed-slack certificate — all argue from a
+    *stalled feasibility sub-problem*: the violation is bounded away from zero
+    and no local move reduces it. That is evidence, not proof. On this model the
+    first row carries `±1e30` coefficients; in the scaled space its slack
+    initializes far from anything `x` can follow, the outer line search walks the
+    violation *up* from an exactly feasible start, restoration burns its
+    3000-iteration budget, and the cycle gate concluded infeasibility from the
+    exhaustion.
+  - Fix: a refutation. `pounce_algorithm::infeasibility_refutation` evaluates the
+    model's own starting point, clamped into the variable box, and withdraws the
+    verdict if it satisfies every row. One concrete feasible point settles the
+    question outright, whatever a local argument concluded.
+  - Applied as **one gate over all four claim sites**, not per-site. The two
+    preceding safeguards in this area were each added to one path and not its
+    twin, and a hole survived both times.
+  - One-directional by construction: it can only ever *withdraw* a verdict, never
+    create one. A model with no feasible point cannot produce a witness, so every
+    correct verdict is untouched — and a failure to evaluate simply declines to
+    refute.
+  - The withdrawn verdict becomes `Error_In_Step_Computation` (AMPL 500), the
+    status this codebase already uses for "the solve broke down and we are *not*
+    claiming infeasibility". The solve still fails on the `±1e30` model — that row
+    is genuinely hostile — but it fails visibly instead of returning a confident
+    wrong answer.
+  - The witness test is **pure relative** (`tol · scale`), not the clamped
+    accepting form. Measured: the clamped form withdrew *correct* infeasibility
+    verdicts on three models at row scalings `1e-12 … 1e-8`, because the clamp
+    reinstates an absolute floor exactly in the down-scaled direction. Caught by
+    `pyomo-pounce/tests/test_scale_invariance.py` before it could ship.
+  - Presolve *certificates* are exempt: a proof is not a numerical inference, and
+    it carries its own, tighter refutation.
+  - Measured: false positives on the numerical path (`solver_selection=nlp
+    presolve=no`) over 400 feasible-by-construction instances go **1 → 0**; over
+    the property test's full option set, **3 → 1** (the remaining one is seed 223
+    on the presolve path, unrelated to this fix). `MAX_FALSE_POSITIVES` ratchets
+    from 4 to 1.
+  - Not a bug: the CLI's MC64 second-opinion re-solve did not overturn these. That
+    guard exists for *hypersensitivity* — two backward-stable scalings falling
+    into different basins — and this failure reproduces under both scalings, so
+    the second opinion correctly agreed. The refutation runs before it, so the
+    wasted re-solve is now skipped too.
+  - Regression: `crates/pounce-cli/tests/false_local_infeasibility.rs` gains both
+    reported shapes and pins that the convex route still solves them, so the two
+    routes cannot drift back into disagreeing about whether the feasible set is
+    empty.
+
 ### Fixed — an infeasible model's verdict depended on the user's `tol` (follow-up to #372)
 
 - **A genuinely infeasible model now reports `Infeasible_Problem_Detected` (AMPL
--- a/crates/pounce-algorithm/src/application.rs
+++ b/crates/pounce-algorithm/src/application.rs
@@ -1010,6 +1010,23 @@ impl IpoptApplication {
             ),
         };
 
+        // Same gate as the IPM path: an infeasible-subproblem exit is a
+        // numerical inference, and a feasible starting point disproves it
+        // (gh #379). Only the infeasibility verdict is rewritten — every other
+        // status passes through untouched, so the pair stays in lockstep.
+        let refuted = withdraw_infeasibility_if_refuted(
+            &tnlp,
+            solver_status,
+            self.nlp_lower_bound_inf(),
+            self.nlp_upper_bound_inf(),
+            self.user_tol(),
+        );
+        let (app_status, solver_status) = if refuted == solver_status {
+            (app_status, solver_status)
+        } else {
+            (solver_return_to_app_status(refuted), refuted)
+        };
+
         // Forward to the user TNLP's finalize_solution. We pass
         // the SQP iterate and recovered multipliers via the
         // OrigIpoptNlp's lifting hooks. Failure here is silent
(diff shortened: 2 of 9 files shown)


jkitchin:
Merged `main` in (`edd0fc1`) to pick up #389, which landed in the same area — the DOF gate now consults presolve's bound-propagation certification instead of reporting `NotEnoughDegreesOfFreedom`.

No conflict, textual or semantic. #389's verdict is a *proof* carried on `TNLP::presolve_infeasibility_proof`, which this PR's gate explicitly exempts, and #389's own witness refutation is the tighter one. The two withdrawal mechanisms sit on different paths and do not overlap.

Re-ran the interacting suites on the merge:

| suite | result |
|---|---|
| `false_local_infeasibility` | 7 passed |
| `issue_387_overdetermined_infeasible` (CLI) | 2 passed |
| `presolve_certified_infeasibility` | 8 passed |
| `infeasible_status_tol_invariance` | 3 passed |
| `issue_372_infeasible_bounds_status` | 4 passed |
| `scaled_feasible_not_infeasible` | 1 passed |
| `test_scale_invariance.py` + `test_infeasibility_no_false_positives.py` | 2 passed |

Note #389 lowered `BASELINE_WRONG["inf_eq"]` from 13 to 3; that tightened baseline holds with this change on top, so the refutation does not push any scale-invariance cell back to wrong.

The "verified biting on the parent commit" evidence in the PR body was measured against `91804a4`, which is still an ancestor of this branch — that claim is unchanged.

---

Issue #398, opened 2026-07-28
A one-sided constraint bound beyond ±1e19 is read as a crossed bound pair → Invalid_Problem_Definition on a feasible model


jkitchin:
Split out of #396, where it surfaced. #396's fix stops presolve *certifying* this model infeasible; this is the reason the same model still does not solve.

## Summary

`TNLPAdapter` rejects a model with a legitimate one-sided constraint bound whose magnitude exceeds `nlp_lower_bound_inf` / `nlp_upper_bound_inf` (default `±1e19`), reporting `Invalid_Problem_Definition` — AMPL `solve_result_num` **504**, Pyomo `internalSolverError`.

The model is feasible. `pounce check-x0` measures its declared starting point at exactly zero violation, and the convex QP route solves it to optimality.

## Mechanism

`crates/pounce-nlp/src/tnlp_adapter.rs:392`:

```rust
} else if lo > hi {
    return Err(SolverException::new(
        ExceptionKind::INCONSISTENT_BOUNDS, …
```

For a `<=`-only row the `.nl` reader fills the absent lower bound with the sentinel `-1e19`. If the row's genuine upper bound is more negative than that sentinel, the pair reads as crossed:

```
row c[3]:  -1e30·x[0] <= -5.0000000000000007e20
           g_l = -1e19   (sentinel, absent)
           g_u = -5.0000000000000007e20   (real)
           -1e19 > -5e20  ->  INCONSISTENT_BOUNDS
```

Nothing is actually inconsistent. `-5e20` is a perfectly ordinary finite upper bound; it is only "beyond infinity" under a **symmetric** magnitude reading of the sentinel. The correct reading is directional — a lower bound is absent at or below `-INF_BOUND`, an upper bound at or above `+INF_BOUND` — which is the convention `pounce_presolve::bound_tighten` already uses (`x_l[j] <= -INF_BOUND`, `x_u[j] >= INF_BOUND`).

#396 fixed exactly this confusion inside presolve's witness gate. The adapter has the same confusion, independently.

## Reproduction

Fixture is already in the tree as of #396: `crates/pounce-cli/tests/fixtures/feasible_x0_sentinel_bound.nl` (property-test seed 223).

```sh
pounce crates/pounce-cli/tests/fixtures/feasible_x0_sentinel_bound.nl \
  -AMPL --sol-output out.sol solver_selection=nlp presolve=no print_level=0
[...]

PR #399, opened 2026-07-28
Never certify infeasibility from an absent-bound sentinel

### LINKED ISSUE #396: Presolve certifies infeasibility (201) on a feasible model whose x0 POUNCE itself measures at exactly zero violation
Tracking issue for the last false positive left in the feasible-by-construction property sweep. #392 took the *numerical* path to zero over 400 instances; this one is on the **presolve certification** path, which #392 deliberately did not touch (a proof is not a numerical inference, and it carries its own refutation).

Filing rather than folding into #392 because it is a different mechanism, and because `MAX_FALSE_POSITIVES` in `pyomo-pounce/tests/test_infeasibility_no_false_positives.py` currently sits at **1** purely to accommodate this — it can ratchet to 0 once this is fixed.

## Summary

Seed 223 of the property generator is feasible by construction. With presolve on, POUNCE reports:

```
POUNCE 0.9.0: InfeasibleProblemDetected (detected by presolve: bound propagation)
objno 0 201
```

`201` means **proved**, which is a stronger claim than the numerical `200` — and this model has a feasible point.

## The model is feasible, by POUNCE's own arithmetic

`pounce check-x0` on the model's declared starting point, evaluated with POUNCE's own evaluator:

```
  x0 vs bounds:
    violations: 0  on-bound components: 2

  initial constraint violation:
    rows violated: 0  max violation: 0.000e0
```

Not "within tolerance" — **exactly zero**, on both bounds and rows. The convex route agrees:

| route | status | `solve_result_num` |
|---|---|---|
| `presolve=yes presolve_fbbt=yes …` (property test's option set) | `InfeasibleProblemDetected (detected by presolve: bound propagation)` | **201** |
| `solver_selection=nlp presolve=no` | `InvalidProblemDefinition` | 504 (see below) |
| default auto-route (convex QP IPM) | `Optimal Solution Found` | 0 |

## Why this is surprising

`pounce_presolve::witness_refutes_infeasibility` exists for exactly this case, is applied as a single gate over both certification paths, and evaluates **the model's own starting point first** — precisely because sampling geometry alone was known to miss witnesses at extreme coefficient scale. Here that starting point is exactly feasible and the verdict still escapes.

So something upstream of the tolerance question is going on. Two candidates, neither verified — I did not want to assert a root cause I hadn't confirmed:

1. The gate is not reached on this path (an early return, or the proof is recorded at a site that bypasses it).
2. The gate is reached but sees different inputs than `check-x0` does — `inner_x_l`/`inner_x_u` after Phase-1 tightening rather than the declared box, or `eval_g` declining.

Worth checking (1) first: the console emits `pounce: FBBT detected infeasibility (witness constraint 0)` while the report attributes the proof to *bound propagation*, which suggests more than one detection site is live on this model.

## What makes the model hostile

`check-x0` reports the Jacobian spanning **1e-320 to 1e30 — about 350 decades**, with `1e-320` being subnormal (roughly four significant digits of mantissa left). All three rows are exactly active at the witness:

- `c[1]`: coefficients `-1e-320` on `x[1]`, `x[2]`; `<= 5.2104e-320`. Subnormal on both sides.
- `c[2]`: `1.8056e17*x[0] - 1e18*x[2] >= 1.29139404770787e18`.
- `c[3]`: `-1e30*x[0] <= -5.0000000000000007e20`, i.e. `x[0] >= 5e-10`, against `x[0] ∈ [0, 1e-9]`.

Zero-slack active rows plus subnormal arithmetic is the regime where outward-rounded interval arithmetic loses the precision its conclusion rests on. That is the documented reason the refutation gate exists.

## Secondary finding, possibly separate

The same model with `presolve=no` returns **AMPL 504 `InvalidProblemDefinition`** on the numerical path. That is also wrong — the model is well-formed and feasible — but it falls outside the `200..299` band, so the property test does not flag it and #392's refutation never sees it. May share the 350-decade root cause, may be independent. Splitting it out is fine if it turns out unrelated.

## Reproduction

Self-contained `.nl` (property generator seed 223, `_build(223)` from `pyomo-pounce/tests/test_infeasibility_no_false_positives.py`):

```
g3 1 1 0	# problem unknown
 3 3 1 0 0 	# vars, constraints, objectives, ranges, eqns
 0 1 0 0 0 0	# nonlinear constrs, objs; ccons: lin, nonlin, nd, nzlb
 0 0	# network constraints: nonlinear, linear
 0 3 0 	# nonlinear vars in constraints, objectives, both
 0 0 0 1	# linear network variables; functions; arith, flags
 0 0 0 0 0 	# discrete variables: binary, integer, nonlinear (b,c,o)
 5 3 	# nonzeros in Jacobian, obj. gradient
 4 4	# max name lengths: constraints, variables
 0 0 0 0 0	# common exprs: b,c,o,c1,o1
C0	#c[1]
n0
C1	#c[2]
n0
C2	#c[3]
n0
O0 0	#o
o54	# sumlist
3	# (n)
o5	#^
o0	#+
v0	#x[0]
n-5e-10
n2
o5	#^
o0	#+
v1	#x[1]
n3.9190147700811693
n2
o5	#^
o0	#+
v2	#x[2]
n1.2913940476175876
n2
x3	# initial guess
0 5e-10	#x[0]
1 -3.9190147700811693	#x[1]
2 -1.2913940476175876	#x[2]
r	#3 ranges (rhs's)
1 5.2104e-320	#c[1]
2 1.29139404770787e+18	#c[2]
1 -5.0000000000000007e+20	#c[3]
b	#3 bounds (on variables)
0 0.0 1e-09	#x[0]
0 -3.9190147705811693 -3.9190147695811692	#x[1]
0 -6.291394047617588 3.7086059523824124	#x[2]
k2	#intermediate Jacobian column lengths
2
3
J0 2	#c[1]
1 -1e-320
2 -1e-320
J1 2	#c[2]
0 1.8056493234311766e+17
2 -1e+18
J2 1	#c[3]
0 -1e+30
G0 3	#o
0 0
1 0
2 0
```

```sh
pounce f223.nl -AMPL --sol-output a.sol \
  solver_selection=nlp presolve=yes presolve_fbbt=yes presolve_auxiliary=yes \
  presolve_linear_eq_reduction=yes presolve_redundant_constraint_removal=yes print_level=0
grep objno a.sol          # -> objno 0 201

pounce check-x0 f223.nl   # -> rows violated: 0  max violation: 0.000e0
```

## Suggested next steps

1. Establish which of the two candidates above is actually happening — instrument `witness_refutes_infeasibility` on this model and see whether it runs and what it computes.
2. If it runs and passes the rows, the tolerance form is the suspect. Note #392 found the mirror of this on the numerical side: the clamped `is_negligible` reinstates an absolute floor for `scale < 1`, which is wrong in the down-scaled direction. Presolve's refutation still uses `is_negligible`. That is the *accepting* direction so the clamp is defensible there, but it is worth re-deriving rather than assuming, especially against subnormal row magnitudes.
3. Decide whether the `presolve=no` → 504 result is the same bug or its own issue.
4. Once fixed, drop `MAX_FALSE_POSITIVES` to 0 and consider vendoring this model as a fixture next to `feasible_x0_extreme_row.nl`.

## Scope

Presolve is off by default, so a user on defaults never sees the 201. It requires `presolve=yes`. That bounds the practical impact — but `201` asserts a *proof*, and this one is false, which makes it the more serious of the two codes to get wrong.

Found by the property test from #377; the numerical half of the original three was fixed in #392 (#379).

jkitchin:
Fixes #396.

## Problem

`solve_result_num` **201** claims a *proof* — the strongest thing POUNCE says. This one was false.

Property-test seed 223 is feasible by construction. POUNCE's own `check-x0` measures its declared starting point at **exactly** zero violation, and the convex QP route solves it to optimality:

```
  x0 vs bounds:                   violations: 0
  initial constraint violation:   rows violated: 0  max violation: 0.000e0
```

| route | status | `solve_result_num` |
|---|---|---|
| `presolve=yes presolve_fbbt=yes`, before | `InfeasibleProblemDetected (detected by presolve: bound propagation)` | **201** |
| `presolve=yes presolve_fbbt=yes`, after | `InvalidProblemDefinition` | 504 (see *Still open*) |
| default auto-route (convex QP IPM) | `Optimal Solution Found` | 0 |

## Cause

`witness_refutes_infeasibility` exists to stop exactly this, is applied as a single gate over both certification paths, and evaluates the model's own starting point first. I confirmed it **was** reached and **did** run — there is no early return between the certification site (`lib.rs:985`) and the gate. It simply could not find the witness sitting in front of it.

POUNCE stores an absent constraint bound as the sentinel `±INF_BOUND` (1e19), not as an infinity, so every use of a bound has to ask whether it is real first. The gate asked that question **twice, differently**, and got both wrong on the same row — `c[3]`, `-1e30·x[0] <= -5.0000000000000007e20`. Instrumented, at `x0`:

```
row 2: g=-5.0000000000000007e20  g_l=-1e19  g_u=-5.0000000000000007e20
       viol=4.9e20   scale=5.0e20   negligible=false
```

`g` and `g_u` are identical — the row sits *exactly* on its only real bound, true violation zero. Yet:

- The violation term was `(g_l - v).max(v - g_u)` with **no** presence test, so it scored `g_l - v = 4.9e20` against a lower bound the row does not have.
- The magnitude term used a **symmetric** test, `|b| < INF_BOUND`, which called the row's genuine upper bound `-5e20` absent and discarded a real constraint.

Either error alone sinks the witness. With no candidate able to pass, the certification escaped.

## Fix

One **directional** predicate, matching the convention `pounce_presolve::bound_tighten` already uses (`x_l[j] <= -INF_BOUND`, `x_u[j] >= INF_BOUND`): a lower bound is absent below `-INF_BOUND`, an upper bound above `+INF_BOUND`. The witness gate was the odd one out.

Deriving the magnitude *and* the violation from the same predicate is the load-bearing part — two spellings of one question on adjacent lines is what produced the defect, so the fix has to remove the possibility rather than correct both spellings.

Strictly one-directional, like the gate it sits in: it can only ever *withdraw* a verdict, never create one.

**A symmetric presence test on the violation term only.** My first attempt added the existing `|b| < INF_BOUND` test to `viol`, leaving the magnitude side alone. It fixes seed 223's headline symptom — 201 disappears — but for the wrong reason: it makes row `c[3]` *vacuously* satisfied by discarding its real upper bound, rather than satisfied because the point meets it. A unit test written against that version exposed it (`present(g_u) = false` for a bound POUNCE should honour), which is what led to the directional form. Worth recording because the symptom-level check does not distinguish the two.

## Blast radius

Only solves that would have reported a presolve-certified infeasibility, and only those where a candidate point satisfies every row. A model with no feasible point cannot produce a witness, so correct verdicts are untouched — confirmed by the must-still-detect tests below.

The `.nl` reader's sentinel convention itself is unchanged; this is a read-side fix inside one predicate.

Corpus sweep: **not run** — the benchmark `.nl` inputs are gitignored and absent from this checkout, same gap as #392. The property sweeps and the full suite are what this rests on.

Measured over 400 feasible-by-construction instances:

| configuration | before | after |
|---|---|---|
| full presolve option set | 1 (seed 223) | **0** |
| `solver_selection=nlp presolve=no` | 0 | **0** |

`MAX_FALSE_POSITIVES` ratchets **1 → 0**. That is a floor, not a target: a single false positive means a feasible-by-construction model is being called infeasible.

## Tests

**`crates/pounce-presolve/src/lib.rs`** — two unit tests on `witness_refutes_infeasibility` directly.

`absent_bound_sentinel_does_not_manufacture_a_violation` is the reproducer. Verified biting on the parent commit (`3a511fb`) by grafting it into a worktree at that commit:

```
test tests::absent_bound_sentinel_does_not_manufacture_a_violation ... FAILED
  x0 = 5e-10 puts the row exactly on its only real bound — that is a witness, and
  the -1e19 sentinel standing in for the absent lower bound must not turn it into
  a 4.9e20 violation
test result: FAILED. 232 passed; 1 failed
```

`a_real_bound_beyond_the_sentinel_still_blocks_the_witness` guards the direction the fix could plausibly have broken — admitting the bound must not make real violations invisible. It **passes on the parent** (via the phantom lower-bound violation rather than the real upper-bound one), so it is a direction-guard, not a reproducer; its doc comment says so.

**`crates/pounce-cli/tests/presolve_certified_infeasibility.rs`** — `a_feasible_model_is_never_certified_infeasible` runs the real model end to end, under both `presolve=yes` and `presolve=no`. It asserts the AMPL *band*, not a specific code, deliberately — see below.

Must-still-detect, all passing: `presolve_proves_the_contradiction_and_says_so`, `float_noise_infeasibility_is_never_certified`, `user_declared_crossed_box_is_still_rejected`, `interval_overflow_is_never_mistaken_for_a_proof`, `genuinely_infeasible_problem_is_still_detected`, `contradictory_equalities_report_infeasible_band` (#389's).

Full workspace suite (`--exclude pounce-hsl`, which needs `COINHSL_DIR` and does not link here): **158 test binaries, exit 0, zero failures**. `test_scale_invariance.py` and `test_infeasibility_no_false_positives.py` both pass. `cargo fmt --all -- --check` clean; `cargo clippy --all-targets` adds no new warnings.

## Still open on this model — #398

It does not solve. It now returns `Invalid_Problem_Definition` (504), because `TNLPAdapter` reads the same out-of-range bound as a crossed constraint pair: `g_l = -1e19` against `g_u = -5e20` trips its `lo > hi` check at `tnlp_adapter.rs:392`. That is the **identical** symmetric-sentinel confusion, living independently in the adapter, and it affects any model with a one-sided constraint bound beyond `±1e19`.

Filed as #398 rather than fixed here: it is a different crate and a different mechanism, and changing bound classification for every model is not something this diff should smuggle in. The regression test asserts the band rather than a code precisely so it does not need rewriting when #398 lands.

I did not audit for further copies of the predicate. Two spellings turned up on one line here, so more are plausible; #398 notes it.

---

- [ ] Book page under `docs/src/` updated — not needed; no user-facing behaviour is newly *documented*, and `solution-output.md`'s existing description of `201` as a proof is what this restores rather than changes

https://claude.ai/code/session_01DyT66Zbb3fZEPQgtEK9tPp

---
_Generated by [Claude Code](https://claude.ai/code/session_01DyT66Zbb3fZEPQgtEK9tPp)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,47 @@ changes.
   - Docs: [Solution output](docs/src/solution-output.md) gains a
     `solve_result_num` band table and the proved-vs-local distinction.
 
+### Fixed — presolve *certified* a feasible model infeasible, from an absent-bound sentinel (#396)
+
+- **A model with a feasible point is no longer reported `solve_result_num` 201.**
+  `201` claims a *proof*, which is the strongest thing POUNCE says; this one was
+  false. The model (property-test seed 223) is feasible by construction, and
+  `pounce check-x0` measures its own declared starting point at **exactly** zero
+  violation — `rows violated: 0, max violation: 0.000e0` — while the convex QP
+  route solves it to optimality.
+  - Root cause is the absent-bound sentinel, mishandled in **both directions on
+    the same row**. POUNCE stores an absent constraint bound as `±1e19`
+    (`INF_BOUND`), not as an infinity, so every use has to ask whether a bound is
+    real. `witness_refutes_infeasibility` asked twice, differently, and got both
+    wrong for row `-1e30·x[0] <= -5.0000000000000007e20`:
+    - The violation term was `(g_l - v).max(v - g_u)` with **no** presence test,
+      so it scored `4.9e20` against a lower bound the row does not have — at a
+      point whose true violation is zero.
+    - The magnitude term used a **symmetric** test, `|b| < INF_BOUND`, which
+      called the row's genuine upper bound `-5e20` absent and discarded a real
+      constraint.
+  - Either error alone sinks the witness, so no candidate point could ever
+    refute the verdict, and the certification escaped.
+  - Fix: one **directional** predicate, matching the convention
+    `pounce_presolve::bound_tighten` already uses (`x_l <= -INF_BOUND`,
+    `x_u >= INF_BOUND`) — a lower bound is absent below `-INF_BOUND`, an upper
+    bound above `+INF_BOUND` — derived once and used for both the magnitude and
+    the violation, so the two cannot drift apart again.
+  - Strictly one-directional, like the gate it sits in: it can only ever
+    *withdraw* a verdict. Every certified-infeasibility test still passes,
+    including the must-still-detect cases.
+  - **The property sweep is now at zero.** Over 400 feasible-by-construction
+    instances, false positives in the AMPL infeasible band go `1 → 0` on the full
+    presolve option set (and stay at 0 on the numerical path from #379).
+    `MAX_FALSE_POSITIVES` ratchets `1 → 0` — a floor, not a target.
+  - **Still open on this model:** it does not solve. It now returns
+    `Invalid_Problem_Definition` (504), because `TNLPAdapter` reads the same
+    out-of-range bound as a crossed constraint pair — `g_l = -1e19` against
+    `g_u = -5e20` trips the `lo > hi` check. That is a separate defect in the
+    sentinel convention for constraint bounds, affecting any model with a
+    one-sided bound beyond `±1e19`; it is tracked on its own. The regression test
+    here asserts the *band*, not a specific code, for that reason.
+
 ### Fixed — the NLP path reported `Infeasible_Problem_Detected` on models whose own starting point is feasible (#379)
 
 - **No numerical path may now claim infeasibility while holding a point that
--- a/crates/pounce-cli/tests/fixtures/feasible_x0_sentinel_bound.nl
+++ b/crates/pounce-cli/tests/fixtures/feasible_x0_sentinel_bound.nl
@@ -1,0 +1,61 @@
+g3 1 1 0	# problem unknown
+ 3 3 1 0 0 	# vars, constraints, objectives, ranges, eqns
+ 0 1 0 0 0 0	# nonlinear constrs, objs; ccons: lin, nonlin, nd, nzlb
+ 0 0	# network constraints: nonlinear, linear
+ 0 3 0 	# nonlinear vars in constraints, objectives, both
+ 0 0 0 1	# linear network variables; functions; arith, flags
+ 0 0 0 0 0 	# discrete variables: binary, integer, nonlinear (b,c,o)
+ 5 3 	# nonzeros in Jacobian, obj. gradient
+ 4 4	# max name lengths: constraints, variables
+ 0 0 0 0 0	# common exprs: b,c,o,c1,o1
+C0	#c[1]
+n0
+C1	#c[2]
+n0
+C2	#c[3]
+n0
+O0 0	#o
+o54	# sumlist
+3	# (n)
+o5	#^
+o0	#+
+v0	#x[0]
+n-5e-10
+n2
+o5	#^
+o0	#+
+v1	#x[1]
+n3.9190147700811693
+n2
+o5	#^
+o0	#+
+v2	#x[2]
+n1.2913940476175876
+n2
+x3	# initial guess
+0 5e-10	#x[0]
+1 -3.9190147700811693	#x[1]
+2 -1.2913940476175876	#x[2]
+r	#3 ranges (rhs's)
+1 5.2104e-320	#c[1]
+2 1.29139404770787e+18	#c[2]
+1 -5.0000000000000007e+20	#c[3]
+b	#3 bounds (on variables)
+0 0.0 1e-09	#x[0]
+0 -3.9190147705811693 -3.9190147695811692	#x[1]
+0 -6.291394047617588 3.7086059523824124	#x[2]
+k2	#intermediate Jacobian column lengths
+2
+3
+J0 2	#c[1]
+1 -1e-320
+2 -1e-320
+J1 2	#c[2]
+0 1.8056493234311766e+17
+2 -1e+18
+J2 1	#c[3]
+0 -1e+30
+G0 3	#o
+0 0
+1 0
+2 0
--- a/crates/pounce-cli/tests/presolve_certified_infeasibility.rs
+++ b/crates/pounce-cli/tests/presolve_certified_infeasibility.rs
@@ -65,6 +65,64 @@ fn solve(tag: &str, extra: &[&str]) -> String {
     std::fs::read_to_string(&sol).expect("read .sol")
 }
 
+fn solve_fixture(model: &str, tag: &str, extra: &[&str]) -> String {
+    let sol = std::env::temp_dir().join(format!("pounce_presolve_cert_{tag}.sol"));
+    let _ = std::fs::remove_file(&sol);
+
+    let out = Command::new(pounce_exe())
+        .arg(fixture(model))
+        .arg("-AMPL")
+        .arg("--sol-output")
+        .arg(&sol)
+        .arg("print_level=0")
+        .args(extra)
+        .output()
+        .expect("spawn pounce");
+
+    assert_eq!(out.status.code(), Some(0), "-AMPL must exit 0");
+    std::fs::read_to_string(&sol).expect("read .sol")
+}
+
+/// **gh #396.** A *proof* of infeasibility is the strongest claim POUNCE makes;
+/// it must never be issued for a model that has a feasible point.
+///
+/// `feasible_x0_sentinel_bound.nl` is property-test seed 223, feasible by
+/// construction. `pounce check-x0` measures its declared starting point at
+/// **exactly** zero violation — `rows violated: 0, max violation: 0.000e0` — and
+/// the convex QP route solves it to optimality. Presolve reported
+/// `solve_result_num` 201, "detected by presolve: bound propagation".
+///
+/// The witness gate that exists to stop precisely this was mishandling the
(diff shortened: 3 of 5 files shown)


jkitchin:
Merged `main` in (`1966cc0`) to pick up #394, which landed `WitnessRule` in `witness_refutes_infeasibility` — the same closure this PR changes. It conflicted; both intents are kept.

**What #394 added:** a rule dispatch, with the DOF-gate path using `DeclaredRowRelative` (pure relative against the row's declared magnitude, no clamp), plus a `declared` binding factored out so that test has something to be relative to.

**What this PR adds:** directional presence tests for the absent-bound sentinel.

**Resolution:** `declared` is now computed from *present* bounds via the directional tests, and both violation terms are gated the same way; #394's `match rule` is untouched.

The two compose, and the combination is stronger than either alone. #394's `DeclaredRowRelative` falls back to the absolute floor when a row has "no declared magnitude" — deliberately, and its comment explains why that is the fail-closed direction. Under the old symmetric test, a row could report `declared == 0` while carrying a perfectly real bound of `-5e20`, taking that fallback **by accident** rather than for the documented reason. The directional test makes "no declared magnitude" mean what it says. I noted that inline where the fallback lives.

Re-ran on the merge:

| suite | result |
|---|---|
| `pounce-presolve` (incl. #394's new cases) | 238 passed |
| `presolve_certified_infeasibility` | 9 passed |
| `issue_387_overdetermined_infeasible` (CLI + algorithm) | 2 + 2 passed |
| `false_local_infeasibility` | 7 passed |
| `infeasible_status_tol_invariance` | 3 passed |
| `issue_372_infeasible_bounds_status` | 4 passed |
| `scaled_feasible_not_infeasible` | 1 passed |
| `test_scale_invariance.py` + `test_infeasibility_no_false_positives.py` | 2 passed |

400-instance sweep, re-measured on the merge: **0 false positives** on both the numerical (`presolve=no`) and full-presolve option sets. #394 lowered `BASELINE_WRONG["inf_eq"]` to 0; that holds with this change on top.

The "fails on the parent commit" evidence in the PR body was measured against `3a511fb`, still an ancestor of this branch — unchanged.

---

Issue #403, opened 2026-07-28
verify / check-x0 / Python preflight under-report violations against a bound past the sentinel


jkitchin:
Split out of the audit requested in #398. Lower severity than #401 and #402 —
these surfaces report rather than decide — but `verify` exists precisely to be
the independent check on a `.sol`, so a checker that under-reports is worth more
than its blast radius suggests.

## `pounce verify` — a band test used for both sides

`crates/pounce-cli/src/verify.rs:195-197`
```rust
pub(crate) fn is_finite_bound(b: Number) -> bool {
    b > NLP_LOWER_BOUND_INF && b < NLP_UPPER_BOUND_INF
}
```

This is a *band* membership test, applied to `lo` and `hi` alike. A real upper
bound of `-5e20` fails it (`-5e20 > -1e19` is false), so in `box_violation`
(`:253-262`):

```rust
let above = if is_finite_bound(hi) { v - hi } else { Number::NEG_INFINITY };
...
below.max(above).max(0.0)
```

`above` is `-inf` and the violation reads `0.0`. **`pounce verify` reports
ACCEPTED for a `.sol` that violates a real declared bound.** Mirror case for a
real lower bound of `+5e20`.

The same predicate also drives `row_magnitude` (`:214-218`), so a row written at
`5e20` scale reports a magnitude that ignores its own bound — which then feeds
the scale-relative feasibility comparison.

The existing coverage at `verify.rs:1112` pins only the *sentinel* case, so it
passes either way.

## `pounce check-x0` — same helper, re-exported

`crates/pounce-cli/src/check_x0.rs:35` imports `is_finite_bound` from `verify`
and uses it at `:525-528` (`at_lo` / `at_hi` bound-activity counting) and `:722`.
`clamp_to_interior` matches on `(is_finite_bound(lo), is_finite_bound(hi))`, so a
real out-of-range bound is not clamped against — the reported starting point is
not the one the solver would actually start from.

There is an irony worth noting: `check-x0` measuring the #398 fixture at exactly
zero violation is what established that model was feasible. On a model with a
bound past the sentinel, that measurement would itself be wrong.

## Python twins

`python/pounce/_preflight.py:189-190`
```python
def _finite_bound(b: float) -> bool:
    return bool(np.isfinite(b)) and -_BOUND_INF < b < _BOUND_INF
```
Exact twin of `verify.rs:195`, used at `:196-197` (`_box_violation`, under-reports
[...]

PR #407, opened 2026-07-28
Refuse to run the ratchet probes against an unvouched binary

jkitchin:
Follow-up to #403, where this surfaced. Not a bug in the solver — a hole in how
two of its correctness ratchets choose the binary they measure.

## What happened

`test_infeasibility_no_false_positives` and `test_scale_invariance` resolved the
solver with a bare `shutil.which("pounce")` + `subprocess.run`, so they measured
whatever happened to be installed. While working #403 they reported **49 of 200**
feasible-by-construction models in the AMPL infeasible band — against a ratchet
whose limit is `0`. That reads exactly like a #396-class regression. It wasn't:
the binary was a pip install from the previous day.

## Why neither existing guard caught it

#366 built real machinery for this — `check_binary()`, `_bundled_path()`, and
the `conftest` hook. It cannot reach these two, for two independent reasons:

1. **They never touch the pyomo plugin.** Zero `SolverFactory` uses in either
   file, so `_default_executable()` and the bundled-binary preference never run.
2. **Their failure mode is the one the hook deliberately excludes.** It fires
   only on `ApplicationError: ... did not exit normally` — the solver producing
   *no result*. The stale binary ran fine and wrote a valid `.sol` every time; it
   just returned wrong verdicts. #366's docstring is explicit that a guard broad
   enough to swallow wrong answers makes the suite meaningless, and that is still
   the right call. This is a gap beside it, not a defect in it.

## Resolution path is not evidence; the build is

My first cut preferred the wheel-bundled binary, on `conftest`'s own documented
reasoning that it is "by construction the build under test."

**That reasoning is CI-only.** In a working tree `python/pounce/bin/pounce` is
gitignored and survives across days and commits — and the binary that produced
the 49 false positives was sitting exactly there, one day and six commits stale.
The first cut would have trusted it and reproduced the whole problem. Caught by
running the guard against my own tree rather than assuming.

Both builds reported `0.9.0`. Only the embedded commit told them apart
(`10a6fe0c+dirty` vs `e17b0279`) — which is precisely what `_build_id` was
written for, per its own docstring about the #271/#272 dual-sign case.

## The guard

A session-scoped `pounce_exe` fixture: resolve (bundled, then `PATH`), then
compare the binary's build id against the checkout's `HEAD`, and skip with a
message naming both. Observed output:

```
SKIPPED [1] refusing to measure '/Users/.../python/pounce/bin/pounce': it is
build 10a6fe0c+dirty, but this checkout is e17b0279+dirty. These probes are
correctness ratchets — a different build's verdicts are not evidence either
way, in either direction. Rebuild and stage it (see this module's docstring),
or set POUNCE_TEST_EXE to choose a binary deliberately.
```

Two deliberate asymmetries, both about not making things worse:

- **Refuses only on a *proven* mismatch** — both ids readable and different. An
  unreadable id must never skip: a skipped ratchet proves nothing, which is the
  same silent loss by another route. `crates/pounce-cli/build.rs` embeds
  `unknown` when built outside a git checkout (a wheel built in a container
  without `.git`), and **CI is where a silent skip would hurt most**. I nearly
  shipped the fail-closed version before checking `build.rs`.
- **`+dirty` is ignored**, so a clean build against an edited tree still runs.
  That staleness is real but narrow, already covered by the source-mtime guard,
  and failing on it would fire constantly during edit-build-test — which trains
  people to bypass the guard.

`POUNCE_TEST_EXE` names a binary explicitly and steps aside. Explicit beats
bypassing: it records *which* binary was meant, where `PATH` manipulation
records nothing.

## Verification

Four scenarios exercised against the real tree:

| scenario | result |
|---|---|
| stale binary staged as bundled | refused, message above |
| build matching `HEAD` staged | runs, both probes pass |
| `POUNCE_TEST_EXE` set to a valid binary | runs |
| `POUNCE_TEST_EXE=/nope/pounce` | skipped, `is not a file` |

The guard is load-bearing, so it has its own coverage —
`pyomo-pounce/tests/test_exe_guard.py`, 10 cases pinning both directions,
including the two asymmetries above.

Full pyomo suite: **144 passed** (was 134; +10 guard tests). No new lint: the
one `F401` in `test_scale_invariance.py` is pre-existing on `main` and untouched.

## Note

This is a test-infrastructure change only — no solver behaviour is affected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,50 @@ changes.
 
 ## [Unreleased]
 
+### Tests — two correctness ratchets silently measured whatever `pounce` was on `PATH`
+
+- **`test_infeasibility_no_false_positives` and `test_scale_invariance` now
+  refuse to run against a binary this checkout cannot vouch for.** Both resolved
+  the solver with a bare `shutil.which("pounce")` and `subprocess.run`, so they
+  measured whatever happened to be installed. Found while fixing #403: a
+  pip-installed `pounce` from the previous day reported **49 of 200**
+  feasible-by-construction models in the AMPL infeasible band, against a ratchet
+  whose limit is `0`.
+- **Neither existing guard could reach this.** The #366 `conftest` hook covers
+  tests that go through pyomo's `SolverFactory` and fires only on
+  `ApplicationError: ... did not exit normally` — the solver producing *no*
+  result. These two use neither: no `SolverFactory`, so the plugin's
+  bundled-binary resolution never runs; and a foreign binary that answers
+  *incorrectly* exits cleanly and writes a valid `.sol`. That narrowness is
+  deliberate and still correct — #366's docstring argues a guard broad enough to
+  swallow wrong answers makes the suite meaningless — it just leaves this gap.
+- **Resolution path is not evidence; the build is.** The first cut preferred the
+  wheel-bundled binary, on the documented reasoning that it is "by construction
+  the build under test". That holds in CI, which stages it fresh, and fails in a
+  working tree, where `python/pounce/bin/pounce` is gitignored and survives
+  across days and commits — the binary that produced the 49 false positives was
+  sitting exactly there, one day and six commits stale. Both builds reported
+  `0.9.0`; only the embedded commit distinguished them (`10a6fe0c+dirty` vs
+  `e17b0279`), which is what `_build_id` exists for.
+- The new `pounce_exe` fixture compares the resolved binary's build against the
+  checkout's `HEAD` and skips with an actionable message naming both. Two
+  deliberate asymmetries:
+  - **It refuses only on a *proven* mismatch** — both ids readable and
+    different. An unreadable id must never skip, because a skipped ratchet
+    proves nothing and is the same silent loss by another route.
+    `crates/pounce-cli/build.rs` embeds `unknown` when it builds outside a git
+    checkout (a wheel built in a container without `.git`), and CI is where a
+    silent skip would hurt most.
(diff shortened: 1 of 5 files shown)


jkitchin:
Post-merge note, for the record.

CI on this PR showed the pyomo job dropping from **114s to 4.88s** while gaining
10 tests. That looked alarming enough to check before merging — a 20x speedup on
a suite whose slowest members are two subprocess-heavy probes reads like those
probes silently stopped running.

They did not. Two independent confirmations:

1. **Counts.** Baseline (#406) was `86 passed, 2 skipped`; this PR is
   `96 passed, 2 skipped`. Exactly +10 passed and +0 skipped — the 10 new
   `test_exe_guard.py` cases. Had the probes flipped to skipped it would read
   `94 passed, 4 skipped`. The 2 skips are pre-existing and identical before and
   after (they are collection-time, so they never appear in the progress line —
   also true on the baseline).

2. **The speedup has a mundane cause.** `shutil.which("pounce")` resolved the
   *console-script wrapper* — a Python shim that starts an interpreter and
   imports `pounce` before exec'ing the binary. The new fixture resolves
   `_bundled_path()`, i.e. the binary itself. Measured locally, 20 invocations:

   ```
   direct binary      0.118s
   console wrapper    4.702s
   ```

   ~235ms of interpreter startup per call versus ~6ms. Across the ~300
   subprocess solves these two probes perform, that is ~70s — which accounts for
   the gap.

So the guard also removed ~110s of pure Python-startup overhead from every CI
run, as a side effect of asking for the right binary. Not the point of the
change, but worth recording so the next person reading this timing does not
draw the conclusion I nearly did.

Issue #401, opened 2026-07-28
The convex route silently drops constraints whose bound lies past the opposite sentinel


jkitchin:
Split out of the audit requested in #398. That issue fixed `TNLPAdapter`; this
is the same confusion on the convex/LP/QP route, where it is worse — it does not
misclassify a constraint, it **deletes** it, and the solver then returns
`Optimal` at a point the model excludes.

## The predicate

Two independent copies, both symmetric:

`crates/pounce-cli/src/qp_extract.rs:34-38`
```rust
const NL_INF: f64 = 1e19;

fn is_finite_bound(v: f64) -> bool {
    v.abs() < NL_INF
}
```

`crates/pounce-cli/src/dispatch.rs:95-99`
```rust
const NL_INF: f64 = 1e19;

#[inline]
fn is_finite_bound(v: f64) -> bool {
    v.abs() < NL_INF
}
```

Presence is directional — a lower bound is absent iff `lo <= nlp_lower_bound_inf`,
an upper bound iff `hi >= nlp_upper_bound_inf`. A real upper bound of `-5e20` is
an ordinary finite bound; `|-5e20| < 1e19` is simply the wrong question.

### 1. A real variable bound is dropped from the box

`qp_extract.rs:173-181`
```rust
if is_finite_bound(xu) {
    let gr = next_row(&h);
    g.push(Triplet::new(gr, i, 1.0)); // x_i ≤ xu
    h.push(xu);
}
if is_finite_bound(xl) {
    let gr = next_row(&h);
    g.push(Triplet::new(gr, i, -1.0)); // −x_i ≤ −xl
    h.push(-xl);
}
```

`x_u = -5e20` fails the test, so `x_i <= -5e20` never enters `G`. The QP is
solved over a strictly larger box and reports `Optimal` at a point outside the
declared bounds. Mirror case for a real `x_l = +5e20`. Same code at `:495-500`.

### 2. An equality row vanishes from *both* `A` and `G`

`qp_extract.rs:134`
```rust
if lo == hi && is_finite_bound(lo) {
    // Equality row.
```

A genuine equality row at `±5e20` (both bounds present and equal) fails
`is_finite_bound`, falls into the `else`, and there both `is_finite_bound(hi)`
(`:144`) and `is_finite_bound(lo)` (`:155`) are false too — `upper` and `lower`
are both `None`. The row is gone entirely, contributing nothing to `A` and
nothing to `G`. Same code at `:459`.

[...]

PR #408, opened 2026-07-28
Grade the active-set QP against Maros-Meszaros ground truth

jkitchin:
Points the Maros-Mészáros set at the active-set QP engine, which had almost no
benchmark exposure. Suggested by Jeff Renfro, who uses active-set SQP for online
optimization and observed that a degeneracy-rich published set would surface
issues. He was right.

## The gap

Both NETLIB LP and Maros-Mészáros were **already** in `benchmarks/` with ground
truth. Nothing pointed them at this solver:

- `qp-active-set` was graded on **3 hand-constructed problems** in
  `adversarial_convex.md`
- `qp-ipm` was graded on all **138** MM instances with BPMPD optima

So this adds a fourth column to the existing runner rather than a new harness —
same `.nl` files, same ground truth, same correctness predicate. Renamed
`three_way` → `four_way` since it now reports four solvers.

## Result

| solver | correct | solved-but-wrong | not solved |
|---|---|---|---|
| qp-ipm | **137/138** | 0 | 1 |
| clarabel | 124/138 | **9** | 5 |
| nlp | 105/138 | **8** | 25 |
| qp-active-set | **46/138** | **0** | 92 |

Failure modes:

```
72  SearchDirectionBecomesTooSmall
14  TimeOut
 4  InternalError
 1  MaximumIterationsExceeded
 1  ParseError
```

Two things make this more tractable than 46/138 suggests:

- **One dominant symptom**, not 92 unrelated ones, and it is an
  *initialization* failure: **all 72 fail at `iterations = 0`**. The engine
  never takes a step. (The 46 it solves finish in 1-2 iterations in 44 cases.)
  So the defect is in phase 1 — building the initial working set or the first
  search direction — not in convergence or cycling. One root cause to chase,
  and a narrow one.

  On `QSHARE2B` (n=79, m=175, opt=11703.692), same `.nl`, both engines:

  ```
  qp-ipm          18 iterations →  1.1703691721e+04   correct
  qp-active-set    0 iterations →  0.0000000000e+00   Search Direction is becoming Too Small
  ```
- **Zero solved-but-wrong.** Every failure is loud. Clarabel has 9 wrong
  objectives here and the NLP path 8; the active-set engine never once claimed
  success with a wrong answer. The dangerous failure mode is absent.

And it is not uniformly weak — it wins **10 problems outright**, including 4 of
the LISWET family (`LISWET1`, `7`, `9`, `12`) where *both* Clarabel and the NLP
path return wrong objectives.

I verified failures at the CLI rather than trusting the table, because ~50
report objective exactly `0.0`, which looked like the solve never happening —
and in the sense that matters, it doesn't:

```
Problem class: convex QP. Selected solver: active-set QP (pounce-qp)
Overall NLP error: 6.9e+03
EXIT: Search Direction is becoming Too Small.
```

Correctly classified, correctly routed, then fails to converge. Not a routing
artifact.

## Not caused by this week's merges

`qp_extract.rs` feeds this path and was touched by #401, so I checked rather
than argued. A binary built at `bebfce44` (pre-#398) gives **bit-identical**
status and objective on 8 sampled failures. Long-standing, previously
unmeasured.

## Harness fixes, each made after being misled by it

- **Progress now flushes.** Redirected to a file, stdout is block-buffered, so
  progress lagged ~55 problems behind reality and read exactly like a hang on
  the tail of a long sweep. I killed a perfectly healthy run over it.
- **Speed-set membership is `is not None`, not truthiness.** `0.0` is falsy, so
  the old test silently dropped every sub-millisecond solve — which was *every*
  correct active-set solve — and crashed `summarize` when `geomean` of an empty
  list returned `None`. This bug was pre-existing in the three-way basis.
- **The active-set speed section reports the data as unavailable** instead of
  synthesizing a number. That path writes `total_wallclock_time_secs: 0.0` into
  `--json-output` while actually taking 0.568s on AUG2D (`qp-ipm` reports
  0.678s correctly on the same problem). A geomean over that field would read
  "instantaneous"; falling back to harness wall-clock would include process
  startup and compare a different quantity than the other three columns.
- **Reports record the producing binary's `--about` line.** The previous
  committed report was from 2026-06-13, and its NLP column had drifted
  129/138 → 105/138 on binary drift alone with nothing in the file to reveal
  it. Same lesson as #407.

## Two follow-ups this surfaces

Not fixed here — this PR only measures:

1. **The active-set phase-1 failures** — 72 × `SearchDirectionBecomesTooSmall`
   at `iterations = 0` on convex QPs with published optima. **`qp-ipm` solves
   all 72 correctly** from the identical `QpProblem`, so the problem data and
   the extraction are exonerated; the defect is in the active-set setup. Every
   one of the 72 is solved correctly by at least one other engine — none are
   intrinsically hard. That is a ready-made regression corpus with known
   answers and a same-codebase oracle.
2. **`total_wallclock_time_secs` is not populated on the active-set path**,
   making that column's timing unmeasurable.

## Verification

Full 138-problem sweep at the default 120s limit, against a binary built from
this checkout. No new lint (the 2 `ruff` findings are pre-existing on the
pre-rename file).


jkitchin:
### Narrowing the 72 phase-1 failures

Follow-up analysis on the failure set, for whoever picks this up.

**All 72 fail at `iterations = 0`.** Not a convergence or cycling problem — the
engine never takes a step. The 46 it solves finish in 1–2 iterations in 44 of
those cases. So the defect is in phase 1: initial working set / first search
direction.

**It is not input density.** The failures are *sparser* than the successes:

| group | N | med n | med m | med P density | med G density |
|---|---|---|---|---|---|
| FAIL (iters=0) | 72 | 1001 | 1522 | 2.1e-03 | 2.9e-03 |
| OK (correct) | 46 | 230 | 238 | 4.3e-03 | 3.2e-02 |

**Nor is it size.** `AUG2D` (n=20200) succeeds; `QSHARE2B` (n=79) fails.

**The strongest signal is equality constraints:**

```
          eq>0   eq==0
FAIL       68      4      -> 77% fail when equalities are present
OK         20     26      -> 13% fail when they are absent
```

Still not absolute: `UBH1` succeeds with 12012 equalities, `QSHARE2B` fails with
13. But note *which* problems survive many equalities — `UBH1`, `DTOC3`,
`AUG2D/2DC`, `AUG3D/3DC`: all structured control/PDE problems with banded
sparsity. The failures are Netlib-derived (`Q*`) and `CVXQP*`, with irregular
sparsity.

**Hypothesis (unconfirmed):** an equality-constrained phase 1 forming a reduced
KKT or null-space basis stays sparse under a banded structure but fills in
catastrophically on irregular Netlib sparsity. That would explain why input
density and problem size both fail to separate the groups while provenance
does. Confirming or killing this needs the phase-1 code, not more statistics.

**Cheapest reproducer:** `QSHARE2B` — n=79, m=175, 13 equalities, fails in
milliseconds at iteration 0, and `qp-ipm` solves it correctly from the
*identical* `QpProblem`:

```
qp-ipm          18 iterations ->  1.1703691721e+04   (opt 11703.692)
qp-active-set    0 iterations ->  0.0000000000e+00   Search Direction is becoming Too Small
```

All 72 are solved correctly by at least one other engine, and by `qp-ipm` in
every case — so there is a same-codebase oracle for each one.

jkitchin:
### Correction + profiler evidence: it is repeated *symbolic* factorization

Superseding the fill-in hypothesis in my previous comment. I profiled it instead
of reasoning about it, and two things in that comment were wrong.

**1. "Fails in milliseconds" was wrong.** I took that from
`total_wallclock_time_secs`, the field this PR documents as broken on this exact
path (always `0.0`). Measured wall clock:

| problem | n | m | qp-ipm | qp-active-set | iters |
|---|---|---|---|---|---|
| QSHARE2B | 79 | 175 | 0.02s | 0.30s | 0 |
| QSCTAP1 | 480 | 780 | 0.04s | 1.81s | 0 |
| Q25FV47 | 1571 | 2391 | 1.36s | **19.52s** | **0** |

It burns 19.5s to complete zero iterations on a problem qp-ipm solves in 1.36s,
scaling as ~`(n+m)^2.1`.

**2. The cost is not dense Gram-Schmidt, and not fill-in.** `macOS sample` over
the Q25FV47 run:

```
ParametricActiveSetSolver::solve            100.0%
  solve_elastic (phase 1)                    65.7%
    factorize_with_inertia_control           57.0%
      FERAL factorize_and_solve              56.7%
        symbolic_factorize_with_method       32.0%   <-- should run ONCE
```

**A third of total runtime is symbolic factorization** — fill-reducing ordering
and MC64 matching — redone on every KKT factorization inside the phase-1 elastic
loop, rather than analyzed once and reused numerically.

That inverts the premise of the method. Active-set is chosen for warm-started
sequences *because* refactorization is cheap; paying full symbolic analysis per
factorization makes it strictly more expensive per step than the IPM it exists
to beat.

Chain: repeated symbolic refactorization in phase 1 → no usable direction →
`SearchDirectionBecomesTooSmall` at outer iteration 0. Whether the re-analysis
causes the bad direction or merely accompanies it needs the code; the cost is
measured.

Worth noting `independent_active_subset` (solver.rs:1978) already anticipates a
related trap — its comment calls dense MGS "the operation that grinds large
degenerate LPs (the NETLIB GEN family) to a halt" and prefers a sparse
rank-reveal. FERAL does provide that probe (`provides_degeneracy_detection() ->
true`), so that guard appears to be working; the cost is one level down, in the
factorization itself.

Reproducer unchanged: `QSHARE2B`, n=79, 13 equalities, 0.30s to do nothing,
with `qp-ipm` solving the identical `QpProblem` in 0.02s.

PR #421, opened 2026-07-31
estimate() baseline is the solve point, not the Param's current value

### LINKED ISSUE #420: estimate() measures its perturbation from the Param's current value, silently returning the unperturbed solution
`estimate(model, perturb)` computes each step as `new_value - pyo.value(param)`, against the Param's current value on the model. The held factorization describes exactly one point, the solve point, and the two coincide only while the Param is untouched after the solve.

The natural receding-horizon pattern touches it: solve at a predicted value, write the arriving measurement into the Param, then ask for the estimate at that measurement.

```python
# solved with m.p == 2.0, p declared
m.p.set_value(2.2)               # measurement arrives
est = estimate(m, [(m.p, 2.2)])  # delta = 2.2 - 2.2 = 0
# est is the p == 2.0 solution, unperturbed, silently
```

No error, no warning, and the output is a valid solution (at the prediction), so nothing looks wrong downstream; a controller consuming it applies uncorrected inputs. Found while building an advanced-step controller whose specified call sequence is exactly this pattern.

Two remedies:

1. **Measure the step from the solve point** (proposed). The declared Param's pinned variable already holds the solve-time value in the stored primal vector, so no new information is needed. Untouched Params give identical numbers; write-then-ask and ask-then-write agree.
2. Detect a changed Param and raise. This forces callers to ask for corrections before recording measurements, which inverts the receding-horizon flow `estimate()` exists to serve.

A PR implementing remedy 1 with a regression test follows.

devin-griff:
Fixes #420.

`sens_solve` builds `pin_vars`, a map from each declared ParamData to its pinned variable's row in the stored primal vector, beside the existing `pins` (constraint-row) map. `estimate()` measures each delta against `session.base_x[pin_vars[pd]]` instead of `pyo.value(pd)`: the solve-time value, the one point the held factorization describes. A Param untouched since the solve gives identical numbers, so existing callers are unchanged; the cost is one row index per declared parameter.

The docstring states the baseline semantics and the CHANGELOG entry carries the defect. Regression test `test_estimate_baseline_is_the_solve_point`: solve, write the new value into the Param, ask at that value, compare against a re-solve there; fails on the old baseline (zero delta), passes with the fix.

Suite: 143 passed, 2 skipped; all 34 pre-existing sens tests pass unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ changes.
 
 ## [Unreleased]
 
-### Fixed — pyomo-pounce: `estimate()` measured its perturbation from the Param's current value (#TBD)
+### Fixed — pyomo-pounce: `estimate()` measured its perturbation from the Param's current value (#420)
 
 - `estimate(model, perturb)` computed each step as `new value` minus the
   Param's current value on the model. The factorization the step runs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,11 @@ changes.
   zero and `estimate()` silently returned the unperturbed solution. No
   error, no warning, and the output is a valid solution (at the
   prediction), so nothing looked wrong downstream.
-- The baseline is now the declared Param's pinned-variable entry in the
-  stored primal vector: the solve-time value, already retained. A caller
-  that has not touched the Param sees the same numbers as before; storage
-  is one row index per declared parameter. The docstring states the
-  baseline semantics.
+- The baseline is now the pin constraint's stored right-hand side: the
+  value the perturbation actually shifts, holding the Param's solve-time
+  value exactly, already retained. A caller that has not touched the
+  Param sees the same numbers as before; no new state is stored. The
+  docstring and `docs/src/sensitivity.md` state the baseline semantics.
 - Regression: solve, write the new value into the Param, ask at that
   value, compare against a re-solve there. Fails on the old baseline
   (zero delta), passes now.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,27 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — pyomo-pounce: `estimate()` measured its perturbation from the Param's current value (#TBD)
+
+- `estimate(model, perturb)` computed each step as `new value` minus the
+  Param's current value on the model. The factorization the step runs
+  through describes exactly one point, the solve point, so the two agree
+  only while the Param is untouched. In the receding-horizon pattern the
+  caller solves at a prediction, writes the arriving measurement into the
+  Param, then asks for the estimate at that value: the delta came out
+  zero and `estimate()` silently returned the unperturbed solution. No
+  error, no warning, and the output is a valid solution (at the
+  prediction), so nothing looked wrong downstream.
+- The baseline is now the declared Param's pinned-variable entry in the
+  stored primal vector: the solve-time value, already retained. A caller
+  that has not touched the Param sees the same numbers as before; storage
+  is one row index per declared parameter. The docstring states the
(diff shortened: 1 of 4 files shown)


jkitchin (commented):
Correct fix, and the right one of the two remedies in #420. I built the extension and CLI from this branch and ran the suite rather than reading it.

## What I verified

| | result |
|---|---|
| `pyomo-pounce/tests/test_sens.py` on this branch | 35 passed |
| same tests with `sens.py` reverted to the base commit (`dd6a745`) | `test_estimate_baseline_is_the_solve_point` **fails**, 34 pass |
| full `pyomo-pounce/tests` | 97 passed, 2 skipped (networkx) |

So the regression test genuinely bisects the defect and nothing else moved. `pyomo-pounce-smoke` runs `pytest pyomo-pounce/tests`, so the new test is covered in CI.

I also checked two paths the new test does not reach. Both now agree with a re-solve under write-then-ask, where the old baseline returned the unperturbed solution:

- **indexed Param** — `m.p[1]` written to 3.0, asked at 3.0: `[2.666667, 0.666667]`, matching the re-solve
- **fixed Var declared as a sens param**: `2.5` vs re-solve `2.5`

And `python/notebooks/25_pyomo_sensitivity.ipynb` asks before writing, so its printed outputs are unchanged.

## Notes — none blocking

**1. The reference the step actually perturbs is the pin row's RHS, not the pin variable's value.** `parametric_step` shifts the pin *constraint*, and that row already carries the Param's solve-time value directly — on the test model I read `g_l = g_u = 2.0` on `_SENSITIVITY_TOOLBOX_DATA.paramConst[1]`. This PR measures from the pin *variable's* solved value, which is that RHS plus the row's primal residual. The measured residual was exactly `0.0` for both declared params, and it is bounded by the constraint tolerance on any converged solve, so this is theoretical. Raising it only because `session.nl.g_l[_param_pin(session, pd)]` is that exact value and needs no new session field at all — and because the docstring and CHANGELOG both say "the solve point", which these two readings satisfy slightly differently.

**2. `pin_vars` is attached after construction** (`pyomo-pounce/pyomo_pounce/sens.py:584`) while `pins` is a constructor argument and `base_x` / `moved_bounds` get defaults in `__init__`. `estimate()` now dereferences it unconditionally, so a `_Session` built any other way raises `AttributeError` rather than the clean "no sensitivity session" error. `self.pin_vars = ComponentMap()` beside `self.pins` (line 259) closes that. It matches how `fit_rows` / `res_rows` are already done, so take it or leave it.

**3. `docs/src/sensitivity.md` does not get the new semantics.** That page documents `estimate()` and links the NMPC notebook, so it is read by exactly the audience that hits this bug; the baseline rule currently lives only in the docstring and the CHANGELOG. One sentence around line 98 would put it where people look.

## Adjacent, out of scope

`covariance()`'s `n_data=` branch reads SSR as `pyo.value(objective)` off the current model (`sens.py:1057`), so a Param mutated after the solve silently changes the reported variance — the same class of staleness as #420. The `declare_residual` path is unaffected (it uses `base_x`). Probably worth its own issue.

---

devin-griff:
All three notes taken, in 95b5c3a:

1. The baseline is now `session.nl.g_l[_param_pin(session, pd)]`: the pin row's RHS, the value the step actually shifts, exact rather than exact-minus-residual. `pin_vars` is gone entirely, so there is no new session state, which also retires note 2.
2. Absorbed by 1.
3. `docs/src/sensitivity.md` now states the baseline rule beside the `estimate()` introduction, and the docstring and CHANGELOG both say "the pin constraint's stored right-hand side".

Suite after the rework: 143 passed, 2 skipped.

The adjacent `n_data` SSR staleness is filed as #426.

Issue #296, opened 2026-07-23
[Parity] .sol omits bound multipliers (ipopt_zL_out/zU_out); reduced costs unavailable to AMPL/Pyomo/GAMS


jkitchin:
> Surfaced while auditing the dual/marginal surfaces for #271 / #272. An Ipopt-parity gap for AMPL/Pyomo/GAMS users doing sensitivity work.

## The gap

pounce's `.sol` writer emits the constraint duals (now with the correct sign, #287) but **no bound multipliers**. Ipopt writes `ipopt_zL_out` / `ipopt_zU_out` suffix blocks — the reduced costs / bound sensitivities — which Pyomo exposes as `model.ipopt_zL_out[var]` / `model.ipopt_zU_out[var]` and AMPL as variable `.rc`.

Verified on `main` (bound-constrained model, `.sol` written under `-AMPL`):

```
pounce .sol suffix blocks: 0
ipopt  .sol suffix blocks: 2   (ipopt_zL_out, ipopt_zU_out)
```

An earlier adversary run also noted the Pyomo `rc` suffix comes back `None` from pounce. The underlying multipliers **exist** internally (`mult_x_L` / `mult_x_U`, confirmed correct against Ipopt during the #271 audit) — they are simply not written to the `.sol`.

## Why it matters

Reduced costs are the standard way to ask "how much would relaxing this variable's bound improve the objective" — bread-and-butter LP/QP sensitivity, and a routine part of parameter-estimation and design workflows. A user migrating from Ipopt (pounce advertises drop-in Ipopt compatibility) who reads `model.ipopt_zL_out` gets nothing back, with no error — a silent behavioral gap rather than a wrong answer, but a surprising one.

## Toward SOTA

- Write `ipopt_zL_out` / `ipopt_zU_out` variable-suffix blocks in the `.sol` writer (the `SolutionFile` already carries suffix support; the values are the internal `mult_x_L` / `mult_x_U`, with the same AMPL-marginal sign treatment applied to constraint duals in #287).
- Populate the Pyomo `rc` suffix in `pyomo-pounce` from those, matching the ASL driver's convention.
- A regression test against Ipopt's `zL`/`zU` on a bound-active model (the sign/scale can be pinned analytically for a textbook box-constrained QP).

Related: #271, #272, #287.

PR #433, opened 2026-08-01
Sens path: forward solver options; warm-start from the model's suffixes

### LINKED ISSUE #432: the sens path drops solver options, and with them the warm start
`PounceSolver.solve()` reroutes any model carrying declarations to the in-process session, and everything the caller asked for is dropped silently on that path: `sens_solve` has no options parameter, so per-call `options={...}` and factory-level `solver.options[...]` alike run the solve at defaults. No warning, and the failure is invisible by construction: declaring one sensitivity parameter reroutes every subsequent solve, so options that worked yesterday stop working the day a declaration is added. Repro: a declared model solved with `options={"max_iter": 1}` reports `optimal` after running to completion.

Behind it sits a missing capability the dropped options were masking: even with options flowing, `warm_start_init_point=yes` has nothing to consume. The ASL path seeds initial multipliers from `dual` / `ipopt_zL_in` / `ipopt_zU_in`; the in-process path calls `Solver.solve(x0)` bare, though the session API already accepts `lagrange`/`zl`/`zu` and the core has the switch.

One deliberate divergence proposed for the wiring, rather than inherited: through ASL an absent suffix entry reads as a zero multiplier, because a dense array cannot express "unknown". A zero bound multiplier on an active bound is a contradictory KKT certificate; measured on one warm-started problem, sparsely populated suffixes gave 293 iterations where densely seeded ones gave 9. A suffix knows which entries exist, so the in-process path can honor an explicit zero and fall back to the solver's own defaults for absent entries. Dense round-trips behave identically to ASL; sparse seeds get a sane default instead of a poisoned certificate.

A PR with both fixes follows.

devin-griff:
Fixes #432.

**Options.** `sens_solve` gains `options=`; `solve()` forwards factory options with per-call `options=` on top (the ASL layer's bookkeeping `solver` key excluded), applied after the tee default so an explicit `print_level` wins. Regression: `max_iter=1` stops a declared model's solve, from both option sources.

**Warm start.** With `warm_start_init_point=yes`, `_warm_start_from_suffixes` reads `dual` / `ipopt_zL_in` / `ipopt_zU_in` by component name into the session's initial multipliers; entries naming components the solve doesn't carry (e.g. constraints replaced by the declared-parameter surgery) are skipped. Absent entries fall back to the solver's defaults (`bound_mult_init_val` for bound multipliers, 0 for equality duals), never zero; explicit zeros are honored; the rationale and the divergence from ASL's zero-fill are stated in the docstring and CHANGELOG. The sens path's results object now carries the iteration count (`statistics.black_box.number_of_iterations`), which the warm-start regression needs. Tests: a unit test pins the fallback semantics (explicit zero kept, absent entry defaulted); end to end, a plain solve's exported multipliers fed back as suffixes make the declared warm re-solve beat the cold one.

Suite: 149 passed, 2 skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,18 +25,27 @@ changes.
 
 ### Added — pyomo-pounce: in-process warm starts from the model's suffixes (#432)
 
-- With `warm_start_init_point=yes` among the options, the sens path
-  reads the model's `dual` / `ipopt_zL_in` / `ipopt_zU_in` suffixes
-  into the session's initial multipliers, matched by component name.
-- Entries the user did not supply fall back to the solver's own
-  defaults (`bound_mult_init_val` for bound multipliers, 0 for
-  equality duals), never zero. Through ASL an absent entry reads as a
-  zero multiplier because a dense array cannot say "unknown", and a
-  zero bound multiplier on an active bound is a contradictory
-  certificate the solver must first recover from. A suffix knows which
-  entries exist, so an explicit zero is honored and absence means
-  "initialize normally": a deliberate divergence from the ASL path's
-  accidental zero-fill.
+- With `warm_start_init_point=yes` (or `True`; `add_option` maps them
+  alike) among the options, the sens path reads the model's `dual` /
+  `ipopt_zL_in` / `ipopt_zU_in` suffixes into the session's initial
+  multipliers, matched by component name, with a constraint replaced
+  by the declared-parameter surgery reached through its clone alias.
+  Both external sign conventions are crossed on the way in: `dual`
+  holds the AMPL marginal `-λ` (#271) and `ipopt_zU_in` Ipopt's
+  negative-at-upper `z_u` (#296); the session wants the internal
+  `+λ` and non-negative `z_u`.
+- Entries the user did not supply are seeded NaN, a new "unseeded"
+  marker in the warm-start contract: the initializer substitutes its
+  own resolved defaults (`bound_mult_init_val`, including the
+  Mehrotra override, for bound multipliers; 0 for equality duals), so
+  the defaults live in one place. Through Ipopt's ASL interface an
+  absent entry reads as a zero multiplier because a dense array
+  cannot say "unknown", and a zero bound multiplier on an active
+  bound is a contradictory certificate the solver must first recover
+  from; a suffix knows which entries exist, so an explicit zero is
+  honored and absence means "initialize normally". (POUNCE's own CLI
+  reads no `ipopt_zL_in`/`ipopt_zU_in` at all today, so the
+  comparison is with Ipopt-via-ASL, not this project's binary.)
 - The sens path's results object now reports the iteration count
   (`statistics.black_box.number_of_iterations`).
 - Regression: a plain solve's exported multipliers, fed back as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,41 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — pyomo-pounce: the sens path dropped solver options (#432)
+
+- `SolverFactory("pounce").solve(m, options={...})` and factory-level
+  options were silently ignored the moment a model carried
+  declarations: the reroute to the in-process session forwarded
+  nothing. `max_iter`, tolerances, scaling, everything ran at defaults
+  with no signal, so a model stopped being tunable the day it gained a
+  declaration.
+- Options now flow: factory options first, per-call `options=` on top,
+  applied after the tee default so an explicit `print_level` wins. The
+  ASL layer's bookkeeping `solver` key is excluded.
+- Regression: `max_iter=1` must stop a declared model's solve, from
+  both option sources.
+
+### Added — pyomo-pounce: in-process warm starts from the model's suffixes (#432)
+
+- With `warm_start_init_point=yes` among the options, the sens path
+  reads the model's `dual` / `ipopt_zL_in` / `ipopt_zU_in` suffixes
+  into the session's initial multipliers, matched by component name.
+- Entries the user did not supply fall back to the solver's own
+  defaults (`bound_mult_init_val` for bound multipliers, 0 for
+  equality duals), never zero. Through ASL an absent entry reads as a
+  zero multiplier because a dense array cannot say "unknown", and a
+  zero bound multiplier on an active bound is a contradictory
+  certificate the solver must first recover from. A suffix knows which
+  entries exist, so an explicit zero is honored and absence means
+  "initialize normally": a deliberate divergence from the ASL path's
+  accidental zero-fill.
+- The sens path's results object now reports the iteration count
+  (`statistics.black_box.number_of_iterations`).
+- Regression: a plain solve's exported multipliers, fed back as
+  suffixes, make the declared warm re-solve beat the cold one; the
+  reader's fallback semantics are unit-tested (explicit zero kept,
+  absent entry defaulted).
+
 ### Added — post-solve activity classification on `Solver` (#362, covariance roadmap item 0)
 
 - `Solver.classify_activity()` (Rust: `pounce_sensitivity::activity`)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,10 +35,15 @@ changes.
   negative-at-upper `z_u` (#296); the session wants the internal
   `+λ` and non-negative `z_u`.
 - Entries the user did not supply are seeded NaN, a new "unseeded"
-  marker in the warm-start contract: the initializer substitutes its
-  own resolved defaults (`bound_mult_init_val`, including the
-  Mehrotra override, for bound multipliers; 0 for equality duals), so
-  the defaults live in one place. Through Ipopt's ASL interface an
+  marker in the warm-start contract of `Problem.solve` /
+  `Solver.solve`'s `lagrange`/`zl`/`zu` arguments: the warm-start
+  initializer substitutes its own resolved defaults
+  (`bound_mult_init_val`, including the Mehrotra override, for bound
+  multipliers; for equality duals the warm path's existing 0, which
+  is not the cold path's least-squares estimate), so the defaults
+  live in one place. The contract covers the warm-start initializer
+  only: the batched solver's multiplier seeds and the SQP
+  working-set arrays do not route through it and must not carry NaN. Through Ipopt's ASL interface an
   absent entry reads as a zero multiplier because a dense array
   cannot say "unknown", and a zero bound multiplier on an active
   bound is a contradictory certificate the solver must first recover
--- a/crates/pounce-algorithm/src/alg_builder.rs
+++ b/crates/pounce-algorithm/src/alg_builder.rs
@@ -353,6 +353,12 @@ pub struct WarmStartOptions {
     pub target_mu: Number,
     pub entire_iterate: bool,
     pub same_structure: bool,
+    /// The value a NaN-seeded bound multiplier takes: NaN in a
+    /// user-supplied `z`/`v` seed means "unseeded, use the default".
+    /// Threaded from `builder.init.bound_mult_init_val` at build time
+    /// so the Mehrotra override and any user setting stay the single
+    /// source of truth.
+    pub bound_mult_init_val: Number,
 }
 
 impl Default for WarmStartOptions {
@@ -367,6 +373,7 @@ impl Default for WarmStartOptions {
             target_mu: 0.0,
             entire_iterate: false,
             same_structure: false,
+            bound_mult_init_val: 1.0,
         }
     }
 }
@@ -1021,7 +1028,9 @@ impl AlgorithmBuilder {
 
         let init: Box<dyn crate::init::r#trait::IterateInitializer> = if self.warm_start_init_point
         {
-            Box::new(WarmStartIterateInitializer::with_options(self.warm.clone()))
+            let mut wopts = self.warm.clone();
+            wopts.bound_mult_init_val = self.init.bound_mult_init_val;
+            Box::new(WarmStartIterateInitializer::with_options(wopts))
         } else {
             let mut d = DefaultIterateInitializer::with_eq_mult_calculator(Box::new(
                 LeastSquareMults::new(),
--- a/crates/pounce-algorithm/src/alg_builder.rs
+++ b/crates/pounce-algorithm/src/alg_builder.rs
@@ -124,6 +124,32 @@ pub struct AlgorithmBundle {
     pub search_dir: Option<PdSearchDirCalc>,
 }
 
+#[cfg(test)]
+mod warm_threading_tests {
+    use super::*;
+
+    #[test]
+    fn warm_options_take_the_init_default_not_their_own() {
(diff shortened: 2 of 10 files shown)


jkitchin (changes requested):
## What this does

Two changes to the sens path (the in-process reroute `SolverFactory("pounce").solve()` takes when a model carries declarations):

1. **Options forwarding** — `sens_solve` gains `options=`; `PounceSolver.solve` merges factory options (minus ASL's bookkeeping `solver` key) with per-call `options=`, applied after the `tee`-derived `print_level` default.
2. **Warm start** — with `warm_start_init_point=yes`, `_warm_start_from_suffixes` reads `dual` / `ipopt_zL_in` / `ipopt_zU_in` into `lagrange`/`zl`/`zu` seeds by component name, absent entries falling back to `bound_mult_init_val` rather than zero. Plus `number_of_iterations` on the results object.

The options half is correct and I'd merge it as-is. The warm-start half has multiplier-convention bugs.

---

## Blocking

### 1. `dual` is fed into `lagrange` without the sign flip

`pyomo_pounce/sens.py:371` does `y[r] = float(val)` and passes `y` as `lagrange=`, which is pounce's **internal** convention (`L = f + λ'g`). But `model.dual[c]` is the AMPL **marginal** `d obj/d b = −λ`. This repo negates on the way out in `crates/pounce-cli/src/nl_writer.rs:160-166` (gh #271), and `sens.py:686` — in this same file — already spells the rule out:

> `mult_entry` rows come from … the internal Lagrange multiplier, whereas `m.dual[con]` holds the AMPL *marginal* `d obj / d b = -lambda` (gh #271)

So every equality multiplier is seeded with the wrong sign — strictly worse than not warm-starting, since the solver starts from a valid dual point reflected through zero. Needs `y[r] = -float(val)`.

### 2. `ipopt_zU_in` is fed into `zu` without the sign flip

`sens.py:372-378` treats `ipopt_zL_in` and `ipopt_zU_in` identically. They aren't: `crates/pounce-cli/tests/issue_296_bound_multiplier_suffixes.rs:26` pins `ipopt_zU_out = −z_u` (Ipopt's convention — negative at an active upper bound), while `zu=` expects the internal non-negative `z_u` (`crates/pounce-py/src/warm_start.rs:30`: "`z_l`, `z_u` … non-negative each").

Under the exact round-trip this PR's own test performs (`zU_out` → `zU_in`), an active upper bound seeds a **negative** `z_u`, which `WarmStartIterateInitializer` then clamps up to `mult_bound_push`. With the `warm_start_mult_bound_push=1e-9` the test sets, that is effectively zero — precisely the "contradictory KKT certificate the solver must first recover from" that the docstring says this feature exists to avoid.

### 3. The dual channel is a no-op on the models this path exists to serve

`con_row.get(cd.name)` matches against `con_names` read from the **clone's** `.row` file. The declared-parameter surgery replaces every constraint mentioning a declared param with a copy on `_SENSITIVITY_TOOLBOX_DATA` — which is exactly why `con_alias` exists (`sens.py:292-295`, built at `sens.py:580-585`). So for any model with `sens_params`, the constraints most likely to carry meaningful duals are silently skipped.

The PR body frames this as intentional ("entries naming components the solve doesn't carry … are skipped"), but the aliasing information is already computed from `block._replaced_map` and is available before the solve. Skipping is a workaround for a translation the code can already do. At minimum `_warm_start_from_suffixes` should take the alias map and apply `con_alias.get(name, name)`, the way `mult_entry` does.

This also explains why 1 and 2 don't show up in CI: in `test_warm_start_from_suffixes_reduces_iterations` the only constraint `m.c` mentions `m.p`, so its dual is dropped; `m.x` has only a lower bound and `m.y` is free, so every `zU_out` entry is ≈0. **The only channel the end-to-end test actually exercises is `zl`, which happens to be the one with the correct sign.**

---

## Non-blocking

- **`warm_start_init_point=True` enters warm-start mode with no seeds.** `str(...).lower() == "yes"` rejects `True`, but `Problem.add_option` maps Python `True` → `"yes"` (`crates/pounce-py/src/problem.rs:170-177`). A user writing `options={"warm_start_init_point": True}` — natural in Pyomo — gets the warm-start initializer *with an empty `warm` dict*, i.e. `warm_start_bound_push` applied to a cold point. Accept bools (and `"y"`).
- **`bound_mult_init_val` default duplicated as `1.0`.** `crates/pounce-algorithm/src/application.rs:2255` raises it to `10.0` under `mehrotra_algorithm`, so the Python-side fallback silently disagrees there, and it drifts if the Rust default changes.
- **`test_factory_options_reach_the_sens_path` doesn't test its docstring.** It claims "the per-call `options=` wins on conflict" but never passes a per-call option. Set `solver.options["max_iter"] = 1` and call with `options={"max_iter": 500}`, asserting *optimal* — that actually pins the precedence.
- **`assert its_warm < its_cold` is a bare strict inequality** on an iteration count for a 2-variable problem, no margin. Once the sign bugs and the alias translation land, a direct unit assertion on the `warm` dict for the `zU`/`dual` channels (`test_warm_start_reader_fallback_semantics` is the right shape) would catch regressions the iteration count can't.
- **No docs.** `docs/src/pyomo.md` / `docs/src/sensitivity.md` are untouched, though this is user-visible behavior with a deliberate semantic divergence from the ASL path. That rationale currently lives only in a private function's docstring and the CHANGELOG.
- The CHANGELOG says absence-vs-zero is indistinguishable "through ASL" — worth noting pounce's own CLI doesn't read `ipopt_zL_in`/`ipopt_zU_in` at all, so via pounce-via-ASL `warm_start_init_point=yes` consumes no suffixes whatsoever. The claim is about Ipopt's ASL interface, not this project's.
- Nit: `for _k, _v in (options or {}).items()` — the leading underscores read as "unused" and aren't the convention elsewhere in this file.

---

## Verdict

Request changes. The options half is good. The warm-start half seeds two of its three channels with the wrong sign convention and drops the third for the models it targets — and the suite can't see any of it, because the end-to-end test's geometry exercises only the one correct channel.

Findings 1–3 come from reading the conventions pinned in `nl_writer.rs`, `issue_296_bound_multiplier_suffixes.rs`, `warm_start.rs`, and `sens.py:686`; I did not reproduce them by running the suite.

---

devin-griff:
All taken, in 458690e:

**Blocking.** (1) `dual` negates on the way in (the AMPL marginal → internal λ). (2) `ipopt_zU_in` negates (Ipopt's negative-at-upper → the internal non-negative `z_u`); `zL` stays positive in both conventions. (3) The replaced-constraint alias map is built before the solve (`_replaced_aliases`, from `block._replaced_map`) and the reader translates through it, so declared models' duals reach their rows; the session stores the same map it did before.

**The default-duplication note got the deeper fix.** Rather than mirroring `bound_mult_init_val` (and its Mehrotra override) in Python, the warm-start contract gains NaN-means-unseeded: the reader seeds absent entries NaN, and `WarmStartIterateInitializer` resolves them against `builder.init.bound_mult_init_val` (bound multipliers) or 0 (equality duals) before its clamps — the clamp block now always runs, so NaN resolves even with both clamps disabled. The default lives in exactly one place; unit-tested at the initializer.

**The rest.** `warm_start_init_point` accepts `True`/`"y"`/`"yes"`, matching what `add_option` maps; the factory-options test now pins precedence (factory `max_iter=1`, per-call `500`, asserting optimal); the reader unit test asserts every channel directly (dual negated, `zU` negated, explicit zero honored, absent entries NaN, an aliased dual landing on its row); a new partial-seed end-to-end test seeds duals only, rides every bound multiplier through the NaN channel, and comes back optimal — the pipe your review showed the old test never exercised; `docs/src/sensitivity.md` gains a section on options forwarding and warm starts with the divergence rationale; the CHANGELOG attributes the absence-vs-zero claim to Ipopt's ASL interface and notes POUNCE's own CLI reads no `zL_in`/`zU_in` today; the `_k`/`_v` nit.

Suites after rebuilding from the branch: pyomo-pounce 153 passed, pounce-algorithm 319 passed.

jkitchin (commented):
Re-reviewed at `458690e`. **All three blocking items from my previous review are resolved**, and the non-blocking list is fully addressed. Everything below is non-blocking — I'd be happy to see this merged once the `CompoundVector` gap and the doc placement are settled.

## Resolved

- **`dual` sign** — `y[r] = -float(val)`, with the gh #271 rationale in the docstring. ✅
- **`ipopt_zU_in` sign** — handled via the `(suffix, array, sign)` table, `-1.0` for zU, citing gh #296. ✅
- **Alias translation** — `_replaced_aliases(clone, si)` extracted and built *before* the solve; the reader translates through it and the session reuses the same map rather than rebuilding it. ✅
- Plus: `True`/`"y"` accepted with a unit test, the factory-options test now actually pins per-call precedence, `docs/src/sensitivity.md` documents the feature, the CHANGELOG attributes the zero-fill claim to Ipopt-via-ASL, `_k`/`_v` cleaned up.

The `bound_mult_init_val` duplication got a better fix than I suggested — NaN-as-unseeded resolved against `builder.init.bound_mult_init_val` beats copying the constant into Python. But it grows the diff into the engine's warm-start contract, and that's where the remaining comments are.

## On the NaN-sentinel design

- **NaN resolution silently no-ops on non-dense vectors.** `clone_clamped` does `out = v.make_new()` (same concrete type as `v`), then `if let Some(d) = out.as_any_mut().downcast_mut::<DenseVector>()`. `CompoundVector` is the other `Vector` impl; for it the downcast fails and NaN rides straight through `element_wise_min`/`element_wise_max` into the iterate — NaN poisoning instead of a default. The user-seed path is safe today (`seed_from_nlp` builds DenseVectors unconditionally), but the re-optimize path (`set_warm_start_iterate`) reuses whatever the prior solve's spaces produced. Worth an explicit `else` / `debug_assert!` rather than a silent skip.
- **The contract is public but documented in one place only.** `Problem.solve(lagrange=, zl=, zu=)` in `crates/pounce-py/src/solver.rs` and the module docs in `problem.rs` still describe those arguments with no mention of NaN, and no `docs/src/` warm-start page is touched — only `sensitivity.md`, under the pyomo heading. Same for the CHANGELOG entry. Anyone driving `pounce.Problem` directly is now subject to a semantics change they cannot discover.
- **Inert on the other seed-consuming paths.** `nlp_batch.rs` pulls `mult_x_L` / `mult_x_U` seeds and the SQP working-set path takes `z_l` / `z_u`; neither routes through `WarmStartIterateInitializer`, so NaN there is garbage-in, not "unseeded". Either reject NaN at those entry points or state the limitation.
- **"The default lives in one place" isn't quite true yet** — `WarmStartOptions::default()` re-hardcodes `bound_mult_init_val: 1.0`, a third copy alongside `DefaultInitOptions::default()`. Seeding it from the latter would close it. Relatedly, nothing tests the threading at `alg_builder.rs:1031-1033` or the Mehrotra override; `tests_nan_seed` calls `clone_clamped(…, 7.0)` directly, so the piece most likely to regress silently is the untested piece.
- **"0 for equality duals" isn't the solver's default.** The cold path is `DefaultIterateInitializer::with_eq_mult_calculator(LeastSquareMults::new())`, so an unseeded dual on a cold solve gets a least-squares estimate. NaN→0 matches what `seed_from_nlp` already did, so this is not a regression — but "the initializer substitutes its own resolved defaults … 0 for equality duals" reads as if 0 *were* the default, and the PR's own "absence must not silently mean zero" argument applies here verbatim. Either route NaN duals through the least-squares calculator or say plainly that the dual channel keeps the warm path's zero.
- **Comment overstates.** The clamp block's new "cap = inf, floor = 0, an identity clamp" isn't identity — floor 0 still clamps negative `z`/`v`. Harmless in practice (the default `mult_bound_push` is 1e-3, so the block always ran anyway), but the comment is wrong as written.
- **Doc inconsistency.** `_warm_start_from_suffixes`'s docstring correctly notes an explicit zero is "then floored at `warm_start_mult_bound_push`"; `docs/src/sensitivity.md` still says flatly "an explicit zero is honored."

## One thing I'd still push on

`test_warm_start_partial_seed_solves_clean` seeds only `dual` and asserts `optimal` — but if the alias translation were broken, `y` would be all-NaN → resolved to 0 → **still optimal**. So the end-to-end effect of the alias fix is not asserted anywhere; `test_warm_start_reader_semantics` covers it only against synthetic names (`["c1", "c2_replaced"]` with a hand-built map), never against a map `_replaced_aliases` produced from a real surgery block. That is the same blind spot as the original finding, one level up. A test that runs `sens_solve` on a declared model and checks the dual actually landed on the aliased row would close it.

---

devin-griff:
All taken, in b0d25b5 + c34d4c7:

**CompoundVector.** `clone_clamped` now fails loudly (`debug_assert` on a NaN dot) when NaN reaches a non-dense vector, instead of silently skipping resolution; the comment states why NaN there is a contract violation rather than an unseeded entry.

**Contract visibility.** `Problem.solve` documents the seed arguments' conventions and the NaN semantics; `Solver.solve` cross-references it; `docs/src/python.md` carries the contract for direct API users beside the batch warm-start section; the CHANGELOG states the same scope. For the other seed paths I stated the limitation rather than adding NaN rejection at their entry points, keeping the diff out of `nlp_batch` and the working-set path — say the word if you'd rather they reject.

**The default.** `WarmStartOptions::default()` now seeds from `InitOptions::default()`, and the build-time threading is extracted into `resolved_warm_options`, unit-tested directly, including that a stale warm-side copy loses to the init value (the Mehrotra case).

**The dual channel, said plainly.** Unseeded duals keep the warm path's zero, which is not the cold path's least-squares estimate; docstring, engine comment, and docs all say so now, with least-squares routing noted as a possible refinement. Both wording fixes taken (the floor-still-clamps comment; the docs' explicit-zero now mentions the `warm_start_mult_bound_push` floor).

**The alias, end to end.** A new test runs `sens_solve` on a declared model, takes `session.con_alias` from the real surgery block, and asserts the dual lands negated on the aliased row, not NaN. Doing so surfaced that `.row` files carry the objective name as a trailing entry and the surgery aliases the objective too, so the reader now guards `r < m`: an objective-keyed suffix entry can no longer index out of the dual vector.

Suites: pounce-algorithm 321, pyomo-pounce 154, green after rebuilding from the branch.

jkitchin (approved):
Approving — this clears my earlier blocking review. All 5 checks are green at `c34d4c7`.

Both sign conventions are crossed correctly, the alias path is asserted end to end against a map a real surgery block produced, and the NaN-as-unseeded contract is now documented at every surface a caller would look at and explicitly scoped to the warm-start initializer.

Three loose ends are being handled in a follow-up rather than held against this PR:

1. The `.row` trailing-objective fix is applied only in the warm-start reader (`r < nl.m`), but `_row_index(con_names)` is shared with the pin map and `_Session.mult_entry` — trimming `con_names` to the `nl.m` constraint rows at the read site fixes every consumer at once.
2. `debug_assert!` is compiled out of release builds, so the `CompoundVector` NaN case is documented rather than guarded in shipped wheels; resolving NaN generically (dense scan + compound recursion) makes the contract hold instead of asserting it cannot be violated.
3. `mod warm_threading_tests` sits above the types it tests, and `default_warm_matches_default_init` is tautological now that the default *is* `InitOptions::default().bound_mult_init_val`.

---

PR #436, opened 2026-08-01
covariance() membership from the barrier geometry (roadmap item 1)

### LINKED ISSUE #362: covariance() does not project a fitted variable pinned by an active constraint

devin-griff:
Closes #362. Item 1 of the covariance/information roadmap (#262); items 2-4 follow separately.

**The defect.** `covariance()` tested activity against NL variable bounds only, so a fitted parameter held by an active constraint *row* was treated as free: the same cap written as a bound gave `std_err = 0.0` with the nonstandard-asymptotics warning, while written as a Constraint it gave the barrier's leftover variance silently (`7.9e-6` in the issue's report; `1.8e-7` re-measured on current main, zero warnings). On this branch #362's own reproduction returns exactly `0.0`, warned, in both spellings — `test_issue_362_row_pinned_variable_projects`, verified causally by swapping only `sens.py` back to main's.

**Membership.** `covariance()` classifies bound and constraint activity through item 0's rule applied at the reduced fitted block, where a fitted parameter's curvature actually lives: in the residual-variable idiom its raw Lagrangian diagonal is zero, so the per-coordinate report honestly says `unidentified` and cannot decide membership alone. The slack-only test and its moved-bounds re-injection are gone. Dispositions: strongly active projects (zero variance, conditional, warned); weakly active is **kept** at its full finite variance, where the slack test deleted it and the raw factor would have halved it; ambiguous and unidentified stay with warnings.

**Value correction.** The fitted rows' retained barrier diagonal (`var_sigma`/`row_sigma`, new in the activity report) is subtracted from the factor's reduced Hessian: a weakly active kept parameter reports its true curvature q, not the factor's 2q. The report's Σ is the solver's scaled-space value while the factor block is natural-units (kkt_solve, pounce#128), so the correction normalizes by the objective scale — the per-row constraint scale cancels algebraically — tested against a solve with gradient-based scaling engaged.

**Row projection.** A strongly active inequality row over the fitted block pins a combination, not a coordinate: both accessor paths project on the null space of the binding normals, the matrix goes singular by one per binding row, and the warning names the constraint, the combination, and its conditional information. A bound moved onto a row (#357) is the single-coordinate case and returns exactly the variable disposition. The restricted normal is honest only while the row's support outside the fitted block is pinned (pin columns cannot move); a row reaching the fitted parameters through *free* eliminated variables pins a direction the restricted normal cannot represent — and the reduced-level ratio is equally blind — so such rows take item 0's raw classification, stay unprojected, and warn explicitly. The general treatment (the reduced normal through elimination) is item 2 territory, recorded in the roadmap. `Solver.row_normal(j)` supplies the projection direction; declaration-triggered solves set `bound_relax_factor=0` for the classifier's precondition.

**Tests** pin analytics, and each pillar is mutation-verified (the value correction, the row projection, and the reduced-level membership were each disabled in turn; only their own tests fail, and all do): the weakly active kept variance equals the unconstrained σ²(XᵀX)⁻¹ entry in both the bound and row spellings; binding `a + b <= cap` matches restricted least squares with correlation exactly −1 and a rank-one drop; an objective-scaled solve returns the same numbers; the mixed `a + r_k` row warns and stays full-rank; an inactive far bound changes nothing; Gauss-Newton agrees with projection included.

Suites: Rust 57, Python activity 26, pyomo-pounce 156, all passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,47 @@ changes.
 
 ## [Unreleased]
 
+### Changed — pyomo-pounce: `covariance()` membership from the solve's barrier geometry (#362, covariance roadmap item 1)
+
+- Bound and constraint activity on the fitted parameters now classifies
+  through the item-0 rule applied at the reduced fitted block, where a
+  fitted parameter's curvature actually lives (its raw Lagrangian diagonal
+  is zero in the residual-variable idiom, so the per-coordinate report
+  alone cannot decide membership there). The slack-only test
+  (`tol = 1e-6 * (1 + |x|)`) and its moved-bounds re-injection are gone.
+- Dispositions per the item-1 table: strongly active projects (zero
+  variance, conditional, warned); weakly active is KEPT at its full
+  finite variance, where the slack test deleted it and the raw factor
+  would have halved it; ambiguous and unidentified are kept and warned.
+- The value correction subtracts the fitted rows' retained barrier
+  diagonal (`var_sigma`, new in the activity report along with
+  `row_sigma`) from the factor's reduced Hessian: a weakly active kept
+  parameter reports its true curvature `q`, not the factor's `2q`, and
+  inactive rows shed an O(μ) drift. On kept rows Σ is at most the
+  curvature's own size, so nothing cancels; pinned coordinates and
+  binding directions never enter (excluded by the free restriction and
+  annihilated by the projection, which also annihilates the binding
+  rows' huge barrier weight exactly).
+- A strongly active inequality row whose normal touches the fitted
+  block pins a combination, not a coordinate: both accessor paths
+  (Lagrangian and Gauss-Newton) project on the null space of the
+  binding normals, the matrix goes singular by one per binding row,
+  and the warning names the constraint, the pinned combination, and
+  its conditional information. A bound moved onto a row
+  (jkitchin/pounce#357) is the single-coordinate case and returns
+  exactly the variable-bound disposition, so the two spellings of one
+  limit agree in the matrices (#362), not only in the classification.
+- Declaration-triggered solves set `bound_relax_factor=0`: the
+  classifier reads slacks as distances to the user's own bounds.
+- `Solver.row_normal(j)`: a constraint row's gradient at the converged
+  iterate in user variable order, serving the projection direction.
+- Tests pin the analytics: the weakly active kept variance equals the
+  unconstrained `σ²(XᵀX)⁻¹` entry; a binding `a + b <= cap` matches
+  restricted least squares with corr −1 and a rank drop; bound and row
+  spellings agree to 1e-6; an inactive far bound changes nothing to
+  1e-7; Gauss-Newton matches on the linear model, projection included.
+
+
 ### Fixed — feasible convex QPs reported locally infeasible on the NLP path
 
 - With `solver_selection=nlp`, POUNCE reported `Converged to a point of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,11 +49,13 @@ changes.
   equally blind, so such a row takes item 0's raw classification, is
   kept unprojected, and warns explicitly. The general treatment is
   the row's reduced normal through the elimination, item 2 territory.
-- The report's `Σ` is the solver's scaled-space value while the
-  factor block is natural-units (kkt_solve, pounce#128): the value
-  correction normalizes by the objective scale (the per-row
-  constraint scale cancels algebraically), tested against a solve
-  with gradient-based scaling engaged.
+- The report's `var_sigma`/`row_sigma` and `row_normal` follow the
+  documented natural-units contract like every other sensitivity
+  output: classification runs on the solver's scaled quantities (the
+  ratio is scale-invariant), the exported values are unscaled at the
+  boundary. Tested by pinning the weak scalar row's `Σ = 1/c²` under
+  both scaling modes and `row_normal` returning the user's own
+  coefficient.
 - Declaration-triggered solves set `bound_relax_factor=0`: the
   classifier reads slacks as distances to the user's own bounds.
 - `Solver.row_normal(j)`: a constraint row's gradient at the converged
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,21 @@ changes.
   (jkitchin/pounce#357) is the single-coordinate case and returns
   exactly the variable-bound disposition, so the two spellings of one
   limit agree in the matrices (#362), not only in the classification.
+- The restricted normal is honest only when the row's support outside
+  the fitted block is pinned (declared-parameter pin columns do not
+  count as outside: they cannot move). A binding row reaching the
+  fitted parameters through FREE eliminated variables pins a
+  direction the restricted normal cannot represent (`a + r_1 <= cap`
+  with `r_1 = y_1 - a - b x_1` pins a `b`-direction while the
+  restricted normal reads `e_a`), and the reduced-level ratio is
+  equally blind, so such a row takes item 0's raw classification, is
+  kept unprojected, and warns explicitly. The general treatment is
+  the row's reduced normal through the elimination, item 2 territory.
+- The report's `Σ` is the solver's scaled-space value while the
+  factor block is natural-units (kkt_solve, pounce#128): the value
+  correction normalizes by the objective scale (the per-row
+  constraint scale cancels algebraically), tested against a solve
+  with gradient-based scaling engaged.
 - Declaration-triggered solves set `bound_relax_factor=0`: the
   classifier reads slacks as distances to the user's own bounds.
 - `Solver.row_normal(j)`: a constraint row's gradient at the converged
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,7 +65,10 @@ changes.
   restricted least squares with corr −1 and a rank drop; bound and row
   spellings agree to 1e-6; an inactive far bound changes nothing to
   1e-7; Gauss-Newton matches on the linear model, projection included.
-
+- `_classify_ratio` mirrors the Rust rule in Python, so a drift test
+  checks the two against each other on real solves: every classified
+  entry of a `classify_activity()` report must re-derive its own
+  status from the reported `(ratio, mu)` through the Python rule.
 
 ### Fixed — feasible convex QPs reported locally infeasible on the NLP path
 
--- a/crates/pounce-py/src/solver.rs
+++ b/crates/pounce-py/src/solver.rs
@@ -422,7 +422,15 @@ impl PySolver {
     /// - `"var_sigma"`, `"row_sigma"`: ndarray of the barrier
     ///   diagonal `Σ` itself (both sides summed; 0 where nothing was
     ///   classified), the quantity item 1 of the covariance roadmap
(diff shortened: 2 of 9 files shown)


jkitchin (commented):
## Review

The design is careful and the reasoning in the comments is unusually good — the mixed-row escape hatch and the scale-cancellation argument in particular. The mutation-verified test claims match what I see: each pillar has a test that isolates it, and CI is green. `Closes #362` is in the body per the repo's closing-keyword rule.

### Issues worth addressing

**1. The new outputs violate the documented "natural units" contract, silently.** `docs/src/sensitivity.md` (units section) states *"All sensitivity outputs are in natural (unscaled) units… pounce undoes that scaling in every held-factor back-solve"*. `var_sigma`, `row_sigma`, and `row_normal` are the first exports that are **not** — they come straight off `cq.curr_*` / `cq.curr_jac_{c,d}()` in the algorithm's scaled space. The Python here proves it: it divides Σ by `nlp_scaling["obj"]`, and the row branch depends on `row_normal` returning `dg·∇g` for the `dg` cancellation to work. None of the three doc comments (`activity.rs` `var_sigma`/`row_sigma`, `pounce-py` `row_normal`) mention scaling, and `row_normal`'s reads flatly as "the gradient of user constraint row `j`". A user calling `row_normal` on a gradient-scaled solve gets a silently rescaled vector. Please state the convention in all three docstrings plus the units section of `sensitivity.md`, or unscale at the boundary.

**2. Gauss-Newton doesn't get the value correction.** `group_jacobians()` recovers `J = Z_r · minv()`, i.e. off the **uncorrected** `M`; only the Lagrangian branch uses `R_corr`. The GN parity test (`test_binding_row_projects_the_combination`) is the *strongly active row* case, where Σ is annihilated by the projection — so it can't catch this. A weakly active **bound** under `hessian="gauss-newton"` is untested and, if the barrier weight survives into `Z_r`, would report roughly a quarter of the right variance in that direction. Either add the GN counterpart of `test_weakly_active_bound_kept_with_true_variance`, or, if Σ genuinely cancels in the `Z_r = J·M` identity, say so in a comment — the asymmetry currently reads as an oversight.

**3. A row classified `unidentified` is silently ignored.** In the non-mixed row branch, `strongly_active` / `weakly_active` / `ambiguous` each warn; `unidentified` falls through with no warning and no correction. Both the variable loop and the *mixed*-row branch do warn on `unidentified`.

**4. Inactive rows keep their O(μ) weight; inactive bounds don't.** `sig_fit` is subtracted for every fitted variable regardless of status, but `R_corr -= sig_row·aaᵀ` only fires for `weakly_active`/`ambiguous` rows. So the bound and row spellings of one *inactive* limit agree only to O(μ) — and `test_inactive_bound_changes_nothing` only tests the bound spelling. `sig_row` is already computed; subtracting it for inactive rows too would make the symmetry exact.

**5. Two inversions on the most common path.** The Lagrangian branch used to return `Mc = M` directly when nothing was active; it now always computes `inv(inv(M) − diag(σ))`. For unbounded fitted parameters `var_sigma` is exactly 0, so this is a pure round-trip that costs conditioning on an ill-conditioned `M`. Worth short-circuiting when `bind_normals` is empty, `free == all`, and `sig_fit` is all-zero.

**6. `bound_relax_factor=0` is forced on every declaration-triggered solve**, not just ones that end in `covariance()` — `pounce_solver.py` reroutes any model with declarations through `sens_solve`. Removing Ipopt's default relaxation can cost convergence on models sitting exactly on bounds with round-off, and pure-sensitivity users get the change without ever calling `covariance()`. It's in the CHANGELOG but not in `docs/src/sensitivity.md`. Consider documenting it, and possibly retrying with the default on a convergence failure.

**7. Performance/scaling.** `row_normal(j)` is called for every non-equality row; each call allocates two dense vectors and does a full transpose mat-vec, and the d-block position is found by an O(`user_row`) scan over `full_g_to_c_block` — so an all-inequality model costs O(m²) index lookups plus O(m·n) traffic before any row is known to touch the fitted block. Fine for parameter estimation (residuals are equalities and get skipped), but a batched accessor, or reusing the report's existing `d_pos` scan, would avoid a surprise on wide models.

### Smaller notes

- `if na == 0.0: continue` — exact float comparison for "row does not touch the fitted block", while the `mixed` test right below correctly uses a relative tolerance. Use one there too.
- `_classify_ratio` re-implements `activity.rs::classify` in Python. It's branch-tested, but the two will drift. Exposing the rule from `pounce-py` (classify a supplied Σ, q, μ) would leave one implementation. Note also the deliberate divergence — Python's `q < floor and ri <= 1e1` guard vs Rust's unconditional `q < floor → UNIDENTIFIED` — which is explained in a comment but not in the "mirrors pounce_sensitivity::activity" docstring.
- `R_corr` is mutated inside the row loop, so a later row classifies against an already-corrected block. Deterministic, but order-dependent across multiple kept rows; a classify-then-correct two-pass would remove the coupling.
- `q_red = np.abs(diag(R_W) - sig_fit)` discards the sign, where the Rust classifier keeps `q_sign` precisely so an indefinite direction is "reported rather than hidden". Negative reduced curvature now classifies as if positive.
- **Stale comment**: `pyomo-pounce/pyomo_pounce/sens.py:144` still says *"covariance()'s active-bound projection reads NL bounds, which for these rows now hold the reader's no-bound sentinel"* — that re-injection is exactly what this PR deleted.
- **Stale docs**: `docs/src/sensitivity.md:271-276` still describes the old behavior ("a fitted parameter sits on a bound at the optimum → projected out"), with no mention of weakly-active-kept, row projection, or the conditional-information warning. The CHANGELOG is thorough; the user-facing doc wasn't updated.
- `test_inactive_bound_changes_nothing` asserts `not w` — any unrelated warning from numpy/Pyomo fails it. Filter for `"covariance:"` instead.
- Scaling coverage: `test_objective_scaling_does_not_move_the_correction` exercises `df` only, through a *bound*. The per-row `dg` cancellation in the row branch — the subtler half of the argument — has no test, and the Rust-level tests all run under `_options()` with scaling effectively inert.

### Verified, no issue found

- `row_normal`'s d-block position scan matches the report's own scatter (`activity.rs` `d_pos` ascending over non-c-block rows) — the two conventions agree.
- Zero-variance directions from a row-pinned coordinate don't produce NaN correlations: `Covariance.__init__` already suppresses and zeroes non-finite entries.
- `_free_nullspace` with an empty `free` is unreachable — the `if not free:` early return precedes it.
- No security surface; no new dependencies.

**Verdict:** sound approach, well-tested where it's tested. I'd want (1) units documentation on a public API and (2) the GN correction/test resolved before merge; the rest are reasonable follow-ups.

---

devin-griff:
All addressed, in dabbd1c:

**1. Units contract.** Resolved by unscaling at the boundary rather than documenting an exception, so the natural-units contract stays unconditionally true: `var_sigma` exports Σ/df, `row_sigma` exports Σ·dg²/df, and `row_normal` divides out the per-row scale. Classification still runs on the scaled quantities (the ratio is invariant either way); only the exported values changed, and the pyomo layer's df bookkeeping is deleted rather than annotated. All three docstrings state the convention, and the units section of `sensitivity.md` now covers the classifier's exports. The dg half of the argument you noted as untested is the new test's core: the weak scalar row pins Σ = 1/c² under both `gradient-based` and `none`, and `row_normal` must return the user's own coefficient — both fail by construction against the scaled exports.

**2. Gauss-Newton.** The asymmetry is correct, by an exact identity rather than luck: the residual rows of the K-inverse columns are J times the W-based parameter sensitivities, `Z_r = J @ M`, so `Z_r @ inv(M) = J` and the factor's barrier weight cancels regardless of Σ. The Lagrangian branch corrects `R_W` because it uses the W-based reduced Hessian; GN rebuilds from the exact J instead. Stated as a comment where the recovery happens, and pinned by the GN counterpart of the weakly-active analytic test, which matches σ²(XᵀX)⁻¹ because of the identity.

**3. `unidentified` rows** now warn like every other kept status.

**4. Inactive rows.** Here I kept the O(μ) asymmetry deliberately, because exactness collides with your point 7: subtracting an inactive row's weight requires its normal, and fetching every normal is the O(m·n) sweep. The geometric weight of an inactive row is O(μ) by the same invariance argument as everything else — the order the validation already accepts throughout — so inactive rows stay skipped, the choice is documented at the skip, and the previously missing test now pins both spellings of an inactive limit agreeing at that tolerance. If you weigh exactness over the sweep, it's a two-line change and I'll take it.

**5. Clean path** short-circuits to `Mc = M` when nothing is active, no corrections apply above noise scale, and no rows bind.

**6. `bound_relax_factor`.** Documented in `sensitivity.md`, including that it applies to every declaration-routed solve and the escape hatch: an explicit `bound_relax_factor` in `options=` wins (tested), and `covariance()` then refuses with a clear error. I stopped short of the retry-on-failure suggestion — a solve that silently re-runs under different options felt like its own surprise — but if you'd rather have the retry, say so and I'll add it with a warning when it engages.

**7. Performance.** Item-0's `row_status` now prefilters which rows fetch normals, so equality, unbounded, and inactive rows — the bulk of a wide model — cost nothing. The batched accessor stays a follow-up.

**Smaller notes:** the `na == 0` exact compare is gone (relative tolerance, with a fully-outside row handled as the extreme mixed case); the two-pass classify-then-correct removes the row-order coupling; the |q| classification is deliberate parity with the Rust rule, now said in a comment, with indefiniteness surfacing through the existing negative-variance warning; the stale re-injection comment and the stale dispositions docs are rewritten; the `not w` assertion filters for `covariance:`; the scaling coverage gap is the new natural-units test above. On `_classify_ratio` I kept the Python mirror for this PR and documented the deliberate `q < floor` divergence in its docstring — exposing the rule from pounce-py so one implementation serves both is queued for item 2, which reworks that surface anyway.

Suites: pyomo-pounce 167, Rust 57, Python activity 27; the value-correction mutation still kills exactly its three tests through the restructure.

jkitchin:
@devin-griff is this ready for review / merge (I can fix the conflicts)

Issue #390, opened 2026-07-28
Equality rows still have no scale-relative runtime feasibility measure (residual of #387)


jkitchin:
Residual of #387 (defect 2), split out after PR #389 fixed the linear case at
the DOF gate.

Equality rows still have no scale-relative *runtime* feasibility measure.
POUNCE folds an equality's right-hand side into `c(x) = 0`, so `|c_i|` *is*
the violation and carries no independent magnitude —
`IpoptCq::curr_relative_primal_infeasibility_max` (added in #386) therefore
skips the `c` block entirely, and every runtime feasibility decision on an
equality row is against an absolute tolerance.

## Why this still matters after #389

PR #389 handles contradictions that **linear bound propagation** can prove, by
certifying them at the DOF-gate failure path. Out of its reach, and still
scale-dependent, are:

- genuinely **nonlinear** equality contradictions (e.g. `x^2 == 0.2*s` with
  `x^2 == 0.8*s` shapes, or anything FBBT-invisible on the default path), where
  the verdict comes from the restoration phase's infeasibility test at
  runtime; and
- any equality-constrained model where the solve *does* run and acceptance /
  infeasibility declarations compare `|c_i|` against absolute thresholds.

Down-scaling such rows shrinks `|c_i|` under the absolute tolerance and flips
a true infeasibility verdict to `Solve_Succeeded`; the fix that made
inequality rows scale-relative cannot reach equality rows as built.

## Sketch (from #387)

Plumb the **pre-fold RHS** (or an equivalent declared row magnitude,
`max(|b_i|, 1)`-style) to where the measure is computed — the same
"declared, not live" pattern `IpoptNlp::declared_d_bounds` established for
inequality rows in #386. Then let
`curr_relative_primal_infeasibility_max` cover the `c` block with
`|c_i| / max(1, |b_i|)`.

Note the same absolute-clamp question will surface here as in the
witness-refutation discussion (see the sibling issue on the remaining
`inf_eq` cells): a homogeneous row (`b_i = 0`) has no declared magnitude and
must keep the absolute floor, or float-noise residuals become "violations".

## Acceptance

A nonlinear equality-contradiction model added to
`pyomo-pounce/tests/test_scale_invariance.py` (the linear `inf_eq` is now
covered by the presolve route and cannot exercise this) whose wrong-cell
count drops when the measure lands, ratcheted via `BASELINE_WRONG`.

Refs: #387, #386, #389.

Issue #387, opened 2026-07-28
Equality rows have no scale-relative feasibility measure — inf_eq is 13/13 FAIL at every scale


jkitchin:
Follow-up to #385 / #386 (the scale-invariance migration), and the one model
class it deliberately leaves untouched.

`inf_eq` in `pyomo-pounce/tests/test_scale_invariance.py` — two contradictory
equalities, `x == 0.2` and `x == 0.8` — reports `FAIL` (5xx) at **all 13 row
scalings**, baseline `13/13` wrong. Two distinct defects stack:

1. **At unit scale it is already wrong**: a two-equality contradiction on one
   variable should be `INFEAS` (it is about as provable as infeasibility
   gets — presolve certifies the inequality analogue), yet it exits in the
   5xx failure band. Scale-invariantly wrong, but wrong.
2. **No scale-relative measure can reach it as built**: POUNCE folds an
   equality's right-hand side into `c(x) = 0`, so `|c_i|` *is* the violation
   and carries no independent magnitude. `IpoptCq::
   curr_relative_primal_infeasibility_max` (added in #386) therefore skips
   the `c` block entirely. Fixing this needs the pre-fold RHS (or an
   equivalent declared row magnitude) plumbed to where the measure is
   computed — the same "declared, not live" pattern
   `IpoptNlp::declared_d_bounds` established for inequality rows.

Worth checking first whether presolve should simply certify this model: both
rows are linear equalities over a box, and bound propagation handles the
`x >= 0.7`-style contradiction today (201). If Phase 1 learns to propagate
equality pairs, cell 1 collapses without touching the solver, and only the
genuinely nonlinear equality contradictions remain for the measure.

PR #443, opened 2026-08-01
Port IpRestoConvCheck layer 2: a verdict when the restoration sub-problem converges

### LINKED ISSUE #438: Restoration has no verdict when its sub-problem converges: IpRestoConvCheck layer 2 is missing
Found while fixing the restoration-doorway port gap (`ACCEPTABLE_POINT_REACHED`). That fix stops pounce *entering* restoration from an acceptable point; this issue is about what happens once it is inside and cannot succeed.

## The gap

`IpRestoConvCheck.cpp:CheckConvergence` is two layers. Pounce ports the first and omits the second.

**Layer 1 — can the trial point leave restoration?** The `kappa_resto` reduction guard, the square-problem fast path, and the outer-filter acceptance test. Pounce has all of this in `crates/pounce-restoration/src/conv_check.rs:75-125`, faithfully, with upstream line references.

**Layer 2 — has the restoration sub-problem itself converged?** Upstream, `IpRestoConvCheck.cpp:200-240`:

```cpp
if( status == CONTINUE )
{
   status = OptimalityErrorConvergenceCheck::CheckConvergence(false);
   if( status == CONVERGED || status == CONVERGED_TO_ACCEPTABLE_POINT )
   {
      if( orig_trial_inf_pr <= 1e2 * IpData().tol() && IpData().tol() > 1e-1 * orig_ip_data->tol() )
      {   // tighten the resto tolerance once more and keep going
          IpData().Set_tol(1e-2 * IpData().tol());
          status = CONTINUE;
      }
      else if( orig_ip_cq->IsSquareProblem() && orig_trial_inf_pr <= orig_constr_viol_tol_ )
      {   status = CONVERGED_TO_ACCEPTABLE_POINT; }
      else if( orig_trial_inf_pr <= 1e2 * IpData().tol() )
      {   THROW_EXCEPTION(RESTORATION_CONVERGED_TO_FEASIBLE_POINT, ...); }
      else
      {   THROW_EXCEPTION(LOCALLY_INFEASIBLE, ...); }
   }
}
```

Pounce's `check_convergence` never asks the question. Its final arm is:

```rust
if acceptable_to_outer() {
    RestoConvergenceStatus::Converged
} else {
    RestoConvergenceStatus::Continue
}
```

So a restoration whose sub-problem has fully converged — which is the moment it has provably done everything it can — is indistinguishable from one still making progress. It returns `Continue`, forever, until an iteration cap.

## What it costs

Mittelmann `qcqp1000-1nc` (n=1000, m=154). The outer enters restoration at iteration 188 with `orig_curr_inf_pr = 5.96e-8`. The kappa target is therefore `0.9 x 5.96e-8 ~ 5.4e-8`. Restoration immediately moves the iterate to `theta ~ 5e-3` and never gets back, so:

```
orig_trial_inf_pr (5e-3)  >  kappa_resto * orig_curr_inf_pr (5.4e-8)   ->  Continue
```

is true at every subsequent iteration, unconditionally. Meanwhile the sub-problem is sitting at its own KKT point — from the run log, `inf_du = 4.28e-09` and `lg(mu) = -9.0` pinned, objective creeping in the seventh digit, `||d|| ~ 6.9e3`:

```
2963r-1.7953054e+06 5.05e-03 4.28e-09  -9.0 6.96e+03    - 1.00e+00 1.00e+00h  1
2964r-1.7953027e+06 6.82e-03 4.28e-09  -9.0 6.96e+03    - 1.00e+00 1.00e+00h  1
2965r-1.7953000e+06 6.74e-03 4.28e-09  -9.0 6.96e+03    - 1.00e+00 1.00e+00h  1
```

2780 restoration iterations chasing a target five orders out of reach, on a problem whose optimum the outer already held at iteration 187. Upstream's layer 2 would have rendered `LOCALLY_INFEASIBLE` (or `RESTORATION_CONVERGED_TO_FEASIBLE_POINT`) the first time the sub-problem converged, and the line search would then have restored the acceptable point (`IpBacktrackingLineSearch.cpp:611`).

## Why the existing caps do not cover it

`conv_check.rs:41-42` sets `maximum_iters = 3000` and `maximum_resto_iters = 3000`. Both are hit *eventually* — the run above was at 2965 when the 300 s wall clock killed it — but they are backstops sized to the outer iteration budget, not a verdict. They convert "restoration cannot succeed" into "the whole solve ran out of iterations", which is the wrong status and three orders of magnitude too much work.

## Suggested fix

Port layer 2. The pieces are all present: `RestoConvCheck` already holds `orig_constr_viol_tol`, the inner solver already owns an `OptErrorConvCheck` for the sub-problem, and `RestorationOutcome::LocallyInfeasible` already exists and is already handled by the caller (`ipopt_alg.rs:2787`). What is missing is the query and the four-way verdict.

Note the tolerance-tightening arm is bounded by construction (`IpData().tol() > 1e-1 * orig_tol` with a `1e-2` factor per round, so at most a couple of rounds) — worth keeping that guard, since without it the arm is an infinite loop.

## Not a bounded-restoration budget

Worth recording, since it was the first thing considered: an iteration budget on restoration is the wrong instrument. It answers "how long have we spent" when the question is "can this still succeed", it is not scale-free, and its constant has no derivation, so it ends up tuned to whatever instance prompted it. Upstream carries no such budget — restoration is bounded by having a defined success target plus a total verdict at sub-problem convergence. Layer 2 *is* the bound.

jkitchin:
Fixes #438

## Problem

A restoration whose sub-problem has converged — the moment it has provably done everything it can — was indistinguishable from one still making progress. It reported the sub-problem's own status and let the outer re-enter, or, when the κ_resto target was out of reach, ground on until an iteration cap converted "restoration cannot succeed" into "the solve ran out of iterations".

`issue_372_infeasible_bounds.nl` (`0 <= x <= 0.6` with `x >= 0.7`) is the smallest instance of the shape available in-tree. Its user-facing status was already right, but only because a post-hoc heuristic reconstructed it:

| | status | route to the status | outer iters |
|---|---|---|---|
| before | local infeasibility | `tiny_step_locally_infeasible` heuristic in `resto_inner_solver` (inner exits `StopAtTinyStep`, gate infers the diagnosis from the terminal status + KKT residual) | 56 |
| after | local infeasibility | layer-2 verdict at inner iteration 10, on a point the sub-solve itself certified | 56 |

No oracle comparison: I have no Ipopt binary here, and the instance from the issue (`qcqp1000-1nc`) needs an AMPL CE licence to translate, so I could not run it — see **Tests** for what that leaves unpinned.

## Cause

`IpRestoConvCheck::CheckConvergence` is two layers. Layer 1 — the κ_resto reduction guard, the square-problem fast path, and the outer-filter acceptance test — answers *can the trial point leave restoration?*. Layer 2 (`IpRestoConvCheck.cpp:200-240`) asks whether the restoration *sub-problem* has converged in its own right and, when it has, renders one of four verdicts.

Pounce ported layer 1 faithfully and omitted layer 2. In the live path (`RestoConvCheckAdapter::check_convergence_with_state`, the adapter `run_inner_resto` installs as the nested IPM's `conv_check`) step 3 delegated to the sub-problem's own `OptErrorConvCheck` and returned whatever it said. Nothing ever asked the layer-2 question.

That matters most exactly when layer 1 can never succeed. The outer enters restoration at a nearly-feasible point, so the κ target is a fraction of an already-tiny infeasibility; restoration moves the iterate away; the reduction test then fails at every subsequent iteration unconditionally. The remaining exits are the `maximum_iters` / `max_resto_iter` caps — the wrong status, after orders of magnitude more work than the diagnosis needed. Layer 2 *is* the bound.

## Fix

The four arms are factored into a pure `resto_orig_verdict` (`conv_check.rs:80`) so each is unit-testable without a live NLP, and rendered inside the restoration sub-solve — in `RestoConvCheckAdapter` (the live path) and in the scalar `RestoConvCheck` the port is documented against. `run_inner_resto` treats an inner `LocalInfeasibility` as an authoritative verdict.

The existing post-hoc gates (`strict`/`alt`/`cycle`/`step_failure`/`tiny_step`) stay. They cover the exits where the inner never gets to render a verdict at all — line-search failure, step explosion, tiny step, iteration cap — and layer 2 does not subsume them.

The tightening arm keeps upstream's `tol > 1e-1 · orig_tol` guard. Without it the arm is an infinite loop rather than the at-most-one-round step it is meant to be; `layer2_tightening_arm_is_bounded_by_construction` pins that as a property rather than as a fixed count.

The violation the arms compare against is measured **unscaled**, because their thresholds (`1e2 · tol`, `constr_viol_tol`) are absolute user-facing magnitudes. The κ guard keeps its scaled measurement, where the ratio cancels the scaling anyway. This is the same units trap documented at `eval_orig_inf_pr_at_inner_curr`, and the locally-infeasible arm carries the same guard `resto_inner_solver` applies to its gates: never claim infeasibility at a point the solver's own convergence test would call feasible.

Two deliberate deviations, both scoped so this changes only the arm that had no verdict at all:

* **converged-to-a-feasible-point.** Upstream throws `RESTORATION_CONVERGED_TO_FEASIBLE_POINT`, which surfaces as a restoration failure. Pounce hands the recovered — and by that arm's own test, feasible — point back to the outer phase. That is what it already did, and discarding a feasible point to report a failure is strictly worse.
* **square problems are exempt from the locally-infeasible arm.** `MinC_1NrmRestorationPhase` returns the recovered point unconditionally on a square problem (`IpRestoMinC_1Nrm.cpp:357-371`), and short-circuiting that regresses PFIT3/PFIT4 — the same carve-out `strict_locally_infeasible` already makes, for the same reason.

Not attempted: a bounded-restoration iteration budget. The issue rules it out and I agree — it answers "how long have we spent" when the question is "can this still succeed", it is not scale-free, and its constant would end up tuned to whichever instance prompted it.

`POUNCE_DBG_RESTO_LAYER2=1` with `RUST_LOG=pounce::restoration=debug` traces the verdict, alongside the existing `POUNCE_DBG_RESTO_KAPPA` / `POUNCE_DBG_RESTO_LOCINF`.

## Blast radius

Measured over the 12 CLI `.nl` fixtures that reach the solver, against parent `c35edca`. Bit-identical on 11 of 12 — status and outer iteration count both. The one difference:

| fixture | before | after | note |
|---|---|---|---|
| `deb7` | 180 iters, `Error in step computation` | 189 iters, `Error in step computation` | tightening arm fires once (tol 1e-8 → 1e-10) then terminates; same pre-existing status |

`deb7` is the only fixture that exercises the tightening arm, and it is the arm's expected cost: layer 1 refused, the orig NLP was near-feasible (`inf_pr = 8.0e-13`), so upstream spends one round chasing a 100× tighter sub-problem tolerance before declaring the point feasible.

Not a corpus sweep — the Mittelmann harness needs an AMPL CE licence to translate `.mod` → `.nl` and the instances are gitignored, so I could not run one. Full workspace suite (171 test binaries), `cargo fmt --all -- --check`, and the CI clippy gate are all clean.

## Tests

`crates/pounce-cli/tests/issue_438_resto_layer2_verdict.rs` — three end-to-end tests through the binary. Verified against parent `c35edca` (parent's `crates/pounce-restoration/src/` checked out under the new test file):

* `sub_problem_convergence_renders_a_locally_infeasible_verdict` — **fails on the parent**, for the stated reason: no verdict is rendered at all, so the trace is empty where it should read `LocallyInfeasible`. This is the one that pins the fix.
* `locally_infeasible_verdict_surfaces_as_the_infeasibility_status` — passes on the parent. It pins that the user-facing status did not move, not the fix.
* `no_verdict_while_the_sub_problem_is_still_making_progress` — passes on the parent (trivially, the trace does not exist there). It pins that layer 2 stays silent on a restoration that leaves via layer 1, which is the regression this could plausibly cause.

Unit tests in `pounce-restoration` cover the four arms individually, the tightening arm's boundedness, layer 1's precedence over layer 2, and that an unconverged sub-problem still yields `Continue`.

**Deliberately not pinned, and a real gap:** whether this bites on `qcqp1000-1nc`, the instance in the issue. Layer 2 fires only once the sub-problem's own convergence check reports convergence, and in that trace the sub-problem ran 2965 iterations without doing so. That is consistent with two readings — it converges late (layer 2 now catches it), or it never converges by pounce's `nlp_err` measure (layer 2 is correct but silent there, and the residual gap is in the sub-problem's own convergence check). I could not distinguish them without the instance. Worth a confirming run before closing #438.

---

- [x] Tests fail on the parent commit for the stated reason (one of the three; the other two are non-regression pins and say so above)
- [ ] Book page under `docs/src/` updated — none applies; no user-facing option or status is added or changed

---
_Generated by [Claude Code](https://claude.ai/code/session_01RcQrUMw3gs5D3fDS6UoK31)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,38 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — restoration had no verdict when its sub-problem converged
+
+- `IpRestoConvCheck::CheckConvergence` is two layers, and pounce ported
+  only the first. Layer 1 — the κ_resto reduction guard, the
+  square-problem fast path, and the outer-filter acceptance test —
+  answers *can the trial point leave restoration?*. Layer 2
+  (`IpRestoConvCheck.cpp:200-240`) answers *has the restoration
+  sub-problem itself converged?*, and it is the only thing that bounds a
+  restoration which can never satisfy layer 1 (#438).
+- Without it, a sub-problem sitting at its own KKT point — the moment it
+  has provably done everything it can — was indistinguishable from one
+  still making progress. When the κ_resto target is out of reach (the
+  outer enters restoration at a nearly-feasible point, restoration moves
+  the iterate away, and the reduction test then fails at every subsequent
+  iteration), the only remaining exits were the `maximum_iters` /
+  `max_resto_iter` caps: the wrong status, reported after orders of
+  magnitude more work than the diagnosis needed.
+- Layer 2's four-way verdict is now rendered inside the restoration
+  sub-solve: tighten the sub-problem tolerance once and continue,
+  converged-to-acceptable on a feasible square problem,
+  converged-to-a-feasible-point, or `LOCALLY_INFEASIBLE`. The tightening
+  arm keeps upstream's `tol > 1e-1 · orig_tol` guard, without which it is
+  an infinite loop.
+- Two deliberate deviations from upstream, both scoped so that #438
+  changes only the arm that had no verdict at all: the
+  converged-to-a-feasible-point arm hands the recovered point back to the
+  outer phase instead of throwing (upstream reports a restoration
+  failure), and square problems are exempt from the locally-infeasible
+  arm, matching the carve-out the post-hoc detection in
+  `resto_inner_solver` already made for them.
+- Set `POUNCE_DBG_RESTO_LAYER2=1` with `RUST_LOG=pounce::restoration=debug`
+  to trace the verdict.
 ### Fixed — port gap: no `ACCEPTABLE_POINT_REACHED` at the restoration doorway
 
 - **Upstream refuses to enter restoration from an acceptable point; pounce did
--- a/crates/pounce-cli/tests/issue_438_resto_layer2_verdict.rs
+++ b/crates/pounce-cli/tests/issue_438_resto_layer2_verdict.rs
@@ -1,0 +1,129 @@
+//! Regression test for pounce#438: the restoration convergence check had
+//! no layer 2.
+//!
+//! `IpRestoConvCheck::CheckConvergence` is two layers. Layer 1 — the
+//! `kappa_resto` reduction guard, the square-problem fast path, and the
+//! outer-filter acceptance test — answers *can the trial point leave
+//! restoration?*. Layer 2 (`IpRestoConvCheck.cpp:200-240`) answers the
+//! complementary question, *has the restoration sub-problem itself
+//! converged?*, and renders one of four verdicts when it has.
+//!
+//! Pounce ported layer 1 and omitted layer 2, so a restoration whose
+//! sub-problem had provably done everything it could was indistinguishable
+//! from one still making progress: it reported the sub-problem's own
+//! status and let the outer re-enter, or — when the kappa target was out
+//! of reach, which it is at every iteration once restoration moves the
+//! iterate away from a nearly-feasible entry point — ground on until an
+//! iteration cap turned "restoration cannot succeed" into "the solve ran
+//! out of iterations".
+//!
+//! This pins the verdict end-to-end through the binary, because the unit
+//! tests in `pounce-restoration` cover [`resto_orig_verdict`]'s arms
+//! individually and only a real solve proves the query is wired to the
+//! nested IPM's convergence check at all.
+//!
+//! `issue_372_infeasible_bounds.nl` (`0 <= x <= 0.6` with `x >= 0.7`) is
+//! used because it is a one-inequality contradiction: restoration drives
+//! the sub-problem to its KKT point in ten inner iterations at an
+//! original-NLP violation of `1e-1`, which is exactly layer 2's
+//! locally-infeasible arm and nothing else.
+
+use std::path::PathBuf;
+use std::process::Command;
+
+fn pounce_exe() -> PathBuf {
+    PathBuf::from(env!("CARGO_BIN_EXE_pounce"))
+}
+
+fn fixture_nl(name: &str) -> PathBuf {
+    let mut p = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    p.push("tests");
+    p.push("fixtures");
+    p.push(name);
+    p
+}
+
+/// Run a fixture with the layer-2 trace enabled and return `(stdout, stderr)`.
+fn run_with_layer2_trace(fixture: &str) -> (String, String) {
+    let output = Command::new(pounce_exe())
+        .arg(fixture_nl(fixture))
+        .arg("--no-sol")
+        .env("POUNCE_DBG_RESTO_LAYER2", "1")
+        .env("RUST_LOG", "pounce::restoration=debug")
+        .env("NO_COLOR", "1")
+        .output()
+        .expect("spawn pounce");
+    (
+        String::from_utf8_lossy(&output.stdout).into_owned(),
+        String::from_utf8_lossy(&output.stderr).into_owned(),
+    )
+}
+
+/// Every distinct layer-2 verdict the run logged, in first-seen order.
+fn logged_verdicts(stderr: &str) -> Vec<String> {
+    let mut seen: Vec<String> = Vec::new();
+    for line in stderr.lines() {
+        let Some(rest) = line.split("[PN_RESTO_LAYER2] verdict=").nth(1) else {
+            continue;
+        };
+        let v = rest
+            .split_whitespace()
+            .next()
+            .unwrap_or_default()
+            .to_string();
+        if !seen.contains(&v) {
+            seen.push(v);
+        }
+    }
+    seen
+}
+
+/// The core of #438: layer 2 is reached, and on a model whose restoration
+/// sub-problem converges at a violated point it renders the
+/// locally-infeasible verdict rather than no verdict at all.
+#[test]
+fn sub_problem_convergence_renders_a_locally_infeasible_verdict() {
+    let (_, stderr) = run_with_layer2_trace("issue_372_infeasible_bounds.nl");
+    assert_eq!(
+        logged_verdicts(&stderr),
+        vec!["LocallyInfeasible".to_string()],
+        "pounce#438: once the restoration sub-problem converges, layer 2 \
+         must render a verdict; a violation of 1e-1 against tol=1e-8 is \
+         upstream's `LOCALLY_INFEASIBLE` arm \
+         (`IpRestoConvCheck.cpp:240`).\n--- stderr ---\n{stderr}",
+    );
+}
+
+/// The verdict must reach the user as the status it means. Before #438
+/// this model was diagnosed by a post-hoc heuristic in
+/// `resto_inner_solver` that reconstructed the same conclusion from the
+/// inner solver's terminal status and KKT residual (the `tiny_step` gate,
+/// added for gh #372); the verdict now supplies it directly, and the
+/// user-facing status is unchanged.
+#[test]
+fn locally_infeasible_verdict_surfaces_as_the_infeasibility_status() {
+    let (stdout, stderr) = run_with_layer2_trace("issue_372_infeasible_bounds.nl");
+    let combined = format!("{stdout}{stderr}");
+    assert!(
+        combined.contains("local infeasibility"),
+        "expected the local-infeasibility exit status, got:\n{combined}",
+    );
+}
+
+/// Layer 2 must stay silent while restoration is still working — it is
+/// gated on the sub-problem's *own* convergence check, not on elapsed
+/// iterations. `pooling_rt2stp.nl` enters restoration, leaves it via
+/// layer 1, and solves; no verdict should ever be rendered.
+#[test]
+fn no_verdict_while_the_sub_problem_is_still_making_progress() {
+    let (stdout, stderr) = run_with_layer2_trace("pooling_rt2stp.nl");
+    assert!(
+        logged_verdicts(&stderr).is_empty(),
+        "layer 2 fired on a restoration that left via layer 1: {:?}",
+        logged_verdicts(&stderr),
+    );
+    assert!(
+        format!("{stdout}{stderr}").contains("Optimal Solution Found"),
+        "fixture is expected to solve; layer 2 must not have changed that",
+    );
+}
--- a/crates/pounce-restoration/src/conv_check.rs
+++ b/crates/pounce-restoration/src/conv_check.rs
@@ -79,23 +79,41 @@ pub enum RestoOrigVerdict {
 /// loop, which is why it is part of the port rather than an optimisation.
 pub fn resto_orig_verdict(
     orig_trial_inf_pr: f64,
+    orig_trial_rel_inf_pr: f64,
     tol: f64,
     orig_tol: f64,
     orig_constr_viol_tol: f64,
+    rel_viol_threshold: f64,
     is_square_problem: bool,
 ) -> RestoOrigVerdict {
+    // Every arm but the last asks "is this nearly feasible for the orig NLP?"
+    // and upstream asks it with an absolute threshold alone. That is not
+    // scale-invariant: `s*g(x) >= s*b` is the same constraint for every s > 0,
+    // but its residual carries s. Measured on `inf_372` (0 <= x <= 0.6 with
+    // s*x >= s*0.7, infeasible by 14% of the row at every s): the violation is
+    // 1.0e-1 at s = 1 and 1.0e-9 at s = 1e-8, so the same model clears
+    // `1e2 * tol` at small s and the arms read a decisively infeasible point as
+    // nearly feasible. gh #390 (residual of #387) settled this for the outer
+    // check; the same invariant is needed the moment layer 2 judges
+    // feasibility. Rows with no declared magnitude contribute 0 and fall back
+    // to the absolute test, which is already invariant for them.
+    let nearly_feasible =
+        orig_trial_inf_pr <= 1e2 * tol && orig_trial_rel_inf_pr <= rel_viol_threshold;
     // `IpRestoConvCheck.cpp:212` — nearly feasible for the orig NLP, and
     // there is tolerance budget left to spend chasing the rest.
-    if orig_trial_inf_pr <= 1e2 * tol && tol > 1e-1 * orig_tol {
+    if nearly_feasible && tol > 1e-1 * orig_tol {
         return RestoOrigVerdict::TightenAndContinue;
     }
     // `IpRestoConvCheck.cpp:222` — a square problem has nothing to
     // optimise, so any point feasible to `constr_viol_tol` is the answer.
-    if is_square_problem && orig_trial_inf_pr <= orig_constr_viol_tol {
+    if is_square_problem
+        && orig_trial_inf_pr <= orig_constr_viol_tol
+        && orig_trial_rel_inf_pr <= rel_viol_threshold
+    {
         return RestoOrigVerdict::ConvergedToAcceptable;
     }
     // `IpRestoConvCheck.cpp:227`.
-    if orig_trial_inf_pr <= 1e2 * tol {
+    if nearly_feasible {
         return RestoOrigVerdict::ConvergedToFeasiblePoint;
     }
     // `IpRestoConvCheck.cpp:240`.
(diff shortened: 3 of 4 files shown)


jkitchin:
## Ran the open question. **Reading two is correct: layer 2 never fires on `qcqp1000-1nc`.**

I filed #438 and had the instance and the mittelmann corpus locally, so I ran the confirming experiment this PR asks for. Built `9750d393` in a worktree, ran `qcqp1000-1nc.nl` with `POUNCE_DBG_RESTO_LAYER2=1 RUST_LOG=pounce::restoration=debug`, 320 s cap.

**Zero layer-2 verdicts. 3001 iteration lines, 2814 of them restoration. Timed out.**

```
2997r-1.7952309e+06 6.83e-03 4.28e-09  -9.0 7.00e+03  - 1.00e+00 1.00e+00h  1
2998r-1.7952296e+06 5.16e-03 4.28e-09  -9.0 7.00e+03  - 1.00e+00 5.00e-01h  2
2999r-1.7952269e+06 6.83e-03 4.28e-09  -9.0 7.00e+03  - 1.00e+00 1.00e+00h  1
```

Byte-for-byte the same grind as on the parent — `inf_du` pinned at `4.28e-09`, `lg(mu)` at the `-9.0` floor, `‖d‖ ≈ 7e3`, `inf_pr` oscillating in `5–7e-3`, objective creeping in the seventh digit.

**Positive control, so the silence means something.** The same binary and the same environment on this PR's own fixture emits the verdict as documented:

```
$ POUNCE_DBG_RESTO_LAYER2=1 RUST_LOG=pounce::restoration=debug \
    pounce issue_372_infeasible_bounds.nl --no-sol
DEBUG pounce::restoration: [PN_RESTO_LAYER2] verdict=LocallyInfeasible orig_inf_pr=9.999998e-2 ... square=false tightenings=0
```

Two lines there, zero on `qcqp1000-1nc`. The instrumentation is live and the invocation is right, so this is a real negative and not a missing flag.

### What that means

Of the two readings in the PR body, it is the second: **layer 2 is correct here but structurally silent.** It is gated on the sub-problem's own convergence check reporting convergence, and on this instance that check never fires in 2814 restoration iterations — so the gate never opens. The residual gap is in the restoration sub-problem's own convergence, not in layer 2.

`‖d‖ ≈ 7e3` at `mu` floor with full steps accepted is the shape of a sub-solve that is wandering rather than converging, which is consistent with that reading and worth its own issue. I would not close #438 on the strength of `qcqp1000-1nc`; I would close it on what this PR actually fixes, which is the missing verdict, and file the sub-problem convergence separately.

### Interaction with #442 — worth knowing before either merges

#442 ports the *other* half of upstream's restoration handling: `ACCEPTABLE_POINT_REACHED` at the doorway (`IpBacktrackingLineSearch.cpp:568`), which refuses to *enter* restoration from an acceptable point. On `qcqp1000-1nc` that fires at iteration 187 and the instance never reaches restoration at all — `Solved To Acceptable Level`, 300 s → 22 s.

So the two are complementary, not overlapping:

| | fixes | on `qcqp1000-1nc` |
|---|---|---|
| #442 | never enter restoration from an acceptable point | fires at iter 187, resolves it |
| #443 | render a verdict once inside and converged | never fires |

Mechanically they are clean: **no conflict** (`git merge-tree` reports none), and disjoint files apart from `CHANGELOG.md` — #443 is `pounce-restoration/`, #442 is `pounce-algorithm/ipopt_alg.rs` + `pounce-cli/`.

One consequence for whoever merges second: once #442 is in, `qcqp1000-1nc` stops being a witness for layer 2, because it no longer enters restoration. If you want the instance to keep exercising this path in future, that has to be a deliberate test construction rather than something the benchmark corpus happens to provide.

### On the PR itself

The blast-radius work is the right shape — 11 of 12 fixtures bit-identical including outer iteration count, and the one difference (`deb7`, tightening arm firing once, same pre-existing status) explained rather than waved at. The three-way test split is honest about which test pins the fix and which two are non-regression pins. Both deviations from upstream are argued rather than asserted, and the square-problem carve-out matches the existing `strict_locally_infeasible` precedent.

I have not reviewed the code itself here — this comment is only the experimental answer to the open question.

jkitchin:
## CI is red, and it is a real regression. It is the **tightening arm**, and it is scale-dependent.

`wheel smoke (pyomo-pounce)` → `test_scale_invariance.py` fails, deterministically, on both runs. **It also failed on the pre-rebase commit `9750d393`, so this is not from the rebase** — the PR has never been green. The Rust suite passes; this is the Python-side ratchet.

```
scale invariance regressed — a model's verdict is wrong at more row scalings than before.
    inf_372:   0 -> 3 wrong cells    INFEAS FAIL FAIL FAIL INFEAS ...
    inf_clear: 0 -> 2 wrong cells    INFEAS FAIL FAIL INFEAS ...
    inf_two:   0 -> 2 wrong cells    INFEAS FAIL FAIL INFEAS ...
```

`FAIL` is `solve_result_num >= 500`. Three infeasible models now report a **solver failure** instead of certifying infeasibility, at `k = -10, -8, -6` in `KS = range(-12, 13, 2)` — i.e. when the row is multiplied by `1e-10 ... 1e-6`. `inf_372` is this PR's own fixture.

### Which arm

Reproduced locally on the rebased branch and traced with `POUNCE_DBG_RESTO_LAYER2=1`, one `.nl` per scaling of `_inf_372` (`0 <= x <= 0.6` with `s*x >= s*0.7`):

| `s` | layer-2 verdict | status | correct |
|---|---|---|---|
| `1e-12` | *(none — never reaches the arm)* | local infeasibility | yes |
| `1e-10` | **`TightenAndContinue`** | **`Restoration Failed!`** | **no** |
| `1e-8` | **`TightenAndContinue`** | **`Restoration Failed!`** | **no** |
| `1e-6` | **`TightenAndContinue`** | **`Restoration Failed!`** | **no** |
| `1` | `LocallyInfeasible` | local infeasibility | yes |

Not the locally-infeasible arm — the **tightening** arm. The trace, same model, two scalings:

```
s=1     verdict=LocallyInfeasible    orig_inf_pr=9.999998e-2  orig_inf_pr_scaled=9.999998e-2  tol=1.000000e-8
s=1e-8  verdict=TightenAndContinue   orig_inf_pr=1.005193e-9  orig_inf_pr_scaled=1.005193e-9  tol=1.000000e-8
```

### Mechanism

The arm's guard is upstream's, `orig_trial_inf_pr <= 1e2 * tol`, i.e. `<= 1e-6` at the default `tol`. The violation is `0.1` in the model's natural units. Multiply the row by `1e-8` and that same violation reads `1.005e-9`, which slips under the threshold — so a model that is infeasible by a wide margin looks *nearly feasible*, the arm tightens instead of rendering `LocallyInfeasible`, exhausts its rounds, and ends as `Restoration Failed`.

`1e2 * tol` is an absolute threshold, and the quantity it is compared against carries the user's own row scaling. `s*x >= s*0.7` is the same problem for every `s > 0`, which is exactly what the harness exists to enforce.

This contradicts the reasoning in the PR body:

> The violation the arms compare against is measured **unscaled**, because their thresholds (`1e2 · tol`, `constr_viol_tol`) are absolute user-facing magnitudes.

Measuring unscaled is what makes it scale-*dependent* here — the "unscaled" magnitude is precisely where the user's row scaling lives. Note also that in this trace `orig_inf_pr` and `orig_inf_pr_scaled` are **identical** at both scalings, so the scaled/unscaled distinction the argument rests on does not separate these cases at all.

### Is this just upstream's behaviour?

Yes — upstream has the same absolute test, so the port is faithful. But faithful is not sufficient here: pounce carries a scale-invariance ratchet upstream has no equivalent of, and its baselines for these three models are `0`. The tree has been deliberately migrating these gates toward scale-aware tolerances (`feat/scale-aware-tolerance-primitive`, `feat/resto-gates-scale-aware`, `test/scale-invariance-harness`), and `pounce_common::tolerance::is_negligible(value, scale, tol)` already exists for exactly this shape — widening an absolute tolerance by the magnitude of the thing being measured.

So the arm needs the violation compared against the row's own magnitude rather than against a bare `1e2 * tol`. Porting upstream's constant unchanged is what reintroduced the scale dependence the ratchet is there to prevent.

Two smaller notes from the same trace:

* `s = 1e-12` escapes only because it never reaches the arm at all (no verdict logged), not because the guard holds there. The correct answer at that scaling comes from a pre-existing gate.
* The layer-2 verdict at `s = 1` fires twice with identical values, which is expected given the arm is consulted per inner iteration, but worth a glance if the tightening counter is meant to advance between them.

### Not merging

I was asked to merge this on green and it is not green. The rebase onto `main` (post-#442) is done and clean — `MERGEABLE` at `36cdf9cd`, CHANGELOG conflict resolved keeping both entries, and #442's tests plus this PR's own three tests all pass locally (972 Rust tests, 0 failures). The only thing standing in the way is this regression, which predates the rebase and is in the fix itself.

jkitchin:
## Proposed fix, plus one thing I got wrong trying to implement it

Following up my diagnosis above (tightening arm, absolute threshold, scale-dependent). I attempted the fix, and it does **not** work as written — the reason is interesting and probably saves you the same hour.

### The proposal

Every arm of `resto_orig_verdict` except the last asks one question — *is this point nearly feasible for the orig NLP?* — and asks it with an absolute threshold only:

```rust
if orig_trial_inf_pr <= 1e2 * tol && tol > 1e-1 * orig_tol { TightenAndContinue }
if is_square_problem && orig_trial_inf_pr <= orig_constr_viol_tol { ConvergedToAcceptable }
if orig_trial_inf_pr <= 1e2 * tol { ConvergedToFeasiblePoint }
LocallyInfeasible
```

An absolute test cannot answer that question scale-invariantly, because a user may write any constraint as `s·g(x) ≥ s·b` for `s > 0` without changing the feasible set, and the residual then carries `s`.

**This tree already settled exactly this**, for the outer convergence check, in gh #390 (residual of #387): `IpoptCq::curr_relative_primal_infeasibility_max`, `max_i |c_i| / |b_i|` and `max_i |d_i − s_i| / max(|d_i|, |d_l_i|, |d_u_i|)`. Its docstring is worth reading here, because the worked example in it *is* the failing fixture:

> An absolute violation measure cannot tell "satisfied" from "violated by 10% of everything the row is" once the row's numbers are small — `x >= 0.7` written as `1e-12·x >= 0.7e-12` has an absolute violation of `1e-13` at `x = 0.6` …

So the shape of the fix is: require the point to be small **both** absolutely and as a fraction of the violated row's own magnitude, before any arm calls it nearly feasible. Rows with no declared magnitude (homogeneous, or an NLP that does not track one) contribute `0` to the relative measure and fall back to the absolute test, which is already invariant for them. `1e-2` is the natural band — it is what `OptErrorConvCheck::relative_viol_threshold` uses for the same judgement on the outer side.

That is a two-line change to the predicate. The hard part is supplying the quantity.

### Where it fails, concretely

I implemented it: added `inf_pr_relative` to `OrigTrialMeasures`, computed it in `eval_orig_trial_measures` the same way `IpoptCq::relative_{c,d}_infeasibility_max` does, threaded it through `resto_orig_verdict` and the adapter, and traced it.

**It comes back `0.0`**, so the guard never engages and the behaviour is unchanged:

```
verdict=TightenAndContinue orig_inf_pr=1.005193e-9 orig_inf_pr_scaled=1.005193e-9 orig_inf_pr_relative=0.000000e0
```

Both blocks abstain. `declared_c_rhs()` returns `None` for this NLP, which per its contract means abstain. For the `d` block I tried both `declared_d_bounds()` and the live-`d_l`/`d_u` fallback its docstring prescribes, and still got `0`.

My reading — and this is where someone who knows the restoration block layout should take over — is that the quantity `eval_orig_trial_measures` is measuring is `d(x_orig) − s_inner`, where `s_inner` is the *restoration* problem's slack, not the original row's bound. The residual is therefore not "row value vs its declared bound", and the magnitude to divide by is not reachable from the vectors in hand. Getting a genuine relative measure here probably means keying off the original row's declared bound through the `BLOCK_X` mapping, rather than reusing the `d − s` residual.

I have reverted all of this; the branch is untouched at `36cdf9cd`.

### Options, in the order I'd consider them

1. **Supply the relative measure properly** — right fix, needs the block-layout knowledge above.
2. **Narrow layer 2 to where it is provably safe.** The verdict that motivated #438 is `LocallyInfeasible`, and that arm is the *fall-through* — it is reached when the point is decisively infeasible, where absolute and relative agree. The three arms that misfire are the ones claiming near-feasibility. Rendering only `LocallyInfeasible` from layer 2 and leaving the other three to the existing post-hoc gates would fix `qcqp1000-1nc`'s class without touching the scale-sensitive judgements at all. Smaller, and it is the half of layer 2 the issue actually asked for.
3. **Raise the harness baselines** — I would not. The baselines are `0` and the docstring above shows this exact model was the motivating case for making them `0`. Moving them re-accepts a defect the tree deliberately fixed.

Option 2 looks like the best value for the risk, but it is your design call — you own the layer-2 shape, and I have only been holding the experiment.

### Where the branch stands

Rebased onto `main` (post-#442), `MERGEABLE` at `36cdf9cd`, CHANGELOG conflict resolved keeping both entries. 972 Rust tests pass including this PR's three and #442's five. The only blocker is the scale-invariance regression, which predates the rebase (`9750d393` failed the same check).

PR #445, opened 2026-08-02
initialize() shares one incidence walk across its passes

### LINKED ISSUE #444: initialize() derives the same incidence three times per call
An enhancement with a measured case, not a defect: results are correct; the same symbolic analysis runs several times within one call.

Profiling `pyomo_pounce.initialize` on a 25,276-constraint collocation model (3-point Radau, distillation; `fill=None, project=False`, so the numbers are almost purely the plan/analyze/block machinery):

- **345.7 s** for the call, essentially all symbolic; the numeric solving is a rounding error next to it. (Default fill/project: 639.9 s, same multiplicities.)
- `IncidenceGraphInterface.__init__` runs **3 times**, 247.6 s cumulative: the plan, the re-plan (below), and the analyze pass each construct whole-model incidence from scratch.
- `block_repair_plan` runs **twice**, 263.6 s cumulative: `initialize()` computes the plan itself, then `block_initialize(repair="auto")` recomputes it after the decisions are already fixed. The second run is structurally a no-op — every candidate is already fixed, so nothing can prune or pin — but it pays a full graph construction and a full denominator-rule sweep to conclude that.
- `get_incident_variables` runs exactly **6× per constraint** (151,656 calls, 192.2 s), each call regenerating the constraint's standard repn and walking its expression again. (Cumtimes overlap: the derivations run inside the constructions.)
- `block_init.py:494` already carries a TODO naming the graph sharing.

Fix shape, two independent steps:

1. **The no-op re-plan**: `initialize()` already did the repair, so it can pass `repair="off"` downstream. Behavior-identical by construction; removes one graph build and one denominator sweep, about a third of the call above.
2. **Within-call sharing** for the remaining two graphs, with one subtlety a naive version gets wrong: the plan's graph is built before the decisions are fixed, the analyze graph after, and incidence excludes fixed variables, so the two views legitimately differ. The share therefore happens at the structural level — walk each constraint's expression once and filter by current fixed status per pass. Outputs identical by construction; no cross-call caching, no invalidation question.

Regression evidence: repair plans and initialized values identical; an instrumentation assertion that incidence construction happens once per call, so the redundancy cannot quietly return; before/after timing belongs in the PR body rather than CI.

A PR with the fix follows.

devin-griff:
Fixes #444.

One `initialize()` call built whole-model incidence three times and re-derived each constraint's incident variables six times over: the plan pass, a re-plan inside `block_initialize(repair="auto")` that is structurally a no-op (every candidate already fixed) yet paid a full construction and denominator sweep, and the analyze pass.

**The change.** One structural walk per call (`structural_incidence`, fixed variables included, so it is independent of fix-state), filtered per pass to the currently-unfixed variables through `IncidenceGraphInterface.subgraph`, which copies the stored graph and never re-inspects constraints. The plan filters pre-fixing and the analyze pass post-fixing, so each sees exactly the node sets, in the same order, that a fresh construction would have built — verified by an identity test (same plan, same square decomposition, same block order). `initialize` passes `repair="off"` downstream since its own plan already ran. `block_repair_plan`, `block_analyze`, and `block_initialize` accept the shared graph as an optional `igraph=` argument and behave exactly as before when it is omitted.

**Measured** on the 25,276-constraint collocation model from #444, symbolic path (`fill=None, project=False`): **345.7 s → 51.5 s**. The default path's remaining time is dominated by the projection NLP solve itself, which this change does not alter.

**Regressions.** The fresh-versus-shared identity test above, and a counter test pinning exactly one model-walking incidence construction per `initialize()` call, so the redundancy cannot quietly return (`subgraph` re-enters `__init__` with a prebuilt graph tuple and is deliberately not counted; the counter fails if the re-plan is restored — checked by mutation). Full suite: 155 passed, 2 skipped.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -225,9 +225,13 @@ jobs:
           sccache: "true"
       - name: Install + smoke
         run: |
-          # pyomo + numpy (sensitivity needs numpy); the locally-built
-          # pounce-solver wheel provides the `pounce` Python package + CLI.
-          python -m pip install pyomo numpy pandas
+          # pyomo + numpy (sensitivity needs numpy); networkx + scipy so
+          # the incidence surface (test_repair.py / test_block_init.py)
+          # actually runs instead of importorskip-ing away — without
+          # them the redundancy guard of gh #444 never executes here.
+          # The locally-built pounce-solver wheel provides the `pounce`
+          # Python package + CLI.
+          python -m pip install pyomo numpy pandas networkx scipy
           python -m pip install python/dist/*.whl
           # pyomo-pounce itself installed without deps: its declared
           # pounce-solver dep is satisfied by the wheel installed just above.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,16 @@ changes.
   filters it to the currently-unfixed variables through
   `IncidenceGraphInterface.subgraph`, which copies the stored graph
   without re-inspecting constraints. The plan filters pre-fixing and
-  the analyze pass post-fixing, so each sees exactly the view a fresh
-  construction would have built, in the same order; and `initialize`
-  passes `repair="off"` downstream since its own plan already ran.
+  the analyze pass post-fixing. The shared view is structural where a
+  fresh build substitutes fixed values: the edge sets can differ only
+  when a fixed zero cancels a term, which is guarded (affected rows
+  re-derived; a genuine cancellation falls back to a fresh build for
+  that pass), while the variable order within rows may differ on
+  models where fixed values change the linear/nonlinear split.
+  `initialize` passes `repair="off"` downstream since its own plan
+  already ran. On pyomo older than 6.7.1 (no
+  `IncidenceGraphInterface.subgraph`) every pass falls back to the
+  fresh build: old speed, same behavior.
   `block_repair_plan`, `block_analyze`, and `block_initialize` accept
   the shared graph as an optional `igraph=` argument and behave as
   before when it is omitted.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,9 @@ changes.
   when a fixed zero cancels a term, which is guarded (affected rows
   re-derived; a genuine cancellation falls back to a fresh build for
   that pass), while the variable order within rows may differ on
-  models where fixed values change the linear/nonlinear split.
+  models where fixed values change the linear/nonlinear split; order
+  feeds tie-breaks, so equally-valid diagnostics (which variable is
+  reported loose) may resolve differently between the two views.
   `initialize` passes `repair="off"` downstream since its own plan
   already ran. On pyomo older than 6.7.1 (no
   `IncidenceGraphInterface.subgraph`) every pass falls back to the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,13 @@ changes.
   `IncidenceGraphInterface.subgraph`, which copies the stored graph
   without re-inspecting constraints. The plan filters pre-fixing and
   the analyze pass post-fixing. The shared view is structural where a
-  fresh build substitutes fixed values: the edge sets can differ only
-  when a fixed zero cancels a term, which is guarded (affected rows
-  re-derived; a genuine cancellation falls back to a fresh build for
-  that pass), while the variable order within rows may differ on
+  fresh build substitutes fixed values: the edge sets can differ when
+  that substitution cancels a term — a fixed zero is the obvious way
+  but not the only one, since values can cancel across terms with no
+  fixed variable being zero (`a*x - b*x` with `a` and `b` fixed
+  equal). Every row adjacent to a fixed variable is therefore
+  re-derived and compared, and a genuine cancellation falls back to a
+  fresh build for that pass; the variable order within rows may differ on
   models where fixed values change the linear/nonlinear split; order
   feeds tie-breaks, so equally-valid diagnostics (which variable is
   reported loose) may resolve differently between the two views.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,33 @@ changes.
 
 ## [Unreleased]
 
+### Performance — pyomo-pounce: `initialize()` walked the same incidence three times per call (#444)
+
+- One `initialize()` call built whole-model incidence three times and
+  re-derived each constraint's incident variables six times over: the
+  plan pass, a re-plan inside `block_initialize(repair="auto")` that
+  was structurally a no-op (every candidate already fixed) yet paid a
+  full graph construction and denominator sweep, and the analyze
+  pass. On a 25,276-constraint collocation model that was 345.7 s of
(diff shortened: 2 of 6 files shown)


jkitchin (commented):
Reviewed at `c5a57df`, with pyomo 6.10.1 installed to check the claims empirically.

## Overview

The design is right, the perf win is real, and the "no-op re-plan" reasoning checks out: `initialize` validates `repair ∈ {auto, off}`, the outer plan fixes `decisions` + `pinned`, and the inner plan would run with an empty candidate pool over a graph where every pin is already fixed — so `pruned`/`pinned` are necessarily empty. Hardcoding `repair="off"` downstream is safe.

The central claim, though — "each sees exactly the node sets, in the same order, that a fresh construction would have built" — does not hold, and one case is a genuine behavior regression.

## Issues

### 1. Breaks on pyomo < 6.7.1, but the package declares `pyomo>=6.0` (blocking)

`IncidenceGraphInterface.subgraph` was added in pyomo 6.7.1. Verified against the wheels:

```
6.4.4 subgraph: False | 6.6.2: False | 6.7.0: False | 6.7.1: True
```

`_active_view` (`block_init.py:470`) calls it unconditionally whenever anything is fixed — which is the normal path, since `initialize()` fixes the decisions before the analyze pass. On pyomo 6.0–6.7.0 that is an `AttributeError` where the code previously worked. CI installs latest pyomo, so nothing catches it.

Fix: bump `pyomo>=6.7.1` in `pyomo-pounce/pyproject.toml`, or have `_active_view` fall back to a fresh construction when `subgraph` is absent.

### 2. The filtered view is not equivalent to a fresh build when a fixed variable multiplies an unfixed one

With `include_fixed=False`, pyomo's default `standard_repn` method substitutes fixed values; with `include_fixed=True` it does not. Two divergences, both reproduced:

**a. Extra edges from cancellation → false "square", failing block.** A decision fixed at 0 multiplying an unknown (`c1: y == a*x`, `a` fixed at 0.0) drops `x` from `c1` in a fresh build but keeps it in the shared view:

```
FRESH:  system square: False, overconstrained: c2, c1   (nothing solved, nothing broken)
SHARED: system square: True, blocks solved 2
        FAILED: 1x1 block at 'c1' failed (derivative ... very close to zero)
```

The shared path claims a square system, then fails a block and stops the loop, so downstream blocks keep their seeds. This is not a tie-break — it is a worse answer than the shipped behavior. Zero-valued decisions (closed splitter, zero purge, zero recycle) are not exotic in the flowsheet models this targets.

**b. Variable ordering differs.** With `a` fixed at 2.0 in `c1: a*x + y == 1`, `a*x` is linear-in-`x` under substitution but a nonlinear term without it, so the linear/nonlinear split reorders the variable list:

```
fresh: vars ['x','y','z']   c1 -> ['x','y']
view : vars ['y','x','z']   c1 -> ['y','x']
```

`block_repair_plan` documents `gvars` order as an implicit priority for tie-breaking, and DM / block-triangularization inherit it. Observable today: for the same model, `plan.loose_variables` reports `['y']` fresh vs `['x']` shared.

Neither is fatal on well-posed models, but the docstrings on all three functions, the CHANGELOG entry, and the PR body assert exact identity. At minimum, correct the claim to "structural incidence, independent of fixed values" and note that diagnostics on values-dependent models may differ. If exact equivalence matters, the cleaner route is to make both paths purely structural (`method=IncidenceMethod.identify_variables`) so fresh and shared agree — but that changes the fresh path too and deserves its own call.

### 3. Test gap that hides exactly this

`test_shared_graph_matches_fresh_analysis` uses `_split_model` with `feed` **and** `split` both fixed — the one arrangement where the bilinear term has no unfixed factor, so no divergence is possible. Fixing only `split` (or only `feed`) would fail the test today. Worth adding that case plus a zero-valued-decision case, whichever way #2 is resolved.

Separately: `ci.yml` installs `pyomo numpy pandas` but not `networkx` / `scipy`, so `test_repair.py` (and the rest of the incidence surface) is `importorskip`-ed away in CI entirely. The new counter test meant to stop the redundancy from returning never runs there. Adding `networkx scipy` to that install line would give the guard actual teeth.

### 4. Friendly `ImportError` regression

`block_repair_plan` / `block_analyze` / `block_initialize` each probe `networkx` explicitly to turn pyomo's deferred-import failure into an actionable message. `initialize()` now calls `structural_incidence(model)` (`repair.py:221`) *before* any of them, and that function has no probe — so a user without networkx/scipy gets a raw `DeferredImportError` from pyomo instead. Moving the probe into `structural_incidence` restores it.

## Minor

- `structural_incidence` is in `block_init.__all__` but not re-exported from `pyomo_pounce/__init__.py`, unlike its siblings. Since `igraph=` is documented public API, users need `pyomo_pounce.structural_incidence` to use it.
- `_active_view` returns the shared graph itself when nothing is fixed, so plan and analyze can alias it. No pass mutates today, so it is fine — but a one-line note would keep it that way.
- Currently conflicting with `main` (CHANGELOG, after #448).

## Verdict

The approach is sound — sharing the walk, the `repair="off"` reasoning, the counter test, and the documentation of the fix-state contract are all solid. Blocking on #1 (hard break for the declared pyomo range) and #2 (behavior change presented as an identity); #3 and #4 are good follow-ons.

---

devin-griff:
All addressed, in 2203969, point by point:

**1. pyomo < 6.7.1 (blocking).** `_active_view` now falls back to a fresh construction when `IncidenceGraphInterface.subgraph` is absent, so the declared `pyomo>=6.0` range keeps working at the old speed with the old behavior. If you'd rather bump the floor to 6.7.1 and keep the code unconditional, say so and I'll do that instead.

**2. The false identity claim (blocking).** You're right on both counts.
**2a (zero cancellation):** guarded — rows adjacent to a variable fixed at exactly 0 are re-derived with value substitution, and a genuine cancellation falls back to a fresh construction for that pass, so the false-square-then-failed-block outcome cannot occur. The guard is mutation-verified: disabling it fails the new zero-decision test.
**2b (ordering):** not prevented, documented honestly — every docstring and the CHANGELOG now state the structural contract: edge sets match away from cancellations, and since order feeds tie-breaks, equally-valid diagnostics (your `loose_variables` example: which of two variables is reported loose) may resolve differently between the views, not merely list differently. Your cleaner option — both paths purely structural so fresh and shared agree exactly — changes the fresh path's semantics too, so I've left it as your call rather than folding it in here.

**3. Tests and CI.** Both missing arrangements added: the one-factor-fixed case, compared set-level per the corrected contract, and the zero-decision case, which must reproduce fresh diagnostics exactly through the fallback. And `ci.yml` now installs networkx and scipy — your sharpest find, since without them the entire incidence surface including the #444 counter was importorskip-ed away and the guard had no teeth.

**4. ImportError probe.** Moved into `structural_incidence` with the same actionable message; it runs first in the pipeline, so the friendly error is restored for missing networkx/scipy.

**Minors.** `structural_incidence` is re-exported from the package root beside its siblings; the aliasing note is in `_active_view`'s docstring; conflicts absorbed, including over #436's merge.

Suites: 170 passed, 2 skipped on the rebased branch.

PR #454, opened 2026-08-02
information(): the un-inverted sibling of covariance() (roadmap item 2)

devin-griff:
Item 2 of the covariance/information roadmap (#262); items 3 and 4 follow separately.

**What it is.** `information(model, hessian=)` returns the reduced Hessian over the declared fitted block from the same single solve: natural units, no σ² anywhere, so for a homoscedastic Lagrangian fit `covariance()` equals `2σ²·inv(information())` on the free block (tested as an identity). Same `hessian=` selector as `covariance()`.

**The construction.** The Lagrangian form is built by tangent recovery rather than inverting the covariance back or subtracting the barrier off the factor: the K-inverse columns' x-blocks are `T·M` — each satisfies the equalities and has the fitted block as its own coordinates — so `T = Zx·inv(M)` exactly, and `R = TᵀHT` with the exact Lagrangian Hessian through a new session primitive, `Solver.hessian_vec(v)` (user-space, natural units, following the sensitivity-output contract). The barrier weight cancels multiplicatively. Measured on a pinned model with Σ/q ≈ 3×10¹⁰:

| route | max rel. error on R | on S |
|---|---|---|
| tangent recovery | 1.9e-16 | 5.2e-16 |
| factor subtraction | 3.2e-6 | 3.6e-5 |

The recovery requires the equalities to determine the non-fitted variables given the fitted block — the same square estimation structure `covariance()` already assumes — stated in the docstring.

**Dispositions** follow item 1's table, with the sign of one flipped by design: a strongly active parameter returns **S**, the reduction onto the pinned set, not a zero row — zero information is the opposite of what a pinned parameter carries — conditional on the rest of the pinned set, with zero cross blocks. Binding constraint rows project the free block on both sides, the pseudo-inverse of the projected covariance (tested against `pinv`). The Gauss-Newton product is formed over **all** fitted parameters and sliced last, so the pinned rows exist to build S from. An indefinite Lagrangian block returns as computed with a warning naming Gauss-Newton as the PSD alternative; the detector is unit-pinned, since at a genuine minimum the reduced Hessian is PSD by necessity and no honest fixture reaches the branch end to end.

**Shared membership.** The item-1 classification block is extracted into `_classify_fitted_block`, consumed by both accessors, so membership, row handling, and warnings cannot drift between them; `covariance()`'s behavior is unchanged (its entire suite passes untouched). One upgrade it gains: the binding-row conditional-information scalar is now the exact tangent value `aᵀRa` — computed lazily, so only solves with a binding row pay the Hessian products — replacing the documented digits-losing factor subtraction, and pinned against the analytic value in a new test.

**Tests** anchor to analytics: the linear model's information is exactly `2XᵀX` and `eigen()` matches its eigenvalues; the inverse identity against `covariance()`; Gauss-Newton parity; the pinned S exact at 1e-9 with the barrier weight in play, in both forms (the Gauss-Newton assertion is the test that slice-last preserves the pinned rows); an objective-scaling fixture (df ≠ 1 asserted) pinning both forms in natural units, since every other fixture runs at df = 1; the `pinv` relation under a binding row; `hessian_vec` pinned at both the pyomo and pounce levels; error paths including gauss-newton-without-residuals; the indefinite detector directly. Mutation checks: breaking the tangent recovery fails five of the nine information tests, a scaled-space Hessian product fails the scaling test, a slice-first Gauss-Newton fails the pinned test.

Docs: a new "The information matrix" section in `sensitivity.md`. Suites: pyomo-pounce 178, Python 28, Rust 57, all passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,12 @@ changes.
 - Membership, row handling, and their warnings are shared with
   `covariance()` (`_classify_fitted_block`), so the two accessors
   cannot drift.
+- Factor indexing follows the full-x vs var-x discipline of gh #450
+  throughout: fitted and residual rows route through `primal_row`,
+  and the tangent recovery slices the factor's var-x block and
+  scatters back to full-x for `hessian_vec`. Regression: one inert
+  fixed variable ahead of the fitted block changes nothing, both
+  forms.
 ### Fixed — pyomo-pounce: a fixed variable shifted every sensitivity result
 
 - A variable whose bounds are equal is removed from the solve
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,10 +36,12 @@ changes.
   build `S` from. An indefinite Lagrangian block returns as computed
   with a warning naming Gauss-Newton; the detector is unit-pinned
   since a genuine minimum is PSD by necessity.
-- `covariance()`'s binding-row conditional-information scalar is now
-  EXACT via the same tangent construction (lazily, only solves with a
-  binding row pay the Hessian products), replacing the documented
-  digits-losing factor subtraction.
+- `covariance()`'s binding-row conditional-information scalar now
+  comes from the same tangent construction (lazily, only solves with
+  a binding row pay the Hessian products): accurate to ~1e-6 where
+  the factor subtraction lost ten digits, the residue being the
+  binding row's own finite slack-barrier weight in the recovery.
+  Machine-exact when activity couples through equalities only.
 - Membership, row handling, and their warnings are shared with
   `covariance()` (`_classify_fitted_block`), so the two accessors
   cannot drift.
@@ -49,6 +51,7 @@ changes.
   scatters back to full-x for `hessian_vec`. Regression: one inert
   fixed variable ahead of the fitted block changes nothing, both
   forms.
+
 ### Added — POUNCE runs in the browser (WebAssembly)
 
 - The default build has no C or Fortran dependency, so the whole solver —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,41 @@ changes.
 
 ## [Unreleased]
 
+### Added — pyomo-pounce: `information()`, the un-inverted sibling of `covariance()` (covariance roadmap item 2, #262)
+
+- The reduced Hessian over the declared fitted block from the same
+  single solve, natural units, no `sigma^2`: for a homoscedastic
+  Lagrangian fit, `covariance()` equals `2*sigma^2*inv(information())`
+  on the free block (tested). Same `hessian=` selector as
+  `covariance()`.
+- The Lagrangian form is built by TANGENT RECOVERY, not by inverting
+  the covariance back or subtracting the barrier off the factor: the
+  K-inverse columns' x-blocks are `T*M`, so `T = Zx*inv(M)` exactly
+  and `R = T'HT` with the exact Lagrangian Hessian. The barrier
+  weight cancels multiplicatively; measured on a pinned model with
+  `Sigma/q ~ 3e10`, the route is exact to machine precision where the
+  subtraction route loses ten digits. A new session primitive,
+  `Solver.hessian_vec(v)` (user-space, natural units), supplies the
+  products.
+- Dispositions per item 1's table, with the sign of one flipped by
+  design: a strongly active parameter returns `S`, the reduction onto
+  the pinned set, NOT a zero row (zero information is the opposite of
+  what a pinned parameter carries), conditional on the rest of the
+  pinned set, zero cross blocks. Binding rows project the free block
+  on both sides, the pseudo-inverse of the projected covariance
+  (tested against `pinv`). The Gauss-Newton product is formed over
+  ALL fitted parameters and sliced last, so the pinned rows exist to
+  build `S` from. An indefinite Lagrangian block returns as computed
+  with a warning naming Gauss-Newton; the detector is unit-pinned
+  since a genuine minimum is PSD by necessity.
+- `covariance()`'s binding-row conditional-information scalar is now
+  EXACT via the same tangent construction (lazily, only solves with a
+  binding row pay the Hessian products), replacing the documented
+  digits-losing factor subtraction.
+- Membership, row handling, and their warnings are shared with
+  `covariance()` (`_classify_fitted_block`), so the two accessors
+  cannot drift.
+
 ### Changed — pyomo-pounce: `covariance()` membership from the solve's barrier geometry (#362, covariance roadmap item 1)
 
 - Bound and constraint activity on the fitted parameters now classifies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,9 @@ changes.
   a binding row pay the Hessian products): accurate to ~1e-6 where
   the factor subtraction lost ten digits, the residue being the
   binding row's own finite slack-barrier weight in the recovery.
-  Machine-exact when activity couples through equalities only.
+  Bound and equality activity stays machine-exact; the residue is
+  specific to binding inequality rows, which couple through the
+  slack block.
 - Membership, row handling, and their warnings are shared with
   `covariance()` (`_classify_fitted_block`), so the two accessors
   cannot drift.
--- a/crates/pounce-py/src/solver.rs
+++ b/crates/pounce-py/src/solver.rs
@@ -491,6 +491,25 @@ impl PySolver {
         Ok(out)
     }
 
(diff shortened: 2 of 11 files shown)


jkitchin (commented):
Reviewed against the branch at `5edecc0` (`cargo check -p pounce-sensitivity` clean; CI all six checks green).

## Overview

Adds `information(model, hessian=)` to `pyomo-pounce` — the reduced Hessian over the declared fitted block, un-inverted, natural units, no σ². Three supporting changes: a new session primitive `Solver.hessian_vec(v)` (Rust + pyo3), extraction of the shared membership/row classification into `_classify_fitted_block()`, and an upgrade of `covariance()`'s binding-row conditional-information scalar to the new tangent construction.

**The math checks out.** `K⁻¹`'s xx-block is `Z(Zᵀ(W+Σ)Z)⁻¹Zᵀ`, so every column of `Zx` lies in `null(A)`; defining `T = Zx·inv(M)` gives `E_fitᵀT = I` and `AT = 0`, which pins `T` uniquely under the stated square-estimation precondition — and Σ drops out entirely, which is exactly why this beats subtracting the barrier off the factor. The natural-units division is also right: `orig_ipopt_nlp.rs:1743-1748` passes the user `eval_h` multipliers as `c_scale·y_c` and `obj_factor·obj_scale`, and doesn't scale `x`, so `H_int = obj_scale·H_user` and the single `/ obj_scale` in `activity.rs:596` recovers natural units for both the objective and the constraint terms.

Tests anchor to analytic ground truth (`2XᵀX`) rather than to the implementation, the gh #450 full-x/var-x discipline is followed throughout, and the shared-classification extraction is faithful — `covariance()`'s logic is genuinely unchanged.

## Issues

**1. Doc comments were spliced onto the wrong functions (both Rust files).** The new function was inserted between `row_normal`'s doc comment and `row_normal` itself:

- `crates/pounce-sensitivity/src/activity.rs:568-582` — `hessian_vec` now carries a merged block that opens *"The gradient of one user constraint row at the converged iterate…"* and then continues *"The exact Lagrangian Hessian times a user-space vector…"*. `row_normal` at line 627 is left with no doc comment at all.
- `crates/pounce-sensitivity/src/solver.rs:291-303` — same splice, on **public** API. `Solver::hessian_vec` inherits `row_normal`'s paragraph including *"Errors on an out-of-range row"* (it errors on a length mismatch), and `Solver::row_normal` at line 315 keeps only the dangling fragment *"Serves the covariance roadmap's item 1: a binding row's normal…"*.

Move the `hessian_vec` definitions below `row_normal`, or move the doc blocks back.

**2. `covariance:`-prefixed warnings now reach `information()` callers.** `_classify_fitted_block` emits ~10 warnings all prefixed `covariance: …`, and `_minv` (`sens.py:1405`) raises `"covariance: the parameter block…"`. A user who only calls `information()` sees diagnostics labeled with an accessor they didn't call. `_ParamKeyed._loc` (`sens.py:985`) is the same: `Information[bad_key]` says *"not one of the covariance parameters"*. A `who="information"` argument threaded through the three helpers fixes all of it.

**3. The "EXACT" claim under a binding inequality row may be overstated.** The Σ-cancellation argument holds when the rows coupling into the factor's x-block are equalities. A strongly active *inequality* enters through the slack block with a large but finite Σ_s, so the recovered `T` is a Σ-weighted compromise rather than the exact constrained tangent. That's harmless for `information()`'s free block (the `Zb` projection annihilates that direction anyway), but `s_a = aᵀR_exact a` at `sens.py:1276` is evaluated *along* the contaminated direction. Suggestively, the two binding-row tests are written at `rtol=1e-6` while every other fixture asserts `1e-9`. Either that tolerance is just conservative — worth confirming — or the CHANGELOG's *"now EXACT via the same tangent construction"* should be softened to "accurate to ~1e-6 where the subtraction route lost ten digits," which is still the real win.

**4. The tangent-recovery precondition is documented but unchecked** (`sens.py:1073-1075`). If the equalities don't determine the non-fitted variables given the fitted block, `T = Zx·inv(M)` silently isn't the tangent map and `R` is quietly wrong. `_minv` catches an outright singular `M`, but not the wrong-dimension case. A cheap guard (`n_var - n_eq == n_params`) would turn a silent wrong answer into an error.

## Minor

- `CHANGELOG.md:44` — no blank line before `### Added — POUNCE runs in the browser`. Matters here because the GitHub Release body is cut from this file by section.
- `sens.py:1540-1541` — `act = cb.act` and `active = cb.active` are dead after the refactor; nothing in `covariance()` reads either.
- `sens.py:1877` — mangled continuation: `… @` followed by 21 spaces and `np.linalg.solve(` on the same 132-char line. Valid Python, but clearly an editing artifact.
- `hessian_vec` isn't listed in `docs/src/sessions.md`'s session-primitive rundown, or in the natural-units paragraph at `sensitivity.md:472` that names `row_normal(j)` explicitly. (`row_normal` isn't in `sessions.md` either, so this matches item 1's precedent rather than being a regression.)
- `activity.rs:594` — `hv.values_mut().fill(0.0)` before a `mult_vector` call with `beta = 0.0` is redundant.

## Test coverage

Strong where it counts — the inverse identity, GN parity, the pinned-`S` exactness with the barrier weight live, the objective-scaling fixture (with `df != 1` asserted, closing the gap every other fixture leaves open), the `pinv` relation, and the fixed-variable regression. Gaps, all minor: no test for `information()`'s free-block-singular `RuntimeError` (`sens.py:1880`), the all-pinned path where `free` is empty (`sens.py:1886`), or the `kkt_perturbations` warning. The indefinite branch being unit-only is acknowledged in the PR description and is the right call.

## Verdict

Good to merge after fixing the doc-comment splice (#1) — that one ships wrong public rustdoc. #2 and #3 are worth doing in the same pass; the rest are cleanup.


---

devin-griff:
Thanks for the review. All points addressed across 98fdc19 and c559702; point by point:

**1. Doc splice.** Fixed by moving the `hessian_vec` definitions below `row_normal` in both files. `row_normal` has its full original comment back (including the item-1 paragraph in solver.rs), and `hessian_vec` carries only its own.

**2. Warning prefixes.** A `who=` argument now threads through `_classify_fitted_block` and `_minv`, and the key-lookup error reads a class-level `_who` that `Information` overrides, so `information()` callers see `information:`-prefixed diagnostics everywhere, including the singular-M refusal and bad-key errors.

**3. The "EXACT" claim.** You were right, and measurement says the 1e-6 tolerance is load-bearing twice over. The warning prints 6 significant digits, so the parsing test could never see below ~4e-7. Bypassing the print: the tangent R under the binding-inequality fixture carries 8.7e-7 relative error at default tolerance and degrades to 1.1e-4 at mu = 9e-11, the pinned combination driving M toward singularity as mu tightens, exactly the slack-barrier compromise you described. One refinement to the boundary you drew: a strongly active variable *bound* stays machine-exact (the item-2 spike's 1.9e-16 was a pinned variable, and the pinned-S test holds at 1e-9), so the residue is specific to binding inequality *rows*, which couple through the slack block; everything in W cancels multiplicatively. That statement now replaces the exactness claim in the CHANGELOG, the tangent and `information()` docstrings, the docs section, the test module header, the scalar test's name and docstring, and both 1e-6 tolerances carry comments saying why they are what they are.

**4. Precondition guard.** Added as `_estimation_counts()` (factor variables vs equality rows). `information()`'s Lagrangian form refuses with the counts in the message; `covariance()`'s binding-row scalar falls back to the factor subtraction outside the square structure, so it is never worse than before item 2. Both paths tested.

**Minor:** all five done. The CHANGELOG blank line; the dead `act`/`active` assignments (AST-checked that `covariance()` references neither); the mangled `S` continuation reflowed; `hessian_vec` added to the `sessions.md` rundown and the `sensitivity.md` natural-units paragraph, and `row_normal` added to the rundown beside it so the item-1 precedent stops propagating; the redundant pre-fill deleted after confirming `mult_vector` special-cases `beta == 0` as overwrite (`dense_gen_matrix.rs:306`).

**Test gaps:** all three closed, and the singular one turned out nicer than expected. The all-pinned path asserts S is the whole block, `2XᵀX` in both forms. For the singular free block, two interchangeable intercepts beside a pinned slope produce bit-identical tangent columns, so LAPACK hits an exact zero pivot and the refusal genuinely raises instead of returning a garbage Schur complement; the same flat direction forces real inertia corrections into the held factor (`delta_w = 4.1e-7`), which finally exercises the `kkt_perturbations` warning honestly rather than by mock. The non-square-structure tests cover the refusal (with counts), Gauss-Newton's indifference to it, and the scalar fallback.

Suites: pyomo-pounce 191, Python 30, Rust all green, `cargo fmt` clean.

PR #466, opened 2026-08-03
wrt= block selection on both accessors (roadmap item 3)

devin-griff:
Item 3 of the covariance/information roadmap (#262); item 4 (`retain_kkt()`) remains. Pure pyomo-pounce: no core changes.

**What it is.** `covariance(m, wrt=...)` and `information(m, wrt=...)` reduce onto any block of the solve's variables off the held factor, post-solve: a Var (scalar or indexed), an indexed slice (`m.x[2, :]`), a `(Var, iterable)` pair, data objects, or a list mixing these. The declared fitted block is the default, so the no-wrt behavior is untouched — the existing suites pass without modification. Each call re-reduces onto its own argument: one solve, as many blocks as you ask about, each getting that block's **marginal** (everything outside it profiled out). Sigma estimation always divides by the fit's own degrees of freedom — a property of the solve, not the question — so a sub-block's numbers equal the corresponding entries of the default answer exactly (mutation-verified: a block-sized df fails the test).

**Rank-deficient blocks are the prediction-band case.** A block with more coordinates than the fit has degrees of freedom (e.g. `wrt=m.r`, the predicted trajectory) has a perfectly well-defined marginal covariance — tested against the hat matrix, `σ²X(XᵀX)⁻¹Xᵀ` — but no inverse, so `covariance()` returns the homoscedastic Lagrangian marginal with membership handling bypassed, and `information()` refuses toward `covariance()`. The gate is the count, not LAPACK: measured, LAPACK happily inverts a structurally singular 25×25 slice into garbage without raising. Gauss-Newton and per-group noise are refused on such blocks too — their profiled Jacobians need `inv(M)`.

**`information()` never inverts a covariance where it matters.** A block that parameterizes the constraint manifold gets item 2's exact tangent construction. A sub-block of the fitted set gets its marginal by **Schur complement of the exact tangent R over the fitted block** — free fitted variables outside the block are profiled out, pinned ones conditioned on — so a pinned member costs no digits (route-discriminating test: the corrected `inv(M_B)` fallback fails the 1e-9 pin). Other free blocks reduce off the held factor with the item-1 corrections, which is benign there (no barrier term in the slice); fitted-level binding rows fall back the same way, since their projection does not compose simply with marginalization.

**The active set comes back with the matrix.** Strongly active variables outside the block are returned on the result as `.conditioned_on` (both accessors): the block's numbers are conditional on those bounds, not marginal over them. Identification is item 1's reduced-level rule applied per candidate as a singleton block — one backsolve gives `(K⁻¹)ᵢᵢ`, the effective curvature is `|1/(K⁻¹)ᵢᵢ − Σ|` clamped to the cancellation floor exactly as the block-level rule does, and the shipped ratio edges make the call. Scale-invariant where a Σ-threshold heuristic is not (mutation-verified: a 1e-3-scaled pinned variable evades the threshold and is caught by the rule); only near-bound variables pay the backsolve, behind a cheap `Σ > √μ` prefilter.

**Sharing, not forking.** The membership helper now reads the caller's rows instead of re-looking-up `fit_rows`, which is the whole change that lets it classify any block; its diagnostics are now block-relative ("block member", "variables outside the block") since a residual coordinate in a wrt block is not a fitted parameter.

**Tests** (13, all anchored analytically): exact equality with the default answer; the sub-block marginal and its sibling identity; the hat-matrix band; the conditional value with `.conditioned_on` across the plain, scaled, rank-deficient-bypass, and all-pinned cases; accepted forms; the df decoupling; the Schur route with a pinned member at 1e-9 in both Lagrangian and Gauss-Newton forms; an objective-scaling fixture (df ≠ 1 asserted) covering the marginal, band, information, and the singleton call; a fixed-variable fixture covering the three new factor-indexing paths (gh #450 class); the grouped-band refusal; the binding-row fallback's warning plus sibling identity; and the error paths (duplicates, names, empty, foreign, fixed).

Docs: a new "Choosing the block: wrt=" section in `sensitivity.md`. Suites: pyomo-pounce 204, all passing.

### DIFF
--- a/pyomo-pounce/pyomo_pounce/sens.py
+++ b/pyomo-pounce/pyomo_pounce/sens.py
@@ -1027,10 +1027,13 @@ class Covariance(_ParamMatrix):
     ordered like `params`; `sigma_sq` is the residual variance that was
     used. eigen() supports identifiability diagnosis."""
 
-    def __init__(self, params, matrix, sigma_sq):
+    def __init__(self, params, matrix, sigma_sq, conditioned_on=()):
         super().__init__(params, matrix)
         self.params = self._params
         self.sigma_sq = sigma_sq          # float, or {group: float}
+        # strongly active variables OUTSIDE the block: the matrix is
+        # conditional on those bounds (roadmap item 3); empty when none
+        self.conditioned_on = tuple(conditioned_on)
         with np.errstate(invalid="ignore", divide="ignore"):
             se = np.sqrt(np.diag(self.matrix))
             corr = self.matrix / np.outer(se, se)
@@ -1139,14 +1142,13 @@ def _classify_fitted_block(session, params, rows, M, zcols,
     # (pounce#128), and so are the report's sigmas and row_normal
     # (unscaled at the classifier boundary per the same contract), so
     # everything here composes without scale factors.
-    sig_fit = np.array([float(act["var_sigma"][session.fit_rows[p]])
-                        for p in params])
+    sig_fit = np.array([float(act["var_sigma"][r]) for r in rows])
     q_red = np.abs(np.diag(R_W) - sig_fit)
     floor = np.sqrt(np.finfo(float).eps) * max(
         1.0, float(np.abs(np.diag(R_W)).max()))
     active = []
     for i, p in enumerate(params):
-        st = act["var_status"][session.fit_rows[p]]
+        st = act["var_status"][rows[i]]
         if st in ("unbounded", "fixed"):
             continue                       # no variable bound to classify
         ri = float(sig_fit[i]) / max(float(q_red[i]), floor)
@@ -1365,9 +1367,12 @@ class Information(_ParamMatrix):
 
     _who = "information"
 
-    def __init__(self, params, matrix):
+    def __init__(self, params, matrix, conditioned_on=()):
         super().__init__(params, matrix)
         self.params = self._params
+        # strongly active variables OUTSIDE the block: the matrix is
+        # conditional on those bounds (roadmap item 3); empty when none
+        self.conditioned_on = tuple(conditioned_on)
 
     def eigen(self):
         """(eigenvalues, eigenvectors) of the information matrix,
@@ -1430,6 +1435,113 @@ def _nullspace(A):
     return vh[rank:].T
 
 
+def _resolve_wrt(session, wrt, who):
+    """Normalize wrt= into the block: an ordered list of variable data
+    objects with their full-x (.col) rows. Accepted forms: None (the
+    declared fitted block, exactly the prior behavior), a Var component
+    (scalar or indexed: every member), an indexed slice (m.x[2, :]), a
+    (Var, iterable) pair (var[t] for t in the iterable), a single
+    VarData, or an iterable mixing any of these. Duplicates are an
+    error: a repeated coordinate makes the block singular by
+    construction."""
+    if wrt is None:
+        params = list(session.fit_rows.keys())
+        if not params:
+            raise RuntimeError(
+                f"{who}: no fitted parameters were declared; flag the "
+                "fitted variables with declare_fitted() before the "
+                "solve, or select a block explicitly with wrt=")
+        return params, [session.fit_rows[p] for p in params]
+
+    try:
+        from pyomo.core.base.indexed_component_slice import (
+            IndexedComponent_slice,
+        )
+    except ImportError:                    # pragma: no cover
+        IndexedComponent_slice = ()
+
+    def leaves(obj):
+        if (isinstance(obj, tuple) and len(obj) == 2
+                and hasattr(obj[0], "ctype")):
+            comp, idx = obj
+            for t in idx:
+                yield comp[t]
+            return
+        if isinstance(obj, IndexedComponent_slice):
+            # a slice PROXIES attribute access to its members, so the
+            # duck-typing below would misfire; iterate it directly
+            for v in obj:
+                yield v
+            return
+        if hasattr(obj, "ctype"):
+            values = getattr(obj, "values", None)
+            if callable(values):           # a component: every member
+                for v in values():
+                    yield v
+            else:                          # a data object
+                yield obj
+            return
+        if isinstance(obj, str):
+            raise TypeError(
+                f"{who}: wrt takes variables, not names; got {obj!r}")
+        for el in obj:                     # slice or plain iterable
+            yield from leaves(el)
+
+    params, rows, seen = [], [], set()
+    for v in leaves(wrt):
+        name = getattr(v, "name", None)
+        if name is None:
+            raise TypeError(
+                f"{who}: wrt element {v!r} is not a Pyomo variable")
+        try:
+            r = session.var_entry(name)
+        except ValueError as e:
+            raise ValueError(f"{who}: wrt member {e}") from None
+        if id(v) in seen:
+            raise ValueError(
+                f"{who}: wrt lists {name} twice; a repeated coordinate "
+                "makes the block singular by construction")
+        seen.add(id(v))
+        params.append(v)
+        rows.append(r)
+    if not params:
+        raise ValueError(f"{who}: wrt resolved to an empty block")
+    return params, rows
+
+
+def _conditioned_on(session, act, rows):
+    """Strongly active variables outside the block. Their Sigma stays
+    in the held factor and drives the coupling through them to zero as
+    mu falls, so the block's numbers are the values conditional on
+    those bounds, not the marginal over them. Returned with the matrix
+    rather than warned: it is a property of the answer, not a defect.
+
+    Identification: the report's raw status where curvature identifies
+    it, plus the barrier weight alone (Sigma > 1/sqrt(mu), the
+    classifier's strong edge at unit curvature) where the raw call is
+    `unidentified`. In the residual idiom a pinned variable has zero
+    raw Lagrangian curvature, which is exactly why item 1 classifies
+    block members at the reduced level; outside the block there is no
+    reduced level, and the barrier weight is the thing doing the
+    conditioning. The Sigma test is not scale-invariant (no curvature
+    to normalize against); a wildly rescaled variable can evade or
+    overtrip it."""
+    inside = set(int(r) for r in rows)
+    mu = float(act["mu"])
+    edge = 1.0 / np.sqrt(mu) if mu > 0 else np.inf
+    out = []
+    for idx, st in enumerate(act["var_status"]):
+        if idx in inside or st in ("unbounded", "fixed", "equality"):
+            continue
+        pinned = st == "strongly_active" or (
+            st == "unidentified"
+            and float(act["var_sigma"][idx]) > edge)
+        if pinned:
+            v = session.orig_var(session.var_names[idx])
+            out.append(v if v is not None else session.var_names[idx])
+    return tuple(out)
+
+
 def _minv(M, who="covariance"):
     try:
         return np.linalg.inv(M)
@@ -1440,7 +1552,8 @@ def _minv(M, who="covariance"):
             "dependent (structurally unidentifiable)") from e
 
 
-def covariance(model, sigma_sq=None, n_data=None, hessian="lagrangian"):
+def covariance(model, sigma_sq=None, n_data=None, hessian="lagrangian",
+               wrt=None):
     """Asymptotic covariance of the fitted parameters of a
     least-squares problem, from ONE ordinary solve.
 
@@ -1483,6 +1596,23 @@ def covariance(model, sigma_sq=None, n_data=None, hessian="lagrangian"):
     semidefinite, which makes it the safe choice when the covariance
     must stay PSD, e.g. feeding an arrival-cost update in moving
     horizon estimation.
+
+    wrt= selects the block (covariance roadmap item 3): any of the
+    solve's variables, not only the declared fitted ones, given as a
+    Var (scalar or indexed), an indexed slice (m.x[2, :]), a
+    (Var, iterable) pair, data objects, or a list mixing these; None
+    (default) is the declared fitted block, exactly the prior
+    behavior. Each call re-reduces onto its own argument, so one solve
+    serves as many blocks as are asked about, each getting that
+    block's MARGINAL (everything else profiled out). Sigma estimation
+    always divides by the fit's own degrees of freedom, a property of
+    the solve, not the block. A rank-deficient block (more
+    coordinates than the fit has degrees of freedom, e.g. a predicted
+    trajectory: the prediction-band case) gets the homoscedastic
+    Lagrangian marginal, with membership handling bypassed. Strongly
+    active variables OUTSIDE the block come back on the result as
+    .conditioned_on: the matrix is conditional on those bounds, not
+    marginal over them.
 
     Returns a Covariance object keyed by the declared variables'
     data objects: cov[m.A, m.k], cov.std_err[m.A],
@@ -1536,12 +1666,15 @@ def covariance(model, sigma_sq=None, n_data=None, hessian="lagrangian"):
             "no sensitivity session: declare_fitted() (and optionally "
             "declare_residual()) then solve with SolverFactory('pounce') "
             "first")
-    params = list(session.fit_rows.keys())
+    # the block: the declared fitted parameters by default, or any
+    # block of the solve's variables via wrt= (each call re-reduces
+    # onto its own argument, so one solve serves as many blocks as are
+    # asked about, each getting that block's marginal)
+    params, rows = _resolve_wrt(session, wrt, "covariance")
     n_params = len(params)
-    if n_params == 0:
-        raise RuntimeError(
-            "covariance: no fitted parameters were declared; flag the "
-            "fitted variables with declare_fitted() before the solve")
+    # sigma estimation divides by the FIT's degrees of freedom, a
+    # property of the solve, not of the block being asked about
+    n_fit = len(session.fit_rows)
 
     # ── guardrails ────────────────────────────────────────────────────────
     pert = np.asarray(session.solver.kkt_perturbations)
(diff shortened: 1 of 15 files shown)


pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1430,17 +1435,207 @@ def _nullspace(A):
...
+        from pyomo.core.base.indexed_component_slice import (
+            IndexedComponent_slice,
+        )
+    except ImportError:                    # pragma: no cover
+        IndexedComponent_slice = ()
+
+    def leaves(obj):
+        if (isinstance(obj, tuple) and len(obj) == 2
+                and hasattr(obj[0], "ctype")):
+            comp, idx = obj
```
jkitchin, line 1510: **`wrt=(m.a, m.b)` silently returns a one-element block.**

This matches *any* 2-tuple whose first element has `.ctype`, so a tuple of two Vars — the thing a user types right next to the documented `wrt=[m.a, m.b]` — takes the `(Var, iterable)` branch. `ScalarVar.__iter__` yields `[None]` and `m.a[None] is m.a`, so the second Var is consumed as an index set and vanishes.

Verified against Pyomo 6.10.1, running this branch's logic directly:

```
iter scalar var m.b: [None]
(m.a, m.b) ->  ['a']                        # m.b silently dropped
(m.x, m.y) ->  ['x[1]', 'x[2]', 'x[3]']     # m.y silently dropped
```

No error, no warning — just a 1×1 marginal where a 2×2 was asked for. That is the same class of failure the duplicate check on line 1497 exists to prevent: a wrong-shape block that looks plausible.

Disambiguating on the second element fixes both cases (they then fall through to the generic-iterable branch and resolve correctly):

```python
if (isinstance(obj, tuple) and len(obj) == 2
        and hasattr(obj[0], "ctype") and not hasattr(obj[1], "ctype")):
```

Worth an assertion in `test_wrt_accepted_forms` that `wrt=(m.a, m.b)` equals `wrt=[m.a, m.b]` — the current form test only exercises `(m.r, range(N))`, where the second element is genuinely an index iterable, so it can't catch this.

---

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1569,12 +1783,30 @@ def covariance(model, sigma_sq=None, n_data=None, hessian="lagrangian"):
...
+    if wrt is not None and n_params > n_var - n_eq:
+        cb = None
+    else:
+        try:
+            cb = _classify_fitted_block(session, params, rows, M, zcols)
+        except RuntimeError as e:
+            # backup for a within-count but still dependent block
+            if wrt is None or "inverse KKT matrix" not in str(e):
+                raise
+            cb = None
```
jkitchin, line 1881: **The "LAPACK doesn't reliably raise" argument applies below the count too.**

The count gate is justified in the PR body by exactly this: *"LAPACK happily inverts a structurally singular 25×25 slice into garbage without raising."* Agreed — and that is equally true for a block with `n_params <= n_var - n_eq` that is linearly dependent. On that side of the gate, `covariance()` depends on `_minv` raising, and `information()` (line 2137) has no fallback at all. So the silent-garbage path the gate was built to close stays open one step to its left.

A rank test on `M` — SVD rank, or `cond(M)` against a tolerance — would close it with the same reasoning that motivated the count gate, and would make this `except` a genuine last resort rather than the primary defence.

Separately, `"inverse KKT matrix" not in str(e)` is control flow on a message substring, and this PR edits that very message ~370 lines above (`_minv`, "the parameter block" → "the requested block"). It survives only because the edit happened to leave those three words intact. A dedicated exception type (`_SingularBlock(RuntimeError)`) raised by `_minv` would decouple the two.

---

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1675,6 +1907,38 @@ def group_jacobians():
...
+        act = session.solver.classify_activity()
+        flagged = [p.name for i, p in enumerate(params)
+                   if act["var_status"][rows[i]] not in
+                   ("inactive", "unbounded", "fixed", "equality")]
+        if flagged:
+            warnings.warn(
+                f"covariance: block members {flagged} have bound "
+                "activity, but the block is rank-deficient, so the "
+                "membership handling (pinned parameters, binding rows) "
+                "is unavailable; the marginal is returned without it.")
```
jkitchin, line 2013: **These two messages assert a count claim that one of the paths into here contradicts.**

`cb` becomes `None` two ways: the count gate at line 1794, and the `LinAlgError` rescue at line 1799. The second is reached only when `n_params <= n_var - n_eq` — a within-count but dependent block. Both messages here state the block has *"more coordinates than the fit has degrees of freedom"*, which is false on that path.

Given how carefully the rest of this PR states what it does and does not know (the tangent-accuracy notes, the `_tangent_reduced_hessian` docstring), it seems worth distinguishing the two — either a separate message on the rescue path, or wording that covers both ("rank-deficient" without the parenthetical count claim).

---

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1772,10 +2036,11 @@ def embed(cov_ff):
     sig_out = (next(iter(group_sigma.values()))
                if len(group_sigma) == 1 and None in group_sigma
                else group_sigma)
-    return Covariance(params, cov, sig_out)
+    return Covariance(params, cov, sig_out,
+                      _conditioned_on(session, cb.act, rows, "covariance"))
```
jkitchin: **`_conditioned_on` runs eagerly on the `wrt=None` path, which isn't free.**

Called unconditionally here and at line 2232, so a plain `covariance(m)` — untouched by this feature — now scans every full-x variable and issues a KKT backsolve per candidate. The `strongly_active` short-circuit and the `Σ > √μ` prefilter bound it, but the surviving statuses are `weakly_active` / `ambiguous` / `unidentified`, and by `_conditioned_on`'s own docstring *"a pinned variable in the residual idiom has zero RAW curvature, which is why the raw report calls it unidentified"*. On a bound-heavy estimation model (MHE, discretized dynamics with state bounds) that turns `n_fit` backsolves into `O(#near-bound variables)`.

The claim that "the no-wrt behavior is untouched" holds for the *values*, which is the important part — but not for the cost. A lazy cached property on `Covariance`/`Information` would keep the default path where it was and cost nothing at the call sites that read the attribute. If it stays eager, the CHANGELOG should say the default call now does extra backsolves.

---

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1162,25 +1164,25 @@ def _classify_fitted_block(session, params, rows, M, zcols,
         if status == "strongly_active":
             active.append(i)
             warnings.warn(
-                f"{who}: fitted parameter {p.name} is held by its "
+                f"{who}: block member {p.name} is held by its "
```
jkitchin: **Two things about the noun change** (this line and the four other warnings through line 1313, plus "involves non-fitted variables" → "involves variables outside the block").

1. It's user-visible and unlogged. Two existing tests had to be updated to match the new strings, which is direct evidence these messages *are* matched on — so anyone filtering on them downstream breaks silently. This wants a `### Changed` entry in the CHANGELOG alongside the `### Added` one.

2. On the default path the message is now strictly less informative than before: when `wrt is None` the block *is* the fitted set, and "fitted parameter A is held by its bound" said so exactly. Threading a noun alongside `who` — `"fitted parameter"` when `wrt is None`, `"block member"` otherwise — would keep the default message precise for roughly the cost of one more parameter on `_classify_fitted_block`.

The sharing itself is the right call; it's only the flattening of the default case that costs something.

---

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1430,17 +1435,207 @@ def _nullspace(A):
...
+    Rs = R_f[np.ix_(sel, sel)]
+    nb = len(keep)
+    if o_free:
+        A_ = Rs[:nb, :nb]
+        B_ = Rs[:nb, nb:]
+        D_ = Rs[nb:, nb:]
+        try:
+            R_B = A_ - B_ @ np.linalg.solve(D_, B_.T)
+        except np.linalg.LinAlgError:
+            return None
```
jkitchin, line 1628: Two small things in this helper:

**The `LinAlgError` → `None` at 1569-1570 is a silent route change.** The caller turns it into `cb.R_corr`, i.e. the corrected-reduction route with its documented (lower) precision, with nothing said. Every other route change in this PR is either warned or documented — the `bind_normals` decline gets the mixed-row warning, the rank-deficient bypass gets its own warning. A `warnings.warn` here would make the accuracy drop visible the same way.

**`who` misattributes on 1548.** `_classify_fitted_block` is called on the *fitted* block here, but `who="information"`, so a singular `M_f` surfaces as "information: the requested block of the inverse KKT matrix is singular; the block members are linearly dependent" — pointing the user at their `wrt=` argument when the singular matrix is the fitted set. The `primal_row` calls at 1532 already use `f"{who}(fitted)"` for exactly this reason; passing the same string here would make it consistent.

---

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1880,7 +2161,22 @@ def information(model, hessian="lagrangian"):
         R = cb.R_exact
         if R is None:
             n_var, n_eq = _estimation_counts(session)
```
jkitchin: Dead recompute — `n_var, n_eq` were already bound at line 2127 by the new gate, and nothing between reassigns them. Cheap (`_primal_row_map` is cached on the session), but it reads as a leftover from before the gate moved up.

---

jkitchin (changes requested):
## Review — `wrt=` block selection on both accessors

**Scope reviewed:** 6 files, +818/−56, `c98f036...0076179`. Entirely `pyomo-pounce`; no core/Rust changes, consistent with the PR description.

### What it does

Adds `wrt=` to `covariance()` and `information()`, reducing the answer onto an arbitrary block of the solve's variables instead of only the declared fitted set. `_resolve_wrt()` normalizes the argument; `_classify_fitted_block()` now indexes the activity report off the caller's `rows` rather than re-looking-up `session.fit_rows[p]` (the one change that lets it classify a non-fitted block); σ estimation switches to `n_fit` so sub-block marginals match the default answer entry-for-entry; a count gate routes rank-deficient blocks to a marginal-only path or a refusal; `_subblock_information()` gets a fitted sub-block's marginal by Schur complement of the exact tangent `R`; `_conditioned_on()` reports the outside active set on both result types.

The design is coherent and the math is well argued. The routing table (tangent / Schur / corrected-reduction / refuse) is the right decomposition, sharing the membership helper rather than forking it is the right call, and gating rank-deficiency on the count rather than on `LinAlgError` is correct for the stated reason.

### Blocking

**`wrt=(m.a, m.b)` silently returns a one-element block** (inline at `_resolve_wrt`). A 2-tuple of two Vars matches the `(Var, iterable)` branch, and the second Var is consumed as an index set. Verified against Pyomo 6.10.1: `(m.a, m.b)` resolves to `['a']`, `(m.x, m.y)` resolves to x's members only. No error, no warning — a wrong-shape block that looks plausible, which is exactly what the duplicate check exists to prevent. One added condition fixes it.

### Worth addressing before merge

- **A within-count but structurally singular block can still return garbage.** The PR's own justification for the count gate — LAPACK does not reliably raise on a structurally singular slice — applies equally below the count, where `covariance()` depends on `_minv` raising and `information()` has no fallback at all. A rank test on `M` would close it with the same reasoning that motivated the gate. Related: the `"inverse KKT matrix" not in str(e)` rescue is control flow on a message substring that this PR edits.
- **The rank-deficiency messages assert a count claim that the `LinAlgError` rescue path contradicts.**
- **`_conditioned_on` runs eagerly on the `wrt=None` path.** Values are unchanged, but a plain `covariance(m)` now pays `O(#near-bound variables)` backsolves where it paid `n_fit`. A lazy cached property keeps the default path where it was.
- **The warning-noun change is user-visible and unlogged** — two existing tests had to be updated to match, which is evidence these strings are matched on. Wants a `### Changed` CHANGELOG entry, and the default path loses precision it had ("fitted parameter" was exact when `wrt is None`).

Smaller points are inline: the silent Schur fallback, a `who` misattribution in `_subblock_information`, and a dead `_estimation_counts` recompute.

### Tests

Strong — 13 tests anchored to closed-form values rather than to the implementation, and the coverage choices are the right ones: the hat-matrix band, the df-decoupling identity, the objective-scaling fixture with `df != 1` asserted, the fixed-variable fixture covering all three new factor-indexing paths (the gh #450 class), scale-invariance for `conditioned_on`, and the binding-row fallback. The Schur test at `rel=1e-9` genuinely discriminates the route it claims to.

Gaps: nothing covers the tuple form above (which is what lets it through); nothing covers a within-count-but-dependent block; `test_wrt_error_paths` covers duplicates/names/empty/foreign/fixed but not a non-iterable scalar, which escapes as a bare `TypeError: 'int' object is not iterable` with none of the `{who}:` prefixing every other path in `_resolve_wrt` has.

### Security and performance

No security surface — no I/O, no deserialization, no input beyond Pyomo objects. Performance is the `_conditioned_on` eagerness noted above; `_subblock_information`'s extra `dof` backsolves plus `hessian_vec` calls are proportional to work the caller explicitly asked for and are fine. Docs (`sensitivity.md`) and CHANGELOG are otherwise thorough and match the code.

Requesting changes on the tuple bug — it's a silent wrong answer on a plausible call and the fix is one condition plus a test. The rest are judgment calls. The core design, the routing decisions, and the test anchoring are solid work.

---

devin-griff:
Thanks for the review. All points addressed across bf8ac7f, 163d1ef, and 3844011; each inline thread has its own reply. The short version:

The blocking tuple bug is fixed with exactly your condition, pinned by an exact-equality test against the list form (mutation-verified: reverting the guard fails it). The rank gate now exists on both sides of the count, `_SingularBlock` replaces the substring rescue, each refusal path states the reason that is true of it, `conditioned_on` is a lazy cached property, the diagnostics keep their exact "fitted parameter" wording on the default path and speak block-relative only under an explicit `wrt=`, and the three smalls (Schur decline warning, `who` attribution, dead recompute) are done.

One addendum you predicted: between two adjacent CI runs of this response — same test, same numpy 2.4.6, no numeric change in between — the wheel job's singular-`S` refusal test flipped from pass to fail: LAPACK returned garbage on one runner where it raised on another, because whether the pivot lands on exact zero is machine- and BLAS-path-dependent. That is the same non-determinism your rank-gate comment describes, one more step to the left, so the `S` computation now rank-tests the free block before the Schur solve (163d1ef), deterministic across builds, exception clause kept as last resort.

New tests: the tuple form; a within-count dependent block (duplicated design point, bit-identical coordinates below the count: covariance returns the rank-1 marginal against the hat matrix, information refuses with the dependence message); the unprefixed-scalar `TypeError`; and the block-relative wording asserted where it is new. Suites: pyomo-pounce 206, all passing.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Fixed with your condition verbatim: the second element must not be a component, so `(m.a, m.b)` falls through to the generic-iterable branch and resolves as the two-member block. The forms test asserts exact equality with `wrt=[m.a, m.b]`, and reverting the guard fails it.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Both taken. A `matrix_rank` test on M routes a within-count dependent block to the same bypass and refusal, and `_minv` now raises a dedicated `_SingularBlock`, so the rescue no longer reads message text. Regression: a duplicated design point makes two residual coordinates bit-identical below the count; covariance returns the rank-1 marginal against the hat matrix, information refuses. One honest boundary: the rank test catches fp-detectable dependence; softer near-dependence (an inertia-corrected collinear pair) still rides the `_SingularBlock` backstop. Your argument then proved itself a third time in CI: the item-2 singular-`S` refusal test flipped pass-fail-pass across adjacent runs with the same numpy and no numeric change, i.e. whether LAPACK raises there is runner-dependent, so the `S` solve now rank-gates the free block too (163d1ef).

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Split as suggested: the count gate keeps the "more coordinates than the fit has degrees of freedom" parenthetical, and the rank gate and the `_SingularBlock` backstop say "linearly dependent coordinates". Each path now states only what is true of it.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Took the lazy cached property: both result types store a thunk and compute on first read, so a plain `covariance(m)` pays exactly what it paid before this PR unless the attribute is accessed. One consequence stated in the property docstring (3844011): until first access the pending thunk keeps the session, and so the held KKT factor, alive — reading the attribute or discarding the result releases them.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Took the threading suggestion, which dissolves point 1 as well: the default path is restored to the exact "fitted parameter" / "non-fitted variables" strings (byte-identical, so nothing user-visible changed and the two test anchors reverted), and the block-relative nouns appear only under an explicit `wrt=` via a wording table on the helper. Since the default strings are unchanged there is no `### Changed` entry; the `### Added` entry documents the wrt-path wording as part of the new surface, and the wrt smoke test asserts it where it is genuinely new.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Both done. The `LinAlgError` decline now warns, naming the fallback and its precision consequence, matching how every other route change announces itself. The fitted-level classification inside the helper attributes to `information(fitted)`, consistent with the `primal_row` calls beside it.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Removed; the branch reuses the gate's bindings.

PR #468, opened 2026-08-03
retain_kkt(): factor retention without declarations (roadmap item 4)

devin-griff:
The final item of the covariance/information roadmap (#262). Pure pyomo-pounce, and small: the infrastructure was already in place.

**What it is.** The solve factors the KKT matrix to solve the NLP; the only question is whether that factor is kept for post-solve queries. Today any declaration keeps it. `retain_kkt(model)` keeps it with **nothing declared**, so `covariance(model, wrt=block)` and `information(model, wrt=block)` work on any block without a declared default — the MHE case, where the arrival state and the parameters are each queried by `wrt=` and neither is *the* fitted set. Implementation is a registry flag that `has_declarations` honors and the registry's deepcopy carries through `model.clone()`; `sens_solve` already tolerated empty declaration sets, so the retained path is the existing in-process solve with nothing else new.

**The roadmap's table, every row tested:**

| setup | factor kept | `covariance(model)` | `covariance(model, wrt=T)` |
|---|---|---|---|
| nothing | no | error | error |
| `declare_fitted(S)` | yes | over S | over T |
| `retain_kkt()` only | yes | error, no default | over T |
| `retain_kkt()` + `declare_fitted(S)` | yes | over S | over T |

An undeclared solve without the call takes the ordinary path — no registry, no session, no cost — and the accessors' no-session errors now name `retain_kkt()` as the declaration-free route. Retain alone serves any `wrt=` query, matching a declared solve of the same model entry for entry (marginal, information, and the prediction band asserted). `covariance(model)` with no block stays the no-default error. Retain beside declarations changes nothing. The clone carries the intent. And `retain_kkt()` keeps the factor, not a noise model, so the `sigma_sq=`/residuals requirements stand (tested).

**The retention policy, stated once** (docstring and a new docs section): the factor is kept if anything is declared or `retain_kkt()` was called, and a `Covariance`/`Information` result whose lazy `conditioned_on` has not been read keeps the session alive through its pending computation until first access — the #466 lifetime note folded into one policy instead of three accidents.

**Demo notebooks for the whole roadmap.** 31 (information and identifiability) and 32 (one solve, many questions) cover items 0 through 4, written for a reader new to NLP: the fitted models stated in math before the code, every term either standard, a code identifier, or plain English, and one message per figure — the joint-uncertainty line-versus-box picture, the eigenvector perturbation panels, the prediction band on undeclared variables, and the pinned-versus-free uncertainty comparison. Both are committed executed.

Tests: 6 new, table-driven. Suites: pyomo-pounce 211, all passing. Docs: "Keeping the factor: retain_kkt()" in `sensitivity.md`; CHANGELOG entry included.

With this, items 0 through 4 of #262 are complete.


jkitchin (commented):
## Review

The design is right and minimal, but I don't think this should merge as-is. One real statistical defect sits directly on the flow this PR advertises, and the headline notebook mislabels a confidence band as a prediction band.

**Verified good:** merges cleanly into `main` (2 commits behind, no conflicts); all 6 CI checks green; full `pyomo-pounce` suite on the branch = 209 passed / 2 skipped (the PR's "211"); all four rows of the truth table reproduce; the retain flag is honored in exactly one place (`has_declarations`) and carried by `_Registry.__deepcopy__` without carrying the session — a clone correctly never answers from its parent's factor. Adversarial probing found no session leak (40 re-solves → 1 live session, +2.6 MB RSS), clean errors on an infeasible solve, and bound-activity classification under retain-only matching a declared solve exactly.

---

### 1. sigma^2 uses the wrong degrees of freedom whenever nothing is declared fitted — `sens.py:1849`

`n_fit = len(session.fit_rows)`, then `ssr/(n_g - n_fit)`. Under retain-only there is no `declare_fitted`, so `n_fit == 0` and the estimate silently becomes SSR/n instead of SSR/(n-p). Reproduced on a 25-point linear fit:

```
declared  sigma^2 = 0.06406681  == SSR/(N-2)   ok
retain    sigma^2 = 0.05894147  == SSR/N       wrong
var(a) ratio declared/retain = 1.0870  == N/(N-2)
```

Variances are biased low by exactly n/(n-p) — 8% here, unbounded as p approaches n. The `n_data=` branch is the same, and the guard `if n_data <= n_fit` degenerates to `n_data <= 0`, so nonsense is accepted:

```
n_data=25: sigma^2 = 0.0589415   ACCEPTED
n_data= 2: sigma^2 = 0.736768    ACCEPTED
n_data= 1: sigma^2 = 1.47354     ACCEPTED   (= SSR)
```

The arithmetic is pre-existing (`declare_residual` alone with no retain does the same), but this PR promotes the empty-fitted-set case to a **documented first-class path**, and `sensitivity.md` explicitly blesses both broken routes:

> `retain_kkt()` keeps the factor, not a noise model, so `covariance()` still needs `sigma_sq=`, `n_data=`, or declared residuals.

Of those three, only `sigma_sq=` is correct under retain-only — and that is the only one `test_retain_kkt.py` exercises. `test_retain_only_sigma_still_required` checks the *missing*-sigma error, never the estimated-sigma branch.

**Suggested fix:** when sigma must be estimated and `n_fit == 0`, raise — something like "no fitted parameters were declared, so the degrees of freedom for the noise estimate are unknown; pass `sigma_sq=`, or `declare_fitted()`" — rather than silently dividing by n. Plus tests for retain + `declare_residual` and retain + `n_data`.

### 2. Notebook 32 section 2: "prediction band" is a confidence band on the mean response

`covariance(m, wrt=m.yhat)` is Var(yhat) — the sampling uncertainty of the *fitted curve*. A prediction band needs the observation noise added: sqrt(se^2 + sigma_hat^2). The notebook's own figure shows the error: sigma_hat = 0.058 while 2*se(yhat_0) = 0.073, and only **12 of 30 data points (40%)** fall inside the band that is plotted with the data drawn on top of it.

Affects the notebook title, the section 2 heading, the cell 5 prose, and the `fill_between` label. (The "prediction-band" wording at `docs/src/sensitivity.md:482` came in with #466 — worth fixing there too, but this PR is what makes it visibly wrong.) Either relabel it as a confidence band on the mean response, or widen it to an actual prediction band.

### 3. Notebook 31 committed outputs are not reproducible

`info.eigen()` is `np.linalg.eigh`; the smallest eigenvector's sign is an arbitrary LAPACK convention. Re-executing gives the exact negation in cell 7, and because the model is nonlinear the downstream perturbation is not sign-symmetric — cell 9 prints `SSR after the step, least identified: 0.01147` committed vs `0.01212` on re-run. `python/notebooks/README.md:5` promises "the saved outputs match what you will see". Everything else in nb31 reproduces exactly. Pinning the sign (e.g. force the largest component positive) before plotting/perturbing would fix it.

### 4. Notebooks 31 and 32 are not registered in `python/notebooks/README.md`

Every notebook 01-30 has a table row (25, 26, 28 included). These two belong right after the `26_parameter_covariance.ipynb` row under "Differentiating through the solver".

### 5. Notebook 31: two claims do not match their own numbers

- **Cell 0 intro** advertises "when a parameter sits at a bound ... the covariance reports zero variance", but the `b >= 2.11` case actually built in cells 14/15 is the *counter*-example: the classifier calls it ambiguous and `b` keeps finite variance 0.532899. The zero-variance demo is a different parameter (`A`) in a different model (cells 16/17). The intro attaches the claim to the wrong example.
- **Cell 19 bar chart** compares two different statistics. The "free" bar (9.73) is the diagonal of the full information block — conditional, with `k` held fixed. The "pinned (S)" bar (4.29) is the Schur complement with `k` profiled out. The like-for-like free quantity is `information(ms, wrt=[ms.A])[ms.A] = 5.3796`. The plotted drop 9.73 -> 4.29 overstates the loss; the honest comparison is 5.38 -> 4.29.

---

### Minor

- **`retain_kkt()` silently reroutes the solve.** Flipping `has_declarations` routes `solve()` into `sens_solve`, which discards every other Pyomo solve kwarg. Confirmed: adding one `retain_kkt(m)` line makes `solve(m, load_solutions=False)` load the solution anyway (`a` = 2.000 vs 0.500 on the ordinary path), leaves an IMPORT `dual` Suffix empty, and swallows misspelled kwargs that the ordinary path rejects loudly. This is identical with `declare_fitted`, so it is pre-existing — but the docstring and docs frame `retain_kkt()` as merely keeping a factor "the solve computes anyway", which invites dropping it into an existing working script. The truth table has a "factor kept" column but none for solve fidelity. Worth a sentence in the docs section.
- `import warnings` is unused in `pyomo-pounce/tests/test_retain_kkt.py:7`.
- No off-switch, and no way to release a retained session (pre-existing for declarations; `retain_kkt` adds a second door to it).
- nb31 cell 6 says the least-identified direction shifts "the rates", but the printed eigenvector has `k1 = +0.059` (negligible) and `k2 = +/-0.764`. Cell 0 states it correctly.
- nb32 cell 9 says the pinned band is "exactly zero" at t=0; measured 1.23e-05.
- Both notebooks carry `language_info.version = 3.12.3`; every other notebook in the repo says `3.12.11`.
- The branch is behind `main` by `b926f538` / `25dea714` (the #466 review follow-up). It merges clean, but a rebase would test against the final rank-gate scaling.

---

devin-griff:
Thanks for the review. All points addressed at e5887fe (rebased onto main, so the diagonal-scaled rank gates from your #467 are underneath); point by point:

**1. Wrong degrees of freedom under retain-only.** You're right, and it's the defect that mattered. Both estimation routes now refuse when nothing is declared fitted: `declare_residual`-based estimation and `n_data=` each raise with direction toward `sigma_sq=` or `declare_fitted()`, so the silent SSR/n bias and the degenerate `n_data <= 0` guard are gone. Tests cover both refusals (retain + residuals, retain + `n_data`), and the docs section now blesses only `sigma_sq=` under retain-only and says why.

**2. Confidence band mislabeled.** Correct, and your 40%-coverage number was the giveaway. Rather than just relabeling, section 2 now shows both objects in one figure: the covariance of the prediction variables labeled as the confidence band on the fitted curve, which is not supposed to contain the data, and the widened `sqrt(se^2 + sigma^2)` prediction band, which is; the committed figure shows the data inside the wide band and scattered around the narrow one, and the prose explains the distinction. Title, heading, labels, and the `sensitivity.md` wording from #466 all updated — and the same mislabel in the #262 design record is fixed on that branch as well.

**3. Non-reproducible outputs.** Each eigenvector's largest component is now pinned positive before printing and perturbing. Verified by executing twice and diffing all text outputs: identical except the solver's wall-clock line in the tee'd log, which varies in every executed notebook in the repo.

**4. README.** Both notebooks have rows after 26 under "Differentiating through the solver".

**5. Claims vs numbers.** The intro now attaches the zero-variance claim to the strongly active case and announces the declined-pin border case as what it is. The bar chart compares like for like: information about `A` alone in both bars, 5.38 free against S = 4.29 pinned, with the prose saying the drop is modest — which is the honest punchline anyway.

**Minors:** the solve-fidelity consequence is now stated in both the docs section and the `retain_kkt` docstring (routing through `sens_solve` changes the `solve()` keyword surface, your `load_solutions=False` finding); the unused import is gone; "shifting the rates" now says moving `k2`; "exactly zero" is now zero to solver tolerance; `language_info` is normalized to the repo's 3.12.11 (the executing kernel here is 3.12.3, for the record). On the missing off-switch and session release: agreed it's real and pre-existing for declarations too; I'd propose `release_kkt()` as a small follow-up rather than scope here, if you want it.

**On your suite count:** the two skips you saw are the build-stamp ratchet tests, which refuse to run against a CLI binary whose stamp doesn't match the checkout; with a locally staged build they run, which is the 211 (now 213 with the new refusal tests).

jkitchin (commented):
## Re-review: all substantive findings fixed

Re-verified against `224b6c67`. Every finding from the review above is fixed; one cosmetic contradiction remains.

| # | Finding | Status |
|---|---|---|
| 1 | sigma^2 divided by n instead of n-p under retain-only | **Fixed** |
| 2 | "prediction band" was a confidence band on the mean | **Fixed** |
| 3 | nb31 committed outputs not reproducible | **Fixed** |
| 4 | Notebooks absent from the README index | **Fixed** |
| 5b | Bar chart compared conditional vs profiled information | **Fixed** |
| — | All minors (unused import, undocumented solve-path reroute, "exactly zero", `language_info` version) | **Fixed** |

**Finding 1** was the only correctness bug in shipped code. Both estimation routes now raise; my original reproductions error cleanly:

```
retain + declare_residual → ValueError: ...no fitted parameters were declared, so the
  degrees of freedom for the estimate are unknown; pass sigma_sq= or declare_fitted()
n_data = 25 / 2 / 1      → same ValueError (all three, including the n_data=1 nonsense)
```

And it breaks nothing: the declared path is numerically identical to before (`var(a) = 0.009659304182827918`), and `declare_residual`-only still hits the earlier `_resolve_wrt` error, so no released behavior changes. Full suite on that exact commit: **211 passed, 2 skipped**.

**Finding 2** verified by coverage — prediction band `fit +/- 2 sqrt(se^2 + sigma_hat^2)` contains 30/30 data points, confidence band `fit +/- 2 se` contains 12/30. Exactly the intended contrast, and the prose now says outright that the confidence band is not supposed to contain the data. The `sensitivity.md` and CHANGELOG wording followed.

**Finding 3** was pinned deterministically, not papered over. Raw `np.linalg.eigh` on my machine returns `[-0.459, +0.450, -0.059, -0.764]` — the exact negation of the committed vector — and the added loop flips it back, which is why the fix commit left the cell 7 and cell 9 outputs unchanged. Full re-execution is byte-identical apart from ANSI color depth in the solver banner.

**Finding 5b** verified by recomputation: both bars are now literally `information(model, wrt=[A])[A]` — free 5.379605 (= the Schur complement `9.73058 - 9.61740^2/21.25831`), pinned 4.288907. The misleading conditional diagonal 9.73 is gone, and cell 18's markdown was updated to match.

### Still open (cosmetic, non-blocking)

nb31 cell 0 still reads:

> When a parameter is HELD at a bound, a limit like $b \ge 2$ that the classifier calls **strongly active**, the covariance reports zero variance...

That bound is `mp.b.setlb(b_hat + 1.5)` = 2.1099, which cell 15 shows the classifier calling **ambiguous**, keeping `b` at variance 0.532899. The strongly-active / zero-variance demo is `A >= A_hat + 0.4` in the second model (cells 16/17), so the intro cites the notebook's own counter-example. One-symbol fix: `$b \ge 2$` -> `$A \ge 2.4$`. Fine as a follow-up.

### Out of scope, filed separately

`Covariance.eigen()` and `Information.eigen()` still return raw `np.linalg.eigh`, so the eigenvector sign non-determinism that bit nb31 is unfixed for every other caller; nb31 works around it locally. Pre-existing, not introduced here — filed as a separate issue.

LGTM once CI is green.

---

jkitchin:
`1ac7f1cc` closes the last open item. The intro now cites `$A \ge 2.4$` — the strongly active bound from cells 16/17 — instead of the `b` bound that the classifier calls ambiguous.

Checked the number rather than taking it on faith: cell 17 sets `m2.A.setlb(A_hat + 0.4)`, and re-fitting that model gives `A_hat = 1.9963`, so the bound is **2.3963**. "$A \ge 2.4$" is accurate, and cell 17's own warning confirms the classifier calls it strongly active with zero variance — exactly what the sentence claims.

**Everything raised in this review is now fixed**, across findings 1-5 and every minor. CI is green on `1ac7f1cc` (all six checks).

One follow-up left deliberately out of scope, filed as #471: `Covariance.eigen()` and `Information.eigen()` still return raw `np.linalg.eigh`, so the eigenvector sign non-determinism that bit notebook 31 is unfixed for every other caller — the notebook pins it locally. Pre-existing, not introduced here.

Merging.

---

PR #470, opened 2026-08-03
Expose nl_tape to Python: in-memory construction, dense/sparse HVPs, and an erf TapeOp

### LINKED ISSUE #469: Expose nl_tape capabilities to Python: in-memory construction, HVP, and an erf TapeOp
## Context

[discopt](https://github.com/jkitchin/discopt) is moving its NLP subproblem derivatives off JAX and onto pounce's Rust AD tape (`pounce-nl/src/nl_tape.rs`). pounce is *already* the NLP subsolver — `discopt/solvers/nlp_pounce.py` builds cyipopt-shaped callbacks and hands them to pounce — but those callbacks are currently computed by a JAX evaluator. The change is a callback-provider substitution: one tape per model, replacing `jax.grad`/`jacfwd`/`hessian`.

Verified this works in principle: `pounce.read_nl()` on a 580-var/510-constraint instance reads in 8 ms and its `objective`/`gradient` agree with the JAX path **exactly** (relative difference 0.00e+00, gradient max-abs-diff 0.0).

Three gaps block the swap. All three are capabilities that already exist in the Rust core and are simply not reachable from Python.

## 1. No in-memory problem construction (biggest blocker)

`pounce-py` exposes only file-path `read_nl` (`crates/pounce-py/src/nl_problem.rs:421`). Confirmed by grep — `crates/pounce-py/src/` has **zero** references to:

- `parse_nl_text` (`crates/pounce-nl/src/nl_reader.rs:322`, `pub fn`)
- `Tape::build` / `Tape::build_multi` (`crates/pounce-nl/src/nl_tape.rs:180`, `:1912`, `pub fn`)

discopt models built through its Python modeling API never touch disk, so today the only route is writing a temporary `.nl` file. That is not just slow — it is **lossy**. discopt's `.nl` writer refuses several operators that pounce's tape *already supports*:

| operator | in pounce `TapeOp`? | survives discopt's `.nl` export? |
|---|---|---|
| `atan2` | yes (`Atan2`) | no — writer has no 2-arg funcall path |
| `min` / `max` | yes (`Min`, `Max`) | no — raises "requires DNLP model type" |

So round-tripping through `.nl` would *lose* capability that the tape has. Either binding solves it; `Tape::build(expr)` is the more direct fit if the expression IR is exposable, otherwise `parse_nl_text(&str)` at least avoids the filesystem.

**Ask:** a pyo3 binding for in-memory construction — `parse_nl_text(text) -> NlProblem`, and/or a tape-builder entry point.

## 2. No Hessian-vector product exposed

`hessian_directional` (`crates/pounce-nl/src/nl_tape.rs:696`) is `pub fn` with zero references in `pounce-py`. `pounce.NlProblem`'s Python surface is:

```
con_names, constraints, g_l, g_u, gradient, hessian, hessian_structure,
jacobian, jacobian_structure, m, minimize, n, nnz_hess, nnz_jac,
obj_constant, objective, var_names, variant, x0, x_l, x_u
```

— no HVP. discopt uses one (`_jax/nlp_evaluator.py:485-493`, a `jax.grad` + `jax.jvp` composition) for large models where the dense Lagrangian Hessian is impractical.

Everything else discopt needs is already exposed and matches well: the Ipopt-style `hessian(x, lam, obj_factor)` plus `hessian_structure`/`jacobian_structure` map directly onto discopt's `evaluate_hessian_values`/`hessian_structure`, and the sparse structures replace discopt's JAX sparsity probe and CPR-coloring helpers outright.

**Ask:** expose `hessian_directional` on `PyNlProblem`.

## 3. `erf` has no `TapeOp`

Diffing discopt's 30 DAG operators against pounce's `TapeOp` variants: **20 covered, 10 missing** — but 9 of the 10 are decomposable into ops pounce already has, and discopt can do that decomposition on its side (it already does exactly this in its `.nl` writer for `log1p`, `sigmoid`, `softplus`):

| missing | decomposition | needs pounce change? |
|---|---|---|
| `log1p`, `log2`, `sigmoid`, `softplus` | `log`/`exp` compositions | no |
| `entropy`, `centropy` | `x·log(x)`, `x·log(x/y)` | no |
| `prod` | `Mul` chain | no |
| `sign`, `signpower` | `Cmp` + `Select` (both present) | no |
| **`erf`** | **not elementary** | **yes** |

`erf` is the only genuine primitive gap. Its derivative is cheap and closed-form:

```
erf'(x) = 2/sqrt(pi) * exp(-x^2)
erf''(x) = -4x/sqrt(pi) * exp(-x^2)
```

**Ask:** an `Erf` `TapeOp` with value/tangent/adjoint arms. If that's unwanted, a loud unsupported-operator error is a fine answer — discopt would then refuse `erf` models on this path rather than silently mis-solve.

## Notes

- Items 1 and 2 are pure bindings for existing `pub fn`s — no behavior change to any existing API.
- Item 3 is the only new numerics.
- Not urgent: the discopt side is staged behind a default-off flag and gated on a differential correctness panel (`incorrect_count = 0`, no bound above its reference optimum) before anything graduates.

For reference, the full derivative surface discopt needs from the tape: objective, gradient, constraints, Jacobian (values + structure), Lagrangian Hessian (values + structure), and HVP. All but the last are already exposed on `PyNlProblem`.

### Suggested order

1. `hessian_directional` binding — smallest, unblocks large-model Hessian-free mode.
2. `parse_nl_text` binding — unblocks the Python-API model path without a `Tape` IR decision.
3. `Tape::build` binding — the better long-term seam; avoids `.nl` as an intermediate entirely.
4. `Erf` TapeOp — only needed for `erf` models; a loud refusal is an acceptable interim.

jkitchin:
Fixes #469

> **Revised after review** ([#470 review](https://github.com/jkitchin/pounce/pull/470#pullrequestreview-1)). Four claims in the original body were wrong and are corrected below; the Tests section now records what was measured rather than what I assumed. Fixes are in `39af24e`.

## Problem

A modeling frontend with its own expression DAG (discopt) wants pounce's Rust AD tape as its derivative provider instead of JAX. The evaluators already agree exactly — `read_nl()` on a 580-var/510-constraint instance matched the JAX path to 0.00e+00 relative difference — but three capabilities that exist in the Rust core had no Python door, so the only route was writing a temporary `.nl` file.

That round trip is not just slow, it is **lossy**. `.nl` writers routinely refuse operators this tape already differentiates:

| operator | in `TapeOp`? | survives a `.nl` export? |
|---|---|---|
| `atan2` | yes | no — no two-argument funcall path |
| `min` / `max` | yes | no — forces a DNLP model type |
| `erf` | **no** (until this PR) | no — AMPL has no `erf` opcode |

And there was no Hessian-vector product at all, which is what a large model needs when the dense Lagrangian Hessian is impractical.

## Cause

Not a bug — a reachability gap. `parse_nl_text`, `Tape::build`, and `Tape::hessian_directional` are all `pub fn` in `pounce-nl` with zero references from `crates/pounce-py/src/`. `erf` was the one genuine primitive gap: of the 10 operators discopt has that pounce lacked, 9 decompose into ops pounce already had, and `erf` does not.

## Fix

**In-memory construction.** `NlProblem::from_expressions` (with `NlProblemParts`) assembles a problem straight from `Expr` trees. The linear/nonlinear split is left empty on purpose — the tape treats a linear term like any other subexpression, so `.nl`'s `J`/`G` segments are a file-format optimization, not an evaluator requirement. `validate_expr` rejects two things the tape cannot recover from later: a `Var` index at or past `n` (an out-of-bounds read in the forward sweep) and an `Expr::Funcall` (there is nowhere to put the `F`-segment declarations that bind a funcall id, so any funcall on this path is unresolvable — see Review fixes). The walk memoizes `Cse` bodies by pointer identity.

On top of it, two Python entry points — `pounce.build_nl_problem` with a `pounce.NlExpr` node type, and `pounce.parse_nl_text` for a frontend that already emits `.nl` text. Both return the same `NlProblem` class `read_nl` does.

Comparison is spelled `NlExpr.compare("<", a, b)` rather than overloading `<`: returning an expression from `__lt__` breaks every ordinary use of an object in a container. Rejected alternative — exposing a bare `Tape` pyclass, which the issue floats as the long-term seam. A scalar tape only offers value/gradient/Hessian of one expression; everything a caller actually needs (Jacobian, structures, HVP, `variant`, `solve_nlp_batch`) lives at problem level, so the problem-level builder is the real seam and `Tape::build` is reached through it. It is also exposed directly as `NlExpr.eval` / `.gradient` / `.variables` for checking a subexpression in isolation.

`NlExpr` operands are deep-copied, so `+` chains are O(N²) and nesting is capped at `NlExpr.max_depth = 1000`. Both are covered under Review fixes.

**Hessian-vector products.** `NlTnlp::hessian_vector_products` seeds `Tape::hessian_directional` with the caller's directions. `eval_h` runs one such pass *per Hessian color* and decodes compressed columns into the sparse triangle; this runs one per direction, O(tape ops), independent of the chromatic number.

The block form is a separate entry point rather than a loop over the single-vector call because the forward sweep depends only on `x`: a block runs it once per tape and reuses `vals` across all `k` directions, where `k` separate calls redo it `k` times. All-zero directions are skipped, and the liveness scan is computed once up front rather than per (tape, direction) pair.

Exposed as `NlProblem.hessian_vector_product(x, v, lam=None, obj_factor=1.0)`, where `v` may be a dense vector, a dense `(n, k)` block, or any SciPy sparse vector or matrix; the result matches the input's shape. Everything routes through `numpy.asarray(..., float64)` so lists, integer arrays, and strided views (`V[:, 0]`, `w[::2]`) work rather than tripping the contiguity requirement. Sparse is recognized by duck-typing `.toarray()` — `pounce-py` must build without scipy, and anything that hands back a dense array is a legitimate input.

Two deliberate limits, both documented:

- **Sparse `v` is densified.** It is only `n × k`, and the sparsity worth exploiting is the model's, which the tapes already exploit — every pass is O(tape ops), never O(n²), whichever way `v` arrives.
- **The result is always dense.** `∇²L·v` is dense in general even when both factors are sparse; a sparse return type would advertise an economy the product does not have. The sparse Hessian itself is already available as `hessian_structure()` + `hessian(x)`.

A `(1, n)` row vector raises instead of being guessed at: for a square-ish block it is indistinguishable from `k` directions of the wrong length. Note this catches `csr_matrix(v_1d)`, which SciPy shapes as `(1, n)`.

**`Erf` op.** `TapeOp::Erf` / `UnaryOp::Erf` with closed-form arms — `erf'(u) = 2/√π·exp(-u²)`, `erf''(u) = -2·(u·erf'(u))` — factored into shared `erf_d1` / `erf_d2` helpers so the ~18 match arms cannot drift apart. Value via `libm` (the rust-lang port of musl's libm), not the Abramowitz–Stegun series: at ~1e-7 accuracy the series would cap the achievable KKT residual, which is the wrong tradeoff inside a residual. Reachable only through the in-memory path, since no `.nl` opcode maps to it.

## Review fixes (`39af24e`)

**Must-fix — SIGSEGV from ordinary Python.** Every consumer of an `Expr` recurses per nesting level (`Tape::build`, the derived `Clone`, the derived `Drop`), so a deep enough tree overflowed the stack with no exception to catch. Reproduced and bracketed on Linux with an 8 MB stack:

| build | survives | dies |
|---|---|---|
| release | 32 000 | 36 000 |
| **debug** | **4 000** | **8 000** |

Debug is the binding constraint, and a Python thread with a smaller stack is worse. `NlExpr` now tracks depth incrementally (O(1) per operation) and refuses past `NlExpr.max_depth = 1000`. Enforcing at *construction* is what makes the guarantee cover all three consumers — an expression that cannot be built can never be cloned or dropped into a crash. The reviewer's exact repro is now a `ValueError` naming the depth and pointing at `NlExpr.sum`.

1000 sits far below the smallest observed failure and far above what is reachable in practice, because building depth `D` through the operators is O(D²) anyway — the adjacent finding. Operators deep-copy their operands, so `e = e + t` in a loop copies everything built so far each iteration; `NlExpr.sum` builds one n-ary node, one level regardless of width, so the cap never binds on wide models. Documented on the class, in the error text, and in the book.

**`from_expressions` accepted an unusable `Expr::Funcall`.** There is nowhere to put the `F`-segment declarations that bind a funcall id, so any funcall on this path is unresolvable — and it surfaced as "AMPLFUNC is not set", advice the user cannot act on, since setting `AMPLFUNC` only moves the failure to "no `F<id>` declaration". Now rejected up front, pointing at `read_nl` / `parse_nl_text`, which do carry externals.

**The CSE memo protected only its own walk.** `collect_vars` and `collect_funcall_ids` re-entered a shared body once per reference rather than once per body — Θ(2^depth), and `get_variables_linearity` reaches `collect_vars` on every presolve. Both now memoize on `Arc` pointer identity, which cannot change either answer since both collect into a set.

**`erf_d2` returned NaN where the limit is 0.** `(-2u)·erf'(u)` is `inf * 0.0` for `|u| > f64::MAX/2`. Reassociated to `-2·(u·erf'(u))`.

**scipy floor.** `sp.csr_array(v_1d)` is only `(n,)` on scipy >= 1.14 — `(1, n)` on 1.11, raises on 1.13 — against a declared floor of >= 1.11. CI installs latest, which is how it hid. The 1-D sparse-array cases are now version-gated, with version-proof `(n, 1)` columns covering every floor.

Minor, all addressed: `erf → FbbtOp::Opaque` is pinned by a test; `__array_ufunc__ = None` makes `np.array([...]) * expr` raise instead of silently returning a `dtype=object` array of `NlExpr`; `__copy__`/`__deepcopy__` work and pickling is documented as unsupported; `decode_vec` accepts a strided `x`/`lam` instead of emitting numpy's unnamed "not contiguous"; the per-call liveness `Vec` moved to a reusable field; the `hessian_vector_product` rustdoc no longer says "in either shape"; `NlProblemParts` documents that an empty linear split makes genuinely linear rows report `NonLinear` to presolve; and the CHANGELOG's breaking-change note moved from `Added` to its own `Changed` heading.

## Corrections to the original PR body

Four claims were wrong. Recorded rather than quietly edited, since the review checked them:

| original claim | correction |
|---|---|
| libm is "MIT/Apache" | **MIT only** at the resolved 0.2.16 (it was `MIT AND (MIT OR Apache-2.0)` through 0.2.11, before libm moved into `rust-lang/compiler-builtins`). Permissive either way; the workspace `Cargo.toml` comment now says so accurately. |
| "`cargo clippy` clean on the touched crates" | Was **not** true — 15 net new warnings, mostly `useless conversion to the same type: pyo3::PyErr`. That lint fires once per `PyResult`-returning `#[pymethods]` function and points at the return type, where there is nothing to remove — the conversion is in macro-generated code. Now suppressed crate-wide next to the existing pyo3-macro allow, so the count cannot creep with each new binding. Measured after: **158 warnings across `pounce-py` + `pounce-nl`, identical to `main`'s 158** (down from 214 on the previous PR head). |
| "60 test binaries, all pass" | **175 targets, 2390 tests, 0 failures.** The 60 came from a `grep \| tail -60` truncating the log before I counted it — the count was of my own pipeline, not of the run. |
| "an `erf` model transparently uses the tape interpreter — the same fallback already taken for `tan`/`atan`" | `HessianProgram` has **zero callers**; the fast path is not wired into dispatch, so there is no runtime fallback to speak of. The `unreachable!` soundness argument still holds (`program_supports_op` is a positive allowlist, gated up front). Corrected here and in the CHANGELOG, where it was user-facing. |

## Blast radius

Additive. No existing API changed behavior.

- **Python package:** purely additive — same wheel, same `import pounce`, no new module. Three new names (`NlExpr`, `build_nl_problem`, `parse_nl_text`) and one new `NlProblem` method. Users need a rebuilt wheel since it is an extension change, not pure Python.
- **Rust:** `TapeOp` / `UnaryOp` gained a variant. Source-breaking only for an out-of-workspace crate matching on them exhaustively; nothing in-repo is affected. Now under its own `### Changed` CHANGELOG heading.
- **FBBT:** `erf` translates to `FbbtOp::Opaque`. It is monotone and bounded in (-1, 1), so it *could* propagate intervals exactly, but `FbbtOp` has no kernel for it; opaque means no tightening, never a wrong interval. A known gap, now pinned by a test.
- **Presolve:** `collect_vars` is no longer exponential on a shared DAG — a strict improvement on a pre-existing path.
- **New dependency:** `libm` 0.2 on `pounce-nl` — rust-lang, no_std, MIT, no transitive deps. Already in `Cargo.lock` transitively (the diff is a single `+ "libm"` line). `scripts/check-release-consistency.sh` passes.

## Tests

**These do not fail on the parent commit for an interesting reason** — they reference APIs that do not exist there, so they fail to compile / raise `AttributeError`. This is new capability, not a regression fix. What follows is how I know they bite instead.

Three mutations run:

| mutation | result |
|---|---|
| flip the sign of `erf_d2` | `erf_grad_and_hessian_match_fd` **FAILED**; `erf_directional_hessian_matches_accumulated` passed |
| scale `gpp` by 1.5 in the `hessian_directional` arm only | `erf_grad_and_hessian_match_fd` passed; `erf_directional_hessian_matches_accumulated` **FAILED** |
| un-memoize `collect_vars`' `Cse` arm | `shared_dag_walks_are_memoized_not_exponential` **hangs** (90 s timeout; instant with the memo) |

The first two splitting is the point of having both, and worth stating plainly: the cross-check is a *consistency* test, not a correctness one — with both sweeps calling the shared helper, a wrong shared formula agrees with itself. Finite differences catch that. Conversely the FD test only exercises `hessian_accumulate`, so it cannot see one match arm diverging from the other. Each covers the other's blind spot.

Also pinned: `erf` values against reference C99 doubles at six points plus oddness and saturation; `erf_d2` finite at 1e19 / 1e150 / 1e300 / `f64::MAX`; HVP against a densified `eval_h` Hessian on unit *and* mixed seeds; maximize sign (`max x²` → `∇² = -2`, a disagreement would have a Hessian-free step climbing where the sparse path descends); block form vs. `k` separate calls with a zero column between live ones; out-of-range `Var` driven through `Cse`, nested `Cse`, `Cond` branches, `MinList` and `Sum`; and `Funcall` rejection both bare and buried behind a `Cse` in a constraint.

The 55 Python cases run every direction shape against **two model shapes**, because they hide different bugs. A chain objective gives a tridiagonal Hessian — the IPM shape, asserted at `2n-1` entries, where most of the matrix is structural zero and leaked coupling between distant variables would vanish in a 3×3 test; the densify-via-HVP test asserts every entry *outside* the stored sparsity is exactly zero, which a sparse-only comparison cannot check. A fully coupled objective gives a dense Hessian (`n(n+1)/2` entries) where nothing is masked by a zero. Both are checked against the sparse COO triangle *and* its dense expansion.

Not mutation-tested: the Python direction decoder, which needs a wheel rebuild per mutation. Its coverage is structural instead — blocks are non-square (n=10/6, k=4), so a fill-order or transposition bug cannot pass the shape check, and each `block[:, c]` is compared against an independent single-vector call as well as against a scipy-assembled `H @ V`. The identity-block test alone could *not* catch a transposition, since the Hessian is symmetric; the non-square block test is what pins column order.

Verification run: `cargo fmt --all -- --check` clean; CI's clippy gate (`-D clippy::correctness -D clippy::suspicious`) passes on `--workspace --exclude pounce-hsl`; warning count at parity with `main` (158 = 158); `cargo test --workspace --exclude pounce-hsl` — **175 targets, 2390 tests, 0 failures**; `pytest python/tests` — **662 passed, 33 skipped**. `pounce-hsl` is excluded because it cannot link without a CoinHSL install, which is equally true on the parent commit; CI excludes it for the same reason.

---

- [ ] ~~Tests fail on the parent commit for the stated reason~~ — N/A for new capability; mutation results given above instead
- [x] Book page under `docs/src/` updated — `docs/src/python.md`, already linked from `SUMMARY.md`


jkitchin (commented):
## Review

The engineering is strong and the core math is right. One must-fix crash, three should-fixes, and four claims in the PR body that are not true.

Reviewed with four independent passes (Rust tape/erf, HVP algorithm, Python bindings, claims audit), each building and running code rather than reading it.

## What checks out

- **HVP is correct.** Verified against 36 real `.nl` fixtures with random `x`, mixed-sign `lam`, `obj_factor = 0.85`, and a 3-column block containing a zero column — worst relative error **9.07e-16**. Also verified against an *independent* finite-difference reference assembled from `eval_grad_f` + `eval_jac_g` (which does not share `hessian_directional`), on a 5-var / 3-constraint model with `exp`/`pow`/`sin`/`log`/`erf`/`sqrt`: 20/20 comparisons agreed. No triangle double-count, no transpose, no stale scratch — `out.fill(0.0)` precedes everything and `hessian_directional` only `+=`s, so a skipped column stays exactly zero (probed with `out` pre-poisoned to 9.99e30). Maximize sign is consistent with `eval_h` *and* `eval_grad_f`, including the mixed case where the objective block negates and the constraint block does not.
- **The performance claim is real**, measured on release builds: `hessian()` / `hessian_vector_product()` = 5.0x at n=30, 23.8x at n=100, **55.2x at n=200** — the ratio tracks the chromatic number as claimed. Block form beats `k` separate calls by 1.30x-5.40x. (Worth knowing: in a *debug* build the block form measured slightly slower, 0.80x.)
- **erf math is right.** `erf'(u) = 2/sqrt(pi)*exp(-u^2)` and `erf''(u) = -2u*erf'(u)` both correct. `TapeOp::Erf` appears at exactly the same 16 `TapeOp` match sites as the reference unary `Atanh`, plus the 2 `UnaryOp -> TapeOp` mappings = the claimed 18. No `_ =>` catch-all swallows it. The tangent and adjoint arms agree with analytic `erf''` to < 1e-14 relative at 14 points.
- **The `Var` bounds check is complete** — out-of-range indices were driven through `Cse`, `Cse(Cse(..))`, `Cond.then_`, and `Funcall` args; all rejected. The pointer-identity CSE memo is sound (the `Arc`s stay alive for the whole function body).
- **Tests are unusually good** — 43 cases covering strided views, Fortran-order columns, empty `(4, 0)` blocks, integer dtypes, and exact error-message shapes. All 5 `docs/src/python.md` examples run verbatim and produce correct output. Full workspace: **2386 tests, 0 failures, 0 new rustc warnings.**

## Must fix

### 1. SIGSEGV from ordinary Python — no exception, no panic, the interpreter dies

`crates/pounce-py/src/nl_expr.rs:434-454` (`Tape::build`, reached from `eval` / `gradient` / `variables`) and the operator methods at `:290-346` (recursive `Expr::clone` / `Drop`) recurse over expression depth with no guard. Reproduced independently:

```python
e = pounce.NlExpr.var(0)
for _ in range(40000):
    e = e + 1.0
e.variables()        # Segmentation fault: 11   (exit 139)
```

30 000 survives; 40 000 dies on the query; construction alone dies between 40 000 and 50 000 (macOS, 8 MB main stack). This is specific to the single-expression path the PR advertises for "checking a subexpression in isolation" — `build_nl_problem` on the *same* 40 000-deep expression is fine, because the problem-level tape flattens Add chains. `__repr__` is safe (capped by `node_count_capped`).

Fix: an explicit depth cap raising `ValueError`, an iterative walk, or a spawned thread with a large stack.

**Adjacent, and the same road:** the "cheap to clone" note at `nl_expr.rs:53-59` is misleading. Each operator deep-copies the LHS, so accumulating in a loop is **O(N^2)** in wall time — measured 17 s at N = 20 000 and 43 s at N = 40 000, versus ~0 s for `NlExpr.sum` over 200 000 terms. `e += ...` is the first thing a modeler writes, and it ends at the segfault above. Worth an explicit "use `NlExpr.sum`; `+` chains are quadratic" next to the depth fix.

## Should fix

### 2. `from_expressions` accepts an `Expr::Funcall` it can never evaluate

`crates/pounce-nl/src/nl_reader.rs:349` hardcodes `imported_funcs: Vec::new()`, and `NlProblemParts` has no field for them, so *any* `Funcall` on this path is guaranteed unusable. It is accepted anyway, and fails later with:

```
problem uses external functions but AMPLFUNC is not set;
set AMPLFUNC to a newline-separated list of AMPL shared-library paths
```

`AMPLFUNC` cannot fix it — setting it just moves the failure to `nl_external.rs:83` (`funcall id 0 has no F<0> declaration`). This is the same class of error the PR's own doc comment at `nl_reader.rs:74-76` says must be "caught while it is still a diagnosable user error". `max_var_index` already walks `FuncallArg::Real`, so rejecting `Expr::Funcall { .. }` there is a two-line fix.

### 3. A new test needs scipy >= 1.14, but the declared floor is >= 1.11

`python/tests/test_nl_build.py:499-500` uses `sp.csr_array(dense_v)` with the comment `# 1-D sparse array (n,)`, while `python/pyproject.toml:28` declares `scipy>=1.11`. Measured against real installs:

| scipy | `sp.csr_array(v_1d).shape` | `sp.coo_array(v_1d).shape` | outcome |
|---|---|---|---|
| 1.11.4 | `(1, 3)` | `(1, 3)` | **test FAILS** — decoder rejects the row vector |
| 1.13.1 | raises `ValueError` | `(3,)` | **errors inside scipy** |
| 1.14.1 / 1.18 | `(3,)` | `(3,)` | passes |

CI installs latest, which is exactly how this hides. Fix: raise the floor, guard by version, or use the version-proof `sp.csc_matrix(dense_v[:, None])` already present in the same tuple.

### 4. `erf_d2` returns NaN where the limit is 0

`crates/pounce-nl/src/nl_tape.rs:148`: `-2.0 * u * erf_d1(u)` evaluates left-to-right as `(-2*u) * erf_d1(u)`, so for `|u| > f64::MAX/2 ~ 9e307` that is `inf * 0.0 = NaN`. Verified:

```
u=1.798e+308   erf_d1=0.0   as-written = nan     reassociated = -0.0
u=1.000e+150   erf_d1=0.0   as-written = -0.0    reassociated = -0.0
```

Reassociating to `-2.0 * (u * erf_d1(u))` fixes every finite magnitude (`u = +/-inf` stays NaN either way). The value and gradient arms are correct at every magnitude; only the Hessian arms are affected. Low practical severity — `.nl`'s unbounded sentinel is +/-1e19 — but it is free to fix, and an unguarded `x` now reaches it through the new Python binding.

### 5. The CSE memo protects only its own walk

`nl_reader.rs:325-328` justifies the pointer-identity memo by the exponential blowup "on a DAG a frontend built by sharing subexpressions". It fixes that walk, but the next two things a caller touches have no memo and are still exponential on the same input: `collect_vars` (`nl_reader.rs:1686`) and `collect_funcall_ids` (`nl_external.rs:125`). `get_variables_linearity` reaches `collect_vars`, and **pounce-presolve calls it on every solve** (`crates/pounce-presolve/src/lib.rs:855`). Measured on a release build, balanced share-DAG of depth `d`:

| depth | `from_expressions` | `NlTnlp::try_new` | `get_variables_linearity` |
|---|---|---|---|
| 20 | 4 us | 5.56 ms | 7.09 ms |
| 22 | 3 us | 24.1 ms | 27.9 ms |
| 26 | 101 us | 512 ms | **3.00 s** |

4x per +2 depth = Theta(2^d), vs O(1) for the memoized walk. Pre-existing in those helpers, but `from_expressions` is the new door that makes such a DAG trivially constructible from a frontend — so either memoize them the same way, or drop the claim that the memo protects this input class.

## Claims in the PR body that are not true

| claim | reality |
|---|---|
| libm is "MIT/Apache" | `Cargo.lock` resolves **libm 0.2.16, which is MIT only** (crates.io confirms; 0.2.11 was `MIT AND (MIT OR Apache-2.0)`, before libm moved into `rust-lang/compiler-builtins`). Fine for an EPL project — just state it accurately. Everything else in that bullet is confirmed: rust-lang, no_std, zero transitive deps, already in `Cargo.lock` at 610 on main (the diff is a single `+ "libm"` line), and `check-release-consistency.sh` passes. |
| "`cargo clippy` clean on the touched crates (the workspace's pre-existing warnings are all in untouched files)" | **15 net new warnings.** Baseline on main for `pounce-py`: 11 in `nl_problem.rs`. On this head: 26 across `nl_problem.rs` + the new `nl_expr.rs`, mostly `useless conversion to the same type: pyo3::PyErr`. CI's actual gate (`-D clippy::correctness -D clippy::suspicious`) **does** pass, so the build is safe — the statement is what is wrong. |
| "60 test binaries, all pass" | **152** test binaries plus 23 doc-test targets (175 targets, 2386 tests). All pass, which is the substantive claim; the number is just wrong. |
| "an `erf` model transparently uses the tape interpreter — the same fallback already taken for `tan`/`atan`" | `HessianProgram` has **zero callers**: `git grep` across the branch returns only `pub mod nl_hessian_program;` in `crates/pounce-cli/src/lib.rs:24`. The `unreachable!` soundness argument holds (`program_supports_op` is a positive allowlist of 15 ops, gated up front at `:358`), but there is no runtime fallback, because the fast path is not wired into dispatch. This same framing is repeated in the CHANGELOG, where it is user-facing. |

Also: the rustdoc at `crates/pounce-py/src/nl_problem.rs:433` — "any SciPy sparse vector or `(n, k)` matrix, **in either shape**" — contradicts both the implementation and the PR's own stated design decision (a `(1, n)` row raises). That string is what `help(NlProblem.hessian_vector_product)` prints at the REPL. Concretely:

```
csr_matrix(np.ones(2))  shape (1,2) -> ValueError (row-vector)
csc_matrix(np.ones(2))  shape (1,2) -> ValueError (row-vector)
coo_array(np.ones(2))   shape (2,)  -> OK
csr_array(np.ones(2))   shape (2,)  -> OK
```

Only the 1-D sparse *array* API works, and that needs scipy >= 1.14 (see finding 3). `docs/src/python.md` is better — its table says "SciPy sparse vector, `(n, 1)` column, or `(n, k)` matrix" — but the `(1, n)` exception lands 33 lines later at `:319`.

## Minor

- `decode_vec` (`nl_problem.rs:76`) does not route `x` / `lam` through `np.asarray`, so a strided `x` gives a bare `TypeError: The given array is not contiguous` — no argument name, no shape advice — in the very same call that accepts a strided `v`, under a docstring promising "any dtype or stride".
- No test pins `erf -> Opaque`. `inverse_trig_translate_to_opaque` still loops only `[Tan, Atan, Acos]`. (The translation itself is correct and sound: `Opaque` yields `Interval::ENTIRE` forward and nothing reverse, and ENTIRE contains (-1, 1).)
- `np.array([1., 2.]) * x[0]` silently returns a `dtype=object` array of `NlExpr` (the forward form `x[0] * np.array(...)` correctly raises). `__array_ufunc__ = None` on the pyclass makes the reflected form raise too.
- `NlExpr` cannot be pickled or deepcopied (`TypeError: cannot pickle 'pounce.NlExpr' object`) — a frontend cloning its DAG hits this; undocumented either way.
- The CHANGELOG's breaking-change note lives under `### Added` rather than `### Changed`, so a reader scanning headings for breakage will miss it.
- `live: Vec<bool>` (`nl_reader.rs:2608`) is heap-allocated per call in a crate that has an explicit no-per-call-allocation test for this path (`crates/pounce-nl/tests/tape_gradient_no_alloc.rs`); the headline Newton-Krylov use is `k = 1`, where it is pure overhead.
- The empty linear split makes a genuinely linear row report `NonLinear` (`get_constraints_linearity` tags `Linear` only when `con_nonlinear[i]` is literally `Const(0.0)`), and presolve consumes that. Conservative-safe — it just loses tightening — but `NlProblemParts`'s doc explains the evaluator half and omits the metadata cost.
- Worth documenting, not a defect: HVP is deliberately **not** bit-equivalent to a dense `H @ v` when `v` carries NaN/Inf, because structural zeros do not propagate NaN through AD. With a block-diagonal Hessian and `v = [nan, 0, 1, 0]`, dense gives `[nan nan nan nan]` and HVP gives `[nan nan 2.42 3.08]`. Arguably better semantics — but any future "matches `H @ v` exactly" claim would be false.

Filing the segfault separately, since it is a crash in a newly public API and worth tracking independently of this PR's merge.

---

jkitchin:
All findings addressed in `39af24e`. The four claims audit was fair — three were sloppy, and one of those was a measurement of my own tooling rather than the code. PR body corrected in place with the corrections recorded rather than quietly edited.

## Must fix

**1. SIGSEGV.** Reproduced and bracketed here on Linux with an 8 MB stack, which puts the binding constraint somewhere you didn't measure:

| build | survives | dies |
|---|---|---|
| release | 32 000 | 36 000 |
| **debug** | **4 000** | **8 000** |

A debug-built extension fails at an order of magnitude lower depth, so the cap had to have real margin. `NlExpr` now carries a depth field, maintained incrementally (O(1) per operation, no walk), and refuses past `NlExpr.max_depth = 1000`. Your exact repro is now:

```
ValueError: NlExpr: expression nesting would reach depth 1001, past the limit of 1000
```

Chose construction-time enforcement over a big-stack thread specifically because of the second and third consumers you named: a spawned thread fixes `Tape::build`, but `Drop` runs at GC time on whatever thread the interpreter happens to be on, and `Clone` likewise. Capping at construction means an expression that cannot be built can never be cloned or dropped into a crash — one invariant covering all three.

On the adjacent O(N²): agreed, and it is now in three places — the class docstring, the book, and the error text itself, since that error is exactly where a modeler writing `e += ...` will land. Worth noting the two findings are load-bearing for each other: the cap is only cheap *because* the quadratic makes depth >1000 impractical anyway, so it costs no real capability. `NlExpr.sum` is depth 1 regardless of width, and there is a test that 200 000 terms through it is fine.

## Should fix

**2. `Funcall`.** Fixed — `max_var_index` became `validate_expr` and rejects it, pointing at `read_nl` / `parse_nl_text`. Your framing of *why* is what I put in the doc comment: `AMPLFUNC` cannot fix it, so surfacing that message is advice the user cannot act on. Tested bare and buried behind a `Cse` in a constraint.

**3. scipy floor.** Fixed by version-gating the 1-D sparse-array cases, keeping the version-proof `(n, 1)` columns unconditional so coverage holds at the declared floor. Your table was the useful part — I would not have guessed 1.13 *raises* rather than returning the 1.11 shape.

**4. `erf_d2` NaN.** Reassociated to `-2·(u·erf'(u))`, with a test at 1e19 / 1e150 / 1e300 / `f64::MAX` / `f64::MAX/2`.

**5. CSE memo scope.** You are right that the claim outran the fix. Memoized both `collect_vars` and `collect_funcall_ids` rather than dropping the claim — the `get_variables_linearity` → presolve-on-every-solve path made that the better side to fix. Both collect into a set, so skipping a repeat visit cannot change the answer. Pinned by a depth-30 share-DAG test, which I verified bites: un-memoizing the `Cse` arm makes it hang (90 s timeout), instant with the memo.

## Claims

- **libm license** — MIT only at 0.2.16. Corrected in the PR body and in the workspace `Cargo.toml` comment, with the 0.2.11 history noted so the next person does not "fix" it back.
- **clippy** — you were right, and the lint is a pyo3 macro artifact: it fires once per `PyResult`-returning `#[pymethods]` function and points at the return type, where there is nothing to remove. Suppressed crate-wide next to the existing pyo3-macro allow, so the count cannot creep with each new binding. Measured after: **158 across `pounce-py` + `pounce-nl`, identical to `main`'s 158** (down from 214 on the previous head — it also clears the 11 that predate this PR).
- **test count** — 175 targets, 2390 tests, matching your 175 exactly. My "60" came from `grep ... | tail -60`: I counted my own pipeline's output, not the run. Instructive error, and the reason the count is now taken from a full log.
- **`HessianProgram`** — zero callers, so "transparently falls back" describes a dispatch path that does not exist. Corrected in the body and in the CHANGELOG, since that one was user-facing.
- **`in either shape` rustdoc** — fixed, and the `csr_matrix(v_1d)` → `(1, n)` trap you enumerated is now called out in both the rustdoc and the book, immediately under the shape table rather than 33 lines later.

## Minor

All taken: `decode_vec` accepts a strided `x`/`lam`; `erf → Opaque` is pinned by a test (added to `inverse_trig_translate_to_opaque`); `__array_ufunc__ = None` so the reflected form raises like the forward one; `__copy__`/`__deepcopy__` work and pickling is documented as unsupported; the liveness `Vec` moved to a reusable field; `NlProblemParts` documents the `NonLinear` tag cost to presolve; CHANGELOG breaking note moved to its own `### Changed` heading. The NaN/Inf semantics are now documented in the book with your block-diagonal example — good catch that it forecloses a future "matches `H @ v` exactly" claim.

Two things I did **not** change, both deliberate:

- **The result of `hessian_vector_product` stays dense**, including for sparse input, and sparse `v` is still densified on the way in. `H·v` is dense in general even when both factors are sparse.
- **`FbbtOp` still has no `Erf` kernel.** You confirmed `Opaque` is sound; adding a kernel is a separate change with its own correctness surface, and I would rather not smuggle it in here.

Thanks for running it rather than reading it — the debug-vs-release stack split, the scipy 1.13 behavior, and the `HessianProgram` caller count are all things a read-through would have missed.

Verification after the fixes: 175 targets / 2390 tests / 0 failures; 662 Python tests passed, 33 skipped; CI's clippy gate passes; warning count at parity with `main`.

---

PR #482, opened 2026-08-04
Fix pounce hanging using WASM

GermanHeim:
## Problem

I was trying to create an oximo-pounce example that runs in the browser using WASM, similar to the pyomo demo.
Currently, pounce hangs using wasm after:

```text
Number of nonzeros in equality constraint Jacobian...:        0
Number of nonzeros in inequality constraint Jacobian.:        4

Total number of variables............................:        2
                variables with lower and upper bounds:        2
                     variables with only upper bounds:        0
Total number of equality constraints.................:        0
Total number of inequality constraints...............:        2
        inequality constraints with only lower bounds:        1
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:        1
```

## Cause

The causes are various:
A first Wasm-specific stall is exposed after bypassing the optional multiplier solve. Pounce hangs while installing its per-callback thread-local inspector context. That context exists only for C API inspect current iterate calls and is not needed by the builder.

Another hang was `pounce_observability::iteration_event_wanted()` touching a capture TLS slot on wasm32-wasip1.

Another stall was inside `feral::Solver::factor`, after the KKT matrix is assembled. This is the parallel FERAL driver attempting to initialize/use Rayon in a single- threaded WASI instance.

Your .nl demo avoided enough of that path to look fine. 

## Fix

I’ve made those facilities no-op/serial only on Wasm. Native behavior remains unchanged.

## Blast radius

None of the actual changes affect the model or the main IPM. They only remove startup heuristics unavailable in this host.

**Note:** The issues were diagnosed using ChatGPT Sol 5.6.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,8 @@ jobs:
   # wasm32 (a new `libc`/`dlopen`/thread call somewhere in the tree), and one
   # that compiles but traps at run time on a syscall WASI does not have. The
   # smoke test drives the same C ABI the page uses — load a `.nl`, solve it,
-  # check the objective — under Node's WASI.
+  # check the objective — under Node's WASI. It also calls a pounce-rs builder
+  # entry point.
   wasm:
     name: WebAssembly build + smoke
     runs-on: ubuntu-latest
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1152,6 +1152,7 @@ dependencies = [
  "pounce-nl",
  "pounce-nlp",
  "pounce-presolve",
+ "pounce-rs",
  "pounce-solve-report",
  "serde_json",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,7 @@ pounce-nl = { path = "crates/pounce-nl", version = "0.9.0" }
 pounce-hsl = { path = "crates/pounce-hsl", version = "0.9.0" }
 pounce-feral = { path = "crates/pounce-feral", version = "0.9.0" }
 pounce-algorithm = { path = "crates/pounce-algorithm", version = "0.9.0" }
+pounce-rs = { path = "crates/pounce-rs", version = "0.9.0" }
 pounce-restoration = { path = "crates/pounce-restoration", version = "0.9.0" }
 pounce-presolve = { path = "crates/pounce-presolve", version = "0.9.0" }
 pounce-l1penalty = { path = "crates/pounce-l1penalty", version = "0.9.0" }
--- a/crates/pounce-algorithm/src/application.rs
+++ b/crates/pounce-algorithm/src/application.rs
@@ -1846,13 +1846,6 @@ impl IpoptApplication {
         // per-problem `.opt` files can flip backend knobs without
         // rebuilding pounce.
         let mut feral_cfg = feral_config_from_options(&self.options);
-        // wasm32-wasip1 browser and Node hosts used by pounce-wasm have no
-        // native threads. FERAL's Rayon-backed factor driver can block while
-        // creating its pool, so make the backend explicitly serial.
-        #[cfg(target_arch = "wasm32")]
-        {
-            feral_cfg.parallel = Some(false);
-        }
         // Block-triangular / Schur KKT partition (pounce#180 item 2). Configure
         // the Schur block solvers from the *base* feral cfg: a full-KKT external
         // ordering (item 1) is sized for the whole system and cannot apply to
@@ -3108,6 +3101,13 @@ pub fn feral_config_from_options(
         if let Some(s) = pounce_feral::parse_scaling_strategy(&v) {
             cfg.scaling = s;
         }
+    }
+    // wasm32-wasip1 browser and Node hosts used by pounce-wasm have no
+    // native threads. FERAL's Rayon-backed factor driver can block while
+    // creating its pool, so make the backend explicitly serial.
+    #[cfg(target_arch = "wasm32")]
+    {
+        cfg.parallel = Some(false);
     }
     cfg
 }
--- a/crates/pounce-algorithm/src/application.rs
+++ b/crates/pounce-algorithm/src/application.rs
@@ -1846,6 +1846,13 @@ impl IpoptApplication {
         // per-problem `.opt` files can flip backend knobs without
         // rebuilding pounce.
         let mut feral_cfg = feral_config_from_options(&self.options);
+        // wasm32-wasip1 browser and Node hosts used by pounce-wasm have no
+        // native threads. FERAL's Rayon-backed factor driver can block while
+        // creating its pool, so make the backend explicitly serial.
+        #[cfg(target_arch = "wasm32")]
+        {
+            feral_cfg.parallel = Some(false);
+        }
         // Block-triangular / Schur KKT partition (pounce#180 item 2). Configure
         // the Schur block solvers from the *base* feral cfg: a full-KKT external
         // ordering (item 1) is sized for the whole system and cannot apply to
--- a/crates/pounce-algorithm/src/application.rs
+++ b/crates/pounce-algorithm/src/application.rs
@@ -3105,7 +3105,7 @@ pub fn feral_config_from_options(
     // wasm32-wasip1 browser and Node hosts used by pounce-wasm have no
     // native threads. FERAL's Rayon-backed factor driver can block while
     // creating its pool, so make the backend explicitly serial.
-    #[cfg(target_arch = "wasm32")]
+    #[cfg(target_os = "wasi")]
     {
         cfg.parallel = Some(false);
     }
--- a/crates/pounce-algorithm/src/application.rs
+++ b/crates/pounce-algorithm/src/application.rs
@@ -381,6 +381,8 @@ impl IpoptApplication {
     /// populated with one [`pounce_nlp::solve_statistics::IterRecord`]
     /// per accepted iterate. Off by default — the `pounce_sens` and
     /// `pounce` binaries opt in when `--json-output` is passed.
+    /// The `wasm32-wasip1` build has no in-process tracing collector, so its
+    /// iteration vector remains empty even when this flag is enabled.
     pub fn enable_iter_history(&mut self) {
         self.record_iter_history = true;
     }
--- a/crates/pounce-algorithm/src/intermediate.rs
+++ b/crates/pounce-algorithm/src/intermediate.rs
@@ -42,13 +42,21 @@ pub struct CtxGuard {
 
 impl CtxGuard {
     pub fn install(ctx: IntermediateContext) -> Self {
+        // wasm32-wasip1 browser/Node hosts are single-threaded here, and
+        // touching this destructor-bearing TLS slot can block indefinitely.
+        // The context only serves optional C-API inspector calls during the
+        // callback.
+        #[cfg(not(target_arch = "wasm32"))]
         CURRENT_CTX.with(|c| *c.borrow_mut() = Some(ctx));
+        #[cfg(target_arch = "wasm32")]
+        let _ = ctx;
         Self { _private: () }
     }
 }
 
 impl Drop for CtxGuard {
     fn drop(&mut self) {
+        #[cfg(not(target_arch = "wasm32"))]
         CURRENT_CTX.with(|c| *c.borrow_mut() = None);
     }
 }
@@ -59,10 +67,23 @@ pub fn with_current<F, R>(f: F) -> Option<R>
 where
     F: FnOnce(&IntermediateContext) -> R,
 {
+    #[cfg(target_arch = "wasm32")]
+    {
+        // Live-iterate inspection is unavailable on the Wasm build because
+        // its callback context is deliberately not installed above.
+        let _ = f;
+        None
+    }
+    #[cfg(not(target_arch = "wasm32"))]
     CURRENT_CTX.with(|c| c.borrow().as_ref().map(f))
 }
 
 /// Whether a context is currently installed.
 pub fn is_active() -> bool {
+    #[cfg(target_arch = "wasm32")]
+    {
+        false
+    }
+    #[cfg(not(target_arch = "wasm32"))]
     CURRENT_CTX.with(|c| c.borrow().is_some())
 }
--- a/crates/pounce-algorithm/src/intermediate.rs
+++ b/crates/pounce-algorithm/src/intermediate.rs
@@ -18,6 +18,7 @@ use crate::ipopt_cq::IpoptCqHandle;
 use crate::ipopt_data::IpoptDataHandle;
 use crate::ipopt_nlp::IpoptNlp;
 use std::cell::RefCell;
+use std::mem::MaybeUninit;
 use std::rc::Rc;
 
 /// Snapshot stashed in TLS for the duration of one
@@ -29,8 +30,44 @@ pub struct IntermediateContext {
     pub nlp: Rc<RefCell<dyn IpoptNlp>>,
 }
 
+/// TLS storage without compiler-registered destructor glue.
+/// Keeping the slot non-`Drop` also makes it safe for wasm32-wasip1 hosts.
+struct ContextSlot {
+    value: MaybeUninit<Option<IntermediateContext>>,
+    initialized: bool,
+}
+
+impl ContextSlot {
+    const fn new() -> Self {
+        Self {
+            value: MaybeUninit::uninit(),
+            initialized: false,
+        }
+    }
+
+    fn get(&self) -> Option<&Option<IntermediateContext>> {
+        self.initialized.then(|| {
+            // SAFETY: `initialized` is set only after `value` is initialized.
+            unsafe { self.value.assume_init_ref() }
+        })
+    }
+
+    fn get_mut(&mut self) -> &mut Option<IntermediateContext> {
+        if !self.initialized {
+            self.value.write(None);
+            self.initialized = true;
+        }
+        // SAFETY: the branch above initializes `value` before this access.
+        unsafe { self.value.assume_init_mut() }
+    }
+
+    fn replace(&mut self, value: Option<IntermediateContext>) -> Option<IntermediateContext> {
+        std::mem::replace(self.get_mut(), value)
+    }
+}
+
 thread_local! {
-    static CURRENT_CTX: RefCell<Option<IntermediateContext>> = const { RefCell::new(None) };
+    static CURRENT_CTX: RefCell<ContextSlot> = const { RefCell::new(ContextSlot::new()) };
 }
 
 /// RAII guard — installs `ctx` on construction, clears on drop. Used
@@ -42,22 +79,18 @@ pub struct CtxGuard {
 
 impl CtxGuard {
     pub fn install(ctx: IntermediateContext) -> Self {
-        // wasm32-wasip1 browser/Node hosts are single-threaded here, and
-        // touching this destructor-bearing TLS slot can block indefinitely.
-        // The context only serves optional C-API inspector calls during the
-        // callback.
-        #[cfg(not(target_arch = "wasm32"))]
-        CURRENT_CTX.with(|c| *c.borrow_mut() = Some(ctx));
-        #[cfg(target_arch = "wasm32")]
-        let _ = ctx;
+        CURRENT_CTX.with(|c| {
+            c.borrow_mut().replace(Some(ctx));
+        });
         Self { _private: () }
     }
 }
 
 impl Drop for CtxGuard {
     fn drop(&mut self) {
-        #[cfg(not(target_arch = "wasm32"))]
-        CURRENT_CTX.with(|c| *c.borrow_mut() = None);
+        CURRENT_CTX.with(|c| {
+            c.borrow_mut().replace(None);
+        });
     }
 }
 
@@ -67,23 +100,20 @@ pub fn with_current<F, R>(f: F) -> Option<R>
 where
     F: FnOnce(&IntermediateContext) -> R,
 {
-    #[cfg(target_arch = "wasm32")]
-    {
-        // Live-iterate inspection is unavailable on the Wasm build because
-        // its callback context is deliberately not installed above.
-        let _ = f;
-        None
-    }
-    #[cfg(not(target_arch = "wasm32"))]
-    CURRENT_CTX.with(|c| c.borrow().as_ref().map(f))
+    CURRENT_CTX.with(|c| c.borrow().get().and_then(|ctx| ctx.as_ref().map(f)))
 }
 
 /// Whether a context is currently installed.
 pub fn is_active() -> bool {
-    #[cfg(target_arch = "wasm32")]
-    {
-        false
+    CURRENT_CTX.with(|c| c.borrow().get().is_some_and(Option::is_some))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::ContextSlot;
+
+    #[test]
+    fn context_slot_has_no_thread_local_drop_glue() {
+        assert!(!std::mem::needs_drop::<ContextSlot>());
     }
-    #[cfg(not(target_arch = "wasm32"))]
-    CURRENT_CTX.with(|c| c.borrow().is_some())
 }
--- a/crates/pounce-observability/src/lib.rs
+++ b/crates/pounce-observability/src/lib.rs
@@ -38,8 +38,8 @@
 
 #![cfg_attr(test, allow(clippy::unwrap_used, clippy::expect_used))]
 
-#[cfg(not(target_arch = "wasm32"))]
 use std::cell::RefCell;
+use std::mem::MaybeUninit;
 
 use pounce_common::types::{Index, Number};
 use pounce_nlp::solve_statistics::IterRecord;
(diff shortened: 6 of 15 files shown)


jkitchin (commented):
Thanks for chasing this down — getting a browser solve to run end to end is worth having, and the native-safety discipline here (every hunk behind a `cfg`, nothing changed for native builds) is exactly right.

**Verified locally:** fetched `pull/482/head` (be283cc), ran `rustup target add wasm32-wasip1 && cargo check -p pounce-wasm --target wasm32-wasip1` — clean, zero warnings. All 6 CI checks green.

## What it does

1. `application.rs:1849` — force `feral_cfg.parallel = Some(false)` on wasm32.
2. `intermediate.rs` — don't install the per-callback `CURRENT_CTX` TLS slot; `with_current`/`is_active` become `None`/`false`.
3. `pounce-observability` — `CAPTURE` TLS and every accessor compiled out; `iteration_event_wanted()` returns `false` (after the `JSON_LOGGING` check, so the console/JSON stream is preserved), `IterCaptureGuard` becomes a no-op returning an empty `Vec`.
4. `pounce-wasm/build.sh` — add `--enable-nontrapping-float-to-int --enable-simd` to `wasm-opt`.

My concerns are about *which* of these is actually load-bearing, and the absence of anything that would catch a regression.

## Blocking-ish

**1. The PR body and the diff disagree — cause #1 has no fix.** The body names `calculate_y_eq` (the optional least-squares eq-multiplier init) as the first stall, and "Blast radius" lists "the optional least-squares multiplier initializer" as one of two wasm omissions. There is no such change in the diff. Either the body is stale or a commit was dropped. This matters beyond bookkeeping: if that stall dissolved once the feral-serial fix landed, the same is plausibly true of changes 2 and 3. Could you re-test each of the three code changes reverted individually against the reproducer and keep only what's required? Two of them remove real functionality, and the diagnosis chain hasn't been narrowed by bisection.

**2. No regression test, and CI cannot see this bug.** `.github/workflows/ci.yml` already has a `WebAssembly build + smoke` job that builds for wasm32-wasip1 and solves a `.nl` under Node's WASI — green on `main` and green here. This PR would pass CI identically with all four changes reverted. Given the failing case is a 2-variable / 2-inequality problem, adding it as a `.nl` alongside `tests/simple.nl` plus a second case in `smoke.mjs` looks cheap. If the hang only reproduces through the `pounce-rs` builder path (which is what an `oximo` integration would use) rather than the `.nl` entry point, it's worth saying so explicitly in the body — that's an important clue about the real cause, and it means the reproducer needs a small wasm-side entry point.

**3. The feral override is in the wrong layer.** `application.rs:1852` mutates the *fallback* config, but line 1874 is `self.linear_backend_factory.take().unwrap_or_else(|| default_backend_factory_with_sink(feral_cfg, …))` — any caller that installs its own factory bypasses it, and `pounce-cli`, `pounce-cinterface`, and `pounce-py` all do. `pounce-wasm` and `pounce-rs` don't, so the fix works for today's consumers, but it's a trap. Moving it into `feral_config_from_options` (`application.rs:3052`) covers every caller from one place, including `FeralBackendPool::serial`. Semantics are safe — `crates/pounce-feral/src/lib.rs:1093-1122` documents and tests serial vs. parallel factors as bit-identical.

## Should address

**4. Silent regression in a documented public API.** `pounce_rs::…::capture_iterations()` (`crates/pounce-rs/src/builder.rs:204-206`) promises "one `IterRecord` per Newton iteration"; on wasm it now returns an empty vec, indistinguishable from "you didn't enable it" (`solve_statistics.rs:125` documents empty as exactly that). Same for `enable_iter_history()` via `pounce-py`/CLI. `docs/src/wasm.md` has a **Limitations** section (line 183) with precisely this kind of entry — two bullets there would cover it: no in-process iteration-history capture, and no live-iterate inspection (`GetIpoptCurrentIterate`/`GetIpoptCurrentViolations` report "not available").

**5. `target_arch = "wasm32"` is broader than every comment justifying it.** All five comments say wasm32-wasip1, but the cfg also catches `wasm32-unknown-emscripten` — which *does* have pthreads and working TLS — and `wasm32-unknown-unknown`. `target_os = "wasi"` matches the stated reasoning. Not hypothetical: `docs/src/wasm.md` and `web-python/README.md` both discuss Pyodide, and an emscripten build of `pounce-py` would silently lose iteration history for no reason.

**6. Root-cause skepticism on the TLS claim.** Both slots are `const {}`-initialized (`intermediate.rs:33`, `pounce-observability/src/lib.rs:66`), so there is no lazy-init path to block on — the only wasm-specific machinery left is drop-glue registration, and "blocks indefinitely" is not a known failure mode for that on wasip1 (a trap or abort is). If what you saw was a trap that the host shim surfaced as an apparent hang, these cfgs mask it rather than fix it. If it genuinely is dtor registration, storing a non-`Drop` payload would fix it on all targets with no cfg at all.

## Minor

- **`--enable-simd` is unexplained** and unrelated to the comment above it. `wasm32-wasip1` doesn't emit `simd128` without `-C target-feature=+simd128`, so it's a no-op at best. Worth dropping, or adding a sentence — the surrounding code is unusually good about explaining every flag.
- **The `wasm-opt` branch never runs in CI.** The `wasm` job installs no binaryen, so `command -v wasm-opt` fails and `build.sh` falls through to the `cp`. The validation failure this fixes reproduces only locally. An `apt-get install binaryen` step in that job would make the optimized artifact actually exercised.
- **Asymmetry:** `pounce-observability` compiles out its `CAPTURE` thread_local, but `intermediate.rs:32-34` leaves `CURRENT_CTX` declared with all uses gated out (no warning today, but gating both the same way would read better). Same for `IterCaptureGuard::prev`, now dead on wasm.
- **No CHANGELOG entry.** The repo keeps a detailed `## [Unreleased]` section and every recent fix has one.
- The `#[cfg]`-on-tail-expression pattern in `with_current`/`is_active` is valid but subtle; two `#[cfg]`-selected function bodies would read more plainly. Style call.

## Security / performance

No security surface. Native codegen is untouched — every hunk is behind a cfg. Serial FERAL on wasm costs nothing in practice (no threads there anyway) and, per feral's own parity test, doesn't change results.

---

Net: the feral-serial change (relocated into `feral_config_from_options`) plus the build flag look ready to land. Before removing observability and inspector support on wasm, I'd want the reproducer test in CI and a bisect confirming the two TLS changes are actually needed.

---

GermanHeim:
Feel free to review @jkitchin. No. 1 wasn't an actual issue, that's why there is no commit with it.

jkitchin:
I haven't been able to reproduce the hang so far. Can you provide some more detail on the browser + version, rustc version, and whether the hang reproduces with pounce_builder_regression alone or only inside oximo?

GermanHeim:
Sure.

- OS: Fedora 44, Linux 7.1.5, KDE w/ Niri
- Browser: Chromium 150.0.7871.186 (Helium)
- Node: v22.23.1
- Binaryen wasm-opt 126
- Toolchain: rustc 1.99.0-nightly (dc3f85158, 2026-07-26) or rustc 1.99.0-nightly (73dc9167f, 2026-08-01) (I've been switching because of an enzyme/std::autodiff issue that I am trying to diagnose.

I have added a repo where I was facing the issue, in case you want to try it. It uses nightly-2026-08-02 + zola:
https://github.com/GermanHeim/oximo-pounce-site

It patches locally to `pounce-rs = { path = "../pounce/crates/pounce-rs" }` so you might need to change that

jkitchin (commented):
Follow-up on the five new commits. Thanks for turning these around so fast — `fc47dde` and the new test are exactly what I asked for. But I went looking for the hang before signing off, and what I found reframes most of the PR.

## I could not reproduce the hang on wasm32-wasip1, in any configuration

All against **unfixed `main`**, built with `crates/pounce-wasm/build.sh`:

| Setup | Result |
|---|---|
| `pounce_builder_regression` (this PR's own fixture), Node WASI | solves — and captures **30 iteration records** |
| `two_inequalities.nl` (this PR's own fixture) | solves |
| `init_subscriber()` + default `print_level` — full banner, matching your transcript — Node WASI | solves, ~50 ms |
| Same, under this repo's browser shim `web/wasi.js` | solves, ~70 ms |

The subscriber was the one difference I could see between my earlier run and your transcript: that banner means a global tracing subscriber was installed, and `pounce-wasm` never installs one. Adding it changes nothing.

Concretely, this PR's new test suite passes on `main` with no fixes applied. The only failing assertion is `captured_iterations === 0` — which on `main` is 30.

## The mechanism is real, and it is rayon in feral

`feral-0.14.0/src/numeric/solver.rs:972`, inside `factor()`:

```rust
let pool = if self.use_parallel { self.ensure_parallel_pool() } else { None };
```

That runs on **every** factorization, independent of problem size and independent of the `min_parallel_flops` dispatch gate. `ensure_parallel_pool` (`solver.rs:461`) calls `rayon::current_num_threads()` — which initializes rayon's global registry — then `ThreadPoolBuilder::build()`, which spawns *n* workers and blocks until they check in. `solve_refined` (`solver.rs:1540`) installs on that same pool, so the solve path depends on it too.

On wasm32-wasip1 this degrades gracefully. Measured with a stock toolchain:

```
available_parallelism = Err(Unsupported: "getting the number of hardware threads is not supported ...")
spawn                 = Err(Unsupported: "operation not supported on this platform")
```

`spawn` fails, `build()` returns `Err`, `ensure_parallel_pool` returns `None`, the factor runs serial. My successful runs above took exactly that path — `cfg.parallel` is `None` on `main` and neither host provides `FERAL_PARALLEL`, so `use_parallel` was `true` and the pool really was attempted.

The case that blocks is a host where the spawn call itself succeeds but workers never get scheduled — a threads-enabled wasm target whose host has not wired up its worker pool. That stalls at the first KKT factorization: exactly where your transcript stops, after the dimensions header, before iteration 0.

## Which means the gate is currently aimed the wrong way

```
wasm32-wasip1          → target_os="wasi"     target_family="wasm"
wasm32-unknown-unknown → target_os="unknown"  target_family="wasm"
```

`fc47dde` gates on `target_os = "wasi"`. So the override fires on wasip1, where feral already degrades gracefully and the fix is unnecessary — and does **not** fire on wasm32-unknown-unknown, the target where it would be load-bearing. `pounce-rs` does compile for that target; I checked.

That narrowing came from my earlier review point 5, and I got it wrong: my reasoning was right for the code as documented — every comment claimed wasm32-wasip1 — but the documented target looks like it was never the failing one. `target_family = "wasm"` is the correct gate, and it also covers emscripten and wasm64.

Where it does fire, the fix is otherwise complete: `use_parallel = false` short-circuits both blocking sites, and `feral_config_from_options` is the single funnel every production caller uses (the only direct `FeralConfig::default()` constructions are in `#[cfg(test)]` modules).

## What that means for the rest

`446b8c5`, `0788004`, and the TLS half of `d101725` address a mechanism that does not exist. Both TLS slots are hit on *every* iteration on `main` — `iteration_event_wanted()` at `ipopt_alg.rs:1528`, `CtxGuard::install` at `ipopt_alg.rs:1338` — and the wasm CI job has been green throughout. Capture demonstrably works on wasip1: 30 records.

The `collector_scope()` no-op is therefore not a workaround, it is a regression — it removes working functionality — and the new test asserts the loss as intended behaviour, so anyone who later restores capture will see this test fail and read it as their bug.

I also checked whether dense linear algebra could be the culprit, since that was the other candidate: faer 0.24 and gemm 0.19 carry no rayon in this build (rayon's only reverse deps here are feral and pounce-algorithm), feral's iterative refinement is capped at 10 + 1 with early exit, and there are no unbounded `loop {}` or float-predicate `while` constructs in feral. Same kernels ran to completion in every run above.

## Suggested landing shape

- **Keep `fc47dde`**, regate to `target_family = "wasm"`, and rewrite the comment to name the real mechanism (`ensure_parallel_pool` blocking in `feral::Solver::factor`) rather than "can block while creating its pool".
- **Keep `51c3e08`'s fixtures**, but flip `captured_iterations === 0` to `> 0`.
- **Drop `446b8c5`, `0788004`, and the TLS half of `d101725`.**
- **Keep `e48a5fe`**, minus the unexplained `--enable-simd`.

## One question, and it decides the above

What target triple did the failing build use, and were `+simd128` or atomics enabled? Everything here turns on that. If it was wasm32-unknown-unknown via wasm-bindgen, the rayon story fits your transcript exactly and `target_family` is the fix. If it really was wasm32-wasip1, then I cannot reproduce your hang at all and we are both still looking for it — in which case a build that fails would be more use than any of this analysis.

Worth noting either way: no test in this repo can cover wasm32-unknown-unknown, so a `cargo check --target wasm32-unknown-unknown` step would at least pin compilation.

## Upstream

The durable fix is in feral, not here: `Solver::new` hardcodes `use_parallel: true` (`solver.rs:430`). Defaulting it from `available_parallelism()` would get every threadless host right with no target `cfg` in feral or downstream, and pounce would need nothing at all. I have a filed-issue draft written up for that but no access to `jkitchin/feral` from this session, so it is not filed yet.

---

GermanHeim:
Ok, I did some experiments with the oximo wasm (in the one with the repo above):

1. With only `cfg.parallel = Some(false)` removed, the oximo WASI smoke times out after the model dimensions and before iteration 0.
2. Restoring only the intermediate TLS slot also reproduces the same pre-iteration timeout with FERAL still serial. So in this host, both independently affect the hang.
3. With FERAL remaining serial and the intermediate slot back to the PR version also times out, but later, with ordinary observability TLS restored (and FERAL still serial), it reaches iteration 0 and then stalls.

With current PR, all three oximo models solve successfully. The target is `wasm32-wasip1`. If you run the repo I sent  you, but using the main branch, do you not get the hangs? I am confused on why it would happen only on my PC... I will try my laptop, just in case.

GermanHeim:
Tested in my laptop, running CachyOS:

1. Checked out origin/main (`53c8f19`).
2. Picked only `51c3e08`.
3. Built `pounce-wasm` for `wasm32-wasip1`.
4. Invoked only `pounce_builder_regression()` under Node WASI with timeout 30s.

It timed out (exit 124).

Then I invoked the same export from the current build:
{"captured_iterations":0,"iterations":29,"status":"SolveSucceeded","success":true}
Result: exit 0.

jkitchin (commented):
@GermanHeim — thank you for your patience through this, and for the bisection and the second-machine run. That last experiment is what cracked it. I reproduced your hang, and my previous review was wrong on its central claim.

## The variable was the toolchain

This box was on **stable rustc 1.94.1**. You are on **1.99.0-nightly**. Same source, same host, same target — different compiler, opposite result. `wasm32-wasip1`, Node 22 WASI, `main` + only `51c3e08`:

| Build | Toolchain | `pounce_builder_regression` | `.nl` path |
|---|---|---|---|
| main + test | stable 1.94.1 | solves, 44 ms, 29 iters | solves |
| main + test | nightly-2026-08-02 | **hangs** | **hangs** |
| PR head `0788004` | nightly-2026-08-02 | solves, 54 ms, 29 iters | **hangs** |

The PR-head row matches your laptop output exactly — `iterations: 29`, `captured_iterations: 0`. So your changes are load-bearing on your toolchain, your bisection was sound, and "these fix a mechanism that does not exist" was an artifact of my environment. Sorry for the runaround that caused.

The rayon/`ensure_parallel_pool` theory from my last review is also dead — see below, the hang is upstream of feral entirely.

## What the hang actually is

Instrumenting the call boundary: under nightly, `pounce_load` never returns. That path parses the `.nl` text and builds the TNLP — it touches no capture TLS, no intermediate context, no feral, no rayon.

It is a **CPU spin**, not a block: 20 s wall, 20.0 s user. A wasm module under Node WASI has nothing to block on, so this is an infinite loop — which rules out TLS destructor registration as the mechanism. Consistent with that, a minimal drop-bearing `const`-initialized `thread_local!` compiled with the same nightly runs fine.

So the shape is: this nightly produces a pounce-wasm that spins in at least two distinct places, and the PR's edits perturb codegen enough to avoid one of them. **`pounce_load` still hangs on PR head under nightly** — so the PR is symptom relief, not a root-cause fix, and there is a second spin it does not reach.

Two candidate causes, which I have not separated: a codegen regression in that nightly for wasm32, or latent UB in pounce that the newer compiler exploits differently. The second deserves weight — pounce-wasm has hand-rolled `unsafe` in its allocator/payload ABI, and `446b8c5` adds two more `unsafe` blocks. Miri over the workspace plus a careful audit of the wasm ABI code would separate them.

## I was also wrong about `--enable-simd`

I told you to drop it as "a no-op at best." I ran Binaryen 126 against a stable-built module:

```
2656 × "SIMD operations require SIMD [--enable-simd]"
 374 × "all used features should be allowed"   (i32.trunc_sat_f64_*)
```

Rust emits v128 into this module. Both your flags are required. And it is worse than that: with `main`'s current `build.sh` (`--enable-bulk-memory` only) `wasm-opt` fails validation, and `set -euo pipefail` means **anyone with binaryen installed cannot build pounce-wasm at all today**. CI misses it only because the `wasm` job installs no binaryen.

`e48a5fe` is a real bug fix on its own merits. I would like to land it immediately, separately from the rest of this — happy for you to split it out, or I can.

## One question, and then a decision

Does `main` + `51c3e08` pass for you on **stable**? If it does, this is confined to nightly, and the decision becomes a judgement call I would rather make with you than impose:

- **Option A** — land `e48a5fe` now, file a rustc issue with the `pounce_load` spin as the reproducer (it is a clean one: build pounce-wasm on nightly-2026-08-02, call `pounce_load`, watch it spin), and unblock your site by pinning stable until that resolves.
- **Option B** — land the target `cfg`s as well, accepting permanent workarounds in the tree for a compiler bug that never reaches stable users or CI.

I lean A, because B leaves us maintaining cfgs whose stated rationale we now know is not the real mechanism, and it still would not fix `pounce_load` for you. But you are the one blocked, so if A leaves your site broken for longer than you can wait, say so and we will find a middle path — a documented, clearly-labelled temporary workaround with a link to the rustc issue is a defensible thing to carry.

Either way, the two CI gaps this exposed are worth closing on their own: install binaryen in the `wasm` job so the optimized artifact is actually built, and consider a nightly leg so toolchain regressions surface here rather than in your browser.

---

GermanHeim:
I confirmed the split locally. `origin/main` plus only `51c3e08` succeeds under stable 1.97.1 (SolveSucceeded, 29 iterations), while the same built with `nightly-2026-08-02` times out.

I am fine keeping stable for my toolchain. We could keep `51c3e08` and `e48a5fe` now if you want to go Route A.
Keep in mind that filing an issue in `rustc` would require creating a MWE (possibly without AI because of the [new policy](https://blog.rust-lang.org/inside-rust/2026/08/05/rust-langrust-is-adopting-an-llm-policy/)). Given the complexity of this codebase (Port of C++ + AI), that might be difficult.

That said, I am not opposed to doing route B because it would only affect wasm targets which are probably not that many pounce users.

Honestly, since this is your codebase, I think that it is your call.
My only real gripe is that Route A would break things if I wanted to use `std::autodiff `(Enzyme) in oximo + pounce. However, I am not using it for my current demos anyway, since I need to compile the Enzyme component for WASM from scratch (and maybe? likely? to fail). For now, I am sticking to finite differences.

GermanHeim:
By the way, this part:
> "... file a rustc issue with the pounce_load spin as the reproducer (it is a clean one: build pounce-wasm on nightly-2026-08-02, call pounce_load, watch it spin)..."

Rust maintainers won't accept this as a reproducer. You would need a MWE without (probably) pounce.

jkitchin:
I am going to look at this again tonight I hope.

GermanHeim:
Hi @jkitchin, I am out of town this week for a conferece, so I don't think I'll be able to do many commits.

In my opinion:
- If you decide to go route A, I would advise to force push `51c3e08` and `e48a5fe`. You can edit this PR.
- If you decide to go route B, then I think merging as is is fine.

Let me know if I can help you with anything!

jkitchin:
I reproduced the hang locally and tracked it to a root cause. Short version: **this is an upstream rustc/std regression on `wasm32-wasip1`, it is nightly-only, and it is not in POUNCE.** Your instinct that this was "the TLS issue" was right — it's a single upstream bug, which is why it kept reappearing at a new TLS site each time you removed one.

## Root cause

Sampling profile of the spinning module — 99.9% of ticks, one stack:

```
pounce_load
 └ std::sys::thread_local::native::eager::Storage<RefCell<Option<SolveOutcome>>>::get_or_init_slow
    └ std::sys::thread_local::destructors::list::register
       └ std::sys::thread_local::key::racy::LazyKey::lazy_init
          └ __pthread_key_delete        ← spins here forever
```

The first touch of a `thread_local!` **whose value needs drop glue** makes std register a TLS destructor. On `wasm32-wasip1` that now goes through the pthread-key shim, and `LazyKey::lazy_init` calls `pthread_key_delete`, which never returns in wasi-libc on a threadless build.

That is why the hang appeared to have "various causes". Every `cfg(wasi)` you added removed one TLS *use*, so the first-touch simply moved to the next one — the inspector context, then the observability capture slot, then `LAST_SOLVE`. Same single bug each time. I hit the identical moving target while bisecting, which is what put me onto it.

## Minimal reproducer (no dependencies, ~12 lines)

Two ingredients are required: a **cdylib reactor** (`wasi.initialize()`, so `std::rt::init` never runs) and a **thread_local whose value has drop glue**.

`Cargo.toml`:
```toml
[package]
name = "tls-hang"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["cdylib"]
```

`src/lib.rs`:
```rust
use std::cell::RefCell;

// The value needs drop glue, so std must register a TLS destructor on the
// first access. Swap in `Cell<u32>` (no destructor) and the hang disappears.
thread_local! {
    static SLOT: RefCell<Option<String>> = RefCell::new(None);
}

#[unsafe(no_mangle)]
pub extern "C" fn touch() -> u32 {
    eprintln!("before first touch");
    SLOT.with(|s| *s.borrow_mut() = Some(String::from("hello")));
    eprintln!("after first touch");
    1
}
```

`run.mjs`:
```js
import { readFileSync } from 'node:fs';
import { WASI } from 'node:wasi';

const wasi = new WASI({ version: 'preview1', args: [], env: {}, stdio: 'inherit' });
const inst = await WebAssembly.instantiate(
  await WebAssembly.compile(readFileSync('target/wasm32-wasip1/release/tls_hang.wasm')),
  wasi.getImportObject(),
);
wasi.initialize(inst);            // reactor module: no _start
console.log('touch ->', inst.exports.touch());
```

Run it:
```console
$ cargo +1.97.1 build --release --target wasm32-wasip1 && node run.mjs
before first touch
after first touch
touch -> 1

$ cargo +nightly-2026-08-02 build --release --target wasm32-wasip1 && node run.mjs
before first touch
        ← hangs, 100% CPU, never returns
```

Verified on node v25.9.0, macOS arm64.

## Scope

| toolchain | result |
|---|---|
| 1.96.0 stable | ok |
| 1.97.1 stable | ok |
| 1.98.0-beta.7 (2026-08-09) | ok |
| 1.99.0-nightly (2026-07-27 .. 2026-08-09) | **hangs** |

Bisected to a one-day commit window: `dc3f85158` (2026-07-26) good → `09ee43b2d` (2026-07-27) bad. I have not identified the specific PR in that range yet.

Both ingredients are load-bearing:

- `Cell<u32>` instead of `RefCell<Option<String>>` (no destructor to register) → works on every toolchain tested.
- A `bin` with `_start` instead of a cdylib reactor → works on every toolchain tested.

Good news on timing: 1.98 beta is clean, so this has **not** reached beta. If it's filed now it can be fixed before it ships in 1.99 stable.

## If you want to see it in POUNCE itself

The profile above came from POUNCE on `main`, not from the MWE. To reproduce that view, the one gotcha is that `crates/pounce-wasm/build.sh` passes `strip="symbols"`, which erases the wasm name section and leaves you staring at `wasm-function[2691]`. Build without it:

```console
$ cargo +nightly-2026-08-02 build -p pounce-wasm --release --target wasm32-wasip1 \
    --config 'profile.release.debug=false'
$ node --prof <your-harness>.mjs target/wasm32-wasip1/release/pounce_wasm.wasm   # kill after ~30s
$ node --prof-process --preprocess isolate-*.log > prof.json
```

then resolve the `wasm-function[N]` indices against the module's name-section function names.

One thing worth flagging, because it cost me time: an `opt-level=0` build still hangs, and that is **not** evidence about our codegen. Without `-Z build-std` every profile links the same precompiled `libstd`, so a bug living in std is invariant to opt-level. I initially read that backwards.
## Already extracted: your `wasm-opt` commit

`e48a5feb` is an independent, real bug with nothing to do with the hang, so I have
cherry-picked it out to **#571**, with you as the commit author. Without those flags
`wasm-opt` fails validation, and since `build.sh` runs under `set -euo pipefail` that
is a hard failure — anyone with binaryen installed currently cannot build
`pounce-wasm` at all. Confirmed locally: `Fatal: error validating input` and exit 1
before, exit 0 and a working module after.

Two things happened to it there that you should know about, since it is your commit:

- **CI now installs binaryen** in the `wasm` job. It previously did not, so `build.sh`
  took its silent `cp` fallback and the wasm-opt path was never exercised — which is
  exactly how this bug reached `main` unnoticed.
- **The explicit feature list became `-all`.** Installing binaryen in CI immediately
  proved the enumerated list insufficient: the runner has binaryen 108 (Ubuntu noble,
  2022) against 131 on a current Homebrew, and 108 additionally rejected sign-extension
  ops (`i32.extend8_s`). Chasing the feature set per binaryen version is not winnable,
  so it is now `wasm-opt -Oz -all`. That only widens what wasm-opt will *accept*; the
  emitted feature set is still whatever rustc produced.

#571 is now merged, so that part of this PR is done and can come out of the branch.
Do shout if you disagree with the `-all` change, since it broadens your commit's
original intent — easy to revisit.

## What I think this means for the rest of this PR

1. **The `cfg(wasi)` / `cfg(target_arch = "wasm32")` workarounds can come out.** They
   each disable a real feature (the inspector context, the observability capture slot)
   to dodge a compiler bug that no stable release ever had, and they would be permanent
   scar tissue for a transient nightly regression.
2. **Pin CI to stable** for the wasm job, which unblocks you now.
3. `51c3e08d`'s regression test asserts `captured_iterations === 0`. That encodes the
   *symptom*. On a good toolchain iterations really are captured — the smoke test on
   #571 reports `10 iterations` — so `> 0` is the thing actually worth guarding.

## On filing it upstream

I am not going to file the rust-lang issue, but the MWE above is yours to use if you
would like to. One thing to flag first, so it does not cause you a problem: **that
reproducer was derived with the help of Claude**, as was this analysis. rust-lang has
policies covering AI-assisted contributions, and I do not want to hand you something
that trips over them without you knowing its provenance. If you do file it, please
check their current policy on disclosure, and I would suggest reproducing it
independently on your own machine first so you can vouch for it as your own report
rather than relaying someone else's. An independent reproduction on different hardware
and a different OS would make it a materially stronger bug report anyway.

jkitchin:
@GermanHeim I am not sure where this leaves this PR. Let me know if you can confirm the root cause I described above.

GermanHeim:
@jkitchin I suspect this may be the cause, though I can't say for sure. 
Feel free to close this PR if it isn't needed anymore. I can easily patch the crate locally if I move to nightly because of autodiff. 

Thanks for looking into it!

jkitchin:
I will go ahead and close it; I don't think there is anything we should do on the pounce side if the mechanism above is correct. If you find otherwise later, feel free to open a new one.

PR #490, opened 2026-08-06
presolve: eliminate variables determined by linear equality rows

### LINKED ISSUE #487: presolve: eliminate variables determined by linear equality rows
**The measurement.** A dynamic-optimization flowsheet (a solvent-extraction stage assembled for NMPC) declares 23,681 variables and 23,138 equality constraints. Pyomo's NL-v2 writer with its linear presolve on — the default for the `ipopt_v2` interface — hands the solver 10,284 variables and 9,741 rows. The same model through pyomo-pounce reaches the solver at full size: `sens.py` writes through `model.write()`, which does not eliminate (verified on a two-variable equality fixture: both columns written), and the presolve phases here drop rows but do not remove columns on this path.

**What the NL-v2 presolve does.** A work list over three shapes, iterated to a fixed point so chains propagate: variables fixed by equal bounds are substituted as constants; singleton linear equality rows fix their variable (`x := b`, with a bounds check that raises on trivial infeasibility); two-variable linear equality rows substitute one variable for the other (`x := a·y + b`) with no anchoring requirement, so free-free pairs — arc equalities, `Reference` aliases, unit-conversion links — aggregate away. The writer returns the recovery map: `NLWriterInfo.eliminated_vars`, an ordered list of `(VarData, expression)` pairs arranged so each entry evaluates from solver-returned variables or earlier entries.

**What pounce-presolve covers today.** Phases 1–5 tighten bounds against linear rows, drop redundant rows behind the reinstate-on-finalize contract, check LICQ, emit warm-start hints, and project metadata. Variable elimination exists in one form: the auxiliary-equality pipeline from #53 (incidence → matching → Dulmage-Mendelsohn → block-triangular → linear block solve) solves *determined* square linear blocks and clamps their variables (`x_l = x_u = value`), which the fixed-variable classification then removes from the algorithm. Three limits keep it from closing the gap above: it handles determined blocks, not free-pair aggregation, so an arc equality between two free interior variables stays; nonlinear blocks are deferred; and the wrapper is opt-in, with sensitivity solves declining it entirely. Gauss-elimination on the linear equality rows was listed in scope by #20's table and is the missing piece.

**Two candidate homes.**

*(a) pyomo-pounce writes through NL-v2 with the presolve on.* Cheapest to reach and matches the `ipopt_v2` interface exactly. The cost lands on the sensitivity session, which is the pyomo path's reason to exist: the `.col` space no longer contains the eliminated variables, so every positional map — the KKT row and column indexing, warm-start suffix placement, bound-activity reads, `wrt=` block resolution — needs a third variable space beside full-x and the factor's, plus value recovery through the elimination map at report time. Answering `covariance(m, wrt=[an eliminated variable])` would mean propagating through the substitution expressions. This is the class of silent cross-space indexing risk gh #450 was about, multiplied by one more space.

*(b) an aggregation phase in pounce-presolve.* One implementation serves every path — CLI, GAMS, the C interface, and Pyomo — behind the same reinstate-on-finalize contract the existing phases use, with #53's pipeline (incidence, matching, block machinery, reduction frames) already in place as scaffolding. Sensitivity solves keep declining the wrapper, so the natural-units machinery is untouched: sensitivity runs at full size, exactly as today. The costs: the elimination work is implemented rather than borrowed, and the sensitivity path gains no speedup from it.

**Recommendation: (b).** The elimination belongs where every frontend benefits and where the reinstatement contract already exists; option (a) buys the same reduction only for Pyomo solves at the price of threading a third variable space through machinery whose correctness currently rests on there being exactly two. If the sensitivity path later wants the reduction too, that is the moment to design the index mapping deliberately — as its own stage, with the elimination map a first-class object — rather than as a side effect of switching writers.

jkitchin:
Fixes #487

## Problem

`presolve_linear_eq_reduction` was a registered option with nothing behind it — `PresolveOptions` read the key and no code ever consulted it.

The gap it left is the one #487 measured: a solvent-extraction NMPC flowsheet declares 23,681 variables and 23,138 equality rows, and Pyomo's NL-v2 writer — with the linear presolve its `ipopt_v2` interface turns on by default — hands the solver 10,284 columns and 9,741 rows. POUNCE reached the solver at full size on every path.

Reproduced here on a synthetic alias-heavy model of the same shape (24,000 columns / 21,600 rows, release build):

| | columns to the solver | rows | wall time | objective |
|---|---|---|---|---|
| before | 24,000 | 21,600 | 1.63 s | 1.61e-09 |
| after | **2,400** | **0** | **0.30 s** | 8.06e-09 |

Both are converged points of a deliberately flat quartic; the objectives agree to the reported tolerance and the `.sol` body is 45,600 entries either way.

## Cause

Two of the three shapes NL-v2 recognises had no home in POUNCE.

The auxiliary-equality pipeline (#53) eliminates *determined* square linear blocks. A row linking two otherwise-free interior variables — an arc equality, a `Reference` alias, a unit-conversion link — determines neither of its columns, so it fell straight through. That free/free pair is most of the reduction on a flowsheet.

## Fix

Option (b) of #487's two candidate homes: an aggregation phase in `pounce-presolve`, so the CLI, GAMS, the C interface, and Pyomo all get one implementation — and the sensitivity session is not asked to thread a third variable space through index maps whose correctness currently rests on there being exactly two (the #450 hazard).

Three shapes, iterated to a fixed point so chains propagate: a variable fixed by equal bounds, a singleton row `a·x = b`, and a two-variable row `a₁·x + a₂·y = b` with **no anchoring requirement**. Rows reduced to `0 = 0` are dropped as redundant.

Two design choices worth recording, because both had a cheaper alternative that was tried and rejected:

**The wrapper stacks outside `PresolveTnlp`, not inside it.** Phases 0–5 are written against a fixed column space; threading a variable reduction through them means touching every one of ~15k lines of carefully-tested presolve. Stacking on top means the new layer is the only thing that changes `n`, and each layer undoes exactly its own reduction at `finalize_solution`.

**Duals are recovered by a reverse sweep, not a factorization.** Restricted to the eliminated columns, the dual system is `J_D[:,e]ᵀ λ_D = −(∇f_e + J_K[:,e]ᵀ λ_K)` — square, sparse, and not something anyone wants inside presolve. Taken one elimination at a time in reverse order it is triangular instead: every other row of the problem-as-it-stood-then survives that step, so stationarity at that step's column has exactly one unknown multiplier. That makes the recovery a linear sweep over the elimination forest. The pivot tie-break uses union-by-size so alias chains stay shallow, since postsolve walks ancestor chains per dropped-row nonzero.

Correcting my own first attempt: I had the recovery in the `∇f − Jᵀλ` convention that `reduction_frame.rs` documents. What `finalize_solution` actually delivers is Ipopt's `∇f + Jᵀλ − z_l + z_u = 0`, verified empirically against a bare solve. The shipped code uses the latter.

## Blast radius

**Off by default**, and a no-op in three further ways: the wrapper falls into a verbatim-forwarding passthrough when the option is off, when the inner TNLP declines to publish its Hessian structure, and when the plan finds nothing to remove. Every existing presolve test is unchanged and green.

Fails closed. A contradictory equality system, or a model whose every column is determined, abandons the plan entirely and hands the model over untouched — an elimination pass is the wrong place to be the first and only voice calling a model infeasible.

Three limits worth stating plainly. Each is now filed as a follow-up:

- **LP and convex-QP models never see this phase** (#494), because the CLI dispatches to `pounce-convex` before any presolve wrapper is built. That is pre-existing and applies to the whole presolve layer, not just this phase. It does *not* mean those models go unpresolved — `pounce-convex` has its own PaPILO-style reduction stack, on by default. The real gap is narrower: that stack has singleton-row and free-column-singleton reductions but no **doubleton aggregation**, which is the shape that carries a flowsheet. Its own module docs name it as a known gap.
- **The `.nl` path cannot reach the row-constant branch** (#492). `get_constraints_linearity` tests whether a row's expression segment is *identically* zero, not whether it is affine, so a row carrying a bare constant is tagged NonLinear and the pass declines it. Verified directly; the branch is live only on the general TNLP path. The same predicate makes such a model classify as NLP rather than LP, so folding the constant into the bound at parse fixes both.
- **An eliminated column whose bound is active reports that multiplier on its survivor** (#493), because the bound was transferred there during planning. Primal, objective, and full-space stationarity are unaffected; only that one dual's attribution differs from a no-presolve solve. Documented at the head of `linear_eq_elim.rs`. The absent `ipopt_zL_out` / `ipopt_zU_out` entry is a symptom of this, not a `.sol` writer defect — sparse suffixes are the AMPL convention and readers zero-initialize, so the omission is lossless.

## Tests

New tests exercise new code, so "fails on the parent commit" is not available in the usual sense — they would not compile there. The evidence that they bite is mutation testing instead: five defects injected one at a time into the shipped code, each reverted after.

| injected defect | caught by | signal |
|---|---|---|
| drop the Hessian symmetric-pair ×2 factor | derivative probe | 573 / 1200 |
| ignore the sign of α when transferring bounds | plan probe | 600 / 1200 |
| forget the row constant `c` in `g_r = Σaⱼxⱼ + c` | primal-fidelity check | 459 / 1200 |
| lift an eliminated column without its offset β | plan probe | 573 / 1200 |
| skip the dual recovery (report λ = 0) | integration test | 2 tests |

The row-constant defect escaped **every** probe on the first pass, which was the most useful result of the exercise: derivatives are blind to an affine offset, since `∇(αy + β) = α` either way. The probe now pushes a reduced point through `finalize_solution` and checks the consumed rows still hold at the lift.

Layers:

- 30 unit tests on the planner and the transforms.
- `presolve_phase6_linear_eq_elim.rs` — solves through the full stack and checks full-space KKT stationarity of the returned duals, plus exact agreement with the bare solve. Also covers `wrap_with_presolve`, the composition point every library caller, the C interface, and the GAMS link go through.
- `linear_eq_reduction.rs` + a new `.nl` fixture — the `.sol` blocks come back at the original length and order with the consumed row's multiplier recovered.
- `adversary/fuzz` `linelim` probe — 12,000 randomized instances across three seeds, generated around a known feasible point the pass never sees. 0 failures.
- `adversary/runs/2026-08-06_nlp_linear_eq_reduction_verify.py` — 120 randomized `.nl` models through the CLI, checked by `pounce verify` (independent KKT/feasibility against the original file) and by in-script re-evaluation of every row. 0 failures.

Everything re-run after rebasing onto #489, which changed `.nl` derivative evaluation. Full workspace suite, `cargo fmt --check`, the CI clippy gate, and `check-release-consistency.sh` all clean.

Deliberately not pinned: the 23,681-variable flowsheet itself is not in the repo, so the headline measurement is reproduced on a synthetic model of the same shape rather than the original.

---

- [x] Tests fail on the parent commit for the stated reason — see the mutation table above; the tests target new code that does not exist on the parent
- [x] Book page under `docs/src/` updated (`options.md`, new Phase 6 section; no new page, so `SUMMARY.md` unchanged)

https://claude.ai/code/session_016JAvpSWoMKnBjNEoviZgVW

---
_Generated by [Claude Code](https://claude.ai/code/session_016JAvpSWoMKnBjNEoviZgVW)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,44 @@ changes.
 
 ## [Unreleased]
 
+### Added — presolve eliminates variables determined by linear equality rows (#487)
+
+- `presolve_linear_eq_reduction` was a registered option with no
+  implementation behind it. It now runs a **variable-elimination phase**
+  (Phase 6) in `pounce-presolve`, the first pass in the layer that removes
+  *columns* rather than only rows. Off by default; requires `presolve=yes`.
+- The measurement in #487: a solvent-extraction NMPC flowsheet declares
+  23,681 variables and 23,138 equality rows, and Pyomo's NL-v2 writer — with
+  the linear presolve its `ipopt_v2` interface turns on by default — hands
+  the solver 10,284 columns and 9,741 rows. POUNCE reached the solver at
+  full size on every path. It now recognises the same three shapes, iterated
+  to a fixed point so chains propagate: variables fixed by equal bounds,
+  singleton rows `a·x = b`, and two-variable rows `a₁·x + a₂·y = b`.
+- The two-variable shape is the one that mattered and the one nothing in
+  POUNCE could reach. The auxiliary-equality pipeline (#53) eliminates
+  *determined* square blocks; a row linking two otherwise-free interior
+  variables — an arc equality, a `Reference` alias, a unit-conversion link —
+  determines neither of them and so survived. Phase 6 substitutes one for
+  the other with no anchoring requirement.
+- Implemented in `pounce-presolve` (option (b) of #487's two candidate
+  homes) rather than by switching the Pyomo path onto NL-v2's presolve, so
+  the CLI, GAMS, the C interface, and Pyomo all get the same reduction from
+  one implementation — and so the sensitivity session is not asked to
+  thread a third variable space through index maps whose correctness rests
+  on there being exactly two (the #450 hazard).
+- Duals for the consumed rows are **recovered, not zeroed**. Taken as a
+  whole the recovery is a square sparse solve; taken one elimination at a
+  time in reverse order it is triangular, because every other row of the
+  problem-as-it-stood-then survives that step. So it is a linear sweep over
+  the elimination forest, not a factorization. `.sol` and JSON solution
+  blocks come back at the original model's length, in the original order,
+  and can still be read positionally by AMPL and Pyomo.
+- Two documented caveats, both in `docs/src/options.md`: an eliminated
+  variable's *bound* multiplier is reported on the survivor that inherited
+  the bound (the same attribution trade the Phase-2 row drop makes), and a
+  contradictory equality system makes the pass stand down entirely rather
+  than be the first and only voice calling a model infeasible.
+
 ### Fixed — the C working-set API validated nothing structural, and reported the wrong row order (#484 follow-up, round 4)
 
 - `IpoptSetWarmStartWorkingSet` range-checked its status codes and stopped
--- a/adversary/fuzz/Cargo.lock
+++ b/adversary/fuzz/Cargo.lock
@@ -10,6 +10,8 @@ dependencies = [
  "pounce-common",
  "pounce-feral",
  "pounce-linalg",
+ "pounce-nlp",
+ "pounce-presolve",
  "pounce-qp",
 ]
 
--- a/adversary/fuzz/Cargo.toml
+++ b/adversary/fuzz/Cargo.toml
@@ -15,6 +15,8 @@ pounce-feral = { path = "../../crates/pounce-feral" }
 pounce-common = { path = "../../crates/pounce-common" }
 pounce-linalg = { path = "../../crates/pounce-linalg" }
 pounce-cinterface = { path = "../../crates/pounce-cinterface" }
+pounce-presolve = { path = "../../crates/pounce-presolve" }
+pounce-nlp = { path = "../../crates/pounce-nlp" }
 
 [profile.release]
 debug = true
--- a/adversary/fuzz/src/linelim_probe.rs
+++ b/adversary/fuzz/src/linelim_probe.rs
@@ -1,0 +1,920 @@
+//! Adversary probe for Phase 6 — linear-equality variable elimination
+//! (`presolve_linear_eq_reduction`, gh#487).
+//!
+//! Two independent attacks, because the pass has two independent ways to be
+//! silently wrong.
+//!
+//! # 1. The plan (`plan` mode)
+//!
+//! Every instance is generated **around a known feasible point** `x*`: the
+//! right-hand sides are computed from `x*`, and every box is drawn to contain
+//! it. That makes `x*` an oracle the pass has never seen, and it pins five
+//! properties that a wrong substitution cannot fake:
+//!
+//! * **Fidelity** — projecting `x*` onto the survivors and lifting must
+//!   reproduce `x*` exactly. `x*` satisfies every row the pass consumed, so
+//!   any substitution derived from those rows must reproduce it. A sign slip
+//!   or a mis-composed chain shows up here immediately.
+//! * **Feasible-point retention** — `x*`'s survivor coordinates must lie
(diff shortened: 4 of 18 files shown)


devin-griff:
@jkitchin this is excellent. Thank you!

Issue #336, opened 2026-07-23
[Adversary] SOLVER_LIMITATION: HSDE exp/power driver reports numerical_failure on correct answers under extreme scaling (unnormalized complementarity gap)


jkitchin:
> Discovered by the pounce adversary agent (automated solver correctness testing).

## Family / solver
`exp` and `power` — `pounce.solve_socp` non-symmetric HSDE driver (shared by the
exponential and power cones). Version 0.9.0 @ b2aa052.

## Summary
Under extreme-but-legitimate data scaling, the HSDE conic driver reports
`status = numerical_failure` (`success = False`) at points that are
**primal-feasible, dual-feasible, and objective-correct**. The cause is that its
convergence/status criterion uses the **raw (unnormalized) complementarity gap
`s·z`** as `kkt_error`. When the optimal cone variables are large (e.g. a GP with
large coefficients, whose optimal cone values are ~1e6), the absolute
complementarity floor scales with them and can never reach the absolute
tolerance — so a relatively-converged solution is mislabeled a failure. A user
calling `solve_socp` gets `success=False` and discards a correct answer.
`scaling_warning` is `None` throughout.

This is the conic-driver analogue of the scale-normalization issues #324 (QP) and
#327 (NLP). Here the returned answer is *correct* (within standard tolerance), so
it is a `SOLVER_LIMITATION` (false-failure status + accuracy plateau), not a
wrong-answer bug.

## Case A — exponential cone (GP)
`min x + K/x`, `x>0`, with `K=1e12`. AM-GM optimum `x*=sqrt(K)=1e6`,
`f*=2 sqrt(K)=2e6`. Optimal cone variables `t1=t2 ~ 1e6`.

| solver | status | objective | rel err |
|---|---|---|---|
| pounce (exp) | `numerical_failure` | 1.99999025e6 | 4.87e-6 |
| cvxpy ECOS | optimal | 2.00000000e6 | 9.7e-10 |
| cvxpy SCS | inaccurate | 7.29e5 | 0.64 (SCS also fails; ECOS is the reliable oracle) |

Residuals at the returned point:
```
primal_infeasibility = 4.9e-6      (feasible)
dual_infeasibility   = 5.4e-8      (dual-feasible)
complementarity      = 3.84        (== reported kkt_error)
scaling_warning      = None
```
`3.84 / (1+|obj|)` = 1.9e-6 — i.e. the *normalized* gap is well inside tol.

[...]

PR #501, opened 2026-08-06
fix(convex/presolve): stop certifying infeasibility on a rounding residual

### LINKED ISSUE #496: [Adversary] SOLVER_BUG: false primal-infeasible certificate on rank-deficient equality rows inconsistent by one ULP
> Discovered by the pounce adversary agent (automated solver correctness testing), while probing gh#492. **Unrelated to that change** — reproduced on `origin/main` at `cad2b13`, and the reproducer contains no `C`-segment content at all.

## Family / solver

`lp` — `pounce-convex` LP interior-point, reached via the CLI on a `.nl` file (`solver_selection=auto` routes it there).

## Problem

Two equality rows on the same single column. The system is rank-deficient (rank 1, two rows) and **consistent** — but the two rows imply values of `x0` that differ by `6.94e-18`, one ULP at that magnitude.

```
min  -x0
s.t. -2.830268 x0 =  0.13596324445199998
      2.470924 x0 = -0.11870071803600002
     -10 <= x0 <= 10
```

```
x0 from row 0 = -0.048039
x0 from row 1 = -0.048039000000000005
difference    = -6.938893903907228e-18

row 0 residual at x0 = row1's value: 2.78e-17
row 1 residual at x0 = row0's value: 1.39e-17
```

Both residuals are at the level of floating-point rounding — around 1e-17 absolute, on rows whose coefficients are O(1).

## Expected vs actual

| | status | objective |
|---|---|---|
| pounce (`auto` → LP IPM) | **Problem is primal infeasible**, `iters=0` | 0 |
| scipy HiGHS (`linprog`) | optimal | `-14.060284628789999` on the 7-column instance this was reduced from |
| known feasible point | `x0 = -0.048039` satisfies both rows to ~1e-17 | — |

The certificate is emitted at **iteration 0**, before any IPM step, so this is a presolve/factorization-stage inconsistency test rather than an iterate-based conclusion. The `Constraint violation: 6.906...` printed alongside is the violation at the *initial* point, not evidence about feasibility.

## What makes it fire

The tolerance for declaring a rank-deficient equality system inconsistent appears to be effectively **zero** — it rejects at 1 ULP. Making the right-hand sides exactly consistent fixes it, and so does an exactly-duplicated row:

| variant | result |
|---|---|
| as generated (inconsistent by 6.9e-18) | **primal infeasible** |
| row 1 rhs replaced by `a1 * (b0/a0)` (exactly consistent) | Optimal |
| row 1 replaced by an exact duplicate of row 0 | Optimal |
| `2 x0 = 2` and `3 x0 = 3` (clean integers) | Optimal |
| both rows scaled by 1000 | **primal infeasible** |
| both right-hand sides zero | Optimal |

So it is not the redundancy itself, and not the sign pattern — it is that the redundancy is inexact at the last bit.

## Why it matters

Redundant equality rows are ordinary in real models — a duplicated balance, an alias plus its defining equation, a unit-conversion link stated twice. Their right-hand sides are usually *computed* rather than typed, so exact bit-agreement is the exception. A false infeasibility certificate is the worst failure mode available here: it is confident, it is fast, and it is wrong.

## Cross-check

- `scipy.optimize.linprog` (HiGHS) solves the model this was reduced from and reports `fun = -14.060284628789999`.
- A known feasible point exists by construction (the instance was generated around it) and satisfies every row to ~1e-17.
- Reproduced identically on `origin/main` at `cad2b13` and on the gh#492 branch, so it is not a regression from that change.
- The reduction from the original 7-column, 6-row instance to these two rows was done by exhaustive subset search; rows 1 and 5 alone are enough.

## Reproduce

Save as `m.nl` and run `pounce m -AMPL`:

```
g3 1 1 0	# two rank-deficient, consistent equality rows on one column
 1 2 1 0 2 	# vars, constraints, objectives, ranges, eqns
 0 0 0 0 0 0	# nonlinear constrs, objs
 0 0	# network
 0 0 0 	# nonlinear vars in constraints, objectives, both
 0 0 0 1	# linear network vars; functions; arith, flags
 0 0 0 0 0 	# discrete
 2 1 	# nonzeros in Jacobian, obj gradient
 0 0	# max name lengths
 0 0 0 0 0	# common exprs
r
4 0.13596324445199998
4 -0.11870071803600002
b
0 -10 10
k0
J0 1
0 -2.830268
J1 1
0 2.470924
G0 1
0 -1
```

Expected: `Optimal Solution Found`, `obj = 0.048039`.
Actual: `POUNCE (LP IPM, pounce-convex): Problem is primal infeasible. iters=0`.

The randomized family that surfaced it is `adversary/runs/2026-08-06_nlp_row_constant_fold.py` (seed `987654321`, trial 101); it hit roughly 1 instance in 250.

jkitchin:
Fixes #496

## Problem

Two equality rows on one column, redundant and consistent, but with right-hand sides that imply values of `x0` differing by `6.9e-18` — one ULP. POUNCE certifies the model primal infeasible at iteration 0.

```
min  -x0
s.t. -2.830268 x0 =  0.13596324445199998
      2.470924 x0 = -0.11870071803600002
     -10 <= x0 <= 10
```

| | status | objective | constraint violation | iters |
|---|---|---|---|---|
| before | Problem is primal infeasible | 0 | 1.36e-01 (at the *initial* point) | 0 |
| after | Optimal Solution Found | 0.04803900 | 1.39e-17 | 0 |
| oracle (scipy HiGHS `linprog`) | optimal | `-14.060284628789999` on the 7-column instance this was reduced from | — | — |

A known feasible point exists by construction and satisfies both rows to ~1e-17.

## Cause

`crates/pounce-convex/src/presolve.rs`. Presolve fixes `x0` from the first singleton equality row and substitutes it into the second, which empties to `0 = rhs`. That feasibility test was written against exact zero:

```rust
if row.rhs.abs() > 0.0 { return Err(()); }   // -> PresolveOutcome::Infeasible
```

But `rhs` at that point is `b - Σ aⱼ vⱼ`, a *computed* difference, so it carries the subtraction's rounding error. The residual here is 1.4e-17 and reads as a contradiction.

That also accounts for the full variant table in the issue, which the redundancy-based explanations do not: exact duplicates, clean integers, and both-rhs-zero all leave a residual of exactly 0, so they never fired; scaling both rows by 1000 scales the residual by 1000, which a fixed zero threshold cannot survive.

## Fix

Each row tracks `Row::scale` — the largest term that went into its `rhs` (the original `|b|` and every substituted `|aⱼvⱼ|`). That is the scale the rounding error lives on, so the emptied-row check compares against `EMPTY_ROW_TOL * (1 + scale)` with `EMPTY_ROW_TOL = 1e-9`, matching `ACTIVITY_TOL` — the same feasibility question asked of a row that kept its coefficients.

Scaling the cancellation scale rather than using a flat absolute tolerance is what makes the verdict invariant under multiplying the rows through by a constant; a flat threshold would just move the issue's reproducer to a different magnitude, which is exactly the failure the `x1000` row of the issue's table demonstrates.

Two deliberate limits on the blast radius:

- **Empty rows present in the *input* keep their exact check** (`prob.b[row] != 0.0`). No substitution happened, so there is no rounding to forgive, and `0 = 1e-18` written literally by a model is a statement about the model, not an artifact. Those rows are also the only ones that reach that check — `build_rows` drops emptied rows, so a fixpoint pass never feeds one back in.
- **The `0 <= rhs` inequality branch gets the same treatment**, for consistency rather than to fix a live defect — see Tests.

## Blast radius

Behaviour changes only for a row that (a) empties after substitution and (b) has a residual in `(0, 1e-9*(1+scale)]`. Every other input is bit-identical: the tolerance is consulted on exactly one branch, one that previously compared to zero.

Real conflicts are unaffected — injected disagreements of 1e-6, 1e-3, and 1.0 are still certified infeasible at presolve time, against a 1e-9 relative bar.

I did **not** run the benchmark corpus sweep. The models it contains do not exercise the changed branch except by the coincidence this PR is about, and a sweep would report no diff without that being evidence of anything. The evidence offered instead is the full workspace suite plus the 2000-instance randomized sweep below.

## Tests

New: `crates/pounce-convex/tests/issue496_ulp_inconsistent_equalities.rs`, six tests.

- the issue's exact instance, asserting first that the 1-ULP gap is really there (so the test cannot silently degrade into an exactly-consistent case);
- the same instance at scales 1e-3 / 1 / 1e3 / 1e6, which is what pins the residual tolerance to the cancellation scale rather than to a constant;
- the exactly-consistent variants (bit-exact rhs, duplicated row, clean integers, both rhs zero), guarding against a fix that trades one direction for the other;
- genuine conflicts at gaps 1e-6 / 1e-3 / 1 / 100, still infeasible;
- a 2000-instance deterministic sweep (in-file splitmix64) reproducing the generator that surfaced this — instances built around a random feasible point so both right-hand sides are *computed* rather than typed. This is the shape the adversary run hit at roughly 1 in 250.

**Three of the six fail on the parent commit** (`cad2b13`), for the stated reason — a `PresolveOutcome::Infeasible` on a feasible instance. Verified by stashing only `presolve.rs` and re-running.

One gap, stated so it is a known one: `ulp_violated_emptied_inequality_is_feasible` passes with *and* without the fix, so it is not a regression test. The activity-bound pass sees that row while `x0` is already fixed and has always used `ACTIVITY_TOL`, so it drops the row before `build_rows` looks. The `0 <= rhs` branch is only reachable via fixings made *after* the activity pass (forcing rows, dominated columns, bound tightening), which I did not find a compact instance for. The test is kept as a guard against that divergence reopening if the pass order changes, and its doc comment says exactly this rather than letting it read as a reproducer.

---

- [ ] ~Book page under `docs/src/`~ — n/a; the tolerance is internal to presolve and not a documented option
- [x] `cargo fmt --all -- --check`, CI's clippy gate (`-D clippy::correctness -D clippy::suspicious`), and `cargo test --workspace --exclude pounce-hsl` all clean. `pounce-hsl` excluded as CI does — it FFI-links the licensed CoinHSL, absent here.
- [x] Every claim in this body is true of the diff as it stands

---
_Generated by [Claude Code](https://claude.ai/code/session_01NRUVKiyXAfdSM9GzJXZ6hw)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,37 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — false primal-infeasible certificate on redundant equality rows that disagree by one ULP (#496)
+
+- The convex presolve fixes a variable from a singleton equality row and
+  substitutes it into the remaining rows. A row that empties in the process
+  becomes the feasibility test `0 = rhs`, and that test was written against
+  **exact zero**. But `rhs` at that point is `b − Σ aⱼ vⱼ`, a computed
+  difference — so it carries the subtraction's rounding error, and a second
+  equality that pins the same variable to the same value fails the test by
+  one ULP.
+- Found by the adversary agent while probing #492; unrelated to that change
+  and reproduced on `main`. Two rows on one column, `−2.830268 x₀ =
+  0.13596324445199998` and `2.470924 x₀ = −0.11870071803600002`, imply
+  values of `x₀` that differ by `6.9e-18`. POUNCE answered *primal
+  infeasible* at iteration 0; HiGHS solves the model it was reduced from.
+  Making the right-hand sides bit-exact, or duplicating a row verbatim, made
+  it optimal again — so the redundancy was never the trigger, only its
+  inexactness.
+- The residual is now compared against a tolerance scaled by the magnitude
+  of the terms that cancelled — the scale the rounding error actually lives
+  on — so the verdict does not change when the same rows are multiplied
+  through by 1000. Real conflicts are unaffected: rows that disagree by
+  anything a solve could act on are still certified infeasible at presolve
+  time, and the empty rows present in the *input* (no substitution, no
+  rounding) keep their exact check. The emptied-inequality test `0 ≤ rhs`
+  gets the same treatment.
+- This is the worst failure mode available to a presolver: a redundant
+  balance, an alias plus its defining equation, or a unit conversion stated
+  twice is ordinary in real models, and its right-hand sides are usually
+  computed rather than typed. Regression coverage in
+  `crates/pounce-convex/tests/issue496_ulp_inconsistent_equalities.rs`.
+
 ### Fixed — `solve_qp(method="active-set")` panicked across the FFI on a reversed box (#491)
 
 - `pounce.solve_qp(..., method="active-set")` with a crossed bound
--- a/adversary/log.org
+++ b/adversary/log.org
@@ -10,7 +10,7 @@ Counts update as runs land. The agent picks the LEAST-tested family each run.
 
 | Family          | Tested | Last run   | Open findings |
 |-----------------+--------+------------+---------------|
-| nlp             |     14 | 2026-08-06 | #348 (sqp_hessian=exact Internal_Error); #349 (SQP filter no-SOC Maratos/HS6); #495 (Phase 6 reports no bound multiplier for an eliminated column whose bound is active — pre-existing, off-by-default option); #496 (LP IPM false primal-infeasible on rank-deficient equalities inconsistent by 1 ULP) |
+| nlp             |     14 | 2026-08-06 | #348 (sqp_hessian=exact Internal_Error); #349 (SQP filter no-SOC Maratos/HS6); #495 (Phase 6 reports no bound multiplier for an eliminated column whose bound is active — pre-existing, off-by-default option); #496 FIXED 2026-08-06 (LP IPM false primal-infeasible on rank-deficient equalities inconsistent by 1 ULP; the row-constant-fold probe's baseline-skip count goes 1 -> 0 at seed 987654321) |
 | lp              |     11 | 2026-08-06 | —             |
 | qp              |     11 | 2026-08-06 | —             |
 | qp-active-set   |     14 | 2026-08-05 | both probes now CLEAN. Fixed this run: false-Infeasible (14 -> 0), `Optimal` at violating points, rank-deficiency hard errors, unvalidated working-set status codes, and IpoptGet/SetWorkingSet reporting the SQP's internal row order instead of the caller's. Nothing open. |
--- a/crates/pounce-convex/src/presolve.rs
+++ b/crates/pounce-convex/src/presolve.rs
@@ -16,7 +16,10 @@
 //! Reductions implemented:
 //! - **Empty rows** (equality / inequality with no nonzeros): a
 //!   feasibility check, then drop. Their dual is zero. Detects trivial
-//!   primal infeasibility (`0 = b≠0` or `0 ≤ h<0`).
+//!   primal infeasibility (`0 = b≠0` or `0 ≤ h<0`). A row emptied by
+//!   *substitution* is judged against the rounding error of the terms
+//!   that cancelled, not against exact zero — otherwise two equalities
+//!   that agree to all but the last bit read as contradictory (gh#496).
 //! - **Fixed-variable elimination** from a singleton equality row
 //!   (`a·x_k = b ⇒ x_k = b/a`): substitute `x_k` out of `P`, `c`, `A`,
 //!   `G` (adjusting the objective constant and the row right-hand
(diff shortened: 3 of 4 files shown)


jkitchin:
Status: **CI green** (12/12 on `3ea4904`), no review comments. Resolved a merge conflict and added one piece of independent confirmation.

### Merge conflict — resolved

`main` moved to `c1205c4` (#498 / gh#492) while this was open. Only `CHANGELOG.md` conflicted, both sides having added an `[Unreleased]` entry. Kept both, newest first. Merge commit `2efef42`; full suite re-run green on the merged tree, including #498's new `issue_492_row_constant_fold.rs`.

### The probe that filed this now runs clean

`main` also brought in `adversary/runs/2026-08-06_nlp_row_constant_fold.py`, the probe that discovered gh#496. It carries a `skipped, baseline unsolved` counter that exists specifically because of this bug. Running it at the seed named in the issue:

| `presolve.rs` | instances checked | baseline skips |
|---|---|---|
| this branch | 250 | **0** |
| reverted to `cad2b13` | 249 | **1** |

```
python3 adversary/runs/2026-08-06_nlp_row_constant_fold.py 250 987654321
```

Both directions confirmed against a release build, so this is the CLI + `.nl` reader path end to end — a different route to the defect than the in-crate tests, which construct `QpProblem` directly. `VERDICT: PASS`, with the sentinel and decline batteries also clean.

That also retires the caveat in my original Tests section about the randomized sweep being the only distributional evidence: the probe's counter now tracks the fix on its own.

### One file you may want to overrule me on

`26ba54c` marks gh#496 `FIXED` in the `adversary/log.org` findings index — one table cell, following the `#336 FIXED <date>` convention already in that column, since the "Open findings" list would otherwise be wrong the moment this merges.

I deliberately left the run report and the probe's own inline comment untouched. Both describe gh#496 as pre-existing and unfixed, which is an accurate record of what that run found at the time; rewriting them would be editing history rather than updating an index. If you'd rather the adversary agent own that index line too, drop `26ba54c` — nothing else depends on it.

---

jkitchin:
CI red on `08f0c97`, but not from this diff — GitHub Actions is failing to resolve action downloads.

Three jobs have failed so far, all identically, in GitHub's own `Prepare all required actions` step before checkout:

| job | run | region | failed |
|---|---|---|---|
| Python tests (torch) | 31116887785 | centralus | 15:43:06 |
| Python tests (torch) | 31116893754 | northcentralus | 15:44:17 |
| WebAssembly build + smoke | 31116893754 | westus | 15:44:32 |

```
Getting action download info
Failed to resolve action download info. Error: Service Unavailable
##[error]Service Unavailable
```

None of them checked out the repository or executed a line of project code.

The comparison that settles it: `WebAssembly build + smoke` in run `31116887785` **succeeded** at 15:42:24 — same commit, same workflow, same code — while the identical job in the sibling run failed at 15:44:32. The failures all begin after ~15:41:58, so the variable is wall-clock time against a GitHub-side incident, not anything in the commit.

No fix to push; a commit here would be a no-op whose only effect is cancelling the ten jobs still in flight. I'll re-run the failed jobs once the incident clears and the runs finish, and report back.

Local verification on this exact tree, rebased onto `a694ea3`: `cargo fmt --all -- --check` clean, the six `issue496_*` regression tests pass, and the workspace suite is 124 binaries in with zero failures.

---

Issue #494, opened 2026-08-06
convex presolve: add doubleton-equality aggregation to the reduction stack


jkitchin:
Follow-up to #487 / #490 (limit 1 of 3).

## What's wrong

#490 added linear-equality variable elimination as Phase 6 of `pounce-presolve`, which covers the general NLP path — CLI, GAMS, the C interface, Pyomo. LP and convex-QP models never reach it, because `crates/pounce-cli/src/main.rs` dispatches to `pounce-convex` several hundred lines before any presolve wrapper is built.

That is *not* the same as those models getting no presolve. `pounce-convex` has its own — `crates/pounce-convex/src/presolve.rs`, a PaPILO-style reversible transaction stack — and it is on by default (`resolve_convex_presolve`, `main.rs:2011`; `presolve=no` / `qp_presolve=no` turns it off).

The real gap is one missing reduction. That catalog has:

- fixed-variable elimination from a **singleton** equality row (`a·x_k = b`), and
- **free-column singleton** substitution (a free variable in exactly one row),

but no **doubleton-equality aggregation**: a row `a₁·x + a₂·y = b` linking two otherwise-free interior variables, with no anchoring requirement on either. That is the arc equality / `Reference` alias / unit-conversion link, and on a flowsheet it is most of the available reduction. Its own module docs already name it as a known gap (`presolve.rs:94`: "PaPILO is the catalog to mine for the next reductions — singleton / doubleton rows…").

So a genuinely *linear* flowsheet — the LP relaxation of the model #487 measured — gets the rest of the catalog but not the reduction that matters most for it.

## Fix

Add an `Aggregate` variant to the existing `Reduction` enum. The transaction stack is the right home; postsolve already replays in reverse, which is exactly the structure this reduction's dual recovery needs.

**Share the planner, not the wrapper.** `linear_eq_plan::PlanInput` is pure data — `n_vars`, per-row `(col, coeff)` lists, `row_const`, `g_l`/`g_u`, an `eligible` mask, and the box. No TNLP appears anywhere in it. So `pounce-convex` can depend on `pounce-presolve` for `build_plan` and write only its own postsolve, inheriting the fixed-point iteration over the three shapes, the union-by-size pivot rule that keeps alias chains shallow, and the fail-closed-on-contradiction behaviour — duplicating none of it.

Work then reduces to:

1. Feed the `QpProblem` equality rows into `build_plan`.
[...]

Issue #493, opened 2026-08-06
presolve Phase 6: attribute a transferred bound's multiplier to the column it came from


jkitchin:
Follow-up to #487 / #490 (limit 3 of 3). Closes the last documented caveat on the linear-equality elimination.

## What's wrong

`crates/pounce-presolve/src/linear_eq_elim.rs:60-72` records the assumption the dual recovery rests on: an eliminated column is interior to its own bounds at the optimum, so `z_l = z_u = 0` there.

When an eliminated column's bound *is* active, that bound was transferred onto its survivor during planning, so the IPM reports the multiplier on the **survivor's** bound instead. The primal point, the objective, and full-space stationarity are all unaffected — only the attribution of that one dual differs from a no-presolve solve.

### Not the `.sol` writer

An earlier framing of this (in #490's body) blamed `sol_writer.rs:203` dropping exact zeros from suffix blocks, so an eliminated column gets no `ipopt_zL_out` / `ipopt_zU_out` entry. That is a red herring and needs no change: sparse suffixes are the AMPL convention and readers zero-initialize, so the omission is lossless. The missing entry is a *symptom* of the multiplier having been attributed elsewhere.

## Fix

**Record bound provenance, then split at postsolve.**

`finish()` (`crates/pounce-presolve/src/linear_eq_plan.rs:671-682`) intersects each cluster's boxes down into `bounds.lo/hi` and discards which member supplied the winning bound:

```rust
for &j in &vars_kept {
    x_l_red.push(if bounds.lo[j].is_finite() { bounds.lo[j] } else { input.x_l[j] });
    ...
}
```

Add `x_l_src: Vec<usize>` / `x_u_src: Vec<usize>` to `EliminationPlan`, recording the originating full-space column per reduced bound — one extra write wherever the propagated box is tightened, so the cost is nil.

Then in `finalize_solution` (`linear_eq_elim.rs:933-942`), instead of unconditionally parking the multiplier on the survivor:

```rust
z_l_full[full] = sol.z_l[red];
z_u_full[full] = sol.z_u[red];
```

[...]

PR #504, opened 2026-08-06
presolve: recover the bound multiplier a collapsed reduced box hides

### LINKED ISSUE #495: [Adversary] SOLVER_BUG: Phase 6 produces no bound multiplier for an eliminated column whose bound is active (nlp)
> Discovered by the pounce adversary agent (automated solver correctness testing),
> while validating the gh#493 fix. Not caused by it: the residuals are identical
> with and without that change.

## Family / solver

`nlp` — CLI on `.nl`, `presolve_linear_eq_reduction=yes` (Phase 6, #487).

## What's wrong

When Phase 6 removes a column whose **declared bound is active at the optimum**,
no bound multiplier is produced for it, and the reported full-space duals are
not a KKT multiplier set. The primal, the objective, and the row multipliers are
fine; the bound duals are simply missing, and the stationarity residual they
should have carried is left uncancelled.

This is distinct from #493. #493 is a multiplier delivered to the *wrong column*
— it shows up as a complementarity violation, a non-zero multiplier on a slack
bound. This is a multiplier that is *never produced at all*, so re-attribution
cannot fix it: there is nothing to re-attribute.

Two routes in, both observed:

1. **A survivor whose reduced box collapses to a point.** Accumulated bound
   transfers can squeeze `[x_l_red, x_u_red]` down to a single value. The
   reduced column is then a *fixed* variable, which the solver drops from its
   internal problem, so it reports no bound multiplier for it — while the
   full-space cluster it stands for may have several simultaneously active
   bounds needing several multipliers.
2. **A column pinned to a constant whose value lands on one of its own declared
   bounds**, where the consumed row's multiplier does not absorb the whole
   residual.

Once one multiplier is missing, the row multipliers the reverse sweep builds on
top of it are off too, so the residual spreads across the cluster — in one
generated instance onto eight of nine columns.

## Expected vs actual

Minimal, hand-checkable:

```
min (x0-4)^4 + (x1-4)^4   s.t.   x0 - x1 = 0,   x0 in [1,5],   x1 in [-5,1]
```

The transfer squeezes x1's reduced box to the single point {1}. Optimum is
x = (1,1); by hand `grad = (-108,-108)`, and e.g. `lambda = 108`,
`z_l[0] = 0`, `z_u[1] = 216` is a valid KKT point.

| | x | lambda | z_l | z_u | KKT |
|---|---|---|---|---|---|
| `presolve=no` | (1, 1) | -108.14 | (0.1357, 0) | (0, 216.14) | OK |
| Phase 6 on | (1, 1) | -108.00 | (0, 0) | (0, 0) | **stationarity off by 216** |

The primal and the objective agree exactly. Only the bound duals are absent.

## Cross-check

The oracle is the textbook KKT system — primal feasibility, stationarity
`grad f + J' lambda - z_l + z_u = 0`, dual feasibility, and bound
complementarity — evaluated in Python from the same arrays used to write the
`.nl`, on the `ipopt_zL_out` / `ipopt_zU_out` suffixes that come back. The
same oracle certifies the `presolve=no` solve of every one of these instances,
which is what rules out a miscalibrated check.

`pounce verify` does **not** catch this, by construction: it reports
*bound-projected* stationarity (`verify.rs`, `stationarity_residual`), which
projects out exactly the component a bound multiplier would carry, and it never
reads the `zL`/`zU` suffixes. It returns 0 on every affected instance.

Scale: 80 randomized instances per seed, 9 seeds. 10-15 instances per seed hit
this, on both the pre-#493 and post-#493 builds, with identical residuals.

## Reproduce

```
cargo build --release -p pounce-cli
cd adversary/fuzz && cargo build --release --example plan_dump && cd ../..
python3 adversary/runs/2026-08-06_nlp_phase6_bound_attribution.py 80 11
```

The probe classifies this family separately from #493 and reports it under the
line `explained by a collapsed reduced box`. Its own verdict stays keyed to
#493, so this issue neither masks a #493 regression nor is mistaken for one.

## Note on scope

Phase 6 is off by default (`presolve_linear_eq_reduction=no`), so no default
user is affected. Anyone turning it on and reading `ipopt_zL_out` /
`ipopt_zU_out` — Pyomo's `model.ipopt_zL_out`, AMPL's `.rc` — can be.

jkitchin:
Fixes #495.

> **Rebased twice, onto #503 and then #499.** This started on the pre-#493 base. #503 landed mid-review, so the branch was rebuilt on top of it and the fix re-applied *onto* `attribute_bound_multiplier` rather than merged textually with it; #499 landed after that, and its interaction is in the comment below. The parent is `37631ec`, and every measurement here was re-taken against it.

## Problem

With `presolve_linear_eq_reduction=yes`, Phase 6 could return duals that are not a KKT multiplier set for the model as written. The primal, the objective and the row multipliers were right; a bound multiplier was simply absent, and stationarity was off by exactly its size. Anyone reading `ipopt_zL_out` / `ipopt_zU_out` — Pyomo's `model.ipopt_zL_out`, AMPL's `.rc` — saw zeros where a bound was carrying load.

The issue's minimal model, solved end to end through the CLI — `min (x0−4)⁴ + (x1−4)⁴` subject to `x0 − x1 = 0`, `x0 ∈ [1,5]`, `x1 ∈ [−5,1]`:

| | x | λ | `ipopt_zL_out` | `ipopt_zU_out` | stationarity |
|---|---|---|---|---|---|
| no presolve | (1, 1) | −108.57 | `0: 0.565` | `1: −216.57` | holds |
| before | (1, 1) | −108 | *(block empty)* | *(block empty)* | **off by 216** |
| after | (1, 1) | −108 | *(empty)* | `1: −216` | holds |

The reduced answer is exact where the bare solve carries interior-point slack; both are valid KKT points of the same problem.

On the probe filed with the issue — `adversary/runs/2026-08-06_nlp_phase6_bound_attribution.py`, which #503 brought onto `main`, 80 instances per seed:

| | seed 11 | seed 7 |
|---|---|---|
| `explained by a collapsed reduced box` (gh#495) before | 21 | 18 |
| after | **0** | **0** |
| `KKT violations (Phase 6 only)` (gh#493), before and after | 0 | 0 |
| `KKT violations (other phases, no P6)` — the probe's own scope check — before and after | 31 | 29 |

That last row is the useful one for blast radius: it does not move, so nothing outside Phase 6 changed.

## Cause

#493 fixed a multiplier being handed to the wrong column. This is the case where the solver produced none at all, so re-attribution has nothing to work with.

Phase 6 does not throw an eliminated column's box away, it *transfers* it onto the survivor. Accumulated transfers can squeeze `[x_l_red, x_u_red]` down to a single value — and a variable with equal bounds is a **fixed** variable, which `pounce_nlp::tnlp_adapter` drops from the problem as a parameter under `make_parameter`. It comes back with `z_l = z_u = 0` because the reduced problem never had it.

The dual sweep could not make that up. It closes stationarity at every column it consumed, but only because it is free to choose the row multiplier it is solving for; it has no such freedom over the bound multipliers of columns the reduced problem no longer has. In the model above the transfer collapses `x1`'s reduced box to `{1}`, and the 216 that `x1`'s upper bound should carry stayed in the residual.

## Fix

**Report the residual the sweep cannot close, and place it on the declared bound the point is resting on.**

`recover_dropped_multipliers` now returns the full-space residual, and zeroes the consumed rows' multipliers on entry so it can be re-run without compounding. That residual is zero at every column the sweep consumed, so what survives sits on a *surviving* column — and since the reduced problem's own stationarity holds at every column it still had, a non-zero residual there means the solver never enforced stationarity for that column at all. That identity is what makes the leftover interpretable rather than a heuristic: it *is* the missing multiplier.

`attribute_bound_residual` then places it, sign-aware (`∇f + Jᵀλ − z_l + z_u = 0`, so a positive residual is a lower-bound multiplier and a negative one an upper-bound multiplier):

* on the survivor's own bound where that is the active one;
* otherwise on the cluster member whose box supplied the collapsing side. `x_i = coeff·x_rep + offset`, so a multiplier of `resid / coeff` there moves exactly `resid` off the survivor — the same chain rule `attribute_bound_multiplier` applies to a multiplier that does exist, and a negative `coeff` therefore sends it to the opposite side of that column's box. Placing it changes what the consumed rows have to carry, so the sweep is redone with it in hand.

That the hand-off is a *single* column and needs exactly one re-sweep is not luck: `A` annihilates the dropped rows, so `Σ_cluster coeff_i · resid_i` is invariant under any choice of row multipliers, which fixes the size of the correction and guarantees the re-sweep lands on zero.

**It composes with #493 rather than overlapping it.** #503 already changed the sweep to start from `∇f + Jᵀλ − z_l + z_u`, which is exactly the entry point this needs — a multiplier placed here is absorbed by the same mechanism that absorbs a re-attributed one. The two are disjoint in what they can reach: one moves a reported multiplier, the other derives an unreported one. And #494 reached the same technique independently on the convex side; see the comment below.

Three things worth recording because the cheaper version was considered and rejected:

**Nothing is invented.** A residual with no active declared bound to carry it is left where it was. Parking it anyway would trade a stationarity error for a complementarity error, which is no improvement. A sub-tolerance residual (below `1e-9·‖∇f‖∞`) is float noise from the sweep and is likewise left alone, so an ordinary solve's duals do not acquire 1e-14 multipliers.

**Only surviving columns are examined.** An eliminated column with an elimination step already had its residual closed by that step's row multiplier; one without a step is a column the model declares fixed. So the first loop needs no special-casing to be safe — it structurally cannot disturb a multiplier the sweep depends on.

**A column the model declares fixed (`x_l == x_u`) is deliberately out of scope.** Those are dropped as parameters whether or not the reduction runs, and a no-presolve solve reports no multiplier for them either. Attributing one here would make Phase 6 *diverge* from the baseline rather than match it. The test is exact equality because that is the test that decides it — the adapter's `lo == hi` — so a column a hair wider reaches the solver and its multiplier is one the reduction really did lose. `a_declared_fixed_column_reports_exactly_what_the_bare_solve_reports` pins this boundary; if it should change, that belongs to `fixed_variable_treatment`, not here.

## Blast radius

Confined to `linear_eq_elim.rs` and to a code path that is off by default (`presolve_linear_eq_reduction=no`). Where nothing collapsed, the duals are unchanged: the new pass only fires on a residual the old sweep left behind, and `an_interior_reduction_reports_exactly_what_it_did_before` asserts that directly. The probe's out-of-scope counter (31 / 29) is identical before and after, and every one of #503's own unit tests passes untouched.

`recover_dropped_multipliers` gained a return value and now zeroes the consumed rows' multipliers on entry, which reaches the one external caller #499 added — `pounce_convex::aggregate::postsolve`. That caller already maintained the same invariant by hand, so it is unaffected; details in the comment below, and `cargo test -p pounce-convex` is green.

Full workspace suite (`cargo test --workspace --exclude pounce-hsl`), `cargo fmt --all -- --check`, the CI clippy gate, `check-docs-consistency.sh` and `check-release-consistency.sh` are all clean. The gh#487 probe (`2026-08-06_nlp_linear_eq_reduction_verify.py`, 80 instances) reports 0 failures.

## Tests

**Unit, in `linear_eq_elim.rs`** (7 new, alongside #503's 5): the residual lands on the survivor's active bound; a residual the survivor cannot carry moves to a cluster member; the hand-off rescales by the substitution coefficient and flips sides when it is negative; a residual with nowhere to go invents nothing; a declared-fixed column is not given a multiplier; a sub-tolerance residual is left alone; the sweep reports its residual and can be re-run without compounding.

**Through the solver stack**, `crates/pounce-algorithm/tests/presolve_phase6_missing_bound_multiplier.rs`. The fixture owns its own analytic derivatives, so the oracle is the textbook KKT system — feasibility, stationarity, dual feasibility, complementarity — not anything the solver says about itself. `pounce verify` cannot stand in: it reports *bound-projected* stationarity, which projects out exactly the component under test. Covers the issue's model, both sides of the box, a negative coefficient, a chain, the cross-cluster hand-off, and 60 randomized clusters built around a shared witness point so they are feasible by construction and a stand-down is a generator bug rather than something to skip past.

The randomized sweep asserts **all four** KKT conditions. On the pre-#503 base it had to leave complementarity out, because #493 broke it independently; with #503 in, it passes whole — which is a second, independent check that the two fixes compose.

Two of the eight assert that nothing moved — the interior reduction and the declared-fixed boundary — and both pass on the parent commit, which is the point of them.

**End to end**, `crates/pounce-cli/tests/linear_eq_collapsed_box_duals.rs`: the issue's `.nl`, through the CLI, read back out of the `ipopt_zL_out` / `ipopt_zU_out` suffix blocks. `presolve_bound_tightening=no` is not incidental — Phase 1 otherwise pins both columns from the same intersection and the model never reaches Phase 6.

Six of the eight stack tests and the CLI test fail on the parent commit with the missing multiplier.

**Six defects injected one at a time into the shipped code**, each reverted after:

| injected defect | caught by |
|---|---|
| no hand-off across the cluster | 2 unit tests, 2 stack tests |
| hand-off drops the `1/coeff` rescale | 1 unit test, the randomized sweep |
| sweep ignores the attributed multipliers | 2 unit tests (one of them #503's), 2 stack tests |
| no re-sweep after a hand-off | 2 stack tests |
| parks without checking the bound is active | 4 unit tests, 2 stack tests |
| ignores the residual tolerance | 1 unit test |

**Deliberately not pinned:** no new Python probe. #503's `2026-08-06_nlp_phase6_bound_attribution.py` already classifies this family under its own line, which is what the before/after table above measures; a second script would duplicate it.

---

- [x] Tests fail on the parent commit for the stated reason — 6 of 8 stack tests plus the CLI test, with the missing multiplier; plus the six injected defects above
- [x] `CHANGELOG.md` `[Unreleased]` entry, in the user's terms — a bullet in the #487 section, matching how #493's correction to the same unreleased feature was recorded
- [x] Book page under `docs/src/` updated (`options.md`, Phase 6's caveat list; no new page, so `SUMMARY.md` unchanged)
- [x] Every claim in this PR body and in the code comments is true of the code as it stands now — the body was rewritten after each rebase and the measurements re-taken on `1510159`

https://claude.ai/code/session_01SQgwy4qgX3NwNYHJJkR9iT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,6 +119,22 @@ changes.
   no-presolve solve. Where the survivor's own bound is active *as well*, the
   split is genuinely non-unique and the multiplier stays on the survivor —
   still a valid KKT point, and documented as such in `docs/src/options.md`.
+- A bound multiplier the reduced solve never *produced* is now recovered
+  too (#495). Re-attribution can only move a multiplier that exists, and
+  accumulated bound transfers can leave a survivor's reduced box as a
+  single point — a fixed variable, which the solver drops as a parameter,
+  so it reports `z_l = z_u = 0` even though the cluster it stands for is
+  sitting on a bound carrying load. Stationarity in the original space was
+  off by exactly the missing multiplier, and `ipopt_zL_out` /
+  `ipopt_zU_out` came back empty where Pyomo and AMPL expect a value.
+  Postsolve now reads the residual its row-multiplier sweep cannot close —
+  which is precisely that multiplier — and places it on the declared bound
+  the point is resting on, the survivor's own or, through the same `α`
+  rescale, the column the bound was borrowed from. A residual with no
+  active declared bound to carry it is left alone: trading a stationarity
+  error for a complementarity error is not a repair. A column the *model*
+  declares fixed is untouched, since the solver drops those with or without
+  the reduction and a no-presolve solve reports nothing for them either.
 - One documented caveat remains, in `docs/src/options.md`: a contradictory
   equality system makes the pass stand down entirely rather than be the
   first and only voice calling a model infeasible.
--- a/crates/pounce-algorithm/tests/presolve_phase6_missing_bound_multiplier.rs
+++ b/crates/pounce-algorithm/tests/presolve_phase6_missing_bound_multiplier.rs
@@ -1,0 +1,623 @@
+//! Phase 6 acceptance (issue #495): every column whose declared bound is
+//! active at the optimum must come back with a bound multiplier, so that
+//! what postsolve reports is a KKT multiplier set for the *original*
+//! problem.
+//!
+//! #493 moved a multiplier the solver *reported* to the column that owns
+//! the bound. This is the case where the solver reported none: accumulated
+//! bound transfers squeeze a survivor's reduced box down to a single point,
+//! the solver treats that column as a fixed variable and drops it, and no
+//! bound multiplier is produced at all — while the full-space cluster it
+//! stands for is sitting on a bound that needs one. There is nothing to
+//! re-attribute, so postsolve has to derive it.
+//!
+//! The oracle here is the textbook KKT system, evaluated from the fixture's
+//! own analytic derivatives: primal feasibility, stationarity
+//! `∇f + Jᵀλ − z_l + z_u = 0`, dual feasibility, and bound
+//! complementarity. `pounce verify` cannot stand in for it — it reports
+//! *bound-projected* stationarity, which projects out exactly the
+//! component a bound multiplier carries.
+
+use pounce_algorithm::application::IpoptApplication;
+use pounce_common::types::Number;
+use pounce_nlp::tnlp::{
+    BoundsInfo, IndexStyle, IpoptCq, IpoptData, Linearity, NlpInfo, Solution, SparsityRequest,
+    StartingPoint, TNLP,
+};
+use pounce_presolve::{LinearEqElimTnlp, PresolveOptions, PresolveTnlp};
+use std::cell::RefCell;
+use std::rc::Rc;
+
+/// `min Σ (x_j − t_j)^p` over linear equality rows and a box.
+///
+/// Separable and polynomial, so every derivative the KKT oracle needs is
+/// available in closed form on the test side rather than read back out of
+/// the solver.
+#[derive(Clone)]
+struct Model {
+    targets: Vec<Number>,
+    x_l: Vec<Number>,
+    x_u: Vec<Number>,
+    /// `Σ a_j x_j = b`, one entry per row.
+    rows: Vec<(Vec<(usize, Number)>, Number)>,
+    power: i32,
+}
+
+impl Model {
+    fn n(&self) -> usize {
+        self.targets.len()
+    }
+
+    fn grad(&self, x: &[Number]) -> Vec<Number> {
+        let p = self.power;
+        (0..self.n())
+            .map(|j| p as Number * (x[j] - self.targets[j]).powi(p - 1))
+            .collect()
+    }
+
+    /// `∇f + Jᵀλ − z_l + z_u`, in POUNCE's `finalize_solution` convention.
+    fn stationarity(&self, c: &Captured) -> Vec<Number> {
+        let mut r = self.grad(&c.x);
+        for (i, (entries, _)) in self.rows.iter().enumerate() {
+            for &(j, a) in entries {
+                r[j] += a * c.lambda[i];
+            }
+        }
+        for j in 0..self.n() {
+            r[j] += -c.z_l[j] + c.z_u[j];
+        }
+        r
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+struct Captured {
+    x: Vec<Number>,
+    z_l: Vec<Number>,
+    z_u: Vec<Number>,
+    lambda: Vec<Number>,
+}
+
+struct Fixture {
+    model: Model,
+    captured: Option<Captured>,
+}
+
+impl TNLP for Fixture {
+    fn get_nlp_info(&mut self) -> Option<NlpInfo> {
+        let nnz_jac: usize = self.model.rows.iter().map(|(e, _)| e.len()).sum();
+        Some(NlpInfo {
+            n: self.model.n() as i32,
+            m: self.model.rows.len() as i32,
+            nnz_jac_g: nnz_jac as i32,
+            nnz_h_lag: self.model.n() as i32,
+            index_style: IndexStyle::C,
+        })
+    }
+
+    fn get_bounds_info(&mut self, b: BoundsInfo<'_>) -> bool {
+        b.x_l.copy_from_slice(&self.model.x_l);
+        b.x_u.copy_from_slice(&self.model.x_u);
+        for (i, (_, rhs)) in self.model.rows.iter().enumerate() {
+            b.g_l[i] = *rhs;
+            b.g_u[i] = *rhs;
+        }
+        true
+    }
+
+    fn get_starting_point(&mut self, sp: StartingPoint<'_>) -> bool {
+        for (j, slot) in sp.x.iter_mut().enumerate() {
+            // Inside the box, but nowhere near the answer.
+            let lo = self.model.x_l[j].max(-1e3);
+            let hi = self.model.x_u[j].min(1e3);
+            *slot = 0.5 * (lo + hi);
+        }
+        true
+    }
+
+    fn get_constraints_linearity(&mut self, types: &mut [Linearity]) -> bool {
+        for t in types.iter_mut() {
+            *t = Linearity::Linear;
+        }
+        true
+    }
+
+    fn eval_f(&mut self, x: &[Number], _new_x: bool) -> Option<Number> {
+        Some(
+            (0..self.model.n())
+                .map(|j| (x[j] - self.model.targets[j]).powi(self.model.power))
+                .sum(),
+        )
+    }
+
+    fn eval_grad_f(&mut self, x: &[Number], _new_x: bool, g: &mut [Number]) -> bool {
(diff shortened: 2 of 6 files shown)


jkitchin:
Rebased again, onto #499 (`37631ec`). Two things in that merge touch this PR, both worth flagging rather than leaving to be discovered:

**`recover_dropped_multipliers` is now `pub`, and `pounce-convex` calls it.** This PR changes its signature — it returns the residual, and it zeroes the consumed rows' multipliers on entry. The convex caller is unaffected: `aggregate::postsolve` already hands it a `y` with the consumed rows zeroed, and already does `y = y_kept;` by hand before its second call. So the entry-zeroing is exactly the invariant that caller was maintaining manually; it now holds by construction for both. The returned `Vec` is ignored there, which is fine. `cargo test -p pounce-convex` is green, including `presolve_aggregation.rs`.

**#494 arrived at the same technique independently, on the convex side.** `aggregate.rs`'s module docs describe re-attributing "the leftover reduced cost at each survivor … to whichever cluster member is actually sitting on its own declared bound, and the sweep is re-run with that force in place" — which is what `attribute_bound_residual` does here, in the same order, with `LEFTOVER_TOL` / `on_bounds` playing the parts of `resid_tol` / `BOUND_ACTIVE_TOL`. That page already says the NLP path "reaches the same answer by a different route (gh #493, fixed by #503)" — after this PR that is only half true: for a multiplier the solver *reported*, the NLP path still uses recorded provenance; for one it never produced, both paths now read it off the leftover. Happy to reword that paragraph here if you'd rather it not go stale, but it seemed better to leave another module's docs alone unless asked.

Measurements re-taken on the rebased head, not carried over: the issue's probe at seed 11 still reports `explained by a collapsed reduced box: 0` (21 before), `KKT violations (Phase 6 only): 0`, and its out-of-scope counter unchanged at 31. Full workspace suite, fmt, clippy gate, and both consistency scripts clean.

---

PR #509, opened 2026-08-06
fix(status): ask "is this violation real" with constr_viol_tol, not tol

### LINKED ISSUE #508: Infeasible model returns solve_result_num=500 (internalSolverError): the LocalInfeasibility/ErrorInStepComputation threshold is built from tol, not constr_viol_tol
On an infeasible model, the choice between `Converged to a point of local
infeasibility` and `Error in step computation` is made by a threshold built
from `tol` — a tolerance on the **KKT error** — and it never consults
`constr_viol_tol`, the option that declares what a violated constraint is.

`crates/pounce-algorithm/src/ipopt_alg.rs:2517`:

```rust
let outer_tol_for_cycle = self.bundle.conv_check.tol_or_default();
let reference_theta_unscaled = self.cq.borrow().curr_unscaled_primal_infeasibility_max();
let cycle_exit = if reference_theta_unscaled > (100.0 * outer_tol_for_cycle).max(1e-4) {
    SolverReturn::LocalInfeasibility
} else {
    SolverReturn::ErrorInStepComputation
};
```

`ErrorInStepComputation` writes `solve_result_num = 500` into the `.sol`, which
Pyomo maps to `internalSolverError` — a status that tells the user their solver
broke, not that their model has no solution.

## Reproducer

```
min (x - 5)^2   s.t.   x^2 + delta == 0
```

Infeasible for every `delta > 0`. `½‖c‖²` has a strict local minimum at `x = 0`
where the violation is exactly `delta` and `Jᵀc = 2x(x²+delta) = 0`, so
restoration converges there, no local move reduces the violation, and the
honest verdict at every `delta` is "converged to a point of local
infeasibility". Because the reported violation *is* `delta`, the threshold is
measurable to the digit.

Probe (60 cells, deterministic across three runs):
`adversary/runs/2026-08-06_nlp_infeasible_gap_status.py`, report alongside it
in `.org`.

## What it measures

**1. `constr_viol_tol` does not move the boundary.** `tol` fixed at 1e-6,
`constr_viol_tol` swept four orders:

| constr_viol_tol | delta=1e-5 | delta=3e-5 | delta=1e-4 |
|---|---|---|---|
| 1e-8 | infeasible | FAILURE(500) | FAILURE(500) |
| 1e-6 | infeasible | FAILURE(500) | FAILURE(500) |
| 1e-5 | infeasible | FAILURE(500) | FAILURE(500) |
| 1e-3 | infeasible | FAILURE(500) | FAILURE(500) |

Identical. The bottom row is the sharpest: at `constr_viol_tol=1e-3` a
violation of `1e-4` is *inside* the user's declared feasibility tolerance, and
the answer is an internal error. This part needs no oracle — it is an internal
contradiction in POUNCE's own option semantics.

**2. Loosening `tol` makes it worse.** The failure band tracks
`max(100·tol, 1e-4)` exactly, so the standard user reaction to a struggling
solve widens it:

| tol | threshold | FAILURE(500) cells of 15 |
|---|---|---|
| 1e-8 | 1e-4 | 4  (delta = 1e-9, 1e-7, 1e-6, 1e-4) |
| 1e-6 | 1e-4 | 3  (delta = 1e-9, 3e-5, 1e-4) |
| 1e-5 | 1e-3 | 6  (delta = 1e-9, 3e-6, 3e-5, 1e-4, 3e-4, 1e-3) |
| 1e-4 | 1e-2 | 7  (delta = 1e-8, 3e-6, 1e-4, 3e-4, 1e-3, 3e-3, 1e-2) |

At `tol=1e-4`, five consecutive gaps from `3e-4` to `1e-2` come back as
internal errors — including a model infeasible by a full **1%**. The comparison
is `>`, so a violation landing exactly on the threshold takes the failure
branch (`delta=1e-4` at `tol=1e-6` and `tol=1e-8`).

The failing cells are not short-circuiting: within `delta ≤ 1e-2` the
`infeasible` verdicts take 22-48 iterations and the failures take 65-802.

**3. Against IPOPT 3.14.19**, same `.nl`, same options, classified off its own
`.sol` the same way: POUNCE returns 500+ on 20 of 60 cells, IPOPT on 7 (all
`501`, restoration failure). On **11 of 60** IPOPT returns a clean `200` —
locally infeasible, the interpretable and correct answer — where POUNCE reports
that it broke. IPOPT is not clean here and this is not the load-bearing
evidence; (1) and (2) are.

## Why the threshold is the wrong shape

`tol` and `constr_viol_tol` are different quantities in different units. A
threshold on the constraint violation built out of `tol`, with a hard-coded
`1e-4` floor and no reference to `constr_viol_tol`, cannot express "is this
violation real" — which is exactly the question the branch is asking. The
natural spelling is a multiple of `constr_viol_tol` (the same
`infeas_viol_kappa · constr_viol_tol` band the rapid-infeasibility detector
already uses), and `>=` rather than `>` at the boundary.

A second question worth settling in the same change: whether
`ErrorInStepComputation` is ever the right answer for a *stalled restoration on
a genuinely infeasible model*. The step computation did not error — restoration
converged to a local minimum of the violation. `LocalInfeasibility`, or
`StopAtAcceptablePoint` where one is stored, describes what happened; `500`
does not.

## Secondary: console banner disagrees with the `.sol`

Two cells print the MC64 re-solve's second `EXIT:` banner while the `.sol`
keeps the first solve's verdict, so the terminal and the modelling layer
disagree about the same solve:

| tol | delta | console | .sol |
|---|---|---|---|
| 1e-4 | 1e-9 | `Error in step computation.` | infeasible (200) |
| 1e-4 | 1e-1 | `Maximum Number of Iterations Exceeded.` | infeasible (200) |

Two banners is expected and announced; the two disagreeing is not. The second
row also ran 3000 iterations and ended with a reported violation of **1.49**
against `delta = 0.1` — an order of magnitude away from the `x=0` minimum every
other cell in that column reaches in ~40 iterations.

## Not a duplicate of #505

Same family, different defect, and fixing either one leaves the other standing.

|  | #505 | this |
|---|---|---|
| model | feasible; a usable point | infeasible by construction |
| statuses in play | `200` returned where `100` was right | `500` returned where `200` was right |
| decision site | the infeasibility exit firing ahead of the acceptable-level exit | the `LocalInfeasibility`/`ErrorInStepComputation` ternary at `ipopt_alg.rs:2517` |
| what the user is told | "your model may be infeasible" (it isn't) | "your solver broke" (it didn't) |

#505's suggested remedy — prefer the acceptable exit when the current iterate
already satisfies the acceptable-level criteria — does not touch this one. In
the sweep above `acceptable_tol=1e-12`, so no acceptable point exists to prefer;
the choice is still between `200` and `500`, still decided by a `tol`-derived
threshold. Conversely, giving `:2517` the right units does nothing for #505,
whose exit returns `200` already.

The shared ancestry is real but sits one level up: both are the status-decision
region choosing badly among exits, which is the family #372 and #446 belong to.

Control arm in the same probe: 7 feasible ill-conditioned models in gh#505's
residual regime, zero false infeasibility, and POUNCE reaches the optimum on
the three where IPOPT hits its iteration limit. The infeasibility machinery is
not firing indiscriminately — the failure is specific to which of two failure
statuses is chosen.

jkitchin:
Fixes #508

## Problem

On a model with no solution, POUNCE chooses between `LocalInfeasibility` (AMPL 200 — "your model has no solution") and `ErrorInStepComputation` / `RestorationFailure` (AMPL 500, Pyomo `internalSolverError` — "your solver broke"). That choice was made against a threshold built from `tol`, and never consulted `constr_viol_tol`.

Probe model, from the issue: `min (x-5)² s.t. x²+δ = 0`. Infeasible for every δ>0; `½‖c‖²` has a strict local minimum at `x=0` where the violation is exactly δ and `Jᵀc = 2x(x²+δ) = 0`, so restoration converges there, no local move reduces the violation, and the honest verdict at every δ is "converged to a point of local infeasibility". Because the reported violation *is* δ, the threshold is measurable to the digit.

`solve_result_num` at `acceptable_tol=1e-12`, **before** (parent commit `ec70fab`):

| `constr_viol_tol` (tol=1e-6) | 1e-6 | 3e-6 | 1e-5 | 3e-5 | **1e-4** | 3e-4 | 1e-3 | 1e-2 |
| 1e-8 | 200 | 200 | 200 | 200 | **500** | 200 | 200 | 200 |
| 1e-6 | 200 | 200 | 200 | 200 | **500** | 200 | 200 | 200 |
| 1e-5 | 200 | 200 | 200 | 200 | **500** | 200 | 200 | 200 |
| 1e-4 | 200 | 200 | 200 | 200 | **500** | 200 | 200 | 200 |
| 1e-3 | 200 | 200 | 200 | 200 | **500** | 200 | 200 | 200 |

Four orders of `constr_viol_tol`, a bit-identical column — the knob that declares what a violated constraint is has no effect on the question "is this violation real". The `δ=1e-4` column is 500 at every setting because the comparison was `>` against a threshold that landed exactly on it.

**after**:

| `constr_viol_tol` (tol=1e-6) | 1e-6 | 3e-6 | 1e-5 | 3e-5 | 1e-4 | 3e-4 | 1e-3 | 1e-2 |
| 1e-8 | 200 | 200 | 200 | 200 | 200 | 200 | 200 | 200 |
| 1e-6 | 200 | 200 | 200 | 200 | 200 | 200 | 200 | 200 |
| 1e-5 | 200 | 200 | 200 | 200 | 200 | 200 | 200 | 200 |
| 1e-4 | 200 | 200 | 200 | 200 | 200 | 200 | 200 | 200 |
| 1e-3 | 200 | 200 | 200 | 200 | **500** | **500** | 200 | 200 |

The boundary now tracks the knob. The bottom row's two 500s are the deliberate direction of the fix, not a leftover — see "Fix".

The other half: sweeping `tol` moved the boundary a great deal, and in the wrong direction. The failure band tracked `max(100·tol, 1e-4)`, so at `tol=1e-4` it widened to `1e-2`:

| `tol` (constr_viol_tol=1e-4), δ ≥ 1e-4 | 1e-4 | 3e-4 | 1e-3 | 3e-3 | 1e-2 | 1e-1 |
|---|---|---|---|---|---|---|
| before, tol=1e-5 | 500 | 500 | 500 | 200 | 200 | 200 |
| before, tol=1e-4 | 500 | 500 | 500 | 500 | 200 | 200 |
| before, tol=1e-3 | 500 | 500 | 500 | 500 | **500** | 200 |
| after, any `tol` | 200 | 200 | 200 | 200 | 200 | 200 |

A model infeasible by a full percent answering "your solver broke". Loosening `tol` is the standard reaction to a struggling solve, so the failure widened exactly when the user tried to help.

Not short-circuiting: within δ ≤ 1e-2 the `infeasible` verdicts take 22–48 iterations and the failures 65–802.

## Cause

`tol` is a tolerance on the **KKT error**. The quantity being tested is a constraint violation. Different quantity, different units, so `constr_viol_tol` could not move the verdict and `tol` moved it for no reason.

Six sites, not one. `ipopt_alg.rs`'s restoration-cycle exit is the one the issue names:

```rust
let outer_tol_for_cycle = self.bundle.conv_check.tol_or_default();
let cycle_exit = if reference_theta_unscaled > (100.0 * outer_tol_for_cycle).max(1e-4) { … }
```

It matters disproportionately because square problems are carved out of the `strict` restoration gate (`!is_square_problem`) so the outer gets another shot — which makes this cycle exit their *only* locally-infeasible safety net, and the probe model is square.

Fixing it alone left a hole. At `tol=1e-3` the δ=1e-2 model still returned 500, now via `Restoration_Failed`, because the five locally-infeasible gates in `resto_inner_solver.rs` share the same defect in the shape `is_significant(orig_inf_pr_at_final, violation_scale, 100.0 * outer_tol)`. Same tolerance standing in for the same wrong quantity.

## Fix

All six now test the constraint violation against `constr_viol_tol`, exposed on the `ConvCheck` trait as `constr_viol_tol_or_default()` (default `1e-4`, matching upstream) alongside the existing `tol_or_default()`. In `resto_inner_solver` the floor is read off the same inner conv-check options that already carry `constr_viol_tol` for the layer-2 verdict.

The boundary comparison is `>=`, not `>`. A violation landing exactly on the tolerance the user declared too large is a violation, and the probe hits the boundary to the digit.

**Rejected: `infeas_viol_kappa · constr_viol_tol`**, which the issue suggests for consistency with the rapid-infeasibility pre-filter. Two reasons, both measured:

- It is `1e-2` at defaults, a hundredfold loosening of today's effective `1e-4` threshold, so it would hand back 500 across the whole band in between — the wrong direction for the bug being fixed.
- The kappa margin exists because that detector fires *mid-solve* off a streak heuristic and must not convict an iterate that is still converging. Here restoration has already demonstrably cycled; the certainty comes from the cycle evidence, not from extra violation headroom.

Plain `constr_viol_tol` also reproduces current default behaviour exactly (`max(100·1e-8, 1e-4) = 1e-4 = ` the default `constr_viol_tol`), so the default trajectory of every model in the corpus is untouched.

**Deliberately not changed: below `constr_viol_tol`, 500 stays.** This is the issue's `constr_viol_tol=1e-3 / δ=1e-4` cell, and it answers the report's second question. After the fix, this exit no longer returns 500 on any model whose infeasibility POUNCE can *see* — a violation at or above the user's declared feasibility tolerance is always 200. Below that tolerance the iterate is primal-feasible by the user's own declaration; certifying infeasibility there would be the #505-family false-infeasibility failure from the opposite direction, which is a worse answer than an unhelpful one. The corresponding half of the invariant is pinned by its own test.

A `theta > 0.0` conjunct guards the `>=`: the options layer registers `constr_viol_tol` with a strict lower bound of zero, but a library embedder setting `ConvCheckOptions` directly is not bound by that, and `0 >= 0` would turn an exactly-feasible iterate into an infeasibility certificate.

**Secondary defect, also fixed.** After a non-promoted MC64 hypersensitivity re-solve, the terminal's last `EXIT:` banner was the retry's while the `.sol`, the summary and the JSON report all carried the kept first-solve verdict. `validation/p3_control.py` reads a log exactly that way — last `EXIT:` line, paired with the `.sol` — so it could record a status the `.sol` never held. The CLI now re-emits the verdict that shipped as the final banner, gated on `print_level >= 1` to match `Application::emit_end_summary` (at `print_level 0` there are no banners to disagree).

## Blast radius

At default options the change is bit-identical: the old threshold `max(100·tol, 1e-4)` evaluates to `1e-4` at the default `tol=1e-8`, which is exactly the default `constr_viol_tol`. Behaviour moves only when the user has retuned `tol` or `constr_viol_tol` away from their defaults, or when a violation lands exactly on the boundary.

The `resto_inner_solver` gates get *stricter* at default `tol` (floor `100·1e-8 = 1e-6` → `1e-4`), which is the intended direction: a violation the user calls feasible should not be certified infeasible. Every restoration-verdict test in the repo passes unchanged — `issue_372_infeasible_bounds_status`, `infeasible_status_tol_invariance` (which sweeps `tol` over nine values across four infeasible fixtures), `issue_387_overdetermined_infeasible`, `issue_438_resto_layer2_verdict`, `issue_439_required_infeasibility_reduction`, `false_local_infeasibility`, `scaled_feasible_not_infeasible`, `presolve_certified_infeasibility`.

No CUTE corpus sweep was run — none is available in this environment. The suites above are the closest proxy and are the ones that encode the CUTE-derived regressions the touched comments name (PFIT1–4, DECONVBNE, CRESC100, ACOPR14/30, qcqp750-2nc).

## Tests

`crates/pounce-cli/tests/issue_508_infeasibility_gap_status.rs`, six tests over two vendored fixtures (δ=1e-4 and δ=1e-2 instances of the probe model, ~30 lines of `.nl` each). Verified against the parent commit by reverting `crates/*/src` to `ec70fab` and keeping the tests:

| test | on parent |
|---|---|
| `a_percent_wide_infeasibility_gap_is_tol_invariant` | **FAILED** — `tol=1e-3`, `solve_result_num=500` (`RestorationFailed`) |
| `a_violation_exactly_at_constr_viol_tol_is_infeasible_not_an_internal_error` | **FAILED** — 500 at the `>` boundary |
| `constr_viol_tol_moves_the_boundary` | **FAILED** — the knob has no effect |
| `a_gap_inside_constr_viol_tol_is_not_claimed_infeasible` | passes — direction guard, pins the half of the invariant the fix must *not* break |
| `the_gap_models_are_never_reported_solved` | passes — explicit "never trade one wrong verdict for a worse one" guard |
| `the_last_exit_banner_matches_the_sol_after_a_non_promoted_mc64_retry` | passes — see gap below |

Plus `constr_viol_tol_accessor_tracks_the_option` in `conv_check::opt_error`, which does not compile against the parent (the accessor is new) and covers the debugger's live `set_tolerance` hot-swap path.

**Known gap, stated deliberately.** The banner test is an invariant guard, not a bite-on-parent pin. It asserts the right thing and fails loudly if a future change stops exercising the non-promoted retry path, but it passes pre-fix: every non-promoted retry reachable from this repo's fixture corpus returns `InfeasibleProblemDetected`, the same verdict the `.sol` keeps, so the two banners agree by luck rather than by construction. A sweep over `tol`, `max_iter` and every `.nl` under `tests/fixtures/` found no divergent retry status. The fix does not depend on reproducing one — the retry's status is unconstrained and `scaling_retry_promoted` is false for most statuses, so the mismatch is a reachable state either way, and the reporter observed it on their own model. Vendoring a model that diverges would upgrade this to a real pin. The gap is written into the test's doc comment so it is not rediscovered.

Also not addressed, and out of scope: the report's observation that one cell ran 3000 iterations to a violation of 1.49 against δ=0.1. That is the MC64 re-solve's trajectory under a different scaling — the hypersensitivity the retry exists to probe — and is a separate question from status selection.

---

- [x] Tests fail on the parent commit for the stated reason — three of six do, with the per-test breakdown and the one known gap above
- [ ] Book page under `docs/src/` updated — none needed; no user-facing option or documented behaviour is added or renamed, and `constr_viol_tol` / `tol` keep their documented meanings (this change makes the code honour them)
- [x] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean — full workspace green except `pounce-hsl`, which cannot link its HSL system libraries in this container (pre-existing, unrelated); clippy clean on the three touched crates

---
_Generated by [Claude Code](https://claude.ai/code/session_01VZSkXGDWFkCxc8HugCJ619)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,49 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — infeasible models reported `internalSolverError` because the infeasibility threshold was built from `tol` (#508)
+
+- On a model with no solution, POUNCE chooses between *converged to a point
+  of local infeasibility* (AMPL 200 — "your model has no solution") and
+  *error in step computation* / *restoration failed* (AMPL 500, Pyomo
+  `internalSolverError` — "your solver broke"). That choice was made against
+  a threshold built from `tol`, a tolerance on the **KKT error**, and never
+  consulted `constr_viol_tol`, the option that declares what a violated
+  constraint is. Different quantity, different units.
+- Two measured consequences, on `min (x−5)² s.t. x²+δ = 0` — infeasible for
+  every `δ > 0`, with the reported violation exactly `δ`, so the threshold is
+  visible to the digit. Sweeping `constr_viol_tol` across four orders of
+  magnitude moved the boundary **not at all**: at `constr_viol_tol=1e-3` a
+  violation of `1e-4`, comfortably inside the user's own declared feasibility
+  tolerance, still came back 500. And sweeping `tol` moved it a great deal,
+  in the wrong direction — the failure band tracked `max(100·tol, 1e-4)`, so
+  at `tol=1e-4` every gap from `3e-4` to `1e-2` answered "internal error",
+  including a model infeasible by a full percent. Loosening `tol` is the
+  standard reaction to a struggling solve, so the failure widened exactly
+  when the user tried to help.
+- Every one of these sites now tests the constraint violation against
+  `constr_viol_tol`, and at the boundary with `>=` rather than `>` — a
+  violation *at* the tolerance the user declared too large is a violation.
+  Six thresholds changed: the restoration-cycle exit in `ipopt_alg`, which is
+  the only locally-infeasible safety net square problems have, and the five
+  locally-infeasible gates in `resto_inner_solver` that shared the same
+  `100·tol` floor. The invariant now holds across the whole sweep: a
+  violation at or above the user's declared feasibility tolerance lands in
+  the AMPL infeasible range at every `tol`, and `constr_viol_tol` is what
+  moves it.
+- Where the remaining violation is *below* `constr_viol_tol`, POUNCE still
+  declines to certify infeasibility — the iterate is primal-feasible by the
+  user's own declaration, and claiming otherwise is the false-infeasibility
+  failure from the opposite direction.
+- Also fixed alongside: after a non-promoted MC64 hypersensitivity re-solve,
+  the terminal's last `EXIT:` banner was the retry's while the `.sol`, the
+  summary and the JSON report all carried the kept first-solve verdict. Two
+  banners is expected and announced; the two disagreeing is not, and
+  `validation/p3_control.py` reads the log exactly that way. The CLI now
+  re-emits the verdict that actually shipped as the final banner.
+- Reported by the adversary agent. Regression coverage in
+  `crates/pounce-cli/tests/issue_508_infeasibility_gap_status.rs`.
+
 ### Fixed — false primal-infeasible certificate on redundant equality rows that disagree by one ULP (#496)
 
 - The convex presolve fixes a variable from a singleton equality row and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,12 @@ changes.
   `validation/p3_control.py` reads the log exactly that way. The CLI now
   re-emits the verdict that actually shipped as the final banner.
 - Reported by the adversary agent. Regression coverage in
-  `crates/pounce-cli/tests/issue_508_infeasibility_gap_status.rs`.
+  `crates/pounce-cli/tests/issue_508_infeasibility_gap_status.rs`. The banner
+  ordering is covered by an invariant guard rather than a bite-on-parent pin:
+  every non-promoted MC64 retry reachable from the fixture corpus happens to
+  return the same verdict the `.sol` keeps, so the two banners agree there by
+  luck; the mismatch is still a reachable state of the code. Noted in the test
+  so the gap is a known one.
 
 ### Fixed — false primal-infeasible certificate on redundant equality rows that disagree by one ULP (#496)
 
--- a/crates/pounce-algorithm/src/conv_check/opt_error.rs
+++ b/crates/pounce-algorithm/src/conv_check/opt_error.rs
@@ -658,6 +658,10 @@ impl ConvCheck for OptErrorConvCheck {
         self.tol
     }
 
+    fn constr_viol_tol_or_default(&self) -> Number {
+        self.constr_viol_tol
+    }
+
     fn acceptable_constr_viol_tol_or_default(&self) -> Number {
         self.acceptable_constr_viol_tol
     }
@@ -1106,6 +1110,27 @@ mod tests {
         // Gradient flat but violation below threshold → nearly
         // feasible, does not count.
         assert!(!c.is_infeasible_stationary(1e-3, 0.0, 1e-9));
+    }
+
+    /// gh #508: the status-decision sites that ask "is this violation real"
+    /// read `constr_viol_tol` off the policy, so a user setting has to reach
+    /// them — the defect was a threshold built from `tol` that no
+    /// `constr_viol_tol` value could move. `set_tolerance` is the debugger's
+    /// live hot-swap path and must be visible through the accessor too.
+    #[test]
+    fn constr_viol_tol_accessor_tracks_the_option() {
+        let mut c = OptErrorConvCheck {
+            tol: 1e-6,
+            constr_viol_tol: 1e-3,
+            ..Default::default()
+        };
+        assert_eq!(c.constr_viol_tol_or_default(), 1e-3);
+        // Independent of `tol` — retuning convergence must not retune what
+        // counts as a violated constraint.
+        c.tol = 1e-10;
+        assert_eq!(c.constr_viol_tol_or_default(), 1e-3);
+        assert!(c.set_tolerance("constr_viol_tol", 1e-7));
+        assert_eq!(c.constr_viol_tol_or_default(), 1e-7);
     }
 
     #[test]
--- a/crates/pounce-algorithm/src/conv_check/trait.rs
+++ b/crates/pounce-algorithm/src/conv_check/trait.rs
@@ -143,6 +143,22 @@ pub trait ConvCheck {
         1e-8
     }
 
+    /// Primal-feasibility tolerance `constr_viol_tol`, in the unscaled max-norm
+    /// space `curr_unscaled_primal_infeasibility_max` reports — the option that
+    /// declares what a *violated* constraint is.
+    ///
+    /// Read by the status-decision sites that have to answer "is this violation
+    /// real" (gh #508). That question is about the constraint violation, so it
+    /// must be asked with the constraint-violation tolerance; asking it with
+    /// `tol` — a tolerance on the **KKT error**, a different quantity in
+    /// different units — makes the answer move when the user retunes
+    /// convergence and stand still when they retune feasibility. Default `1e-4`
+    /// matches upstream's default `constr_viol_tol`; policies that track no such
+    /// tolerance keep it.
+    fn constr_viol_tol_or_default(&self) -> Number {
+        1e-4
+    }
+
     /// Acceptable-level primal-feasibility band `acceptable_constr_viol_tol`, in
     /// the unscaled max-norm space `curr_unscaled_primal_infeasibility_max`
     /// reports. Read by the best-acceptable fallback's feasibility-aware ranking
--- a/crates/pounce-algorithm/src/ipopt_alg.rs
+++ b/crates/pounce-algorithm/src/ipopt_alg.rs
@@ -2488,12 +2488,49 @@ impl IpoptAlgorithm {
             .as_ref()
             .expect("curr set before invoke_restoration")
             .clone();
-        // Helper: when the cycle detector fires and the orig cv is
-        // bounded away from outer tol (e.g. PFIT1's 2.73e-2), the
-        // outer is stuck at a feasibility-stationary point and the
-        // honest exit is `LocalInfeasibility`. Below that threshold
-        // the failure is numerical, not algorithmic, so retain
-        // `ErrorInStepComputation`.
+        // Helper: when the cycle detector fires and the orig cv is a
+        // violation the *user* calls a violation (e.g. PFIT1's 2.73e-2),
+        // the outer is stuck at a feasibility-stationary point and the
+        // honest exit is `LocalInfeasibility`. Below that threshold the
+        // iterate is primal-feasible by the user's own declaration, so there
+        // is no infeasibility to certify — the failure is numerical, not
+        // algorithmic, and `ErrorInStepComputation` is retained.
+        //
+        // The threshold is `constr_viol_tol`, and *only* `constr_viol_tol`
+        // (gh #508). The question this ternary asks — "is this violation
+        // real?" — is a question about the constraint violation, so it has to
+        // be asked with the option that declares what a violated constraint
+        // is. The previous form, `max(100·tol, 1e-4)`, was built from `tol`, a
+        // tolerance on the **KKT error**: different quantity, different units,
+        // and it never consulted `constr_viol_tol` at all. Two consequences,
+        // both measured on `min (x-5)² s.t. x²+δ = 0` (infeasible for every
+        // δ>0, reported violation exactly δ):
+        //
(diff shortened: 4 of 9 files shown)


jkitchin:
CI status: red for infrastructure, not for this diff — and the same failure is on `main`.

**On this branch** ([run 1954](https://github.com/jkitchin/pounce/actions/runs/31124409849), `097a471`), every job that got a runner passed:

| job | result |
|---|---|
| Test (ubuntu-latest) — release-consistency, docs-consistency, fmt, clippy, build, full test suite, `pounce-hsl` compile-check | ✅ |
| WebAssembly build + smoke | ✅ |
| Python tests (torch) | ✅ |
| Python tests (jax) | ✅ |
| wheel smoke (pyomo-pounce) | ✅ |
| wheel smoke (pounce-solver) | ⛔ cancelled — queued 17:52→19:03 UTC, never assigned a runner |

The run's `failure` conclusion is that one cancelled job.

**Same thing on the base branch.** [Run 1939](https://github.com/jkitchin/pounce/actions/runs/31119017600) on `main` @ `ec70fab` — the commit this PR is based on — is also red: `wheel smoke (pounce-solver)` and `Python tests (jax)` both failed **in "Set up job"**, step 1, before any repo code ran; `Python tests (torch)` and `WebAssembly` were cancelled with no runner assigned. `Test (ubuntu-latest)` passed there too. `Deploy docs` and `release-docker` are red on the same commit.

**Runner starvation is repo-wide and ongoing.** Five runs on `claude/issue-506-i6ugje` have been stuck `queued` since ~17:00 UTC, and run 1954 is the newest run created anywhere in the repository — nothing has been scheduled since 17:52 UTC. Two pushes to this branch after that (`0696903`, then an empty `3213ea1` specifically to trigger one) produced no workflow run at all, which is why the PR currently shows no checks against its head SHA. I'm not pushing further no-op commits; that isn't the lever here.

Nothing to fix on this diff as far as CI can currently tell. Locally on the head commit: full workspace `cargo test` green (except `pounce-hsl`, which can't link its HSL system libraries in this container — the same crate CI covers with a compile-check), `cargo fmt --check` clean, `cargo clippy` clean on the three touched crates. I'll re-check periodically and re-run once runners recover.

---

PR #525, opened 2026-08-07
fix(convex): stop a near-forcing row from falsifying a feasible LP (#523)

### LINKED ISSUE #523: pounce-convex: false Infeasible_Problem_Detected at iteration 0 on bore3d / QBORE3D (NLP path and Ipopt both solve them)
`pounce-convex` reports `Infeasible_Problem_Detected` at **iteration 0** on two
feasible, well-posed benchmark problems that both the NLP path and Ipopt-MA57
solve to optimality. Found in the v0.9.0 benchmark sweep (commit `a664dc05`,
FERAL linear solver).

The two problems are the same underlying model — Netlib `bore3d` and its
Maros-Mészáros quadratic twin `QBORE3D`, identical size (n=315, m=233), the
latter just adding a convex quadratic objective — so the LP-IPM and QP-IPM
entry points are both tripping on the same feasible region. Almost certainly
one bug, not two.

## Repro

```sh
pounce $POUNCE_BENCH_DATA/lp/nl/bore3d.nl  --no-sol
pounce $POUNCE_BENCH_DATA/qp/nl/QBORE3D.nl --no-sol
```

```
Problem class: LP. Selected solver: convex QP interior-point (pounce-convex) [solver_selection=auto].
POUNCE (LP IPM, pounce-convex): Problem is primal infeasible.  obj=0.00000000  iters=0  (0.005s)
```

Forcing the NLP path solves both, and matches the committed Ipopt-MA57
reference objective to ~10 significant figures:

```sh
pounce $POUNCE_BENCH_DATA/lp/nl/bore3d.nl  --no-sol solver_selection=nlp
pounce $POUNCE_BENCH_DATA/qp/nl/QBORE3D.nl --no-sol solver_selection=nlp
```

| problem | convex path (auto) | `solver_selection=nlp` | ipopt-ma57 reference |
|---|---|---|---|
| `bore3d`  | Infeasible_Problem_Detected, 0 iters | Solved To Acceptable Level, obj 1373.0803843798531, 331 iters | Solved_To_Acceptable_Level, obj 1373.0803839123494 |
| `QBORE3D` | Infeasible_Problem_Detected, 0 iters | Optimal Solution Found, obj 3100.2007904156926, 178 iters | Solve_Succeeded, obj 3100.2007904156912 |

So the feasible set is fine and the infeasibility certificate is false.

## Where to look

`iters=0` means nothing in the IPM loop ran — the claim comes from a
pre-iteration screen, not from a Farkas certificate the iteration derived.
Candidates, in `crates/pounce-convex/src`:

- `presolve.rs:2100` — `PresolveOutcome::Infeasible => trivial(QpStatus::PrimalInfeasible)`
- `ipm.rs:2596` — the setup-time screen that builds a `PrimalInfeasible`
  solution ("Build a `PrimalInfeasible` solution reported by a **setup-time**
  screen")
- `qp.rs:234` — the empty-box-by-inspection screen

A `print_level=8` run emits no additional diagnostic on the way out, so
whichever screen fires does so silently — worth logging the triggering row /
column / bound alongside the fix, since a false infeasibility with no trace is
expensive to localize.

Note the reported termination block is itself internally inconsistent with the
claim: constraint violation 1.79e+01 and dual infeasibility 3.35e+02 at the
returned point, i.e. it reports a point that is neither feasible nor a
certificate.

## Impact

This is the worst failure class a solver has — a confident wrong answer on a
feasible problem, returned in 5 ms with no warning. It is currently the **only**
non-matching failure in the `lp` suite (367/371, other 3 misses are problems
Ipopt also fails) and the **only** failure in the `qp` suite (137/138).

## Suggested guard

Beyond the specific fix: a pre-iteration infeasibility screen should be
required to hand back a certificate that is checked before the status is
emitted, so a buggy screen degrades to "solve normally" rather than to a false
`Infeasible_Problem_Detected`.

jkitchin:
Fixes #523

## Problem

`pounce-convex` returned `Infeasible_Problem_Detected` at iteration 0, in ~5 ms with no diagnostic, on two feasible problems: netlib `bore3d` and its Maros-Mészáros quadratic twin `QBORE3D` (both n=315, m=233 — the same model, the latter adding a convex quadratic objective). They were the only non-matching failure in the `lp` suite and the only one in the `qp` suite.

`bore3d`:

| | status | objective | iters |
|---|---|---|---|
| before | Infeasible_Problem_Detected | 0.0 | 0 |
| after | Optimal | 1373.0803943 | 21 |
| oracle (ipopt-ma57) | Solved_To_Acceptable_Level | 1373.0803839 | — |

`QBORE3D`:

| | status | objective | iters |
|---|---|---|---|
| before | Infeasible_Problem_Detected | 0.0 | 0 |
| after | Optimal | 3100.2008018 | 21 |
| oracle (ipopt-ma57) | Solve_Succeeded | 3100.2007904 | — |

The reported termination block was internally inconsistent with its own status — constraint violation `1.79e+01`, dual infeasibility `3.35e+02` at the returned point, i.e. a point that is neither feasible nor a certificate. That inconsistency is enforcement item 1 of `dev-notes/termination-status-invariants.md`, already written down, and it was printing.

The benchmark corpus is not vendored, so these were reproduced by fetching `bore3d` from the COIN-OR netlib mirror and `QBORE3D.mat` from `qpsolvers/maros_meszaros_qpbenchmark`, converted to `QpProblem` directly. Both reproduced the reported failure exactly (`PrimalInfeasible`, `obj=0`, `iters=0`).

## Cause

Presolve's **forcing-constraint** reduction — none of the three screens the issue nominated.

Bound tightening propagates a group of nonnegative variables' upper bounds geometrically toward their true limit of zero, round after round; the cascade converges but never terminates, so the fixpoint runs to its layer cap. By round 21 those boxes are ~1e-8 wide, and one equality row's activity range `[-6.29e-10, 4.87e-8]` comes within `ACTIVITY_TOL` (1e-9) of its right-hand side `0` at the min vertex:

```
row 11 (b = 0):  8.17e-2·x0 + 5.32e-2·x1 + 8.3e-3·x2 − 1.14e-1·x3 − 5.7e-3·x4 + x5 = 0
boxes: [0, 1.98e-7] [0, 3.04e-7] [0, 2.21e-8] [0, 4.76e-9] [0, 1.52e-8] [0, 1.62e-8]
amin = −6.29e-10   amax = 4.87e-8
```

Forcing read that as "this row can hold only at that vertex" and pinned all six variables to bounds — `x0, x1, x2, x5` to `0`, and `x3, x4` (negative coefficients) to *upper* bounds `4.76e-9` and `1.52e-8`. But `b = 0` lies strictly inside `[amin, amax]`; the row is not forcing, it merely looks forcing at 1e-9.

The key quantity: a residual gap `g` licenses a displacement of `g/|coefⱼ|` on variable `j`, not `g`. Here `6.29e-10` bought `1.52e-8` — 24× larger — because the coefficients are small. Substituted forward, the next row those two appear in (`x0+x1+x2+x3+x4−x5 = 0`) read as inconsistent by `2.0e-8` against a tolerance of `1.0e-9`, and presolve called a feasible problem infeasible.

## Fix

**1. Forcing must be tight, not merely close.** A forcing row now has to clear a stronger bar than "the gap is small": the gap must be small enough that *every* pinned variable lands within `FORCING_PIN_TOL` of where the pin claims it is — `min(gap/|coefⱼ|, widthⱼ) ≤ tol`, capped by the box width since a variable cannot leave its box either way. A row that genuinely touches (`gap == 0`) is unaffected, which is every legitimate forcing row in the existing suite.

**2. An infeasibility verdict is re-derived before it is emitted** — the guard the issue asked for. When any screen reports infeasible, presolve re-runs from the original problem with the two reductions that *fix a variable at a tolerance-chosen value* withheld (forcing constraints, dominated columns), and only a verdict that pass reaches on its own is returned as `PrimalInfeasible`. Otherwise the reduction is solved normally and the discarded claim is kept on the handle (`Presolve::discarded_infeasibility`).

Considered and rejected: withholding *all* tolerance-based reductions in the confirming pass. That reads as the safer line but is the wrong one — it costs real detection power. Bound tightening is what proves `x ≤ −gap ∧ x ≥ 0` infeasible (`contradictory_singleton_rows_are_seen_as_an_empty_box`), and the activity screens are what prove `x0+x1 = 5` over `[0,1]²` infeasible; both would have downgraded to a full solve. The line that actually matters is **fix vs. report**: a wrong report is one wrong answer, while a wrong *fixing* propagates into every row the variable touches and surfaces somewhere else entirely, in a screen behaving correctly on data that is not. Only forcing and dominated columns have that shape. Everything that narrows, drops, or screens is retained, so no infeasibility class detectable before is lost.

**3. The silent exit is gone.** `PresolveOutcome::Infeasible` carries an `InfeasibleTrigger` (the screen, plus the row/column/bound and the compared values), printed by the CLI for both the proved and the discarded case. Previously even `print_level=8` emitted nothing on the way out — which is most of why this cost so much to localize.

Layer 2 was verified to stand alone: with layer 1 reverted, both problems still solve to the reference objective, reporting

```
Presolve: discarded an unconfirmed infeasibility claim — row emptied by substitution is
inconsistent (equality row 33: residual -1.992965e-8 exceeds tolerance 1.000000e-9
(cancellation scale 1.517033e-8)); solving normally
```

**Deliberately out of scope:** the bound-tightening cascade itself. It never converges on this model — it shaves those boxes geometrically until the layer cap cuts it off, burning ~20 fixpoint rounds. That is sound (it is approaching the true bound of zero) and no longer dangerous now that forcing will not pin on it, but it is wasted work. Worth a separate issue; fixing it here would have mixed a performance change into a correctness fix. The sweep below shows no runtime cost, so it is not urgent.

## Blast radius

- **`pounce-convex` presolve**, all callers (CLI convex path, `pounce-py`, `solve_with_presolve`). The forcing change only withholds pins that were previously made on a near-touch; an exact touch behaves identically. The re-derivation runs only when a verdict is raised — a cold path — so there is no cost on any solve that does not claim infeasibility.
- **Breaking (`pounce-convex` API):** `PresolveOutcome::Infeasible` is now a tuple variant. `matches!(…, PresolveOutcome::Infeasible)` becomes `PresolveOutcome::Infeasible(_)`. 24 in-tree call sites updated, all mechanical; noted in the CHANGELOG.
- **Corpus sweep: run, clean** ([comment](https://github.com/jkitchin/pounce/pull/525#issuecomment-5219879702)). `lp` 368/371 (baseline 367/371), `qp` 138/138 (baseline 137/138), against `a664dc05` on the same host and corpus. Across all **509 problems**: `FIXED=2` (the two intended), `BROKE=0`, `obj-drift=0`. Runtime within noise both ways (`lp` 580s vs 583s, `qp` 220s vs 218s), consistent with the re-derivation being a cold path. This is the check that would catch a legitimate forcing pin the fix now declines; nothing in either corpus had one declined. Presolve still reports `forcing 17` on `bore3d` after the fix — the reduction is not disabled on the model that motivated the change.
- **NLP suites not swept**, but the exposure is ruled out by static reachability rather than reasoning: `pounce_convex::presolve` has exactly one production call site (`crates/pounce-cli/src/main.rs:2124`, inside `run_convex_qp`), which has exactly one caller in the convex dispatch branch, and both `main.rs` hunks in this diff are inside that function. `pounce-py` never touches the presolve module; `pounce-algorithm` names `pounce_convex` only in a comment. The NLP path's presolve is `pounce-presolve`, a separate crate, untouched.
- **`presolve_conic` got the same guard and no sweep touched it** — the conic path is `cblib`/SOCP/SDP, not `lp`/`qp`. Same three-line wrapper as the orthant path, and the in-tree conic presolve tests pass, but that gap is real and named here rather than left implicit.

## Tests

`crates/pounce-convex/tests/issue523_false_forcing_infeasibility.rs`, four tests:

- `near_forcing_row_does_not_falsify_a_feasible_lp` — the `bore3d` geometry distilled to the two rows that matter, with the exact coefficients and box widths from round 21. **Fails on the parent commit**, and for the right reason twice over: with the guard active it fails on the `discarded_infeasibility().is_none()` assertion (the guard caught it and downgraded); with both layers reverted it fails on `panic!("feasible LP reported primal infeasible by presolve")`, the original bug. That is how each layer was verified independently.
- `an_exactly_touching_row_still_forces` — over-correction guard; a row with `gap == 0` still fires and still fixes.
- `a_real_infeasibility_is_reported_with_its_trigger` — a genuine infeasibility still proves out, and names its screen and row.
- `a_genuine_forcing_infeasibility_is_still_confirmed` — the guard's cost, stated as a test: an infeasibility only reachable through a forcing pin must still come out infeasible (domain propagation re-derives it), or at minimum survive to the solver.

**Not pinned:** `bore3d` and `QBORE3D` themselves. The `.nl` inputs are ~2 GB of gitignored corpus, too large to vendor, so the regression test carries the distilled geometry instead and the full instances stay a benchmark-sweep check — which is the sweep recorded above.

---

- [x] Book page under `docs/src/` updated (`lp-qp-routing.md`, convex presolve section — the reported output changed; no new page, so `SUMMARY.md` is unchanged)
- [x] `cargo fmt --all -- --check`, `cargo clippy` (CI gate: `-D clippy::correctness -D clippy::suspicious`), `cargo test --workspace --exclude pounce-hsl` clean

`dev-notes/termination-status-invariants.md` records this as the fifth instance of that note's invariant — the note exists so a fifth would be recognised rather than rediscovered, and this one is the first in a different subsystem, by a mechanism the other four do not share (a reduction, not a threshold). The generalisation the earlier four suggest — *audit every threshold* — would not have found it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,56 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — `pounce-convex` reported false `Infeasible_Problem_Detected` at iteration 0 on `bore3d` / `QBORE3D` (#523)
+
+- Netlib `bore3d` and its Maros-Mészáros quadratic twin `QBORE3D` (both
+  n=315, m=233, both feasible) came back from the convex path as
+  `Infeasible_Problem_Detected` in ~5 ms with `iters=0` and no diagnostic,
+  while the NLP path (`solver_selection=nlp`) solved both to the committed
+  Ipopt-MA57 reference objectives. They were the only non-matching failure
+  in the `lp` suite and the only one in the `qp` suite.
+- The claim came from presolve's **forcing-constraint** reduction. Bound
+  tightening propagated a group of nonnegative variables' upper bounds
+  geometrically toward their true limit of zero; by round 21 of the
+  presolve fixpoint those boxes were ~1e-8 wide, and one equality row's
+  activity range `[-6.29e-10, 4.87e-8]` came within `ACTIVITY_TOL` (1e-9)
+  of its right-hand side `0`. Forcing read that as "this row can hold only
+  at its min vertex" and pinned all six of its variables to bounds — two of
+  them, with coefficients `-1.14e-1` and `-5.7e-3`, to *upper* bounds
+  `4.8e-9` and `1.5e-8` from zero. A gap of `6.29e-10` had licensed a
+  displacement of `1.5e-8`. Substituted into the next row those two
+  appeared in, the residual was `2.0e-8` against a tolerance of `1.0e-9`,
+  and a feasible problem was declared infeasible.
+- **Forcing now requires the pin to be tight.** A row whose activity range
+  only *approximately* touches its right-hand side may still spend the
+  residual gap, moving each variable up to `gap / |coefⱼ|` (capped by its
+  box width) off the bound the pin claims it must occupy. Forcing fires
+  only when that displacement is within `FORCING_PIN_TOL` for **every**
+  variable in the row, so a pin is a deduction rather than a guess. A row
+  that genuinely touches (gap 0) is unaffected.
+- **An infeasibility verdict is now re-derived before it is emitted.**
+  Two reductions — forcing constraints and dominated columns — fix a
+  variable at a value chosen from a tolerance judgment, and a wrong one is
+  substituted into every row that variable touches until some row reads as
+  contradictory. When any screen reports infeasible, presolve re-runs from
+  the original problem with those two withheld; only a verdict that pass
+  reaches on its own is returned as `PrimalInfeasible`. Otherwise the
+  reduction is solved normally and the discarded claim is kept on the
+  handle (`Presolve::discarded_infeasibility`). Everything that only
+  *reports* — empty rows, activity ranges, parallel rows, emptied-row
+  residuals — is retained, so no infeasibility class detectable before is
+  lost. Verified: with the forcing fix reverted, the guard alone still
+  solves both problems to the reference objective.
+- **The screen that fired is now named.** `PresolveOutcome::Infeasible`
+  carries an `InfeasibleTrigger` (the screen plus the row / column / bound
+  and the compared values), and the CLI prints it — `Presolve: proved
+  primal infeasible — <screen> (<detail>)`, or `Presolve: discarded an
+  unconfirmed infeasibility claim — …; solving normally`. Previously even
+  `print_level=8` emitted nothing on the way out.
+- **Breaking (pounce-convex API):** `PresolveOutcome::Infeasible` is now a
+  tuple variant; `matches!(…, PresolveOutcome::Infeasible)` becomes
+  `PresolveOutcome::Infeasible(_)`.
+
 ### Fixed — tightening `constr_viol_tol` made POUNCE more likely to report local infeasibility (#519)
 
 - Rapid infeasibility detection counted an iterate toward its streak when
--- a/crates/pounce-cli/src/main.rs
+++ b/crates/pounce-cli/src/main.rs
@@ -2194,6 +2194,17 @@ fn run_convex_qp(
     } else if presolve_on {
         match presolve(&qp) {
             PresolveOutcome::Reduced(ps) => {
+                // A screen claimed infeasibility and the re-derivation without
+                // the speculative fixings would not reproduce it, so presolve
+                // solved on instead (gh #523). Say so: the guard turns a false
+                // `Infeasible_Problem_Detected` into a normal solve, and this
+                // line is the only trace of the reduction that misfired.
+                if let Some(trigger) = ps.discarded_infeasibility() {
+                    println!(
+                        "Presolve: discarded an unconfirmed infeasibility claim — \
+                         {trigger}; solving normally"
+                    );
+                }
                 let st = ps.stats();
                 if st.reduced_anything() {
                     println!(
@@ -2221,7 +2232,13 @@ fn run_convex_qp(
                 };
                 ps.postsolve(&red)
             }
-            PresolveOutcome::Infeasible => trivial(QpStatus::PrimalInfeasible),
+            PresolveOutcome::Infeasible(trigger) => {
+                // Name the screen and what it tripped on. A presolve
+                // infeasibility arrives with no iteration behind it, so this
+                // line is the whole record of *why* (gh #523).
+                println!("Presolve: proved primal infeasible — {trigger}");
+                trivial(QpStatus::PrimalInfeasible)
+            }
             PresolveOutcome::Unbounded => trivial(QpStatus::DualInfeasible),
         }
     } else if use_active_set {
--- a/crates/pounce-convex/examples/presolve_reductions.rs
+++ b/crates/pounce-convex/examples/presolve_reductions.rs
@@ -16,7 +16,7 @@ fn backend() -> Box<dyn SparseSymLinearSolverInterface> {
 fn report(name: &str, prob: &QpProblem) {
     print!("{name:<34} {}×{} → ", prob.n, prob.m_eq() + prob.m_ineq());
     match presolve(prob) {
-        PresolveOutcome::Infeasible => println!("INFEASIBLE (detected in presolve)"),
+        PresolveOutcome::Infeasible(_) => println!("INFEASIBLE (detected in presolve)"),
         PresolveOutcome::Unbounded => println!("UNBOUNDED (detected in presolve)"),
         PresolveOutcome::Reduced(ps) => {
             let r = &ps.reduced;
--- a/crates/pounce-convex/src/presolve.rs
+++ b/crates/pounce-convex/src/presolve.rs
@@ -162,11 +162,39 @@ pub enum PresolveOutcome {
     Reduced(Presolve),
     /// Presolve proved the problem primal-infeasible (e.g. an empty row
     /// `0 = b` with `b ≠ 0`, contradictory fixed bounds, or duplicate
-    /// equality rows with different right-hand sides).
-    Infeasible,
+    /// equality rows with different right-hand sides). Carries the screen
+    /// that fired and what it tripped on — see [`InfeasibleTrigger`].
+    Infeasible(InfeasibleTrigger),
     /// Presolve proved the problem unbounded below (a free column with a
     /// nonzero objective coefficient).
     Unbounded,
+}
+
+/// Which screen concluded that the problem is primal-infeasible, and what it
+/// tripped on.
+///
+/// A presolve infeasibility comes back in milliseconds with no iteration
+/// behind it, so when it is wrong it is the most expensive failure the solver
+/// has: a confident wrong answer with no trace. Carrying the trigger makes
+/// the claim auditable — the CLI prints it, and a bug report names a row, a
+/// column and the numbers that were compared instead of `iters=0` (gh #523).
+#[derive(Debug, Clone, PartialEq)]
+pub struct InfeasibleTrigger {
+    /// The screen that fired, e.g. `"empty equality row"`.
+    pub screen: &'static str,
+    /// The row / column / bound it tripped on, with the compared values.
+    pub detail: String,
+}
+
+impl std::fmt::Display for InfeasibleTrigger {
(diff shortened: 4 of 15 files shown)


jkitchin:
## Sweep run — clean, no regressions

The `lp`/`qp` sweep this PR asks for before merge has been run. Both suites,
every problem, against a baseline captured from `main` earlier today on the
same host with the same corpus.

```
[lp] 368/371 solved (baseline 367/371) | solver 580s vs baseline 583s | FIXED=1 BROKE=0 obj-drift=0
  FIXED  bore3d:  Infeasible_Problem_Detected -> Solve_Succeeded
[qp] 138/138 solved (baseline 137/138) | solver 220s vs baseline 218s | FIXED=1 BROKE=0 obj-drift=0
  FIXED  QBORE3D: Infeasible_Problem_Detected -> Solve_Succeeded
```

Across all **509 problems**, the only two status changes are the two intended
fixes. No problem that solved before stopped solving, and no objective moved
more than `1e-6` relative on the 507 solved by both.

That is the specific risk the PR flags — "the check that would catch a
legitimate forcing pin this now declines." Nothing in the LP or QP corpus had
one declined.

Runtime is within noise in both directions (`lp` 580s vs 583s, `qp` 220s vs
218s), consistent with the re-derivation being a cold path that only runs when
a verdict is raised.

### Setup

| | |
|---|---|
| PR binary | `d11500d7`, release, FERAL |
| Baseline | `a664dc05` (`main`), release, FERAL — the run that produced the v0.9.0 sweep |
| Host | Darwin 25.5.0 arm64, single-threaded per the driver |
| Corpus | `$POUNCE_BENCH_DATA` — `lp` 371 `.nl`, `qp` 138 `.nl` |
| Driver | `make lp-rerun / qp-rerun POUNCE_BIN=<pr binary>` |

Both target problems end-to-end on the PR binary:

```
bore3d    Optimal Solution Found.  obj=1373.08039427  iters=21  (0.021s)
QBORE3D   Optimal Solution Found.  obj=3100.20080182  iters=21  (0.019s)
```

matching the committed Ipopt-MA57 references (`1373.0803839`, `3100.2007904`)
to ~4e-9 relative.

Worth noting for the record: presolve still reports `forcing 17` on this model
after the fix. The reduction is not disabled on the problem that motivated the
change — 17 rows still clear the tighter bar and still pin. That is the
distinction between "forcing must be tight" and "forcing is too dangerous to
run," and the sweep is consistent with the former.

### Not covered

Only `lp` and `qp` were run, which is what this PR asked for. The change is
confined to the convex presolve, so the NLP-path suites (`vanderbei`,
`mittelmann`, …) should be untouched — but that is reasoning, not measurement,
and no sweep was run to confirm it.

jkitchin:
Thanks — that clears the pre-merge check this PR asked for.

`forcing 17` surviving on `bore3d` is the number I'd have wanted to see. The change was meant to raise the bar for a pin, not retire the reduction, and a model where forcing went to zero after the fix would have meant the tightness test was keying on something structural rather than on the gap. 17 rows still clearing it on the very model that motivated the change is the right shape.

## The NLP-suite gap

You flagged this as reasoning rather than measurement, so I checked the call graph instead of leaving it at reasoning. It holds, and for a narrower reason than "the change is confined to the convex presolve":

`pounce_convex::presolve` has **exactly one** production call site in the workspace —

```
crates/pounce-cli/src/main.rs:2124   (inside run_convex_qp)
```

— and `run_convex_qp` has exactly one caller, `main.rs:1014`, in the convex dispatch branch. Of the three crates that depend on `pounce-convex`: `pounce-py` uses the QP/SOS entry points and never the presolve module, and `pounce-algorithm` names `pounce_convex` only in a comment. Both `main.rs` hunks in this diff are inside `run_convex_qp`.

The NLP path's presolve is `pounce-presolve`, a separate crate, untouched by this diff.

So the NLP suites cannot reach any changed line. That is a static-reachability argument, not a sweep — it rules out the code being *executed*, which is the whole of the exposure here, but it is worth naming as the weaker claim it is.

## One thing the sweep does not cover, for the record

`presolve_conic` got the same re-derivation guard, and neither `lp` nor `qp` exercises it — the conic path is `cblib`/SOCP/SDP. It is the same three-line wrapper as the orthant path and the in-tree conic presolve tests pass, so I did not treat it as a separate risk, but no sweep touched it either.


---

PR #526, opened 2026-08-07
fix(cli): make the local-infeasibility re-solve vary the barrier trajectory, not just the scaling (#524)

### LINKED ISSUE #524: Vanderbei: cresc4 (false local-infeasibility) and steenbrf (3000-iter stall) — the only 2 problems Ipopt solves and POUNCE does not; both fixed by mu_strategy=adaptive
These are the only two problems in the 733-problem Vanderbei suite where POUNCE
fails and the committed Ipopt-MA57 reference succeeds. Found in the v0.9.0
benchmark sweep (commit `a664dc05`, FERAL). Overall the suite is net positive
against the reference — 701/733, and POUNCE solves 14 problems Ipopt fails —
so this is about closing the last two, not a broad regression.

They look unrelated (a false local-infeasibility verdict vs. an iteration-limit
stall) but they share a cause worth investigating: **both are fixed by
`mu_strategy=adaptive`**, and neither is a capability gap.

| problem | n | m | POUNCE (defaults) | Ipopt-MA57 |
|---|---|---|---|---|
| `cresc4`   | 6   | 8   | Infeasible_Problem_Detected, 74 iters, 0.06s | Solve_Succeeded, obj 0.8718975393, 71 iters |
| `steenbrf` | 468 | 108 | Maximum_Iterations_Exceeded, 3000 iters, 3.5s | Solved_To_Acceptable_Level, obj 1321.652, 1846 iters |

## The shared workaround

```
                          cresc4                    steenbrf
default                   Infeasible_Problem_Det.   Maximum_Iterations_Exceeded
mu_strategy=adaptive      Optimal Solution Found    Solved To Acceptable Level
nlp_scaling_method=none   Optimal Solution Found    Maximum_Iterations_Exceeded
```

Under `mu_strategy=adaptive`:

- `cresc4` → obj **0.87189752899987727**, matching the Ipopt reference
  0.8718975393 to 8 significant figures.
- `steenbrf` → obj **282.678**, 567 iterations, 0.57s, dual infeasibility
  8.7e-08, constraint violation 4.5e-13 — a *better* point than Ipopt's
  acceptable-level answer at obj 1321.65, and 3x fewer iterations.

That both fall to the same knob suggests the monotone-µ default is the common
factor, not two independent bugs.

## `cresc4` — false local-infeasibility on a 6-variable problem

The smallest possible reproduction of the worst failure class: a confident
infeasibility verdict on a problem that is feasible and that Ipopt solves in 71
iterations.

```sh
pounce $POUNCE_BENCH_DATA/vanderbei/nl/cresc4.nl --no-sol
```

```
Number of Iterations....: 74
Objective...............:   2.8658120760032343e-09
Dual infeasibility......:   1.0000000000237212e+00
Constraint violation....:   5.1235431712319179e-01
Complementarity.........:   5.9536696090209468e+01
pounce: MC64 re-solve did not recover (InfeasibleProblemDetected); keeping the
original local-infeasibility verdict (now corroborated by a second scaling).
```

Two things to note:

1. The MC64 re-solve guard fired and *corroborated* the wrong verdict. That
   guard exists to catch exactly this, and a second scaling agreeing gave false
   confidence — but `nlp_scaling_method=none` solves the problem, so a third
   scaling would have disagreed. Worth asking whether the re-solve is diverse
   enough to be evidence.
2. It converges to a point with constraint violation 0.51 and calls that a
   point of local infeasibility, on a problem with 6 variables and 8
   constraints. At that size the whole trajectory is inspectable.

Note the neighbours `cresc50` / `cresc100` / `cresc132` are reported infeasible
by *both* solvers, so the family is genuinely nasty; `cresc4` is the one where
ground truth is known and POUNCE is wrong.

## `steenbrf` — line-search stall at a feasible point

```sh
pounce $POUNCE_BENCH_DATA/vanderbei/nl/steenbrf.nl --no-sol
```

Hits the 3000-iteration cap. The tail shows the shape of the stall — feasible
throughout, but with collapsing step sizes and 20+ line-search backtracks per
iteration:

```
2997  3.9781835e+02 3.08e-09 8.20e+02  -3.8 4.48e-03  2.4 1.00e+00 2.05e-07h 23
2998  3.9781835e+02 3.08e-09 2.35e+03  -3.8 1.27e-02  1.9 1.00e+00 5.06e-08h 24
2999  3.9781835e+02 3.08e-09 3.84e+03  -3.8 4.98e-02  1.4 1.00e+00 3.29e-09h 27
3000  3.9781835e+02 3.08e-09 5.39e+03  -3.8 2.24e-01  1.0 1.00e+00 8.21e-08f 21
```

Final state: constraint violation 3.1e-09, complementarity 1.2e-02, dual
infeasibility 5.4e+03. The objective is pinned at 397.818 for hundreds of
iterations while the dual infeasibility *oscillates upward*. It burns 39,379
objective evaluations to move nowhere.

Two separate questions here:

- Why does the monotone-µ path stall at a feasible point when adaptive-µ walks
  straight to a KKT point in 567 iterations?
- Should a stall this flat — objective unchanged to 8 digits over hundreds of
  iterations, primal feasible the whole way — be detected and terminated early
  rather than grinding to the iteration cap? `acceptable_iter=15` does not
  trip it, because dual infeasibility never reaches the acceptable tolerance.

The sibling `steenbrc` also hits the iteration cap (Ipopt returns
Unknown_Error there, so it is not counted above), while `steenbrb`, `steenbrd`,
`steenbre`, and `steenbrg` all solve cleanly. Whatever the trigger is, it is
specific within the family.

jkitchin:
Fixes #524 — both halves. `cresc4`'s false local-infeasibility exit is fixed by
the second-opinion ladder below; `steenbrf`'s stall turned out to be a separate
port defect (the watchdog's shortened-step counter surviving a successful
restoration) and is fixed in `327cc2fd`. There is also a corpus-entry finding
about `steenbrf` — a dropped index set — that is independent of both and is
worth reading before trusting that benchmark row.

Rebased on `5421971` (#525).

## Problem

CUTE `cresc4` — smallest-area crescent containing four points; 6 variables, 8
constraints, nonconvex, **feasible**, with a known optimum. POUNCE on defaults
reports `Infeasible_Problem_Detected`.

| | status | objective | constr viol | NLP error | iters |
|---|---|---|---|---|---|
| before | Infeasible_Problem_Detected | −9.483e−10 | 4.17e−01 | 1.18e+05 | 61 |
| after | Optimal Solution Found | 0.87189752729 | 4.90e−16 | 1.12e−09 | 81 |
| oracle (Ipopt-MA57) | Solve_Succeeded | 0.8718975393 | | | 71 |
| oracle (LOQO, SNOPT) | optimal | 0.8718976 | | | |

The objective at the bad exit is ~0 because the objective *is* the crescent's
area: the solver converges to a degenerate zero-area crescent, which cannot
contain the points, and then reports the feasible set empty.

A false *infeasible* is the dangerous verdict — a branch-and-bound driver
silently prunes a node that may hold the optimum.

## Cause

Not the infeasibility detector, and not restoration: the final restoration
converges properly (its own dual infeasibility → 2.91e−11 at a pinned
`inf_pr`), so it really did find a stationary point of the constraint
violation. The defect is in the guard that exists to catch exactly this.

A local-infeasibility verdict was second-guessed by one re-solve,
`feral_scaling=mc64`. That rung perturbs only the **linear algebra**. Measured
on the committed fixture, MC64 reproduces the failing trajectory
character-for-character through iteration 15 and diverges at iteration 16 in
the eighth significant digit:

```
orig  16  2.2741476e-08 2.98e+02 6.31e+05  -1.0 5.47e+02  1.6 1.59e-04 8.75e-05h  1
mc64  16  2.2739801e-08 2.98e+02 6.31e+05  -1.0 5.47e+02  1.6 1.59e-04 8.75e-05h  1
```

That is exactly the ULP-hypersensitivity signature the guard was written for
(`discs.nl`) — and it lands in the same basin anyway. Whether the perturbation
escapes is luck. On `discs.nl` it was lucky; here it was not, and the CLI
printed *"now corroborated by a second scaling"*, converting "I re-ran it and
nothing changed" into confidence. That message is the actual bug: it is the
step that turns a numerical failure into a reported fact.

`nlp_scaling_method=none` also solves `cresc4`, so a third scaling *does*
disagree — the two-way agreement was never about scaling at all.

A side finding worth knowing: **the verdict depends on the surface encoding.**
Six of twelve equally-valid AMPL encodings of the identical model fail.
Constraint declaration order and whether the radius term sits on the RHS each
flip it; variable declaration order does not (the `.nl` writer reorders
anyway). So on this problem POUNCE's answer to "is the feasible set empty"
depends on the order the user typed their constraints.

## Fix

The retry becomes a two-rung ladder:

| rung | option | varies |
|---|---|---|
| `feral_scaling=mc64` | `feral_infeasibility_scaling_retry` (existing) | the linear algebra |
| `mu_strategy=adaptive` | `infeasibility_mu_strategy_retry` (new, default on) | the barrier trajectory |

Three properties are deliberate:

- **Rungs are not cumulative.** Rung 2 re-asserts the baseline `feral_scaling`
  first. This is load-bearing, not tidiness: on the fixture
  `mu_strategy=adaptive` recovers the optimum and `mu_strategy=adaptive` *with*
  `feral_scaling=mc64` still reports local infeasibility. A stacked ladder
  would have thrown the fix away — and would have looked correct in a test that
  only checked the final status on the other encoding, where the stacked
  combination happens to work. The baseline tag is read from the **resolved**
  strategy, not the option string, so a `POUNCE_FERAL_SCALING`-configured run is
  restored rather than overridden.
- **Promotion is verified, not trusted.** A rung overturns the verdict only on
  `Solve_Succeeded` / `Solved_To_Acceptable_Level` — the retry's own KKT check.
  Adaptive µ is never assumed right; it is only allowed to produce a point that
  passes the same convergence test as everything else. So this cannot
  manufacture a false *optimal*, which would be worse than the bug being fixed.
- **A no-op rung is skipped**, so nothing is spent re-deriving an answer under
  settings the baseline already had.

Rejected: retuning the infeasibility detector. The point reached genuinely *is*
a stationary point of the constraint violation, so no threshold on the
stationarity surrogate separates it from a correct verdict — the same
conclusion `false_local_infeasibility.rs` already records for the HS13 case.
Retrying a different barrier strategy is also the remedy IPOPT's own docs give
a user who gets an infeasibility verdict on a problem they believe is feasible;
this just spares them the round trip.

## Blast radius

CLI-only (`crates/pounce-cli/src/main.rs` plus the option registration);
library embedders that inject their own restoration provider are unaffected, as
with the existing rung.

The ladder runs **only** on a solve that already ended in
`Infeasible_Problem_Detected`, so a successful solve is untouched and costs
nothing — now measured, not argued: a 733-problem Vanderbei sweep on the
reporter's machine shows no objective drift on any of the 700 problems both
solve, nothing broken, and `cresc4` as the only status change from the solver.

On the infeasible path the cost is real, and this body originally understated
it as "nothing". Two extra solves on a hard infeasible problem: `cresc132`
3.65s -> 73.4s (20x), `cresc100` 0.97 -> 5.95, `cresc50` 2.33 -> 4.56, suite
total 1430s -> 1523s (+7 %). See the follow-up comment for the full table and
why capping the rungs' `max_iter` is a separate change.

On runs that do reach the ladder:

- genuinely infeasible fixtures still report infeasible — the ladder runs out
  of rungs and keeps the original verdict (`infeasible_equalities.nl`,
  `issue_508_infeasible_gap_1em2/1em4.nl`, all still in AMPL band 200, 0.04–0.25 s);
- presolve-*certified* infeasibility stays exempt (a proof is not something
  either knob can overturn);
- the interactive debugger still suppresses the whole ladder.

Console output changed, which broke two tests that grepped for the old strings
— see Tests.

~~No corpus sweep: `$POUNCE_BENCH_DATA` is not available in this environment,
so the blast radius above is argued from the code path and the fixture corpus,
not measured over the corpus.~~ Done since, on the reporter's machine — see
the follow-up comment. It also turned up a benchmark-driver bug this PR
triggers (the suite mis-recorded `cresc100` as `Maximum_Iterations_Exceeded`),
fixed in the later commits along with a new machine-readable `Status:` line
from the CLI.

## Tests

`cargo test --release`: **193 suites, 2674 passed, 0 failed, 6 ignored** (run
pre-rebase; the three affected suites re-verified after rebasing onto #525).
`cargo fmt --all -- --check` clean. Clippy: no findings in any file this PR
touches; the workspace has ~150 pre-existing `collapsible_if`/`expect_used`
findings at rust 1.94 clippy, identical on the parent commit (verified by
running clippy on `a664dc0` in a separate worktree).

New, in `false_local_infeasibility.rs` alongside the HS13 cases:

- `cresc4_is_not_reported_infeasible` — the headline. **Fails on the parent
  commit for the stated reason**, verified by building `a664dc0` in a separate
  worktree and running it on the committed fixture: it returns
  `Converged to a point of local infeasibility`, then prints
  `MC64 re-solve did not recover … now corroborated by a second scaling`.
- `cresc4_needs_the_barrier_rung_not_the_scaling_rung` — pins the *mechanism*,
  so a future change that collapses the ladder back to one MC64 rung says so
  directly instead of leaving the headline test to fail unexplained.
- `ladder_rungs_are_skipped_when_the_baseline_already_has_them`.

Plus 5 unit tests on ladder construction (ordering, the baseline-scaling reset,
per-rung opt-out, no-op skipping, and the untagged-scaling case).

Two existing guards were re-anchored — worth reading, because they failed in
opposite ways:

- `issue_508_…::the_last_exit_banner_matches_the_sol_after_a_non_promoted_…`
  **failed loudly**, which is what it was designed to do: its first assertion
  exists to fail when the fixture stops exercising the non-promoted retry path.
- `presolve_certified_infeasibility::certified_infeasibility_skips_…` is the
  one that mattered. It asserts `!contains("re-solving once with MC64")`.
  Deleting that string from the CLI made the negative assertion **vacuously
  true** — it would have kept passing while guarding nothing, with no test
  failure to say so. Found by grepping for the strings I had changed, not by
  the suite. Both now key off the ladder's own announcement/outcome lines
  rather than any one rung's message, so adding or renaming rungs cannot
  silently defuse them again.

**Deliberately not pinned:** the fixture is transcribed from CUTEst
`mastsif/CRESC4.SIF` and generated by the committed `cresc4.py`; it is *not*
byte-identical to the corpus's AMPL-generated `cresc4.nl`, which was not
reachable from this environment (`vanderbei.princeton.edu` is blocked by the
proxy and the issue carries no `cresc4` attachment). It reproduces the reported
signature — same status, near-zero objective, constraint violation near a half,
recovery under both `mu_strategy=adaptive` and `nlp_scaling_method=none` — but
exits at 61 iterations where the corpus file exits at 74. The generator is
committed so the transcription can be audited against the SIF rather than
taken on trust.

## ~~`steenbrf`: not a solver defect~~ — withdrawn, and now fixed

**This section's conclusion is wrong and is retracted; see the follow-up
comments.** It was written in a container where the default path happens to
converge at 2570 iterations. On the reporter's machine, on the same bytes at
merged `main`, POUNCE hits the 3000 cap — and `max_iter=6000` fails at 3039
with `Restoration Failed`, so the "430 iterations of margin" is not margin.

There is a real solver defect, and it is now fixed in `327cc2fd`: a successful
restoration phase left the watchdog's consecutive-shortened-step counter
running (upstream clears it at `IpBacktrackingLineSearch.cpp:624-631`), so runs
of shortened steps on either side of a restoration accumulated as if
consecutive and armed the watchdog where Ipopt never does. `steenbrf` now
solves — `Solve_Succeeded` in **481** iterations at objective 282.678, better
than the Ipopt-MA57 reference's acceptable-level 1321.65 at 1846. The corpus
sweep also gains `brainpc2` and loses 7.3 % of its total iterations.

A second reading is withdrawn along with the first: this is **not** "the
`cresc4` shape". The 62 restoration episodes were the symptom of the watchdog
wrecking the line search, not a shared root cause. `cresc4`'s "Also not in this
PR" item is unaffected.

What survives below: the corpus-entry findings (the dropped index set, the
STEENBRB control). Those were always about the benchmark, not the solver. The
claim to read past is "It does not stall here."

Original text follows.

### `steenbrf`: not a solver defect (as filed)

Worked against the reporter's own file (uploaded into the container; `sha256
bba26942…78d0ab`, matching the issue comment's checksum). No code change; two
findings.

**It does not stall here.** At the reporter's own commit, on the byte-identical
file, POUNCE converges: `Optimal`, **2570 iterations**, objective 282.678 — 430
short of the cap the reporter hit. Not noise (three runs give 2570 exactly;
`RAYON_NUM_THREADS=1` and `=4` both give 2570) and not this branch (the parent
binary is bit-identical). The adaptive column *does* match the reporter to the
iteration, 567, which is what says the file and setup are right. So the
trajectory is platform-dependent and the margin to `max_iter` is ~15%.

**Why it crawls: 360 of its 468 variables carry no objective at all.** Straight
off the `.nl` header — `steenbrb` has 468 objective-gradient nonzeros,
`steenbrf` has 108 — and the `.col` map names them: the 36 capacities plus the
flows of **commodities 11 and 12 only**. CUTEst `STEENBRF` sums all twelve, so
Vanderbei's `.mod` looks like a dropped index set. Those 360 variables enter the
barrier problem only through their `log` terms, with zero objective gradient and
zero Hessian contribution, so monotone µ must walk all of them down on schedule
with nothing pulling them — exactly the trace the issue quotes (objective pinned
to eight digits, feasible throughout, 20+ backtracks per iteration).

The controlled comparison — same network, data and start point, differing only
in whether the other ten commodities are priced:

| model | congestion term | iterations (defaults) |
|---|---|---|
| CUTEst `STEENBRF` (transcribed) | all 12 commodities | **53** |
| Vanderbei `steenbrf` (reporter's file) | commodities 11–12 only | **2570** |

This also makes "different model" a measurement rather than an argument:
dropping the non-negative congestion term from the CUTEst model leaves a
min-cost multicommodity-flow LP that POUNCE solves at **8250.0**, so that model
cannot reach 282.678 at all — it is not a second local minimum.
`dev-notes/issue-524/floor_bound.py` computes it. Vanderbei's `steenbrb.nl`,
also supplied, confirms the control outright: his file and the transcription
solve to the same objective in the same 49 iterations, so the machinery is
sound and `steenbrb` is *not* affected by the defect.

POUNCE's answer here agrees with the best reference (nitro's 282.7578) to four
digits. What is worth doing is checking the corpus entry — `steenbrf` is also
one of the few table rows with `solvers_agree: false`, which a large cost-free
subspace would explain — not changing the solver.

## Also not in this PR

- The underlying `cresc4` question: why monotone µ lets the main loop trade
  constraint violation for a degenerate crescent, undoing restoration's
  progress across six restoration entries. The ladder is a second opinion, not
  a cure. Six variables, so the trajectory is fully inspectable.
- The one robustness item `steenbrf` does raise: a *post-hoc* diagnostic on the
  `max_iter` + frozen-µ exit that says "this looks like a degenerate crawl, try
  `mu_strategy=adaptive`". `dev-notes/issue-131-monotone-lbfgs-stall.md`
  establishes that the mid-solve version is an unprincipled patience knob; the
  post-hoc version is unbuilt and would have turned the reporter's
  3000-iteration wall into a one-line answer.
- Pre-existing: on a non-promoting ladder the `.sol` carries the last rung's
  `x` paired with the original verdict and statistics. True before this change
  too; adding a rung only changes which failing point it is.

---

- [x] Book page under `docs/src/` updated (`troubleshooting.md`; existing page,
      no `SUMMARY.md` change needed)
- [x] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean — clippy
      caveat stated above (pre-existing workspace findings, none in touched files)
      as it stands now


jkitchin:
Reviewed with the corpus available, so this covers the two gaps the PR names:
the real `cresc4.nl` (the transcribed fixture's counterpart) and the sweep.
Built `67b14e9b` release/FERAL on Darwin 25.5.0 arm64; baseline is
`54219714` (`main`, post-#525), same host.

## The `cresc4` half checks out, on the corpus file

The PR could not reach `vanderbei.princeton.edu` and flags the fixture as not
byte-identical to the corpus's `cresc4.nl`. Run against the corpus file itself:

```
rung 1  feral_scaling=mc64      → InfeasibleProblemDetected (no recovery)
rung 2  mu_strategy=adaptive    → Optimal Solution Found, obj 0.87189752899987727  ← promoted
```

against `54219714`, which still reports `Infeasible_Problem_Detected`. That
objective matches the committed Ipopt-MA57 reference (`0.8718975393`) to 9
significant figures. The ladder fixes the problem as it actually appears in the
corpus, and the fixture is a faithful stand-in for it.

**Vanderbei sweep** (733 problems, PR binary vs baseline, same host):

```
702/733 solved (baseline 701/733) | FIXED=1 BROKE=0 obj-drift=0
  FIXED  cresc4: Infeasible_Problem_Detected -> Solve_Succeeded
```

No regressions, no objective drift on the 700 solved by both. The genuinely
infeasible entries (`cresc50`, `cresc132`, `launch`, and `cresc100` — see below)
all still reach the infeasible verdict, so the ladder is running out of rungs
and keeping the verdict as designed.

## 1. The per-rung banners break log scrapers — including this repo's own

The sweep recorded `cresc100` as `Maximum_Iterations_Exceeded`, where the
baseline had `Infeasible_Problem_Detected`. **The solver is right and the
harness is wrong** — on the console the final verdict is correct:

```
pounce: second opinion — re-solving with mu_strategy=adaptive…
EXIT: Maximum Number of Iterations Exceeded.            ← rung 2's own banner
pounce: mu_strategy=adaptive re-solve did not recover (MaximumIterationsExceeded).
pounce: keeping the original local-infeasibility verdict; it survived 2 independent re-solve(s).
EXIT: Converged to a point of local infeasibility.      ← the actual verdict
```

`benchmarks/scripts/run_nl_bench.sh` greps the whole log in a fixed priority
order and returns on first match — `"Maximum Number of Iterations Exceeded"` at
line 92, `"Converged to a point of local infeasibility"` at line 94. Rung 2's
banner matches first, so the scrape never reaches the line that would be right.

The driver bug is pre-existing, but this PR is what triggers it: a rung that
fails *differently* from the final verdict is new. On `main` the single MC64
rung reproduces the same failure mode, so first-match happened to be
harmless.

Two reasons this is worth fixing before merge rather than after:

- **It is silent.** A wrong status, no error, no warning. It corrupts
  `BENCHMARK_REPORT.md` — the artifact this project uses to check itself — in
  the direction of *understating* a correct infeasibility verdict as an
  iteration-limit failure.
- **The driver already has the right mechanism and it is not being fed.**
  `pounce_status_from_log` first tries `grep -oE '^[Ss]tatus:[[:space:]]+\w+' |
  tail -1` — a machine-readable, last-one-wins line — and only falls back to
  the free-form priority scrape. These runs emit no `Status:` line at all
  (verified: zero matches on `cresc100`), so the fallback is what runs. Emitting
  one final `Status:` line would fix this class outright and is strictly better
  than reordering greps, which only patches the instance.

Either side is a small change: emit the machine-readable final status, suppress
per-rung `EXIT:` banners (keep the `pounce:` narration), or key the driver off
the last `POUNCE 0.9.0:` line. Only `cresc100` is affected in this corpus, but
any downstream consumer of pounce's stdout has the same exposure.

## 2. The infeasible path costs more than the blast radius says

"Bit-identical except the targeted case" is right for successful solves — those
never enter the ladder and I measured no drift on 700 of them. The other side
is unquantified in the PR:

| problem | baseline | this PR |
|---|---|---|
| `cresc132` | 3.65s | **73.84s** (20×) |
| `cresc100` | 0.97s | 6.05s |
| `cresc50` | 2.33s | 4.74s |
| `launch` | 0.10s | 0.25s |
| `cresc4` | 0.06s | 0.08s |
| vanderbei suite | 1430s | 1511s (+6%) |

Two extra solves on a hard infeasible problem is inherent to the design, and the
suite-level +6% is acceptable. But 20× on `cresc132` is the number a user with a
branch-and-bound driver pruning many infeasible nodes will feel, and it belongs
in the blast-radius section. Worth considering whether rung 2 should inherit a
reduced `max_iter`, given that a rung only ever counts when it *converges*.

## 3. The `steenbrf` section does not hold up on the reporter's platform

I am the reporter; this is measured on the same host and file as the issue
(`sha256 bba26942…78d0ab`).

**It stalls here, and the margin the PR infers is not there.** At `54219714`:
3000 iterations, `Maximum_Iterations_Exceeded` — not 2570/Optimal. Not
threading (`RAYON_NUM_THREADS=1` gives 3000 too). And raising the cap does not
rescue it: `max_iter=6000` fails at iteration **3039 with Restoration Failed**.
So on this platform there is no 15% margin to a cap — the trajectory ends in
restoration failure. The PR's platform-dependence observation is real, but the
conclusion drawn from it ("430 short of the cap") is the *lucky* side of that
dependence.

**The µ trajectory says solver, not model.** Both solvers walk µ down
identically through six barrier levels, then diverge at the seventh:

| lg(µ) | 3.3 | 2.6 | 1.9 | −1.0 | −1.7 | −2.5 | **−3.8** | −5.7 | −8.6 |
|---|---|---|---|---|---|---|---|---|---|
| Ipopt-MA57 | 2 | 5 | 3 | 43 | 133 | 422 | **29** | 18 | 1192 |
| POUNCE | 2 | 5 | 3 | 37 | 78 | 242 | **2696** | — | — |

POUNCE reaches lg(µ) = −3.8 at iteration 366 and never leaves it: 2696 of 3000
iterations at one barrier level. Ipopt clears that level in 29. They enter it in
near-identical states —

```
ipopt   608  5.1501578e+02 1.48e-07 3.23e+03  -3.8 2.69e+01  -3.0 8.76e-01 1.48e-01f  1
pounce  366  3.6948567e+02 1.52e-07 3.81e+03  -3.8 4.27e-01  -1.4 1.00e+00 1.49e-01f  1
```

— and then part two iterations later, where POUNCE takes unregularized steps of
norm 5.4e+03 and 6.9e+03 (`-` in the `lg(rg)` column) that raise the objective
from 369 to 393. Ipopt's largest step anywhere in that phase is 3.0e+02. Ipopt
drives `inf_du` to 1.85e-03 and drops µ; POUNCE bottoms out at 6.0e+00, never
satisfies the µ-update test, and enters restoration 92 times over the run.

**The structural finding is correct but does not carry the conclusion.** The
missing objective terms check out directly from the `.nl` headers — `steenbrb`
has 468 objective-gradient nonzeros, `steenbrf` has 108, so 360 of 468 variables
are unpriced. That is a real corpus defect and explains why the problem is hard.
It cannot explain the stall, because Ipopt solves *that same file*, with the same
360 unpriced variables, and its µ advances.

So: the corpus-entry finding is worth keeping and reporting upstream, and the
post-hoc "degenerate crawl, try `mu_strategy=adaptive`" diagnostic listed under
"Also not in this PR" looks like the right follow-up. But "not a solver defect"
is stronger than the evidence supports, and #524 should not be closed on it.

## Recommendation

The `cresc4` half is sound — mechanism is well argued, the non-cumulative rung
ordering is load-bearing and correctly justified, promotion is gated on the
retry's own KKT check so it cannot manufacture a false optimal, and it is now
confirmed on the real corpus file with a clean 733-problem sweep behind it.

I would fix item 1 before merging: it is a few lines, and left alone it silently
degrades the benchmark data this project uses to check itself. Item 2 is a
documentation fix to the blast radius. Item 3 is about what #524 gets closed
on, not about this PR's code.

jkitchin:
Picked this up on the reporter's machine, where `$POUNCE_BENCH_DATA`, the
committed Ipopt-MA57 reference logs and a real build are all available — the
three things the container review flagged as unverifiable. Five commits. **Two claims in the PR body above are now wrong and are
corrected here**; I'll edit the body itself rather than leave them standing.

## 1. The log-scrape hazard — fixed at both ends (`9d3c1768`, `09641663`, `a386477d`)

Confirmed real, on the real corpus. `benchmarks/scripts/run_nl_bench.sh`
ranks the phrases it greps for, so on a ladder run it reads whichever banner
ranks highest anywhere in the log rather than the one the solve exited with.
`cresc100` is the case: the rungs' own `EXIT: Maximum Number of Iterations
Exceeded` outranks the final `EXIT: Converged to a point of local
infeasibility`, and the suite records `Maximum_Iterations_Exceeded` for a
problem POUNCE correctly called infeasible. Silent, and it lands in the
committed `BENCHMARK_REPORT`.

Fixed twice, on purpose:

**The CLI now prints its status.** `pounce` emits one machine-readable line at
the end of a solve —

```
Status: Solve_Succeeded
```

— using the upstream C enumerator spelling, from a new
`ApplicationReturnStatus::upstream_name()`. It is gated on `print_level >= 1`
and off in `--json` mode. Three tests pin it, including one that asserts
`upstream_name().replace('_', "") == format!("{status:?}")` across all 20
variants, so a new status cannot be added with a name that drifts from the
enum. This is the durable fix: any parser, ours or a user's, now has a
contract instead of prose to scrape.

**The driver no longer needs it.** `pounce_status_from_log` became three
tiers: `Status:` line if present, else the **last** `EXIT:` banner (which gh
#508 already made trustworthy — the shipped banner is re-emitted when a retry
is not promoted), else the existing phrase greps. So the benchmark corpus
reads correctly against older binaries too. `ipopt_status_from_log` untouched.

The banner table for that middle tier is now complete: all 20 messages, each
round-tripped against `status_message()` in
`crates/pounce-solve-report/src/console.rs` — feed it any variant's message
and you get that variant's `upstream_name()` back. It had ten arms before,
and its `Diverging Iterates` arm matched nothing, since the banner actually
reads `Iterates diverging; problem might be unbounded.`

Verified at all three tiers on the real `cresc100` log — with the `Status:`
line, with it stripped, and with both it and the last `EXIT:` stripped — all
three return `Infeasible_Problem_Detected`.

## 2. The ladder's cost — measured, and it is not zero (`98779468`, `9b08f12e`)

The body above says "bit-identical except the targeted case … a successful
solve is untouched and costs nothing". The first half is now measured and
holds; the second half was silent about the infeasible path, where the cost is
real. From a 733-problem Vanderbei sweep, PR binary vs merged `54219714`, same
host:

| problem | `54219714` | with the ladder |
|---|---|---|
| `cresc132` | 3.65 s | **73.4 s** (20×) |
| `cresc100` | 0.97 s | 5.95 s |
| `cresc50` | 2.33 s | 4.56 s |
| `launch` | 0.10 s | 0.22 s |
| `cresc4` | 0.06 s | 0.07 s (and now solves) |
| suite total | 1430 s | 1523 s (+7 %) |

No objective drift on any of the 700 problems both solve, nothing broken, and
exactly one status change from the solver — `cresc4` — so the "bit-identical
on success" claim is now a measurement rather than an argument from the code
path. (Four further rows change, none of them the solver: `indef`/`static3`
`Solver_Error` → `Diverging_Iterates` and `grouping`/`lewispol`
`Solver_Error` → `Not_Enough_Degrees_Of_Freedom`, all four the driver fix in
item 1 now classifying logs it previously could not.) The 20× on `cresc132`
is the number a branch-and-bound driver pruning many infeasible nodes would
feel: the barrier rung runs to `max_iter` there and contributes nothing,
because a rung only counts when it *converges*. Capping the rungs' `max_iter`
would bound that, at the cost of declining recoveries that genuinely need the
iterations — a separate change needing its own sweep, not folded in here.

**`cresc4` is also now confirmed on the corpus file itself**, not just the
committed transcription: rung 2 returns Optimal at `0.87189752899987727`,
matching the Ipopt-MA57 reference `0.8718975393` to nine significant figures.
Merged `main` still reports local infeasibility on that same file.

## 3. `steenbrf` — the "not a solver defect" conclusion is withdrawn

This is the one the body above gets wrong. On the reporter's machine, on the
byte-identical file, at merged `main`:

| | container, `a664dc0` | reporter's machine, `54219714` |
|---|---|---|
| defaults | Optimal, 2570 iters | **Maximum_Iterations_Exceeded, 3000** |
| defaults, `max_iter=6000` | — | **Restoration_Failed at 3039** |
| `mu_strategy=adaptive` | Acceptable, 567 | Acceptable, 567 |

So it *does* stall, and the "430 iterations of margin" the container seemed to
show is not margin — raising the cap does not let the default path finish.
Same commit, same bytes, two answers: the trajectory is platform dependent,
and neither side is run-to-run noise.

What the trajectory shows. Counting non-restoration iterations per barrier
level against the committed reference log:

| lg(µ) | −1.0 | −1.7 | −2.5 | **−3.8** | −5.7 | −8.6 |
|---|---|---|---|---|---|---|
| Ipopt-MA57 | 43 | 133 | 422 | **29** | 18 | 1192 |
| POUNCE, defaults | 37 | 78 | 242 | **2360** | — | — |

They enter −3.8 in near-identical states; Ipopt clears it in 29 iterations,
POUNCE never leaves. And it is not a crawl — `inf_du` reaches 6.2e-03 at
`inf_pr` = 2e-08, one short step from the µ-update test, restoration is
entered, and it returns at 3.4e+03. That cycles for 2600 iterations; the
objective swings between 395 and 436, it is not pinned.

The discriminator is restoration *episodes*:

| run | iters | restoration iters | **episodes** | outcome |
|---|---|---|---|---|
| POUNCE, defaults | 3000 | 346 | **62** | Maximum_Iterations_Exceeded |
| POUNCE, `mu_strategy=adaptive` | 567 | 10 | **1** | Acceptable |
| Ipopt-MA57 | 1846 | 10 | **1** | Acceptable |

Adaptive µ and Ipopt have identical restoration profiles. That is what demotes
the 360-cost-free-variables finding from cause to context: it explains why the
model is hard, but both solvers face it and only one stalls, so it cannot
explain the failure.

This is the `cresc4` shape — restoration progress the main loop does not hold
on to — at 468 variables instead of 6, which makes the issue's original
shared-cause hypothesis look right after all. **#524 should not be closed on
`steenbrf`.** The corpus-entry findings (the dropped index set, the STEENBRB
control) stand unchanged; they were always about the benchmark, not the solver.

Also withdrawn, from my own review comment on this PR: that POUNCE takes large
unregularized steps Ipopt does not. Ipopt takes larger ones here (7.9e+04,
1.1e+05). Not the discriminator.

## Verification

- `cargo test --workspace --exclude pounce-hsl`: 200 suites ok, 0 failures.
  (`pounce-hsl` fails to *link* on this host — no HSL library — unrelated to
  this branch.)
- `cargo fmt --all -- --check` clean; clippy clean on the new code.
- 733-problem Vanderbei sweep re-run against this branch's binary **and** this
  branch's driver, diffed against both merged `main` and this PR as filed:
  `cresc4` FIXED, nothing BROKE, no objective drift, and the spurious
  `cresc100 → Maximum_Iterations_Exceeded` from item 1 is gone.
- Live: `cresc4` → `Status: Solve_Succeeded`, `cresc100` →
  `Status: Infeasible_Problem_Detected`.

jkitchin:
## `steenbrf` is fixed, and it was a port defect

Follow-up to my last comment, which said the solver defect on `steenbrf` was
real but unfixed and wanted its own issue. It has a cause and a fix, and both
are narrower than that comment guessed. One commit, `327cc2fd`.

**The defect.** The watchdog arms after `watchdog_shortened_iter_trigger`
(default 10) *consecutive* shortened line-search steps. Upstream zeroes that
counter when the restoration phase succeeds —
`IpBacktrackingLineSearch.cpp:624-631`, alongside `in_soft_resto_phase_` and
`soft_resto_counter_` — because an iterate returned by restoration is a
different point, so a run of shortened steps does not continue through it.

POUNCE had no equivalent, and the reason is structural rather than a misread.
Upstream calls `PerformRestoration()` from *inside* `FindAcceptableTrialPoint`,
so those four assignments sit in scope immediately after it. POUNCE returns
`Outcome::Failed` and lets `IpoptAlgorithm::invoke_restoration` run
restoration — so there was no site where that code could land, and nothing on
the recovery path told the line search a restoration had happened.

**What it does on this problem.** Straight off the log:

| iters | what | counter |
|---|---|---|
| 381–385 | five shortened steps (`ls` = 11, 12, 13, 15, 16) | 1 → 5 |
| 386–390 | soft restoration (`s`) — neither side touches the counter | 5 |
| 391–394 | full restoration, succeeds | upstream → **0**; POUNCE → **5** |
| 399–403 | five more shortened steps | upstream 5; POUNCE → **10 = trigger** |
| 404 | — | POUNCE arms the watchdog |

```
 403  3.6779355e+02 1.22e-08 4.61e+02  -3.8 1.27e-03  4.1 1.00e+00 1.53e-05f 17
 404  3.6744474e+02 3.68e-07 9.11e+01  -3.8 2.29e-03  3.7 1.00e+00 1.00e+00w  1
 405  3.6709336e+02 1.23e-07 1.87e+02  -3.8 2.86e-03  3.2 1.00e+00 9.88e-01w  1
 406  3.6675112e+02 1.35e-07 1.97e+02  -3.8 4.23e-03  2.7 1.00e+00 9.30e-01w  1
 407  3.6779355e+02 1.22e-08 4.72e+02  -3.8 7.70e-03  2.2 1.00e+00 4.77e-07f 21
 408  3.6779355e+02 1.22e-08 4.89e+04  -3.8 8.68e-02  1.8 1.00e+00 4.04e-08f 20
 409  3.6779355e+02 1.22e-08 1.08e+04  -3.8 1.27e-01  1.3 7.83e-01 2.91e-08f 23
```

404–406 are exactly `watchdog_trial_iter_max = 3` trial iterations. Iteration
407's objective and `inf_pr` are bit-identical to 403's — that is `StopWatchDog`
reverting to the snapshot. The three `w` steps bought nothing, and coming out of
the revert the line search backtracks 20+ times to `alpha` ~1e-08. That cycle
ran 105 times until `max_iter`.

Two independent confirmations that this is the mechanism and not a coincidence:

- Ipopt-MA57's longest run of consecutive shortened steps on the same file is
  **6**. It never reaches the trigger — exactly what the missing reset predicts.
- Ablation, before writing any code: `soft_resto_pderror_reduction_factor=0`
  changes nothing (still `Maximum_Iterations_Exceeded` at 3000), while
  `watchdog_shortened_iter_trigger=0` alone gives `Solve_Succeeded` in 481. The
  two runs are iteration-identical through 403 and diverge exactly at the arming
  point.

**Result on `steenbrf`:**

| | before | after | Ipopt-MA57 ref |
|---|---|---|---|
| status | `Maximum_Iterations_Exceeded` | **`Solve_Succeeded`** | `Solved To Acceptable Level` |
| iterations | 3000 | **481** | 1846 |
| objective (unscaled) | 397.82, not converged | **282.678** | 1321.65 |
| constr. violation | — | 4.4e-11 | — |
| `w` / `s` markers | 334 / 472 | **0 / 12** | 0 / 0 |

So it now converges to full tolerance at a *lower* local minimum than the
reference reaches at acceptable level.

**Corpus.** 733-problem Vanderbei sweep, same host, this branch's binary and
driver:

- **2 fixed** — `steenbrf`, and `brainpc2` (`Maximum_Iterations_Exceeded` at
  3000 → `Solved_To_Acceptable_Level` at 1003), which nobody was looking at.
- **0 broken**, and no objective drift above 1e-6 relative on any problem both
  runs solve.
- Total solve time **−6.5 %**, total iterations **−7.3 %**.

That last line is the tell that this was a defect and not a tuning preference: a
heuristic firing when it should not costs iterations everywhere, quietly.
`pfit3`, `pfit4` and `scon1dls` — the three fixtures the watchdog comments cite
— are all in this suite and are unchanged, iteration for iteration.

**Withdrawn from my previous comment:** that `steenbrf` is "the `cresc4` shape —
restoration progress the main loop does not hold on to" and wants a shared
issue. The repeated restoration entry was the symptom; restoration kept being
re-entered because the watchdog kept wrecking the line search. Both problems do
respond to `mu_strategy=adaptive`, which is what made the shared-cause reading
attractive, but a heuristic that only fires under the monotone schedule produces
that signature without sharing anything else. `cresc4`'s "Not fixed here" item
stands on its own, unaffected.

**#524 can now close on both halves.**

## Verification

- `cargo test --workspace --exclude pounce-hsl`: 0 failures. (`pounce-hsl` fails
  to *link* on this host for want of an HSL library — unrelated to this branch.)
- New unit test `restoration_resets_the_shortened_iter_counter` pins the
  invariant with steenbrf's own five-restoration-five arithmetic.
- `cargo fmt --all -- --check` clean; no new clippy warnings in the two changed
  files.

Issue #486, opened 2026-08-05
Variable scaling in the core, and the remaining registered-but-unread options (#483 follow-up)


jkitchin:
Tracking issue for what remains after #483 / #485.

#483 asked for user scaling in three stages. **Stage 1 shipped in #485**: the `scaling_factor` suffix reaches the solver on both paths, per-variable factors are refused rather than discarded, and the sensitivity accessors are proven unaffected. Auditing it turned up five sibling defects, also fixed there.

Stage 1 makes everything below **safely deferrable**: the rejections are in place, so nobody is silently getting conditioning they did not ask for while these decisions are made. There is no pressure to pick the locally cheaper option.

---

## 1. Stage 2 — the core models variable scaling  ✅ shipped in #529

#483 asked for design input before code. Here is the fork, with the evidence behind it.

### Where the change of variables lives

**A — inside `OrigIpoptNlp`**, beside `obj_scale_factor` / `c_scale` / `d_scale`.

*For:* matches upstream (`NLPScalingObject` applied in `OrigIpoptNLP`), which matters in a codebase that is explicitly a port; future ports line up. `natural_units_conj` already lives here and already translates `df`/`dc`/`dd`, so stage 3 extends an understood mechanism instead of inventing a boundary.

*Against:* it touches the iterate on every path — each `eval_*`, `get_starting_point`, `lift_x_to_full`, `finalize_solution_{x,z_l,z_u}`, `relax_bounds`. Each is an opportunity to miss one, and a missed path is *silently* wrong conditioning — the exact defect #483 is about. It also collides with #485's `declared_x_l`/`declared_x_u` snapshots and the `honor_original_bounds` projection, which would become snapshots in scaled space.

**B — inside `TNLPAdapter`.** Rejected: the adapter is shared with the SQP engine, its job is structure (splitting, compressing) rather than units, and it has no compensating benefit over A or C.

**C — a `ScalingTnlp` wrapper, mirroring `PresolveTnlp`.** ← recommended

*For:*
- One transform path, and a **loud** failure mode. If the wrapper is wrong everything is wrong obviously, rather than one forgotten branch being quietly unscaled.
- Diagonal scaling leaves sparsity **unchanged**, so every transform is an elementwise multiply: `grad_f[i] /= d[i]`, `jac[k] /= d[col(k)]`, `h[k] /= d[row]·d[col]`, bounds and start `*= d`, and at finalize `x /= d`, bound duals `*= d`, λ unchanged.
[...]

Issue #528, opened 2026-08-07
IPM exits Search_Direction_Becomes_Too_Small on feasible, BOUNDED LPs once data magnitude reaches ~1e7 (scipy/HiGHS solves them)


jkitchin:
## Summary

On plain LPs with O(1) coefficients but right-hand sides of magnitude ~1e7 or
larger, POUNCE's IPM exits with Ipopt code **3 (`Search_Direction_Becomes_Too_Small`)**
on problems that are feasible and bounded, and that scipy's HiGHS solves to a
finite optimum without difficulty. **9 of 48** runs in the sweep below.

Nothing here is a wrong *answer* — code 3 is an honest "I stalled", not a false
certificate. It is a robustness/conditioning gap: the LPs are well posed (O(1)
matrix entries, a nonempty feasible set, every variable inside a finite box), so
the only unusual property is the magnitude of `b` and of the upper bound.

Reported from downstream discopt work (jkitchin/discopt#940), but the
reproduction below is standalone — `pounce` + `scipy` only, no discopt.

## Reproduction

```python
import numpy as np, pounce
from scipy.optimize import linprog

_INF = 1e20

class LP:
    """cyipopt-style callbacks for min c'x with constant Jacobian A."""
    def __init__(self, c, A):
        self.c, self.A = c, A
        m, n = A.shape
        self.jac = A.ravel().astype(float)
        r, cc = np.meshgrid(np.arange(m), np.arange(n), indexing="ij")
        self.rows, self.cols = r.ravel(), cc.ravel()
    def objective(self, x):   return float(self.c @ x)
    def gradient(self, x):    return self.c
    def constraints(self, x): return np.asarray(self.A @ x, dtype=float)
    def jacobian(self, x):    return self.jac
    def jacobianstructure(self):  return self.rows, self.cols
    def hessian(self, x, lagrange, obj_factor): return np.empty(0)
    def hessianstructure(self):
        return np.empty(0, dtype=np.intp), np.empty(0, dtype=np.intp)

seed, n, m, scale = 0, 20, 10, 1e7
rng = np.random.default_rng(seed)
A = np.abs(rng.uniform(0.5, 5.0, size=(m, n)))          # A x >= b
b = scale * np.abs(rng.uniform(0.5, 2.0, size=m)) * n * 0.25
c = np.abs(rng.uniform(1.0, 10.0, size=n))
hi = 10.0 * scale

[...]

Issue #527, opened 2026-08-07
pounce-convex presolve: the "fixpoint" never converges on bore3d — it always exits on MAX_ROUNDS, and the cap silently sets how much reduction you get


jkitchin:
Follow-up from #523 / #525. That PR fixed the correctness bug (a near-forcing row pinning on rounding-noise boxes). This is the mechanism that *set it up*, left deliberately out of scope there so a performance change would not ride along with a correctness fix.

`presolve` documents itself as "iterating the reduction passes to a **fixpoint**". On `bore3d` it does not reach one. It exits on the `MAX_ROUNDS` cap, every time, at any cap.

## Measured on `main` (`5421971`), netlib `bore3d`, release + FERAL

```
MAX_ROUNDS=32   EXIT: HIT MAX_ROUNDS cap at chain.len()=32   Reduced -> n=128 m_eq=75 m_ineq=2
MAX_ROUNDS=64   EXIT: HIT MAX_ROUNDS cap at chain.len()=64   Reduced -> n=128 m_eq=75 m_ineq=2
MAX_ROUNDS=200  EXIT: HIT MAX_ROUNDS cap at chain.len()=200  Reduced -> n=128 m_eq=75 m_ineq=2
```

Never `EXIT: fixpoint`. The loop is chasing a geometric sequence: two equality rows mutually imply ever-tighter upper bounds on the same group of nonnegative variables, each round shrinking them by a factor of ~26 toward their true limit of zero, which they approach but never reach.

Structural progress stops long before the cap does. Per-round dimensions at the default cap:

| rounds | n / m_eq / m_ineq | what happened |
|---|---|---|
| 0–2 | 315/214/19 → 136/79/4 | the real reductions |
| 3–22 | 136/79/4 | **nothing but bound shaving** |
| 22→23 | → 128/75/2 | boxes finally tight enough for 8 more eliminations |
| 24–28 | 128/75/2 | nothing but bound shaving, until the cap cuts it off |

Cost, presolve alone (mean of 20 calls):

| cap | presolve | reduced size |
|---|---|---|
| 4 | 1.2 ms | 136/79/4 |
| 8 | 2.7 ms | 136/79/4 |
| 32 (default) | 10.5 ms | 128/75/2 |
| 200 | 64.3 ms | 128/75/2 |

Linear in the cap, and past 32 it buys nothing at all on this model. For scale, the whole `bore3d` solve is ~21 ms — so presolve is about half of it, and most of that half is the cascade.

[...]

PR #529, opened 2026-08-07
Apply per-variable scaling as a change of variables (#486 stage 2)

devin-griff:
Stage 2 of #486, after #485 delivered the objective and constraint halves. Stage 3, which carries the factors into the sensitivity translation, is untouched.

**What it does.** `nlp_scaling_method=user-scaling` now applies per-variable `scaling_factor` entries. A new `ScalingTnlp` wraps the user's problem and presents `x̃ = d·x` upward, mirroring how `PresolveTnlp` transforms coordinates and inverts at `finalize_solution`. The substitution is diagonal, so sparsity is untouched and every transform is an elementwise multiply or divide: the starting point scales, bounds scale where present, the objective gradient and Jacobian columns divide, Hessian entries divide by both their row's and their column's factor, and on the way out the solution divides back while bound multipliers multiply and constraint multipliers stay put. That multiplier rule is derived from stationarity in the module docs rather than asserted.

Option (b) of the fork you wrote up, with the four open decisions resolved as: `tol` keeps comparing scaled quantities, matching upstream; parity means agreement on the solution and unscaled duals rather than iterate for iterate; negative and zero factors raise, since a sign flip reverses a variable and swaps its bounds; and the wrapper is installed at `optimize_tnlp`, so with presolve in play it sits outermost and the factors it reads are already in reduced variable indices. That is consistent rather than accidental: `linear_eq_elim` maps each surviving column's factor through `vars_kept`, the row-only reducer leaves columns alone, and `finalize_solution` unscales in those same reduced indices before the presolve layer beneath expands back.

**Installed at `optimize_tnlp`**, the single funnel every path passes through, so the CLI, the Python bindings and the batch solvers inherit it from one insertion, and only under `user-scaling`. An unscaled solve pays nothing.

**Absent bounds are the subtle case.** A variable with no bound arrives carrying the `nlp_lower_bound_inf` / `nlp_upper_bound_inf` sentinel, which is an ordinary finite number and not an infinity, so scaling it by a factor below 1 moves it inside the threshold and hands a free variable a barrier term and multipliers it never declared. Bounds are scaled only where present, using the directional test `classify_bounds` already uses. The mirror case is the same defect pointing the other way: a present bound scaled past a threshold reads as absent and disappears. `get_bounds_info` returns a bare bool and cannot explain itself, so that one is checked once at setup and refused naming the threshold it crossed.

**One finding that bears on stage 3.** Choosing the wrapper over scaling inside `OrigIpoptNlp` moved where the consumer audit is needed rather than removing it. Anything reading the algorithm's iterate rather than the `finalize_solution` payload bypasses every wrapper in the chain, and the CLI's `on_converged` hook is one, so the `.sol` file, the JSON report and the bound-multiplier suffixes carried scaled coordinates. The pyomo suite caught it as an answer exactly three times too large, three being the factor. They now undo the substitution explicitly, via a new trait accessor through which the application reports the factors it applied.

**Every surface that takes factors now applies them.** `Problem.set_problem_scaling` raised on any non-unit `x_scaling` while the core beneath it handled them; it solves through `optimize_tnlp`, so removing the raise was the whole fix. The C `SetIpoptProblemScaling` already stored and forwarded correctly and only its documentation was stale.

**The sensitivity accessors still refuse a variable-scaled solve**, and the refusal sits on `gradient`, `estimate`, `covariance` and `information` rather than on the solve. A declared model may want the scaled solve and never ask a sensitivity question; what cannot be answered is a query against the held factorization, which belongs to the scaled problem. This came out of testing rather than design: on a real drto model the transform declares sensitivity parameters, so a solve-time guard blocked the very solve the feature exists to enable.

**Tests.** Derivative checks are finite differences against the wrapper's own objective and constraints in scaled coordinates, not restatements of the transform table, so a multiply where a divide belongs, or a row factor used where a column was meant, fails. Two defects surfaced that way: the wrapper panicked when a values call preceded a structure call, and it claimed `is_presolve_wrapper`, which asks whether a presolve wrapper sits *below* rather than whether coordinates moved, so claiming it would have silently disabled presolve on every scaled problem. End-to-end tests solve the same skewed model with and without factors and require the answers to agree in user units, and a free variable is checked for the bound multipliers it must not have.

The refusal message is tested too, which it had not been. It was wrong three times in three different spellings, each surviving because no test in any suite ever rendered it; a Pyomo-written fixture carrying `scaling_factor[x1] = -3` reaches it now. Every test pinning a fixed defect was confirmed to fail against the unfixed code rather than assumed to.

**Measured on a real model.** A solvent-extraction NMPC flowsheet with 8,356 variable factors spanning 1e-6 to 1e6 solves to optimal in 44s natively, against roughly 60s through `core.scale_model`'s clone under `ipopt_v2`, and without the clone or the `propagate_solution` step.

Suites: Rust 2740 across 204 binaries, pyomo-pounce 249, pounce Python 750. The local failures are environment, not code: two need NumPy 2 for `np.trapezoid`, one reads a source file as cp1252 on Windows, one Rust test writes to a POSIX `/tmp` path, and one convex SDP test hits an iteration limit in debug that it clears in release. Docs updated in `scaling.md`, `pyomo.md`, `python/README.md` and `pounce.h`. Merged with main through #531.


jkitchin (commented):
Reviewed the code rather than the description, and verified three defects with throwaway probes against this branch.

The design is sound and unusually well documented — the multiplier rule is *derived* from stationarity rather than asserted, the finite-difference tests check the transform instead of restating it, and catching that `is_presolve_wrapper` asks "is a presolve wrapper *below*" rather than "did coordinates move" avoided silently disabling presolve on every scaled problem. All of this PR's own tests pass locally: `scaling_tnlp` 10, `variable_scaling` 3, `user_scaling_variable_factors` 2.

Three things I'd want fixed before merge.

## Blocking — scaling the bounds turns "absent" bounds into real ones

`ScalingTnlp::get_bounds_info` does `x_l[i] *= s; x_u[i] *= s`, with the comment "infinities scale to themselves." That holds for `f64::INFINITY`, but the `.nl` path never produces one: `nl_reader.rs:595-598` writes the **±1e19 sentinel**, and `nlp_lower_bound_inf` is documented as "the magnitude at or below which a bound is treated as absent" (default ∓1e19, `tnlp_adapter.rs:26`). Multiplying a sentinel by `s < 1` moves it *inside* the threshold.

Probing an unbounded variable with `d = 1e-6`:

```
scaled x_l = -1e13   (threshold -1e19)
scaled x_u =  1e13   (threshold  1e19)
lower: absent before=true after=false
upper: absent before=true after=false
```

A free variable silently becomes bounded, gaining a barrier term and `z_L`/`z_U` multipliers it should not have. This is the primary path — a Pyomo model carrying `scaling_factor` with any unbounded variable — and the headline model here spans 1e-6 to 1e6, so `s < 1` is the ordinary case.

The current tests cannot catch it: `variable_scaling.rs` uses `±1e20` bounds, which `1e-6` maps to `∓1e14` — finite, but so far from the optimum `(5, 2e6)` that the answer is unaffected, so the test passes while the problem's structure has quietly changed.

Worth noting `nl_reader.rs:826` already carries a comment about a previous bug of exactly this shape ("would turn the ±1e19 sentinel into a real bound for `c < 0`").

Fix: compare against the infinity thresholds before scaling and pass sentinels through untouched. That needs the thresholds inside the wrapper, which currently reads options nowhere — probably the least pleasant part of the change.

## Should fix — `variable_scaling()` is never cleared

`install_variable_scaling` assigns only in the `Ok(Some(_))` branch. The `method != "user-scaling"` and `Ok(None)` paths return early without resetting, so the field survives into the next solve on the same `IpoptApplication`:

```
after SCALED solve:   variable_scaling() = Some([4.0])
after UNSCALED solve: variable_scaling() = Some([4.0])
```

The accessor is `pub` and documented as "the factors applied to the last solve," so it now reports the wrong thing. Today's blast radius is small — only the CLI consumes it, and that builds a fresh app per run — but `pounce-cinterface` holds `info.app` across `IpoptSolve` calls (`lib.rs:704`), which is exactly the reuse pattern the Ipopt C API invites. A one-line `*self.variable_scaling.borrow_mut() = None;` at function entry.

## Should fix — the error message is mangled

The `format!` in `install_variable_scaling` spans lines without `\` continuations, so the source indentation lands in the output:

> `pounce: nlp_scaling_method=user-scaling supplied per-variable                  scaling factors that cannot be applied. … Correct the                  factors, or drop nlp_scaling_method=user-scaling.`

Two runs of ~18 spaces, plus a trailing newline inside the literal. The `MSG` constant this replaced used `\` correctly — looks like it was lost in the port.

## Minor

- **`x_scratch` is dead weight.** Every caller writes `self.unscale_x(x).to_vec()`, allocating a fresh `Vec` anyway — five per iteration (`eval_f`, `eval_grad_f`, `eval_g`, `eval_jac_g`, `eval_h`). The scratch buffer saves nothing as written. `std::mem::take` it for the duration of the inner call and put it back, or drop the field.
- **Silent truncation on length mismatch.** `unscale_x` uses `d.len().min(x.len())`, and both `finalize_solution` and the new CLI hook use `zip`. A wrong-length vector yields quietly wrong numbers rather than an error. An assert would be cheap given how much of this change rests on index alignment.
- **The PR body's claim about wrapper placement doesn't match the code.** It says the wrapper "sits closest to the user's TNLP so factors are read in original variable indices," but it is installed at the top of `optimize_tnlp`, and the CLI passes an already-built `PresolveTnlp` (`set_presolve_already_applied(true)`, `main.rs:1313`) — so it sits *outermost*, and the factors are in reduced indices. As far as I can tell this is benign: `PresolveTnlp` only reduces rows, and `linear_eq_elim` projects `x_scaling` through `vars_kept`. But the stated reasoning doesn't describe the actual composition, and stage 3 will be read against that description.

## Not covered

I did not exercise the pyomo end-to-end path (`test_user_scaling.py`) — it needs this branch's CLI wired into pyomo-pounce — so the claimed 249-test suite is unverified on my side. The `sens.py` change reads well: moving the refusal from solve time onto the four accessors, on the argument that a variable-scaled solve is correct and only a query against the held factorization would be answered in the wrong space.

---

devin-griff:
Thanks, this was a good catch on the bounds. All of it is fixed, plus a case in the same class you did not raise and one site I missed on the first pass.

**Bounds.** You are right that the sentinel is the primary path and that the existing tests could not see it. Bounds are now scaled only where present, using the directional test `classify_bounds` already uses: a lower bound is absent at or below `lo_inf`, an upper bound at or above `up_inf`. The thresholds come from `nlp_lower_bound_inf()` / `nlp_upper_bound_inf()` at the install site, which is the unpleasant part you predicted.

The mirror case is the same defect pointing the other way: a *present* bound scaled past a threshold reads as absent and vanishes, with nothing downstream able to tell. `get_bounds_info` returns a bare bool and cannot explain itself, so that one is checked once at setup and refused naming the threshold it crossed.

Your point about `variable_scaling.rs` not catching this was the useful half. The unit test pins `get_bounds_info` directly, and there is an end-to-end one on top: the `±1e20` fixture makes both variables free, so it now asserts they come back with no bound multipliers, which is the user-visible harm. The solution can land in the right place while the structure quietly changes, which is exactly why the old assertions passed.

**Clearing.** Done at function entry, with your `pounce-cinterface` reuse as the reason in the comment. The test reuses one application across a scaled then an unscaled solve.

**Message.** This one deserves more than a line. My first fix was also wrong: the literal ended up carrying `\n` escapes rather than continuations, so it would have printed real newlines and the source indentation. My second was wrong a third way: the caller emits with `eprint!`, and the trailing newline the original literal carried was mine to preserve, so the refusal would have run into the next line of output.

All three survived for one reason, which is the actual finding: nothing in any suite ever rendered that message. Two of the three were caught by reading it, never by running it. There is now a Pyomo-written fixture carrying `scaling_factor[x1] = -3` that reaches the refusal, and a test asserting the sentence arrives whole, on one line, with no run of source indentation, and terminated. I confirmed each of the three spellings fails it.

**Minor.** `x_scratch` is moved out and back with `mem::take`, so the five per-iteration allocations are gone. On the truncations, you listed three sites and I initially fixed two: the CLI hook still had its `zip` and is now asserted along with the others. The PR body claim is corrected, and the actual composition is in the module docs so stage 3 is not read against the wrong description. Your read that it is benign checks out: `linear_eq_elim` maps each surviving column's factor through `vars_kept` and the row-only reducer leaves columns alone.

Every test pinning a fixed defect was confirmed to fail against the unfixed code rather than assumed to.

**One thing outside your review that this turned up.** `Problem.set_problem_scaling` still raised on any non-unit `x_scaling` while the core underneath now handles it. That surface solves through `optimize_tnlp`, so the raise was the only thing blocking it, and there was no test either way, which is how it survived. The C entry point already forwarded correctly and only its docs were stale.

**On the pyomo path you could not exercise:** `pyomo-pounce/tests` is 247 passed, 2 skipped. `python/tests` is 750 passed with 3 failures that are all environment on my machine, two needing NumPy 2 for `np.trapezoid` and one reading a source file as cp1252 on Windows. I verified the NumPy ones fail identically against the pre-change extension rather than assuming.

Merged with main through #531.

jkitchin:
Merged `main` in and resolved the conflict (`ebe16df2`). Only `CHANGELOG.md`
conflicted — 43 lines this side against 221 that had accumulated on `main`,
resolved keep-both with this PR's `### Added` entry at the top of
`[Unreleased]`. `application.rs` and `main.rs` auto-merged.

`main` has moved a long way since this branch was cut (#519/#521, #523,
#524, #527, #528, #532, #533, #534, #535, #540/#541 via #542/#543/#544), so
the post-merge state was re-verified rather than assumed:

**Full suite: 2800 passed, 0 failed** across 210 binaries (`pounce-hsl`
excluded; it fails to link in this environment on `main` too). Release
build clean.

**End-to-end probe of the substitution**, run against the merged tree
rather than read off the diff:

| | `.sol` x | unscaled objective |
|---|---|---|
| no option | `1.54383575462520661`, `2.72808212336050637`, `2.22808212336050637` | `8.7686701206267281e-01` |
| `nlp_scaling_method=user-scaling` | `1.54383575387169936`, `2.72808212307087805`, `2.22808212307087761` | `8.7686701206267337e-01` |

Same point in the model's own units, 9 digits on the iterate and 15 on the
objective, with the scaled column showing `8.7686701206267344e+01` — so the
substitution is being applied and inverted, and the `.sol` carries natural
units. That was the caught defect, and it stays caught.

The refusal path renders correctly through the CLI:

```
pounce: nlp_scaling_method=user-scaling supplied per-variable scaling
factors that cannot be applied. scaling: variable 0 has scaling factor -3;
factors must be finite and positive (a negative factor would reverse the
variable and swap its bounds). Correct the factors, or drop
nlp_scaling_method=user-scaling.
```

All three review points confirmed present with tests behind them:
`absent_bounds_stay_absent_under_scaling` and
`a_free_variable_gains_no_bound_multipliers_from_scaling` (the sentinel
guard), `a_real_bound_scaled_past_the_threshold_is_refused` (the mirror
case), and `the_reported_factors_do_not_survive_into_the_next_solve` (the
cross-solve reset).

No closing keyword, correctly — #486 stays open for stage 3, as the PR
body states.

PR #530, opened 2026-08-07
fix(convex): make the presolve fixpoint terminate on its own, and say when it doesn't (#527)

### LINKED ISSUE #527: pounce-convex presolve: the "fixpoint" never converges on bore3d — it always exits on MAX_ROUNDS, and the cap silently sets how much reduction you get

jkitchin:
Fixes #527.

## Problem

`presolve` documents itself as iterating the reduction passes to a **fixpoint**. On netlib `bore3d` it never reached one — it exited on the `MAX_ROUNDS` layer cap on every solve, at every cap. An arbitrary defensive constant, not the algorithm, was choosing which of two different reduced problems the solver got, and nothing recorded that it had.

Measured on `bore3d` (315 vars / 233 rows), release + FERAL, presolve alone, mean of 20 calls:

| | layers | exit | presolve | reduced |
|---|---|---|---|---|
| before | 32 | cap (unreported) | ~14 ms | 128v / 77r |
| before, cap lifted | 330 | fixpoint | ~112 ms | 128v / 77r |
| after | 32 | cap (**reported**) | **~5 ms** | 128v / 77r |
| after, cap lifted | 238 | **fixpoint** | ~28 ms | 128v / 77r |

The reduced problem is identical in all four rows — same 61 fixings, 17 forcing rows, 64 aggregations. Only boxes differ.

## Cause

Bound tightening is the one reduction that **consumes nothing**. A fixing removes its column, an aggregation removes its row, so those can fire at most `n + m` times however long the loop runs. Narrowing a box leaves the column *and* the row in place, so a pair of rows that mutually imply ever-tighter bounds fires every round forever.

Instrumenting each tightening showed **two** mechanisms, not the one the issue names:

1. **The geometric cascade** (cols 0–4, 133, ratio exactly `3.887e-2` every round — the #523 geometry). This one self-terminates: the boxes collapse toward zero and stop registering once they reach `BOUND_FEAS_TOL`, around round 24.
2. **A slow drift** at O(1) magnitude (ratios like `0.99999990`) that never reaches the absolute floor, plus plain **serialized propagation** — the disjoint-source rule lets only ~3 tightenings fire per layer, so traversing the propagation graph takes hundreds of rounds. Variables are still receiving their *first* finite bound at layer 320.

Mechanism 2 is what actually runs to the cap, and it is why the issue's "detect the geometric signature" direction would not have been enough on its own.

## Fix

**`MAX_BOX_REFINEMENTS` (12)** — a per-`(column, side)` refinement budget carried across rounds and renumbered through each layer's `kept_cols`. With bound tightening finite like every other reduction, termination follows from the algorithm; with the cap lifted, `bore3d` reaches a real fixpoint for the first time.

It is a **count, not a magnitude**, so no scale-dependent constant of the kind #523 came out of enters. A minimum-improvement threshold does not work here, and the two versions fail on *different* models — which is worth stating precisely, because an earlier revision of this body got it wrong:

- an **absolute** floor already exists (`BOUND_FEAS_TOL`) and does stop mechanism 1, but not a cascade converging to a limit of `1e3`, where the improvements stay far above any floor;
- a **relative** floor fails on `bore3d` itself, whose cascade shrinks each bound to `3.887e-2` of its previous value — a **96%** relative improvement every round.

The value was swept, not picked (table in the doc comment): 8 is where the full reduction appears — the deepest box that unlocks a structural reduction on `bore3d` is refined seven times first — and nothing past 8 buys anything. 12 leaves margin.

**`FixpointExit` on `PresolveStats`** (plus `rounds`), and the exit reason on the CLI summary line. This is the part the issue asked for directly.

**Dedup memoization.** Duplicate/parallel-row merging reads coefficients and right-hand sides only, never a box, so a round that follows a bounds-only round and itself fixes, substitutes and drops nothing is re-deriving a merge it already performed. Skipping it there is a memoization, not a heuristic. That hashing pass was ~70% of presolve's cost on a cascade — it is where the 14 ms → 5 ms comes from.

### Rejected: raising `MAX_ROUNDS`

Left at 32, measured rather than assumed. 128 layers costs ~16 ms and reduces `bore3d` to *exactly* the same problem 32 layers reach, while multiplying the chain's retained memory — every layer holds a clone of its input, which #527 itself flags as a concern. A deeper search with no measured reduction to show for it and a real memory cost is not a trade worth making blind.

**So `bore3d` still exits on the cap.** What changed is that it now says so, that the loop no longer depends on the cap to terminate, and that the cap costs nothing here — which is now a measurement rather than luck. The residual cause is serialization (mechanism 2 above), which is a different problem from the one this issue names and is documented in the module for whoever picks it up.

## Blast radius

`pounce-convex` only, plus the CLI presolve summary line and two book sections.

**Corpus sweep** (run by @jkitchin, [review comment](https://github.com/jkitchin/pounce/pull/530#issuecomment-5222929127)), baseline `54219714` vs `cbf6624e`: 0 status changes and 0 objective drifts over 371 LP and 138 QP models. Diffing the per-model `Presolve:` line, 370 of the 394 models that presolve at all are byte-identical and the remaining 24 differ in `tightened` **only** — 0 differ in vars, rows, fixings, aggregations, forcing rows or dominated columns. So the refinement budget changes only boxes, never the structural reduction, measured rather than argued from `bore3d` alone.

Two caveats on that sweep, both post-dating it: the CLI summary-line format changed in `867d5df` (the exit reason is now a suffix on it), so a re-run needs a re-baseline of that diff; and the branch has since merged `main`, whose #524 changed solver retry behaviour on infeasibility verdicts, so a re-baseline should be `6bfc4e7` rather than `54219714`.

The dual recovery is untouched, and the tests that would catch a break in it — the randomized KKT roundtrips over *overlapping* tightening chains in `presolve_bound_tightening.rs`, which is what previously rejected relaxing the disjoint-source rule — pass. The `#523` / `#525` tests pass. `presolve_conic` takes an unlimited budget and `structure_stable: false`: a single pass cannot cascade its own tightening, so its behaviour is bit-identical.

## Tests

`crates/pounce-convex/tests/issue527_tightening_fixpoint.rs`:

- `a_contracting_bound_cascade_reaches_a_fixpoint` — two rows implying each other's bound with a contraction factor of `0.999`, which needs ~28,000 rounds to shave its last improvement below `BOUND_FEAS_TOL`. Asserts `Fixpoint`, and that it took a couple of dozen layers, so a future cap raise cannot make it pass vacuously. **Fails on the parent commit for the stated reason**: with `MAX_BOX_REFINEMENTS` disabled it comes back `RoundCap` at the layer cap, which is the pre-fix behaviour.
- `the_refinement_budget_does_not_change_the_answer` — presolved and unpresolved solves agree, and the recovered point respects the *original* box.
- `the_absolute_floor_would_take_tens_of_thousands_of_rounds_here` — walks the propagation arithmetic. Deliberately scoped to the *absolute* floor: this model's relative step is 0.1% and decays, so it is not evidence about relative thresholds, and the test says so. The relative half rests on `bore3d`'s measured 96%, recorded in `presolve.rs`.
- `structural_reductions_still_cascade_across_rounds` — a 20-link singleton-equality chain still fully collapses; the budget is on box refinement only.

`dedup_memo_tests` in `presolve.rs` (needs internals), behind a `DedupMemo::{Enabled,Disabled}` seam: 200 randomized bounded LPs carrying duplicated and positively-scaled rows, a long bounds-only cascade, and a model where a substitution *creates* a parallel pair. Each asserts the memoized and non-memoized builds produce a byte-identical reduced problem and identical reduction counts.

**Stated honestly, because it is not what it looks like:** no weakening of the memoization gate I tried makes those tests fail — including deleting the safety conjuncts down to `rows_unchanged = structure_stable`. The round trace shows why, and it relocates the soundness argument: a column removal is always preceded by a round that changed dimensions, so the `bounds_only` dimension check is what carries it, and the stack / `*_dropped` conjuncts are belt and braces. They are kept, and the tests stand as a guard against future divergence — but this is not a demonstration of a caught bug, and the code says so rather than claiming otherwise. Discussion in the [review thread](https://github.com/jkitchin/pounce/pull/530#issuecomment-5223139132).

Deliberately not pinned: the `bore3d` numbers above. They need an 8 KB MPS fixture from the COIN-OR netlib mirror that is not vendored, so they live in the module docs and this body rather than in a test. `contracting_pair` is that geometry distilled to the two rows that matter.

---

- [x] Book page under `docs/src/` updated (`convex-solver.md`, `lp-qp-routing.md`; no new page, so `SUMMARY.md` unchanged)
- [x] Every claim in this body and in the code comments is true of the code as it stands now — re-read against the final diff after the review fixes in `867d5df` and the `main` merge in `67cf1d5`; the test names, the relative-improvement figure, and the sweep status above were all corrected in that pass


jkitchin:
## Review — the corpus sweep the body asks for, plus five notes

The body says the `.nl` corpus was unavailable and the `lp` / `qp` suites had
not been run, and that it was worth doing before merge. Both suites are
available, so I ran them. **Nothing broke, and the reduction is provably
unchanged where it matters.**

### The sweep

Baseline `54219714` (this branch's exact merge-base) vs head `cbf6624e`, same
host, same driver, 120 s limit.

| | LP (netlib + Meszaros, 371) | QP (Maros-Mészáros, 138) |
|---|---|---|
| status changes | **0** | **0** |
| objective drift > 1e-6 rel | **0** | **0** |
| total solve time | 481.8 s → 481.9 s | 228.6 s → 229.1 s |
| total iterations | 9512 → 9521 | 2628 → 2631 |

More useful than the aggregate: I diffed the per-model `Presolve:` line, so
this compares the *reduction itself* rather than only the answer.

| | LP | QP |
|---|---|---|
| models that presolve at all | 321 | 73 |
| byte-identical reduction | **307** | **63** |
| differ in `tightened` only | 14 (4 up, 10 down) | 10 |
| differ in vars / rows / fixed / aggregated / forcing / dominated | **0** | **0** |

So the refinement budget changes **only boxes**, never the structural
reduction — which is what the module docs claim from `bore3d` alone, now
measured on 394 real models. `bore3d` itself is identical on both builds
(`315 → 128` vars, `233 → 77` rows, 61 fixings / 64 aggregations / 17 forcing,
objective `1.3730803942727209e+03`), with the truncation line added.

Also: 379 `pounce-convex` tests pass in release, `cargo fmt --check` is clean,
and clippy on `presolve.rs` reports only the pre-existing pedantic set
(`expect`/`unwrap`, collapsible `if`) — nothing the diff introduced.

I could not confirm the `bore3d` presolve `~14 ms → ~5 ms` claim end-to-end:
that model's whole solve is 53 ms on both builds, so presolve time is inside
the noise of a wall-clock measurement.

### 1. "On most models the cap is never reached" is not true on this corpus

`docs/src/lp-qp-routing.md`. The new line fires on **172 of 371 LP models
(46%)** and **34 of 138 QP models (25%)** — 206 of the 394 models that presolve
at all, i.e. a slight majority of them. It is not the rare event the docs and
the message's own framing ("Nothing is wrong when you see it") describe, and it
prints unconditionally on stdout.

Given the sweep above shows the truncation costs nothing structural anywhere,
the honest version is closer to "the cap is commonly reached and measurably
costs only box tightness" — which is a *better* story than the current one, and
it is the one the data supports. Worth reconsidering whether a line that
appears on half of all solves belongs on stdout by default.

### 2. Nothing tests the dedup memoization

It is the one change in the diff that can silently alter the reduced problem,
and all four new tests are about the budget and the exit reason.

I convinced myself it is sound by reading: when `rows_unchanged` holds, the
all-`BoundTightening` stack implies no column was fixed or substituted (every
`dropped_col` path pushes a stack entry — `FixedVar` 1119, `ForcingRow` 1298,
`DominatedColumn` 1425/1428, `FreeColSingleton` 1168, `FreeColumnFixed` 1786),
and the two `*_dropped` flags cover the row removals that carry no stack entry.
So `build_rows` reproduces the input rows verbatim, and the input is the prior
round's dedup output. `structure_stable` starts `false`, so round 1 always
dedups; `aggregate_once` clears it; `dedup_rows` only ever drops rows and never
rewrites one, so it is idempotent and the induction closes.

That is a five-step argument spread over 800 lines, resting on an
exhaustiveness claim ("every column removal has a stack entry") that a future
reduction can break without touching any of this code. A test that runs a
presolve with the skip forced off and asserts the identical reduced problem
would pin it.

### 3. Three different numbers for the same quantity

And it is the crux of "why not a relative threshold":

- `issue527_tightening_fixpoint.rs:23` — "the *relative* improvement is a hair
  under 100% every round"
- `presolve.rs` module doc — "the relative improvement is ~96% every round"
- the test's own assertion, `(relative - 0.001).abs() < 1e-5` — 0.1%

CHANGELOG repeats the first. The argument survives whichever is right (0.1% is
still far above any relative floor anyone would set), but a reader checking the
reasoning hits a contradiction inside a single file.

### 4. `rounds` is 0 for a one-round fixpoint, not just for a single pass

The `chain.len() == 1` path returns the layer itself, whose own `chain` is
empty, so `stats()` falls through to `stats_once()` and `rounds` keeps its
`Default` of 0. The field doc says "`0` for a single-pass presolve", which
reads as the conic path only. Cosmetic — `RoundCap` needs ≥32 layers, so the
CLI message can never print a misleading count.

### 5. `allows()` fails silent on an out-of-range column

`side.get(col).is_some_and(...)` returns `false`, so a length or numbering
mismatch would quietly *disable* bound tightening rather than panic — the exact
"a constant, not the algorithm, decided the reduction, and nothing said so"
failure class this PR exists to eliminate. The invariant holds today (`charge`
runs before `remap`, and `remap` always re-lengths to `kept_cols.len() ==
reduced.n`), so this is defensive: a `debug_assert_eq!` on the length would
keep it honest.

### Verdict

The mechanism is right and the reasoning holds up: `allows()` gates the
mutation at `presolve.rs:1622`, not just the stack record, so the reduced box
and the `BoundTightening` entry cannot drift apart; every applied tightening
moves the box by more than `BOUND_FEAS_TOL` and so always produces exactly one
entry per `(col, side)`, which is what `charge` counts. Termination follows.
`presolve_conic` correctly opts out with `structure_stable = false` and an
unlimited budget.

Approve. Item 1 is worth fixing before merge since the sweep contradicts a
shipped doc claim; 2 is the one I would most want as a test; 3–5 are small.

jkitchin:
Thanks — the sweep is what the body was missing, and item 3 was a real error on my part, not just an inconsistency. All five addressed in `867d5df`. One of them came out differently than either of us expected; details below.

### 1. The docs claim the sweep contradicts

You're right and the doc was wrong. Fixed in both places, and I took the second half of your suggestion too: the truncation is now a **suffix on the existing summary line** rather than a line of its own.

```text
Presolve: 315 → 128 vars, 233 → 77 rows (fixed 61, ..., tightened 158, cap-truncated after 32 layers)
```

A separate line on 46% of LP solves reads as a warning about something working as designed. The suffix keeps the one thing #527 actually needs — that a truncated reduction be distinguishable from a converged one — without that framing. `lp-qp-routing.md` now leads with your numbers and the finding that it costs only box tightness, which as you say is the better story.

### 2. The memoization test — and what trying to break it showed

Added `dedup_memo_tests` in `presolve.rs` (it needs internals) behind a `DedupMemo::{Enabled,Disabled}` seam: 200 randomized bounded LPs carrying deliberately duplicated and positively-scaled rows, a long bounds-only cascade, and a targeted model where a substitution *creates* a parallel pair that did not exist before. Each asserts the memoized and non-memoized builds produce a byte-identical reduced problem and identical reduction counts.

**But I could not make any of them fail under a broken gate, and I think that is worth more than the tests are.** I tried three weakenings, including deleting the safety conjuncts entirely down to `rows_unchanged = structure_stable`. All three still produced identical reductions. The round trace on the targeted model says why:

```text
round: chain=2 n=3 m_eq=0 m_ineq=3  cat_changed=true bounds_only=false agg=true
round: chain=3 n=2 m_eq=0 m_ineq=1  cat_changed=true bounds_only=false agg=false   <- the merge
```

The aggregation collapses the `x₃ → x₂` chain in round 1, so the substitution and the merge both land in round 2 — whose predecessor changed the shape, so `structure_stable` was never set. That looks general: a column removal is always preceded by a round that changed dimensions, so **the `bounds_only` dimension check is what carries the soundness argument**, and the stack / `*_dropped` conjuncts you traced are belt and braces rather than the load-bearing part.

That reframes your concern rather than dismissing it — the exhaustiveness claim over the catalog turns out not to be what the skip rests on. I kept the conjuncts (a conservative boolean costs nothing) and kept the tests as a guard against future divergence, but I've written both up in the code as *"no weakening tried so far makes this fail"* rather than "verified to bite", because the latter would have been false. If you can see a shape where a catalog reduction removes a column on a round whose predecessor was bounds-only, that is the case that would settle it, and I'd take the hint.

### 3. Three numbers — and the argument was weaker than I wrote it

Not just an inconsistency; I had conflated two different models, and the test was overclaiming. Checked:

- **`bore3d`'s real cascade**: ratio `3.887e-2`, so a **96%** relative improvement every round. No relative threshold stops that. The `presolve.rs` figure was the right one.
- **The synthetic model in the test file**: `u ← 1 + 0.999u`, whose step is 0.1% of the bound and *decays toward zero* as it nears its limit. A relative floor would eventually catch it.

So `neither_an_absolute_nor_a_relative_threshold_would_stop_this` was asserting something its own model does not support. Renamed to `the_absolute_floor_would_take_tens_of_thousands_of_rounds_here`, which is what it actually establishes, with the relative half attributed to the measured `bore3d` ratio where it belongs. CHANGELOG and the test header now say the absolute and relative versions fail on *different* models, which is the honest form of the argument.

### 4. `rounds`

`stats_once` now reports `1` — a single-layer handle is one pass however it was reached — and the field doc names all three cases (conic path, no-op, one-round fixpoint) instead of implying only the first.

### 5. `allows()`

`debug_assert_eq!(refinements.len(), n)` at each `presolve_once` entry, with the reasoning in the comment: a short budget would silently switch bound tightening off, which is the failure class this PR exists to remove.

---

`cargo fmt --check` clean, no new clippy warnings (the test-only `DedupMemo::Disabled` is `cfg_attr`-gated so release builds don't warn), 382 `pounce-convex` and 442 `pounce-cli` tests pass in release. I have not re-run the corpus sweep — none of these touch the reduction path except the memoization test seam, but the CLI summary-line format did change, so if your sweep diffs that line it will need a re-baseline.


---

Issue #446, opened 2026-08-02
solver_selection=nlp reports local infeasibility on 15 feasible Maros-Meszaros QPs


jkitchin:
## Summary

With `solver_selection=nlp`, POUNCE reports

```
EXIT: Converged to a point of local infeasibility. Problem may be infeasible.
```

on 15 Maros-Meszaros convex QPs that are feasible and that POUNCE's own
`qp-ipm` path solves to optimality.

This is a confident wrong answer rather than a failure: a caller that
trusts the status will conclude its model has no solution. Pyomo maps
`Infeasible_Problem_Detected` into the infeasible family, so it surfaces
as a modelling error rather than a solver limitation.

## The clearest case: QSCSD1

The NLP path terminates at a point that satisfies its own convergence
tolerances by a wide margin, and then labels it locally infeasible:

```
Number of Iterations....: 30
Objective...............:   8.6666498134108796e+00
Dual infeasibility......:   1.1546319456101628e-14
Constraint violation....:   1.1102230246251565e-16
EXIT: Converged to a point of local infeasibility. Problem may be infeasible.
```

Constraint violation is 1.1e-16 and dual infeasibility is 1.2e-14 — this
is a converged KKT point, not a point of local infeasibility. The
objective (8.66665) agrees with the `qp-ipm` optimum (8.666667) to five
figures. Whatever renders the verdict is not consulting the residuals at
the point it is judging.

## Reproduce

```
pounce benchmarks/qp/nl/QSCSD1.nl solver_selection=nlp     # locally infeasible
pounce benchmarks/qp/nl/QSCSD1.nl solver_selection=qp-ipm  # obj=8.66666667
```

(`.nl` inputs live under `POUNCE_BENCH_DATA` when set.)

## Affected problems

All 15 are solved to optimality by `qp-ipm`. `n`/`m` and the reference
objective are from `benchmarks/qp_convex/convex.json`; iteration counts
are the NLP path's.

| problem | n | m | nlp iters | qp-ipm obj | qp-ipm iters |
|---|---:|---:|---:|---:|---:|
| DTOC3 | 14999 | 9998 | 4 | 235.262481 | 12 |
| QBANDM | 472 | 305 | 84 | 16352.34204 | 19 |
| QETAMACR | 688 | 400 | 99 | 86760.36963 | 32 |
| QFFFFF80 | 854 | 524 | 324 | 873147.4605 | 28 |
[...]

Issue #385, opened 2026-07-28
Scale-invariance migration: Step 6 (outer convergence acceptance) and Step 4 (presolve thresholds, post-#377) remain


jkitchin:
**Correctness follow-up to the scale-invariance migration** — the harness
(#381), the significance primitives (#382), the `pounce verify` migration
(#383), and the restoration-gate migration (#384). Those landed Phases 0–1 and
Steps 3 and 5. Two pieces of the plan remain; until this issue they existed
only in commit message prose (see `3f419c94`, PR #384), not anywhere
discoverable. The stakes are wrong verdicts: an infeasible model still reports
`SOLVED` at down-scaled rows (`k = -12, -10, -8`), the same false-feasibility
class as #376.

## Step 6 — make the outer convergence acceptance scale-relative

The remaining wrong cells in the scale-invariance harness
(`pyomo-pounce/tests/test_scale_invariance.py`) are decided **before** the
restoration gates see the point: the outer convergence check accepts it first.
After Step 5, on the infeasible models:

- `k = -12, -10, -8` still report `SOLVED` (a false feasibility verdict at
  down-scaled rows), and `k = -6` still reports `FAIL`;
- three of the seven harness models remain scale-dependent;
- overall harness state: 25/91 wrong cells (was 28/91 before Step 5).

Note recorded for whoever picks this up (from the Step 5 commit message): with
the row magnitude derived as `1/dc` from the solver's own scaling, the
scale-relative test on the *unscaled* violation reduces algebraically to an
absolute test on the *scaled* violation — the same condition the #380 guard
already applies one level up. Step 5 therefore made the restoration gates agree
with that guard; Step 6 is about the outer acceptance path, and should use the
same `is_significant` / `is_negligible` primitives in
`crates/pounce-common/src/tolerance.rs` (mind the two directions: accepting
uses `is_negligible`'s clamped form, proving uses `is_significant`'s pure
relative form — they are deliberately not negations of each other, see the
module docs).

## Step 4 — migrate presolve certification thresholds (was blocked on #377)

Step 4's targets, `certify_margin` and the witness-refutation slack in
`witness_refutes_infeasibility` (`crates/pounce-presolve/src/lib.rs`), only
exist on the #377 branch, so it was skipped rather than done. Once #377 merges
this becomes actionable: those thresholds are absolute
(`certify_margin = max(tol, 1e-12)`; the witness check already uses an ad-hoc
clamped relative form) and should be reviewed against the shared primitives so
[...]

Issue #292, opened 2026-07-22
[Adversary] SOLVER_BUG: NaN gradient reports Solve_Succeeded — amax drops NaN before the finiteness guard


jkitchin:
> Discovered by the pounce adversary agent, root-caused while fixing #275.

## Family / solver

`pounce.minimize` (NLP path) — but the root cause is in shared `pounce-linalg` code.

## Problem

A user objective or gradient that returns `NaN` is reported as a successful solve:

```python
pounce.minimize(lambda x: float(x[0]**2), np.array([0.5]), jac=lambda x: np.array([np.nan]))
# -> success=True, message='Solve_Succeeded', fun=0.25, nit=0
```

This is the most dangerous shape of the family: `fun` is **finite** (the value at `x0`), so the caller gets no signal at all. The NaN-`x0` and infinite-bound shapes were fixed in #275; this one was deliberately left out because the cause is different and the blast radius is wider.

## Root cause

The max-reduction behind the dual-infeasibility norm **silently drops NaN**:

- `crates/pounce-linalg/src/compound_vector.rs:235` uses `if v > m { m = v }` — `NaN > m` is `false`, so a NaN component leaves the running max untouched.
- the dense path (`dense_vector.rs:383`) delegates to BLAS `iamax`, which is likewise NaN-indifferent.

So a NaN gradient is laundered into `0.0` before any guard sees it. Confirmed end-to-end with `print_level=5`:

```
   0  2.5000000e-01 0.00e+00 0.00e+00   -1.0 0.00e+00      - 0.00e+00 0.00e+00   0
Dual infeasibility......:   0.0000000000000000e+00    0.0000000000000000e+00
Overall NLP error.......:   0.0000000000000000e+00    0.0000000000000000e+00
EXIT: Optimal Solution Found.
```

There **is** already a guard at `crates/pounce-algorithm/src/ipopt_alg.rs:1315`:

```rust
let nlp_err = self.cq.borrow().curr_nlp_error();
if !nlp_err.is_finite() {
    return IterateOutcome::Terminate(SolverReturn::InvalidNumberDetected);
}
```

It never fires, because the value it checks was already laundered to `0.0` upstream. The guard is correct; the norm feeding it is not.

## Expected vs actual

| input | expected | actual |
|---|---|---|
| `jac` returns NaN | `Invalid_Number_Detected` | `Solve_Succeeded`, `fun=0.25`, `nit=0` |
[...]

PR #531, opened 2026-08-07
fix(alg): judge the KKT primal term against its floating-point noise floor (#528)

### LINKED ISSUE #528: IPM exits Search_Direction_Becomes_Too_Small on feasible, BOUNDED LPs once data magnitude reaches ~1e7 (scipy/HiGHS solves them)

jkitchin:
Fixes #528

## Problem

Feasible, bounded LPs with `O(1)` matrix entries but right-hand sides of magnitude `~1e7` and above exited `Search_Direction_Becomes_Too_Small` — **while holding the correct optimum**. 15 of the reporter's 48-run sweep on this build (the issue reports 9 of 48 on 0.9.0); none at data scale `1e6`, more of them at `1e8`.

Reporter's seed 0, `n=20`, scale `1e7`:

| | status | objective | NLP error | iters |
|---|---|---|---|---|
| before | `Search_Direction_Becomes_Too_Small` | 5.105014873e+07 | 1.4901e-08 | 73 |
| after | `Solve_Succeeded` | 5.105014873e+07 | 1.4901e-08 | 69 |
| scipy/HiGHS | optimal | 5.105014924e+07 | — | — |

The objective was already right to eight significant figures before the fix. Nothing here was a wrong answer; code 3 is an honest "I stalled". What was missing was the certificate.

## Cause

The KKT error the convergence test compares against `tol` is

```
max( ‖∇L‖∞ / s_d ,  max(‖c‖∞, ‖d − s‖∞) ,  ‖compl‖∞ / s_c )
```

The dual and complementarity terms are normalised by `s_d` / `s_c`. The primal one — upstream Ipopt's too — is a bare absolute residual. But `c_i = g_i(x) − b_i` and `d_i − s_i` are each a **difference of quantities the row's own size**, so they are quantised in units of `eps · |b_i|`: no iterate can place either residual strictly between `0` and one ulp of the row's magnitude.

The failing solve above reported an NLP error of exactly `1.4901161193847656e-08`. That is `ulp(9.718e7)` — one ulp of the largest right-hand side in the model — against the default `tol = 1e-8`.

So `nlp_err <= tol` had stopped being a statement about the iterate and become a bet on the residual landing on a bitwise-exact `0` rather than on one ulp. That is the mechanism behind the pattern the reporter found puzzling: the failures scatter across seeds instead of tracking scale uniformly, because at `1e8` many runs still happen to land on the exact zero (10 of 16 did). An iterate that lost the bet could never certify at all, so the solve kept recomputing a point it could not improve until the search direction collapsed.

The `nlp_scaling_method="none"` and `constr_viol_tol` observations in the issue both follow: gradient-based scaling picks `dc = 1` here (row gradients are `0.5…5`, well under `nlp_scaling_max_gradient = 100`), so there is no scaling to turn off; and `constr_viol_tol` gates a different quantity than the aggregate that was pinned.

## Fix

The **strict** gate now judges the primal term against the finest residual each row can represent — `IpoptCalculatedQuantities::curr_primal_infeasibility_above_noise`, which counts a row's residual only where it rises above that row's own floor. The floor is the larger of two irreducible effects:

* **Placing `x`** — the existing `row_noise_floor`, `eps · ‖x‖∞` passed through the row at `max_j |∂g_i/∂x_j|`. This is the same per-row model #390 / #446 already use to decide when a residual is too fine to be real, carried over deliberately rather than inventing a second constant with a second rationale.
* **Forming the residual** — `eps · m_i`, with `m_i` the magnitude of the quantities subtracted (`|b_i|` on the equality block, `max(|d_i|, |s_i|)` on the inequality block). `row_noise_floor` models `x`-placement only, which understates the `d − s` quantum on wide rows.

One adjustment was needed to reuse `row_noise_floor`: it returns `INFINITY` for a row with an empty Jacobian, meaning "abstain". That is the *safe* direction for the relative measures it was written for and the *unsafe* one here — silencing a constant row `0 = b` would forgive a genuine infeasibility outright — so non-finite floors are read as `0` and such a row is judged on its residual alone.

The safety factor is the **`primal_noise_floor_kappa`** option (default `64`). **Setting it to `0` disables the floor entirely**, restoring upstream Ipopt's bare-absolute primal term bit-for-bit — the escape hatch for anyone who wants the old strict semantics.

**Rejected alternatives, and why:**

* *Normalise the primal term in `curr_nlp_error` itself (an `s_p` alongside `s_d` / `s_c`).* This is the theoretically tidy answer and it is what Ipopt does for the other two terms, but `curr_nlp_error` feeds the μ-update oracles, the filter, the acceptable-level band and the rapid-infeasibility pre-filter. Changing the number every one of those reads, to fix a gate none of them are, is a blast radius out of all proportion to the defect.
* *Relax the exit status instead — upgrade a tiny-step exit that satisfies `acceptable_tol` to `Solved_To_Acceptable_Level`.* This treats the symptom. These points carry a real strict certificate once the aggregate is measured correctly.
* *Also floor the `constr_viol_tol` gate.* Deliberately not done. `constr_viol_tol` is a tolerance the user states; silently widening it to a machine-precision floor would change the meaning of an explicit option.

## Blast radius

**Corpus sweep, baseline `54219714` (this branch's merge-base) vs head, same host and driver.**

* **LP (371 netlib + Meszaros) and QP (138 Maros-Mészáros): bit-identical.** Not merely "no status changes" — identical model by model in both the objective bit pattern and the iteration count (9512 and 2628 iterations on each side). On data that is `O(1)` no row is near its resolution limit, so the floored aggregate and the raw one are the same number and the gate is upstream's, unchanged.
* **Vanderbei (733 NLPs): 8 status changes, every one an improvement, none a regression.**

| model | base `54219714` | head |
|---|---|---|
| `orthrege` | 2652 iters, `Solved_To_Acceptable_Level` | **84 iters**, `Solve_Succeeded` |
| `steenbrf` | 3000 iters, `Maximum_Iterations_Exceeded` | **481 iters**, `Solve_Succeeded` |
| `brainpc2` | 3000 iters, `Maximum_Iterations_Exceeded` | **1003 iters**, `Solved_To_Acceptable_Level` |
| `cresc4` | `Infeasible_Problem_Detected` | **`Solve_Succeeded`**, obj `0.8719` |

`orthrege` is the issue's pathology exactly: thousands of iterations spent at a point the solver could not improve because the certificate was unreachable.

**The earlier claim of "bit-identical except the targeted case" was wrong in both directions** — exactly right on LP/QP, and materially understated on nonlinear problems. Iterate trajectories *do* change on already-succeeding large-`|b|` models: `DenseLp::new(0, 20, 10, 1e8)` goes 31 → 22 iterations, because the base spent iterations 22–31 in a pure stall, re-rolling until the primal residual happened to land on an exact `0.0` and won the certificate. The honest claim is **no status regressions on 1242 corpus models, several improvements, and changed trajectories on nonlinear problems**.

**Locally-infeasible models can cost more.** `cresc50` goes 2 attempts / 1544 iterations → 3 / 4950 for the same `Infeasible_Problem_Detected`; `launch` adds a third attempt. The mechanism is `application.rs:1761`: the ℓ₁ fallback's outer loop breaks when an inner solve returns non-Optimal, on the reasoning that "escalation won't fix numerical failure at this ρ". The old false `Search_Direction_Becomes_Too_Small` was being read as exactly that numerical failure. With the inner solve certifying honestly, `inner_ok` is true, the slacks still do not collapse (the model really is infeasible), ρ is below `l1_penalty_max`, and the loop runs the escalation it was always designed to. Same verdict; the base's saving was a false economy.

Three deliberate containments:

* Only the strict gate reads the floored value. `constr_viol` is still tested against `constr_viol_tol` on the **full, unfloored** residual, so what the floor admits is bounded by the user's own feasibility tolerance. The scale-relative feasibility veto (#385/#390) still sees the unfloored residual too.
* The acceptable-level band keeps the raw `nlp_err`. It sits two decades above `tol`, clear of any realistic quantum.
* A non-finite `nlp_err` passes through untouched. `f64::min` returns the *other* operand at `NaN`, which would have laundered exactly the `Invalid_Number_Detected` signal #292 built the `has_valid_numbers` sweep to raise.

A fourth containment turned up while writing the boundary test, and it is stronger than the `constr_viol_tol` argument: **the floor cannot buy an infeasible model a certificate at any kappa.** Re-running the boundary case with `ROW_NOISE_KAPPA` raised a thousandfold — a floor that silences the row's residual outright — still refuses. On a model with no feasible point the filter and the restoration phase reach a verdict on their own criteria, and the convergence gate the floor feeds is never the deciding vote. The floor only ever participates at a point the rest of the algorithm already believes is converged.

The extra work is computed only on iterations where it can change the verdict. That laziness matters more than the original note implied: the accessor is not two extra Jacobian sweeps on a cached number, it is `nlp_error(true)`, a second evaluation of the whole KKT error (`optimality_error_scaling`, both `grad_lag` terms uncached, all four complementarity vectors, the `has_valid_numbers` sweep) plus the two `compute_row_amax` sweeps. The comment now says so, for whoever next considers reusing it somewhere hotter.

**Reporting.** The summary block prints the tested aggregate under `Overall NLP error` when it differs from the reported one, so a run that ends `EXIT: Optimal Solution Found` beside an error above `tol` accounts for the gap rather than merely presenting it. Solves where the two agree — every `O(1)` model, and every run with the floor off — keep the block byte-identical to upstream's.

**Known limitation:** a user who tightens `constr_viol_tol` *below* the row's ulp (e.g. `1e-8` on data at `1e8`) still will not get a certificate. That is the tolerance gate doing what it was asked, not the defect recurring.

## Tests

`crates/pounce-algorithm/tests/issue_528_large_rhs_kkt_noise_floor.rs`, plus unit tests in `ipopt_cq.rs`.

The issue's LP family is restated so it needs no reference solver: it is **scale-equivariant**. Mapping `b → σ·b` and `ub → σ·ub` maps the LP onto itself under `x → σ·x`, so `f*(σ) = σ·f*(1)` exactly, and `σ = 1` — the regime POUNCE has always handled — is its own oracle.

* `a_large_right_hand_side_does_not_cost_the_certificate` — 24 LPs across `σ ∈ {1e6, 1e7, 1e8}`. **Fails on the parent commit for the stated reason:** 12 of the 24 lose their certificate there, all at `σ ∈ {1e7, 1e8}`, every one holding the right optimum.
* `the_floor_refuses_a_violation_just_above_the_quantum` — the near side of the guard, with the shortfall an order of magnitude *under* `constr_viol_tol` and *over* the quantum. Two findings are recorded in its doc: `bound_relax_factor` has to be pinned to `0`, because at `ub = 1e9` its default already widens every bound by ~10 and absorbs violations far larger than the floor forgives (the model is then genuinely feasible and certifying it is correct); and the refusal survives a thousandfold kappa, which is the containment noted above.
* `a_violation_just_above_the_noise_quantum_is_still_refused` and `an_infeasible_large_scale_row_is_still_refused` — the `constr_viol_tol` containment at two distances, now asserting a verdict that is neither a certificate nor a crash, and checking the returned point genuinely violates.
* `switching_the_noise_floor_off_restores_the_old_verdicts` — pins the option: at `primal_noise_floor_kappa = 0` the issue's LPs lose their certificates again.
* `the_reported_aggregate_and_the_tested_one_diverge_at_large_scale` — pins that the two numbers really do differ here, which is what makes the summary line fire.
* Unit tests for the **equality block**, which had no cover at all before: the integration LP is all-inequality, so `c.dim() == 0` and the `declared_c_rhs()` branch never ran. `a_sub_quantum_equality_residual_is_silenced_and_a_coarser_one_is_not` drives it through the real accessor; `the_equality_floor_rides_the_row_scaling` pins that a floor built from the declared RHS is comparable with `curr_c()` at three row scalings; `a_zero_kappa_switches_the_floor_off_completely` pins the opt-out at the accessor.
* Existing unit tests for `amax_above_floor` and `strict_overall`.

Beyond the repo suites, the reporter's own Python reproduction was run against a wheel built from this branch: **48/48 `Solve_Succeeded`** at the scipy optimum, out to a data scale of `1e10`.

**Not pinned:** the `1e11` regime — a conditioning limit unrelated to this gate.

---

- [x] Book page under `docs/src/` updated — `docs/src/options.md` gains a section on large constraint values and `primal_noise_floor_kappa`
- [x] `cargo fmt --all -- --check`, `cargo clippy` (`-D clippy::correctness -D clippy::suspicious`), `cargo test` clean — 2686 passed, 0 failed; `check-release-consistency.sh` and `check-docs-consistency.sh` OK


jkitchin:
## Review — the corpus sweep the body asks for, plus six notes

The body says the bench data was unavailable and a sweep against the parent
"would strengthen it and is worth running before merge if a mirror is to hand".
It is, so I ran it. **The fix is worth considerably more than the PR claims,
and the blast radius is larger than it claims.** Both halves of that matter.

The diagnosis is right and I could not break the mechanism. Sweep first, then
what I verified by reading, then the notes.

### The sweep

Baseline `54219714` (this branch's exact merge-base) vs head `cbd359b5`, same
host, same driver.

**LP (371 netlib + Meszaros) and QP (138 Maros-Mészáros): bit-identical.** Not
"no status changes" — *identical*, model by model, in both the objective bit
pattern and the iteration count (9512 and 2628 iterations on each side). The
floor never fires on either corpus. That is as clean a confirmation of "the
common path is unchanged" as this could get.

**Vanderbei (733 NLPs): 8 status changes, every one an improvement, none a
regression.** I re-ran each serially with no competing load, so these are not
harness noise:

| model | base `54219714` | head `cbd359b5` |
|---|---|---|
| `orthrege` | 2652 iters, `Solved_To_Acceptable_Level` | **84 iters**, `Solve_Succeeded` |
| `steenbrf` | 3000 iters, `Maximum_Iterations_Exceeded` | **481 iters**, `Solve_Succeeded` |
| `brainpc2` | 3000 iters, `Maximum_Iterations_Exceeded` | **1003 iters**, `Solved_To_Acceptable_Level` |
| `cresc4` | `Infeasible_Problem_Detected` | **`Solve_Succeeded`**, obj `0.8719` |
| `hs057` | 22 iters | 21 iters, same objective |

`orthrege` is the one to put in the CHANGELOG: 31× fewer iterations and a real
certificate instead of the acceptable-level fallback, same objective to eight
figures. It is the #528 pathology exactly — thousands of iterations spent at a
point the solver cannot improve.

**There is a cost side.** On locally-infeasible problems the multistart driver
now runs *more* attempts before giving up: `cresc50` goes 2 attempts / 1544
iterations → 3 attempts / 4950 iterations for the same
`Infeasible_Problem_Detected`; `launch` adds a third 321-iteration attempt for
the same verdict. Worth understanding before merge — an extra attempt is a
plausible consequence of a stricter-passing continued point changing what the
restart loop sees, but it is a 3× cost on those models and the PR does not
predict it.

(Four rows in the raw sweep — `grouping`, `indef`, `lewispol`, `static3`
reading `Solver_Error` at base — are harness artifacts: those base runs were
killed mid-solve under concurrent load, and `indef`/`static3` printed
`Diverging_Iterates` on both sides. Not differences.)

### What holds up

**Units.** The one thing that could have quietly poisoned this is the formation
floor being in a different space from the residual it bounds. It isn't:
`declared_c_rhs()` reapplies `c_scale` (`orig_ipopt_nlp.rs:2006`, pinned by
`declared_c_rhs_carries_the_row_scaling`), so `eps·|b_i|` lands in the same
internally-scaled space as `curr_c()`. Same for `max(|d_i|, |s_i|)` against
`d − s`.

**The laziness is exact, as claimed.** `components_pass` replicates the last
three conjuncts of `passes_component_tols`, so on `!components_pass` the result
is `false` whatever the aggregate is; and on `nlp_err <= tol` the floored value
is smaller still. No iterate changes verdict from the short-circuit.

**`strict_overall`'s `min` compares like with like.** `ipopt_alg.rs:900` passes
`curr_nlp_error()` — scaled — matching `curr_nlp_error_above_primal_noise()`.
(The per-component tolerances at both sites are *unscaled*; that asymmetry is
pre-existing and already documented at `opt_error.rs:509`.)

**The containment is real, at every data scale.** `passes_component_tols` still
tests the raw, unfloored, unscaled `constr_viol` against `constr_viol_tol`, so
nothing above `1e-4` can certify no matter how large `|b|` grows and how loose
`64·eps·|b|` gets. The floor only ever narrows what the aggregate admits
*inside* that band.

**Both magnitudes err conservative.** `|b_i|` understates the quantum of
`g_i(x) − b_i` when `|g_i(x)| ≫ |b_i|` — under-forgiving, the safe direction,
and the placement floor covers exactly that regime. And `max(|d_i|, |s_i|) → 0`
as an active zero-bound row converges, so the floor vanishes precisely where a
floor would be dangerous (the `|d_i|` trap `relative_d_infeasibility_max`'s doc
warns about, just below in the same file).

561 `pounce-algorithm` tests pass in release, `cargo fmt --check` is clean,
clippy on the crate shows only the pre-existing pedantic set, and all 12 CI
checks are green.

### 1. "Bit-identical except the targeted case" is wrong in both directions

The sweep above settles it: exactly right on LP/QP, and materially understated
on nonlinear problems. Even inside the PR's own fixture there is a
counterexample. `DenseLp::new(0, 20, 10, 1e8)` — seed 0, one of the 12 the body
says *already* returned `Solve_Succeeded` — at `54219714` and at head with
`print_level 5`:

| | base `54219714` | head `cbd359b5` |
|---|---|---|
| iterations | **31** | **22** |
| status | `Solve_Succeeded` | `Solve_Succeeded` |
| Constraint violation (reported) | `0.0` | `2.3841857910156250e-07` |
| Overall NLP error (reported) | `2.506e-09` | `2.384e-07` |
| objective | `4.9054816965230238e+08` | `4.9054816965230244e+08` |

Base's iterations 22–31 are a pure stall — objective frozen, dual infeasibility
frozen at `2.51e-14`, `‖d‖` oscillating around `1e-7`, `H` and `t` steps — nine
iterations spent re-rolling until the primal residual happened to land on an
exact `0.0` and won the certificate. Base's final `‖d‖` is `2.38e-07`, the same
number head reports as its violation.

That is the mechanism caught in the act, and it is the best evidence in this PR
that the diagnosis is correct — but it also means already-green large-`|b|`
models change iteration count and final point. The honest claim is "no status
regressions on 1242 corpus models, several improvements, and iterate
trajectories do change on nonlinear problems".

### 2. The end-of-run summary now contradicts itself, and a comment is stale

That same run prints, with `tol = 1e-8`:

```
Constraint violation....:   2.3841857910156250e-07
Overall NLP error.......:   2.3841857910156250e-07
EXIT: Optimal Solution Found.
```

An NLP error 24× the tolerance next to a success status, with nothing in the
output accounting for the gap. `application.rs:2504` reports `curr_nlp_error()`
— the raw value — and the comment immediately above it says *"'Overall NLP
error' is `curr_nlp_error`, which is also the convergence gate
(`ipopt_alg.rs`)"*. After this PR that sentence is false. #528's complaint was
"my converged LP says it failed"; as it stands this trades it for "my LP says
error 2.4e-7 against tol 1e-8 and claims optimal". Printing the floored value
too, or a line when the two differ, would close the loop.

To be fair to the change: this is largely cosmetic, and the reason is a good
argument the PR isn't making. `inf_pr` is `9.81` on **both** builds — at
`|g_l| ~ 1e9` the default `bound_relax_factor = 1e-8` has already moved every
row by ~10. Absolute primal feasibility at this scale was never being measured
to `1e-8` by anything. The forgiveness this PR adds (~`7e-6` on that model) is
six orders of magnitude *tighter* than a relaxation the solver already applies
unconditionally at the same scale. That is the strongest available answer to
"isn't this loosening feasibility?" and it belongs in the doc.

### 3. The equality half of the new code has no test at all

`curr_primal_infeasibility_above_noise` has two branches. The integration test's
LP is `A x >= b` with `g_u = 2e19`, so every row is an inequality, `c.dim() ==
0`, and the `declared_c_rhs()` branch never executes. The three new unit tests
don't reach it either: `amax_above_floor_*` exercise the helper with
hand-supplied floors, and `nlp_error_above_primal_noise_matches_on_ordinary_residuals`
asserts the floored and raw values are *equal*.

So nothing anywhere exercises a floor that actually silences a row through the
real code path, and nothing at all exercises the `c` block — the branch with the
scaling subtlety, where the floor `eps·|c_scale_i·b_i|` has to match `eval_c`'s
`c_scale_i·(g_i(x) − b_i)`. `MockNlp` already carries `c_rhs` and `c_scale`, so
a fixture with `c_rhs = [1e8]` and a residual on either side of `64·eps·1e8`
pins both the branch and the scaling in a few lines.

### 4. The negative test doesn't probe the boundary

`an_infeasible_large_scale_row_is_still_refused` leaves every row short by
`~1e8` against a floor of `~1e-5` — thirteen orders of margin. It would pass
against a floor a trillion times too loose, so it constrains the fix very
little. The question that matters is the near side: is a violation at, say,
`10 · 64·eps·|b|` still refused? Also `assert_ne!(status, SolveSucceeded)`
passes equally on `MaximumIterationsExceeded` or `InvalidNumberDetected`, so it
can't tell "correctly refused" from "fell over".

### 5. The cost note understates the work

`opt_error.rs` says "the two extra `compute_row_amax` sweeps stay off the hot
path". `curr_nlp_error_above_primal_noise()` is `nlp_error(true)`, which
recomputes the *whole* KKT error: `optimality_error_scaling`,
`curr_grad_lag_x`/`_s` (each a fresh allocation plus two mat-vecs, uncached —
`ipopt_cq.rs:621`), all four complementarity vectors, and the eight-vector
`has_valid_numbers` NaN sweep, on top of the two Jacobian sweeps. `nlp_error`
has no entry among the caches at `ipopt_cq.rs:79`.

It doesn't matter in practice — the laziness confines it to iterations at or
near convergence and `current_passes_strict` runs once per solve — but as
written the sentence would be read as "two extra sweeps" by whoever next
considers reusing the accessor somewhere hotter.

Related: `strict_overall`'s `min` is provably a no-op. `nlp_error(true)` shares
`dual` and `compl` with `nlp_error(false)`, and `amax_above_floor` returns at
most `v.amax()` on every path including its three fallbacks, so
`above_primal_noise <= nlp_err` always. Fine as belt-and-braces; the doc reads
as though the two could cross.

### 6. No lever

The floor is unconditional — no option gates it, and `ROW_NOISE_KAPPA` is a
compile-time constant. A user who wants the old strict-absolute semantics has
nowhere to go. Probably right (the constant's doc shows the verdicts are flat
from 8 to 1024 on the measured set), but it is a permanent widening of the
strict gate with no escape hatch, and worth a deliberate decision rather than a
default.

### Verdict

Approve. The mechanism is sound, the containment via `constr_viol_tol` is real
and holds at every scale, and the units line up where they had to — and the
sweep says it fixes considerably more than the issue it was opened for, at zero
regressions across 1242 models.

Before merge I'd want 3 (half the new logic is shipping with no test) and the
`cresc50` / `launch` extra-attempt behaviour understood. 1 and 2 are accuracy
fixes to claims the PR makes about itself. 4–6 are small. The sweep numbers and
`orthrege` in particular deserve to be in the PR body and the CHANGELOG — they
are a much better argument for this change than the synthetic LP family is.

jkitchin:
All six addressed in `10c3dc5`, plus the merge of `main` (#530 had landed; only `CHANGELOG.md` conflicted, both `[Unreleased]` entries kept). 2686 tests pass, fmt/clippy/both consistency scripts clean.

**3 — the equality branch.** You were right that nothing reached it. Three unit tests now drive `declared_c_rhs()` through the real accessor: sub-quantum residual silenced, above-quantum counted in full (not net of the floor), and the floor riding the row scaling at `k ∈ {1, 4, 0.25}` — which, composed with `declared_c_rhs_carries_the_row_scaling`, pins the chain you flagged. `x = (4, 3)` puts `d − s` at an exact `0` so the `c` block is what the accessor returns.

**`cresc50` / `launch`.** Your guess was right, and the specific line is `application.rs:1761`. The ℓ₁ outer loop breaks when an inner solve returns non-Optimal — "escalation won't fix numerical failure at this ρ" — and the old false `Search_Direction_Becomes_Too_Small` was being read as exactly that failure. Now the inner solve certifies honestly, `inner_ok` is true, the slacks still don't collapse (the model really is infeasible), ρ is under `l1_penalty_max`, so it escalates and runs the round it was always designed to. Same verdict, and the base's 2-attempt "saving" was the defect cutting the loop short.

**4 — the boundary.** Writing this turned up two things worth more than the test:

- At `ub = 1e9` the default `bound_relax_factor = 1e-8` absorbs a `7.8e-3` shortfall outright, so my first attempt returned `Solve_Succeeded` and was *correct* to — the model is feasible as posed. That is your note 2's argument arriving as a test failure. The test pins the factor to `0`.
- With the shortfall under `constr_viol_tol` and over the quantum, the point is still refused **at `ROW_NOISE_KAPPA` raised a thousandfold** — a floor that silences the row entirely. On a model with no feasible point the filter and restoration reach a verdict on their own criteria and the convergence gate never votes. So the floor cannot buy an infeasible model a certificate at *any* kappa. That is a stronger containment than the `constr_viol_tol` one, and it is why the floor's own boundary is pinned by unit tests rather than through a solve. Both asserts now also distinguish "refused" from "fell over", and check the returned point genuinely violates.

**1 and 2 — the claims.** Body and CHANGELOG rewritten around your sweep: LP/QP bit-identical model by model, 8 Vanderbei improvements with `orthrege` (2652 → 84) leading, trajectories changing on nonlinear problems, and the `cresc50` cost stated rather than omitted. The stale `application.rs:2504` comment is fixed, and your `bound_relax_factor` argument is now in both the code comment and `docs/src/options.md` — you're right that it is the strongest available answer to "isn't this loosening feasibility?".

**5.** The cost note now says what `nlp_error(true)` actually recomputes, and `strict_overall`'s doc says the `min` is provably a no-op on finite input, kept so the gate can't be loosened by an unnoticed change to either accessor.

**6 — the lever.** Added rather than argued away: **`primal_noise_floor_kappa`**, default `64`, `0` disables the floor and restores upstream's bare-absolute term bit-for-bit. It threads through the accessor (rescaling the placement floor too, so both contributions carry it), short-circuits so the opt-out costs nothing, and `switching_the_noise_floor_off_restores_the_old_verdicts` pins that the issue's LPs lose their certificates again with it off — the one test that would still pass on a revert, which is the point of it.

Also took the reporting half of note 2: the summary prints the tested aggregate under `Overall NLP error` **only when the two differ**, so `O(1)` models and floor-off runs keep the block byte-identical to upstream's, and a large-`|b|` solve stops reading as `error 2.4e-7` beside `Optimal` with nothing accounting for it.

Not done: the `1e11` regime is still unpinned, and `constr_viol_tol` below a row's ulp still won't certify — both unchanged and deliberate.

---

PR #544, opened 2026-08-08
fix(linsol): don't answer an unmeasurable inertia test with delta_w (#540)

### LINKED ISSUE #540: eigena2: inertia correction re-escalates at the final iteration, losing the superlinear tail (Acceptable instead of Optimal)
`eigena2` (Vanderbei) stops at `Solved To Acceptable Level` with dual infeasibility stuck at ~3e-7, where Ipopt certifies `Optimal` at 5.2e-9. The final objective is correct to 12 digits — this is purely a failure to converge the last two orders of magnitude of the dual residual.

The interesting part is that POUNCE follows Ipopt's endgame almost exactly and then diverges in the inertia correction at the second-to-last iteration.

## Reproduce

Both at `270a0502`, single-threaded (`OMP_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 VECLIB_MAXIMUM_THREADS=1 RAYON_NUM_THREADS=1`):

```
./target/release/pounce $POUNCE_BENCH_DATA/vanderbei/nl/eigena2.nl
ipopt eigena2          # homebrew ipopt 3.14, same .nl
```

## Final status

```
              POUNCE (29 it)              Ipopt (27 it)
objective     8.2499999999748326e+01      8.2500000000000160e+01
dual inf      3.3144912958031115e-07      9.3066869792816317e-09
constr viol   2.4123147923660326e-11      2.4424906541753444e-15
complement.   0.0                         0.0
EXIT          Solved To Acceptable Level  Optimal Solution Found
```

Feasibility is at machine precision and complementarity is exactly zero in both. Only `dual_inf` separates them, and it needs to clear `tol = 1e-8`.

## Where they diverge

Both solvers hit a line-search stall in the same place and both enter restoration from an *already feasible* point — POUNCE at `19r` with `inf_pr = 2.41e-11`, Ipopt at `18r` with `1.13e-09`. That part is arguably its own question, but it is not what breaks this run: both recover, and both then walk the inertia-correction parameter down the same ladder.

Line the `lg(rg)` columns up after restoration exit:

```
Ipopt    2.0   1.5   1.0   0.5   0.1  -0.4  -0.9  -1.4      -> dual inf 5.17e-09, EXIT Optimal
POUNCE   2.1   1.6   1.1   0.6   0.2  -0.3  -0.8   1.4  1.0 -> dual inf 7.73e-08, EXIT Acceptable
                                             ^^^^  ^^^
```

POUNCE tracks Ipopt to within 0.1 in `lg(rg)` for six consecutive iterations. Then:

```
POUNCE
  25  8.2500000e+01 1.42e-14 1.12e-06   -8.6 1.14e-07    0.2 1.00e+00 1.00e+00h  1
  26  8.2500000e+01 2.44e-15 2.93e-07   -8.6 4.52e-08   -0.3 1.00e+00 1.00e+00h  1
  27  8.2500000e+01 1.11e-15 4.63e-07   -8.6 2.23e-08   -0.8 1.00e+00 1.00e+00h  1   <-- dual inf INCREASES
  28  8.2500000e+01 4.44e-16 2.24e-07   -8.6 8.20e-09    1.4 1.00e+00 1.00e+00h  1   <-- delta_w jumps back up
  29  8.2500000e+01 8.88e-16 7.73e-08   -8.6 8.47e-09    1.0 1.00e+00 1.00e+00h  1

Ipopt
  25  8.2500000e+01 8.70e-14 1.31e-06  -8.6 2.63e-07   -0.4 1.00e+00 1.00e+00h  1
  26  8.2500000e+01 2.15e-14 3.39e-07  -8.6 1.27e-07   -0.9 1.00e+00 1.00e+00h  1
  27  8.2500000e+01 2.44e-15 5.17e-09  -8.6 1.23e-07   -1.4 1.00e+00 1.00e+00h  1   <-- superlinear, done
```

At iteration 27 POUNCE's step *increases* dual infeasibility (2.93e-07 -> 4.63e-07). The inertia correction responds by re-escalating `delta_w` from `10^-0.8` to `10^1.4` — roughly 25, versus the `10^-1.4` ~ 0.04 Ipopt is using at the same point. That is ~250x more regularization added to the Hessian, which damps the Newton step into a near-gradient step: `||d||` collapses to 8.2e-09 and 8.5e-09 at iterations 28-29 while Ipopt is still taking 1.23e-07 steps. The superlinear tail never happens.

Note also that the run is walking *down* the correct ladder — the regularization is not stuck high from the start. It reaches a good `delta_w`, takes one bad step, and then the escalation logic throws away the progress. So this looks like the `delta_w` update rule reacting to a single non-monotone iteration rather than a systematically wrong choice of regularization.

## Why it matters

`eigena2` is one of the rows in the report's **Regressions (Ipopt Optimal, POUNCE not Optimal)** table. The four convergence-criteria PRs that just merged (#532, #533, #534, #535) were aimed at exit criteria and could not help here — #533's acceptable-streak test does fire, pushing the run from 18 to 29 iterations before it gives up, but the extra iterations do not converge because the step itself is over-damped. The fix has to be in the step computation.

Sister issue: `eigenb2` fails on the same suite with what looks like the opposite symptom in the same subsystem — it applies *no* regularization at an iteration where Ipopt applies `10^2.9`. Worth checking whether both trace to one root cause in the inertia-correction update.

### LINKED ISSUE #541: eigenb2: no inertia correction applied at iteration 3 where Ipopt applies delta_w=10^2.9; run stalls and needs 67 iterations vs 21
`eigenb2` (Vanderbei) stops at `Solved To Acceptable Level` with dual infeasibility stuck at ~4.7e-7 after 67 iterations; Ipopt certifies `Optimal` in 21. The objective is right — POUNCE's `1.6000000000000176` is actually closer to the exact `1.6` than Ipopt's `1.60000000000722`. This is purely a convergence-path problem.

The first three iterations are bit-identical to Ipopt. They split at iteration 3, in the inertia correction.

## Reproduce

Both at `270a0502`, single-threaded (`OMP_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 VECLIB_MAXIMUM_THREADS=1 RAYON_NUM_THREADS=1`):

```
./target/release/pounce $POUNCE_BENCH_DATA/vanderbei/nl/eigenb2.nl
ipopt eigenb2          # homebrew ipopt 3.14, same .nl
```

## Final status

```
              POUNCE (67 it)              Ipopt (21 it)
objective     1.6000000000000176e+00      1.6000000000072200e+00
dual inf      4.6872471493042411e-07      4.1531569294761539e-09
constr viol   4.4408920985006262e-16      1.4921397450962104e-13
complement.   0.0                         0.0
EXIT          Solved To Acceptable Level  Optimal Solution Found
```

## The divergence: iteration 3

Iterations 0-2 agree to every printed digit — same objective, same `inf_pr`, same `inf_du`, same `||d||`, same `lg(rg)`:

```
   1  5.9533607e+02 2.26e+00 2.37e+02  -1.0 5.56e-01   2.0 1.00e+00 1.00e+00f  1
   2  2.3547661e+02 3.97e-01 4.61e+02  -1.0 4.10e-01   3.3 1.00e+00 1.00e+00f  1
```

Then:

```
POUNCE  3  1.5991320e+01 3.56e-02 9.09e+01  -1.0 3.28e-01     -  1.00e+00 1.00e+00f  1
Ipopt   3  1.3159697e+02 3.59e-02 1.55e+02  -1.0 1.60e-01   2.9  1.00e+00 1.00e+00f  1
                                                            ^^^
```

Ipopt applies `delta_w = 10^2.9` ~ 800. POUNCE applies **none** (`-` in the `lg(rg)` column) and takes a step twice as long, `||d|| = 3.28e-01` vs `1.60e-01`. From the same iterate, with the same KKT matrix, the two solvers disagree about whether the matrix needs inertia correction at all.

Superficially POUNCE's step looks better — it lands at objective 15.99 against Ipopt's 131.6, much nearer the optimum of 1.6. But it is an unregularized step in a region where the reduced Hessian is not positive definite, and the run never recovers a clean Newton path afterward.

## Downstream: the stall

From iteration 10 onward POUNCE cannot take a full step. Thirteen iterations accept `alpha_pr` of 1/8 to 1/64 after 4-7 line-search trials, and ten consecutive iterations make essentially no progress at all:

```
  48  1.6000016e+00 9.87e-07 1.61e-05   -8.6 4.10e-03      - 1.00e+00 1.56e-02h  7
  ... (identical to 3 digits for ten iterations, alpha_pr = 1/64, ls = 7) ...
  57  1.6000015e+00 9.27e-07 1.40e-05   -8.6 4.15e-03      - 1.00e+00 1.56e-02h  7
  58  1.6000552e+00 3.45e-05 3.48e-06   -8.6 4.16e-03      - 1.00e+00 1.00e+00w  1   <-- watchdog breaks out
  59  1.6000000e+00 2.98e-10 1.49e-03   -8.6 6.40e-06    2.4 1.00e+00 1.00e+00h  1
```

`inf_pr` moves from 9.87e-07 to 9.27e-07 across ten iterations. The watchdog step at 58 escapes and the run converges hard to `inf_du = 1.88e-07` by iteration 64 — then stalls again at 65-67 with `ls = 7` and `ls = 6` and terminates on the acceptable criteria.

Ipopt, by contrast, takes `alpha_pr = 1.00e+00` with `ls = 1` on *every single iteration* from 0 to 21, never enters restoration, and drives `lg(rg)` monotonically to zero (-3.4, -3.8, -4.3, -4.8, -5.3). There is no stall to recover from because it never leaves the good path.

So the line-search stall is a symptom, not the bug. The bug is upstream at iteration 3.

## Why it matters

`eigenb2` is one of the rows in the report's **Regressions (Ipopt Optimal, POUNCE not Optimal)** table. The recently merged convergence-criteria work (#532, #533, #534, #535) cannot reach this — the exit criteria are behaving correctly given the iterate they are handed; the iterate is what is wrong. The 67-vs-21 iteration count is also a straightforward 3x performance cost on top of the missed certification.

Sister issue: `eigena2` on the same suite fails in the same subsystem with the opposite symptom — it re-escalates `delta_w` to `10^1.4` at the second-to-last iteration where Ipopt is at `10^-1.4`, and loses its superlinear tail. Worth checking whether both trace to one root cause in the inertia-correction update.

jkitchin:
Fixes #540.
Fixes #541.

## Problem

CUTE `eigena2` stops at `Solved To Acceptable Level` with the dual infeasibility stuck two orders above `tol`. The objective is already correct to twelve digits; only the last two orders of the dual residual are missing.

| | status | objective | dual inf | constr viol | iters |
|---|---|---|---|---|---|
| before | `Solved_To_Acceptable_Level` | 8.2499999999748326e+01 | 3.3144912958031115e-07 | 2.41e-11 | 29 |
| after | `Solve_Succeeded` | 8.2499999999999986e+01 | 5.7300837831461680e-10 | 2.22e-16 | 27 |
| Ipopt 3.14 (homebrew, MUMPS) | `Optimal Solution Found` | 8.2500000000000160e+01 | 9.3066869792816317e-09 | 2.44e-15 | 27 |

Ipopt's iteration count, at a residual an order of magnitude tighter.

The reported symptom is `δ_w` re-escalating from `10^-0.8` to `10^1.4` at the second-to-last iteration, damping the Newton step from `1.2e-7` to `8.2e-9`.

## Cause

**Not the `δ_w` update rule.** It is an exact port of upstream's `get_deltas_for_wrong_inertia` and every rung of that ladder is what upstream's rule produces from the state it was in. The issue's reading — "the escalation logic throws away the progress after a single non-monotone iteration" — describes the consequence. The input was wrong.

`--dump kkt:all`, then an exact `numpy.linalg.eigvalsh` on each dumped triplet (dump's own iteration numbering; the last four rows are the four factorizations of the iteration that breaks the run):

| iter | δ_x | feral `n_neg` | exact `n_neg` | `\|λ\|` below 1e-10 | min `\|λ\|` |
|---:|---|---:|---:|---:|---|
| 020 | 0 | 64 | 61 | 51 | 5.1e-16 |
| 022 | 0 | 60 | 47 | 53 | 4.7e-18 |
| 024 | 0 | 56 | 40 | 54 | 9.2e-19 |
| 026 | 0 | 64 | 52 | 54 | 6.6e-19 |
| 026 | 5.35e-2 | 58 | 45 | 45 | 1.1e-18 |
| 026 | 4.28e-1 | 62 | 62 | 45 | 7.1e-11 |
| 026 | 27.4 | 55 | 55 | 45 | 7.1e-11 |

The expected count is 55 on a `165 × 165` KKT. Down the whole tail **45–54 eigenvalues sit at `1e-16` or below on a matrix of norm 240** — the system is singular to working precision and neither middle column means anything. Note where they start agreeing: the last two rows, the factorizations taken *after* `δ_c` was applied.

`eigena2`'s constraint Jacobian degenerates as the iterate converges. Its singular values at iteration 27 are ten at `O(1)` and forty-five at `~1e-8`, halving every iteration. This is a property of the model — the same collapse happens under `nlp_scaling_method=none`. With `δ_c = 0` those near-dependent directions produce no pivot at all; with `δ_c = 7.07e-11` each contributes exactly one negative pivot, the count becomes 55, and the smallest pivot rises to `5.8e-9`.

Three routes to `δ_c` were all closed:

1. **`jac_degenerate = Degenerate`** — the probe only runs while a flag is `NotYetDetermined`, and both latch to `NotDegenerate` on the first successful unperturbed factorization (iteration 0, where the Jacobian is full rank). A Jacobian that degenerates only near the solution can never be detected. Upstream has the same gap.
2. **`residual_ratio_singular` → `pretend_singular`** — refinement converges here (the RHS happens to lie in the range), so the ratio stays under `residual_ratio_max` and this never arms.
3. **The linear solver reporting `Singular`** — `feral_singular_pivot_floor` (`1e-20`, the MA57 `CNTL(2)` value) is far below the `~1e-16` pivots here, *and* it was tested after the inertia comparison, so on a matrix that is both singular and reports a wrong count — every one of them — `WrongInertia` preempted it. It could only fire where the caller does not need it.

Upstream's MA27 / MA57 / MUMPS interfaces all test singularity before comparing the count (`IpMumpsSolverInterface.cpp`: `if (error == -10) return SYMSOLVER_SINGULAR;` sits above `negevals_ = ...`). Ipopt-with-MUMPS takes route 3 on this model; pounce did not.

## Fix

New `feral_inertia_pivot_floor` (default `1e-12`): when the count *already* disagrees and the smallest accepted pivot is under the floor, report `Singular` instead of `WrongInertia`, so `δ_c` is applied before the ladder starts.

Deliberately a **second, higher floor** rather than a raise of `feral_singular_pivot_floor`, and the distinction is the whole safety argument:

- `feral_singular_pivot_floor` says *this factor is unusable*, and applies to factors that were otherwise going to succeed. Raising it can turn a good factorization into a failure.
- `feral_inertia_pivot_floor` says *this factor cannot measure inertia*, and is consulted only on factors already going to be rejected. It cannot cost a usable factorization — it only chooses which perturbation repairs the rejection.

**Rejected first:** raising `feral_singular_pivot_floor` to `1e-14` also fixes `eigena2`, and was dropped. Measured min-pivot values on the accepted (right-inertia) factorizations down the same tail are `1.7e-12, 4.2e-13, 1.1e-13, 3.4e-14, 4.2e-15, 1.3e-15` — overlapping the rejected ones, so any threshold that catches the bad ones also fails good ones, and the fix would have been a trajectory perturbation dressed as a criterion. Also rejected: adding `δ_c` after N `δ_w` escalations within one system (measured at N = 1, 2, 3 — no effect; by then the iterate is already committed), and a `min/max` pivot ratio test (`max|pivot|` reaches `2e15` on factors that are fine, so the ratio is not a stable scale reference — the existing code comment warning against it is correct).

`1e-12` is the middle of the `n·eps` range (`2.2e-15` at `n = 10` to `2.2e-10` at `n = 10^6`) over which an equilibrated pivot loses its sign. Sensitivity is flat across that band and beyond:

| floor | iters | dual inf | status |
|---|---:|---|---|
| 0 | 29 | 3.31e-07 | Acceptable |
| 1e-16 | 29 | 3.31e-07 | Acceptable |
| 1e-15 | 27 | 1.08e-09 | Optimal |
| 1e-14 | 26 | 5.72e-09 | Optimal |
| 1e-13 | 26 | 3.98e-09 | Optimal |
| 1e-12 | 27 | 5.73e-10 | Optimal |
| 1e-10 | 32 | 8.85e-09 | Optimal |

`PerturbForSingularity` also no longer assumes the degeneracy probe is pristine when a `Singular` verdict arrives. Those are `DBG_ASSERT`s upstream — assumptions, not invariants, since MUMPS `INFO(1) = -10` and MA27 `IFLAG = 3` fire from any rung too. The probe is abandoned and the determined-state path runs, keeping the `δ_w` rung already paid for.

## Blast radius

The trigger fires only on a factorization that already mismatched, so no currently-successful factorization can become a failure; the change is which perturbation a rejection is repaired with. `0` restores the previous routing exactly. Mirrored in the Schur backend, which already ordered its singularity check first.

The routing change can still alter a *trajectory*, which the per-factorization
argument does not by itself cover: `coshfun` and `steenbrc` move from
`Maximum_Iterations_Exceeded` to `Error_In_Step_Computation`. Neither solver
solves either model (Ipopt: `Maximum_Iterations_Exceeded` and `Unknown_Error`),
so no scored row moves, but the failure class does.

**Corpus sweep: run.** Vanderbei, all 733 models, `main` @ `270a0502` vs this
branch, identical harness, single-threaded, `max_cpu_time=300`.
**`Solve_Succeeded` 537 -> 541 (+4), zero losses.**

| model | Ipopt 3.14 | main | this PR |
|---|---|---|---|
| `eigena2` | Optimal 82.50000000000016 | Acceptable, 82.4999999997483, 29 it | Optimal, 82.4999999999999986, 27 it |
| `eigenb2` | Optimal 1.60000000000722 | Acceptable, 1.6000000000000176, 67 it | Optimal, 1.6000000000000008, 68 it |
| `orthrds2` | Optimal 1544.296623334707 | Acceptable, 1461.040832, 308 it | Optimal, 1562.822977, 82 it |
| `brainpc2` | MaxCpuTime 5.293887e-04 | Acceptable, 1.783561e-04, 1003 it | Optimal, 4.333174e-04, 618 it |

Two further models already reported `Solve_Succeeded` on `main` — at the wrong
point. A wrong answer under a strict certificate is a worse failure than a
missing certificate, so these matter as much as the headline (unscaled):

| model | Ipopt (Optimal) | main | this PR |
|---|---|---|---|
| `haldmads` | 0.033283191745894 | 0.032996977590367 (0.9% off, 69 it) | 0.033283194386317 (113 it) |
| `lch` | -4.3182888009530025 | -4.2877177671659830 (0.7% off, 24 it) | -4.3182888058762048 (53 it) |

Cost: `brainpc5` goes 44 -> 982 iterations (still `Solve_Succeeded`) — 22x, and
essentially the whole +985 iteration delta across commonly-solved vanderbei
models.

**The other three corpora were swept too**, same harness and settings:

| corpus | models | `Solve_Succeeded` main -> PR | gains | losses |
|---|---:|---|---:|---:|
| `vanderbei` | 733 | 537 -> 541 | 4 | 0 |
| `mittelmann` | 47 | 39 -> 39 | 0 | 0 |
| `qp` | 138 | 1 -> 1 | 0 | 0 |
| `lp` | 371 | 2 -> 2 | 0 | 0 |

Zero losses across all 1289 models. `qp` and `lp` carry little signal — pounce
solves 1/138 and 2/371 of them before this change — so their "no change" rows
are close to vacuous and should not be read as 509 models of coverage.

`mittelmann` adds two more iteration-cost rows, both still `Solve_Succeeded`:
`arki0009` 355 -> 881 (+526) and `arki0003` 329 -> 433 (+104). So the iteration
cost of this change is four models — `brainpc5`, `pooling_rt2stp`, `arki0009`,
`arki0003` — not one.
`orthrds2` and `brainpc2` certify at a *higher* objective than the uncertified
answer `main` returned; both problems are nonconvex, so a different local
minimum is not a defect, but those two rows change category rather than simply
clear.

**`eigenb2` (#541) is fixed too.** It was not investigated here — the sweep
found it. That is the evidence for the sister-issue conjecture in #540/#541
that both trace to one root cause in the inertia correction.

## Tests

Seven new tests. All were run against the parent commit `270a0502` and fail there, each for its stated reason:

| test | on `270a0502` |
|---|---|
| `eigena2_converges_to_optimal` | FAILED — `SolvedToAcceptableLevel`, dual inf 1.84e-7 |
| `eigena2_dual_infeasibility_clears_the_strict_tolerance` | FAILED — `1.841384053223951e-7 did not clear tol = 1e-8` |
| `untrustworthy_inertia_is_reported_singular_not_wrong` | FAILED — `left: WrongInertia, right: Singular` |
| `a_second_singular_verdict_from_the_ladder_does_not_trip_the_probe` | FAILED — panics at `pd_perturbation.rs:293`, the `DBG_ASSERT` |
| `singular_after_a_delta_x_step_abandons_the_probe` | FAILED — probe carried into `finalize_test` |
| `disabling_the_trigger_reproduces_the_reported_failure` | passes (by construction — it asserts the pre-fix behaviour) |
| `well_conditioned_inertia_mismatch_still_reports_wrong_inertia` | passes (guards the property, does not test the fix) |

The last two are deliberate: the first is the anti-vacuity guard (if a later change fixes `eigena2` some other way, it fails and says the headline tests no longer pin what they describe), the second pins that the ordinary negative-curvature signal the `δ_w` ladder exists to answer is untouched.

`fixtures/eigena2.nl` is the reporter's own `.nl`, attached to the issue. `issue_534_resto_decline_progress.rs` notes `eigena2` was not reproducible in-tree because the archive is gitignored — it is now.

`cargo test` (default members): **2758 passed, 0 failed**. `cargo fmt --all -- --check` clean; the CI clippy gate (`-D clippy::correctness -D clippy::suspicious`) passes on the touched crates; `scripts/check-release-consistency.sh` OK.

Full analysis: `dev-notes/issue-540-eigena2-inertia-noise.md`.

---

- [x] Book page under `docs/src/` updated (`options.md`: option table + env-override table; no new page, so no `SUMMARY.md` change)

https://claude.ai/code/session_01VU6SSip8x4G3CvmgvToCfX

---
_Generated by [Claude Code](https://claude.ai/code/session_01VU6SSip8x4G3CvmgvToCfX)_


jkitchin:
## Review: approve

I verified the argument and then ran the one thing the PR says it could not — the corpus sweep. The archive was available here.

### The reasoning holds

The safety claim is the load-bearing one and it checks out in both backends. The floor is consulted only inside `if check_neg_evals && self.negevals != number_of_neg_evals` (`lib.rs`) and `if check_neg_evals && self.negevals != num_neg_evals` (`schur.rs`), so it is unreachable on a factorization the caller was going to accept. It changes which perturbation repairs a rejection, not whether one happens. `0` restores the previous routing, and the test pins both directions on the same matrix.

The `probe_intact` guard is the right shape too. Replacing the `DBG_ASSERT`s with a state test and falling through to the determined-state path is more defensible than upstream, since the preconditions genuinely are not invariants.

### Corpus sweep — Vanderbei, 733 models

Both binaries built from source, `main` @ `270a0502` vs PR head `a0362be4`, identical harness, single-threaded (`OMP_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 VECLIB_MAXIMUM_THREADS=1 RAYON_NUM_THREADS=1`), `max_cpu_time=300`.

**`Solve_Succeeded` 537 to 541 (+4). Zero losses.**

| model | Ipopt 3.14 reference | main | this PR |
|---|---|---|---|
| `eigena2` | Optimal 82.50000000000016 | Acceptable, 82.4999999997483, 29 it | **Optimal, 82.4999999999999986, 27 it** |
| `eigenb2` | Optimal 1.60000000000722 | Acceptable, 1.6000000000000176, 67 it | **Optimal, 1.6000000000000008, 68 it** |
| `orthrds2` | Optimal 1544.296623334707 | Acceptable, 1461.040832, 308 it | **Optimal, 1562.822977, 82 it** |
| `brainpc2` | MaxCpuTime 5.293887e-04 | Acceptable, 1.783561e-04, 1003 it | **Optimal, 4.333174e-04, 618 it** |

### Three findings the body should record

**1. This also fixes #541, which the PR explicitly disclaims.**

The body says "`eigenb2` (#541) was **not** investigated and this PR makes no claim about it." It fixes it anyway: `Solved_To_Acceptable_Level` to `Solve_Succeeded`, and at a tighter objective than Ipopt's. That is consistent with the diagnosis — I filed #540 and #541 as sister issues and suggested they might share a root cause in the inertia correction, and this is the evidence that they do.

Please add `Fixes #541` to the body. Without a closing keyword in the body the issue is left dangling after merge even though the fix is in.

**2. Two silent-correctness wins that are arguably bigger than the headline.**

Both of these already reported `Solve_Succeeded` on `main` — at the wrong point. Unscaled objectives:

| model | Ipopt (Optimal) | main | this PR |
|---|---|---|---|
| `haldmads` | 0.033283191745894 | 0.032996977590367 (0.9% off, 69 it) | **0.033283194386317** (113 it) |
| `lch` | -4.3182888009530025 | -4.2877177671659830 (0.7% off, 24 it) | **-4.3182888058762048** (53 it) |

`main` was certifying a worse local minimum on both; the PR lands on Ipopt's answer to nine digits. A wrong answer under a strict certificate is a worse failure than a missing certificate, so this is worth stating explicitly.

**3. Costs, so they are on the record.**

- `brainpc5`: 44 to 982 iterations, still `Solve_Succeeded`. 22x, and essentially the entire +985 iteration delta across all commonly-solved models. The single worst outcome in the sweep.
- `coshfun` and `steenbrc` change failure class, `Maximum_Iterations_Exceeded` to `Error_In_Step_Computation`. Neither solver solves either model (Ipopt: MaxIter and `Unknown_Error` respectively), so no scored row moves — but it does show the routing change can alter a trajectory into a breakdown, which the per-factorization safety argument does not by itself cover. Worth a sentence in Blast radius.
- `orthrds2` and `brainpc2` certify at a *higher* objective than the uncertified answer `main` returned, and `orthrds2`'s 1562.82 is above Ipopt's 1544.30 on a minimization. Both problems are nonconvex, so a different local minimum is not a defect — but if the report scores objective agreement, these two rows will move category rather than simply clear.

Remaining suites (Mittelmann, QP, LP) are running and I will follow up; nothing in them is expected to move, since every model that changed here is in Vanderbei.

### Verdict

Approve. The evidence supports the change, the sweep the body could not run comes out net positive with no losses, and the two hidden correctness fixes strengthen the case. The only thing I would block on is the missing `Fixes #541`.

Issue #552, opened 2026-08-08
Performance report: dynamic optimization, pounce vs Ipopt (MA27 / MA57 / MUMPS)


devin-griff:
# Performance report: dynamic optimization, pounce vs Ipopt (MA27 / MA57 / MUMPS)

This report compares pounce against Ipopt, using each of Ipopt's three
linear solvers, on five dynamic optimization models from a Pyomo NMPC
package: two chemical process flowsheets and three textbook ODE control
problems. Every solve ran on an otherwise idle machine, one at a time,
from the same starting point, with the same options.

Two results stand out. pounce is the fastest and the most robust solver
on both flowsheet models, which are the only ones in the set that reach
interesting size naturally, and it is the only solver that returns an
optimal solution on all of them. Its factorization is slower than
Harwell's on chain-structured KKT matrices, by a margin since confirmed
by direct measurement in the thread below.

## Setup

* pounce 0.9.0, release build, FERAL backend (the default).
* Ipopt 3.13.2 from `idaes get-extensions`, which carries HSL, driven
  through Pyomo's `ipopt_v2`. MA27, MA57 and MUMPS were each confirmed
  live from the solver banner.
* Windows 11, one machine, nothing else running. Each figure is the
  median of three runs after a discarded warmup.
* Where scaling is used it is `nlp_scaling_method=user-scaling` with
  factors carried in the `.nl` suffix. Both solvers read that suffix
  natively, so no model is cloned and nothing differs between solvers
  except the solver.
* **A note on clocks.** pounce's timing report prints wall-clock
  seconds. Ipopt prints CPU first and wall in parentheses
  (`0.180 (sys: 0.000 wall: 0.180)`). Reading Ipopt's first number
  against pounce's compares CPU to wall, which is wrong in a way that
  penalises pounce whenever the machine is busy. Every figure below is
  wall clock on both sides, and the totals are measured externally
  around the solve call so they are directly comparable.

## The models

Measured at the largest horizon of each, with the range of sizes tested
alongside:

| model | horizons tested | n at largest | time points | vars / time point |
|---|---|---|---|---|
| hicks | N=10 to 100 | 1,706 | 301 | 5.7 |
| cart_pole | N=10 to 100 | 2,810 | 301 | 9.3 |
| quad_tank | N=15 to 100 | 2,910 | 301 | 9.7 |
| double_column | N=4 to 10 | 25,377 | 31 | 818.6 |
[...]

Issue #476, opened 2026-08-04
Performance: robot_a/b/c (m >> n) dominate the Mittelmann geomean gap — ~3x slower than IPOPT, ~2000x slower than KNITRO


jkitchin:
## Summary

On the [Mittelmann ampl-nlp benchmark](https://plato.asu.edu/ftp/ampl-nlp.html) run of
**3 Aug 2026**, POUNCE-0.9.0 solved **47/47** — a perfect robustness score matched only by
KNITRO-16.0 and COPT-8.0.0 — but placed 5th of 9 on the scaled shifted geometric mean (15.0,
vs COPT 1.00 / KNITRO 1.17 / IPOPT 11.6 / WORHP 11.8).

That geomean gap is not diffuse. It is concentrated in a handful of instances, and the
`robot_a` / `robot_b` / `robot_c` family is the single largest contributor.

## The numbers

| instance | n | m | POUNCE | IPOPT | KNITRO | COPT | best other |
| --- | --- | --- | --- | --- | --- | --- | --- |
| `robot_a` | 1001 | 52013 | **2140 s** | 688 s | 1 s | 3 s | 1 s |
| `robot_b` | 1001 | 52013 | **2122 s** | 717 s | 1 s | 4 s | 1 s |
| `robot_c` | 1001 | 52013 | **2298 s** | 683 s | 1 s | 3 s | 1 s |
| `robot_1600` | 14399 | 9601 | 5 s | 1 s | 1 s | 1 s | 1 s |

POUNCE is ~3.1× slower than IPOPT and ~2000× slower than KNITRO on the three
constraint-dominated instances. Note the contrast with `robot_1600`, which has the opposite
aspect ratio (n ≫ m) and is fine — so this is specific to the **m ≫ n shape (52013
constraints on 1001 variables, ~52:1)**, not to the robot model family as such.

For scale, these three problems plus `WM_CFy` (7053 s, best other 13 s) and `NARX_CFy`
(1864 s, best other 17 s) account for essentially the entire distance between POUNCE's
geomean and IPOPT's.

## Reproduces locally, and harder

Our own checked-in results show both POUNCE and IPOPT failing outright on all three at
default limits:

`benchmarks/BENCHMARK_REPORT.json`:

```
robot_a  pounce_status=Maximum_CpuTime_Exceeded   ipopt_status=Maximum_Iterations_Exceeded
[...]

PR #553, opened 2026-08-08
perf(nl): dense-Hessian-row O(n^2) fix, gated shared-CSE Jacobian, and Pyomo v2 interop

### LINKED ISSUE #554: Performance: ~2.3x per-iteration cost vs IPOPT+MA57 on collocation-discretized models
## Summary

Split out of #476, which is now closed. That issue was opened on `robot_a/b/c` dominating the Mittelmann geomean at ~3× IPOPT; the cause turned out to be the filter's `theta_max` ceiling (#546, fixed in #550) and those three instances are now ones POUNCE wins outright. What the post-fix sweep leaves is a different, coherent problem that deserves its own title.

**POUNCE costs roughly 2.3× more per iteration than IPOPT+MA57, concentrated in collocation-discretized models.** This is *not* the m ≫ n aspect ratio #476 was named after.

## The measurement

Fresh 47-instance Mittelmann sweep, same machine, single-threaded (`OMP/OPENBLAS/VECLIB/RAYON_NUM_THREADS=1`), 1800 s limit, POUNCE `5b336a5c`+#550 vs IPOPT 3.14.20 `linear_solver=ma57`. Process wall-clock captured identically for both by `benchmarks/scripts/run_nl_bench.sh`.

Robustness is not the problem — POUNCE solves **47/47**, IPOPT+MA57 **41/47**.

Speed on the 41 both-solved:

| measure | POUNCE / IPOPT+MA57 |
| --- | --- |
| shifted geomean, shift 10 s | 1.42× |
| plain geomean of per-instance ratios | 2.07× |
| **matched-iteration subset (n = 21)** | **2.30×** |

The last row is the number this issue is about. Those 21 instances are ones where **both solvers take an identical iteration count**, so the convergence path is factored out and the remaining time ratio is cost per iteration and nothing else.

It is *higher* than the overall 2.07× because we buy part of the deficit back with fewer iterations on some families — the QCQPs, where we are 3–5× faster overall (`qcqp750-2nc` 0.21×, `qcqp1000-2c` 0.25×, `qcqp500-3c` 0.30×). Those wins mask the per-iteration cost in an aggregate geomean, which is why the matched-iteration subset is the right statistic to track.

## Where it concentrates

| instance | ratio | POUNCE | IPOPT+MA57 | iterations |
| --- | --- | --- | --- | --- |
| `steering_12800` | 8.88× | 12.7 s | 1.4 s | **23 / 23 (identical)** |
| `rocket_12800` | 8.33× | 15.8 s | 1.9 s | **30 / 30 (identical)** |
| `qssp180` | 8.78× | 65.4 s | 7.5 s | 27 / 24 |
| `corkscrw` | 5.66× | 16.4 s | 2.9 s | 283 / 182 |
| `NARX_CFy` | 4.68× | 273.2 s | 58.4 s | 625 / 506 (per-iter 3.79×) |

`steering_12800` and `rocket_12800` are the cleanest signals available: identical iteration counts, ~8.5× wall. Whatever this is, it is visible without any convergence-behaviour confound.

## Corroboration from a second, independent axis

@devin-griff measured **factorization in isolation** on collocation-discretized ODE control problems (#552) and found a consistent **3.5–4.8×** against MA27/MA57 — on models with m/n between 0.76 and 1.00, nowhere near `robot_a`'s 52:1. That is the same model family as `steering_12800` / `rocket_12800` here.

The two numbers are consistent: a ~4× penalty confined to the factorization kernel dilutes to ~2.3× across a whole iteration once function/derivative evaluation, barrier updates and the line search are included.

Their sweep also rules out several things, which is worth not repeating:

- ordering (`amd`, `amf`, `metis`), scaling (`identity`, `infnorm`, `mc64`), and the parallel dispatch gate (`feral_min_par_flops`) each move factorization time by only 10–20%, in directions inconsistent between models, against a 3.5–4.8× gap — so this is not a mis-selected heuristic;
- disabling tree-level parallelism entirely changed nothing measurable — so it is not fork/join overhead.

Which points at the numeric kernel itself on this sparsity structure: long chains of small blocks, thousands of time points carrying 6–10 variables each. Note the contrast in their report — on flowsheet models with 819 and 913 variables per time point, we are *faster* than MA27 and level with MA57. The penalty is specific to the many-tiny-blocks shape.

## Suggested starting points

1. Reproduce `steering_12800` / `rocket_12800` under a profiler and attribute iteration time across factorization / evaluation / line search — confirm the kernel is actually the majority before optimizing it.
2. Compare symbolic-analysis and numeric-factorization phases separately against MA57 on the same matrices; #552's stage sweep suggests the cost is numeric, but that was not measured against MA57 phase-by-phase here.
3. Check whether the small-block chain structure is defeating supernode/block detection, given #552's finding that large per-time-point blocks are fine.

## Not in scope

- `arki0009` takes 881 iterations to MA57's 353 (12.57× total, ~5.1× per iteration) — the per-iteration part belongs here, but the iteration count is a separate question and should be its own issue if pursued.
- `nql180` `Solver_Error` at iteration 0 is an IPOPT-side failure, not ours.

Data: `benchmarks/mittelmann/pounce.json`, `benchmarks/mittelmann/ipopt_ma57.json`, methodology in `dev-notes/research/mittelmann-post-546-sweep.md`.

jkitchin:
Fixes #554.

Found while exploring #552. **No closing keyword for #552**: this fixes neither
of the two improvement areas that issue reports (per-solve overhead on small
models; FERAL factorization on chain-structured KKT). It fixes three different
problems found while investigating the first of those. #552 should stay open on
its own terms — though the measurement in commit 3 revises what its overhead
section is actually measuring; details posted there.

It does close #554. The dense-Hessian-row fix in commit 1 is the same root cause
#554's per-iteration gap concentrates on: the coloring collapsed to ~one color
per variable on collocation-discretized models, making `eval_h` — not the linear
solver — the dominant per-iteration term, at O(n²) memory. #556 was opened
independently against #554 with the same mechanism and is closed in favor of
this PR; see the review there for why its fixed `max(64, 8 × avg)` threshold
reintroduces the blowup on mixed degree profiles, and why the bound-based
selection here does not. The residual factorization-kernel gap on fixed-horizon
chains stays with #552.

Four commits, independently reviewable:

1. `fa20c20` — a dense Hessian row cost O(n²) in setup and every `eval_h`
2. `0f9a6a3` — `eval_jac_g` on the shared-CSE tape, gated (#476 follow-on)
3. `13e8417` — Pyomo's modern solver interface could not drive POUNCE at all
4. `73edff5` — document the several Pyomo interfaces, and the timing trap

---

# 1. A dense Hessian row cost O(n²) in setup and every `eval_h`

## Problem

`NlTnlp` recovers the Lagrangian Hessian by graph coloring and holds two
`n_colors × n` dense `f64` arrays (`seeds`, `compressed`). A model with a
single **dense Hessian row** — a variable multiplying a sum over all the
others: a total-cost variable, a shared design parameter, a scalar entering
every constraint — makes every pair of columns conflict, so the greedy walk
hands out one color per variable and both arrays go quadratic. The Hessian
itself stays perfectly sparse throughout.

Measured on `min Σ(xⱼ−1)² + x₀·Σxⱼ` alongside a sparse constraint chain,
`nnz_h ≈ 2n`:

| n = 8,010 | before | after |
|---|---|---|
| colors | 8,010 | **6** |
| peak RSS through setup | 1,038 MB | **59 MB** |
| setup wall clock | 2.69 s | **0.18 s** |

| n = 4,010, 10 iterations | before | after |
|---|---|---|
| `LagrangianHessianEvaluations` | 2.069 s | **0.019 s** |
| `OverallAlgorithm` | 2.385 s | **0.284 s** |
| objective | 3.4255625875830601e+03 | *bit-identical* |

Measured peak RSS matched `2·n²·8` bytes to within 1%, so the two arrays were
the entire footprint. That extrapolated to ~6.4 GB at n = 20,000 and an OOM at
the flowsheet sizes in #552. It is now linear in `n`.

## Cause

Two independent O(n²)s, stacked. Fixing only the first leaves RSS quadratic —
which is what happened on the first attempt here.

1. **The coloring.** `greedy_hessian_coloring` is a distance-1 column coloring:
   columns may share a color only when their nonzero rows are disjoint. A dense
   row puts a nonzero in every column at that row, so no two columns are ever
   compatible. `n_colors` collapses to `n`.

2. **`Tape::hessian_sparsity`.** It kept a `BTreeSet` of dependent variables
   alive for *every* tape slot. A long additive chain — precisely what a dense
   row is built from — leaves one full-size set per partial sum and copies the
   whole running set at each union. `split_top_sums` keeps top-level sums out of
   a single tape, so this only bites when the sum feeds an enclosing operator,
   which is exactly the dense case.

## Fix

**Peeling.** Columns whose nonzero-row count exceeds `DENSE_COL_FACTOR` (16)
times the average — never fewer than `DENSE_COL_MIN` (32) entries — become
*candidates* for peeling. A peeled column leaves the coloring and takes a
singleton color: one product with seed `e_d` recovers all of column `d`, and
because `H[d, j] == H[j, d]` that same pass also supplies every entry of row
`d`. Row `d` therefore stops constraining anyone else's color and drops out of
the conflict structure, and the remaining columns color on their real sparsity.
The decode table picks the peeled endpoint's pass when a pair has one.

**Which candidates actually get peeled.** Being far above the average is not the
same as being worth a singleton color. A model with a low-degree bulk and many
medium-degree columns whose row supports are *pairwise disjoint* drags the
average down until the whole medium-degree population clears any fixed
threshold — and peeling them is pure loss, since disjoint supports would have
shared one color. So the candidate list is scored, not thresholded:
`select_peeled_cols` walks a doubling ladder over the degree-sorted candidates
and keeps the prefix minimizing

    peel_color_bound = |peeled| + max surviving row degree

which is a lower bound on the resulting color count (a row with `d` surviving
entries makes those `d` columns pairwise conflict) and costs one O(nnz) pass to
evaluate. `MAX_PEELED_COLS` (256) caps the ladder on top of that.

A plain coloring cannot be used as the comparison baseline here: on the
one-dense-row case the greedy walk is itself O(n²) — for each of `n` columns it
iterates the dense row's `n` columns — which is precisely the blowup peeling
exists to avoid. The decision has to come from a cheap bound.

Measured against a plain greedy coloring of the same pattern, on a degree-3
chain plus pairwise-disjoint dense blocks:

| bulk | blocks | plain | peeled path | peeled |
|---|---|---|---|---|
| 190 000 chain | 100 × 100 | 100 | 100 | 0 |
| 190 000 chain | 200 × 70 | 70 | 70 | 0 |
| 100 000 chain | 50 × 200 | 200 | 200 | 0 |
| 50 000 chain | 20 × 100 | 100 | 100 | 0 |

Exact parity, while the motivating one-dense-row case still collapses to 2
colors (n = 200 and n = 20 000 alike) and three rows of degree 10 000 in a
20 000-variable model still peel — a fraction-of-`n` degree rule would have
refused those.

**Rejected: capping `n_colors` with a fallback evaluator.** It bounds memory but
not the work — the dense summand's tape still gets one directional pass per
color. Peeling removes both, and needs no second code path.

**Slot liveness in `hessian_sparsity`.** Compute last-use per slot up front, then
move a dying operand's set into its consumer instead of copying. A left-leaning
chain now extends one set in place. The `for_each_input` helper driving this is
an exhaustive match with **no catch-all arm**: a future `TapeOp` variant fails to
compile rather than silently reporting "reads nothing", which would drop a live
set and under-report sparsity — a wrong answer that surfaces as a bad search
direction, not a panic. Over-reporting an input is only a missed optimization,
so the doc comment says to list a slot when in doubt.

Also in this commit, as a constant-factor cleanup on the same paths: `parse_expr`
/ `next_data_line` / `eat_segment_header` borrow from the source buffer instead
of allocating a `String` per token and per line (~620k of them for a
20k-variable model); `parse_var_coef` / `parse_bound_line` walk the whitespace
iterator instead of collecting a `Vec` of `&str`; the Hessian-pair,
Jacobian-column and tape-color `BTreeSet`s became `Vec` + `sort_unstable` +
`dedup`; and a `HashMap` built with one entry per Hessian nonzero, read once and
discarded, is gone — decode tables come straight from the sorted pair list,
which also makes the `eval_h` scatter walk `values` forward instead of in
`HashMap` order.

I measured that part against a binary built from the parent commit rather than
estimating it: **total setup on ordinary sparse models moves 1.00–1.07×**. It is
a cleanup, not a speedup; the dense-row case is where it compounds.

---

# 2. `eval_jac_g` on the shared-CSE tape, gated on op ratio

## Problem

`HybridTape` has driven `eval_g` since #476 while `eval_jac_g` stayed on the
flat per-summand tapes, whose CSE bodies are duplicated per reference — so the
4.03× op-count reduction #476 measured was unavailable to the Jacobian.

## Cause of the gate

Wiring it through naively **regresses** low-redundancy models. `eval_g` needs
only values, so the prelude is swept once for the whole constraint block and
the saving is the full op-count ratio. The Jacobian needs a gradient per row,
so only the forward sweep can be shared: the reverse sweep still walks each
summand's `prelude_reach` separately, and it pays a per-op cost the flat tape
does not — a nested `SummandOp` dispatch and an indirected walk over a reach
list instead of a straight loop over a contiguous `Vec` of `TapeOp`.

A clean A/B in one binary (`POUNCE_DBG_NO_HYBRID`), chain model, n ≈ 20,000,
op ratio 2.20×:

| | flat | hybrid |
|---|---|---|
| `eval_g` | 0.975 ms | **0.618 ms** (1.58× faster) |
| `eval_jac_g` | **1.778 ms** | 2.824 ms (1.59× *slower*) |

## Fix

Sweeping body size at CSE redundancy 40 locates the crossover (`eval_jac_g`,
flat → hybrid):

| op ratio | 1.94 | 2.20 | 2.84 | 3.53 | 4.20 | 5.16 | 6.35 | 8.00 |
| speedup | 0.77× | 0.63× | 0.88× | 1.21× | 1.21× | 1.18× | 1.50× | 1.32× |

Crossover near 3, so `HYBRID_JAC_MIN_OP_RATIO` is **4** for margin. With the
gate in place the same sweep reads 1.00–1.03× below it (unchanged, flat path)
and 1.13–1.30× above it. `robot_a` measures 4.03×.

## Not done, and why

**`eval_h` is untouched, and it is the larger prize** — ~69% of AD time on
these models against the Jacobian's ~19%. The dormant
`HybridTape::hessian_summand` is the wrong tool for it: it re-walks
`prelude_reach` once per seed variable, so it shares no prelude work at all,
and it scatters through a `HashMap` lookup per emitted pair where the coloring
path writes into a dense buffer. Given that the *Jacobian* only broke even at
op ratio ≈ 3 with the forward sweep shared, a routine sharing nothing and
adding a hash lookup would be worse. Doing it properly means a new directional
routine that accumulates prelude adjoints across all summands and does one
reverse-over-tangent pass over the prelude per color. That is new AD code, not
a wiring change, and wants its own PR.

---

# 3. Pyomo's modern solver interface could not drive POUNCE at all

`SolverFactory('ipopt_v2', executable=PATH)` — the newer `pyomo.contrib.solver`
interface — failed on every model. Two independent gaps, both places where
POUNCE diverged from what the ASL does and every reader therefore expects:

1. **Quoted option values.** The v2 interface builds each option as
   `option_file_name="/tmp/…/x.opt"` (`_option_to_cmd` in
   `pyomo/contrib/solver/solvers/ipopt.py`) and passes it as one `argv` entry.
   With no shell in between the quotes arrive as literal characters. Ipopt's
   ASL option parser strips them; POUNCE did not, so it looked for a file whose
   name began with `"`, failed to load it, and aborted the run. `parse_kv` now
   drops one matching pair of surrounding `"` or `'`; a quote on one side only
   stays content.

2. **The `.sol` `Options` block.** POUNCE wrote a count of `0`. AMPL and
   Pyomo's *legacy* reader accept that, but no ASL solver emits it: the v2
   reader reads the first two option words unconditionally (to detect the
   documented quirk where a second word of `3` means two extra words follow the
   `z` block) and so asserts `n_opts >= 2`. The block is not a constant: a
   solver echoes back **the model's own AMPL option words**, which arrive on
   `.nl` header line 0 as `g<count> <opt0> <opt1> …`. `NlProblem` now carries
   them and the writer emits them verbatim. Checked byte-for-byte against
   Ipopt 3.14.20 `-AMPL` on three models — `arki0003` (`3 10 1 0`),
   `bearing_400` (`3 1 1 0`), `camshape_6400` (`3 10 1 0`) — POUNCE now writes
   what Ipopt writes on each. A hard-coded `3 / 1 / 1 / 0` would have matched
   only `bearing_400`. Models with no usable header option list fall back to
   `3 / 1 / 1 / 0`, which keeps `n_opts >= 2` satisfied.

With both fixed, one model solved through `SolverFactory('pounce')` and through
`SolverFactory('ipopt_v2', executable=PATH)` returns the same objective to 12
digits and identical primal values. This also makes it possible to benchmark
POUNCE against Ipopt through *one* Pyomo interface, which #552's overhead
comparison could not do — that is how these were found.

---

# 4. Documenting the interfaces, and the timing trap

Pyomo has several NL/SOL solver interfaces and they are different code paths,
not aliases. Verified against Pyomo 6.10.1 by running each one, with `PATH`
below standing for the path to the POUNCE binary:

| call | works | carries `pyomo-pounce`'s extras |
|---|---|---|
| `SolverFactory('pounce')` | yes | **yes** |
| `SolverFactory('ipopt_v2', executable=PATH)` | yes | no |
| `SolverFactory('ipopt', executable=PATH)` | yes | no |
| `SolverFactory('asl', executable=PATH, solver='pounce')` | yes | no |
| `SolverFactory('appsi_ipopt', …)` | no — takes no `executable` | — |

The generic routes run the same solver and return the same answer, but silently
do without the `scaling_factor` suffix handling, the sensitivity path, the
preflight/repair helpers, and the guard against handing a model with live
integer variables to a continuous solver. That is worth a user knowing before
they reach for one.

The section also records the timing trap, since it has already caught someone
(#552): `solver.solve(model)` is Pyomo writing the `.nl`, spawning the process,
reading the `.sol` and loading it. Timing around that call and subtracting the
solver's own reported time leaves a remainder that is mostly *Pyomo's* work,
and it is not the same work on every interface — so comparing
`SolverFactory('pounce')` against `SolverFactory('ipopt_v2', …)` compares two
Pyomo interfaces as much as two solvers.

Two claims came back out of the page while writing it, as unverifiable: which
version number the fixes ship in (it points at the CHANGELOG entry instead),
and that `ipopt_v2` is on track to become the default `ipopt` (that is Pyomo's
roadmap, not something I checked).

---

## Blast radius

Bit-identical except the targeted cases. All **62 `.nl` fixtures** in the
repository return an identical objective and exit status under a parent-commit
binary and this one, including the dense model (same 17 digits, same iteration
count). A banded Hessian peels nothing and colors by its bandwidth exactly as
before; peak RSS on a sparse n = 20,010 chain is 127 MB on both. `pounce verify`
still round-trips a `.sol` POUNCE wrote, and Pyomo's legacy reader — which
accepted the old `Options 0` — accepts the new block too (it is the same shape
every other ASL solver emits).

**These changes do not speed up the models in #552, and I checked rather than
assuming.** Old binary vs new through the same Pyomo path, quad_tank at the
horizons that issue reports, median of 5: 1.00× / 0.99× / 1.03× on total, same
iteration count. The reason is structural — inspecting the `.nl` Pyomo writes
for that model, there are **zero `V` segments** (no shared CSE bodies, so the
hybrid tape is never built and the Jacobian gate cannot fire) and
**`n_colors = 1`** (the coloring was already optimal, so peeling has nothing to
peel), with `avg_ops = 4.0`. The fixes target model shapes these three do not
have. Whether the two flowsheet models in #552 have either shape is unknown and
cheap to check — `POUNCE_DBG_TAPE_STATS=1 pounce model.nl max_iter=0` prints
both numbers — but needs IDAES to build them.

**Coverage gap worth naming:** *no repository fixture has a CSE shared across
constraints* — I checked all 62. So neither the new Jacobian path nor the
existing `eval_g` hybrid path has any fixture coverage; both are covered by
unit tests only, and the 62-fixture sweep proves no regression rather than
proving the new path correct.

`docs/src/pyomo.md` gains a section (commit 4); it is an existing page, so
`SUMMARY.md` is unchanged. No user-facing option is added and no solver default
changes.

## Tests

Hessian coloring:

- `a_dense_hessian_row_does_not_explode_the_coloring` — the real regression
  test. **Verified to fail on the parent commit for the stated reason**: the
  exact model it builds (n = 200) colors in 200 colors there versus 2 here,
  with `nnz_h = 399` both ways, so the `seeds.len() <= 4` assertion fails
  with 200.
- `peeled_dense_column_still_recovers_the_exact_hessian` — every recovered
  entry against the closed-form Hessian, including entries read out of a peeled
  column's pass by symmetry. **Passes on the parent**: a correctness guard on
  the new recovery rule, not a regression test, and the test that would catch a
  wrong peeling rule.
- `a_sparse_hessian_is_colored_by_its_bandwidth_not_peeled` — pins the
  no-regression direction. Does not compile against the parent (the coloring
  signature gained the peeled mask), so "fails pre-fix" does not apply.
- `thousands_of_medium_degree_cols_are_not_peeled`,
  `a_few_truly_dense_rows_are_still_peeled`,
  `peeling_is_capped_and_keeps_the_worst_offenders` — pin the selection rule
  from both sides: a disjoint-block model must peel nothing, three degree-10 000
  rows in a 20 000-variable model must still peel, and the cap must keep the
  highest-degree candidates. The first fails against a fixed threshold with no
  pay-for-itself check.
- `header_option_words_are_kept_verbatim`,
  `a_truncated_option_list_is_dropped_not_padded`,
  `options_block_echoes_the_models_own_words`,
  `options_block_falls_back_when_there_is_no_header` — the header parse and the
  `.sol` emission, including the fallback that keeps `n_opts >= 2` valid.

Jacobian:

- `shared_cse_jacobian_matches_flat_tape_bit_for_bit` — forces the path on
  below the gate; asserts exact equality with the flat tapes plus the closed
  form.
- `a_deep_shared_body_turns_the_jacobian_gate_on_and_still_agrees` — a 40-deep
  shared body over 16 rows trips the gate by itself; asserts exact equality
  with the flat tapes on the gate-on path.
- `a_small_shared_body_leaves_the_jacobian_on_the_flat_path` — pins the gate
  staying off where the hybrid would lose.

Pyomo interop:

- `parse_kv_strips_the_quotes_a_driver_added` — the `key="value"` forms Pyomo
  v2 emits, including the one-sided-quote case that must stay content.
- `writes_basic_primal_dual_block` — updated to pin the ASL option block.

The `for_each_input` exhaustiveness guarantee is enforced by the compiler
rather than by a test, which is the stronger pin — but it means a variant added
with the wrong operand list would still slip through, and no test pins that.
The two Pyomo interop fixes are pinned by unit tests, not by an end-to-end
Pyomo test — Pyomo is not a dev-dependency of this workspace, and I did not add
one for this. The interface table in commit 4 was produced by running each
call against Pyomo 6.10.1 by hand; nothing in CI re-checks it, so it will go
stale when Pyomo renames an interface.

`cargo fmt --all -- --check` clean, `cargo clippy` introduces no new warnings
(checked hunk-by-hunk against the diff ranges), 111 test suites green across
`pounce-nl` / `pounce-cli` / `pounce-algorithm` with zero failures,
`scripts/check-release-consistency.sh` OK. (`hs071_max_cpu_time_terminates`
failed once under concurrent load — it is a CPU-time-limit test; it passes in
isolation and on a clean full run.) CI on this branch is green: 12/12 check runs
on `73edff5`.

## Instruments

Added so the tables above are reproducible rather than asserted:

- `POUNCE_DBG_NO_HYBRID=1` — force the flat per-summand path.
- `POUNCE_DBG_TAPE_STATS=1` now also prints flat-versus-shared constraint op
  counts, which is the gate's own input.
- `cargo run --release -p pounce-nl --example adbench -- model.nl [reps]` —
  times `eval_g` / `eval_grad_f` / `eval_jac_g` / `eval_h` in isolation, with
  no solver dynamics confounding the measurement.

---

- [x] Tests fail on the parent commit for the stated reason — one of eight; the others are scoped above
- [x] `CHANGELOG.md` `[Unreleased]` entries, in the user's terms
- [x] Book page updated — `docs/src/pyomo.md`, an existing page, so `SUMMARY.md` needs no change
- [x] Every claim in this body is true of the code as it stands

https://claude.ai/code/session_01TEe95bNcfxEBpgcvZTvWNx

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,80 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — a single dense Hessian row made `.nl` setup and every `eval_h` O(n²) (#552)
+
+`NlTnlp` recovers the Lagrangian Hessian by graph coloring: columns whose
+nonzero rows are pairwise disjoint share a color, and one directional
+product `H·s_c` per color recovers all of them at once. The seed and
+result buffers are `n_colors × n` dense arrays, which is fine while the
+coloring works.
+
+It does not work when the Hessian has one **dense row**. A variable that
+multiplies a sum over all the others — a total-cost variable, a shared
+design parameter, a scalar entering every constraint — puts a nonzero in
+*every* column at its own row, so no two columns may share a color and
+the greedy walk hands out one color per variable. `seeds` and
+`compressed` then become O(n²), on a model whose Hessian is otherwise
+perfectly sparse.
+
+A second, independent O(n²) sat behind it in `Tape::hessian_sparsity`,
+which kept a `BTreeSet` of dependent variables alive for *every* tape
+slot. A long additive chain — exactly what a dense row is built from —
+left one full-size set per partial sum and copied the whole running set
+at each union. (`split_top_sums` keeps top-level sums out of a single
+tape, so this only bit when the sum fed an enclosing operator.)
+
+**What is new.** Columns whose nonzero-row count exceeds
+`DENSE_COL_FACTOR` (16) times the average — never fewer than
+`DENSE_COL_MIN` (32) entries, at most `MAX_PEELED_COLS` (256) of them —
+are peeled out of the coloring and given a singleton color each. One
+product with seed `e_d` recovers the whole of column `d`, and because
+`H[d, j] == H[j, d]` that same pass also supplies every entry in row `d`.
+Row `d` therefore stops constraining anyone else's color and drops out of
+the conflict structure, leaving the rest to color on their real sparsity.
+`hessian_sparsity` additionally computes slot lifetimes up front and
+moves a dying operand's set into its consumer instead of copying it.
+
+Measured on `min Σ(xⱼ−1)² + x₀·Σxⱼ` discretized alongside a sparse
+constraint chain — one dense row, `nnz_h ≈ 2n`:
+
+| n = 8,010 | before | after |
+|---|---|---|
+| colors | 8,010 | 6 |
+| peak RSS through setup | 1,038 MB | 59 MB |
+| setup wall clock | 2.69 s | 0.18 s |
+
+| n = 4,010, 10 iterations | before | after |
+|---|---|---|
+| `LagrangianHessianEvaluations` | 2.069 s | 0.019 s |
+| `OverallAlgorithm` | 2.385 s | 0.284 s |
+
+Memory is now linear in `n` where it was quadratic, so what used to
+extrapolate to ~6.4 GB at n = 20,000 (and an OOM at flowsheet sizes) is a
+few tens of MB. Models without a dense row are untouched: a banded
+Hessian peels nothing and colors by its bandwidth exactly as before, and
+every one of the repository's 62 `.nl` fixtures returns a bit-identical
+objective and exit status.
+
+### Changed — the `.nl` reader's parse and setup paths allocate far less (#552)
+
+Reading a `.nl` put two heap allocations on every data line: one for the
+`String` the reader handed over, another for the `Vec<&str>` that
+`parse_var_coef` / `parse_bound_line` collected from it. `parse_expr`
+allocated a `String` per expression token — one per node of every
+expression tree in the file, ~620k of them for a 20k-variable model. All
+of these now borrow from the source buffer. Setup also replaced the
+global Hessian-pair `BTreeSet`, the per-row Jacobian column sets, and the
+per-summand color sets with `Vec` + `sort_unstable` + `dedup`, and
+dropped a `HashMap` that was built with one entry per Hessian nonzero,
+read once, and discarded — the decode tables are now built straight from
+the sorted pair list, which also makes the `eval_h` scatter walk `values`
+forward instead of in `HashMap` order.
+
+This is a constant-factor cleanup, not a complexity change: total setup
+on ordinary sparse models moves 1.00–1.07×. The dense-row case above is
+where the compounding shows up.
+
 ### Fixed — the filter's `theta_max` ceiling is now raised on demand instead of blocking a solve forever (#476, #546)
 
 The filter rejects any trial iterate whose constraint violation exceeds
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,47 @@ Hessian peels nothing and colors by its bandwidth exactly as before, and
 every one of the repository's 62 `.nl` fixtures returns a bit-identical
 objective and exit status.
 
+### Changed — `eval_jac_g` takes the shared-CSE path when the shared bodies are big enough to pay for it (#476)
+
+`HybridTape` — the constraint tape that evaluates a CSE body once for the
+whole block instead of once per referencing summand — has driven `eval_g`
+since #476, while `eval_jac_g` stayed on the flat per-summand tapes whose
+CSE bodies are duplicated. The Jacobian now uses it too, but only above a
+measured threshold, because unlike `eval_g` it is not a free win.
+
+The asymmetry is structural. `eval_g` needs only values, so the prelude
+is swept once for the entire constraint block and the saving is the full
+op-count ratio. The Jacobian needs a gradient *per row*, so only the
+forward sweep can be shared — the reverse sweep still walks each
+summand's `prelude_reach` separately, and it pays a per-op cost the flat
+tape does not: a nested `SummandOp` dispatch and an indirected walk over
+a reach list rather than a straight loop over a contiguous `Vec<TapeOp>`.
+
+Measured on chain models with CSE redundancy 40, varying the shared body
+size (`eval_jac_g`, flat → hybrid, n ≈ 20,000):
+
+| flat/shared op ratio | 1.94 | 2.20 | 2.84 | 3.53 | 4.20 | 5.16 | 6.35 | 8.00 |
+|---|---|---|---|---|---|---|---|---|
+| speedup | 0.77× | 0.63× | 0.88× | 1.21× | 1.21× | 1.18× | 1.50× | 1.32× |
+
+The crossover sits near 3, so the gate is set at 4 to keep a margin: a
+model that does not clearly benefit stays on the flat path bit-for-bit.
+With the gate in place the same sweep shows 1.00–1.03× below it (i.e.
+unchanged) and 1.13–1.30× above it. `robot_a` (#476) measures 4.03×.
+
+`eval_h` is **not** changed and remains the larger prize — it is ~69% of
+AD time on these models against the Jacobian's ~19%. The dormant
+`HybridTape::hessian_summand` is not the tool for it: it re-walks
+`prelude_reach` once per seed variable, so it shares no prelude work at
+all, and it scatters through a `HashMap` lookup per emitted pair where
+the coloring path writes into a dense buffer. Amortizing prelude
+second-order work across summands needs a new directional routine, which
+is its own piece of work.
+
+`POUNCE_DBG_TAPE_STATS=1` now also prints the flat-versus-shared op counts
+for the constraint block, and `POUNCE_DBG_NO_HYBRID=1` forces the flat
+path, so the trade above is reproducible on any model.
+
 ### Changed — the `.nl` reader's parse and setup paths allocate far less (#552)
 
 Reading a `.nl` put two heap allocations on every data line: one for the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,36 @@ Hessian peels nothing and colors by its bandwidth exactly as before, and
 every one of the repository's 62 `.nl` fixtures returns a bit-identical
 objective and exit status.
 
+### Fixed — Pyomo's modern solver interface could not drive POUNCE at all (#552)
+
+`SolverFactory('ipopt_v2', executable=<pounce>)` — the v2 interface that
+is becoming Pyomo's default `ipopt` — failed on every model. Two
+independent gaps, both places where POUNCE diverged from what the ASL
+does and every reader therefore expects:
+
+1. **Quoted option values.** The v2 interface builds each option as
+   `option_file_name="/tmp/…/x.opt"` and passes it as a single `argv`
+   entry. With no shell in between, the quotes arrive as literal
+   characters. Ipopt's ASL option parser strips them; POUNCE did not, so
+   it looked for a file whose name began with `"`, failed to load it, and
+   aborted the run. `key=value` now drops one matching pair of surrounding
+   `"` or `'` quotes. A quote on one side only is left as content.
+
+2. **The `.sol` `Options` block.** POUNCE wrote a count of `0`. AMPL and
+   Pyomo's *legacy* reader accept that, but no ASL solver emits it: the
+   v2 reader reads the first two option words unconditionally (to detect
+   the documented quirk where a second word of `3` means two extra words
+   follow the `z` block) and so asserts `n_opts >= 2`. A POUNCE `.sol`
+   could not be parsed through the modern interface at all. POUNCE now
+   writes the same `3 / 1 / 1 / 0` the ASL's own `writesol.c` and Ipopt
+   write.
+
+With both fixed, the same model solved through `SolverFactory('pounce')`
+and through `SolverFactory('ipopt_v2', executable=<pounce>)` returns
+identical objectives and identical primal values. This also makes it
+possible to benchmark POUNCE against Ipopt through *one* Pyomo interface,
+which #552's overhead comparison could not do.
+
 ### Changed — `eval_jac_g` takes the shared-CSE path when the shared bodies are big enough to pay for it (#476)
 
 `HybridTape` — the constraint tape that evaluates a CSE body once for the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,8 @@ tape, so this only bit when the sum fed an enclosing operator.)
 **What is new.** Columns whose nonzero-row count exceeds
 `DENSE_COL_FACTOR` (16) times the average — never fewer than
 `DENSE_COL_MIN` (32) entries, at most `MAX_PEELED_COLS` (256) of them —
-are peeled out of the coloring and given a singleton color each. One
+become *candidates* for peeling, and those that pay for themselves are
+peeled out of the coloring and given a singleton color each. One
 product with seed `e_d` recovers the whole of column `d`, and because
 `H[d, j] == H[j, d]` that same pass also supplies every entry in row `d`.
 Row `d` therefore stops constraining anyone else's color and drops out of
@@ -64,6 +65,29 @@ Hessian peels nothing and colors by its bandwidth exactly as before, and
 every one of the repository's 62 `.nl` fixtures returns a bit-identical
 objective and exit status.
 
+**Which candidates actually get peeled.** Each peeled column costs
+exactly one color, so peeling only pays when it removes more conflict
+than it adds — and simply truncating an over-long candidate list to
+`MAX_PEELED_COLS` does not bound the damage. The columns that miss the
+cut stay in the conflict structure, so the base color count is unchanged
+and the singleton colors are pure addition. On disjoint 50x50 blocks
+scattered through a 200,000-variable Hessian that colors to `50 + 256`
+where a plain walk needs 50 — a 6x regression in the very quantity
+peeling exists to reduce, and 8.5x on 34x34 blocks.
+
+The candidate set is therefore chosen by estimating the result: for
+peeling nothing and for the top `k` candidates by degree over a doubling
+ladder, `|peeled| + max surviving row degree` is a lower bound on the
+resulting color count (a row with `d` surviving entries makes those `d`
+columns pairwise conflict), computable in one O(nnz) pass. The best
+scoring option wins, ties going to fewer peels. On the block patterns
+above every `k > 0` now scores strictly worse than peeling nothing, so
+nothing is peeled and the coloring matches the plain walk exactly; the
+one-dense-row case still collapses n = 20,000 to 2 colors. Note the plain
+coloring cannot serve as a comparison baseline here — on that same
+one-dense-row case the plain walk is itself O(n²), which is the blowup
+peeling exists to avoid — so the choice has to be made from a bound.
+
 ### Fixed — Pyomo's modern solver interface could not drive POUNCE at all (#552)
 
 `SolverFactory('ipopt_v2', executable=<pounce>)` — the v2 interface that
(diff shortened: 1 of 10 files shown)


jkitchin:
Reviewed all three commits. The dense-row diagnosis is right, the two O(n²)s are genuinely independent, and the decode rule is correct in all four peel combinations — I checked `(peeled i, plain j)`, `(plain i, peeled j)`, both, and neither against `H[i,j] == H[j,i]`, and each lands on a column that has a color. The `for_each_input` liveness analysis is sound too: over-reporting an input can only extend a set's lifetime, and that is the safe direction, which the doc comment gets right.

One blocking issue and one smaller one.

## 1. `MAX_PEELED_COLS` truncation is additively harmful, not a damage bound

`MAX_PEELED_COLS` is described as bounding the damage if the heuristic misfires, since each peeled column costs exactly one color. That holds when peeling *works*. When more than 256 columns exceed the threshold, truncation keeps 256 and **leaves the rest in the coloring** — so those still drive the base color count, and the 256 singleton colors are pure addition on top.

Measured by calling `greedy_hessian_coloring` directly against a plain distance-1 greedy with peeling disabled (disjoint dense blocks embedded in a large sparse remainder, so a plain coloring needs only ~`block_sz` colors regardless of block count):

| n | blocks | block size | plain | peeled path | peeled | result |
|---|---|---|---|---|---|---|
| 200,000 | 300 | 50 | 50 | 50 | 0 | fine (threshold 64 > 50) |
| 200,000 | 300 | 40 | 40 | 40 | 0 | fine |
| **200,000** | **100** | **50** | **50** | **306** | 256 | **6.1× more colors** |
| **200,000** | **300** | **34** | **34** | **290** | 256 | **8.5× more colors** |
| 50,000 | 300 | 50 | 50 | 50 | 0 | fine |

`n_colors` is the multiplier on both the `n_colors × n` `seeds` / `compressed` arrays and the per-`eval_h` work — the exact quantities this commit exists to reduce. So on these patterns the fix makes both 6–8× worse than the code it replaces.

Note rows 1 and 4 differ only in block size, 50 vs 34, and flip between "no peeling at all" and "8.5× worse". The threshold `max(32, 16·avg)` is sensitive right where these models sit: at block size 50 the average degree pulls the threshold to 64 and nothing peels; at 34 it sits at the `DENSE_COL_MIN` floor of 32 and everything qualifies.

This is not exotic — it is any model with more than 256 moderately-dense columns inside a large sparse problem. Worth noting that #552's flowsheet models (819 and 913 variables per time point) are safe only because they are *uniformly* dense, so the `16·avg` factor dominates; a model mixing a few hundred medium blocks with a large sparse remainder is the gap.

Neither new test covers it: `a_sparse_hessian_is_colored_by_its_bandwidth_not_peeled` is banded, so degrees are uniform and nothing ever qualifies, and `a_dense_hessian_row_does_not_explode_the_coloring` has exactly one dense column. The untested region is "more columns qualify than the cap allows".

The predicate is what wants revisiting rather than the cap: the motivating case has `deg ≈ n` (a row touching *every* column), while the regression cases have `deg = 50` against `n = 200,000`. `> 16× average` admits far more than "dense". Simply refusing to peel when `dense.len() > MAX_PEELED_COLS` fixes these rows, but it would strand a model with, say, 300 genuinely dense rows back at the O(n²) behaviour — so the density test itself probably needs to be relative to `n`, not to the average.

I'd hold the merge on this one, since it is a regression in the same metric the commit optimizes.

## 2. The `.sol` `Options` block is hard-coded where the ASL echoes the model's own

The commit is right that `Options 0` breaks Pyomo's v2 reader — I read `pyomo/contrib/solver/solvers/asl_sol_reader.py` and both `assert n_opts >= 2` (line 300) and `read_vbtol = ampl_options[1] == 3` (line 302) are exactly as described, and `3 / 1 / 1 / 0` parses correctly through it. The fix works.

But the body says it emits "the same three words Ipopt and the ASL's own `writesol.c` emit", and that isn't what Ipopt does. Ipopt echoes back the model's own `ampl_options` from the `.nl` header line, so the words vary per model:

| model | `.nl` header | Ipopt's `.sol` `Options` |
|---|---|---|
| `bt4` | `g3 0 1 0` | `3 / 0 / 1 / 0` |
| `camshape_6400` | `g3 10 1 0` | `3 / 10 / 1 / 0` |
| `robot_1600` | `g9 …` | **9** option words |
| `svanberg` | `g9 …` | **9** option words |

(Ipopt 3.14.19, `-AMPL`, files written by the runs themselves.) A hard-coded `3 / 1 / 1 / 0` writes `1` where `bt4` should echo `0`, and writes three words where a `g9` model has nine.

Nothing downstream appears to *use* the values — Pyomo v2 just records them as `data.ampl_options` — so I don't think this breaks anything today, and it is much less serious than #1. But it is the same class of divergence-from-the-ASL the commit sets out to remove, and the real values are sitting on header line 0, which `parse_header` already reads and currently discards after checking the leading `g`. Echoing them is a small change and removes the question permanently.

At minimum the PR body and the `format_sol` doc comment shouldn't claim these are the words Ipopt emits.

## Smaller notes

- `parse_kv` / `unquote`: correct, including leaving a one-sided quote as content. `key=""` returning `None` via the existing empty-value check is the right call.
- The gate at `HYBRID_JAC_MIN_OP_RATIO = 4` against a measured crossover near 3 is well-argued, and the A/B instrument (`POUNCE_DBG_NO_HYBRID`) makes it checkable rather than asserted — good.
- The coverage gap you named yourself (no fixture has a CSE shared across constraints, so the 62-fixture sweep proves no-regression rather than proving the new Jacobian path correct) is the right thing to have flagged; I don't think it blocks, given the two exact-equality unit tests on both sides of the gate.

All 12 CI checks are green.

PR #556, opened 2026-08-09
fix(nl): peel dense columns in the Hessian coloring so a dense row can't force ~n colors

### LINKED ISSUE #554: Performance: ~2.3x per-iteration cost vs IPOPT+MA57 on collocation-discretized models

jkitchin:
Fixes #554

## Problem

The per-iteration cost gap #554 tracks concentrates in collocation-discretized models, with `steering_12800` / `rocket_12800` as the cleanest signals (~8.5× IPOPT+MA57 wall at identical iteration counts). Profiling those instances (regenerated at the same `nh` via Pyomo) shows the dominant term is **not** the linear solver: on `rocket_3200`, `LagrangianHessianEvaluations` is 5.07 s of a 6.85 s solve (76%), against 1.20 s of factorization. The cost is also O(n²) in both time and memory — `rocket_12800` (n = 51 205) allocates ~42 GB and is OOM-killed on a 15 GB machine, even though #554 records it solving in 15.8 s on a larger one.

| rocket_3200 (n = 12 805) | status | objective | iters | Hessian eval | total solve | peak RSS |
|---|---|---|---|---|---|---|
| before | Optimal | -1.0128257882438041 | 30 | 5.07 s | 6.85 s | 2 582 MB |
| after | Optimal | -1.0128257882438041 | 30 | 0.28 s | 1.83 s | ~180 MB |

| 12 800-point instances | before | after |
|---|---|---|
| `rocket_12800` | OOM-killed (>15 GB) | Optimal, 30 iters, 7.6 s, 310 MB |
| `steering_12800` | OOM-killed (>15 GB) | Optimal, 15 iters, 7.3 s, 318 MB |

The 30-iteration count on `rocket_12800` matches the 30/30 recorded in #554, so this is the same convergence path the benchmark measured.

## Cause

The `.nl` evaluator recovers the Lagrangian Hessian from compressed Hessian-vector products, one per color of a greedy distance-1 coloring of the column-intersection graph (Coleman–Moré direct recovery). Columns `c1`, `c2` are adjacent iff they share any nonzero row. A column whose Hessian row is **dense** — the free-time `step`/`tf` variable of a COPS-style optimal-control model, which multiplies every collocation equation — makes *every pair* of columns adjacent through that one shared row. The intersection graph degenerates to a clique and the coloring to ~one color per variable: `POUNCE_DBG_TAPE_STATS` reports **n_colors = 12 800** on `rocket_3200` and 9 599 on `steering_3200`, versus 4 with `step` fixed. Since the seed and compressed buffers are each `n_colors × n` doubles, memory is O(n²) (2 × 12 805² × 8 B = 2.6 GB — exactly the observed 2 582 MB plateau) and `eval_h` zeroes all of it per call (~130 ms of the ~168 ms per-eval cost).

## Fix

`greedy_hessian_coloring` now **peels dense columns first**: any column whose full-symmetric degree exceeds `max(64, 8 × average degree)` gets an exclusive color. Its product `H·e_j` is the exact symmetric column j, so every entry in a peeled row/column decodes from the exclusive product. The remaining columns are colored by the same greedy pass as before, except conflicts routed *through a peeled row* are ignored — those entries never decode from a shared color, so direct recovery stays exact: for a pair `(i, j)` with both columns sparse, any column sharing the (non-peeled) row `i` with `j` is still seen by the conflict scan and forced into a different color. With `p` peeled columns the count drops from ~n to `p + χ(rest)` — 5 on `rocket_3200`.

A peeled color costs one extra H·s product per eval, which a dense column costs under any exact scheme — that is the floor, not an overhead.

The threshold is deliberately conservative (`max(64, 8×avg)`): a mesh/stencil model has uniform degree ≈ stencil size and never trips it, so the peeling can only fire on columns that are genuinely far out of family. Rejected alternative: routing dense models to the matrix-free HVP path instead — that changes the algorithm's linear-system interface per model class, where peeling is a strictly local change to color assignment and decode routing with an empty-set no-op on every existing model.

## Blast radius

`.nl`-driven evaluation only (`pounce-nl`); the tape, HVP, gradient, and Jacobian paths are untouched. When no column crosses the threshold, `is_dense` is all-false, the greedy starts at color 0, and both the coloring and the decode tables are **bit-for-bit the historical ones** — models without a dense Hessian row cannot observe this change. On the affected models the recovered Hessian is exact, pinned by the identical iterate path above (same iterations, objective equal to all printed digits, NLP error unchanged).

Not addressed here: the 3.5–4.8× factorization-kernel gap on *fixed*-horizon chains (hicks/quad_tank/cart_pole) measured in #552 lives in the `feral` crate (pinned at 0.14.0, the latest release) and remains with #552. After this fix, factorization is the dominant remaining term on the 12 800-point instances (~5 s of ~7.5 s), so #552 is the next lever. A re-run of the 47-instance Mittelmann sweep would be the definitive check on #554's matched-iteration statistic; this environment cannot fetch the AMPL sources (plato.asu.edu egress-blocked).

## Tests

- `dense_column_peeling_caps_colors` — star pattern (hub + 199 spokes): asserts the hub is peeled, spokes are not, `n_colors ≤ 3`, and the hub's color is exclusive; also asserts a uniform chain peels nothing. Fails on the parent commit for the stated reason: un-peeled, the star colors to 200 (the test does not compile against the parent's 2-tuple signature; ported to it, the `n_colors ≤ 3` assert is what fails).
- `dense_hub_hessian_evaluates_exactly` — end-to-end `eval_h` on `min Σ t·x_i²` (101 vars, hub degree 100): asserts the hub is actually peeled and every recovered entry matches the analytic Hessian to 1e-12. This pins the *new decode routing's* exactness; it passes on the parent too (the old path was exact, just O(n²)) — the performance component is pinned by the coloring test's cap, not by a timing assertion, which would be flaky.

`cargo fmt --all -- --check`, `cargo clippy` (CI lint set), and `cargo test --workspace --exclude pounce-hsl` are clean.

---

- [x] Tests fail on the parent commit for the stated reason (`dense_column_peeling_caps_colors`; the exactness test is a correctness pin, see Tests)
- [ ] Book page under `docs/src/` — not updated: no user-facing option or documented behavior changed; the coloring is internal to the `.nl` evaluator

https://claude.ai/code/session_01HzCfuP37kNMCseKRbAJkqW

---
_Generated by [Claude Code](https://claude.ai/code/session_01HzCfuP37kNMCseKRbAJkqW)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,42 @@ changes.
 
 ## [Unreleased]
 
+### Fixed — Hessian coloring no longer collapses to ~n colors on models with a dense Hessian row (#554)
+
+The `.nl` evaluator recovers the Lagrangian Hessian from compressed
+Hessian-vector products, one per color of a greedy distance-1 coloring of
+the column-intersection graph (Coleman–Moré direct recovery). A column
+whose Hessian row is **dense** — the free-time `step`/`tf` variable of a
+COPS-style optimal-control model, which multiplies every collocation
+equation, or any shared design parameter — makes *every pair* of columns
+adjacent through that one shared row. The intersection graph degenerates
+toward a clique, the coloring toward one color per variable, and since
+the seed and compressed buffers are each `n_colors × n` doubles, memory
+and per-evaluation zeroing cost turn O(n²): `rocket_12800` (n = 51 205)
+needed ~42 GB and was OOM-killed on a 15 GB machine, and on instances
+that fit, `eval_h` — not the linear solver — was the dominant
+per-iteration cost that #554 measured (5.07 s of a 6.85 s
+`rocket_3200` solve).
+
+The coloring now **peels dense columns first**: any column whose degree
+exceeds `max(64, 8 × average degree)` gets an exclusive color, whose
+product `H·e_j` is the exact symmetric column j. The remaining columns
+are colored as before, except that conflicts routed *through a peeled
+row* are ignored — those entries decode from the peeled column's
+exclusive product instead, so recovery stays exact. With `p` peeled
+columns the color count drops from ~n to `p + χ(rest)`. The threshold is
+conservative by design: a mesh/stencil model (uniform degree) never
+trips it and keeps its historical coloring bit for bit.
+
+On `rocket_3200`, colors drop 12 800 → 5, Lagrangian-Hessian time
+5.07 s → 0.28 s, and the whole solve 6.85 s → 1.83 s, with an identical
+iterate path (same 30 iterations, objective equal to all printed
+digits). `rocket_12800` and `steering_12800` — the two cleanest signals
+in #554, previously requiring tens of GB — now solve in ~7.5 s at
+~310 MB peak. The remaining per-iteration gap against MA57 on
+fixed-horizon collocation chains is the factorization-kernel question
+tracked in #552.
+
 ### Fixed — the filter's `theta_max` ceiling is now raised on demand instead of blocking a solve forever (#476, #546)
 
 The filter rejects any trial iterate whose constraint violation exceeds
--- a/crates/pounce-nl/src/nl_reader.rs
+++ b/crates/pounce-nl/src/nl_reader.rs
@@ -2395,7 +2395,7 @@ fn split_top_sums(expr: &Expr) -> Vec<Expr> {
 }
 
 /// Greedy column coloring of a symmetric sparsity pattern stored
-/// as lower-triangle pairs.
+/// as lower-triangle pairs, with dense-column peeling.
 ///
 /// Builds the column-intersection graph: columns `c1` and `c2` are
 /// adjacent iff there exists a row `r` with `H[r, c1] != 0` and
@@ -2405,13 +2405,43 @@ fn split_top_sums(expr: &Expr) -> Vec<Expr> {
 /// pairwise disjoint row supports, so a single H·s product
 /// recovers them all unambiguously.
 ///
-/// Returns `(var_color, n_colors)` where `var_color[k]` is the
-/// color assigned to variable `k`, or `u32::MAX` for variables
-/// not in any Hessian pair (they contribute nothing and don't
-/// need a color).
-fn greedy_hessian_coloring(n: usize, lower_pairs: &[(usize, usize)]) -> (Vec<u32>, usize) {
+/// **Dense-column peeling (pounce#554).** A column whose degree is
+/// far above the average — the free-time `step`/`tf` variable of a
+/// COPS-style optimal-control model couples with *every* state, so
+/// its Hessian row is dense — makes every pair of columns adjacent
+/// through that shared row, collapsing the intersection graph
+/// toward a clique and the coloring toward `n` colors. Since the
+/// seed and compressed buffers are `n_colors × n`, that is O(n²)
+/// memory and O(n²) zeroing per `eval_h` (rocket_12800: ~42 GB,
+/// OOM). The classical remedy: peel such columns out first, giving
+/// each an **exclusive** color whose product `H·e_j` recovers the
+/// whole symmetric column j directly, and color the remaining graph
+/// while ignoring conflicts routed through peeled rows — those
+/// entries are decoded from the peeled column's side instead (see
+/// the decode-table construction in `NlTnlp::try_new`). With p
+/// peeled columns the count drops to `p + χ(rest)`; for the
+/// collocation models that is ~5 instead of ~n.
+///
+/// Recovery stays exact: a lower pair `(i, j)` with `j` (or `i`)
+/// peeled reads `H[i, j]` out of the peeled column's exclusive
+/// product, and a pair with both sparse decodes from `color(j)` at
+/// row `i`, which cannot collide — any column sharing the
+/// (non-peeled) row `i` with `j` was seen by the conflict scan and
+/// forced into a different color.
+///
+/// Returns `(var_color, n_colors, is_dense)` where `var_color[k]`
+/// is the color assigned to variable `k`, or `u32::MAX` for
+/// variables not in any Hessian pair (they contribute nothing and
+/// don't need a color), and `is_dense[k]` marks peeled columns
+/// (each alone in its color). When no column crosses the peeling
+/// threshold the coloring is identical to the historical un-peeled
+/// one.
+fn greedy_hessian_coloring(
+    n: usize,
+    lower_pairs: &[(usize, usize)],
+) -> (Vec<u32>, usize, Vec<bool>) {
     if n == 0 {
-        return (Vec::new(), 0);
+        return (Vec::new(), 0, Vec::new());
     }
 
     // For each variable k, list of rows in which column k has a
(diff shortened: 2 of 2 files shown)


jkitchin:
## Review: this is the same fix as #553, and its selection rule reintroduces the O(n^2) blowup it removes

Two things, and the second is the blocking one.

### 1. Duplicate of #553

PR #553 ("dense-Hessian-row O(n^2) fix, ...", branch `claude/issue-552-uet7dp`) already
rewrites the same function, `greedy_hessian_coloring` in `crates/pounce-nl/src/nl_reader.rs`,
with the same mechanism: peel high-degree columns into exclusive colors, skip conflicts routed
through a peeled row, and re-route the decode table. The decode change is line-for-line
equivalent -- #553 has

    let (c, row) = if peeled[i] && !peeled[j] { (var_color[i], j) } else { (var_color[j], i) };

against this PR's

    let (col, row) = if is_dense[i] && !is_dense[j] { (i, j) } else { (j, i) };

Both PRs also change the signature to a 3-tuple returning the peel mask. They will conflict
hard. #553 is further along (CI running on the review fixes now), so the question is whether
#556 carries anything #553 doesn't -- and on the selection rule it carries the opposite.

I confirmed #553's version fixes the motivating case here: on the star pattern this PR's own
test uses, `greedy_hessian_coloring` under #553 returns **2 colors with 1 column peeled**, at
both n = 200 and n = 20 000. Nothing in #554's Hessian blowup survives #553.

### 2. Blocking: the `max(64, 8 x avg)` rule has no cap and no pay-for-itself check

Every column above the threshold gets an exclusive color, unconditionally. That is safe when
the dense columns are few (the collocation `step`/`tf` case this PR profiles), but the rule
never asks whether peeling *helps*, and nothing bounds how many columns it fires on.

The failure mode is a **mixed degree profile**: a low-degree bulk that drags the average down,
plus many medium-degree columns whose row supports are *pairwise disjoint*. Disjoint supports
mean a plain coloring gives them all one color per block -- but peeling gives each one its own.
Built as a degree-3 chain plus `nb` disjoint dense blocks of size `bs`, running this branch's
`greedy_hessian_coloring` against a plain greedy coloring of the same pattern:

| bulk | blocks | plain colors | this PR | peeled | seed+compressed memory |
|---|---|---|---|---|---|
| 190 000 chain | 100 x 100 | 100 | **10 003** | 10 000 | 0.3 GB -> **32.0 GB** |
| 190 000 chain | 200 x 70 | 70 | **14 003** | 14 000 | 0.2 GB -> **45.7 GB** |
| 100 000 chain | 50 x 200 | 200 | **10 003** | 10 000 | 0.4 GB -> **17.6 GB** |
| 50 000 chain | 20 x 100 | 100 | **2 003** | 2 000 | 0.1 GB -> 1.7 GB |

100x to 200x more colors, and 32-46 GB of `n_colors x n` buffers -- the exact OOM this PR
exists to eliminate, now triggered on models that were previously fine. The arithmetic: with a
degree-3 chain of 190 000 and 100 blocks of size 100, average degree is ~7.9, so the threshold
is `max(64, 63) = 64`, and all 10 000 block columns (degree 100) clear it.

The PR body's argument -- "a mesh/stencil model has uniform degree ~ stencil size and never
trips it, so the peeling can only fire on columns that are genuinely far out of family" -- is
right about *uniform* models and is exactly why the mixed case slips through. Far out of family
is not the same as worth peeling: 10 000 columns of degree 100 are far above a chain-dominated
average, and peeling them is a 100x loss.

I ran the same four structures against #553's version: **peeled = 0 on all four, matching plain
coloring exactly (1.0x)**, while still collapsing the star to 2 colors. #553 gets there by
scoring candidates instead of thresholding them -- a doubling ladder over the degree-sorted
candidate list, keeping the prefix that minimizes `|peeled| + max surviving row degree`, which
is a lower bound on the resulting color count and computable in one O(nnz) pass. (A plain
coloring cannot be used as the comparison baseline, since on the one-dense-row case the greedy
walk is itself O(n^2) -- precisely what peeling exists to avoid. The decision has to come from a
cheap bound.) #553 also caps peeling at 256 columns as a backstop.

### Recommendation

Close #556 in favor of #553, and let #553 close #554 as well -- it fixes the same root cause and
its selection rule is the safe one. If you'd rather land the narrow version, then this branch
needs #553's bound-based selection ported in wholesale, at which point the two are the same
patch and the duplication is the only thing left to resolve.

Smaller notes, only relevant if this branch does land independently:

- `dense_column_peeling_caps_colors` covers the star and a uniform chain -- both cases the
  threshold gets right. It has no mixed-profile case, which is why the regression above is
  invisible to the test suite.
- The body says `rocket_12800` "was OOM-killed on a 15 GB machine, even though #554 records it
  solving in 15.8 s on a larger one". Worth keeping that qualifier in the CHANGELOG entry too;
  as written there it reads as an unconditional OOM.

jkitchin:
Agreed on both counts — closing in favor of #553.

The mixed-profile criticism is correct and I verified the arithmetic independently: with a 190k degree-3 chain and 100 disjoint 100-column blocks, the average degree lands at ~7.9, the `max(64, 8×avg)` threshold at 64, and every one of the 10,000 block columns (degree ~100) clears it — each into an exclusive color, which is exactly the `n_colors × n` blowup this PR set out to remove. My rule has no pay-for-itself check and no cap; #553's scored ladder + `MAX_PEELED_COLS` backstop is the safe selection, and my test matrix (star + uniform chain only) was structurally blind to the failure mode. Nothing here is worth porting over it.

Two things worth carrying to #553's record:

1. **Validation on the actual #554 instances.** I regenerated `rocket_12800` / `steering_12800` at the same `nh` via Pyomo (COPS models, trapezoidal — generator script is small and AMPL-free if wanted for `benchmarks/`). Pre-fix on a 15 GB machine both are OOM-killed (`n_colors` = 12,800 on rocket_3200; buffers extrapolate to ~42 GB at nh=12800). With peeling: rocket_12800 solves Optimal in 30 iterations — matching the 30/30 recorded in #554 — at ~310 MB peak, steering_12800 in 15 iterations at ~318 MB, and on rocket_3200 the Lagrangian-Hessian share of the solve drops 5.07 s → 0.28 s with the objective identical to all printed digits. So #553's commit 1 does resolve #554's two headline instances, where Hessian evaluation — not FERAL — was the dominant per-iteration term (76% of `OverallAlgorithm` via `print_timing_statistics`); the residual factor-kernel gap on fixed-horizon chains stays with #552.

2. **The auto-close.** Per the repo convention (CLAUDE.md, learned on #342), #554 only closes if `Fixes #554` appears in #553's *body* — its body currently has the deliberate no-keyword note for #552 but no #554 reference, so as it stands #554 will dangle on merge.

---

jkitchin:
Closing in favor of #553, which fixes the same root cause.

#553 rewrites the same function with the same mechanism — peel high-degree columns into
exclusive colors, ignore conflicts routed through a peeled row, re-route the decode table to
the peeled endpoint — and carries a selection rule that this branch's fixed
`max(64, 8 x avg)` threshold does not. On the four mixed-degree structures in the review above,
this branch turns 70-200 colors into 10 003-14 003 (up to 46 GB of `n_colors x n` buffers);
#553 peels nothing on all four, matching a plain coloring exactly, while still collapsing the
star pattern from this PR's own test to 2 colors at both n = 200 and n = 20 000.

#553 now carries `Fixes #554`, so #554 closes when it merges. The residual factorization-kernel
gap on fixed-horizon collocation chains — correctly scoped out of this PR — stays with #552.

The profiling in this PR body is the useful part and outlives the branch: the `rocket_3200`
breakdown (`LagrangianHessianEvaluations` 5.07 s of a 6.85 s solve, against 1.20 s of
factorization), the `n_colors = 12 800` vs 4-with-`step`-fixed measurement, and the identical
30-iteration path before and after are what established that `eval_h` and not the linear solver
was the dominant per-iteration term in #554. Nothing here is being discarded as wrong — only
superseded by the safer selection rule.

PR #559, opened 2026-08-09
perf(nl): route eval_h through the shared-CSE tape, amortizing prelude second-order work per color (#557)

### LINKED ISSUE #557: Route eval_h through the shared-CSE tape: amortize prelude second-order work across summands
Follow-on to #476 and #553. `HybridTape` — the constraint tape that evaluates a
shared CSE body once for the whole block instead of once per referencing
summand — now drives `eval_g` (#476) and `eval_jac_g` above an op-ratio gate
(#553). **`eval_h` is still on the flat per-summand tapes, and it is the
largest of the three by a wide margin.**

## Why this one is worth doing

Measured with `cargo run --release -p pounce-nl --example adbench`, on a chain
model with CSE redundancy 10 (n = 20,010, m = 20,000, flat/shared op ratio
2.20×), on merged `main`:

| | flat path | current default |
|---|---|---|
| `eval_g` | 0.936 ms | 0.568 ms (hybrid) |
| `eval_grad_f` | 0.642 ms | 0.625 ms |
| `eval_jac_g` | 1.876 ms | 1.785 ms (gate off at 2.20×) |
| **`eval_h`** | **6.839 ms** | **6.658 ms** |
| total | 10.293 ms | 9.636 ms |

`eval_h` is **66–69% of AD time** depending on which path the others take.
Everything shipped so far has been optimizing the other third.

## The dormant `hessian_summand` is the wrong tool

`HybridTape::hessian_summand` exists and has never had a caller. Do not simply
wire it up:

1. It iterates `s.all_vars` and, **for each seed variable**, walks
   `prelude_reach` forward and backward again. It shares no prelude work
   whatsoever — the op count per seed direction is the same as the flat tape's,
   which already inlines the CSE body.
2. It scatters through `hess_map: &HashMap<(usize, usize), usize>` — a hash
   lookup per emitted pair — where the coloring path writes into a dense
   `compressed[c]` buffer.

#553 is the cautionary data point: the *Jacobian* has the forward sweep shared
and still only broke even at op ratio ≈ 3, because the hybrid representation
carries real per-op overhead (nested `SummandOp` dispatch, indirected walk over
a reach list instead of a straight loop over a contiguous `Vec` of `TapeOp`).
A routine that shares nothing **and** adds hashing would very likely be slower
than what we have.

## The design that should work

The coloring already gives the thing that makes sharing possible: for a given
color `c`, **every summand uses the same seed vector** `s_c`. So per color:

1. Forward-tangent the prelude **once** with seed `s_c`.
2. For each summand with nonzero weight `λ_k`: forward its local ops (pulling
   `prelude_dot` at `Shared` boundaries), then reverse-over-tangent the local
   ops, accumulating **weighted** adjoint/adjoint-dot into the prelude
   accumulators at `Shared` boundaries, and writing variable contributions into
   `compressed[c]`.
3. Reverse-over-tangent the prelude **once**, folding the accumulated adjoints
   into `compressed[c]`.

This is valid because reverse-over-tangent is linear in the adjoint seeds:
accumulating `λ_k ·` adjoints across summands and doing a single prelude sweep
gives the same result as per-summand sweeps, which is just reverse mode on a
shared DAG. Folding `λ_k` into the adjoints means the prelude sweep runs with
weight 1.

Concretely this needs a hybrid analogue of `Tape::hessian_directional` that
accumulates into a dense `out[var]` rather than through `hess_map` — i.e.
porting that routine's inner logic (~470 lines) to the prelude/summand split.
It is new AD code, not a wiring change.

## Requirements

- **Gate it and measure it, do not assume.** #553's crossover table is the
  precedent: the win must be demonstrated against the flat path per model
  class, and low-ratio models must stay bit-identical on the flat path.
- **Bit-identical output.** The existing pattern is exact equality against the
  flat tapes (`shared_cse_jacobian_matches_flat_tape_bit_for_bit`,
  `a_deep_shared_body_turns_the_jacobian_gate_on_and_still_agrees`). Hold the
  same bar here — a wrong Hessian surfaces as a bad search direction, not a
  panic.
- **Coverage.** *No repository fixture has a CSE shared across constraints* —
  all 62 checked. Both existing hybrid paths are unit-test-only. This work
  should land a fixture or a generator so the shared-CSE paths get end-to-end
  coverage rather than adding a third untested one.

## Instruments already in place

- `POUNCE_DBG_NO_HYBRID=1` — force the flat path, for a same-binary A/B.
- `POUNCE_DBG_TAPE_STATS=1` — prints flat-vs-shared constraint op counts
  (`flat_ops`, `prelude_ops`, `local_ops`, ratio), which is the gate's input.
- `cargo run --release -p pounce-nl --example adbench -- model.nl [reps]` —
  times the four evaluators in isolation, no solver dynamics.

## Note on expected impact

This is an AD-throughput change. It will do nothing for models without shared
CSE bodies — `.nl` files Pyomo writes for collocation models have **zero** `V`
segments, so the hybrid tape is never built for them at all (see #552). The
population that benefits is AMPL models with defined variables referenced from
many rows: `robot_a` measures 4.03×.

jkitchin:
Closes #557

> **Updated after review.** All four review items are addressed, and the crossover table below has been re-measured as medians — the original was single runs per point, which this machine's variance does not support. The gate moved 2.5 → 3.0 as a result. See the review reply for detail.

## Problem

`eval_h` is 66–69% of AD time on shared-CSE models, and it was the last evaluator still on the flat per-summand tapes, which inline (and re-differentiate) a shared CSE body once per referencing summand. Everything shipped so far (#476 `eval_g`, #553 `eval_jac_g`) optimized the other third.

## Cause

Not a bug — a missing capability. The dormant `HybridTape::hessian_summand` could not fill it: it re-walks `prelude_reach` forward and backward once **per seed variable**, sharing no prelude work at all, and scatters through a `HashMap` lookup per emitted pair where the coloring path writes into a dense buffer. #553's crossover data (forward-sweep-only sharing broke even at ratio ≈ 3) predicted it would lose outright.

## Fix

The color-level design from the issue. For each Hessian color `c`, every summand uses the same seed vector `s_c`, so per color:

1. forward-tangent the prelude **once** for the whole constraint block;
2. per summand with `λ_k ≠ 0`: forward-tangent the local ops (pulling `prelude_dot` at `Shared` boundaries), reverse-over-tangent them writing `weight · wd` straight into the coloring's dense `compressed[c]` (no hashing), and fold `λ_k` into the adjoints deposited at `Shared` boundaries;
3. reverse-over-tangent the prelude **once** with unit weight — valid because reverse-over-tangent is linear in its adjoint seeds.

New kernels `fwd_dir_step` / `ror_dir_step` mirror `Tape::hessian_directional`'s arms exactly, so the per-summand local arithmetic is bit-identical to the flat tape; the folded prelude sweep is mathematically identical and agrees to rounding (the λ-fold plus cross-summand accumulation reassociates floating-point products — exact bitwise equality of the *shared* portion is unattainable by construction, which the tests pin as tightly as each case admits). `hessian_summand` and its now-unused `fwd_tan_step` / `ror_step` helpers are removed.

Both prelude sweeps run **once per color**, so they iterate the union of that color's summands' `prelude_reach` rather than the whole prelude. Walking all of it would cost `n_colors × |prelude|` where the op-ratio gate assumes `|prelude|` — a discrepancy unbounded in `n_colors` that the gate cannot see. A union of operand-closed ascending sets is itself operand-closed and ascending, which is exactly what the sweeps require.

### Measurements

Median of 5 interleaved flat/hybrid pairs per point, chain models at CSE redundancy 40, m = 20,000. Interleaved because absolute times drift over minutes on this machine while the paired ratio stays tight; medians because single runs do not reproduce to the precision a threshold needs — the 2.54 column spans 0.36× to 1.14×.

| flat/shared op ratio | 1.94 | 2.54 | 3.12 | 3.69 | 4.24 | 5.29 | 6.76 | 8.53 |
| median `eval_h` speedup | 1.00× | 1.04× | 1.18× | 1.23× | 1.31× | 1.44× | 1.36× | 1.49× |
| min–max | 0.91–1.11 | 0.36–1.14 | 1.10–1.33 | 1.16–1.30 | 1.23–1.32 | 1.19–1.54 | 1.27–1.50 | 1.36–1.65 |

`HYBRID_HESS_MIN_OP_RATIO` is **3.0** — the lowest ratio where every sample wins by ≥ 10%. At 1.94 and 2.54 the median is within noise of break-even and individual runs lose. Below the gate `eval_h` stays on the flat path bit-identically, so a gate set high is merely conservative while one set low risks a real regression. Whole AD stack at ratio 4.24: **1.25×** (median of 5 pairs, range 1.14–1.44).

Per-color reach measured in isolation is **neutral on these models** — 1.02× at ratio 4.24, 1.12× at 8.53 — as the review predicted: every color there reaches nearly every body, because a body's row-variables all share its Hessian rows and so pairwise conflict, which is exactly the 42 colors the coloring reports. Its value is removing an unbounded blind spot, not speed here.

One defect found during this work: the `eval_h` accumulators must be **separate buffers** from `eval_jac_g`'s prelude adjoint arena. `gradient_summand` leaves that arena dirty after each row, while the directional sweeps rely on an all-zero-between-colors invariant — sharing the buffer seeds `eval_h` with a stale row gradient on every real solve iteration.

## Blast radius

- Models with no CSE shared across summands (including every `.nl` Pyomo writes — zero `V` segments, see #552) never build the hybrid tape: **bit-identical**, zero cost.
- Models with shared CSEs below ratio 3.0 keep `eval_h` on the flat path: **bit-identical**.
- Above the gate, `eval_h` values move by rounding (few ULPs) relative to the flat path, same in kind as what #553 accepted for the Jacobian. All 62 repository `.nl` fixtures have no shared CSE, so none of them change at all.
- `hessian_vector_products` (matrix-free HVP) intentionally stays on the flat tapes — arbitrary user seeds don't batch by color.
- Removed pub API: `HybridTape::hessian_summand` (never had a caller in-repo or in tests; pre-1.0). `HybridTape::prelude_tangent` / `prelude_reverse_directional` take a `reach: &[u32]` argument.

## Tests

- `shared_cse_hessian_matches_flat_tape_bit_for_bit` — hybrid vs flat with `assert_eq!` on a polynomial model whose inputs and coefficients make **every** operation in both traversals exact, so bitwise equality is a theorem, not luck; plus an analytic Lagrangian-Hessian cross-check and a λ = 0 row to exercise the skip.
- `a_deep_shared_body_turns_the_hessian_gate_on_and_still_agrees` — a bounded `(log ∘ exp)^20` body turns the gate on naturally; hybrid vs flat within 1e-12 relative, with the hybrid Jacobian run first in solve order. This is the test that fails (verified by re-introducing the bug) if the Hessian accumulators share `eval_jac_g`'s arena.
- `per_color_prelude_reach_skips_bodies_the_color_cannot_touch` — two shared bodies of differing width, so surplus colors belong to the wide block alone. Asserts the reach table is strictly smaller than the naive `n_colors × |prelude|` walk, that every reach list is ascending and operand-closed, and that the Hessian still matches flat. Deliberately does *not* claim to catch a revert of the sweep body: walking the whole prelude computes the same values, so only the table's narrowness and the safety invariants are checkable.
- `a_small_shared_body_leaves_the_hessian_on_the_flat_path` — gate off below ratio.
- `directional_hybrid_hessian_matches_flat_directional` — tape-level comparison against `Tape::hessian_directional`, including the accumulators-return-to-zero contract, passing a narrowed reach rather than the full range.
- **End-to-end** (`crates/pounce-cli/tests/issue_557_shared_cse_hessian.rs`): a generated model (one defined variable feeding 16 rows, analytic optimum) solved through the CLI with default gates and with `POUNCE_DBG_NO_HYBRID=1`; both must report Solved at the analytic optimum and agree to 1e-6. Gate states are asserted from the solve's own `[hybrid stats]` line, so raising the constant past this model's 8.25× cannot silently turn it into a flat-versus-flat comparison — verified by doing exactly that and watching both tests fail.

---

- [x] Tests bite: the interleave regression test and the e2e gate assertions were both negatively verified against the defect they guard
- [ ] Book page under `docs/src/` — not applicable: the hybrid tape is internal AD machinery with no user-facing surface, same as #476/#553 which added no page
- [x] `cargo fmt --all -- --check` clean; `cargo clippy` clean at CI strictness (`-D clippy::correctness -D clippy::suspicious`); `cargo test -p pounce-nl` (154 tests) and the CLI e2e tests pass
- [x] Every claim in this body and in the code comments is true of the code as it stands now — the measurement table and the whole-stack figure were re-measured for this revision, and the gate constant updated to match

https://claude.ai/code/session_016s9XGvcs9ST6pnRusL2Z2P

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,59 @@ changes.
 
 ## [Unreleased]
 
+### Changed — `eval_h` routes through the shared-CSE tape, amortizing prelude second-order work across summands (#557)
+
+The follow-on to #476 (`eval_g`) and #553 (`eval_jac_g`): the Lagrangian
+Hessian — the largest of the three evaluators by a wide margin, 66–69% of
+AD time on shared-CSE chain models — now takes the `HybridTape` path too,
+above its own measured op-ratio gate.
+
+Unlike the Jacobian, the Hessian can share **both** second-order sweeps of
+the prelude, which is why its crossover sits lower. The coloring hands
+every summand of a color the same seed vector `s_c`, so per color the
+prelude forward tangent runs once for the whole constraint block; and
+because reverse-over-tangent is linear in its adjoint seeds, each summand
+folds its row multiplier `λ_k` into the adjoints it deposits at `Shared`
+boundaries, and one unit-weight prelude reverse sweep per color replaces
+the per-summand sweeps. The per-summand local ops keep the exact
+arithmetic of the flat tape (`ror_dir_step` mirrors
+`Tape::hessian_directional`'s arms, writing into the dense per-color
+`compressed` buffer — no hashing), so the local contributions are
+bit-identical; the folded prelude sweep is mathematically identical and
+agrees to rounding, which the tests pin as exact equality on
+all-dyadic-arithmetic models and a few-ULP band on transcendental ones.
+
+Measured on chain models with CSE redundancy 40, varying the shared body
+size (`eval_h`, flat → hybrid, n ≈ 21,000, m = 20,000):
+
+| flat/shared op ratio | 1.94 | 2.54 | 3.12 | 3.69 | 4.24 | 5.29 | 6.76 | 8.53 |
+|---|---|---|---|---|---|---|---|---|
+| speedup | 1.04× | 1.11× | 1.30× | 1.36× | 1.33× | 1.56× | 1.50× | 1.63× |
+
+At 1.94 repeated runs straddle 1.0× — break-even inside timing noise,
+never a clear loss. The gate (`HYBRID_HESS_MIN_OP_RATIO`) is set at 2.5,
+below the Jacobian's 4, and low-ratio models stay bit-identical on the
+flat path. At ratio 4.24 with all default gates the whole AD stack drops
+34.6 → 23.8 ms/call (`eval_h` 24.8 → 18.3). Models without shared CSE
+bodies — including everything Pyomo writes, which has no `V` segments —
+never build the hybrid tape and are untouched.
+
+The dormant `HybridTape::hessian_summand` is **removed**. It was the
+wrong tool for this and had never had a caller: it re-walked
+`prelude_reach` once per seed variable (sharing no prelude work at all)
+and scattered through a `HashMap` lookup per emitted pair. Its
+replacements are `prelude_tangent` / `hessian_summand_directional` /
+`prelude_reverse_directional`.
+
+Instrumentation and coverage: `POUNCE_DBG_FORCE_HYBRID_HESS=1` forces the
+Hessian gate on (the crossover table's measurement instrument, paired
+with the existing `POUNCE_DBG_NO_HYBRID=1` flat forcing);
+`POUNCE_DBG_TAPE_STATS=1` now prints both gates' on/off state. The
+shared-CSE paths also gain their first end-to-end coverage — no
+repository fixture had a CSE shared across constraints, all 62 checked —
+via a generated model whose one defined variable feeds 16 rows, solved
+through the CLI both hybrid and flat to the same analytic optimum.
+
 ### Fixed — a single dense Hessian row made `.nl` setup and every `eval_h` O(n²) (#552)
 
 `NlTnlp` recovers the Lagrangian Hessian by graph coloring: columns whose
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,19 +32,34 @@ agrees to rounding, which the tests pin as exact equality on
 all-dyadic-arithmetic models and a few-ULP band on transcendental ones.
 
 Measured on chain models with CSE redundancy 40, varying the shared body
-size (`eval_h`, flat → hybrid, n ≈ 21,000, m = 20,000):
+size (`eval_h`, flat → hybrid, n ≈ 21,000, m = 20,000). Median of 5
+interleaved flat/hybrid pairs per point, with the observed range:
 
 | flat/shared op ratio | 1.94 | 2.54 | 3.12 | 3.69 | 4.24 | 5.29 | 6.76 | 8.53 |
 |---|---|---|---|---|---|---|---|---|
-| speedup | 1.04× | 1.11× | 1.30× | 1.36× | 1.33× | 1.56× | 1.50× | 1.63× |
+| median speedup | 1.00× | 1.04× | 1.18× | 1.23× | 1.31× | 1.44× | 1.36× | 1.49× |
+| min–max | 0.91–1.11 | 0.36–1.14 | 1.10–1.33 | 1.16–1.30 | 1.23–1.32 | 1.19–1.54 | 1.27–1.50 | 1.36–1.65 |
 
-At 1.94 repeated runs straddle 1.0× — break-even inside timing noise,
-never a clear loss. The gate (`HYBRID_HESS_MIN_OP_RATIO`) is set at 2.5,
-below the Jacobian's 4, and low-ratio models stay bit-identical on the
-flat path. At ratio 4.24 with all default gates the whole AD stack drops
-34.6 → 23.8 ms/call (`eval_h` 24.8 → 18.3). Models without shared CSE
-bodies — including everything Pyomo writes, which has no `V` segments —
-never build the hybrid tape and are untouched.
+The medians are what the threshold is set from: single runs on a shared
+machine scatter too widely to place a gate on — the 2.54 column spans
+0.36× to 1.14×. The gate (`HYBRID_HESS_MIN_OP_RATIO`) sits at 3.0, the
+lowest ratio where every sample wins by at least 10%; below it `eval_h`
+stays on the flat path bit-identically, so a gate set high is merely
+conservative while one set low risks a real regression. With all default
+gates the whole AD stack at ratio 4.24 speeds up 1.25× (median of 5
+interleaved pairs, range 1.14–1.44). Models without shared CSE bodies —
+including everything Pyomo writes, which has no `V` segments — never
+build the hybrid tape and are untouched.
+
+The two prelude sweeps run once per Hessian color, so they walk the union
+of that color's summands' `prelude_reach` rather than the whole prelude.
+Walking all of it would cost `n_colors × |prelude|` where the op-ratio
+gate assumes `|prelude|` — a discrepancy the gate cannot see, unbounded
+in `n_colors`. On the chain models above this is neutral (every color
+reaches nearly every body there, so there is nothing to skip: 1.02× at
+ratio 4.24, 1.12× at 8.53); it is insurance for models where a color
+reaches only part of the prelude, which a unit test on two shared bodies
+of differing width exercises directly.
 
 The dormant `HybridTape::hessian_summand` is **removed**. It was the
 wrong tool for this and had never had a caller: it re-walked
--- a/crates/pounce-cli/tests/issue_557_shared_cse_hessian.rs
+++ b/crates/pounce-cli/tests/issue_557_shared_cse_hessian.rs
@@ -105,8 +105,10 @@ fn shared_cse_model_nl(m: usize, pairs: usize) -> String {
 }
 
 /// Solve the generated model, optionally with `POUNCE_DBG_NO_HYBRID=1`,
-/// returning the parsed report.
-fn solve(no_hybrid: bool) -> SolveReport {
+/// returning the parsed report and the `[hybrid stats]` line from stderr
+/// (`POUNCE_DBG_TAPE_STATS=1` is always on so the caller can assert which
+/// paths the solve actually took).
+fn solve(no_hybrid: bool) -> (SolveReport, String) {
     let nl_path = tmp_path("model.nl");
     std::fs::write(&nl_path, shared_cse_model_nl(M, PAIRS)).expect("write model");
     let json_path = tmp_path("report.json");
@@ -115,20 +117,30 @@ fn solve(no_hybrid: bool) -> SolveReport {
     cmd.arg(&nl_path)
         .arg(&sol_path)
         .arg("--json-output")
-        .arg(&json_path);
+        .arg(&json_path)
+        .env("POUNCE_DBG_TAPE_STATS", "1");
     if no_hybrid {
         cmd.env("POUNCE_DBG_NO_HYBRID", "1");
     }
-    let status = cmd.status().expect("spawn pounce");
+    let out = cmd.output().expect("spawn pounce");
     assert!(
-        status.success(),
+        out.status.success(),
         "solve exited nonzero (no_hybrid={no_hybrid})"
     );
+    let stderr = String::from_utf8_lossy(&out.stderr).into_owned();
+    let stats = stderr
+        .lines()
+        .find(|l| l.contains("[hybrid stats]"))
+        .unwrap_or_default()
+        .to_string();
     let text = std::fs::read_to_string(&json_path).expect("read json report");
     let _ = std::fs::remove_file(&nl_path);
     let _ = std::fs::remove_file(&json_path);
     let _ = std::fs::remove_file(&sol_path);
-    serde_json::from_str(&text).expect("deserialize SolveReport")
+    (
+        serde_json::from_str(&text).expect("deserialize SolveReport"),
+        stats,
+    )
 }
 
 /// Analytic optimum: x0 = x1 = 0.5, x_{i+2} = 1 / 2.01.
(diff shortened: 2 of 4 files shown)


jkitchin:
## Review: correctness verified independently; four non-blocking cleanups

I checked this out, read the new code against the flat implementation it has to
agree with, ran the tests, and wrote a throwaway adversarial test of my own
rather than relying on the ones shipped here.

### What I verified

- `cargo test -p pounce-nl --lib` — 153 pass; both new CLI e2e tests pass.
- A scratch test with a summand referencing the same shared CSE **twice**, a
  **nested** promoted CSE (one prelude body referencing another), a CSE-free
  summand, and a zero multiplier — with `local_dot` / `local_adj` /
  `local_adj_dot` **NaN-poisoned and reused across summands and seeds with no
  re-zeroing by the caller**. Hybrid matches `Tape::hessian_directional` to
  1e-12 on three seeds, no NaN leaks, and `prelude_adj` / `prelude_adj_dot`
  come back all-zero. Negative control: dropping the `weight *` from the
  `Shared`-boundary fold makes it fail, so the test bites.
- The λ-fold is sound. Every reverse-over-tangent arm is linear in `(w, wd)`
  jointly and emission is `out[k] += weight * wd`, so folding λ at the boundary
  and running one unit-weight prelude sweep is exact up to reassociation, as
  claimed.
- The `local_dot` no-pre-zero invariant holds: `op_operands` is complete for
  every op a `HybridTape` summand can contain (Select / Min / Max / Funcall are
  rejected at build), so `local_reach` really does contain every operand of
  every reached op.
- `hess_color_summands` matches the flat path's `tape_colors` exactly — same
  `var_color` map, same `u32::MAX` filter, same sort/dedup — and
  `all_vars = local_vars ∪ prelude_vars`, so prelude-only variables are not
  dropped from a summand's color set.
- The reordered multiplications in `fwd_dir_step` (`dot[a] * vals[i]` →
  `vals[i] * dot[a]`) are bitwise no-ops; they align it with
  `hessian_directional`'s inline arms.

Nothing below blocks the merge.

### 1. The gate has no `n_colors` term, and the prelude sweeps are per color

`prelude_tangent` walks the **entire** prelude for every color, and
`prelude_reverse_directional` scans the whole prelude per color too (skipping
zeros, but the scan is still O(|prelude|)). So the hybrid's shared cost is

    n_colors_live × |prelude| + Σ_(color, summand) local_ops

while the gate compares only `flat_ops` against `|prelude| + Σ local_ops`. There
is no `n_colors` anywhere in the decision.

On the chain models in the crossover table this is close to optimal — greedy
coloring on block-disjoint structure hands the same colors to every block, so
each color's summands reach essentially the whole prelude and the sweep is not
wasted. The case that is not covered is a model where a color's summands reach
only a slice of the prelude; there the hybrid pays for prelude ops it does not
use, by up to a factor of `n_colors`, and the op-ratio gate cannot see it.

Cheap way to close it off rather than argue about how reachable it is: build a
per-color union of `prelude_reach` (over that color's summands) at setup and
iterate that in both prelude sweeps instead of `0..prelude.len()`. It is the
same table-building pass that already produces `hess_color_summands`, it makes
the cost proportional to what is actually used, and it removes the need for the
gate to model `n_colors` at all.

### 2. The e2e test does not pin the gate on

`crates/pounce-cli/tests/issue_557_shared_cse_hessian.rs` relies on the
generated model's op ratio (≈ 8) clearing `HYBRID_HESS_MIN_OP_RATIO` (2.5). If
that constant is ever raised past 8, the test silently degrades to comparing the
flat path against itself and still passes — the one thing an e2e test for this
feature must not do.

This PR already added the instrument: run the default solve with
`POUNCE_DBG_TAPE_STATS=1` and assert the `[hybrid stats]` line on stderr reports
`hess_gate=on` (and `jac_gate=on`, which the module doc also claims). Defense in
depth only — `a_deep_shared_body_turns_the_hessian_gate_on_and_still_agrees`
does assert `use_for_hess` at unit level — but the e2e is the one that will
outlive changes to the constant.

### 3. `op_operands`'s `Funcall` arm is the last silent one

Every other unsupported op in `op_operands` is an `unreachable!` with an
explanation. `TapeOp::Funcall(_) => (None, None)` returns silently instead, so a
Funcall reaching that walk would produce a `local_reach` missing its tape
arguments. That was survivable before, when `hessian_summand` zeroed the touched
slots each pass; now that `local_dot` is deliberately **not** pre-zeroed, an
incomplete reach reads stale scratch from a previous summand rather than zero —
a wrong Hessian, not a zero one.

It is unreachable today (`build_into_summand` panics on `Expr::Funcall`, and
`hybrid_promoted_cse_with_funcall_reports_clear_message` covers it), so this is
purely about not leaving one silent path in a function whose contract just got
stricter. Making it `unreachable!` like its neighbours costs nothing.

### 4. Doc nit

`Summand::all_vars` is still documented as the "Hessian j-loop set". The j-loop
went away with `hessian_summand`; the field is now the per-summand color set
that `eval_h`'s color-major table is built from.

---

jkitchin:
## All four addressed — and your item 1 exposed that the crossover table was wrong

Pushed as 15284ce. Thanks for the adversarial test in particular; the NaN-poisoned scratch reuse is the invariant I most wanted an independent check on.

### 1. Per-color prelude reach

Done as you suggested. Both sweeps now take a `reach: &[u32]` and iterate the union of that color's summands' `prelude_reach`, built once at setup in the same pass as `hess_color_summands`. Stored as a flat CSR rather than `Vec<Vec<_>>` — the total is `Σ_c |reach_c|`, proportional to the work it drives, which seemed worth caring about given #552 made this exact struct quadratic with per-color dense arrays.

Then I measured it in isolation, and **your prediction was right, including the part I doubted.** I had guessed the chain models would benefit a lot, reasoning that dense-column peeling would create ~40 singleton colors each reaching one body. That was wrong. The real structure is that a body's 40 row-variables all have nonzeros in the body-variable rows, so they pairwise conflict — which is exactly the 42 colors the coloring reports, reused across all 500 blocks. Every color reaches nearly every body, so there is almost nothing to skip:

| | full-range | narrowed |
|---|---|---|
| ratio 4.24 | 23.56 ms | 23.11 ms (1.02×) |
| ratio 8.53 | 32.60 ms | 29.10 ms (1.12×) |

So it is insurance, not speed, on the measured models — which is how you framed it. `per_color_prelude_reach_skips_bodies_the_color_cannot_touch` builds two shared bodies of differing width so the surplus colors belong to the wide block alone, and asserts the reach is strictly smaller than the naive walk plus the ascending/operand-closed invariants. I've written in the test doc that it deliberately does *not* claim to catch a revert of the sweep body — walking the whole prelude computes the same values, so only the table's narrowness and the safety invariants are checkable, and a timing assertion would be flaky.

### 2. e2e gate assertion

Done, via the `[hybrid stats]` line as you suggested; both `hess_gate=on` and `jac_gate=on` are asserted, and the `POUNCE_DBG_NO_HYBRID` run asserts `hybrid not built` so the two runs are pinned to different paths. Negative control: raising `HYBRID_HESS_MIN_OP_RATIO` to 50 makes both e2e tests fail with `hess_gate=off` in the message, where before they passed silently comparing flat against flat.

### 3 and 4

`op_operands`'s `Funcall` arm is now `unreachable!` with the reason you gave — that an incomplete reach is now a *wrong* Hessian rather than a zero one, because `local_dot` is deliberately not pre-zeroed. `all_vars`'s doc no longer mentions the j-loop.

### The table the gate was set from was not reproducible

Item 1 forced a re-measure, and re-measuring surfaced a problem of my own making. The published table was **one run per point**. On this machine that does not reproduce: re-running a single ratio five times spans 0.36×–1.14×. The non-monotonic points I published as data (3.69 → 1.36 against 4.24 → 1.33) were noise I presented as signal, and "from 2.5 up the win is consistent" was not true of individual runs.

Re-measured as the median of 5 interleaved flat/hybrid pairs — interleaved because absolute times drift over minutes while the paired ratio stays tight (±4% at ratio 4.24 against ±25% unpaired):

| flat/shared op ratio | 1.94 | 2.54 | 3.12 | 3.69 | 4.24 | 5.29 | 6.76 | 8.53 |
| median | 1.00× | 1.04× | 1.18× | 1.23× | 1.31× | 1.44× | 1.36× | 1.49× |
| min–max | 0.91–1.11 | 0.36–1.14 | 1.10–1.33 | 1.16–1.30 | 1.23–1.32 | 1.19–1.54 | 1.27–1.50 | 1.36–1.65 |

**The gate moves 2.5 → 3.0.** 3.12 is the lowest ratio where every sample wins, by ≥ 10%; at 2.54 the median win is 4% and one run in five lost. The asymmetry argues for the conservative side: below the gate `eval_h` stays on the flat path bit-identically, so a gate set high forgoes a marginal gain while one set low risks a real regression.

The whole-stack figure went the same way. "34.6 → 23.8 ms/call" came from one lucky pair; interleaved it is **1.25×** (median of 5, range 1.14–1.44). PR body and CHANGELOG both updated to medians with ranges.

For what it's worth, this is the same failure mode as the #552 overhead retraction — a single unrepeated wall-clock run on a busy machine — which is what prompted me to go back and check my own numbers rather than only the ones under review.

### State

`cargo test -p pounce-nl` 154 pass (+1 for the new test), CLI e2e pass, `cargo fmt --all -- --check` and clippy clean at CI strictness.

One loose thread you flagged on #554 that this PR now makes cheaper: `POUNCE_DBG_TAPE_STATS=1` on `steering_12800` / `rocket_12800` prints `n_colors` **and** the flat/shared op ratio with both gate states on one line. Those are AMPL-written, the population #557 claims benefits, so that single run would settle #554's open question and give this gate its first real-model evidence beyond `robot_a`'s 4.03× — which I inherited from #476 rather than measured. I can't build those models here.

---

Issue #341, opened 2026-07-23
[Adversary] pyomo_pounce silently solves models with Binary/Integer variables as a continuous relaxation, reports optimal


jkitchin:
## Problem

`pyomo_pounce`'s `POUNCE` solver plugin (`pyomo-pounce/pyomo_pounce/pounce_solver.py:106`) subclasses Pyomo's generic `ASL` solver class and never overrides its declared capabilities. `ASL.__init__` unconditionally sets:

```python
self._capabilities.integer = True
self._capabilities.sos1 = True
self._capabilities.sos2 = True
```

(`pyomo/solvers/plugins/solvers/ASL.py:57-63`). POUNCE is a continuous NLP solver — it has no integer-handling machinery — so this is false advertising, and nothing in `pounce_solver.py`'s `solve()` checks the model for discrete-domain variables before dispatch.

## Repro

```python
import pyomo.environ as pyo
import pyomo_pounce  # noqa: F401  (registers the 'pounce' plugin)

m = pyo.ConcreteModel()
m.b = pyo.Var(domain=pyo.Binary)
m.obj = pyo.Objective(expr=(m.b - 0.3) ** 2, sense=pyo.minimize)

solver = pyo.SolverFactory("pounce")
res = solver.solve(m)
print(res.solver.termination_condition)  # optimal
print(pyo.value(m.b))                    # 0.30000000215628075
```

`m.b` is declared `Binary` and the reported "optimal" value is `0.3`, a fractional value outside `{0, 1}` — silently. `res.solver.status` is `ok`, `termination_condition` is `optimal`, and there is no warning anywhere in the log or the `SolverResults` object that integrality was ignored.

`SolverFactory('pounce').has_capability('integer')` also reports `True`, which is simply wrong for this solver.

## Why this matters despite matching Ipopt

Side-by-side with `SolverFactory('ipopt')` on the identical model, Ipopt does the exact same thing (`b = 0.30000000243...`, `termination_condition = optimal`) — this is inherited AMPL/ASL-ecosystem behavior, not a pounce-specific regression, and I initially did not file it for that reason.

[...]

Issue #315, opened 2026-07-23
[Enhancement] pyomo-pounce: detect/warn when SolverFactory('pounce') resolves a stale or shadowing PATH binary


jkitchin:
> Surfaced while building the solver-equivalence validation suite (#301). A silent footgun for Pyomo/AMPL users, made concrete by the #271/#272 dual-sign fix.

## Symptom

`SolverFactory('pounce').solve(model)` can silently run a **different `pounce` binary than the one you installed**, returning results from stale code. It was caught because a stale binary predating the #271 dual-sign fix returned **flipped constraint duals** through Pyomo — every `model.dual[...]` negated — while objectives and primals looked correct, so nothing obviously failed.

## Mechanism (verified on this machine)

There are two resolution paths and only one is hardened:

1. **Plugin path (hardened, works):** when `import pyomo_pounce` has run, `PounceSolver._default_executable` (`pyomo-pounce/pyomo_pounce/pounce_solver.py:61`) already prefers the bundled `pounce/bin/pounce` from the installed wheel over PATH — its own comment notes the binary "can be shadowed by a stale system binary." Good.

2. **ASL fallback path (the gap):** the `@SolverFactory.register("pounce")` decorator only runs when `pyomo_pounce` is imported. If a user calls `SolverFactory('pounce')` **without importing `pyomo_pounce`** (easy to forget — the package name differs from the solver name), Pyomo does not use the plugin at all. It constructs a generic ASL solver that searches `PATH` for an executable named `pounce` — and picks up whatever is first, e.g. a stale `~/.local/bin/pounce`. The plugin's careful resolver never runs, so it cannot warn.

Concretely here: `which -a pounce` returns `~/.local/bin/pounce` first (dated before the #271 merge); it and the current build **both report `pounce 0.9.0`**, so a version-string check cannot tell them apart even though they differ across the dual-sign fix.

## Why this is worth hardening

- pounce advertises drop-in IPOPT compatibility; `SolverFactory('pounce')` is the primary Pyomo entry point, and forgetting `import pyomo_pounce` is a natural mistake.
- The failure is silent and, for the dual-sign case specifically, corrupts sensitivity/shadow-price analysis while leaving the headline objective correct — the hardest class to notice.
- **Version strings collide** (`0.9.0` == `0.9.0` across the fix), so any check must be finer-grained than the version string.

## Proposed enhancement

[...]

PR #560, opened 2026-08-09
feat(pyomo): register a native pyomo.contrib.solver (v2) interface

### LINKED ISSUE #558: pyomo-pounce: register a native pyomo.contrib.solver (v2) interface
`pyomo_pounce.POUNCE` subclasses `pyomo.solvers.plugins.solvers.ASL.ASL` —
Pyomo's **legacy** solver interface — overriding only `solve`, `__init__` and
`_default_executable`. There is no registration against
`pyomo.contrib.solver`, Pyomo's newer interface.

#553 made the POUNCE binary *drivable* by the v2 interface (it previously
failed on every model: quoted option values, and a `.sol` `Options` block the
v2 reader rejects). That is not the same as having a v2 interface — a user
must point `ipopt_v2` at the executable by hand, and doing so silently drops
everything `pyomo-pounce` adds.

## What works today, and what it costs

Verified by running each against Pyomo 6.10.1 (`PATH` = path to the binary):

| call | works | carries `pyomo-pounce`'s extras |
|---|---|---|
| `SolverFactory('pounce')` | yes | **yes** |
| `SolverFactory('ipopt_v2', executable=PATH)` | yes | no |
| `SolverFactory('ipopt', executable=PATH)` | yes | no |
| `SolverFactory('asl', executable=PATH, solver='pounce')` | yes | no |
| `SolverFactory('appsi_ipopt', …)` | no — takes no `executable` | — |

The generic routes return the same answer, but without:

- `scaling_factor` suffix handling (gh#483 / #486),
- the sensitivity path (`declare_sens_param` → in-process `pounce.Solver`),
- the preflight / repair / block-init helpers,
- the guard that refuses a model with live integer variables rather than
  silently solving the continuous relaxation and reporting a fractional value
  as optimal (gh#341),
- bundled-binary resolution, which is what keeps a stale `pounce` on `PATH`
  from being picked up silently (gh#315).

So a user who reaches for `ipopt_v2` today — reasonably, since it is the
modern interface — gets a quietly less-safe solve.

## Scope

- Register a `pyomo.contrib.solver`-style solver alongside the existing legacy
  one, so `SolverFactory('pounce')` keeps working unchanged.
- Decide how the extras map onto the v2 lifecycle. The v2 interface returns a
  `Results` object and loads the solution directly, rather than going through
  the legacy `SolverResults` / `load_from` path, so the sensitivity route in
  particular needs a deliberate translation rather than a copy.
- Keep the two ASL-compatibility behaviours from #553 exercised: quoted option
  values, and the per-model `.sol` `Options` echo.

## Testing gap this is a good moment to close

Pyomo is **not** a dev-dependency of this workspace. The #553 fixes are pinned
by Rust unit tests only, and the interface table above (now in
`docs/src/pyomo.md`) was produced by hand — nothing in CI re-checks it, so it
goes stale the moment Pyomo renames an interface. There is a `wheel smoke
(pyomo-pounce)` CI job already; extending it to solve one model through both
the legacy and v2 routes and compare would cover both this work and the #553
fixes end-to-end.

## Not urgent

Nothing is broken today: `SolverFactory('pounce')` is the documented, supported
route and carries everything. This becomes load-bearing when Pyomo makes the v2
interface the default `ipopt` — I have not verified their roadmap, so this is
worth revisiting rather than scheduling on a guess.

jkitchin:
Fixes #558

## Problem

`pyomo_pounce` registered exactly one solver: `POUNCE`, a subclass of Pyomo's **legacy** `ASL` plugin. #553 made the POUNCE binary *drivable* by `pyomo.contrib.solver` (v2), but that is not the same as having a v2 interface — a user had to point `ipopt_v2` at the executable by hand, and doing so silently dropped everything `pyomo-pounce` adds.

Measured against Pyomo 6.10.1 on this branch's binary, minimising `(b - 0.37)²` over a variable declared `Binary`:

| route | `version()` | `available()` | binary variable |
|---|---|---|---|
| `Ipopt(executable=<pounce>)` (generic v2) | `None` | `NotFound` | returns `b* = 0.370000` as **`optimal`** |
| `contrib.solver` `SolverFactory('pounce')` (this PR) | `(0, 9, 0)` | `FullLicense` | refuses with the #341 error |

Both cells on the top row are the bug. The `available()` one is not fatal — nothing in the v2 solve path gates on the version, so solves still run — but every availability check lies, including the `if not solver.available(): skip` guard most Pyomo suites use. The binary-variable one is the #341 silent wrongness: a fractional value reported as optimal for a variable the user declared discrete.

Beyond the extras, this is also a performance fix. [#552](https://github.com/jkitchin/pounce/issues/552#issuecomment-5229601244) measured the same POUNCE binary through both interfaces on drto/IDAES collocation models:

| model | legacy remainder | v2 remainder |
|---|---|---|
| plain `pyomo.dae` four-tank, n = 3,010 | 0.109 s | 0.104 s |
| drto/IDAES `quad_tank` N=100, n = 2,910 | 0.553 s | 0.301 s |
| drto/IDAES `cart_pole` N=100, n = 2,810 | 0.566 s | 0.295 s |

On that model class the legacy path adds ~0.25 s per solve (~1.8×), and `SolverFactory('pounce')` is the documented route — so those users were paying it with no supported alternative that kept the extras.

## Cause

Nothing was broken; the interface simply did not exist. The one substantive incompatibility uncovered while building it is the version banner: `pounce --version` prints `pounce X.Y.Z`, and Pyomo's `Ipopt._get_version` requires a leading `ipopt` **by design**, precisely so that other ASL executables are not mistaken for Ipopt. POUNCE is ASL/Ipopt-compatible on every other inherited surface — `key=value` options, `option_file_name=<path>`, and a log carrying the `Number of Iterations....:` and `Total seconds in POUNCE =` lines Pyomo's parser reads.

## Fix

`pyomo_pounce.v2.Pounce`, registered alongside the legacy plugin using Pyomo's own `ipopt` / `ipopt_v2` split, so the change is purely additive:

| call | before | after |
|---|---|---|
| `SolverFactory('pounce')` | legacy ASL plugin | **unchanged** |
| `contrib.solver` `SolverFactory('pounce')` | — | v2 interface, all extras |
| `SolverFactory('pounce_v2')` | — | v2 engine, legacy API |

**Subclass `Ipopt`, not `SolverBase`.** That inherits the `.nl` write, `.sol` read, option splitting and log parse, all of which POUNCE is compatible with. Overridden: `_get_version`, the `executable` config default (bundled binary first, #315), and `solve`.

**The extras, translated onto the v2 lifecycle.** The integer guard was extracted to a shared `reject_discrete_vars` — a guard protecting only the legacy route would leave the modern one with exactly the silent wrongness it exists to prevent. Scaling-suffix handling and bundled-binary resolution port directly.

The sensitivity path needed the deliberate translation #558 called for, not a copy. v2 returns a `Results` and hands the solution back through a loader the caller may decline to load; the legacy path writes onto the model as a side effect of solving. So:

- `sens_solve` gains an optional `capture=` mapping recording the raw solve (primal iterate, the engine's `info` with multipliers, `.col`/`.row` order, surgery aliases). Populated before the non-converged early return, so a failed solve — exactly when the session is dropped — is still reportable. The legacy path passes nothing and pays nothing.
- `PounceSensSolutionLoader` serves primals, duals and reduced costs from that capture. Two sign conventions are crossed on the way out, inverting what `_warm_start_from_suffixes` does on the way in: `dual` is the AMPL marginal `-lambda` against the engine's internal `mult_g` (#271), and `ipopt_zU_out` is negative at an active upper bound (#296) against the internal non-negative `mult_x_U`. The two bound multipliers are then combined exactly as `IpoptSolutionLoader` combines the `.sol` suffixes, so `rc` means the same thing on both routes.
- With `load_solutions=False`, variable values are snapshotted and rolled back, since `sens_solve` writes them unconditionally. The retained KKT session is unaffected — it answers from its own `base_x`, not from the model.

**Rejected:** re-deriving the sens `Results` by adapting the legacy `SolverResults` it already returns. The enums are disjoint sets, and the loader contract is the substantive difference, so an adapter would have had to reconstruct the multipliers anyway.

**Pyomo floor.** `contrib.solver.common` landed in 6.9.2; the package declares `pyomo>=6.0`. Rather than bump the floor for every legacy user, the v2 registration is optional — `import pyomo_pounce` still works on older Pyomo, `HAVE_V2_INTERFACE` reports availability, and a `[v2]` extra asks for a new enough Pyomo. The probe is on Pyomo, not on `pyomo_pounce.v2`, so a genuine `ImportError` inside the new module surfaces loudly instead of reading as "unavailable".

## Blast radius

**No Rust changed** — this is entirely `pyomo-pounce/` plus docs and CI. The solver corpus cannot move.

`SolverFactory('pounce')` is bit-identical: the only edit to `pounce_solver.py` is extracting the integer guard into a function the legacy `solve` now calls, same error text. `sens_solve` gains one keyword defaulting to `None`; with it unset the function is unchanged.

The one behaviour users could notice: `pyomo.contrib.solver`'s `SolverFactory` now resolves the name `pounce`, which previously raised. Additive.

## Tests

`pyomo-pounce/tests/test_v2.py`, 18 tests. Full suite **273 passed** against a locally-built CLI and `pounce-solver` wheel — real solves throughout, no mocks.

The tests that carry the weight are the cross-route comparisons, because "both work" is not the property that matters:

- `test_v2_matches_legacy_primals_and_duals` — one model through both interfaces, comparing primals, objective, duals and reduced costs. The legacy ASL path does not synthesise `rc` (combining the bound multipliers is a v2-loader behaviour), so it reads raw `ipopt_zL_out`/`ipopt_zU_out` and does the combination by hand, asserting the multiplier is live (`> 1e-3`) rather than a zero that would compare equal by accident.
- `test_sens_route_matches_the_ordinary_route` — the in-process path against the `.sol` path, same four quantities. This is what actually validates the hand-written sign conversions above; it failed while I had them under review and passes now.
- `test_sens_route_honours_load_solutions_false` — pins the snapshot/rollback. **This one failed for the right reason during development**, catching that `sens_solve` writes onto the model unconditionally.
- `test_instance_options_reach_the_sensitivity_route` — #432 on the in-process path, which reads options itself rather than handing them to a subprocess.

**Honest note on the parent commit:** these are new tests for a new module, so on `main` they fail at collection because `pyomo_pounce.v2` does not exist. That is the trivial case, not evidence of biting. The substantive "before" behaviour is the measured table at the top, which I ran against the generic `ipopt_v2` route.

**Deliberately not pinned:** the per-solve overhead figures. They are hardware- and model-dependent (the plain-model row was measured on Linux, the IDAES rows on Windows), and the drto/IDAES models are not vendorable here. The docs table says so and attributes the measurements to #552 rather than presenting them as a CI-checked property.

**CI.** The `wheel smoke (pyomo-pounce)` job is the only place Pyomo runs at all, so it now also asserts the v2 interface resolves a binary and parses its version before the suite runs — a check a plain `import` would not make, given `available()` fails silently.

---

- [x] Tests fail on the parent commit — but see the honest note above: trivially, because the module is new
- [x] Book page under `docs/src/pyomo.md` updated (existing page; no `SUMMARY.md` change needed)
- [ ] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` — **not run: no Rust changed by this PR**

On that last box: a first draft of the CHANGELOG and one test docstring claimed that without the `_get_version` override "the solver refused to run". I measured it, found solves run fine and only `available()` misreports, and corrected both in f7be47c before opening this.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PmcbDVNJrL4H9MuoJSELdB)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -393,3 +393,42 @@ jobs:
           # AMPL/`.nl` tests exercise the CLI plugin path. Both catch
           # regressions the import smoke test misses.
           python -m pytest pyomo-pounce/tests -q
+      # The v2 interface needs Pyomo >= 6.10.1; the package as a whole
+      # supports pyomo>=6.0 through the legacy plugin. Everything above
+      # installs Pyomo unpinned, i.e. always the newest release, so
+      # nothing there exercises the declared floor — it is a claim, not a
+      # tested property, and the one time it was wrong the failure mode
+      # was `import pyomo_pounce` raising, which takes the LEGACY plugin
+      # down with it (gh #558 review).
+      #
+      # 6.10.0 is the last release below the floor and the one where this
+      # actually broke: `pyomo.contrib.solver.common` exists there, so a
+      # package-level probe passed, but the loader API is still
+      # `SolutionLoaderBase`/`get_primals`. A separate venv keeps this
+      # older Pyomo away from the job's real environment.
+      - name: Below-floor Pyomo — legacy plugin must still import
+        run: |
+          python -m venv /tmp/floorenv
+          /tmp/floorenv/bin/pip install -q "pyomo==6.10.0" numpy pandas
+          /tmp/floorenv/bin/pip install -q python/dist/*.whl
+          /tmp/floorenv/bin/pip install -q --no-deps pyomo-pounce/dist/*.whl
+          /tmp/floorenv/bin/python - <<'PY'
+          import pyomo
+          # Must NOT raise: the v2 registration is optional, and a failed
+          # import here is the regression this leg exists to catch.
+          import pyomo_pounce
+          from pyomo.opt import SolverFactory
+          print('pyomo', pyomo.version.version,
+                '| HAVE_V2_INTERFACE:', pyomo_pounce.HAVE_V2_INTERFACE)
+          assert pyomo_pounce.HAVE_V2_INTERFACE is False, (
+              "6.10.0 is below the v2 floor; the guard should report the "
+              "interface as unavailable rather than claiming it works")
+          assert SolverFactory('pounce') is not None, (
+              "the legacy plugin must keep working below the v2 floor")
+          print('legacy plugin OK below the floor')
+          PY
+          # ...and the v2 test module skips cleanly rather than erroring
+          # at collection.
+          /tmp/floorenv/bin/pip install -q pytest
+          /tmp/floorenv/bin/python -m pytest pyomo-pounce/tests/test_v2.py \
+            -q --no-header -rs 2>&1 | tail -5
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -282,6 +282,19 @@ jobs:
   # end-to-end against that binary. The pytest step exercises actual solves
   # (option forwarding, routing) that a plain import smoke test cannot —
   # without it, routing regressions like pounce#186 ship silently.
+  #
+  # This is also the ONLY job where Pyomo runs at all: it is not a
+  # dev-dependency of the workspace, so nothing else here can catch a
+  # Pyomo-side regression. Since gh#558 that includes both of the
+  # interfaces pyomo-pounce registers — the legacy ASL plugin and the
+  # `pyomo.contrib.solver` (v2) one — and `tests/test_v2.py` solves one
+  # model through both and compares primals, duals and reduced costs.
+  # That comparison is what keeps the interface table in
+  # `docs/src/pyomo.md` honest (it was hand-produced and went stale the
+  # moment Pyomo renamed anything) and what pins the two #553
+  # ASL-compatibility fixes from the Python side: quoted option values
+  # and the per-model `.sol` `Options` echo are exercised on every v2
+  # solve, where before they were covered by Rust unit tests only.
   pyomo-pounce-smoke:
     name: wheel smoke (pyomo-pounce)
     runs-on: ubuntu-latest
@@ -358,6 +371,20 @@ jobs:
           python -m pip install --no-deps pyomo-pounce/dist/*.whl
           which pounce
           python -c "import pyomo_pounce; from pyomo.opt import SolverFactory; s = SolverFactory('pounce'); print('exe:', s.executable())"
+          # Both registrations must resolve a usable binary before the
+          # suite runs. The v2 interface finds its executable and parses
+          # its version through a different code path than the legacy
+          # plugin (Pyomo's own Ipopt version parser rejects POUNCE's
+          # `pounce X.Y.Z` banner, so `available()` is NotFound without
+          # our override) — a plain `import` would not notice.
+          python - <<'PY'
+          import pyomo_pounce
+          from pyomo.contrib.solver.common.factory import SolverFactory as SF2
+          s = SF2('pounce')
+          print('v2 exe:', s.config.executable.path())
+          print('v2 version:', s.version(), 'available:', s.available())
+          assert s.version() is not None, "v2 interface could not parse the POUNCE version"
+          PY
       - name: Run pyomo-pounce tests
         run: |
           python -m pip install pytest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,14 +71,29 @@ through v2, ~1.8×. (On a plain `pyomo.dae` model the same comparison is
 is the documented route, users of that model class were paying the legacy
 path's cost with no supported alternative that kept the extras.
 
+**Requirements, and only for the v2 route** (`pip install
+pyomo-pounce[v2]`): **Pyomo ≥ 6.10.1**, where the `SolutionLoader` /
+`get_vars` API this builds on landed — `pyomo.contrib.solver.common`
+exists from 6.9.2, but 6.9.2–6.10.0 ship the older `SolutionLoaderBase` /
+`get_primals` — and **pounce-solver > 0.9.0**, because Pyomo's
+`asl_sol_reader` asserts `n_opts >= 2` where the legacy reader is lenient
+and so needs the per-model `.sol` `Options` echo added after 0.9.0.
+Neither applies to `SolverFactory('pounce')`: on an older Pyomo the
+legacy plugin behaves exactly as before and
+`pyomo_pounce.HAVE_V2_INTERFACE` reports `False`.
+
 **Testing.** `pyomo-pounce/tests/test_v2.py` solves the same model
 through both interfaces and compares primals, objective, duals and
 reduced costs, and does the same for the ordinary route against the
-sensitivity route. That closes the gap noted in #558: the interface table
-in `docs/src/pyomo.md` was produced by hand and nothing in CI re-checked
-it, and the #553 ASL-compatibility fixes (quoted option values, the
-per-model `.sol` `Options` echo) were pinned by Rust unit tests only —
-the v2 route exercises both on every solve.
+sensitivity route — including that they agree on `solution_status` for a
+limit-stopped solve, which decides whether `solve()` raises. That closes
+the gap noted in #558: the interface table in `docs/src/pyomo.md` was
+produced by hand and nothing in CI re-checked it, and the #553
+ASL-compatibility fixes (quoted option values, the per-model `.sol`
+`Options` echo) were pinned by Rust unit tests only — the v2 route
+exercises both on every solve. A CI leg pinned to a Pyomo *below* the
+floor checks that the legacy plugin still imports there, so the floor is
+a tested property rather than a claim.
 
 ### Fixed — a single dense Hessian row made `.nl` setup and every `eval_h` O(n²) (#552)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,71 @@ changes.
 
 ## [Unreleased]
 
+### Added — `pyomo-pounce` registers a native `pyomo.contrib.solver` (v2) interface (#558)
+
+`pyomo_pounce` previously registered one solver: `POUNCE`, a subclass of
+Pyomo's *legacy* `ASL` plugin. #553 made the POUNCE binary drivable by
+Pyomo's newer `pyomo.contrib.solver` interface, but that is not the same
+as having one — a user had to point `ipopt_v2` at the executable by hand,
+and doing so silently dropped everything `pyomo-pounce` adds.
+
+`pyomo_pounce.v2.Pounce` is now registered alongside the legacy plugin:
+
+    import pyomo_pounce
+    from pyomo.contrib.solver.common.factory import SolverFactory as SF2
+    SF2('pounce')                        # v2 interface
+    SolverFactory('pounce_v2')           # v2 engine, legacy API
+    SolverFactory('pounce')              # legacy plugin, unchanged
+
+The split follows Pyomo's own `ipopt` / `ipopt_v2` convention, so the
+registration is purely additive: `SolverFactory('pounce')` still returns
+the legacy plugin and behaves exactly as before.
+
+**The extras come with it.** That is the point of the class, and each
+needed translating onto the v2 lifecycle rather than copying:
+
+- the guard that refuses a model with live integer variables instead of
+  solving the continuous relaxation and reporting a fractional value as
+  optimal (#341);
+- `scaling_factor` Suffix handling (#483 / #486);
+- bundled-binary resolution, which keeps a stale `pounce` on `PATH` from
+  being picked up silently (#315);
+- the sensitivity path (`declare_sens_param` → in-process
+  `pounce.Solver`). This one is a real translation, not an adaptation:
+  v2 returns a `Results` and hands the solution back through a solution
+  loader the caller may decline to load, where the legacy path writes
+  values onto the model as a side effect of solving. The new
+  `PounceSensSolutionLoader` serves primals, duals and reduced costs out
+  of the in-process solve, converting the engine's internal multipliers
+  to the AMPL marginal convention `dual` carries and to Ipopt's
+  `ipopt_zU_out` sign — the same two conversions the warm-start reader
+  applies in the other direction.
+
+**Inherited deliberately** from Pyomo's `Ipopt` v2 class: the `.nl`
+write, the `.sol` read, option splitting between the command line and the
+`.opt` file, and the solver-log parse. POUNCE is ASL/Ipopt-compatible on
+all of them. The one place it is not is the version banner —
+`pounce --version` prints `pounce X.Y.Z`, which Pyomo's Ipopt parser
+rejects by design — so without the override here `available()` reported
+`NotFound` and the solver refused to run.
+
+**Why it matters beyond API parity.** #552 measured the same POUNCE
+binary through both interfaces on drto/IDAES collocation models: 0.553 s
+of non-solve time per solve through the legacy path against 0.301 s
+through v2, ~1.8×. (On a plain `pyomo.dae` model the same comparison is
+1.05×, so this is model-shape dependent.) Since `SolverFactory('pounce')`
+is the documented route, users of that model class were paying the legacy
+path's cost with no supported alternative that kept the extras.
+
+**Testing.** `pyomo-pounce/tests/test_v2.py` solves the same model
+through both interfaces and compares primals, objective, duals and
+reduced costs, and does the same for the ordinary route against the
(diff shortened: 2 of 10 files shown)


jkitchin:
Reviewed. The design is right — subclassing `Ipopt` rather than reimplementing
the NL/`.sol` plumbing, `legacy_name="pounce_v2"` to keep the split additive,
`reject_discrete_vars` extracted so both routes share the guard, and a solution
loader for the sensitivity path rather than an adapter over the legacy
`SolverResults`. The reduced-cost combination is correct: I diffed it against
Pyomo 6.10.1's `IpoptSolutionLoader.get_reduced_costs` and the arithmetic
(`hi if abs(hi) > abs(lo) else lo`, with `hi = -zu`) matches exactly. The
`capture` hook in `sens.py` is placed where the docstring claims — before the
non-converged early return — and `PounceSensSolutionLoader` sets
`_pyomo_model`, so the base class's `load_import_suffixes` works.

Two things block it, and one behavioural divergence should be decided rather
than inherited.

## 1. The Pyomo floor is 6.10.1, not 6.9.2 — and below it the *legacy* interface breaks

This is the blocker. On Pyomo 6.9.2 through 6.10.0, `import pyomo_pounce`
raises, which takes `SolverFactory('pounce')` down with it. That contradicts
"purely additive" and "the legacy plugin is bit-identical".

Verified on this machine (Pyomo 6.10.0):

```
IMPORT FAILED: ImportError pyomo_pounce.v2 needs Pyomo's
`pyomo.contrib.solver.common` interface, which is Pyomo 6.9.2 or newer. ...
  cause: cannot import name 'SolutionLoader' from
  'pyomo.contrib.solver.common.solution_loader'
```

On `main`, the same interpreter: `import OK, POUNCE`.

The cause is that `SolutionLoader`, `load_solution` and `get_vars` are
**6.10.1** names. In the 6.10.1 sdist they are at `solution_loader.py:28`,
`:128` and `:152`; 6.9.2 and 6.9.5 ship `SolutionLoaderBase` / `load_vars` /
`get_primals` and have no `class SolutionLoader` at all (checked both sdists).

The probe in `__init__.py` tests the wrong name:

```python
import pyomo.contrib.solver.common.factory as _probe
```

`factory.py` has existed since 6.9.2, so the probe passes on every version in
the broken window and then imports `pyomo_pounce.v2` unguarded, letting v2.py's
`ImportError` escape. Broken window: **6.9.2, 6.9.3, 6.9.4, 6.9.5, 6.10.0** —
five releases, all at or above the declared floor.

Two things made this invisible:

- `tests/test_v2.py` does `import pyomo_pounce` at line 21, *before* the
  `HAVE_V2_INTERFACE` skip at line 26. In the broken window the module errors
  at collection instead of skipping, so the guard that exists for exactly this
  case can never fire.
- CI installs `pip install pyomo` unpinned (ci.yml:118, :367), i.e. always the
  newest release. Nothing anywhere exercises the declared floor.

Fix: raise the floor to 6.10.1 in the `[v2]` extra, the v2.py module docstring,
the `ImportError` text, the `__init__.py` comment, the `test_v2.py` skip reason
and the CHANGELOG; make the probe import a 6.10.1-only name (`SolutionLoader`)
rather than the module; and move `import pyomo_pounce` in `test_v2.py` inside a
`try`/`skip` or below the guard. A CI matrix leg pinned to the floor would keep
it honest — right now the floor is a claim, not a tested property.

## 2. The `[v2]` extra under-declares its POUNCE binary requirement

I installed the branch into a clean Pyomo 6.10.1 venv against the released
`pounce-solver` 0.9.0 and the ordinary v2 route fails on every solve inside
Pyomo's strict reader:

```
pyomo/contrib/solver/solvers/asl_sol_reader.py:405: AssertionError
>       assert n_opts >= 2
```

That is the per-model `.sol` `Options` echo from #553 — on `main`, unreleased.
The v2 route needs it because `asl_sol_reader` is strict where the legacy
reader is lenient, so this is a requirement v2 introduces and the legacy plugin
does not have.

**Caveat, stated plainly:** the same venv also fails 102 pre-existing tests
from the same version skew (`'pounce._pounce.Solver' object has no attribute
'primal_rows'`), so this is not evidence the PR is broken — it is evidence that
`v2 = ["pyomo>=6.9.2"]` is missing a `pounce-solver >= <the release carrying
#553>` floor. CI builds the wheel from source and so will never see it; a user
running `pip install pyomo-pounce[v2]` today would.

I could not get a clean `test_v2.py` run for the same reason, so the suite is
unverified by me — only read.

## 3. The two routes disagree on status, in the direction that raises

`_V2_STATUS` was written from the POUNCE status strings; the ordinary route
gets its status from Pyomo's AMPL solve-code table. They do not agree. I ran
`asl_solve_code_to_solution_status` on 6.10.1 directly:

| POUNCE status | ordinary route | `_V2_STATUS` (sens route) |
|---|---|---|
| `Solve_Succeeded` | `convergenceCriteriaSatisfied` / `optimal` | same |
| `Infeasible_Problem_Detected` | `locallyInfeasible` / `infeasible` | same |
| `Feasible_Point_Found` | `error` / `feasible` | `unknown` / `feasible` |
| `Maximum_CpuTime_Exceeded` | `iterationLimit` / **`unknown`** | `maxTimeLimit` / **`noSolution`** |
| `Maximum_Iterations_Exceeded` | `iterationLimit` / **`unknown`** | `iterationLimit` / **`noSolution`** |
| `Diverging_Iterates` | `unbounded` / **`unknown`** | `unbounded` / **`noSolution`** |

The termination-condition differences are the PR being *more* correct, and
deliberately so — Pyomo's own source carries `# this is not always correct` on
the `iterationLimit` line, and the `Feasible_Point_Found` choice is argued in a
comment. Keep those.

The solution-status differences are the problem. `unknown` is loadable;
`noSolution` is not. So with default `load_solutions=True`:

```python
solver.solve(m, solver_options={"max_iter": 1})
```

loads the final iterate on the ordinary route and raises `NoSolutionError` on
the sensitivity route. Same interface, same model, same options — one raises.
`Diverging_Iterates` behaves the same way. POUNCE does return a usable iterate
in both cases, so `noSolution` is also the less accurate answer.

Both tests that touch the iteration limit
(`test_solver_options_reach_the_binary`,
`test_instance_options_reach_the_sensitivity_route`) pass
`load_solutions=False` *and* `raise_exception_on_nonoptimal_result=False`, so
neither can see it. Given that route agreement is the PR's stated goal and
`test_sens_route_matches_the_ordinary_route` exists to enforce it, this wants a
test that solves a limit-capped model both ways with default kwargs and asserts
the same `solution_status`.

## Smaller things

- **`incumbent_objective` is gated differently.** `_sens_solve` uses
  `has_solution = ss is not SolutionStatus.noSolution`, which includes
  `SolutionStatus.infeasible`; `Ipopt.solve` gates on
  `ss in {feasible, optimal}` (ipopt.py:418-419). An infeasible sens solve
  reports an objective the ordinary route leaves as `None`.
- **`get_vars` drops unresolvable names silently.** With `vars_to_load=None`,
  any `var_names` entry `find_component` cannot resolve is skipped, and
  `load_vars` then leaves that variable stale with no diagnostic. Given the
  declared-parameter surgery renames things, a count check (`len(out)` against
  `capture["n"]`) with a warning would be cheap insurance. `get_duals` and
  `get_reduced_costs` have the same shape via `_row_of`.
- **`get_solution_ids` is not overridden**, so `solution()` on the sens loader
  raises `NotImplementedError` from the base rather than something legible.
  One-liner returning `[0]` / `[]`.
- **Config is derived twice on the ordinary route** — `self.config(value=kwds,
  preserve_implicit=True)` in `solve`, then `super().solve(model, **kwds)`
  derives its own. Harmless (no mutation of `self.config`) but worth a comment,
  since the `_check_executable` warning is computed off a config object that is
  then thrown away.

Nothing here touches the design, which I think is the right one. Items 1 and 2
are release-correctness; item 3 is a decision to make explicitly rather than
inherit by accident.

---

jkitchin:
All three addressed in ad19dcd. I reproduced each before changing anything; item 1 was worse than a wrong version number, and one smaller item I implemented differently from the suggestion because the suggested form would have misfired.

## 1. Floor — confirmed, and it was my bug

You are right, including about the mechanism. Of the ten names `v2.py` imports, exactly one is absent on 6.10.0:

```
  ok      common/base.py :: LegacySolverWrapper        ok      common/util.py :: NoOptimalSolutionError
  ok      common/factory.py :: SolverFactory           ok      common/util.py :: NoSolutionError
  ok      common/results.py :: Results                 ok      solvers/ipopt.py :: Ipopt
  ok      common/results.py :: SolutionStatus          ok      solvers/ipopt.py :: IpoptConfig
  ok      common/results.py :: TerminationCondition
  MISSING common/solution_loader.py :: SolutionLoader
```

So `SolutionLoader` is not just *a* 6.10.1-only name, it is the only one that discriminates — and it is the one that tracks the API `v2.py` actually calls. That is what the probe imports now. On 6.10.0 in a clean venv:

```
OLD probe (common.factory):      PASSES  <- why the bug escaped
NEW probe (SolutionLoader):      fails correctly
import pyomo_pounce: OK | HAVE_V2_INTERFACE: False | SolverFactory('pounce'): POUNCE
```

Floor corrected to 6.10.1 in the extra, both docstrings, the `ImportError` text, the `__init__` comment, the skip reason, README, CHANGELOG and the docs page.

**On the test-file ordering** I went the other way from your suggestion, and want to flag it rather than bury it. Wrapping `import pyomo_pounce` in a try/skip would swallow exactly the regression that broke the legacy plugin — a raising import is never an "old Pyomo, nothing to test here" condition. So the import stays plain and loud, `HAVE_V2_INTERFACE` remains the only thing that may legitimately be False, and the guard is exercised by the CI leg instead.

That leg is the substantive part of your "the floor is a claim, not a tested property": a separate venv on **6.10.0** (last release below the floor, and the one where this broke) asserting the import works, the flag is `False`, the legacy factory resolves, and `test_v2.py` skips rather than erroring at collection. Verified locally — 1 skipped, not 1 error.

## 2. `pounce-solver` floor — confirmed

Extra is now `["pyomo>=6.10.1", "pounce-solver>0.9.0"]`. `>0.9.0` rather than a named version deliberately: 0.9.0 is the latest on PyPI and #553 is unreleased, so the constraint should be "newer than the released 0.9.0" whatever the next number turns out to be — no edit needed at release time. Comment in `pyproject.toml` records why it is on the extra and not the base dependency, and that CI builds from source and so structurally cannot catch it.

Your caveat about the 102 pre-existing failures from version skew is noted and I did not treat them as evidence about the PR.

## 3. Status divergence — confirmed, reproduced, fixed

```
ordinary: tc=iterationLimit  ss=unknown  x=1.090226
sens    : RAISED NoSolutionError
```

Same model, same options, `raise_exception_on_nonoptimal_result=False`, default `load_solutions`.

The rule I had missed is `asl_sol_reader.py:292`:

```python
status = SolutionStatus.unknown if sol_data.primals else SolutionStatus.noSolution
```

— overridden only for the optimal/feasible/infeasible bands. So it is "did a primal vector come back", and the in-process route always has one. Limit, time-limit, divergence and user-stop now map to `unknown`. Termination conditions keep their POUNCE-derived values as you suggested, and the table now carries the reasoning for why the two columns follow different rules.

`incumbent_objective` gating moved to `{feasible, optimal}` to match `ipopt.py:418-419` — same root cause, and it showed up in the repro above as the ordinary route reporting `None`.

New test solves a limit-capped model both ways with default `load_solutions` and asserts equal `solution_status`, which is the case the two existing limit tests structurally could not see.

## The unresolved-name check — implemented, but not as specified

Comparing `len(out)` against `capture["n"]` would warn on **every** declared-sens-param model. The solve runs on a surgery clone whose `.col`/`.row` legitimately name components that exist only there:

```
n (solve vars): 3
var_names: ['x', 'y', '_SENSITIVITY_TOOLBOX_DATA.p']
resolved on original model: 2        cons resolved: 1 of m = 2
```

Both shortfalls are by construction — `sens_solve`'s own load-back skips them the same way — so a count check would fire on exactly the models this route exists for. The check now excludes names under the sensitivity-toolbox block, which leaves only genuine mismatches. Two tests: one that it stays quiet on a sens model, one that it still fires for a spliced-in name the surgery cannot explain, so the first is not vacuous.

`get_solution_ids` overridden; comment added on the discarded config derivation explaining why it is safe and why reusing it would mean copying `Ipopt.solve`'s body.

Suite is **277 passed**. `PounceSensSolutionLoader` and the reduced-cost arithmetic are otherwise unchanged, since you diffed those against 6.10.1 directly.

---

PR #563, opened 2026-08-09
perf(feral): cache the triplet → CSC permutation instead of rebuilding on every factor()

### LINKED ISSUE #562: perf(feral): pounce rebuilds the CSC from triplets on every factor() call (2.7 ms clnlbeam, 16.5 ms dtoc2)
`pounce-feral` rebuilds the CSC matrix from triplets on **every** `factor()`
call, even though the sparsity pattern is fixed by `initialize_structure` and
only the values change between IPM iterations. It costs 2.7 ms per
factorization on `clnlbeam` and 16.5 ms on `dtoc2`.

This is a pounce-side cost that sits *inside* the FERAL column of the
factorization table in #552, so part of the gap that issue attributes to FERAL
is ours.

## Where

`crates/pounce-feral/src/lib.rs:593`, in `FeralSolverInterface::factor`:

```rust
let matrix = match CscMatrix::from_triplets(n, &self.rows_0, &self.cols_0, &self.values) {
    Ok(m) => m,
    Err(_) => return ESymSolverStatus::FatalError,
};
```

`rows_0` and `cols_0` are written once by `initialize_structure` and never
change; only `values` is refilled by the caller through `values_array_mut()`.
So every factorization redoes an identical bucket-count / place / sort /
dedup-by-summing pass to arrive at the same `col_ptr` and `row_idx` it
computed last time.

The rebuild is not cheap by construction. `from_triplets` allocates
`col_counts`, `col_ptr`, `row_idx`, `values` and `offsets`, then calls
`sort_and_sum_duplicates`, which allocates a fresh `Vec<(usize, f64)>` **per
column** and sorts it, and then allocates three more output arrays. On
`clnlbeam` (n = 99,999) that is ~100k short-lived heap allocations plus ~100k
sorts per factorization, to reproduce a permutation that is already known.

MA57 has no equivalent step. `Ma57SolverInterface` hands the caller-filled
values array straight to `ma57bd_` (`crates/pounce-hsl/src/ma57.rs:345`,
`self.a.as_ptr()`) — the triplet is consumed in place, and the analysis
products from `ma57ad_` are reused. So this is pure overhead in the FERAL arm
of every A/B we have run.

## Measured

Timing a repeated `CscMatrix::from_triplets` on the same KKT triplet dumped
via `POUNCE_DBG_KKT_DUMP` (iteration 3), against the `factor+solve` medians
already reported in #552 and in
[feral#148](https://github.com/jkitchin/feral/issues/148). Apple M-series,
refinement off.

| matrix | dim | nnz | `from_triplets` | FERAL as measured in #552 | share | best-config FERAL | share | MA57 total |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| clnlbeam | 99,999 | 299,993 | **2.68 ms** | 57.80 ms | 4.6% | 29.41 ms | 9.1% | 7.16 ms |
| dtoc1nd | 9,685 | 223,230 | 1.43 | 16.84 | 8.5% | 15.60 | 9.2% | 3.42 |
| steering_12800 | 115,201 | 422,392 | 3.57 | 47.34 | 7.5% | 33.56 | 10.6% | 18.93 |
| rocket_12800 | 89,601 | 473,589 | 3.27 | 24.70 | 13.2% | 22.24 | 14.7% | 8.88 |
| marine_1600 | 76,807 | 414,583 | 3.71 | 39.38 | 9.4% | 36.07 | 10.3% | 13.37 |
| dtoc2 | 103,920 | 2,051,790 | **16.46** | 139.84 | 11.8% | 121.08 | 13.6% | 84.26 |

("best-config FERAL" is the fastest thread count / driver from feral#148 —
what FERAL is worth once its own configuration problems are set aside. The
rebuild's share of the total *grows* as FERAL's other overheads are removed.)

The blunt framing: on `clnlbeam`, pounce burns 2.68 ms rearranging the matrix
before FERAL is even called — 37% of the time MA57 takes to do the entire
factorization *and* solve. Same story on `rocket_12800` (37%) and `dtoc1nd`
(42%).

## Fix

Cache the triplet→CSC slot permutation once and scatter values through it.
The structure is already available: `self.matrix` is retained across calls
(for `backsolve`'s iterative refinement), and `feral::CscMatrix` has all four
fields public (`n`, `col_ptr`, `row_idx`, `values`), so no feral API change is
needed.

- On the first `factor()` after an `initialize_structure`, build the matrix
  the way we do today, and alongside it record `slot: Vec<usize>` mapping each
  triplet index `k` to its position in `matrix.values`.
- On subsequent calls, reuse the retained `matrix`: `values.fill(0.0)` then
  `for k in 0..nnz { values[slot[k]] += self.values[k] }`. The `+=` preserves
  `from_triplets`'s duplicate-summing semantics when several triplets land in
  the same slot.
- Invalidate `slot` (and `self.matrix`) in `initialize_structure`, which is
  the only place the pattern can change.

That turns an allocate-sort-dedup pass into one O(nnz) gather with no
allocation. It should recover most of the column above; `dtoc2` in particular
would go from 16.5 ms of rearrangement to a couple of ms of scatter.

Two things to be careful about:

- **Explicit structural zeros must survive.** The comment at
  `lib.rs:585-592` is load-bearing — where a multiplier is zero the (2,2)
  diagonal arrives as a structurally-present exact `0.0`, and dropping it
  triggers the feral#46 cascade. A permutation-based refill keeps every slot
  that the first build created, so this is preserved by construction, but it
  deserves a test that asserts `matrix.nnz()` and `row_idx` are bit-identical
  to a fresh `from_triplets` across a values refill that zeroes a diagonal.
- **Verify the pattern really is immutable.** The permutation is only valid if
  no path mutates `rows_0`/`cols_0` without going through
  `initialize_structure`. Worth a debug assertion rather than an assumption.

A cheap correctness net: under `debug_assertions`, rebuild with
`from_triplets` and assert the refilled values match.

## Relationship to #552

This does not overturn #552 — the gap is still overwhelmingly FERAL's, and
feral#148 has the breakdown. But the FERAL column in #552's table is
`multi_solve(new_matrix=true)`, which includes this rebuild, so the ratios
there are 5–15% too generous to MA57's side of the comparison. Fixing this is
also the only part of that gap that is entirely within pounce's control, and
it is a small, self-contained change.

---

jkitchin:
Fixes #562

## Problem

`pounce-feral` rebuilt the `CscMatrix` from triplets on **every** `factor()`
call, even though `initialize_structure` fixes the sparsity pattern and only
the values change between IPM iterations. This is pounce-side cost sitting
*inside* the FERAL column of #552's factorization table.

Isolated cost of the rearrangement, timed on a clnlbeam-shaped KKT
(n = 99,999, nnz = 299,994) in a release build. Medians of 5, this box (x86,
not the M-series of the issue's table — the ratio is the transferable number):

| | per factorization |
|---|---:|
| before — `CscMatrix::from_triplets` | 4.45 ms |
| after — slot scatter | 0.34 ms |

13× on that step. #562's own measurements put the rebuild at 4.6–13.2% of the
FERAL `factor+solve` median as measured in #552, rising to 9.1–14.7% against
feral#148's best-config FERAL, and at 16.5 ms per factorization on `dtoc2`.
There is no oracle to compare against here: the factored matrix is unchanged,
so the solver's numerical trajectory is not what this PR moves.

## Cause

`rows_0` / `cols_0` are written once by `initialize_structure` and never
change; only `values` is refilled by the caller through `values_array_mut()`.
So every factorization redid an identical bucket-count / place / sort /
dedup-by-summing pass to arrive at the same `col_ptr` and `row_idx` it had
computed the iteration before.

The rebuild is not cheap by construction. `from_triplets` allocates
`col_counts`, `col_ptr`, `row_idx`, `values` and `offsets`, then calls
`sort_and_sum_duplicates`, which allocates a fresh `Vec<(usize, f64)>` **per
column** and sorts it, then allocates three more output arrays. On a
100k-column KKT that is ~100k short-lived heap allocations plus ~100k sorts
per factorization, to reproduce a permutation that was already known.

MA57 has no equivalent step — `Ma57SolverInterface` hands the caller-filled
values array straight to `ma57bd_` and the triplet is consumed in place — so
this was pure overhead in the FERAL arm of every A/B we have run.

## Fix

Cache the triplet → CSC slot permutation once and scatter values through it.

- The first `factor()` after an `initialize_structure` builds the matrix
  exactly as before, and now also records `slot: Vec<usize>` mapping each
  triplet `k` to its index in `matrix.values`. `slot_map` finds each slot by
  binary search inside the triplet's column range, which is valid because
  `from_triplets` leaves each column's `row_idx` sorted and duplicate-free.
- Later calls reuse the retained matrix: `values.fill(0.0)`, then
  `values[slot[k]] += self.values[k]`. The `+=` (rather than a store)
  reproduces `from_triplets`'s duplicate-summing when several triplets land in
  one slot; the `fill` is what makes that `+=` safe to repeat.
- `initialize_structure` — the only place the pattern can change — invalidates
  both `slot` and `matrix`.

`self.matrix` was already retained across calls for `backsolve`'s iterative
refinement, and `feral::CscMatrix` has all four fields public, so this needs no
feral API change. The matrix now moves into `self.matrix` *before* the
factorization rather than after; the observable behaviour is the same (it was
stored unconditionally either way), and it keeps the retained matrix consistent
with the values just supplied even when the factor fails.

**On the two hazards named in the issue.**

*Explicit structural zeros survive.* The refill writes only into slots the
first build created, so a (2,2) diagonal that goes to an exact `0.0` when its
multiplier vanishes keeps its stored entry — the property whose loss is the
structurally-absent-diagonal cascade (feral#46) that the "do not strip zeros"
comment exists to protect. This is preserved by construction, and pinned by a
test that compares `nnz`, `col_ptr` and `row_idx` bit-for-bit against a fresh
`from_triplets` across a refill that zeroes a diagonal.

*Pattern immutability is asserted, not assumed.* Under `debug_assertions` every
refill is cross-checked against a fresh `from_triplets`. A mutation of
`rows_0` / `cols_0` behind `initialize_structure`'s back surfaces there as a
`col_ptr` / `row_idx` mismatch rather than as silently wrong arithmetic.

**Rejected: comparing the debug cross-check's values bit-wise.** Where
duplicate triplets are summed, the rebuild adds them in row-sorted order and
the scatter adds them in triplet order, and floating-point addition is not
associative — a bit-wise value check would fire spuriously on any KKT with
duplicate entries. Structure is compared exactly (that is the part the
permutation asserts); values to a relative tolerance.

**Rejected: making a `slot_map` miss fatal.** It returns `Option` and a `None`
leaves the cache unpopulated, so the next `factor()` rebuilds from scratch. A
miss would contradict `from_triplets`'s own postcondition, so it should not
happen — but if an upstream change ever makes it happen, degrading to the old
behaviour beats aborting the solve or, worse, factoring a wrong matrix.

## Blast radius

`FeralSolverInterface::factor` only. The matrix handed to
`feral::Solver::factor` is identical to what the old code built (bit-identical
structure; values identical except for the summation order of duplicate
triplets sharing a slot), so the IPM's numerical trajectory is unchanged and
this is a pure wall-clock change.

Not touched: `schur.rs` has its own `from_triplets` calls on the Schur-complement
path. They are outside this issue's scope and are left as they are.

This does not overturn #552 — the gap there is still overwhelmingly FERAL's,
and feral#148 has the breakdown. It does mean #552's FERAL column
(`multi_solve(new_matrix=true)`, which included this rebuild) was 5–15% too
generous to MA57's side of the comparison.

## Tests

Five new tests in `crates/pounce-feral/src/lib.rs`, over a 3×3 pattern chosen
to hit every branch `from_triplets` takes — an upper-triangle entry that
`initialize_structure` canonicalizes, rows arriving out of order within a
column, and a duplicate pair that must be summed — plus a (3,3) diagonal
standing in for the KKT's (2,2) multiplier block:

- `slot_map_matches_from_triplets_layout` — the permutation against
  `from_triplets`'s actual `col_ptr` / `row_idx`, and replaying it reproduces
  the built values exactly.
- `refill_preserves_structure_when_a_diagonal_goes_to_zero` — `nnz`,
  `col_ptr`, `row_idx` bit-identical to a fresh rebuild across a refill that
  drives the diagonal to an exact zero, and that entry still stored.
- `refill_solves_the_new_matrix_not_the_old_one` — residual check on the
  second solve. Zeroing the diagonal flips the matrix to indefinite, so a
  transposed or stale map cannot pass by accident.
- `re_initializing_the_structure_invalidates_the_cached_permutation` — re-init
  with a different pattern *and* dimension, then solve.
- `repeated_refills_do_not_accumulate` — four factorizations in a row, values
  compared against a rebuild each time.

**How I know they bite.** These do not fail on the parent commit in the usual
sense — this is a perf change, not a bug fix, and the assertions on `slot` /
`matrix` do not compile against the parent, where those fields do not exist.
So I checked them by mutation instead, against the final diff:

| mutation | result |
|---|---|
| `values[s] = v` instead of `+=` (breaks duplicate summing) | 3 tests fail |
| drop `values.fill(0.0)` (values accumulate across factorizations) | 3 tests fail |
| drop the invalidation in `initialize_structure` | 1 test fails |

Each failure is for the stated reason — the first two report the wrong value at
a named CSC slot, the third reports the stale permutation surviving re-init.

The debug cross-check also runs on every refill in every debug-profile test, so
the whole suite exercises it, not just these five.

Not pinned: the timing itself. The measurement above came from a throwaway
release-mode bench, deliberately not committed — a wall-clock assertion in the
test suite would be flaky on CI runners, and the structural and numerical
properties are what a regression here would actually break.

---

- [x] Tests fail on the parent commit for the stated reason — **N/A, see
      "How I know they bite" above**: verified by mutation instead, because a
      perf change has no pre-fix failure and the new assertions do not compile
      against the parent
- [ ] Book page under `docs/src/` updated — no user-facing page describes this
      internal path; `crates/pounce-feral/README.md` and the module docs, which
      did describe the old per-factor rebuild, are updated
- [x] `cargo fmt --all -- --check`, `cargo clippy --workspace --exclude
      pounce-hsl --all-targets -- -D clippy::correctness -D clippy::suspicious`,
      and `cargo test --workspace --exclude pounce-hsl` (60 suites) all clean
      as it stands now

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mrgp9sfhDyChdpS8qxzVjd)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,41 @@ changes.
 
 ## [Unreleased]
 
+### Changed — the FERAL backend stops rebuilding the CSC matrix on every factorization (#562)
+
+`pounce-feral` handed FERAL a `CscMatrix` built by
+`CscMatrix::from_triplets` on **every** `factor()` call, even though
+`initialize_structure` fixes the sparsity pattern and only the values
+change between IPM iterations. Each rebuild re-ran an identical
+bucket-count / place / sort / sum-duplicates pass to arrive at the same
+`col_ptr` and `row_idx` it had computed the iteration before — and that
+pass allocates a fresh `Vec<(usize, f64)>` *per column* and sorts it, so
+on a 100k-column KKT it is ~100k short-lived allocations and ~100k sorts
+to reproduce a permutation that was already known. MA57 has no equivalent
+step (its triplet is consumed in place by `ma57bd_`), so this was pure
+overhead in the FERAL arm of every A/B comparison.
+
+The first factorization after an `initialize_structure` still goes
+through `from_triplets`, and now also records the triplet → CSC slot
+permutation. Later factorizations zero the retained matrix's values and
+scatter the caller's triplets through that permutation — one O(nnz) pass
+with no allocation. The scatter accumulates with `+=` so duplicate
+triplets landing in one slot are still summed, and because it reuses the
+slots the first build created, the explicit structural zeros on the
+(2,2) multiplier diagonal survive a refill (dropping them is the
+FERAL-side cascade that the "do not strip zeros" rule exists to avoid).
+`initialize_structure` — the only place the pattern can change —
+invalidates the permutation and the matrix.
+
+On a clnlbeam-shaped KKT (n = 99,999, nnz = 299,994) the rearrangement
+drops from ~4.4 ms to ~0.34 ms per factorization, a 13× reduction, and
+the saving grows with the matrix: the issue measures 16.5 ms of rebuild
+per factorization on `dtoc2`. Under `debug_assertions` each refill is
+cross-checked against a fresh `from_triplets` — exact on `col_ptr` and
+`row_idx`, to a relative tolerance on the values — which doubles as the
+assertion that nothing mutates the pattern behind
+`initialize_structure`'s back.
+
 ### Fixed — a single dense Hessian row made `.nl` setup and every `eval_h` O(n²) (#552)
 
 `NlTnlp` recovers the Lagrangian Hessian by graph coloring: columns whose
--- a/crates/pounce-feral/README.md
+++ b/crates/pounce-feral/README.md
@@ -19,11 +19,15 @@ The wrapper mirrors `pounce-hsl::Ma57SolverInterface`:
    assembly emits are canonicalized to the lower triangle here, because
    FERAL's `CscMatrix::from_triplets` silently drops upper-triangle
    entries during LDLᵀ.
-3. `multi_solve` rebuilds the CSC matrix from the cached pattern +
+3. `multi_solve` refreshes the CSC matrix from the cached pattern +
    caller-filled values, then dispatches to `feral::Solver::factor` /
-   `solve_many`. FERAL's pattern-fingerprint cache reuses the symbolic
-   factorization across iterates with identical structure — the IPM
-   common case.
+   `solve_many`. The CSC is built by `from_triplets` once per pattern;
+   later factorizations scatter the new values through a cached
+   triplet → slot permutation, which skips the per-column allocate /
+   sort / sum-duplicates pass that would recompute the same `col_ptr`
+   and `row_idx` every iteration (gh#562). FERAL's pattern-fingerprint
+   cache likewise reuses the symbolic factorization across iterates
+   with identical structure — the IPM common case.
 4. `increase_quality` delegates to `feral::Solver::increase_quality`
    and uses MA57's `pivtol_changed` / `CallAgain` protocol so the
    upper-layer reload-and-retry semantics line up.
--- a/crates/pounce-feral/src/lib.rs
+++ b/crates/pounce-feral/src/lib.rs
@@ -9,11 +9,13 @@
 //! * `initialize_structure` caches the 0-based row/col arrays needed by
 //!   FERAL's [`CscMatrix::from_triplets`] and allocates the values
 //!   buffer.
-//! * `multi_solve` rebuilds `CscMatrix` from the cached pattern + caller-
-//!   filled values and dispatches to [`feral::Solver::factor`] /
-//!   [`feral::Solver::solve_many`]. FERAL's pattern-fingerprint cache
-//!   reuses the symbolic factorization across iterates with identical
-//!   structure (the IPM common case).
+//! * `multi_solve` refreshes the `CscMatrix` from the cached pattern +
+//!   caller-filled values and dispatches to [`feral::Solver::factor`] /
+//!   [`feral::Solver::solve_many`]. The CSC is built once per pattern; later
+//!   factorizations scatter the new values through a cached triplet → slot
+//!   permutation instead of re-running `from_triplets` (gh#562). FERAL's
+//!   pattern-fingerprint cache likewise reuses the symbolic factorization
+//!   across iterates with identical structure (the IPM common case).
 //! * `increase_quality` delegates to [`feral::Solver::increase_quality`]
 //!   and uses MA57's `pivtol_changed` / `CallAgain` protocol so the
 //!   upper-layer reload-and-retry semantics line up.
@@ -58,8 +60,24 @@ pub struct FeralSolverInterface {
     values: Vec<Number>,
 
     /// Last factored matrix, retained so `backsolve` can run iterative
-    /// refinement against it (feral's `solve_*_refined` requires `A`).
+    /// refinement against it (feral's `solve_*_refined` requires `A`), and
+    /// re-used as the destination of the values refill (see [`Self::slot`]).
     matrix: Option<CscMatrix>,
+
+    /// Triplet → CSC slot permutation: `slot[k]` is the index into
+    /// `matrix.values` that triplet `k` of `(rows_0, cols_0)` lands in.
+    ///
+    /// The sparsity pattern is fixed by `initialize_structure` and only the
+    /// values change between IPM iterations, so the bucket-count / place /
+    /// sort / sum-duplicates pass inside `CscMatrix::from_triplets`
+    /// reproduces the same `col_ptr` and `row_idx` on every factorization —
+    /// at the cost of one `Vec<(usize, f64)>` allocation and one sort *per
+    /// column* (gh#562: 2.7 ms per factor on clnlbeam, 16.5 ms on dtoc2).
+    /// Recorded on the first `factor()` after an `initialize_structure` and
+    /// replayed as an allocation-free O(nnz) scatter thereafter; `None`
+    /// until then, and reset to `None` by `initialize_structure` — the only
+    /// place the pattern can change.
+    slot: Option<Vec<usize>>,
 
     negevals: Index,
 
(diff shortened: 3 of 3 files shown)


jkitchin:
Reviewed, and verified independently rather than from the body. This is a
clean fix for #562. I also resolved the CHANGELOG conflict and pushed it (see
the bottom).

## Verified end to end

The body's claim is a wall-clock change with an unchanged numerical
trajectory. I built release binaries from `origin/main` and this branch and ran
both on real Mittelmann models, three reps each, Apple M-series:

| model | main | this PR | Δ | iters | objective |
|---|---:|---:|---:|---:|---|
| clnlbeam | 7.70 / 7.76 / 7.68 s | 7.49 / 7.40 / 7.36 s | **−3.9%** | 60 = 60 | `3.4517440170910248e+02` both |
| dtoc2 | 3.64 / 3.64 / 3.59 | 3.48 / 3.46 / 3.49 | **−4.3%** | 5 = 5 | `2.8632990924502875e+00` both |

Objectives are bit-identical to all 17 digits and iteration counts match, so
the trajectory really is unchanged — which is the load-bearing claim, since
this touches the matrix every factorization consumes.

The `dtoc2` saving is the detail I found most convincing. Median drops 0.16 s
over a 5-iteration solve. #562 measured the rebuild at 16.46 ms per
factorization, and `dtoc2` takes ~10 factorizations — 0.16 s predicted. The
end-to-end saving matches the isolated microbenchmark almost exactly, which is
a much stronger result than either number alone.

`cargo test -p pounce-feral` is 36 passed on this box, and since that runs in
the debug profile, the `assert_refill_matches_rebuild` cross-check fires on
every refill in every one of them.

## On the design

The three things I flagged as hazards in #562 are all handled, and two are
handled better than I suggested:

- **`slot_map` by binary search inside the column range** is the right
  construction — it derives the permutation from `from_triplets`'s actual
  output rather than reimplementing its placement logic, so it cannot drift
  from feral's behaviour. Returning `Option` and degrading to a rebuild on a
  miss is the correct failure mode.
- **Comparing the debug cross-check's values with a tolerance rather than
  bit-wise** is a catch I did not make. The rebuild sums duplicates in
  row-sorted order and the scatter sums them in triplet order; a bit-wise
  check would have fired spuriously on any KKT with duplicate triplets. Worth
  noting the new path is arguably *more* deterministic than the old one —
  `sort_and_sum_duplicates` uses `sort_unstable_by_key`, so the old
  duplicate-summation order was unspecified in principle.
- **Moving the matrix into `self.matrix` before the factorization** is
  behaviour-preserving as claimed: the old code stored it unconditionally
  after `factor` regardless of outcome, and nothing inside
  `feral::Solver::factor` reads `self.matrix`. The comment explaining why is
  in the right place.

Scoping out `schur.rs`'s own `from_triplets` calls is the right call for this
PR; if that path turns out to be hot it deserves its own measurement first.

The "how I know they bite" mutation table is the right way to evidence a perf
change that has no pre-fix failure, and the three mutations chosen are exactly
the three ways this could silently break (`=` for `+=`, missing `fill`,
missing invalidation).

## No blocking findings

Two observations, neither worth holding the PR for:

- `debug_assert_eq!(slot.len(), self.values.len())` and
  `debug_assert_eq!(matrix.n, n)` are no-ops in release. They can only fail if
  something changed `nnz`/`dim` without going through `initialize_structure`,
  which would also trip the cross-check right below — so the coverage is
  there, just not in release. Fine as is.
- The duplicate-summation reordering means values can differ in the last ulp
  from the old code on a KKT with duplicate triplets. The body states this. My
  A/B above shows it does not perturb these two models at all (bit-identical
  objectives), but that is evidence, not proof, for models with duplicates.

## Merge conflict resolved

#557 merged to `main` about 20 minutes before this PR was pushed, and both
added an entry to the same `[Unreleased]` section, so the branch went
`CONFLICTING`. Only `CHANGELOG.md` conflicted — no code. I merged `main` in and
kept both entries, newest first (`ba24c83f`), then re-ran `cargo fmt --all --
--check` (clean), `cargo test -p pounce-feral` (36 passed) and `cargo test -p
pounce-nl` (154 passed) on the merged tree before pushing.

---

Issue #452, opened 2026-08-02
Published Linux wheels bundle a glibc-2.39 CLI under a manylinux2014 tag


jkitchin:
Discovered while building the container images for #449.

The published Linux wheels for `pounce-solver` are tagged `manylinux2014`
(glibc 2.17) but the `pounce` CLI binary they bundle requires **glibc 2.39**.
On anything older, the CLI cannot start:

```
$ docker run --rm python:3.12-slim-bookworm sh -c 'pip -q install pounce-solver==0.9.0 && pounce --version'
/usr/local/lib/python3.12/site-packages/pounce/bin/pounce: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by .../pounce/bin/pounce)
```

Reproduced on 0.9.0, **both amd64 and arm64**, on Debian bookworm (glibc 2.36).
Debian trixie (2.41) works.

## Cause

`.github/workflows/release-pounce.yml` builds the bundled CLI on the runner
host and only then hands the tree to maturin:

```yaml
- name: Build pounce CLI binary
  run: cargo build --release --bin pounce --target ${{ matrix.rust_target }}   # <- runs on the HOST

- uses: PyO3/maturin-action@v1
  with:
    manylinux: manylinux2014                                                    # <- container, CLI never enters it
```

So the extension module (`_pounce.so`) genuinely is manylinux2014, but the
CLI beside it is linked against the runner's glibc. `ubuntu-latest` and
`ubuntu-24.04-arm` are both Ubuntu 24.04 / glibc 2.39, which is why both
Linux wheels are affected. auditwheel never inspects `pounce/bin/pounce` —
it is data as far as the wheel is concerned — so nothing catches it.

## Impact

`import pounce` is fine; anything that runs the CLI is not. That includes
the `pounce` console script and **pyomo-pounce**, which shells out to the
binary — so `SolverFactory('pounce').solve(m)` fails on:

- Debian 12, Ubuntu 22.04 and older
- RHEL/Rocky/Alma 8 and 9
- most HPC cluster images, which is where this matters most

The failure is at exec time with a linker error, not at install time, so
`pip install` reports success.

## Possible fixes

1. Build the CLI *inside* the manylinux container — e.g. give
   `maturin-action` a `before-script-linux` that runs the `cargo build`, so
   the binary and the extension share a glibc floor.
2. Build the CLI for `*-unknown-linux-musl` and bundle that: fully static, no
[...]

PR #574, opened 2026-08-10
release: 0.10.0

jkitchin:
Version bump for the 0.10.0 release. No code change — every hunk is a version
string, plus the CHANGELOG cut.

0.10.0 covers 553 commits and 129 merged PRs since `v0.9.0` (2026-07-24). The
CHANGELOG section is 4,903 lines, about three times the size of 0.9.0's.

## Surfaces bumped

- Root `Cargo.toml`: `[workspace.package]` and all 17 `[workspace.dependencies]`
  pins that point at our own crates (they must match, or `cargo publish` resolves
  a dependent against the previous release).
- `python/pyproject.toml`, `pyomo-pounce/pyproject.toml`.
- `docker/Dockerfile.release`: the pinned `ARG POUNCE_VERSION` and the usage
  example above it.
- `Cargo.lock`.
- The `ghcr.io/jkitchin/pounce:0.9.0` tags quoted in `README.md`,
  `docs/src/docker.md`, and `docker/README.md` — the consistency script's
  advisory grep covers exactly these files and would have reported them stale.

`scripts/check-release-consistency.sh` passes at 0.10.0.

## CITATION.cff was two releases stale

It sat at `version: 0.8.0` / `date-released: 2026-07-11` through the entire
0.9.0 cycle, so GitHub's "Cite this repository" widget has been offering 0.8.0
since July. `dev-notes/cargo-release.md` step 4 says to bump it, but
`check-release-consistency.sh` does not verify it — which is why the drift was
silent where every other surface's drift is caught. Now at 0.10.0 / 2026-08-10.

Adding CITATION.cff to that guard is worth a follow-up; a checklist item that
was already missed once will be missed again.

## What deliberately did not move

- `pounce-solver>=0.9.0` in `pyomo-pounce`'s base dependencies — a deliberately
  loose compatibility floor, documented as such in the file.
- `pounce-solver>0.9.0` in the `pyomo-v2` extra — encodes that released 0.9.0
  predates the per-model `Options` echo (#553) the v2 route needs. Bumping it
  with the release would erase a real compatibility fact.
- The `#452` history comment in `Dockerfile.release`, which is about what 0.9.0
  shipped.

## CHANGELOG link references

The reference block stopped at `[0.3.0]`, so every version from 0.4.0 up —
including `[Unreleased]` — rendered as unlinked literal text. Added the seven
missing ones.

## Release preflight, and a broken step in the docs

`scripts/publish-crates.sh --dry-run` **fails at the second crate**, and will
fail for any release that bumps the version:

```
error: failed to prepare local package for uploading
  failed to select a version for the requirement `pounce-common = "^0.10.0"`
  candidate versions found which didn't match: 0.9.0, 0.8.0, 0.7.0, ...
  location searched: crates.io index
```

`cargo publish -p <crate> --dry-run` resolves that crate's workspace siblings
against the **crates.io index**. At a bumped version none of them are there yet,
so only the leaves (`pounce-common`, `pounce-studio-core`) can pass. The
procedure in `dev-notes/cargo-release.md` claims this step "dry-runs every crate
end-to-end, so any breakage appears here, not three crates into the real
release" — that is not achievable this way. Run before the bump it passes
trivially, having validated nothing about the version being shipped; run after
the bump it fails on crate 2. Either way it never did the job described.

`cargo publish --workspace --dry-run` (cargo 1.96) resolves siblings against the
locally packaged crates instead of the index, which is the check the docs
describe. Its result is reported in a comment below. Switching the script's
`--dry-run` path to it is the natural follow-up.

The real publish is unaffected: it publishes in topological order, so each
crate's dependencies genuinely are live by the time it uploads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4899,7 +4899,7 @@ fact improved that family substantially (92 → 38 failures/144) before #361 too
 it to 0.
 
 
-### ### Tests — pyomo-pounce: `test_binary_check.py` failed on any Windows checkout (#366)
+### Tests — pyomo-pounce: `test_binary_check.py` failed on any Windows checkout (#366)
 
 - Three tests were POSIX-only: the shadowing test joined `PATH` with a
   hardcoded `":"` instead of `os.pathsep`, and the fake `pounce` binaries were
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ changes.
 
 ## [Unreleased]
 
+
+## [0.10.0] - 2026-08-10
+
 ### Changed — FERAL 0.15.1, which stops assuming the host has threads
 
 The workspace pin moves `feral` 0.15.0 → 0.15.1. It is a patch release with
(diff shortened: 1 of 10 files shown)


jkitchin:
Dry-run result, following up the preflight note in the description.

`cargo publish --workspace --dry-run` passes clean:

- exit 0
- 20 crates packaged and verified at 0.10.0
- 0 errors, 0 warnings

The set it covered diffs exactly against the `CRATES=(...)` list in
`scripts/publish-crates.sh`, and the three `publish = false` crates
(`pounce-py`, `pounce-studio-pyo3`, `iter-diff`) were excluded without needing
to be named.

So the release is validated end-to-end; the gap was only in the documented
command. For the record, the two behave differently on purpose:

| | sibling deps resolved against | works at a bumped version |
|---|---|---|
| `cargo publish -p <crate> --dry-run` (what the script runs, per crate) | crates.io index | no — fails on the first non-leaf |
| `cargo publish --workspace --dry-run` | locally packaged crates | yes |

Follow-up worth filing: point the script's `--dry-run` path at
`cargo publish --workspace --dry-run` and correct the claim in
`dev-notes/cargo-release.md`. Leaving both alone in this PR — a release bump
should not also change the release tooling.

PR #583, opened 2026-08-12
Add timelimit for convex route

GermanHeim:
## Problem

POUNCE has shared wall/CPU deadline machinery for the NLP engine, but both convex routes currently discard max_wall_time/max_cpu_time. I found this while updating oximo to pounce-rs v0.10.0.

## Fix

I added a `time_limit: Option<Duration>` and `TimeLimit` statuses to both convex engines.

Note: I didn't add a cancellation callback because it would make everything more complex, but feel free to add it if you think it is worth it.

Disclaimer: I did an initial implementation and then used ChatGPT Terra Medium to update all the relevant sites.

*— @GermanHeim (commit `afb9d2fc`)*

---

## Maintainer follow-up

Reviewing the above surfaced a latent wrong-answer bug that the change made
reachable, so rather than send a long review I pushed four commits on top
(agreed with @GermanHeim in the thread below). Two rules came out of it, and
both are now applied uniformly across the convex family.

**Cancellation is an error, not a value** (`00ee7334`)

`factorize_with_inertia_control` returned `Ok(current)` when the deadline
expired mid-shift-escalation, handing an *unsolved* KKT right-hand side back
as if it were a step. The result is `x = -g` with `delta == 0`, which skips
the step-quality guards — so the solve reported `Optimal` on a
feasible-but-wrong point. Under a time limit you could get a silently wrong
answer with a success status.

Now a typed `QpError::DeadlineExpired`, propagated by `?` and softened to
`QpStatus::TimeLimit` only at the public entry points, which let the compiler
find the remaining paths. Regression test in
`crates/pounce-qp/tests/deadline_cancellation.rs`: reverting the fix reports
`x = [2.0, 2.0]` as `Optimal` against a true optimum of `(3.2, 0.8)`.

**A verdict outranks the clock** (`00ee7334`, `ad216bbc`)

Several exits stamped `TimeLimit` over a status the solver had already earned.
A proven `Optimal` / `OptimalInaccurate` / `PrimalInfeasible` /
`DualInfeasible` now survives a deadline crossing; only a give-up status is
relabelled. Someone whose problem solves in 1.001 s under `time_limit = 1`
should not lose the optimum they already have.

Applied in the IPM, both HSDE drivers, the LP crossover, and — `ad216bbc` —
the active-set route, which had the same overwrite at its post-solve exit and
at three retry gates, so `method="active_set"` and the LP path could lose a
verdict the same way. `verify_status` also salvaged an identical KKT point to
`Optimal` when iterations ran out but discarded it when the clock did; both
now share the `mark_timed_out` policy.

**The budgets didn't compose** (`62efc742`)

`max_wall_time` bounded each *engine*, not the solve. When a convex attempt
declined and handed off to the NLP path, the NLP side built its deadline from
the option value — still naming the full budget. Spend 55 s of a 60 s budget
convex-side, get 60 more. A declined attempt is now charged against the
remaining budget before the handover.

Note for anyone touching that code: `max_wall_time` has a *strict* lower bound
of zero, so writing `0.0` for an exhausted budget is rejected as invalid and
silently leaves the full budget in place — exactly the bug being fixed. Hence
the `1e-9` floor, pinned by a test.

**Smaller items** (`82b7572d`)

`TimeLimit` was missing from the CLI's `ALL_STATUSES` retry table (a spent
budget was a reason to try again); `max_wall_time` values are validated before
becoming a `Duration`; the Python status code and message map the new status;
`batch.rs` documents that `time_limit` is *per instance*, not per batch.

**Flaky test, unrelated to this change** (`22f7781d`)

CI failed on the base commit in `test_qp_solve_releases_the_gil` — diagnosed
by @GermanHeim as a timing threshold, and correct. It asserted eight threaded
solves finish in under 0.75x the serial time, which measures whether the
runner has idle cores as much as whether the extension releases the GIL; it
failed at ratio 0.77 on a small runner.

It now observes the property directly: a daemon thread increments a counter in
a tight Python loop, and while another thread holds the GIL that counter
cannot advance at all. Comparing its rate during the solve against its rate
while the main thread sleeps separates the regimes by definition rather than
by throughput — measured 0.92-1.05 released against ~0 held, with the
threshold at 0.25. The `>= 4 core` skip is gone, so the test now runs
everywhere instead of silently opting out on small runners.

## Audited and deliberately unchanged

The NLP filter-IPM family already satisfies both rules via its own
`pounce_common::timing::Deadline` (#242/#244/#245/#246) — its KKT solver
returns `false` on an aborted factorization rather than a bogus success, which
is precisely the pattern the QP path was missing, and `opt_error.rs` checks
the clock only after the convergence verdicts. `simplex.rs` returns `None` on
every deadline exit, so it can never regress a solution.

`pounce-algorithm`'s SQP arm sets no `QpOptions::time_limit` on its inner QP
solves, so those stay bounded by iterations and by the outer `Deadline` at
iteration granularity; documented rather than wired.

## Follow-up

#585 — the Python API can *receive* the `time_limit` status but has no way to
*request* a budget, since the pyo3 layer never sets the field. That is an API
addition and wants its own PR, so it is intentionally not closed by this one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,16 @@ changes.
   spent rather than zero. `pounce_convex::batch`'s `time_limit` is documented
   as per-instance, which is what it has always been.
 
+- **`max_wall_time` now bounds the whole CLI solve, not each engine that gets
+  a turn at it.** When a convex attempt declines the problem and hands it to
+  the NLP path (the gh #535 LP→NLP reroute, or `socp_nlp_fallback`), the NLP
+  solve built its `Deadline` from the option value — which still named the
+  full budget — so a run that spent most of its seconds convex-side was
+  granted them all again and the cap could buy nearly twice the wall clock it
+  promises. The declined attempt is now charged against the option before the
+  handover, the same deduction the convex path already applies internally for
+  extraction and presolve.
+
 
 ## [0.10.0] - 2026-08-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,8 +40,17 @@ changes.
     driver's reduced-accuracy salvage, so a cancelled solve whose last iterate
     satisfies the KKT conditions is still allowed to say so.
 
+  Both rules apply on the LP / convex-QP **active-set** route as well as the
+  IPM one: `solve_qp_active_set` shares the IPM's relabelling policy rather
+  than stamping `TimeLimit` over whatever the inner solve returned, and a
+  cancelled engine result now goes through the same verified KKT check as an
+  iteration-limited one — reaching the answer and then running out of seconds
+  used to be reported as reaching nothing, while running out of *iterations*
+  at the identical point reported `Optimal`.
+
   A cancelled active-set solve also reports the wall-clock time it actually
-  spent rather than zero.
+  spent rather than zero. `pounce_convex::batch`'s `time_limit` is documented
+  as per-instance, which is what it has always been.
 
 
 ## [0.10.0] - 2026-08-11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ changes.
 
 ## [Unreleased]
 
+- Added solve-wide convex-engine wall-clock deadlines. Both
+  `pounce_convex::QpOptions` and `pounce_qp::QpOptions` now expose
+  `time_limit: Option<Duration>` and their status enums expose `TimeLimit`.
+  Automatic LP/QP/active-set/SOCP routing forwards an explicitly set
+  `max_wall_time`, shares it across setup and retry stages, reports AMPL result
+  400 / `MaximumWallTimeExceeded`, and never reroutes a timed-out solve to NLP.
+  This is source-breaking for exhaustive option literals and enum matches.
+
 
 ## [0.10.0] - 2026-08-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,32 @@ changes.
   400 / `MaximumWallTimeExceeded`, and never reroutes a timed-out solve to NLP.
   This is source-breaking for exhaustive option literals and enum matches.
 
+  A deadline changes when a solve stops, never what it answers. Two rules
+  carry that:
+
+  * **Cancellation is an error, not a value.** In `pounce-qp` it travels as
+    the internal `QpError::DeadlineExpired`, so `?` propagation forces every
+    caller to handle it, and only the `QpSolver` entry points turn it back
+    into the soft `QpStatus::TimeLimit`. Without this,
+    `factorize_with_inertia_control` — whose "success" is a right-hand side
+    solved in place — could return `Ok` on a timeout, and
+    `solve_equality_only` would read the *un-solved* `[-g; b]` back as
+    `[x*; λ*]`, skip the guards a zero shift makes moot, and report a
+    feasible-but-wrong point as `Optimal`. Covered by a regression test that
+    reproduces the window deterministically.
+  * **A verdict outranks the clock.** Every deadline check in `pounce-convex`
+    runs after some inner solve has already returned, so the crossing can land
+    between convergence and the check. `Optimal`, `OptimalInaccurate`, and the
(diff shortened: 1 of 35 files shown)


GermanHeim:
I believe that the CI fails because of an inconsistent timing threshold.

jkitchin:
Thanks for this, @GermanHeim — the deadline plumbing and the CLI wiring are the
right shape, and reviewing it turned up a latent correctness bug that your
change made reachable. Rather than send you a long review to act on, I've
pushed four commits on top of your branch (your commit is untouched at the
base). Walking through what they do and why:

**1. `00ee7334` — cancellation is an error, not a value.**

The bug: `factorize_with_inertia_control` returned `Ok(current)` when the
deadline expired mid-shift-escalation. That hands an *unsolved* KKT
right-hand side back to the caller as if it were a step. The result is
`x = -g` with `delta == 0`, which skips the step-quality guards, so the solve
reports `Optimal` on a feasible-but-wrong point. Under a time limit you could
get a silently wrong answer with a success status — worse than hanging.

Fix: a typed `QpError::DeadlineExpired`, propagated by `?`, softened to
`QpStatus::TimeLimit` only at the three public entry points. The compiler then
found the rest of the paths for us. There's a regression test in
`crates/pounce-qp/tests/deadline_cancellation.rs` with a stub linear solver
that stalls past the budget; reverting the fix makes it report `x = [2.0, 2.0]`
as `Optimal` against a true optimum of `(3.2, 0.8)`.

`QpStats::time` now reports elapsed scope time rather than zero on that path.

**2. A verdict outranks the clock.**

The other half: several exits stamped `TimeLimit` over a status the solver had
already earned. If the deadline passes while a solve is finishing, a proven
`Optimal` / `OptimalInaccurate` / `PrimalInfeasible` / `DualInfeasible` should
survive — only a give-up status gets relabelled. Someone whose problem solves
in 1.001 s under `time_limit = 1` should not lose the optimum they already
have. That's `mark_timed_out` in `ipm.rs`, applied across the HSDE drivers
(symmetric and non-symmetric) and the LP crossover.

**3. `ad216bbc` — the same two rules on the active-set / LP route.**

The IPM isn't the only consumer. `active_set.rs` had the same
verdict-overwrite at its post-solve exit and at three retry gates, so
`method="active_set"` and the LP path could lose a verdict the same way.
`verify_status` also had an asymmetry worth flagging: an identical KKT point
was salvaged to `Optimal` when iterations ran out but discarded when the clock
did. Both now share the `mark_timed_out` policy.

**4. `62efc742` — the budgets didn't compose.**

`max_wall_time` bounded each *engine*, not the solve. When a convex attempt
declines and hands off to the NLP path, the NLP side built its deadline from
the option value — still naming the full budget. Spend 55 s of a 60 s budget
convex-side, get 60 more. A declined attempt is now charged against the
remaining budget before the handover.

One trap there worth knowing about: `max_wall_time` is registered with a
*strict* lower bound of zero, so writing `0.0` for an exhausted budget is
rejected as invalid and silently leaves the full budget in place — exactly the
bug being fixed. Hence the `1e-9` floor, pinned by a test.

**Also picked up:** `TimeLimit` was missing from the CLI's `ALL_STATUSES`
retry table (a spent budget was a reason to try again); `max_wall_time` values
are now validated before becoming a `Duration`; the Python status code and
message map the new status; `batch.rs` documents that `time_limit` is *per
instance*, not per batch.

**Audited and left alone:** the NLP filter-IPM family already satisfies both
rules via its own `pounce_common::timing::Deadline` — its KKT solver returns
`false` on an aborted factorization rather than a bogus success, which is
precisely the pattern the QP path was missing. `simplex.rs` returns `None` on
every deadline exit, so it can never regress a solution.

Tests: 111 test binaries across `pounce-cli`, `pounce-convex` and `pounce-qp`,
all passing; clippy clean.

Filed #585 as follow-up: the Python API can *receive* the `time_limit` status
but has no way to *request* a budget, since the pyo3 layer never sets the
field. That's an API addition and wants its own PR, not this one.

Happy to move any of this into a separate PR if you'd rather keep yours
focused — just say so. Otherwise I'll merge this as one feature.

GermanHeim:
You can keep it here, if you want. Also, stacked PRs would work, but whatever is fine.

PR #584, opened 2026-08-13
sens: report what the estimate's linear step does about the bounds

devin-griff:
This implements item 0 of the sensitivity roadmap (#255), which is the
ratio test to the first crossing and the report that carries it. A
caller can then see what happened instead of reading a warning that the
step was clamped.

`estimate()` today takes the linear step, clips any variable that leaves
its bounds, warns, and returns. The warning names the variables and
stops there. A caller cannot tell how far along the perturbation the
active set changed, whether a constraint became active before any
variable did, or whether the gap against a re-solve comes from the
barrier parameter, from a regularized factor, or from relaxed bounds.

## What this adds

`estimate_report(model, perturb)` takes the same perturbation argument
`estimate()` takes and returns an `EstimateReport`. It runs the same
step and measures it. Nothing about `estimate()` changes beyond the
perturbation parsing moving into a helper the two now share.

The report carries the fraction of the perturbation that fits before
the first bound is reached, and which variable or constraint reaches
its bound there. It also carries the coordinates the full step pushes
past a bound and by how much, the constraint violation at the predicted
point, and the activity classification for variables and constraints.
Alongside those are the three things that separate this predictor from
the exact value at the perturbed active set: the barrier parameter, the
factor's inertia-correction perturbations, and whether the solve
relaxed its bounds.

The classification is passed through from `Solver.classify_activity`
rather than recomputed. This is the same classifier the covariance and
information work shipped, which is what left item 0 with only the ratio
test and the report assembly to build. No Rust changed.

## The ratio test excludes coordinates already on a bound

A coordinate on its bound cannot be crossed by a step, and scoring one
produces noise rather than a finding. The remaining gap is the small
slack the barrier leaves, the step component is the same size, and
their quotient can take any value. Left in, it becomes the minimum on
any model carrying an active bound, which is most real ones.

The exclusion reads the classification rather than a distance, because
that slack is `O(mu)` at a strongly active bound but `O(sqrt(mu))` at a
weakly active one, and no single distance threshold separates the
second from a coordinate genuinely close to its bound. A distance
threshold remains as the backstop for a solve with no classification
available.

Equality constraints and the constraints pinning the declared Params do
not take part either. An equality constraint is held to first order and
admits no activity change, and a pin constraint's right-hand side is
what the perturbation moves.

## Which residual this reports

Items 0 and 4 of the roadmap both say "the residual" without saying
which one. This reports the constraint violation at the predicted
point, the primal half, since item 0 is scoped as assembly of
quantities the core already exposes and the NL evaluators supply it
directly.

The dual half belongs to item 4, whose corrector is specified as one
primal-dual iteration returning the residual. A primal-dual iteration
holds the updated multipliers, so the full KKT residual is computed
where they already live and comes back as a number. Assembling it out
here instead would mean exposing the multiplier steps in user space,
where only the primal block and the equality multipliers have a
mapping from the compound vector today.

Worth making explicit in the roadmap note, since nothing in the text
currently distinguishes the two.

## A relaxed solve is reported, not raised

`classify_activity` raises for a solve that ran with a non-zero
`bound_relax_factor`, since relaxed bounds shift the slacks it reads.
Rather than propagate that, the report sets `bounds_relaxed` and leaves
the classification empty, and everything else in it is still measured.
A diagnostic that raised there would give the caller nothing exactly
when they are trying to find out why the estimate and a re-solve
disagree.

## Validation

The suite adds 18 tests, each naming its model and pinning its number:

- The step fraction matches a brute-force scan of the unclamped step,
  on the same coordinate exactly and to 1e-12 relative. The same holds
  with a fixed variable present, which the solve removes and which
  shifts every later factor column.
- The step fraction matches hand-computed crossings to 1e-8, for a
  variable bound and for a constraint that binds while both variables
  are still interior.
- The violation matches direct evaluation at the predicted point to
  1e-12.
- The crossing distances match the unclamped estimate to 1e-9.
- An interior perturbation crosses nothing and reports a fraction above
  one.
- A strongly active bound and a weakly active bound are classified as
  such, and neither is reported as a crossing.
- The barrier parameter and the classification agree with calling the
  classifier directly.
- Constraints are named as the model names them, not by the clone names
  the declared-parameter surgery gives them.
- Two parameters perturbed at once agree with the brute-force scan.
- The report is the same through `SolverFactory('pounce')`,
  `SolverFactory('pounce_v2')` and the contrib `SolverFactory('pounce')`,
  since `Pounce.solve` sends a model carrying declarations down the same
  in-process route.
- An end-to-end case saturates a control with a setpoint move. The
  report names the control and returns a fraction below one, and
  re-asking at exactly that fraction crosses nothing.

The local run passes 295 tests in pyomo-pounce, against a 277 baseline
on main.

## Error message correction

The no-session error in `estimate`, `gradient`, `covariance` and
`information` told the reader to solve with `SolverFactory('pounce')`,
which has been incomplete since the v2 interface was added. All five
messages now name the three routes. Verifying that the report is route
independent is what turned this up.

### DIFF
--- a/pyomo-pounce/pyomo_pounce/sens.py
+++ b/pyomo-pounce/pyomo_pounce/sens.py
@@ -1104,9 +1104,15 @@ class EstimateReport:
         Variable data to the distance by which the full step leaves
         that variable's bound. These are the variables `estimate()`
         clamps.
-    crossed_rows : dict
-        Constraint name to the same distance, for inequality
+    crossed_rows : ComponentMap
+        Constraint data to the same distance, for inequality
         constraints.
+
+    Those two are keyed by model components, because every coordinate
+    that crosses has one. The two classifications below are keyed by
+    solve-space name instead, because they cover every coordinate of
+    the solve, including the ones the declared-parameter surgery
+    created, which have no counterpart on the model.
     violation : float
         Largest constraint violation at the predicted point, measured
         against the perturbed right-hand sides. This is the primal
@@ -1168,12 +1174,21 @@ def _row_step(session, dx):
     """Jacobian at the base point times the step, in row order."""
     rows, cols = session.nl.jacobian_structure()
     vals = np.asarray(session.nl.jacobian(session.base_x))
-    dg = np.zeros(len(session.con_names))
-    np.add.at(dg, np.asarray(rows), vals * dx[np.asarray(cols)])
-    return dg
+    return np.bincount(np.asarray(rows),
+                       weights=vals * dx[np.asarray(cols)],
+                       minlength=len(session.con_names))
 
 
-def _ratio_test(base, step, lo, hi, names, live=None):
+#: No bound at all reaches here as the reader's sentinel rather than an
+#: infinity: `read_nl` seeds every bound vector with +-1e19, so an
+#: `isfinite` test passes on it and scores a crossing at 1e18 times the
+#: perturbation. The magnitude test is the convention the rest of the
+#: package already uses (`preflight`, `block_init`, gh #401 / #403).
+_NO_BOUND = 1e19
+
+
+def _ratio_test(base, step, lo, hi, names, live=None, on_bound=None,
+                mu=float("nan")):
     """Smallest fraction of `step` that reaches a bound, and where.
 
     Only coordinates strictly inside their bounds take part, because a
@@ -1185,28 +1200,47 @@ def _ratio_test(base, step, lo, hi, names, live=None):
     active bound. Activity is reported through the classification
     instead.
 
-    The caller passes the exclusion in `live`, built from that same
-    classification, since that slack is O(mu) at a strongly active
-    bound but O(sqrt(mu)) at a weakly active one and no single distance
-    threshold separates the second from a coordinate genuinely close to
-    its bound. The threshold below is the backstop for when no
-    classification is available.
+    Two things decide that a coordinate is on a bound, and both are
+    needed. `on_bound` carries the classification, which is exact where
+    the classifier commits. It is applied per SIDE, not per coordinate:
+    a coordinate held at one bound can still be carried across its
+    other one, so excluding it outright drops a real crossing.
+
+    The distance test below covers the coordinates the classifier
+    declines to rule on, where the label cannot answer the question
+    because `ambiguous` spans both a coordinate near its bound with
+    room left and one already on it. What separates those is the size
+    of the remaining gap, which is O(mu) at a strongly active bound and
+    O(sqrt(mu)) at a weakly active one, against O(1) room in the
+    interior. So the threshold scales with sqrt(mu), measured four
+    orders of magnitude clear of the interior case. Without a
+    classification there is no `mu` either, and it falls back to a
+    fixed threshold.
 
     Coordinates whose step is below the roundoff of their own magnitude
     also take no part, since dividing by one puts the crossing at an
     enormous multiple of the perturbation, which is not a finding.
     """
     scale = np.maximum(1.0, np.abs(base))
-    bound = np.where(step > 0, hi, lo)
+    toward_hi = step > 0
+    bound = np.where(toward_hi, hi, lo)
     distance = bound - base
-    ok = (np.abs(step) > 1e-12 * scale) & np.isfinite(bound) & (
-        np.abs(distance) > 1e-9 * scale)
+    on_bound_gap = 1e-9 if not np.isfinite(mu) else max(1e-9, 10.0 * mu ** 0.5)
+    ok = (np.abs(step) > 1e-12 * scale) & (np.abs(bound) < _NO_BOUND) & (
+        np.abs(distance) > on_bound_gap * scale)
     if live is not None:
         ok &= live
+    if on_bound is not None:
+        # the nearer bound is the one the coordinate is held at, and it
+        # is only that side the classification rules out
+        at_hi = (hi - base) <= (base - lo)
+        ok &= ~(on_bound & (toward_hi == at_hi))
     if not ok.any():
         return float("inf"), None
     fraction = np.full(base.shape, np.inf)
-    fraction[ok] = distance[ok] / step[ok]
+    # a relaxed solve can settle outside a bound, which leaves no room
+    # to move rather than a negative amount of it
+    fraction[ok] = np.maximum(distance[ok] / step[ok], 0.0)
     i = int(np.argmin(fraction))
     return float(fraction[i]), names[i]
 
@@ -1214,7 +1248,11 @@ def _ratio_test(base, step, lo, hi, names, live=None):
 #: Classifications that put a coordinate ON its bound. Both have the
 #: slack going to zero, so neither can be crossed by a step. Strongly
 #: active leaves an O(mu) gap and weakly active an O(sqrt(mu)) one, and
-#: neither gap is room the step can move through.
+#: neither gap is room the step can move through. `ambiguous` and
+#: `unidentified` are absent on purpose: the first spans coordinates on
+#: their bound AND coordinates near it with room left, and the second
+#: says only that the curvature is below scale, so neither label
+#: answers the question. `_ratio_test`'s distance test rules on those.
 _AT_BOUND = ("strongly_active", "weakly_active")
 
 
@@ -1270,23 +1308,19 @@ def estimate_report(model, perturb):
 
     row_names = _user_row_names(session)
 
-    # The classifier raises for a solve that relaxed its bounds,
-    # because relaxed bounds shift the slacks it reads. Report that as
-    # the fact it is: the rest of the report is still measured, and a
-    # diagnostic that raised here would give the caller nothing
-    # exactly when they are trying to find out why the estimate
-    # disagrees with a re-solve. `mu` comes from the classifier, so it is unavailable
-    # too.
+    # The classifier raises for a solve that relaxed its bounds, since
+    # relaxed bounds shift the slacks it reads. Ask the solve what it
+    # ran under instead of provoking that and reading the message, and
+    # report it as the fact it is. The rest of the report is still
+    # measured, where a diagnostic that raised would give the caller
+    # nothing exactly when they are trying to find out why the estimate
+    # disagrees with a re-solve. `mu` comes from the classifier, so it
+    # is unavailable on that path too.
     mu, activity, row_status = float("nan"), {}, {}
-    bounds_relaxed = False
     var_on_bound = row_on_bound = None
-    try:
+    bounds_relaxed = bool(session.solver.bound_relax_factor)
+    if not bounds_relaxed:
         act = session.solver.classify_activity()
-    except Exception as err:                       # noqa: BLE001
-        if "bound_relax_factor" not in str(err):
-            raise
-        bounds_relaxed = True
-    else:
         mu = float(act["mu"])
         activity = dict(zip(session.var_names, act["var_status"]))
         row_status = dict(zip(row_names, act["row_status"]))
@@ -1294,8 +1328,7 @@ def estimate_report(model, perturb):
         row_on_bound = _on_bound(act["row_status"])
 
     alpha, first = _ratio_test(base, dx, lo, hi, session.var_names,
-                               live=None if var_on_bound is None
-                               else ~var_on_bound)
+                               on_bound=var_on_bound, mu=mu)
     first_kind = None if first is None else "variable"
     dg = _row_step(session, dx)
     g_base = np.asarray(session.nl.constraints(base))
@@ -1303,9 +1336,8 @@ def estimate_report(model, perturb):
     # activity, and a pin constraint moves by construction
     live = g_l < g_u
     live[list(pin_idx)] = False
-    a_row, f_row = _ratio_test(
-        g_base, dg, g_l, g_u, row_names,
-        live=live if row_on_bound is None else live & ~row_on_bound)
+    a_row, f_row = _ratio_test(g_base, dg, g_l, g_u, row_names, live=live,
+                               on_bound=row_on_bound, mu=mu)
     if a_row < alpha:
         alpha, first, first_kind = a_row, f_row, "constraint"
 
@@ -1317,11 +1349,13 @@ def estimate_report(model, perturb):
             crossed[ov] = float(max(lo[i] - x_new[i], x_new[i] - hi[i]))
     g_pred = g_base + dg
     gtol = 1e-9 * np.maximum(1.0, np.abs(g_pred))
-    crossed_rows = {}
+    crossed_rows = ComponentMap()
     out_of_bounds = (g_pred < g_l - gtol) | (g_pred > g_u + gtol)
     for j in np.where(live & out_of_bounds)[0]:
-        crossed_rows[row_names[j]] = float(
-            max(g_l[j] - g_pred[j], g_pred[j] - g_u[j]))
+        oc = model.find_component(row_names[j])
+        if oc is not None:
+            crossed_rows[oc] = float(
+                max(g_l[j] - g_pred[j], g_pred[j] - g_u[j]))
 
     # the perturbation IS the shift of the pin constraints' right-hand
     # sides, so the violation is measured against the shifted ones
--- a/pyomo-pounce/pyomo_pounce/sens.py
+++ b/pyomo-pounce/pyomo_pounce/sens.py
@@ -1200,22 +1200,33 @@ def _ratio_test(base, step, lo, hi, names, live=None, on_bound=None,
     active bound. Activity is reported through the classification
     instead.
 
-    Two things decide that a coordinate is on a bound, and both are
-    needed. `on_bound` carries the classification, which is exact where
-    the classifier commits. It is applied per SIDE, not per coordinate:
-    a coordinate held at one bound can still be carried across its
-    other one, so excluding it outright drops a real crossing.
+    A coordinate the full step carries past a bound is always scored,
+    whatever the exclusion would otherwise say. That case IS a crossing,
+    at a fraction below one by construction, and the caller reads it in
+    `crossed` off the same predicate and the same tolerance. Deciding it
+    twice is what let the two disagree, reporting that 998 times the
+    perturbation fits while naming a coordinate the single step leaves
+    its bound by 0.25.
 
-    The distance test below covers the coordinates the classifier
-    declines to rule on, where the label cannot answer the question
-    because `ambiguous` spans both a coordinate near its bound with
-    room left and one already on it. What separates those is the size
-    of the remaining gap, which is O(mu) at a strongly active bound and
+    So the exclusion below only ever applies to coordinates that do not
+    cross, and cannot cost a crossing. Two things drive it. `on_bound`
+    carries the classification, which is exact where the classifier
+    commits, applied per SIDE rather than per coordinate, since a
+    coordinate held at one bound can still be carried across its other
+    one.
+
+    The distance test covers the coordinates the classifier declines to
+    rule on, where the label cannot answer the question because
+    `ambiguous` spans both a coordinate near its bound with room left
+    and one already on it. What separates those is the size of the
+    remaining gap, which is O(mu) at a strongly active bound and
     O(sqrt(mu)) at a weakly active one, against O(1) room in the
     interior. So the threshold scales with sqrt(mu), measured four
-    orders of magnitude clear of the interior case. Without a
-    classification there is no `mu` either, and it falls back to a
-    fixed threshold.
+    orders of magnitude clear of the interior case. It is capped
+    because it is applied relative to the coordinate's own magnitude,
+    and a loose `mu` at termination would otherwise widen it without
+    limit. Without a classification there is no `mu` either, and it
+    falls back to a fixed threshold.
 
     Coordinates whose step is below the roundoff of their own magnitude
     also take no part, since dividing by one puts the crossing at an
@@ -1225,16 +1236,27 @@ def _ratio_test(base, step, lo, hi, names, live=None, on_bound=None,
     toward_hi = step > 0
     bound = np.where(toward_hi, hi, lo)
     distance = bound - base
-    on_bound_gap = 1e-9 if not np.isfinite(mu) else max(1e-9, 10.0 * mu ** 0.5)
-    ok = (np.abs(step) > 1e-12 * scale) & (np.abs(bound) < _NO_BOUND) & (
-        np.abs(distance) > on_bound_gap * scale)
-    if live is not None:
-        ok &= live
+    present = np.abs(bound) < _NO_BOUND
+    moving = np.abs(step) > 1e-12 * scale
+
+    # the same predicate, and the same tolerance, that fills `crossed`
+    reached = base + step
+    tol = 1e-9 * np.maximum(1.0, np.abs(reached))
+    crosses = present & moving & np.where(
+        toward_hi, reached > bound + tol, reached < bound - tol)
+
+    floor = 1e-9 if not np.isfinite(mu) else min(
+        1e-2, max(1e-9, 10.0 * mu ** 0.5))
+    interior = np.abs(distance) > floor * scale
     if on_bound is not None:
         # the nearer bound is the one the coordinate is held at, and it
         # is only that side the classification rules out
         at_hi = (hi - base) <= (base - lo)
-        ok &= ~(on_bound & (toward_hi == at_hi))
+        interior &= ~(on_bound & (toward_hi == at_hi))
+
+    ok = present & moving & (crosses | interior)
+    if live is not None:
+        ok &= live
     if not ok.any():
         return float("inf"), None
     fraction = np.full(base.shape, np.inf)
--- a/pyomo-pounce/pyomo_pounce/sens.py
+++ b/pyomo-pounce/pyomo_pounce/sens.py
@@ -901,7 +901,8 @@ def _session_for(component):
     if reg is None or reg.session is None:
         raise RuntimeError(
             "no sensitivity session: declare_sens_param() then solve with "
-            "SolverFactory('pounce') first")
+            "SolverFactory('pounce'), SolverFactory('pounce_v2') or the "
+            "contrib SolverFactory('pounce') first")
     return reg.session
 
 
@@ -1005,6 +1006,25 @@ def gradient(target=None, *, wrt):
     return Gradient(session, targets, params)
 
 
+def _perturbation_deltas(session, perturb):
+    """Pin constraints and right-hand-side shifts for a perturbation.
+
+    The shift is measured from the pin constraint's stored right-hand
+    side, which holds the Param's value at the solve, not from the
+    Param's current value on the model.
+    """
+    items = perturb.items() if hasattr(perturb, "items") else perturb
+    pin_idx, deltas = [], []
+    for comp, newval in items:
+        for pd in _iter_data(comp):
+            nv = newval[pd.index()] if comp.is_indexed() and hasattr(
+                newval, "__getitem__") else newval
+            pin = _param_pin(session, pd)
+            pin_idx.append(pin)
+            deltas.append(float(nv) - float(session.nl.g_l[pin]))
+    return pin_idx, deltas
+
+
 def estimate(model, perturb, clamp=True):
     """First-order estimate of the solution at perturbed parameter values.
 
@@ -1031,22 +1051,10 @@ def estimate(model, perturb, clamp=True):
     if session is None:
         raise RuntimeError(
             "no sensitivity session: declare_sens_param() then solve with "
-            "SolverFactory('pounce') first")
+            "SolverFactory('pounce'), SolverFactory('pounce_v2') or the "
+            "contrib SolverFactory('pounce') first")
 
-    items = perturb.items() if hasattr(perturb, "items") else perturb
-    pin_idx, deltas = [], []
-    for comp, newval in items:
-        for pd in _iter_data(comp):
-            nv = newval[pd.index()] if comp.is_indexed() and hasattr(
-                newval, "__getitem__") else newval
-            pin = _param_pin(session, pd)
-            pin_idx.append(pin)
-            # the step shifts the pin constraint's RHS, and that RHS
-            # holds the Param's solve-time value exactly, so it is the
-            # baseline: a caller that has already written the new value
-            # into the Param (the receding-horizon pattern) gets the
-            # same estimate as one that has not
-            deltas.append(float(nv) - float(session.nl.g_l[pin]))
+    pin_idx, deltas = _perturbation_deltas(session, perturb)
 
     # parametric_step returns the factor's x block (var-x); base_x and
     # everything below it (nl.x_l/x_u, var_names) are full-x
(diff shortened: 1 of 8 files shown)


pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1075,6 +1083,265 @@ def estimate(model, perturb, clamp=True):
...
+    else:
+        mu = float(act["mu"])
+        activity = dict(zip(session.var_names, act["var_status"]))
+        row_status = dict(zip(row_names, act["row_status"]))
+        var_on_bound = _on_bound(act["var_status"])
+        row_on_bound = _on_bound(act["row_status"])
+
+    alpha, first = _ratio_test(base, dx, lo, hi, session.var_names,
+                               live=None if var_on_bound is None
+                               else ~var_on_bound)
```
jkitchin: **`alpha` can miss a bound the step actually crosses, and then contradicts `crossed` in the same report.**

The exclusion is per *coordinate*, but a crossing is per *side*: a coordinate sitting on one bound can still be carried across its other one. Excluding it entirely drops that crossing from the ratio test.

Repro (`x` weakly active at `ub = 1`, step drives it to −2, past `lb = −1`):

```
activity[x] = weakly_active
alpha = 8.5      first = 'z'          <-- reported
crossed = {'x': 1.0000381}            <-- same report, same call
```

The true first crossing is 2/3. So the report says "no bound reached until 8.5× the perturbation" while simultaneously reporting that the 1× step leaves `x`'s bound by 1.0 — and a caller taking `alpha * perturb` as the safe step overshoots by 12×.

Not only the weakly-active case: `crossed` non-empty with `alpha >= 1` is reachable whenever any two-sided coordinate is classified on-bound. A cheap invariant for the suite: **`r.crossed` non-empty ⇒ `r.alpha < 1`**.

Fix (verified — see the `_ratio_test` comment below; all 18 tests here stay green and this case becomes `alpha = 0.6667, first = 'x'`): keep the exclusion but apply it only to the side the coordinate is on, passing `on_bound=` instead of folding it into `live=`.

---

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1075,6 +1083,265 @@ def estimate(model, perturb, clamp=True):
...
+    ok = (np.abs(step) > 1e-12 * scale) & np.isfinite(bound) & (
+        np.abs(distance) > 1e-9 * scale)
+    if live is not None:
+        ok &= live
+    if not ok.any():
+        return float("inf"), None
+    fraction = np.full(base.shape, np.inf)
+    fraction[ok] = distance[ok] / step[ok]
+    i = int(np.argmin(fraction))
+    return float(fraction[i]), names[i]
```
jkitchin, line 1275: Two more defects live in these thirteen lines, both of which the test file's own `brute_force_multi` helper already gets right — which is exactly why the 18 tests can't see them.

**1. `np.isfinite(bound)` does not recognize the `.nl` no-bound sentinel.** `read_nl` seeds all four bound vectors with `±1e19` (`crates/pounce-nl/src/nl_reader.rs:602-605`), and this repo's convention is the magnitude test `abs(b) < 1e19` (`preflight.py:35`, `block_init.py:388`, gh #401/#403). `docs/src/sensitivity.md:174` flags this precise trap — *"finite, so an `isinf()` test will not catch it"* — in the section on Param-dependent bounds rewritten as rows, which is one of the cases that lands here. On a model with no real bound in the step direction:

```
alpha = 1.6666e+18   first = 'z'   isinf = False
```

The docstring above promises `inf` / `first = None`. `if not ok.any()` is effectively dead code today. The helper in the tests does `b = v.ub if v.ub is not None else np.inf`, reading the user model instead of the `.nl` vectors, so it never hits this.

**2. Negative `alpha` on the relaxed-bounds path.** A relaxed solve settles *outside* a declared bound, so `distance` goes negative and the quotient with it — on the very path this PR advertises as "the measured half still lands":

```
y* = 5.000000049582308  (ub = 5.0)
bounds_relaxed = True
alpha = -89.03   first = 'y'
```

reported as "the fraction of the requested perturbation that fits". `brute_force_multi` does `a = max((b - x0) / d, 0.0)`; this path doesn't.

Verified fix for both, plus the per-side exclusion from the other comment:

```python
def _ratio_test(base, step, lo, hi, names, live=None, on_bound=None):
    scale = np.maximum(1.0, np.abs(base))
    toward_hi = step > 0
    bound = np.where(toward_hi, hi, lo)
    distance = bound - base
    # bound presence is the .nl reader's magnitude test, not isfinite:
    # every absent bound reaches here as the +-1e19 sentinel
    ok = (np.abs(step) > 1e-12 * scale) & (np.abs(bound) < 1e19) & (
        np.abs(distance) > 1e-9 * scale)
    if live is not None:
        ok &= live
    if on_bound is not None:
        # the classification is per coordinate, the crossing is per
        # side: a coordinate held at one bound can still be carried
        # across the other one
        at_hi = (hi - base) <= (base - lo)
        ok &= ~(on_bound & (toward_hi == at_hi))
    if not ok.any():
        return float("inf"), None
    fraction = np.full(base.shape, np.inf)
    # a solve that relaxed its bounds can sit outside one; no step fits
    # there, rather than a negative one
    fraction[ok] = np.maximum(distance[ok] / step[ok], 0.0)
    i = int(np.argmin(fraction))
    return float(fraction[i]), names[i]
```

with the row call site passing `live=live, on_bound=row_on_bound`. I ran this against the branch: **18/18 tests here still pass**, and the three probes become `alpha=inf, first=None` / `alpha=0.6667, first='x'` (matching the hand computation) / `alpha=0.0`.

---

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1075,6 +1083,265 @@ def estimate(model, perturb, clamp=True):
...
+
+
+#: Classifications that put a coordinate ON its bound. Both have the
+#: slack going to zero, so neither can be crossed by a step. Strongly
+#: active leaves an O(mu) gap and weakly active an O(sqrt(mu)) one, and
+#: neither gap is room the step can move through.
+_AT_BOUND = ("strongly_active", "weakly_active")
```
jkitchin, line 1286: Design question rather than a confirmed bug: `ambiguous` and `unidentified` are left out, so they are treated as crossable.

`AMBIGUOUS` is the classifier's explicit *"undetermined at this μ"* verdict (`crates/pounce-sensitivity/src/activity.rs:72-74`) — the coordinate it declines to rule on. If such a coordinate really is on its bound, scoring it produces exactly the tiny-gap-over-tiny-step noise the PR description argues against, and it becomes the minimum. The distance backstop in `_ratio_test` can't catch it either: `1e-9 * scale` is orders of magnitude tighter than the O(√μ) gap in play (μ = 1e-9 → √μ ≈ 3e-5), which is the same argument the description makes for why a distance threshold cannot stand in for the classification.

Either add both to `_AT_BOUND`, or say in this comment why an undetermined coordinate is treated as interior. I did not construct a repro for this one.

---

pyomo-pounce/pyomo_pounce/sens.py
```diff
@@ -1075,6 +1083,265 @@ def estimate(model, perturb, clamp=True):
...
+                f"crossed={n}, violation={self.violation:.3e}, "
+                f"mu={self.mu:.3e}{flags})")
+
+
+def _row_step(session, dx):
+    """Jacobian at the base point times the step, in row order."""
+    rows, cols = session.nl.jacobian_structure()
+    vals = np.asarray(session.nl.jacobian(session.base_x))
+    dg = np.zeros(len(session.con_names))
+    np.add.at(dg, np.asarray(rows), vals * dx[np.asarray(cols)])
```
jkitchin: Nit: `np.add.at` is numpy's slowest scatter (an unbuffered ufunc loop). The same result, typically an order of magnitude faster on the MPC-scale Jacobians this diagnostic is aimed at:

```python
dg = np.bincount(np.asarray(rows), weights=vals * dx[np.asarray(cols)],
                 minlength=len(session.con_names))
```

---

jkitchin (changes requested):
I built the branch locally (maturin wheel + editable `pyomo-pounce`) and ran it. **All 18 new tests pass**, and CI is green on all six checks. The design writeup is genuinely excellent, and the hard parts are right: the fixed-variable `scatter_x` space handling, passing the classification through instead of recomputing it, excluding pin rows from the ratio test, and measuring `violation` against *shifted* pin right-hand sides. `estimate()` is untouched.

Three defects in the ratio test, though, all reproduced by running the code:

1. **`alpha` can miss a bound the step actually crosses, contradicting `crossed` in the same report.** The classification exclusion is per coordinate, not per side. A variable weakly active at `ub` gets carried past `lb` and the ratio test never looks: `alpha = 8.5, first = 'z'` alongside `crossed = {'x': 1.0}`, where the true crossing is at 2/3.
2. **`np.isfinite(bound)` doesn't recognize the `.nl` `±1e19` no-bound sentinel** that `read_nl` seeds every bound with, and that this repo already handles with `abs(b) < 1e19` (gh #401/#403; `docs/src/sensitivity.md:174` flags this exact trap). Result: `alpha = 1.67e18, first = 'z'` where the documented contract says `inf` / `None`.
3. **Negative `alpha` on the relaxed-bounds path** (`alpha = -89.03`), reported as "the fraction of the perturbation that fits" — on the path the description advertises as still measurable.

Details, repro output, and a verified fix are inline on `_ratio_test` and its variable call site. The fix is contained in that one function plus the two call sites; I ran it and all 18 tests here stay green while the three cases become `inf` / `0.6667` / `0.0`.

Worth noting that the test file's own `brute_force_multi` helper already handles both #2 and #3 correctly (`b = v.ub if ... else np.inf`, and `max(..., 0.0)`) — the production path is missing exactly the two guards the reference has, which is why the suite can't see them. A suggested invariant that would catch #1: **`r.crossed` non-empty ⇒ `r.alpha < 1`**.

Two smaller items not worth inline comments:

- **Docs.** `docs/src/sensitivity.md` documents `estimate()` and isn't updated for a new public API; CONTRIBUTING item 3 makes the book section part of the definition of done, and the CI docs guard only catches orphan pages so it won't fail on its own. `docs/src/pyomo.md:201` also enumerates the four accessors that read the KKT factorization directly — now five.
- **Relaxed-solve detection couples to a Rust string.** `"bound_relax_factor" not in str(err)` silently turns back into a re-raise if that message is reworded. Worth asserting the token in `crates/pounce-sensitivity/tests/solver_session.rs:377`, which today only checks `is_err()`.

Plus two trivia: `crossed` is a `ComponentMap` keyed by var data while `crossed_rows` / `activity` / `row_activity` are dicts keyed by name strings, an asymmetry callers will trip on; and the CHANGELOG's "All five messages" covers four pre-existing sites (the fifth is the new function's own).

Requesting changes on 1–3 only. The `ambiguous` question and the docs item are fine as follow-ups if you'd rather land this.

---

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Fixed, using your version: `on_bound` is now a parameter separate from `live`, masked by `at_hi` so only the side the coordinate is held on is excluded. Both call sites pass it.

Your invariant is in as a test across every fixture in the file: a non-empty `crossed` or `crossed_rows` means the reported fraction is below one. That is the check that would have caught this.

One thing I could not reproduce, which does not change the finding. Reconstructing the case here (`x` bounded `[-1, 1]`, weakly active at `ub` after solving at `p = 1`, then perturbing to `p = -2`), the crossing lands at 4/3 rather than 2/3, because at a degenerate base point the linear step carries the two-sided derivative, half the interior value, so `x` reaches `-0.5` rather than `-2`. Your model may differ from my reconstruction. The defect and the fix are the same either way: before it, `x` was excluded outright and `first` came back `'z'`.

So the model-level test asserts that `x` is scored at all and that the fraction agrees with the brute-force scan, rather than hardcoding a value that depends on how the degenerate step falls. The exact 2/3 arithmetic is pinned separately in a unit test where the arrays are chosen rather than solved for.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Both fixed with your code. The bound-presence test is now the magnitude convention, `abs(bound) < _NO_BOUND`, and the fraction is clamped at zero. You are right that `if not ok.any()` was dead: nothing reached it before, since every absent bound arrived as a finite sentinel.

The observation about `brute_force_multi` is the one worth acting on, and it is why I could not see either of these. The reference read bounds off the Pyomo model and clamped at zero, so it disagreed with the production path in exactly the two places that path was wrong, and the suite could only confirm the agreement. It now reads the same `nl.x_l` / `nl.x_u` arrays and applies the same sentinel test, so it cannot diverge silently again.

Both cases are also pinned as unit tests on constructed arrays, since neither is reachable on demand by solving a model: the sentinel needs a solve that leaves no bound in the step direction, and the negative distance needs the solver to settle a chosen amount outside a bound.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: You found a real defect here, but it is the backstop rather than the membership of `_AT_BOUND`, and I have fixed it there.

`ambiguous` cannot decide this question, because it covers two opposite situations. Below the weakly-active band, between `sqrt(mu)` and 1e-1, the coordinate is between inactive and weakly active: near its bound with real room left, and excluding it drops a genuine imminent crossing. Above the band, between 1e1 and `1/sqrt(mu)`, it is between weakly and strongly active and is on its bound, which is your case. Adding it to `_AT_BOUND` is wrong for the first, leaving it out is wrong for the second. `unidentified` says only that the curvature is below scale, which carries no information about bound proximity at all.

What does separate them is the gap, which is your point that `1e-9` cannot see an O(sqrt(mu)) slack. Measured here at the three regimes:

| regime | distance to bound | distance / sqrt(mu) |
|---|---|---|
| strongly active | 2.5e-9 | 0.00005 |
| weakly active | 4.4e-5 | 1.5 |
| inactive | 5.0e-1 | 9988 |

Nearly four orders between the on-bound cases and the interior one, so the threshold now scales with `sqrt(mu)` and lands in that gap. Neither label needs a special case then, and a future verdict from the classifier is handled without changing this code. The comment on `_AT_BOUND` says why the two are absent.

The direction matters too. Excluding on a label that cannot answer risks your defect 1, a missed crossing, which is a wrong answer. Including risks only a conservative fraction, which the caller can see alongside the coordinate's status in the same report.

The right long-term home is item 2. When path-following needs breakpoint ordering and constraint drop, the ratio test moves into the core, which owns 1e-1, 1e1 and the `sqrt(mu)` edges and can answer "on its bound" from its own constants instead of a copy of them.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Applied. `np.bincount` with `weights` and `minlength`, values identical on the existing fixtures.

devin-griff:
All ten items are addressed, including the two you were happy to take as follow-ups. Details are in the four inline replies. The rest:

**Docs.** `docs/src/sensitivity.md` gains a section under the Pyomo part covering `estimate_report`, the fields, why coordinates on a bound take no part in the ratio test, and the three provenance fields, with the `estimate()` paragraph linking to it. `docs/src/pyomo.md` now says five accessors.

**The option coupling.** Removed rather than guarded. `Solver.bound_relax_factor` is new on the Python solver and returns the value the held solve ran under, so `estimate_report` asks the solve whether its bounds were relaxed instead of provoking `classify_activity` and reading its message. Nothing matches a string now. `classify_activity_refuses_a_solve_that_relaxed_its_bounds` still gains the assertion you asked for, since the token stays load-bearing for anyone else, and the test says why in a comment.

**The keying asymmetry.** `crossed_rows` is a `ComponentMap` keyed by constraint data, matching `crossed`. Every entry has a component, since pin constraints are already excluded. `activity` and `row_activity` stay keyed by solve-space name, because they cover every coordinate of the solve including the ones the surgery created, which have no counterpart on the model. The docstring states that split so it reads as a rule rather than an accident.

**Changelog.** Four, not five.

Locally: 303 pyomo-pounce tests pass, up from 295, with both binaries rebuilt at this head. pounce-py and pounce-sensitivity are green, clippy is unchanged at 442 warnings, and `cargo fmt --check` is clean.

pyomo-pounce/pyomo_pounce/sens.py
jkitchin: Fix verified — I rebuilt the branch and this case now reports `alpha = 0.6666531, first = 'x'`.

**On the number you couldn't reproduce:** the difference is the perturbation size, not the model. Mine perturbs `p` from 1 to **−5**, not −2. Your own explanation is what produces it: the degenerate base point carries the two-sided derivative, `dx/dp = 0.5`, so `Δp = −6` gives `Δx = −3` and `x` lands at −2.0, crossing `lb = −1` at exactly 2/3. At your `p = −2`, `Δx = −1.5` puts `x` at −0.5, which crosses nothing — hence 4/3. Both numbers are right for their own perturbation. Your choice to assert against the brute-force scan rather than hardcode is the better test either way.

**One residual case, and it's a gap in the fix I proposed, not something you introduced.** The invariant I suggested is stronger than the per-side mask I suggested alongside it. The mask handles "held at one bound, carried across the other"; it does not handle "held at a bound, pushed *further* past that same side", where the coordinate is still excluded outright:

```python
m.x = pyo.Var(bounds=(None, 1.0), initialize=0.0)      # weakly active at ub
m.z = pyo.Var(bounds=(-500.0, 500.0), initialize=0.0)
m.obj = pyo.Objective(expr=(m.x - m.p)**2 + (m.z - m.p)**2)   # p: 1 -> 1.5
```

```
activity[x] = weakly_active
alpha = 998.0    first = 'z'
crossed = {'x': 0.249956}      unclamped x = 1.24996  (ub = 1.0)
```

Same contradiction as the original finding: "998× the perturbation fits" alongside "the 1× step leaves x's bound by 0.25". Adding this shape to `test_a_bound_the_step_crosses_is_never_missed` fails it today, so the invariant is currently asserted rather than guaranteed.

The coherent rule, I think, is to derive the exclusion and `crossed` from **one** predicate, since the invariant says they must agree: if the full step leaves a bound by more than the crossing tolerance, that coordinate must be scored (`alpha = 0` when it is already on that bound and moving outward — no room, rather than no bound). That does **not** reintroduce "alpha = 0 on any model with an active bound", and your own fixtures show why: in a strongly active case I measured, the outward step is O(μ) and falls under the same `1e-9 * scale` tolerance `crossed` uses, so it drops out on both sides at once. The tolerance is already doing the noise filtering the exclusion was added for.

Not a blocker for landing, but the test asserting it makes it worth an issue at least.

---

pyomo-pounce/pyomo_pounce/sens.py
jkitchin: Your two-regime argument for `ambiguous` is right, and I withdraw the suggestion to add it to `_AT_BOUND` — `[sqrt(mu), 1e-1)` and `(1e1, 1/sqrt(mu)]` really are opposite situations, and I only had the second in mind. Moving the decision onto the gap is the better answer, and the measured table makes the case.

One thing to watch on the new floor, though: `max(1e-9, 10 * sqrt(mu))` is applied **relative to `max(1, |base|)`**, so its absolute width scales with both `mu` and the coordinate's magnitude, and it is uncapped. Measured on the branch:

| solve | mu | floor (relative) | absolute band at \|base\| = 1000 |
|---|---|---|---|
| ordinary (`bounded()`-style) | 2.5e-9 | 5.0e-4 | 0.50 |
| `tol=1e-2`, `mu_init=1e-1` | 9.1e-6 | 3.0e-2 | 30 |

At μ = 2.5e-9 that is already 500,000× the old `1e-9` band. A large-magnitude coordinate with, say, 0.4 units of genuine room to its bound is now dropped from the ratio test — and a dropped coordinate is a *missed* crossing, the same direction of error as defect 1, not a conservative one.

The exposure grows with anything that leaves μ loose at termination: a loose `tol`, an iteration limit, `mu_target`. Both cases I measured still gave the right `alpha` because the bounds were far away, so this is a hazard rather than a demonstrated wrong answer — but `test_ratio_test_scales_its_distance_floor_with_the_barrier` pins only μ = 1e-9, so nothing would catch it drifting.

Worth considering a cap (`min(1e-2, 10 * sqrt(mu))` keeps the μ = 1e-9 calibration intact while bounding the worst case), or asserting the floor against a loosely-converged solve so the interaction is pinned. Agreed that the real home is item 2, where the core owns those constants instead of a copy of them.

---

jkitchin (commented):
All three requested changes are fixed, and I verified them by rebuilding `432f022` rather than reading the diff — the new `Solver.bound_relax_factor` getter means the wheel had to be rebuilt to run any of this.

| defect | before | now |
|---|---|---|
| per-side exclusion | `alpha = 8.5, first = 'z'` with `crossed = {'x': 1.0}` | `alpha = 0.6666531, first = 'x'` |
| `1e19` sentinel | `alpha = 1.67e18, first = 'z'` | `alpha = inf, first = None` |
| relaxed solve | `alpha = -89.03` | `alpha = 0.0` |

**26 passed** in `test_estimate_report.py`, and **100 passed / 7 skipped** across it plus `test_covariance.py` and `test_sens.py`. CI is green on all six checks including `Test (ubuntu-latest)`.

Replacing the message-matching with the new getter is a better fix than what I suggested — reading the value the solve ran under beats catching an error to learn it — and keeping the token assertion in the core test for the other callers is the right call. The docs section, the `ComponentMap` for `crossed_rows`, and `bincount` all landed too. Making `brute_force_multi` read the same `nl.x_l` / `nl.x_u` arrays the report reads is the change that matters most for the future: the reference disagreeing with the code in exactly the two places the code was wrong is why the suite could only confirm agreement.

Two follow-ups, in the threads rather than here, neither blocking:

- **The invariant is stronger than the fix — including the one I proposed.** "Held at a bound, pushed *further* past that same side" still contradicts `crossed`: `alpha = 998` alongside `crossed = {'x': 0.25}`. Adding that shape to `test_a_bound_the_step_crosses_is_never_missed` fails it today. Suggested direction in the thread: derive the exclusion and `crossed` from one predicate, since the test asserts they agree.
- **The `10*sqrt(mu)` floor is uncapped and relative to `max(1, |base|)`**, so its absolute width grows with both μ and coordinate magnitude — 0.5 units at |base| = 1000 on an ordinary solve, 30 on a loosely-converged one. A dropped coordinate is a missed crossing, the same direction of error as defect 1.

I withdraw the `ambiguous` suggestion — your two-regime argument is right, and moving that decision onto the gap is the better answer.

Not blocking: my three requested changes are resolved.

---

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: Your case reproduces exactly, at `alpha = 998.0` with `crossed = {'x': 0.249956}`. I fixed it the way you suggest, with one predicate deciding both on the same tolerance. Anything the full step carries past a bound is now scored, at a fraction below one by construction. A coordinate already on that bound and moving outward scores 0.0, which says it has no room rather than no bound. Your case returns `alpha = 1.76e-4, first = 'x'`.

The exclusion now only ever reaches coordinates that do not cross, so it cannot cost a crossing. That makes the invariant hold by construction rather than by assertion. Your shape is in the test, perturbed both toward the held bound and away from it.

Your prediction about the strongly active case held. There the outward step is O(mu) and falls under the same tolerance, so the coordinate drops out of `crossed` and out of the ratio test together, and a fraction of zero does not come back.

Four existing tests assumed the previous rule, that a coordinate on its bound stays excluded even when the step carries it past. They now exercise the exclusion with steps that stop short of the bound, which is the case it exists for.

You are right that the difference in the number is the perturbation size. Your 2/3 is correct for `p` going to -5, and mine reached -0.5 and crossed nothing, which put it at 4/3. Asserting against the scan keeps the test independent of that.

This is fixed here, so it needs no issue.

pyomo-pounce/pyomo_pounce/sens.py
devin-griff: The floor is capped at `min(1e-2, max(1e-9, 10 * sqrt(mu)))`, which keeps the mu = 1e-9 calibration and limits the worst case. I took your other suggestion as well and pinned it against a loose solve rather than only at a chosen mu.

That solve shows the cap changes the answer rather than only guarding against a change. On a coordinate at magnitude 1000 with a bound 0.4 away:

| solve | mu | uncapped floor | absolute band |
|---|---|---|---|
| ordinary | 2.5e-9 | 5.0e-4 | 0.50 |
| `tol=1e-2`, `mu_init=1e-1` | 1.8e-6 | 1.4e-2 | 13.6 |

Both bands exceed the 0.4 units of room, so without the cap `x` is dropped and its crossing missed. With it, both solves report `alpha = 0.2`. The test asserts that the uncapped floor still exceeds the gap, so it fails if that stops being true instead of passing without testing the cap.

The other follow-up changes how much this matters. A crossing is now scored whatever the floor says, so a coordinate the floor drops can no longer be a missed crossing. It can only make `alpha` too large above one, where the answer to whether the full step fits is unchanged. The cap limits that residual error.

I agree that item 2 is the real home for the constants.

pyomo-pounce/pyomo_pounce/sens.py
jkitchin: Verified on `7772f23`. My case reports `alpha = 0.000176395, first = 'x'`, and the strongly active case still comes back `first = 'x'` with nothing scored at zero, so the O(μ) outward step drops out of both as predicted. I swept 24 combinations of that fixture (three `z` bound widths × eight perturbations, toward and away from the held bound) and the invariant holds on every one. The capped floor gives `alpha = 0.2` on both the ordinary and the loose solve. 31 tests pass.

"Holds by construction" is right for every path that has a classification. **It does not hold on the relaxed one**, though this predates your change rather than coming from it:

```
base: y = 5.000000049582308            # ub = 5.0, left outside by 5e-8

p -> 4.000000000001   alpha = inf   crossed = {'y': 4.96e-08}   step y = 0.0
p -> 2.0              alpha = 2.0   crossed = {'y': 4.90e-08}   step y = -5.6e-10
```

`bound_relax_factor` lets the solve settle `y` outside its declared bound, so `crossed` reports it — the distance is the relaxation itself, and the step moves `y` by 5.6e-10 or not at all. The ratio test is right to score nothing: in the step direction nothing is reached. The two disagree because `crossed` is measuring where the *solve* left the coordinate, not what the *step* did to it.

Which one to change depends on what `crossed` is for. It is documented as "the variables `estimate()` clamps", and `estimate()` genuinely does clamp `y` — so it is faithful as written, and the honest fix is to scope the invariant to `not bounds_relaxed`. If you would rather `crossed` mean "what the step carried past a bound", subtract the base point's own violation so a pre-existing one does not appear as a crossing. Either way it is worth one line in the test, since the invariant is otherwise stated unconditionally.

Narrow path — `_build_session` pins `bound_relax_factor` to 0 unless the user sets it explicitly — but it is the path this report exists to keep measurable, so I would not leave it silently untrue. Not a blocker.

---

devin-griff:
Both follow-ups are fixed, in `5dff1a7` and `7772f23`. The detail is in replies on the two threads, so the short version here:

**The invariant is now guaranteed rather than asserted.** Your residual case reproduced exactly, at `alpha = 998.0` alongside `crossed = {'x': 0.249956}`. Participation and `crossed` now come from one predicate on the same tolerance, so anything the full step carries past a bound is scored, at a fraction below one by construction. Your case returns `alpha = 1.76e-4, first = 'x'`. The exclusion only ever reaches coordinates that do not cross, so it cannot cost a crossing. Your prediction about the strongly active case held: the outward step there is O(mu) and falls under the same tolerance, so the coordinate leaves `crossed` and the ratio test together.

**The floor is capped** at `min(1e-2, max(1e-9, 10 * sqrt(mu)))`, and pinned against a loose solve as well as at a chosen mu. That test shows the cap changes the answer rather than only guarding against a change: at magnitude 1000 with a bound 0.4 away, the uncapped floor is 0.50 absolute on an ordinary solve and 13.6 under `tol=1e-2, mu_init=1e-1`, so without it that coordinate is dropped and its crossing missed.

The two interact in your favor. Because a crossing is always scored now, a coordinate the floor drops can no longer be a missed crossing, only an `alpha` too large above one, where whether the full step fits is unchanged.

Four existing tests assumed the previous rule and now exercise the exclusion with steps that stop short of the bound. 308 pass locally, and CI is green on five checks with `Test (ubuntu-latest)` still running.

pyomo-pounce/pyomo_pounce/sens.py
jkitchin: To be explicit about the disposition, since this is the last thing I have open: **not a blocker, and I would land the PR without it.**

- The path needs an explicit opt-in. `_build_session` pins `bound_relax_factor` to 0 *before* the user options are applied, so it only exists if someone deliberately sets it. No default solve reaches it.
- Neither field is wrong by itself. `crossed` is documented as the set `estimate()` clamps, and `estimate()` does clamp `y`; the ratio test is right that nothing is reached in the step direction. They answer two different questions, and each answers correctly.
- The magnitude is the relaxation itself, 5e-8 — the solve's own slop, not a modelling-scale number that would mislead a reader.
- The report already announces the degraded path: `bounds_relaxed` is true and both classification maps are empty.

So what is actually defective is narrower than the behavior: the docstring and `test_a_bound_the_step_crosses_is_never_missed` state the invariant unconditionally, and on this one path it does not hold. A clause scopes it:

```python
if (len(r.crossed) or len(r.crossed_rows)) and not r.bounds_relaxed:
    assert r.alpha < 1.0, ...
```

plus a sentence on `crossed` saying that on a relaxed solve it can name a coordinate the *solve* left outside its bound rather than one the *step* carried past. Worth doing because an unqualified invariant is the kind of thing that gets built on later — which is how the original per-side defect survived in the first place — but it costs a line and does not gate the merge.

Everything else I raised is fixed and verified. Nice work on the turnaround.

---

devin-griff: Both are applied, and there was a third statement of the same kind.

The invariant in `test_a_bound_the_step_crosses_is_never_missed` now carries your clause, and `crossed` says it measures both bounds at the predicted point, so on a relaxed solve an entry can be a coordinate the solve left outside its bound rather than one the step carried past.

The `alpha` field carried the same claim and is the more consequential one. It read "At least 1.0 means the full step stays inside every bound", which is exactly false here, since your case returns 4.0 with `y` outside its upper bound. It is scoped the same way now.

The relaxed case also gets its own test rather than only being excluded from the invariant. It asserts what is true there: `crossed` names `y`, the amount is the relaxation itself rather than a modelling-scale number, and `alpha` is at or above one.

I reproduced your case before changing anything and got the same numbers, `crossed = {y: 4.9e-8}` alongside `alpha = 4.0` from `x`. Your reading of the mechanism holds. `crossed` tests both bounds at the predicted point while the ratio test looks only along the step direction, so with the step moving `y` downward each answers its own question correctly.

Agreed on why it earns the line. An unqualified invariant is what gets built on later, which is how the per-side case survived.

jkitchin: Verified on `c0b79e1`. 32 tests pass, and re-running my sweep, the only cases it still flags are the relaxed ones — which is now the documented, separately tested contract rather than an unqualified claim, so that is the right outcome.

Catching the `alpha` docstring yourself is the better half of this. I raised the invariant and `crossed`, and missed that "at least 1.0 means the full step stays inside every bound" made the same promise in the field a caller actually reads to decide whether to take the step. My own sweep had the evidence sitting in it — `alpha = inf` and `alpha = 2.0` on relaxed solves with `y` outside its bound — and I read it as being about `crossed` rather than about `alpha`'s own contract. Scoping all three together is what makes it consistent.

Giving the relaxed path a test that asserts what *is* true there, rather than only excluding it, is also the right call: an exclusion alone would have left the path unpinned.

Nothing further from me. Everything I raised is fixed and verified.

---

PR #587, opened 2026-08-14
sens: fix-relax and the barrier correction, for sIPOPT parity

devin-griff:
This implements item 1 of the sensitivity roadmap (#255), which is both
halves of upstream's fix-relax strategy plus the barrier correction.
Together they are what the roadmap defines as parity.

`estimate()` took the linear step and clipped whatever left a bound.
Clipping costs more than the one variable, because every other variable
keeps a value computed on the assumption that the clipped one was free
to move where the step said. On a model where `y = 2x + 1` and `x` hits
its lower bound, that returns `y = -5`, which is not on the constraint
at all.

## What this adds

`estimate(model, perturb, mode="fix_relax")` repairs the active set the
step implies, rather than clipping it. `mode="linear"` is the default
and keeps clipping. `max_passes` bounds the work.

The refinement is both of the paper's cases (Pirnay, Lopez-Negrete and
Biegler 2012, section 2.5), and they fail differently:

- A variable the step carries past a bound is **pinned** there,
  activating it (their equation 17). On the model above this returns
  `y = 1`, matching a full re-solve.
- A bound multiplier the step drives negative is **set to zero**,
  deactivating that bound so the variable can move (equation 18).
  Without it a variable on a bound stays there however hard the
  perturbation pulls, because the linear step preserves
  complementarity. On a model whose bound wants to release, that is the
  difference between `x = 0` and the correct `x = 1.667`.

Each adds a row to the held factorization and re-solves through the
Schur complement over the added rows, which is the paper's equations 19
through 22. The factorization is never rebuilt.

Separately, every parametric step now carries the **barrier
correction**, the paper's equation 11. The step is taken against a
factorization held at the solve's final `mu`, so on its own it estimates
where the barrier problem's solution moves rather than the original
problem's. This applies on all paths and has no option, since the
barrier problem's answer is not one a caller wants.

## Verified against sIPOPT itself

Driven through `pyomo.contrib.sensitivity_toolbox` against sIPOPT
3.14.19, on cases built so that each isolates one mechanism:

| case | pounce vs sIPOPT | what it would be without |
|---|---|---|
| pin a crossing variable | 2e-8 | clip is off by 0.12 vs a re-solve |
| release a bound | 1e-6 | returns 0.0 where the answer is 1.667 |
| barrier correction, `tol = 1e-3` | 2.4e-7 | 9e-6 |
| barrier correction, `tol = 1e-8` | 4e-10 | 2e-9 |

Running the reference is what found the second and third rows. Every
case I had built before that exercised only the pin, so pounce and
sIPOPT agreed on all of them, and I had concluded twice from my own
measurements that the missing pieces did not matter. They do.

The correction's sign was also set by that comparison rather than
derived. The first attempt doubled the disagreement instead of removing
it, which is how it was caught.

## Behavior changes

**`sens_boundcheck` refines instead of clamping**, across
`--sens-boundcheck`, `SensSolve::with_boundcheck` and
`solve_with_sens(sens_boundcheck=True)`. The option is named after
upstream's, and upstream's refines. This changes what it guarantees. The
clamp always returned a point inside the declared box, and the
refinement does not, because pins are limited by the problem's degrees
of freedom. The CLI's help text no longer promises the box, and its
message reports pinned coordinates rather than clamped ones.

**Existing answers move by `O(mu)`** from the barrier correction. The
shift is below notice at a converged tolerance, and it is the point of
the change where the solve leaves `mu` loose.

**One bound tolerance instead of three.** Three numbers answered "is
this outside its bound", disagreeing across surfaces: 1e-3 on the CLI
flag, 1e-9 in the binding, and a third invented inside `estimate()`. It
now comes from the solve's own `bound_relax_factor`, which is the value
that says how far outside a bound the solve was willing to leave a
point.

**`clamp` means the same thing in both modes**, clipping whatever is
still outside a bound at the end. Under `fix_relax` the pins usually
leave nothing to clip.

## Limits, and what is reported

Pins consume degrees of freedom. Past that no step holds every bound at
once and the augmented system is singular, which a dense LU does not
report, returning a solution around 1e15 instead. Each pass checks that
it achieved the displacement it asked for and drops the pin when it did
not. `estimate()` warns, names the variables still outside, and says
whether the pass budget or the degrees of freedom stopped it, since only
the first can be fixed by asking for more.

Neither pounce nor sIPOPT gives a meaningful bound multiplier at a newly
pinned bound. Measured, sIPOPT's refined answer leaves it near `mu`
where the true multiplier at the perturbed solution is 18.

Fix-relax is also still first order and enforces the active set the step
predicts, not the true one. On one perturbation of upstream's own
example it pins a variable a re-solve leaves interior at 0.25.

## Testing

`crates/pounce-sensitivity/tests/parametric_cpp.rs` checks the
refinement against a full re-solve on upstream's parametric example,
against its baked-in golden output, on a three-degree-of-freedom model
where three coordinates cross at once, and for the refusal when pins
would exceed the degrees of freedom.

`pyomo-pounce/tests/test_fix_relax.py` covers the Pyomo surface: both
halves against a re-solve, the barrier correction pinned to sIPOPT's own
values, behavior under a `user-scaling` change of variables, that the
modes agree exactly where nothing crosses, and both sides of what
`clamp` now means.

Coverage was checked by disabling each mechanism and counting failures.
That found two thin spots and one defect. The release had a single test
and worked only on lower bounds. On an upper bound the guard that
catches an over-determined pin set demanded 1e-6 relative accuracy and
discarded a working pin that landed at 1.4e-6, returning 0.999 where the
answer is -3. The guard is for a singular solve returning around 1e15,
so it is now 1e-3, which still catches that by twelve orders. And one
test could not fail, since the barrier correction was asserted to 1e-6
at a tolerance where the term is 4e-10.

Locally 329 pyomo-pounce tests pass, along with every
`pounce-sensitivity` and `pounce-cli` suite, and `cargo fmt --all
--check` is clean.

## Note on the roadmap

#255 calls the barrier correction negligible, which holds at a converged
tolerance and not at a loose one, and it assigns constraint drop to item
2 when the release is the "relax" in fix-relax and belongs here. Both
are wording fixes to that note rather than to this change.


crates/pounce-sensitivity/src/solver.rs
```diff
@@ -510,8 +514,167 @@ impl Solver {
...
+            crate::boundcheck::expand_bounds(n_x, &nl.px_l(), &nl.px_u(), nl.x_l(), nl.x_u())
+        };
+        // Those bounds bound the algorithm's `x̃ = d ⊙ x`, while
+        // `state.x` and the step are both in the model's own units
+        // (gh#486 stage 3). Undo the change of variables on the bounds
+        // so all three agree, rather than projecting onto the wrong box.
+        // A negative factor reflects the interval, so the sides swap.
+        if let Some(d) = state.backsolver.variable_scaling_full() {
+            for i in 0..n_x {
+                let di = d[i];
```
jkitchin: `variable_scaling_full()` is indexed in full-x, but `i` here runs over var-x (`0..n_x`). With user scaling and any fixed variable present the two index spaces diverge, so every bound from the first fixed variable onward gets un-scaled by some other variable's factor — silently projecting onto the wrong box.

The other two call sites get this right by mapping first. `crates/pounce-cli/src/sens.rs:413`:

```rust
let di = d[nlp_ref.var_x_to_full_x(i as Index) as usize];
```

and `convenience.rs` does the same. This is the only site that indexes the full-x vector directly.

---

crates/pounce-sensitivity/src/solver.rs
```diff
@@ -510,8 +514,167 @@ impl Solver {
...
+                dims[0] + dims[1] + dims[2] + dims[3],
+                dims[0] + dims[1] + dims[2] + dims[3] + dims[4],
+            );
+            let (data, _, _) = state.backsolver.activity_handles();
+            let d = data.borrow();
+            let curr = d.curr.as_ref().ok_or(SolverError::NotConverged)?;
+            let mut out = Vec::new();
+            for (off, v) in [(z_l_off, &curr.z_l), (z_u_off, &curr.z_u)] {
+                for (k, &base) in crate::vec_util::dense_to_vec(&**v).iter().enumerate() {
+                    out.push(crate::boundcheck::BoundMultiplier { row: off + k, base });
```
jkitchin, line 663: `base` is taken raw off `curr.z_l` / `curr.z_u`, which are in the solve's scaled coordinates (`df·z/d`), but it is consumed as the release right-hand side and added to `dx_full[row]`, which the block just above deliberately brings into the model's natural units. Under the default gradient-based scaling `df` and `d` are not 1, so the release RHS is off by `df/d`.

This is worse than a wrong multiplier: the release row enters the Schur complement, so a bad RHS perturbs the whole refined step, every coordinate, not just the released one. It would not show up in the tests that pin `mode="fix_relax"` against sIPOPT if those run unscaled.

Same conversion the bounds get a few lines up needs applying to `base` (divide by `d` for the variable's row, multiply by `df`), or the multipliers need to be handed over in scaled units with `dx_full` converted after.

---

crates/pounce-cli/src/sens.rs
```diff
@@ -412,5 +376,98 @@ fn try_compute_sens_step(
...
+                let vals = v
+                    .as_any()
+                    .downcast_ref::<DenseVector>()
+                    .map(|d| d.expanded_values())
+                    .unwrap_or_default();
+                for (k, &b) in vals.iter().enumerate() {
+                    out.push(pounce_sensitivity::boundcheck::BoundMultiplier {
+                        row: off + k,
+                        base: b,
+                    });
```
jkitchin, line 442: Same unit mismatch as the library path. Lines 411–422 immediately above go to the trouble of converting `x_nat`, `lo` and `hi` out of the solve's coordinates into the step's units, and then `base` is passed through raw from `curr.z_l` / `curr.z_u`, which are still scaled (`df·z/d`).

So on this path the pinned coordinates and the bounds agree with `dx_full`, and the release rows do not.

---

crates/pounce-sensitivity/src/convenience.rs
```diff
@@ -449,32 +451,78 @@ impl SensSolve {
...
-                    if let Some(d) = &d_var {
-                        for (s, &di) in dx_primal.iter_mut().zip(d.iter()) {
-                            *s /= di;
+                    match crate::boundcheck::refine_step_onto_bounds(
+                        &backsolver_for_refine,
+                        &dx_full,
+                        &x_nat[..n_x],
+                        &lo,
+                        &hi,
+                        &[],
```
jkitchin: Empty multiplier slice, so the release half of fix-relax never runs on this path. `refine_step_onto_bounds` will only ever pin here; no bound can be released.

That splits the three surfaces the PR description says are now the same computation. The CLI (`sens.rs:426`) and `parametric_step_bounded` both build a real `mults` vector; `SensSolve::with_boundcheck` and `solve_with_sens(sens_boundcheck=True)` route through here and get pin-only behavior. On the release-case model from the description, this returns `0.0` where `estimate(mode="fix_relax")` returns the correct `1.667`.

It also contradicts the docs — `lib.rs:35-47`, `docs/src/sensitivity.md`, and the CHANGELOG entry all describe one shared refinement across the three surfaces.

The same block that the CLI uses to assemble `mults` from `curr.z_l` / `curr.z_u` and `block_dims()` should work here (with the unit fix from the other thread applied).

---

crates/pounce-sensitivity/src/solver.rs
```diff
@@ -510,8 +514,167 @@ impl Solver {
...
+    /// `bound_relax_factor` outside its bound, so anything within that
+    /// is on the bound. An unrelaxed solve gets a roundoff floor.
+    ///
+    /// Passes stop when nothing is outside its bound by that much, at
+    /// `max_passes`, or when a pin cannot be achieved because the pins
+    /// have exhausted the problem's degrees of freedom. None of those
+    /// is an error: the step returned is the last one computed, and the
+    /// returned pin list says how far the refinement got. `max_passes`
+    /// is a budget, since each pass costs a dense `k × k` solve and the
+    /// point of the refinement is to stay cheaper than a re-solve.
```
jkitchin, line 601: The contract stated here — "none of those is an error: the step returned is the last one computed" — holds for the pass budget and for the achievement check that drops an unachievable pin, but not for an exactly singular Schur solve. That path returns `Err` and discards the last good step, so a caller who hits it loses the refinement *and* the plain step.

Exact singularity is the rare case (the near-singular one is what the 1e-3 achievement guard catches), but it is reachable, and it is the one case where the caller gets nothing back rather than a partially-refined answer. Falling back to the last achieved step, with the pin list saying how far it got, would match what this paragraph promises.

---

crates/pounce-py/src/solver.rs
```diff
@@ -262,6 +262,62 @@ impl PySolver {
...
+    /// which can be outside the box. Clamping a coordinate back to its
+    /// bound leaves every other coordinate at its predictor value, so
+    /// the answer is feasible but no longer consistent with the KKT
+    /// relations. This adds a row pinning the offending coordinate at
+    /// the bound and re-solves, so the others move with it, which is
+    /// the refinement upstream runs under `sens_boundcheck`.
+    ///
+    /// `pinned` lists the variables pinned to reach the returned step,
+    /// worst violator first, in **user-space** indices. It is empty
+    /// when the plain step already respects every bound, and the step
```
jkitchin: `pinned` is not in user-space variable indices. `refine_step_onto_bounds` returns compound-KKT rows: var-x indices for pins, but `z_l` / `z_u` rows (≥ `n_x`, offset by `block_dims()`) for releases.

A caller following this doc will index their variable vector with a release row and either panic or read an unrelated variable. The entries also aren't all "variables pinned" — some are bounds released, which is the opposite action.

Either map back to user space before returning, or document the two kinds of row and how to tell them apart.

---

crates/pounce-sensitivity/src/boundcheck.rs
```diff
@@ -177,142 +162,223 @@ mod tests {
...
+/// is never rebuilt.
+///
+/// Passes stop when nothing is outside its bound by more than `eps`, at
+/// `max_passes`, or when a pin cannot be achieved. That last case is
+/// how an over-determined pin set is caught: once the pins exhaust the
+/// problem's degrees of freedom no step can hold them all, the
+/// augmented system is singular, and a dense LU returns a large
+/// solution instead of reporting it. The check is whether the pass
+/// actually moved the pinned coordinates where it asked, and the step
+/// returned is the last one that did.
```
jkitchin: This 25-line doc block describes `refine_step_onto_bounds` — its `dx_plain` / `x_curr` parameters, its pass loop, its stop conditions — but it is attached to `pub struct BoundMultiplier`, whose own doc starts at line 262 with no separator. The rendered rustdoc for `BoundMultiplier` opens with a description of a function it is not.

Looks like a leftover from moving the function below the struct. The real `refine_step_onto_bounds` doc at line 274 covers the same ground, so this block can most likely just be deleted.

---

crates/pounce-sensitivity/src/boundcheck.rs
```diff
@@ -177,142 +162,223 @@ mod tests {
...
+/// coordinate leaves every other one at its linear-predictor value, so
+/// the result satisfies the bounds and no longer satisfies the
+/// constraints.
+///
+/// `dx_plain` is the unrefined step over the whole compound KKT vector,
+/// `x_curr` and the bounds cover its primal block, and all four are in
+/// the same units. Each pass augments the held factorization with every
+/// pin so far and takes the Schur complement over them, so the cost is
+/// one dense `k × k` solve and a backsolve per pass. The factorization
+/// is never rebuilt.
```
jkitchin: "one dense `k × k` solve and a backsolve per pass" undercounts. The Schur complement is rebuilt from scratch each pass, which costs one backsolve per constrained row, not one per pass. Pass `k` costs `k + 1` backsolves, so the default `max_passes = 16` is 136 backsolves, not 16.

"The factorization is never rebuilt" is the true and load-bearing claim, and it stays true. But the per-pass cost is what a caller uses to decide `max_passes` against a re-solve, and it grows quadratically rather than linearly.

Worth correcting in the other four places the same claim appears (the `parametric_step_bounded` doc, `lib.rs`, `docs/src/sensitivity.md`, and the CHANGELOG).

---

crates/pounce-cli/src/sens.rs
```diff
@@ -412,5 +376,98 @@ fn try_compute_sens_step(
...
+    // re-solves against the factor, which only exists in this scope.
+    if let Some(eps) = boundcheck_eps {
+        let n_x = backsolver.block_dims()[0];
+        let x_curr = {
+            let d = data.borrow();
+            let curr = d.curr.as_ref()?;
+            curr.x
+                .as_any()
+                .downcast_ref::<DenseVector>()
+                .map(|v| v.values().to_vec())
```
jkitchin: `.values()` on a `DenseVector` carries the homogeneous-representation `debug_assert` that the L16 fix removed elsewhere in this repo — it fires on a vector in homogeneous form, which `curr.x` can be. The `mults` block at line 436 already uses the replacement:

```rust
.map(|d| d.expanded_values())
```

`expanded_values()` is the drop-in here too.

---

crates/pounce-cli/src/sens.rs
```diff
@@ -412,5 +376,98 @@ fn try_compute_sens_step(
...
+            &lo,
+            &hi,
+            &mults,
+            eps,
+            16,
+        ) {
+            Ok((refined, pinned)) => {
+                if !pinned.is_empty() {
+                    eprintln!(
+                        "pounce: --sens-boundcheck pinned {} primal coordinate(s)                          at their bounds and re-solved",
```
jkitchin: 26 spaces in the middle of the user-visible string — source indentation that ended up inside the literal rather than wrapping it. Prints as:

```
pounce: --sens-boundcheck pinned 3 primal coordinate(s)                          at their bounds and re-solved
```

Same defect in an assertion message in `crates/pounce-sensitivity/tests/parametric_cpp.rs`.

---

jkitchin (commented):
Reviewed the full diff (14 files, +1762/−385).

The core algorithm holds up. I checked the pin right-hand-side convention (`corr[r] = -rhs[r]`, correct given `S = -B K⁻¹ Aᵀ`), that corrections are measured from `dx_plain` rather than accumulated across passes, the `skip`/`taken` bookkeeping, the achievement guard at 1e-3, and the `pins.pop()` rollback. Those are right. The findings below are all in the plumbing around it — four of them unit or index handling.

**Would block on these four:**

1. `solver.rs:624` — `variable_scaling_full()` (full-x) indexed with a var-x index. Wrong bounds under user scaling + any fixed variable.
2. `solver.rs:657` — `BoundMultiplier::base` read in scaled units, consumed in natural units. Corrupts the entire refined step, not just the released row.
3. `cli/sens.rs:441` — same unit mismatch, on a path that converts `x_curr`/`lo`/`hi` correctly three lines above.
4. `convenience.rs:516` — `&[]` for multipliers, so the release half never runs on `with_boundcheck` / `solve_with_sens(sens_boundcheck=True)`. Those two surfaces still return `0.0` on the release-case model where `estimate(mode="fix_relax")` returns 1.667.

(4) is the one I'd fix first — it means two of the three surfaces the description says now share one computation still don't.

The rest are a doc-vs-behavior mismatch on `pinned`, a stranded doc block, an understated cost claim, a reintroduced `debug_assert`, and a whitespace bug in a user-visible message.

**Checked and cleared**, so they don't come up twice: the release branch does *not* fire spuriously on inactive bounds — their `Δz` is tiny because the parametric RHS is zero on the z rows, and the new barrier correction cancels `base ≈ μ/s`. And the `eps`-vs-`bound_relax_factor` question is moot on the Rust side, since `nlp.x_l()`/`x_u()` already return the relaxed bounds.

---

devin-griff:
All ten are fixed. Two of them differ from what you suggested, and checking the cost claim and the release case each found something worth reporting.

## The four blocking

(1) now reads `variable_scaling()`, which is already var-x, rather than mapping the full-x vector through `var_x_to_full_x`. The result is the same, and it removes the chance of indexing the wrong one.

(2), (3) and (4) turned out to be one fix. Rather than convert `base` at each of the three call sites, I added a defaulted `natural_units_factor()` to `SensBacksolver`, returning the backsolver's own `F`, and `refine_step_onto_bounds` converts the multipliers once on entry. Every caller now passes the raw `curr.z_l` value and none of them needs to know the convention. That is a wider change than the review asked for, and the reason is that the same conversion was wrong at three sites independently, which is why it should not be a caller's responsibility. `BoundMultiplier::base` documents that it is raw and that the refinement converts it.

**On the direction.** Your comment gives the convention as `df·z/d` and then says to divide by `d` and multiply by `df`. Those are inverses of each other, so I took it from the code rather than from either reading. `ScalingTnlp::finalize_solution` divides `x` by `d` and multiplies `z_l` / `z_u` by `d`, and the `conj` field doc gives `F` on the z rows as `d_{px(j)}` from the change of variables and `1/df` from the objective. Natural is `scaled · d / df`, which is your prose and not your parenthetical. Using `F` itself rather than rebuilding the product means the multiplier and the `z` rows of the step cannot disagree.

(4) also has a test. `test_solve_with_sens_boundcheck_releases_a_bound_the_step_leaves` ports the release model to the raw API, so it covers `solve_with_sens(sens_boundcheck=True)`. It returns 5/3 now and returned 0.0 before. I checked that it fails without the fix rather than assuming so. Reverting `convenience.rs` to `&[]` and rebuilding, it fails.

## The rest

The singular Schur solve now drops the condition it could not solve and breaks, so the caller gets the last achieved step and a row list saying how far it got, which is what the paragraph promised. `dx` is only written after the achievement check, so the returned step is the last achieved one rather than a partial write.

`pinned` is documented as compound KKT rows, read against `block_dims()`, with the two kinds and their offsets spelled out. I started with "at or above `n_x`" and corrected it. Rows between `n_x` and the `z_l` offset belong to the `s`, `y_c` and `y_d` blocks and never appear, so that is not how a caller separates them.

The stranded doc block is deleted, `.values()` is `expanded_values()`, and both whitespace defects are fixed. I swept the PR's Rust files for further instances of each. There is no other run of spaces inside a string literal, and the one remaining `.values()` on an iterate vector is at `algorithm_backsolver.rs:859`, which predates this branch and is guarded by an explicit `is_homogeneous()` check.

## Two notes on the review itself

`lib.rs` does not carry the cost claim. You listed it among the other four places, and it is not there under that wording or under "never rebuilt", "Schur", or "budget". The claim appears in four places, all corrected, plus `pounce-py/src/solver.rs`, which you did not list and which had it too.

Your summary lists five items under "the rest", but there are six non-blocking comments. The singular Schur solve is not in that list. I treated it as a finding and fixed it.

## Releasing a bound gets worse as the base solve tightens

I found this while writing the release test, and it is not in your review. These are the errors on the release model, with everything else held fixed:

| base solve `tol` | `dx[0]` error against 5/3 |
|---|---|
| 1e-6 | 2.6e-8 |
| 1e-8 | 7.3e-7 |
| 1e-10 | 2.2e-4 |
| 1e-12 | no step returned at all |

Releasing drives a multiplier to zero against a factorization held at the final mu, and the slack of an active bound goes to zero with mu, so that row is worse conditioned the tighter the solve. The direction is the opposite of what a barrier residual would do, which is what makes me read it as conditioning rather than as the `O(mu)` term.

The test fixture solves at 1e-8 and the comment says why. The last row is the one I would look at separately. Returning no step is worse than returning an imprecise one, and it is reachable from an ordinary request for a tight solve.

## Checks

Rust `pounce-sensitivity` 103 pass, `pyomo-pounce` 329 pass, `python/tests` 792 pass. Clippy is clean on the three crates. The three `python/tests` failures here are this machine's NumPy 1, which CI does not have.

crates/pounce-sensitivity/src/solver.rs
devin-griff: Fixed. It reads `variable_scaling()`, which is already var-x, rather than mapping the full-x vector through `var_x_to_full_x`. Same answer as the other two sites, with one less index space to get wrong.

crates/pounce-sensitivity/src/solver.rs
devin-griff: Fixed, and moved out of the caller. `SensBacksolver` has a defaulted `natural_units_factor()` returning the backsolver's own `F`, and `refine_step_onto_bounds` converts the multipliers once on entry, so this site passes the raw `curr.z_l` value and no caller needs to know the convention. The same conversion was wrong at three sites independently, which is why it is not a caller's job.

One thing on the direction. Your comment states the convention as `df·z/d` and then says to divide by `d` and multiply by `df`, which are inverses of each other, so I took it from the code. `ScalingTnlp::finalize_solution` divides `x` by `d` and multiplies `z_l` / `z_u` by `d`, and the `conj` doc gives `F` on the z rows as `d_{px(j)}` from the change of variables and `1/df` from the objective. Natural is `scaled · d / df`, so your prose rather than your parenthetical. Using `F` itself means the multiplier and the `z` rows of the step cannot disagree.

crates/pounce-cli/src/sens.rs
devin-griff: Fixed by the same change. The conversion now happens inside `refine_step_onto_bounds`, so this site keeps passing the raw value and it agrees with the bounds converted a few lines above.

crates/pounce-sensitivity/src/convenience.rs
devin-griff: Fixed. It builds the same multiplier vector the CLI and `parametric_step_bounded` build, so all four surfaces run the same refinement.

Covered too. `test_solve_with_sens_boundcheck_releases_a_bound_the_step_leaves` ports the release model to the raw API, so it exercises `solve_with_sens(sens_boundcheck=True)`. It returns 5/3 and returned 0.0 before. I checked it fails against the unfixed code rather than assuming so, by reverting this line to `&[]` and rebuilding.

crates/pounce-sensitivity/src/solver.rs
devin-griff: Fixed. It drops the condition it could not solve and stops, so the caller gets the last achieved step and a row list saying how far it got, which is what the paragraph promised. `dx` is only written after the achievement check, so the step returned is the last achieved one rather than a partial write.

crates/pounce-py/src/solver.rs
devin-griff: Documented rather than mapped back, since a released row has no variable index that would mean "pinned". It now says these are compound KKT rows, to be read against `block_dims()`, with both kinds and their offsets.

I first wrote "at or above `n_x`" and corrected it. Rows between `n_x` and the `z_l` offset belong to the `s`, `y_c` and `y_d` blocks and never appear, so that is not how a caller separates them.

crates/pounce-sensitivity/src/boundcheck.rs
devin-griff: Deleted. The block below covers the same ground, as you expected.

crates/pounce-sensitivity/src/boundcheck.rs
devin-griff: Corrected here, in `parametric_step_bounded`, in the binding doc, in `docs/src/sensitivity.md` and in the CHANGELOG. All now say the Schur complement is rebuilt each pass, so pass `k` costs `k + 1` back-solves and the default of 16 is 136 rather than 16.

`lib.rs` does not carry the claim, though. I checked for that wording and for "never rebuilt", "Schur" and "budget". The binding doc in `pounce-py/src/solver.rs` did have it, and that one was not on your list.

crates/pounce-cli/src/sens.rs
devin-griff: Fixed, `expanded_values()`. I swept the rest of the changed files for the same thing. The only other `.values()` on an iterate vector is `algorithm_backsolver.rs:859`, which predates this branch and sits inside an explicit `is_homogeneous()` branch.

crates/pounce-cli/src/sens.rs
devin-griff: Fixed here and in the `parametric_cpp.rs` assertion.

The message was also inaccurate in a second way. It said "pinned N primal coordinate(s) at their bounds", and the list can carry released bounds as well as pins, so it now reads "pinned or released N bound(s) and re-solved".

Issue #591, opened 2026-08-14
Pyomo: SolvedToAcceptableLevel loads as warning while IPOPT reports ok


bernalde:
## Summary

An accepted POUNCE 0.10.0 solve is returned through legacy Pyomo as:

```text
solver.status                 warning
termination_condition        optimal
message                       POUNCE 0.10.0: SolvedToAcceptableLevel
```

Pyomo consequently logs a warning while loading the result. IPOPT returns the
equivalent `Solved To Acceptable Level` solve as `status=ok,
termination_condition=optimal`.

The difference occurs both through the official `SolverFactory("pounce")`
plugin and when the POUNCE executable is driven through Pyomo's generic `ipopt`
ASL interface. It breaks otherwise solver-swappable applications that treat
`status == ok` as part of their accepted-solve contract, even though both
solvers agree that termination is optimal.

Downstream integration/context:
https://github.com/SECQUOIA/LyoPRONTO/pull/147

## Environment

```text
pounce 0.10.0 (commit 066949bf, built 2026-08-11T19:25:22Z)
pounce-solver 0.10.0
pyomo-pounce 0.10.0
Pyomo 6.9.5
Python 3.13.5
target x86_64-unknown-linux-gnu
```

## Reproduction

```bash
git clone https://github.com/SECQUOIA/LyoPRONTO
cd LyoPRONTO
gh pr checkout 147
python -m pip install -e ".[dev,pyomo]"
python -m pip install "pounce-solver==0.10.0" "pyomo-pounce==0.10.0"

python - <<'PY'
import pyomo_pounce

from lyopronto.pyomo_models.paper_ocp import (
    PaperDiscretization,
    solve_paper_problem1,
)

d = PaperDiscretization(
    n_z=10,
    nfe=12,
    ncp=3,
    terminal_drying_fraction=0.995,
)
result = solve_paper_problem1(
    discretization=d,
    solver="pounce",
    solver_options={
        "max_iter": 3000,
        "tol": 1e-5,
        "acceptable_tol": 1e-4,
        "constr_viol_tol": 1e-6,
        "mu_strategy": "adaptive",
        "bound_relax_factor": 1e-8,
        "nlp_scaling_method": "gradient-based",
        "print_level": 0,
    },
    require_success=True,
)
print(result["metadata"])
PY
```

## Observed

Pyomo prints:

```text
WARNING: Loading a SolverResults object with a warning status into
model.name="unknown";
    - termination condition: optimal
[...]

Issue #491, opened 2026-08-06
[adversary] qp-active-set: PanicException across the FFI on crossed bounds (lb > ub); ipm returns primal_infeasible


jkitchin:
adversary-fingerprint: qp-active-set/panic-crossed-bounds

`pounce.solve_qp(..., method="active-set")` **panics across the FFI boundary**
(`pyo3_runtime.PanicException`) when a variable's bounds are crossed
(`lb > ub`), instead of returning `primal_infeasible`. The `ipm` path on the
identical input returns `primal_infeasible` correctly, so the two methods
disagree on a plain input-domain case.

A Rust `panic!` escaping into Python is not catchable by `except Exception`
(PyO3's `PanicException` derives from `BaseException`), so it tears down a
caller's loop even when the caller defensively wraps the solve.

## Reproduction

```python
import numpy as np, pounce
P, c = np.array([[1.0]]), np.array([0.0])
lb, ub = np.array([1.0]), np.array([0.0])   # crossed: lb > ub => empty feasible set
print("ipm       ->", pounce.solve_qp(P=P, c=c, lb=lb, ub=ub, method="ipm").status)
print("active-set->", pounce.solve_qp(P=P, c=c, lb=lb, ub=ub, method="active-set").status)
```

The crossing does not need to be large — `lb=0.0, ub=-1e-12` panics too, while
`lb == ub` is handled fine (`optimal`). So the trigger is strictly `ub < lb`.

## Expected

`status="primal_infeasible"` (an empty box is an infeasible problem), or an
up-front `ValueError` from the wrapper's validation pass — either is fine, but
not a panic. Oracles:

- **The `ipm` path itself** on the same data returns `primal_infeasible`.
- **The CLI / `.nl` route** with a crossed-bound variable also reports
  `Problem is primal infeasible` on *both* `solver_selection=qp-active-set`
  and the auto route (exit 1) — so the engine can express this outcome; only
  the `solve_qp` Python entry point panics.
- **`pounce.minimize(bounds=[(1.0, 0.0)])`** already rejects this with a clean
  `ValueError: bounds[0] is reversed: lower 1.0 > upper 0.0` on every
  `solver_selection`, including `qp-active-set`.
- **clarabel** on the equivalent inequality-form model reports
  `PrimalInfeasible`.

## Actual

```
thread '<unnamed>' panicked at .../core/src/num/f64.rs:1503:9:
min > max, or either was NaN. min = 1.0, max = 0.0
[...]

Issue #295, opened 2026-07-23
[Follow-up #275] Move the infinite-bound guard into the pounce-convex core (bindings/Rust callers still exposed)


jkitchin:
> Follow-up to #275 (fixed in #291). The user-facing surfaces are guarded; this tracks the missing defense-in-depth at the core.

## The gap

#291 rejects sign-inconsistent infinite bounds (`lb = +inf`, `ub = -inf`) and non-finite `x0` — but the fix lives entirely in the Python layer (`python/pounce/qp.py::_validate` and `python/pounce/_minimize.py`). The **root cause** is untouched: the `pounce-convex` core's bound-presence test is sign-agnostic.

`crates/pounce-convex/src/ipm.rs` `expand_bounds`:

```rust
if ub < crate::qp::BOUND_INF { ... }   // treats +inf AND any finite huge ub identically
if lb > -crate::qp::BOUND_INF { ... }  // treats -inf AND any finite huge lb identically
```

So a `QpProblem` built with `lb[i] = +inf` still has that bound silently dropped as "absent" and returns `status="optimal"` at a point that violates it. Any consumer that reaches the core **without** going through `qp.py::_validate` is still exposed:

- the raw `_pounce` pyo3 bindings (`_pounce.solve_qp`, `QpProblem`, …)
- `pounce-convex` used directly as a Rust crate
- any future Python entry point that forgets to call `_validate`

The `.nl` / CLI path is *not* exposed in practice, because AMPL `.nl` encodes an unbounded side as an *absent* bound rather than `±inf`, so the degenerate spelling can't be constructed there. But that is luck of the format, not a guarantee.

## Why it matters

The Python guard is correct and covers the documented surfaces, but "silently returns a wrong optimal on an infeasible bound" is exactly the failure mode #275 was about, and it still lives one layer down. A bindings user or a Rust embedder gets the pre-#275 behavior with no warning.

## Toward SOTA

Move the invariant into the core so every surface inherits it:

- Reject (or verifiably handle) `lb = +inf` / `ub = -inf` in `QpProblem` construction / validation in `pounce-convex`, the same place the finite `lb > ub` infeasibility is already detected (`presolve.rs`, the `tlb[k] > tub[k]` check). A bound that admits no finite point should map to `PrimalInfeasible`, not be dropped.
- Keep the Python-layer check too (it gives a better, index-named error message earlier), but stop relying on it as the *only* line of defense.

Related: #275, #291.

PR #595, opened 2026-08-14
Make `solve_socp_ipm_warm` bounds-aware and improve warmstart

GermanHeim:
## Problem

`solve_socp_ipm_warm` requires callers to encode variable bounds manually as G/h rows. This forces downstream users (including oximo), to maintain a bound-row and dual bookkeeping workaround.

Also, when `QpOptions::use_hsde` was enabled the warm SOCP path could route through HSDE, which ignores warm starts.

## Fix

I added a fix to treat `QpProblem::lb` and `ub` as first-class bounds, and append generated bound rows as a trailing nonnegative cone block.

It also forces the initial warm attempt through the direct non-HSDE driver.

## Blast radius

`pounce-covex` and dependants of it

## Tests

I had to modify a test eps, it was failing on my computer

      since moved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ changes.
 
 ## [Unreleased]
 
+- **SOCP warm starts now support first-class variable bounds.** Bounds are
+  expanded and bound duals normalized internally; warm solves force the direct
+  driver for symmetric cones and may fall back to cold HSDE when enabled.
+  Exponential/power cones use their dedicated cold HSDE route.
+
 ### Added — `estimate(mode="fix_relax")` bends the estimate around a bound instead of clipping it (#587)
 
 `estimate()` takes the linear step, and where that step leaves a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ changes.
   expanded and bound duals normalized internally; warm solves force the direct
   driver for symmetric cones and may fall back to cold HSDE when enabled.
   Exponential/power cones use their dedicated cold HSDE route.
+
 ### Added — `estimate(mode="fix_relax")` bends the estimate around a bound instead of clipping it (#587)
 
 `estimate()` takes the linear step, and where that step leaves a
@@ -255,6 +256,7 @@ willing to leave a converged point.
   GAMS in the loop, so the values are now checked against GAMS's own
   `gams.core.gmo` — `gamsapi[core]` is pure Python and needs no license, and
   CI installs it for exactly this test.
+
 - **An accepted solve no longer loads into Pyomo as a warning** (#591).
   `Solved_To_Acceptable_Level` is written into the `.sol` as AMPL
   `solve_result_num = 1` — IPOPT's own code for the same outcome
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@ changes.
   expanded and bound duals normalized internally; warm solves force the direct
   driver for symmetric cones and may fall back to cold HSDE when enabled.
   Exponential/power cones use their dedicated cold HSDE route.
-
 ### Added — `estimate(mode="fix_relax")` bends the estimate around a bound instead of clipping it (#587)
 
 `estimate()` takes the linear step, and where that step leaves a
@@ -256,7 +255,6 @@ willing to leave a converged point.
   GAMS in the loop, so the values are now checked against GAMS's own
   `gams.core.gmo` — `gamsapi[core]` is pure Python and needs no license, and
   CI installs it for exactly this test.
-
 - **An accepted solve no longer loads into Pyomo as a warning** (#591).
   `Solved_To_Acceptable_Level` is written into the `.sol` as AMPL
   `solve_result_num = 1` — IPOPT's own code for the same outcome
--- a/crates/pounce-convex/src/ipm.rs
+++ b/crates/pounce-convex/src/ipm.rs
@@ -1516,7 +1516,12 @@ fn chordal_reconstruct(sol: QpSolution, recon: &ChordalRecon, _prob1: &QpProblem
 /// Warm-started [`solve_socp_ipm`]: seed the iteration from `warm` (a nearby
 /// SOCP's solution). The warm `(s, z)` are projected into each cone's
 /// interior (orthant positivity / SOC `λ_min` floor); the solution is
-/// start-independent, so warm starting only reduces the iteration count.
+/// start-independent, so warm starting is intended to reduce iterations when
+/// compared with the same direct driver. The default cold solve uses HSDE,
+/// however, while symmetric warm solves are forced onto the direct driver;
+/// SOC-heavy problems can therefore take more iterations than cold HSDE and
+/// may return `OptimalInaccurate`, a truthful reduced-accuracy KKT result with
+/// the same objective contract.
 /// Finite variable bounds are first-class and they are expanded into a trailing
 /// nonnegative cone block and the returned bound multipliers are restored to
 /// `z_lb`/`z_ub`.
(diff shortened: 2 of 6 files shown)


jkitchin (commented):
The bounds work is good and correctly reuses the existing machinery. The warm-start half has a real defect, so I don't think this should merge as-is.

What I checked as sound: `cargo fmt` clean, no new clippy warnings, full `pounce-convex` suite green. The bound plumbing reuses `expand_bounds` / `merge_bound_duals` / `split_bound_duals` exactly as `solve_qp_ipm_warm_inner` does, `screen_variable_box` sits before expansion (matching gh #295 / #491), and the new `finite_or_failed` + `mark_timed_out` wrapper correctly mirrors the QP path. Not equilibrating is right — `QpOptions::equilibrate`'s own doc says the conic driver never does. Dropping the `assert!(!prob.has_bounds())` panic is the right call.

## 1. High — the HSDE fallback can throw the warm result away and make warm *slower* than cold

`crates/pounce-convex/src/ipm.rs:1637`. With default options (`use_hsde: true`), if the direct warm solve lands `OptimalInaccurate`, a **full cold** HSDE solve runs and its answer is returned. Measured on a bounded SOC QP (`P = I`, n = 3, `SOC₃`, finite `lb`/`ub`, θ-shifted `c`):

```
hsde=true  shift 0.1: cold Optimal it=11 | warm Optimal it=11 dobj=0.000e0
hsde=false shift 0.1: cold NumericalFailure it=26 | warm OptimalInaccurate it=13
```

The warm machinery worked — 13 iterations against cold-direct's 26. But the gate fired, a cold HSDE solve ran, and its result was returned: identical iteration count, identical objective to the last digit. Net cost is roughly 24 iterations to return the answer an 11-iteration cold solve gives, which makes `solve_socp_ipm_warm` a pessimization on this class. It also makes the new doc line — "ensuring that a supplied warm point is genuinely used" — false in exactly the cases the fallback exists for.

`OptimalInaccurate` is the wrong member of the trigger set. It is a usable answer, and the comment in `solve_socp_ipm_inner` that this block was copied from says so explicitly. Drop it from the trigger, or gate on a residual comparison rather than on status.

## 2. Medium — exp/power cones with `use_hsde: false` now return a fabricated failure

The `has_nonsym` arm (`ipm.rs:1607`) manufactures a `failed_solution` with `x = 0`, and the retry is gated on `opts.use_hsde`, so with HSDE off there is no fallback at all:

```
exp use_hsde=true:  cold Optimal obj=-10.000000 | warm Optimal obj=-10.000000
exp use_hsde=false: cold Optimal obj=-10.000000 | warm NumericalFailure obj=0.000000
```

The guard itself is a genuine fix — on `main` this path panics in `CompositeCone::from_specs`'s `unreachable!`. It is just half-done: route to `solve_nonsym(prob, cones, opts, …, None)` (ignoring the warm point) instead of fabricating a failure. That is what the cold path does, and it needs no new plumbing. `NumericalFailure` also reads as "the solver broke" rather than "warm start unsupported here".

## 3. Medium — the loosened tolerance is mis-explained

`issue417_adaptive_tau.rs:256`, 1e-5 → 1e-3. That test has `lb: vec[], ub: vec[]`, so the bounds change cannot touch it — only the forced-direct-driver change does. I confirmed it passes at 1e-5 on `origin/main` and fails at 1e-5 on this branch:

```
SOC @ 0.2, x[0]: warm 1.200222163888275 vs cold 1.2001691263533163   (Δ = 5.3e-5)
```

deterministically. So "it was failing on my computer" is not the cause. Before this PR, default `use_hsde: true` sent the "warm" solve through HSDE, which ignores the warm point — warm and cold were bit-identical, and that identity is what the 1e-5 was really pinning. Now the two run different drivers.

The change is defensible (the 1e-6 relative objective assert still holds, and the cross-driver sweep already tolerates 1e-5–1e-6 on objectives), but the test is named `..._warm_solves_are_unaffected` and needs a comment saying what the new tolerance means. The PR body should be corrected as well.

## 4. Medium — the headline claim is untested

Nothing asserts the warm SOCP path saves iterations under default options. `bounded_soc_warm_start_splits_bound_duals` is a real test of the new bound plumbing — I confirmed it takes the direct warm path (7 iterations against cold's 8) — but it only checks the duals and the optimum. Adding `assert!(warm.iters < cold.iters)` would pin the behaviour this PR is adding. The new HSDE-fallback branch has no coverage at all.

## 5. Low

- **Merge blocker**: the branch is 4 commits behind `main` and `mergeable_state` is `dirty` — CHANGELOG conflicts against `[Unreleased]` entries added since. Needs a rebase.
- **Duplication**: the `upgraded` status match (`ipm.rs:1649`) is copy-pasted verbatim from `solve_socp_ipm_inner:1037`. Worth extracting a shared `fn hsde_retry_is_upgrade(direct: QpStatus, retry: QpStatus) -> bool`.
- **Nit**: the new doc sentence chains three `and`s — "Finite variable bounds are first-class and they are expanded into a trailing nonnegative cone block and the returned bound multipliers are restored to `z_lb`/`z_ub`."

---

GermanHeim:
Done @jkitchin

jkitchin (commented):
Re-reviewed at `20e1caa`. All five findings are addressed, and I re-ran the measurements rather than taking the diff at its word. Full `pounce-convex` suite green locally (33 binaries), `cargo fmt --all -- --check` clean, no clippy errors.

**1 — fixed, and I confirmed the behaviour change.** `warm_hsde_retry_needed` drops `OptimalInaccurate` from the trigger. On the bounded SOC QP I used originally (`P = I`, n = 3, `SOC₃`, finite `lb`/`ub`), rebased onto the new base:

```
shift 0.1: cold Optimal it=11 obj=-2.406374141109
           warm OptimalInaccurate it=13 obj=-2.406374141700   dobj=5.9e-10
```

The warm answer is now returned at 13 iterations instead of being thrown away and replaced by a second cold solve. The doc sentence explaining why `OptimalInaccurate` stays out of the set is exactly the right thing to have written down.

**2 — fixed.** Routing to `solve_nonsym` is right, and keeping the nonsym+PSD combination as a clean failure matches the cold entry point. The `exp_gp_two_cones_stays_optimal` addition covers the `use_hsde: false` case that previously returned a fabricated `NumericalFailure`.

**3 — the widened asserts are necessary, and the new comment is accurate.** I verified the status relaxation is not cosmetic — on the current base the fixture genuinely lands `OptimalInaccurate` at one θ:

```
θ=0.05: cold Optimal it=11 | warm OptimalInaccurate it=14 | max|Δx| = 1.35e-5
θ=0.2 : cold Optimal it=12 | warm Optimal         it=9  | max|Δx| = 5.30e-5
```

Worth naming the residual trade this exposes, since it is the honest cost of the design rather than a defect: at θ=0.05 the warm path takes **more** iterations than cold (14 vs 11) and lands at reduced accuracy, because forcing the direct driver gives up HSDE's conditioning on SOC blocks. The objective still agrees to 1e-6 relative and `OptimalInaccurate` is a truthful status, so the contract holds — but "warm starting only reduces the iteration count" in the `solve_socp_ipm_warm` doc is now not quite true for symmetric cones under default options. A clause noting that a warm solve may cost more iterations than a cold HSDE solve on SOC-heavy problems would keep the doc honest. Your call whether that is worth another push.

**4 — good coverage.** `FailOnceBackend` is a nice way to reach the fallback branch deterministically, and `warm.iters < cold.iters` in `bounded_soc_warm_start_splits_bound_duals` pins the claim the PR is making. The `warm_hsde_fallback_tests` unit module covers the status algebra directly.

**5 — done**, and the rebase onto `1a0a6ba` cleared the conflict. Two small CHANGELOG artifacts from that rebase:

- Missing blank line between the new `- **SOCP warm starts…**` block and the `### Added — estimate(mode="fix_relax")…` heading that follows it, so the heading is glued to the bullet.
- A blank line was deleted between the GAMS `SOLVESTAT` entry and the `- **An accepted solve no longer loads into Pyomo as a warning** (#591)` entry (~line 257). That section is unrelated to this PR — looks like an accidental drop during conflict resolution and is worth restoring.

CI was still in progress when I looked (WASM and Python/torch green, four jobs running). Nothing else from me — with the CHANGELOG whitespace tidied this looks good to me.

---

Issue #592, opened 2026-08-14
0.10.0: SolveSucceeded point improves by 25 s after two POUNCE restarts on the same NLP


bernalde:
## Summary

POUNCE 0.10.0 reports `SolveSucceeded` on a fixed-policy NLP from LyoPRONTO's Problem 2 GDP, but restarting POUNCE from the returned primal point improves the objective twice. The second restart reaches the same point as IPOPT.

This is not a comparison between two defensible local optima: POUNCE itself leaves its first reported optimum and improves it by **25.095976 s (0.079%)** on the identical fixed NLP.

Downstream integration/context:
https://github.com/SECQUOIA/LyoPRONTO/pull/147

## Environment

```text
pounce-solver 0.10.0
pyomo-pounce 0.10.0
Pyomo 6.9.5
IPOPT 3.14.16
Python 3.13.5
Linux x86_64
POUNCE linear solver: FERAL
```

## Model and options

The model is LyoPRONTO Paper Problem 2 with the deliberately small GDP test discretization:

```text
n_z=3
nfe_per_phase=2
ncp=2
phase_duration_weights=(3, 2, 1)
GDPopt algorithm=gdpopt.ric
master solver=glpk
NLP solver=pounce
init_algorithm=no_init
policy sequence=interface-velocity tracking -> max heat input -> temperature tracking
```

Every NLP solve uses:

```text
max_iter=3000
tol=1e-5
acceptable_tol=1e-3
print_level=0
```

The policy is fixed in the captured GDPopt subproblem, so the replay is a continuous NLP, not a mixed-integer re-solve.

## Reproduction

```bash
git clone https://github.com/SECQUOIA/LyoPRONTO
cd LyoPRONTO
gh pr checkout 147
python -m pip install -e ".[dev,pyomo]"
python -m pip install "pounce-solver==0.10.0" "pyomo-pounce==0.10.0"
sudo apt-get install glpk-utils
```

Then run:

```python
import pyomo.environ as pyo
import pyomo_pounce  # registers SolverFactory("pounce")

from lyopronto.pyomo_models.paper_gdp import (
    PaperGDPDiscretization,
    create_paper_problem2_gdp_model,
)

captured = []

def capture_first(_solver, subproblem, _util):
    if not captured:
        captured.append(subproblem.clone())

opts = {
    "max_iter": 3000,
    "tol": 1e-5,
    "acceptable_tol": 1e-3,
    "print_level": 0,
}

[...]

PR #596, opened 2026-08-14
fix(conv_check): stop judging a feasible point by a residual it cannot resolve

### LINKED ISSUE #590: 0.10.0 regression: LyoPRONTO Problem 1 f=0.02 returns infeasible after main@880b360 solved it
## Summary

The released `pounce-solver==0.10.0` wheel reports
`InfeasibleProblemDetected` on the Problem 1 `f=0.02` rung of LyoPRONTO's
pseudosteady-limit continuation. The same rung:

- converges with IPOPT 3.14.16;
- converged with POUNCE `main@880b360` during the verification of #505/#519;
  and
- has essentially the same physical endpoint in the 0.10.0 failed iterate.

This looks like a regression between `880b360` and the 0.10.0 wheel build at
`066949bf`, not a recurrence of the original #519 arming-condition defect:
0.10.0 now solves the earlier rungs that #519 fixed, but loses the `f=0.02`
rung.

Downstream integration/context:
https://github.com/SECQUOIA/LyoPRONTO/pull/147

## Environment

```text
pounce 0.10.0 (commit 066949bf, built 2026-08-11T19:25:22Z)
pounce-solver 0.10.0
pyomo-pounce 0.10.0
Pyomo 6.9.5
Python 3.13.5
target x86_64-unknown-linux-gnu
linear solver FERAL (MA57 disabled)
Linux 6.6.114.1-microsoft-standard-WSL2 x86_64
```

## Reproduction

```bash
git clone https://github.com/SECQUOIA/LyoPRONTO
cd LyoPRONTO
gh pr checkout 147
python -m pip install -e ".[dev,pyomo]"
python -m pip install "pounce-solver==0.10.0" "pyomo-pounce==0.10.0"

python -m examples.pseudosteady_limit_study \
  --solver ipopt \
  --solver-executable "$(command -v pounce)" \
  --output /tmp/pounce-0.10.0.json
```

The study warm-starts each rung from the preceding successful rung and stops at
the first failed rung.

## Observed

Problem 1 succeeds at `f = 1, 0.5, 0.2, 0.1, 0.05`, then stops:

```text
factor                       0.02
termination_condition        infeasible
solver_status                warning
solver_message               POUNCE 0.10.0: InfeasibleProblemDetected
loaded endpoint              6.148088152262362 h
max product temperature      243.0000024299997 K
max constraint violation     0.016207770073836655
```

## Comparisons

The committed IPOPT 3.14.16 baseline on the identical discretization, scaling,
options, and warm-start protocol reports:

```text
factor                       0.02
termination_condition        optimal
solver_status                ok
solver_message               Solved To Acceptable Level
endpoint                     6.148076261419493 h
max product temperature      243.0000009999997 K
max constraint violation     0.00805983833698369
```

The loaded POUNCE endpoint differs from IPOPT by only `1.19e-5 h`, although
POUNCE's final violation is larger and its verdict is infeasible.

More importantly, the independent verification posted on #505 for POUNCE
`main@880b360` reported this same rung as converged:

```text
f=0.02   endpoint 6.1481 h   acceptable
```

That run used the same continuation application and was the evidence that #519
fixed the whole Problem 1 ladder through `f=0.02`. The 0.10.0 release is 101
commits ahead of `880b360`.

## Expected

POUNCE 0.10.0 should retain the previously verified behavior at this rung:
converge to an acceptable solution near `6.1481 h`, or at least avoid a
confident infeasibility verdict for this known-feasible NLP.

## Regression-test suggestion

Add this rung (or its exported `.nl` plus the `f=0.05` warm start) as a release
regression tied to #505/#519. The useful assertion is not just the endpoint:
the solve must not return an AMPL/Pyomo infeasible-family status.

I can provide the generated `.nl` and warm-start data if that is more convenient
than regenerating them from the public application.

## Related

- #505 — original LyoPRONTO false-infeasibility report
- #519 — fix that made `main@880b360` walk this rung successfully
- https://github.com/SECQUOIA/LyoPRONTO/pull/147 — downstream
  convergence-quality plumbing and study context

jkitchin:
Fixes #590.

## What was reported

`pounce-solver 0.10.0` returned `Infeasible_Problem_Detected` on the Problem 1
`f = 0.02` rung of LyoPRONTO's pseudosteady-limit continuation study — a
known-feasible optimal-control problem that IPOPT 3.14.16 solves. Through
Pyomo that status is indistinguishable from a genuine infeasibility proof,
which makes it the most damaging verdict the solver can get wrong.

## What is actually happening

The model is written in Landau coordinates, so its conduction rows carry
`1/(H - S)^2` and reach magnitudes near `1e8`. Instrumenting the stall point:

```
scaled KKT error        4.29e-10     (tol = 1e-6)
unscaled constr_viol    1.62e-02     (constr_viol_tol = 1e-6)
relative violation      1.00e-08
rows above their floor  none
```

One ulp of those rows, passed through `||x||`, is `~1e-2`. That `1.62e-2` is
the quantum the rows are measured in, not a distance from feasibility. IPOPT
lands on the same point with `8.06e-3` and calls it `Solved To Acceptable
Level` — which side of `1e-2` a run falls on there is arithmetic luck, not a
property of the iterate.

gh #528 built the per-row noise floor for exactly this effect, but gave it only
to the strict **aggregate** and deliberately left the per-component
`constr_viol` test on the raw residual. Two consequences:

1. The gate became incoherent on its own terms. The component test is a
   refinement of the aggregate, so an unfloored component can veto a
   certificate the floored aggregate has already granted. That is precisely
   what happened here.
2. The rapid-infeasibility detector's absolute arm (`MIN_INFEAS_VIOL_FLOOR =
   1e-2`, an absolute constant in unscaled user units) was left free to convict
   on a quantum — the reported verdict.

## The fix

Both sites now consult the same `curr_primal_infeasibility_above_noise` floor,
in both cases **only when no row rises above it**: one resolvable row anywhere
and `constr_viol_tol` is back in charge. It is passed as a boolean rather than
a floored magnitude because the floor is computed on the internally scaled
residual while `constr_viol` is unscaled — comparing them as numbers is the
units error `infeasible_status_tol_invariance` exists to pin down, while the
*abstention verdict* is scale-invariant by construction and transfers cleanly.

A genuinely infeasible model is untouched: its violation is pinned at its
infeasibility gap, orders above `eps *` the row's own magnitude. The
scale-relative veto still sees the raw `rel_viol`, and
`primal_noise_floor_kappa = 0` opts out of both gates as it already did for the
aggregate.

## Evidence

**The reported rung** — `Optimal Solution Found` in 20 iterations, endpoint
6.14809 h against the issue's expected 6.1481 h. Before: 0.10.0 gave
`Infeasible_Problem_Detected`; `main` gave a 10000-iteration
`Maximum_Iterations_Exceeded`. IPOPT needs 58 iterations to reach only
"acceptable".

**The full ladder** now walks past the reported failure to `f = 0.01`, every
rung `converged_to_tolerance`:

```
f=1      endpoint=6.1865 hr   converged_to_tolerance
f=0.5    endpoint=6.1669 hr   converged_to_tolerance
f=0.2    endpoint=6.1552 hr   converged_to_tolerance
f=0.1    endpoint=6.1512 hr   converged_to_tolerance
f=0.05   endpoint=6.1493 hr   converged_to_tolerance
f=0.02   endpoint=6.1481 hr   converged_to_tolerance   <- the reported rung
f=0.01   endpoint=6.1477 hr   converged_to_tolerance
```

**Direction check on the gh #528 LP family**, extended to the data scales where
the quantum clears the issue's `constr_viol_tol = 1e-6` (scales 1e9-1e14, two
sizes, six seeds): the change moved **23 verdicts and every one of them
improved** — `Solved_To_Acceptable_Level`, `Search_Direction_Becomes_Too_Small`
and `Maximum_Iterations_Exceeded` all becoming `Solve_Succeeded`. No verdict
was degraded anywhere; the pre-existing `Restoration_Failed` cases are
identical before and after.

## Correction to the issue's framing

The issue attributes this to a regression between `main@880b360` and the 0.10.0
wheel. I built `880b360` in a worktree and ran the identical `.nl`: it also
fails to converge this rung, cycling indefinitely (reached iteration 5791
before I killed it). The *stall* is not new in 0.10.0 — only the emitted
non-success status differs. The underlying defect predates both.

## Tests

`crates/pounce-algorithm/tests/issue_590_primal_noise_floor_component.rs`, tied
to #505/#519/#528 as the issue asks. It asserts the invariant rather than an
endpoint, per the issue's own framing ("the solve must not return an
AMPL/Pyomo infeasible-family status"):

- `a_residual_below_every_row_floor_does_not_cost_the_certificate` — the grid
  certifies, and the deviation from the sigma=1 optimum is the *same* at every
  data scale to ten significant figures (fails without this change)
- `a_violation_just_above_the_quantum_is_still_refused` — one decade above the
  floor is still a violation the solver must see
- `a_genuinely_infeasible_large_scale_model_is_still_refused`
- `switching_the_floor_off_loses_the_certificates_again` — the opt-out reaches
  both new gates

Plus unit tests for the two new predicates, and the standing
`infeasible_models_are_never_reported_solved` guard still passes.

https://claude.ai/code/session_01RLnYvbmau12Ep9LzZUwahm


jkitchin:
@bernalde — would you take a look? This is the fix for #590, and since you have
the LyoPRONTO study set up I would value your read on two things in particular.

**Does the rung actually look right to you.** It exits `Optimal Solution Found`
in 20 iterations at endpoint 6.14809 h, and the ladder now walks one rung past
the reported failure to f = 0.01. But "converged" and "the answer you expect"
are different claims, and you are better placed than I am to judge the second.

**One correction to the issue's framing**, flagged because it changes what the
regression test is guarding. I built `880b360` in a worktree and ran the
identical `.nl` — it also fails to converge this rung, cycling indefinitely
(iteration 5791 before I killed it). So the stall is not new in 0.10.0; only the
emitted non-success status differs. The invariant you asked for is still the
right assertion, it just isn't guarding against a recent behavioural change. If
you saw `880b360` genuinely solve this rung, then I have reproduced the wrong
configuration and would like to know.

Short version of the diagnosis: at the stall point every constraint row's
residual sat at or below its own floating-point resolution floor. The Landau
coordinates put `1/(H - S)^2` in the conduction rows, which reach ~1e8, so one
ulp there is ~1e-2 — the 1.62e-2 violation is the quantum those rows are
measured in, not a distance from feasibility. IPOPT lands on the same point at
8.06e-3 and calls it acceptable; which side of 1e-2 you land on is arithmetic
luck. The per-row noise floor to detect this already existed (#528) but had been
scoped to the aggregate test only, leaving the infeasibility detector free to
convict on a quantum.

bernalde:
Thanks, John. I tested the exact PR head
`316ce56a879903688c46963167cb670807f5ee6f` in LyoPRONTO. Direct answers to
the two questions above:

**Yes, the repaired rung looks right.** The complete continuation now finishes
for both applications:

```text
Problem 1: f = 1, .5, .2, .1, .05, .02, .01 -> all converged_to_tolerance
Problem 2: f = 1, .5, .2, .1, .05, .02, .01 -> all converged_to_tolerance

Problem 1 f=.02 endpoint: 6.1481 h
Problem 1 f=.01 endpoint: 6.1477 h
Problem 2 f=.02 endpoint: 8.8281 h
Problem 2 f=.01 endpoint: 8.8280 h
```

At the reported Problem 1 rung, the #596 endpoint is within about `0.043 s` of
the IPOPT endpoint and remains on the same active 243 K product-temperature
ceiling. The next rung is smooth rather than a discontinuous escape. For every
rung that already succeeded with released 0.10.0, the maximum endpoint change
from this patch was only `8.1e-8 s` in Problem 1 and `6.6e-9 s` in Problem 2.
That is strong downstream evidence that this is correcting the certificate,
not moving the physical solution.

**I accept the correction about `880b360`.** My earlier evidence established
that the application had once been reported as walking through this rung, but
your identical-`.nl` replay is the better attribution test. The invariant worth
guarding is therefore: a feasible point whose rows are all at their numerical
resolution floor must not receive a confident infeasibility-family verdict.
This PR tests and repairs that invariant; it should not be described as a new
0.10.0 convergence stall.

The targeted Rust regression passed `4/4`. The complete downstream POUNCE lane
against this exact head was:

```text
179 passed, 4 failed, 1 xfailed, 418 deselected
```

The four failures were expectation-only: #596 upgraded four existing solves
from `accepted_at_acceptable_tol` to `converged_to_tolerance`. None was an
application failure. The xfail was the separate GDP restart issue #592.

There is one new integration detail since that run: #594, which fixes #592,
has now merged to `main`, while this PR is currently conflicting and has no CI
attached. I independently validated merged #594: the full LyoPRONTO lane was
`183 passed` plus only the intended strict XPASS for #592. After #596 is
rebased/resolved onto current `main`, the combined head should be rerun because
both changes affect general NLP convergence trajectories. If that combined
lane retains #596's full ladders and #594's restart-stable GDP point, the two
known LyoPRONTO defects are resolved together.

The downstream comparison/CI PR remains release-pinned to 0.10.0 and will keep
the recorded failures until a release containing the fixes is published:
https://github.com/SECQUOIA/LyoPRONTO/pull/148

jkitchin:
## Combined-head verification, and a direct check against #590 as filed

Resolved onto current `main` (CHANGELOG only; `upstream_options.rs` auto-merged).
Since #594 is already on `main`, this head **is** the combined lane, so I ran both
defects against it rather than only this PR's own evidence.

### 1. The reported case, exactly as filed

Not a proxy — the `f = 0.02` rung captured from the study, with the issue's own
option set (`max_iter=10000 tol=1e-6 acceptable_tol=1e-3 constr_viol_tol=1e-6
mu_strategy=adaptive bound_relax_factor=1e-8 nlp_scaling_method=gradient-based`):

| | verdict | iters | wall | unscaled `constr_viol` |
|---|---|---|---|---|
| `main` | `Maximum_Iterations_Exceeded` | 2000 (capped) | 26.5 s | 1.46e-02 |
| **this head** | **`Solve_Succeeded`** | **20** | **0.31 s** | 9.50e-03 |

Endpoint `6.148088153 h`, against the issue's own loaded value of
`6.148088152262362 h` — agreement to nine decimals. That is the point of the
diagnosis: 0.10.0 was already standing on the answer and refused to certify it.
IPOPT 3.14.16 reaches `6.148076261419493 h` (delta `1.19e-05 h`) and calls it only
*acceptable*, with `8.06e-03` violation — so this head's `9.50e-03` sits in the same
band as the reference solver, while the scaled residual is `9.50e-11` against
`tol = 1e-6`.

### 2. The full ladder, every rung certified

| rung | verdict | iters | endpoint |
|---|---|---|---|
| f=1 | `Solve_Succeeded` | 23 | 6.18654 h |
| f=0.5 | `Solve_Succeeded` | 15 | 6.16690 h |
| f=0.2 | `Solve_Succeeded` | 20 | 6.15516 h |
| f=0.1 | `Solve_Succeeded` | 19 | 6.15124 h |
| f=0.05 | `Solve_Succeeded` | 33 | 6.14927 h |
| **f=0.02** | **`Solve_Succeeded`** | **20** | **6.14809 h** ← the reported rung |

### 3. #594's restart-stable GDP point is retained

The three fixed-policy restarts from #592, run on this head and on plain `main`:

```text
                    combined head            plain main
8_restart1    3.1785744276216461e+04   3.1785744276216461e+04
9_restart2    3.1785744254727051e+04   3.1785744254727051e+04
10_restart3   3.1785744273618799e+04   3.1785744273618799e+04
```

**Byte-identical.** Objective range `2.15e-05 s`, against the released `25.095976 s`
drift. So this change is trajectory-neutral on #592's model, and the two LyoPRONTO
defects are resolved together on one head rather than trading off.

### Against the issue's stated acceptance bar

#590 asks for "converge to an acceptable solution near `6.1481 h`, **or at least
avoid a confident infeasibility verdict**." Both are met, the stronger one: an
`Optimal Solution Found` at `6.14809 h`, not merely a non-infeasible status.

The issue also suggests a regression tied to #505/#519. That is
`crates/pounce-algorithm/tests/issue_590_primal_noise_floor_component.rs`. Worth being
explicit that it pins the **invariant** on a scale-equivariant family rather than
shipping the LyoPRONTO rung itself, because the issue's own framing asks for exactly
that ("the useful assertion is not just the endpoint"). The rung above is the
end-to-end evidence; the committed test is what keeps it from regressing without
carrying a 1.4 MB application model into the repo.

### Downstream

The comparison/CI PR (SECQUOIA/LyoPRONTO#148) stays release-pinned to 0.10.0 and will
keep its recorded failures — including the strict `xfail` on the #592 regression that
now XPASSes — until a release containing both fixes is published. That is the trigger
for removing it, not this merge.

PR #597, opened 2026-08-14
Map every engine exit in the Pyomo and GAMS status tables (#589)

### LINKED ISSUE #589: pyomo-pounce: a restoration failure raises NoSolutionError on the v2 route and returns on the legacy one
`_V2_STATUS` (v2.py:186) and `_STATUS_RESULT` (sens.py:430) each map nine engine exits. `Restoration_Failed` is not among them, so it takes the default at v2.py:592:

```python
tc, ss = _V2_STATUS.get(
    capture.get("status_msg", ""),
    (TerminationCondition.error, SolutionStatus.noSolution))
```

`noSolution` then reaches the load block, which is a line-for-line copy of Pyomo's own `Ipopt.solve`:

```python
has_solution = ss is not SolutionStatus.noSolution
...
if config.load_solutions:
    if not has_solution:
        raise NoSolutionError()
```

So the same failed solve returns a results object under `SolverFactory("pounce")` and raises `NoSolutionError` under `SolverFactory("pounce_v2")`.

## Why the default is the wrong answer here

The comment above `_V2_STATUS` already settles this. It records that the solution-status column deliberately matches what the `.sol` route produces "because disagreeing changes whether a solve raises", and that mapping the limit cases to `noSolution`, which the table did at first, was wrong for a stated reason:

> The in-process route always has a primal iterate -- the engine returns `x` regardless of status -- so they map to `unknown` here too. Mapping them to `noSolution` (as this table first did) made `solve(m, solver_options={'max_iter': 1}, raise_exception_on_nonoptimal_result=False)` load the final iterate on the ordinary route and raise `NoSolutionError` on this one, for the same model and options; `noSolution` was also simply the less accurate answer, since POUNCE does return a usable iterate in both cases.

`Restoration_Failed` has the same property. It is a real exit code (`return_codes.rs:24`, `-2`), the engine returns an iterate with it, and `sens_solve` captures that iterate before the non-converged early return, so the loader is holding a point it declines to hand over. The exit was simply never added to either table, so it inherits a default the reasoning above rejects.

Pyomo's own `ipopt_v2` returns on a comparable failure rather than raising.

## Reproducing

Any model that exits restoration, solved through both names. Observed on a 16,217-variable model where both routes run identically and both exit `Restoration Failed`:

```
SolverFactory("pounce")     -> results.solver.termination_condition = error
SolverFactory("pounce_v2")  -> NoSolutionError raised from v2.py:622
```

## Suggested fix

Add `Restoration_Failed` to both tables. `unknown` on the v2 side matches the rule the comment states, since an iterate exists. The legacy side already behaves correctly through its `(error, error)` default, so an explicit entry there is for symmetry rather than to change behavior.

Worth checking at the same time whether any other engine exit is missing from the tables, since the failure mode is silent: an unmapped exit that returns an iterate becomes a raise on one route only.

## Impact

Every caller on the v2 route loses the iterate on a restoration failure, and gets an exception where the legacy route gives a results object they can inspect. drto is affected on every solve, because its `dynamic_optimization` transform declares sensitivity parameters, which routes all of its models through `_sens_solve`, and its `scaled_solve` now defaults to `pounce_v2`.

jkitchin:
Fixes #589

@devin-griff — could you take a look? You filed this, and the third commit goes past what the issue asked for: it changes what the GAMS links report for several exits, and corrects three GAMS constants in the pip link. Your read on the GAMS side especially would be welcome.

> Reviewed below by @devin-griff; two claims in this description were wrong and have been corrected in place (severity, and `.sol` parity). A CI failure from the same area is fixed in `5994fc7`.

## Problem

`Restoration_Failed` was missing from both `pyomo_pounce` status tables, so it took their defaults. On the legacy route that default is a results object; on the v2 route it was `SolutionStatus.noSolution`, which is not a severity — it turns the solution loader off and raises `NoSolutionError` under the default `load_solutions=True`.

| | `SolverFactory("pounce")` | `SolverFactory("pounce_v2")` |
|---|---|---|
| before | `termination_condition = error`, results object | `NoSolutionError` |
| after | `termination_condition = internalSolverError`, results object | `error` / `unknown`, iterate loaded |
| ordinary `.sol` route (the oracle) | `internalSolverError`, `status = error` | `error` / `unknown` |

The issue asked whether other exits were missing too. Eleven of twenty were — on both routes. Checking the GAMS links against the same enum found the same class of gap there, plus one of their own:

| | modelStat | solveStat | objective row |
|---|---|---|---|
| `Restoration_Failed`, before | 6 | 10 (pip) / 4 (C) | not set — `x.l` published with `0` beside it |
| `Restoration_Failed`, after | 6 | 4 | objective at the iterate |
| `Insufficient_Memory`, before | 13 | 12 (pip) / 11 (C), via `default` | — |
| `Insufficient_Memory`, after | 13 | 10 | — |

## Cause

Three separate things, one shape.

**The tables were never exhaustive.** Nothing checks a status table against the enum it mirrors, and an unlisted exit is silent: no warning, no log line, no failing test. It just takes the default. The comment above `_V2_STATUS` had already settled the principle when the limit exits hit this same bug — the solution-status column follows the `.sol` route's rule, "a primal vector came back", and the in-process route always has one — so `Restoration_Failed` was not decided to be a no-solution case, it was never added.

**`pounce_status_has_solution` gates less than its name suggests.** `gmoSetSolution2` hands GAMS the primal vector and the marginals for every exit, because the engine always fills them. So leaving a status out of that set does not withhold the point — it publishes the point with an objective of `0` beside it, which is worse than either alternative.

**The pip GAMS link's `gmoSolveStat_*` constants were wrong.** `SOLVESTAT_EVAL_ERR` was `11` (`gmoSolveStat_InternalErr`) and `SOLVESTAT_INTERNAL_ERR` was `12` (`gmoSolveStat_Skipped`); four exits used `gmoSolveStat_SolverErr` (10, "Solver Failure") where the C link uses `gmoSolveStat_Solver` (4, "Terminated By Solver") — a verdict, not a crash. The two links are supposed to report identically and disagreed on four statuses. This one is invisible without GAMS in the loop: the link runs, the value is in range, and the `.lst` says the wrong thing.

## Fix

**Pyomo.** Both tables now cover all twenty exits, and nothing maps to `noSolution`. The v2 fallback for an unrecognised status is `unknown` too — that lookup moved into `_v2_status()` so it is directly testable. A future engine exit therefore cannot flip whether the route raises; the coverage test catches it for the termination condition, where the cost of an oversight is precision rather than an exception.

The alternative was to classify exit by exit — `unknown` where the engine reached `finalize_solution`, `noSolution` for the pre-solve refusals where `x` is the zero vector it was initialised with. Rejected: the ordinary route reports `unknown` for those too (`crates/pounce-cli/src/main.rs` writes the `.sol` unconditionally "even on a failed solve", falling back to zero blocks sized from the pre-solve NLP dimensions so the primal block still aligns), and `sens_solve` has already written those values onto the model by the time the v2 route inspects the status — so raising would not spare the caller the zeros, only the results object describing them. `raise_exception_on_nonoptimal_result=True`, the default, still raises `NoOptimalSolutionError` on every one of them.

**GAMS.** Both links cover the enum. `Restoration_Failed` and `Invalid_Number_Detected` join the objective-reporting set, guarded on `isfinite(obj_val)` — which is what makes widening it safe, since POUNCE leaves the objective at NaN when it refused the solve and `Invalid_Number_Detected` is by definition an exit where a value went non-finite. `Diverging_Iterates` stays out (the objective is a step along a ray, and `gmoModelStat_Unbounded` already carries the finding); `Insufficient_Memory` stays out (the link should not vouch for the state of a solve the machine killed).

## Blast radius

**Pyomo, v2 route:** a solve that raised `NoSolutionError` now returns. Nothing that previously returned changes its solution status.

**Pyomo, legacy route:** the eleven added exits reported plain `TerminationCondition.error` and now agree with the severity the `.sol` route gives the same solve, while naming the outcome more precisely than AMPL's bands can:

| | count | termination condition | `.sol` route says |
|---|---|---|---|
| 500 failure band | 8 | `internalSolverError` | same |
| the two definition errors | 2 | `invalidProblem` | `internalSolverError` (the band cannot distinguish them) |
| `Search_Direction_Becomes_Too_Small` | 1 | `minStepLength` | `maxIterations` (400 band, "this is not always correct" in Pyomo's own source) |

**One of the eleven changes severity**, corrected from an earlier claim in this description that none did: `Search_Direction_Becomes_Too_Small` is now `SolverStatus.warning` where the default gave `error`, matching the 400 limit band — a stalled solve is a limit case, not a failure. A legacy caller branching on `status == error` for that exit will see `warning`. The other ten stay `error`; verified by enumerating the table.

This whole column is more than #589 asked for — the issue expected the legacy entries to be symmetry-only. It is the file's own stated principle applied to rows being added anyway. Easy to drop if you would rather the legacy side stay bit-identical; that would mean all eleven at plain `error`.

**GAMS:** `solveStat` changes for six exits on the pip link and three on the C link; the objective row appears for two exits that previously showed `0`. Both links now produce identical `(modelStat, solveStat, hasSolution)` triples for all twenty exits — verified by compiling `map_status_to_gams` / `pounce_status_has_solution` standalone under `-Wall -Wextra -Werror` and diffing the output against the Python table, and independently in review by parsing the C switch and resolving every enumerator through `gams.core.gmo`.

`drto` is the caller most affected: its `dynamic_optimization` transform declares sensitivity parameters, routing every model through `_sens_solve`, and `scaled_solve` defaults to `pounce_v2`.

No Rust changed, so no corpus sweep and no `cargo` run — `crates/pounce-nlp/src/return_codes.rs` is read by the new tests but not modified.

## Tests

`pyomo-pounce/tests/test_issue_589_status_table_coverage.py` (new, 8 tests) and the status half of `python/tests/test_gams_link.py` (rewritten). Both hold their tables to the names in `crates/pounce-nlp/src/return_codes.rs`, parsed from the checkout, so a status added there fails the suite until the links are updated.

They bite on the parent commit, for the stated reasons:

- Pyomo: all 8 fail; the rest of that suite is unchanged before and after.
- GAMS coverage: `['Unrecoverable_Exception', 'NonIpopt_Exception_Thrown', 'Insufficient_Memory'] fall to the status_to_gams default`.
- GAMS mapping: `(3, 10) == (3, 4)`, `(6, 11) == (6, 5)`, `(13, 12) == (13, 11)` — the constant bugs, each as a concrete mismatch.
- The `reports_objective` tests fail on the parent because the guard function is new, not because of a value mismatch. Stating that plainly: those are new-API failures, and the mapping tests above are what pins the behaviour change.

One existing test changed. `test_nonconvergence_returns_mapped_results` builds one free variable against two equalities plus the declared-parameter pin row, so it exits `Not_Enough_Degrees_Of_Freedom` rather than reaching an infeasibility verdict — an exit that had no entry and fell to the `(error, error)` default, which is why `error` satisfied it. Its accepted set is widened rather than the mapping reverted: the test's own premise is "the mapped termination (like the ordinary `.sol` path)", and an over-determined model is not a solver failure. This was a CI failure on the first push, not something noticed by reading.

The old nine-row `test_status_to_gams` is deleted, not extended. It compared the table against the module's own constants, which cannot fail whatever the constants say — that is how three wrong ones survived in a table the suite called covered. Its replacement checks all twenty exits against literal GAMS integers taken from the C link.

`test_status_constants_match_gamsapi` checks every constant against GAMS's own `gams.core.gmo`. That needs `gamsapi[core]` — pure Python, no GAMS install and no license — so CI installs it rather than letting the guard skip; a check that never runs would not have caught this.

Not pinned: the live `gamsapi` adapter (`_write_solution`) still has no test, as before. The finiteness guard is reachable through `reports_objective`, which is why it was factored out.

---

- [x] Tests fail on the parent commit for the stated reason (with the `reports_objective` caveat above)
- [x] Book page updated — `docs/src/gams.md` gains a "Status mapping" section; no new page, so no `SUMMARY.md` change
- [ ] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean — not run; no Rust changed. The C link was syntax- and behaviour-checked with `gcc -Wall -Wextra -Werror`.
- [x] Every claim in this PR body and in the code comments is true of the code as it stands now — two were not, and are corrected above rather than quietly edited away

https://claude.ai/code/session_01UBpnrmUH5HJDdMpdmGYY3x


devin-griff:
The GAMS side is correct, and I verified the parts a reader cannot check by inspection. Two claims in the description need correcting, and I would merge after that.

## GAMS

I installed `gamsapi[core]` and read the constants from `gams.core.gmo`. All three corrections are right, and your account of what the wrong values meant is exact. `gmoSolveStat_Solver` is 4, `gmoSolveStat_EvalError` is 5, `gmoSolveStat_InternalErr` is 11, and `gmoSolveStat_Skipped` is 12, so the old `SOLVESTAT_EVAL_ERR = 11` was `InternalErr` and the old `SOLVESTAT_INTERNAL_ERR = 12` was `Skipped`.

`_C_LINK_STATUS` is the part nothing else pins, because it is transcribed by hand and the pip link's integers have no other source in the repo. I checked it mechanically rather than by eye. I parsed the `map_status_to_gams` switch out of `gams/gams_pounce.c`, resolved every `gmoModelStat_*` and `gmoSolveStat_*` name through `gams.core.gmo`, and compared. The switch has twenty cases, the table transcribes twenty, and every pair is identical. `_C_LINK_REPORTS_OBJECTIVE` matches `pounce_status_has_solution` the same way, twelve for twelve, and the two links' sets agree.

Writing those as literal integers instead of `link.SOLVESTAT_*` is what lets the check catch a wrong constant. A table compared against its own module's constants cannot catch a constant that is itself wrong, which is how three of them survived.

The tests fail when they should. Dropping `Restoration_Failed` from both tables produces four failures, the two coverage tests plus `test_status_to_gams_matches_the_c_link[Restoration_Failed]` and `test_restoration_failure_reports_a_loadable_iterate`.

On `Error_In_Step_Computation` taking `gmoSolveStat_SolverErr` where `Search_Direction_Becomes_Too_Small` takes `gmoSolveStat_Solver`, I think that is right. A step that could not be computed is a failure. A step that became too small is a verdict.

## Two claims to correct

**Severity does change, for one exit.** "SolverStatus is unchanged from what the default gave, so anything branching on severity is unaffected" holds for ten of the eleven. `Search_Direction_Becomes_Too_Small` is `(minStepLength, SolverStatus.warning)`, and the default gave `SolverStatus.error`. A legacy caller branching on `status == error` for a stalled solve now sees `warning`. The value is right and only the sentence needs changing. `warning` is what the `.sol` route gives for AMPL's 400 band, so the entry follows this file's principle better than the default did.

**Three of the eleven do not report what the `.sol` route reports.** I ran each status's `status_to_solve_result_num` code through Pyomo's legacy `.sol` reader in `pyomo/opt/plugins/sol.py`:

| status | code | `.sol` route | `_STATUS_RESULT` |
|---|---|---|---|
| `Search_Direction_Becomes_Too_Small` | 400 | `maxIterations` / warning | `minStepLength` / warning |
| `Not_Enough_Degrees_Of_Freedom` | 504 | `internalSolverError` / error | `invalidProblem` / error |
| `Invalid_Problem_Definition` | 504 | `internalSolverError` / error | `invalidProblem` / error |

The severity agrees in all three, and the termination condition is finer than AMPL's bands can express. I would keep the values and restate the claim as agreeing with the band's severity while being more precise than its termination condition.

## The Pyomo choice is right, and better than what I suggested

My comment on #589 recommended leaving the eight exits returned before the algorithm starts on `noSolution`, and that was wrong. `sens_solve` writes `x` onto the model in the non-converged early return before the v2 route inspects the status, so raising `NoSolutionError` denies the caller the results object while leaving the zeros on their model. Mapping every exit to `unknown` is the better answer, and `raise_exception_on_nonoptimal_result` still raises by default.

## Minor

- `Invalid_Option` cannot reach either table through `sens_solve`. I measured this while writing the issue comment, and the binding raises `RuntimeError` before a status is captured. The entry is still needed for exhaustiveness and by the coverage test, so this is a note rather than a request.
- I could not confirm that the CLI writes a zero-filled `.sol` for a refused solve. The write at `main.rs:1970` is not gated on status, which is the part the argument needs, but I did not read the primal block's contents.

All 85 tests pass here, and CI installing `gamsapi[core]` means `test_status_constants_match_gamsapi` runs rather than skipping.

jkitchin:
Thanks — both corrections are right, and the mechanical check of `_C_LINK_STATUS` is the one I wanted and could not give myself. That table is hand-transcribed and nothing else in the repo pins the pip link's integers; parsing the switch and resolving the enumerators through `gams.core.gmo` is a stronger check than the one I ran (compiling the two functions standalone and diffing their output against the Python table). Twenty for twenty and twelve for twelve agrees with what I measured.

Fixed in 5994fc7, along with a CI failure from the same area.

**Severity.** You are right, and the enumeration confirms it: `Search_Direction_Becomes_Too_Small` is the only one of the eleven that moves, `error` → `warning`. The description now says so explicitly rather than claiming none of them do, and the CHANGELOG calls out that a legacy caller branching on `status == error` for that exit will see `warning`. I also rewrote the comment above the entry so the reason is on the row itself: it sits in the 400 limit band because a stalled solve is a limit case, not a failure, and `warning` is that band's severity.

**The three that deviate.** Restated as you suggested — severity parity with a termination condition finer than AMPL's bands can express — and the description now carries the 8 / 2 / 1 split as a table rather than a sentence that reads as "all eleven match `.sol`" and then lists two that do not. The `_STATUS_RESULT` comments got the same treatment: the `invalidProblem` rows now say outright that they deviate from `internalSolverError` and why, instead of only noting that the severity is unchanged.

**CI caught a third thing, and it is the same one.** `test_nonconvergence_returns_mapped_results` failed on the first push. Its model is one free variable against two equalities plus the declared-parameter pin row, so it exits `Not_Enough_Degrees_Of_Freedom` — not an infeasibility verdict, which is what the name suggests. That exit fell to the `(error, error)` default, which is why `error` was enough for its accepted set; it now reports `invalidProblem`.

Worth saying which way I resolved it, since it is exactly the trade-off you flagged. I widened the accepted set rather than reverting to `internalSolverError`, on two grounds: the test's own premise is "the mapped termination (like the ordinary `.sol` path)", and reverting would not have satisfied it either — `internalSolverError` is outside that set too, so any entry at all breaks it. The set was written when the default was the only reachable value. If you would rather the legacy column stay bit-identical, the change to make is all eleven at plain `error`, and I would rather hear that now than have it found later.

**Your two minor notes.** Both stand, and I acted on the second.

`Invalid_Option` not reaching either table through `sens_solve` — agreed, and it matches what I read in the binding: the option is rejected at `add_option` time, before a solve produces a status. Left in for exhaustiveness, as you say. The `.sol` route can still reach it, which is the other reason to keep the row honest.

The zero-filled `.sol` — this is the load-bearing claim under "`noSolution` would buy nothing", so it should not have rested on an ungated write. It is confirmed at `crates/pounce-cli/src/main.rs`, and the code says it in its own comment: the file is written "unconditionally once a target path is resolved — even on a failed solve", and "when the solve never converged the capture is empty; fall back to zero blocks sized from the pre-solve NLP dimensions so the file still round-trips", with the `m` sized to the original `.nl` so a failed-solve `.sol` still aligns. The `v2.py` comment now cites that location instead of asserting the behaviour, so the next reader can check it in one hop.

On `Error_In_Step_Computation` versus `Search_Direction_Becomes_Too_Small` — that is exactly the distinction I was drawing, and better put than my comment put it. A step that could not be computed is a failure; a step that became too small is a verdict.

---

jkitchin:
Merging this. Recording the one decision it makes that nobody explicitly
asked for, so it is on the record rather than buried in the diff.

**Merging accepts the more precise legacy mapping, including one severity
change.** The eleven exits added to `sens._STATUS_RESULT` go beyond what #589
asked for — the issue expected the legacy entries to be symmetry-only, and
they instead follow the file's own stated principle and name the outcome as
precisely as the row allows. The consequence @devin-griff identified:

> `Search_Direction_Becomes_Too_Small` is `(minStepLength, SolverStatus.warning)`,
> and the default gave `SolverStatus.error`. A legacy caller branching on
> `status == error` for a stalled solve now sees `warning`.

That is the only one of the eleven that moves severity; the other ten stay
`error`, verified by enumerating the table. The new value is the more correct
one — `warning` is what the `.sol` route gives for AMPL's 400 limit band, and a
stalled solve is a limit case rather than a failure — but "more correct" is not
the same as "no change", and a legacy caller keying on severity for that exit
will behave differently after this merge.

**If you would rather the legacy side stay bit-identical**, the change is
small and self-contained: all eleven added rows at plain
`TerminationCondition.error`, which is what the default gave them. That keeps
the v2 fix (the actual #589 defect — `NoSolutionError` on a route that had a
point in hand) and drops only the legacy precision. Say so and I will open it
as a follow-up rather than reverting this; the GAMS half and the coverage tests
are independent of that choice and should stay either way.

Three of the eleven also report a termination condition finer than the `.sol`
route's bands can express (`minStepLength` where the band says `maxIterations`;
`invalidProblem` twice where the band says `internalSolverError`). Severity
agrees in all three, and this is deliberate — recorded here because "agrees
with the `.sol` route" is true of the severity column and not of the
termination column, and the distinction is easy to lose later.

Everything @devin-griff raised is addressed: both description corrections in
`5994fc7`, the `v2.py` comment now cites `crates/pounce-cli/src/main.rs` for
the unconditional `.sol` write instead of asserting it, and `Invalid_Option`
stays in for exhaustiveness as agreed. The `_C_LINK_STATUS` check they ran —
parsing the C switch and resolving every enumerator through `gams.core.gmo`,
twenty for twenty and twelve for twelve — is a stronger verification than the
standalone-compile diff I ran, and is the reason the GAMS half is trustworthy.

Conflict with `main` was CHANGELOG-only: this PR and #594 (the #592 fix) each
added an entry at the top of `## [Unreleased]`. Purely additive, both kept.
`crates/pounce-nlp/src/return_codes.rs` — which the new tests parse — is
untouched by that merge, so the enum these tables are held to is unchanged.

Issue #434, opened 2026-08-01
Homotopy is pure cost when the active set is a large fraction of n — needs a runtime guard, not a predictor


jkitchin:
Follow-up from #413, with independent corroboration from the warm-start benchmark added in #431. Filed as a tracking issue rather than a PR because the discriminating feature is not yet known — and the obvious candidate is already refuted below.

## The problem

The parametric homotopy (#412, on by default in the convex QP driver, `active_set.rs` `use_homotopy: true`) is a net win but not a uniform one. On the Maros-Mészáros active-set column at a 120 s cap it takes **71/138 against 58/138** for the conventional phase-1/phase-2 path — 20 gained, 7 lost. On the losses the path is not wrong, it is **pure cost**: it burns the entire budget without reaching `t = 1`.

| instance | path steps | reached `t=1`? |
|---|--:|---|
| POWELL20 | 7250 | ✗ (t = 0.9997) |
| LISWET1 | 7050 | ✗ (t = 0.976) |
| DTOC3 | 4950 | ✗ (t = 0.995) |
| AUG2D | 4100 | ✗ (t = 0.50) |
| QSHIP08L | 4100 | ✗ (t = 0.655) |

Six instances that solved before the homotopy now hit the cap: `AUG2D`, `AUG2DC`, `CONT-050`, `CONT-100`, `DTOC3`, `STADAT3`. `sqp_qp_use_homotopy=no` recovers them, so the information needed to solve them is available — the driver just has no way to decide when to use it.

## Independent corroboration

The warm-start suite (`benchmarks/warmstart`) runs `-hom` arms differing from their twins by exactly that one option. Cold inner-QP active-set changes:

| family | conventional → homotopy | ratio | mean \|A\| / n |
|---|--:|--:|--:|
| `simplex_proj` | 328 → 523 | 0.63× | 16.4/20 = 82% |
| `rosenbrock_ring` | 29 → 29 | 1.00× | 0.5/10 = 5% |
| `nmpc_vanderpol` | 370 → 1140 | **0.32×** | 46.4/47 = 99% |
| **total** | **727 → 1692** | **0.43×** | |

[...]

PR #614, opened 2026-08-15
qp: keep the working set when solve_parametric declines, and let the homotopy add a bound (#602)

### LINKED ISSUE #602: Improve `solve_parametric` eligibility
I think we could improve `solve_parametric` eligibility. Currently it verifies only shape and Hessian equality at `/pounce-qp/src/solver.rs`, even though the homotopy interpolates only the gradient and row bounds.

I think it should also require unchanged:
- Constraint matrix structure and values.
- Variable bounds (unless box-bound interpolation is implemented?)
- Hessian inertia declaration.

Otherwise Pounce may spend time following a path whose initial point is not on the claimed previous problem manifold.
Opening this as an issue to see if we can improve based on these conditions.

jkitchin:
> [!NOTE]
> **The gating Maros-Mészáros sweep has been run — it passes on correctness.**
> 57/138 correct on both arms (the same 57), zero solved-but-wrong, no instance
> correct on one arm only in either direction. Cost is +9.5% iterations, and
> ablation attributes every loss to the bound-adding ratio test — the
> correctness fix paying for itself — and every gain to the reconciliation.
> [Full results and method](https://github.com/jkitchin/pounce/pull/614#issuecomment-5305253101);
> the one outsized instance (`CVXQP2_M`, ×5.42) is filed as #615 with attribution
> already done. Note the sweep used a substituted instrument: the documented
> `homotopy_sweep` example cannot run anywhere, because per
> `dev-notes/issue-434-homotopy-cost.md:280-282` its `.qp` corpus and the `.mat`
> converter were never committed, so the 138-problem `.nl` corpus went through
> the CLI with `solver_selection=qp-active-set`, which enters the same
> `use_homotopy: true` driver.

Fixes #602.

## Problem

#602 observed that `solve_parametric` admits the §4.2 homotopy on a guard
checking only shape and `H`, while the path interpolates only `g` and the row
bounds — so `A`, `xl`/`xu` and `hessian_inertia` can all differ with the path
modelling the difference not at all, and it proposed guarding on all three.

Measuring that turned up three things, only one of which was the thing asked
for:

| | finding |
|---|---|
| the framing | a **cost** question for the three quantities named — all 17 measured routes agree with a cold solve to `1e-14`, because `trace_path` never reports its own point and the corrector re-solves under `audit_and_repair` |
| the real waste | not the guard. The *declined* branch cold-solved, discarding the working set the caller had just handed over — `xu = 0.5` cost 241 ms against 12 ms for simply reusing it |
| a missing algorithm piece | the tracer had **no bound-adding event at all**. Bounds could be dropped, never added, so `x(t)` walked out of the box with nothing capping the step and `worst_path_violation` skipping the box, so nothing reported it — on the **cold** arm as much as the parametric one |

And separately, found in review: a **wrong-answer** hazard in how working sets
are carried between problems at all. See fix (2) below.

## Cause

**The fallback.** `solve_parametric_scoped`'s ineligible branch was
`self.solve(qp_new, None, opts)`. The primal genuinely does not carry between
two QPs — that is why a homotopy exists — but the *discrete* state does, and it
is far more stable than the iterate: it records which constraints bind, not
where the solution is.

**The missing events.** `Event` carried
`AddRowLower/AddRowUpper/DropRow/DropBound` and the primal ratio test looped
`for i in 0..m`. A crossing is absorbing (the ratio test only ever *prevents* a
violation, never repairs one), so a bound crossed once stayed crossed to `t = 1`.

**The topology hazard.** `ConsStatus::Equality` and `BoundStatus::Fixed` are not
"active" markers — they assert `bl == bu` / `xl == xu` about the problem they
came from, and no drop test can retract either. Carried onto a problem where the
row is a range, `pin_working_set` pins it to the `-1e20` sentinel at a point the
feasibility audit accepts, and the solve reports `Optimal`. No check downstream
can catch it: the point *is* optimal for the problem the working set describes,
and it is the working set that describes the wrong problem.

## Fix

1. **`solve_parametric`'s declined branch keeps the working set** via
   `solve_with_working_set`, falling all the way back to cold only when that
   hint is unusable. Gated on `sol_prev.status == Optimal` and dimensional
   validity — without the latter a shape change returns
   `WarmStartDimensionMismatch`, turning "no warm start available" into "solve
   failed".
2. **`WorkingSet::reconciled_with`, applied on all three routes a hint can take.**
   [@GermanHeim found](https://github.com/jkitchin/pounce/issues/602#issuecomment-5303884538)
   that `min ½x² s.t. x == 1` re-solved as `min x² s.t. x ≤ 2` returned
   **`Optimal` at `x = -1e19`, objective `1e38`** against a true optimum of 0.
   The first fix covered only the declined branch; [his review of this PR](https://github.com/jkitchin/pounce/pull/614#issuecomment-5304670936)
   found the other two. With `H` identical the guard *admits*, so the traced
   path clones the stale set straight through to the corrector — three of four
   topology-change cases were wrong, and the first round of tests missed it
   because every one of them perturbed `H` and so only exercised the fallback.
   Now: a `same_topology` condition on the guard (the row type does not
   interpolate — a row that is an equality at `t = 0` is a range at every
   `t > 0` — so the path cannot re-type it and declining is the only honest
   answer), plus reconciliation inside `solve_with_working_set` itself, which
   closes the hazard for external callers.
3. **`AddBoundLower`/`AddBoundUpper`** plus the ratio test over `j in 0..n`, a
   `tabu_bounds` mirroring `tabu_cons` (the rank repair can now fight the new
   test the way it did for rows), and the box added to the feasibility report.

Note that (2)'s guard is a **correctness** guard, which is why it stands where
the `A`/box guard below was declined on cost grounds. Different question,
different bar.

### Rejected, with what ruled each out

Three things were built, measured and backed out, all recorded in full in
`dev-notes/issue-602-parametric-eligibility.md`.

**The guard #602 actually asked for.** Requiring unchanged `A`/`xl`/`xu`/inertia
is better or equal in 14 of 14 rows at `n = 30` and **worse in 9 of 14 at
`n = 20`** — same family, same generator, `A` perturbed 10% going from 2
working-set changes to 34. That is #434's situation restated. The reasoning that
looked sound and was not: the path does not *model* `A`, so tracing it is
unjustified extrapolation — true, but rejection is not a return to correctness,
it is a switch to the working-set heuristic, which has its own failure mode. The
inertia condition is the clearest case against, having no upside even in
principle: nothing on this path reads `hessian_inertia`, so declining on it only
swaps a working path for the fallback — 2 changes became 5, for nothing.

**A runtime signal for "is this working set worth keeping".** Built
(`hint_pin_quality`: pin the hint, count the rows and bounds outside it that the
pinned point violates — one factorization `solve_with_working_set` already pays)
and swept over 360 pairs on both arms. Against the conventional cold path there
is nothing to decide: always-keep is wrong on 2 of 360 and sits 0.05% off the
per-case oracle, which is a positive result about (1). Against the homotopy there
is a real 27% prize and **every threshold scores worse than having no rule**. The
refutation is direct, not statistical — identical signal values carry opposite
answers:

```
active=20 violated=1   n30 da0      ws  3  cold 21   → keep
active=20 violated=1   n30 da0.05   ws 64  cold 22   → go cold
```

Structurally: the signal measures "is the hint good", the decision is "is the
hint better than the alternative", and on that arm the alternative's cost is a
property of the *path*. Instrument kept as `#[cfg(test)]`; no rule ships.

**Interpolating `A`.** Out of scope, and not an extension of this tracer: with
`A(t)` the KKT matrix is `t`-dependent, so `(x(t), λ(t))` stops being affine
along a segment and both ratio tests in `t` stop being exact. That needs a real
predictor–corrector continuation.

## Blast radius

| surface | effect |
|---|---|
| `solve_parametric` | (1), (2). **No production caller** — only two test files; the SQP driver warm-starts through `solve_with_working_set` because each linearization moves `A` |
| `solve_with_working_set` | (2)'s reconciliation — public API, and the SQP driver's warm-start route |
| homotopy tracer (both arms) | (3) |
| `pounce-algorithm` SQP driver | `sqp_qp_use_homotopy` defaults **no** |
| `pounce-convex` active-set QP driver | sets `use_homotopy: true` — **measured, see the note at the top** |

**Maros-Mészáros, 138 problems, `pounce-convex`'s active-set driver** (the
gating run; [method and full tables](https://github.com/jkitchin/pounce/pull/614#issuecomment-5305253101)):

| | `main` | this PR |
|---|---|---|
| correct vs published optimum | 57 / 138 | **57 / 138** (same 57) |
| solved-but-wrong | 0 | **0** |
| iterations over the 57 both solve | 103,810 | 113,623 (+9.5%) |

18 instances move, 13 worse and 5 better; excluding `CVXQP2_M` the aggregate is
+2.0%. Ablation splits it cleanly with no overlap: **every loss is the
bound-adding ratio test** (disabling it returns each losing instance to `main`'s
exact iteration count) and **every gain is the reconciliation**. That is the
expected shape — a path that could previously walk out of a stationary box now
stops at each crossing, so it takes more events to reach `t = 1`. Eleven
instances change failure mode; none of them solves correctly on either arm, and
four move `MaxIter` → `InternalError`, which callers branching on status will
see.

**Warm-start suite** (`benchmarks/warmstart`, whose `-hom` arms differ from
their twins by exactly one option — #434's instrument), 42 family×scale
combinations, inner-QP working-set changes, across the whole PR:

| arm | base → final |
|---|---|
| `cold-sqp`, `warm-sqp`, all IPM arms | **bit-identical** |
| `cold-sqp-hom` | 36726 → **27828** (−24%) |
| `warm-sqp-hom` | 2240 → **1755** (−22%) |

Solved counts unchanged (855/855 SQP, 549/549 QP). In #434's framing the
homotopy's excess inner work over the conventional path falls from 2.82× to
2.13× cold.

**CLI fixture sweep** (`scripts/sweep-fixtures.sh`, 57 fixtures): empty on the
default arm. Five lines move on the SQP arms, all on fixtures that were already
failing, and one stops failing — `jit1` on the homotopy arm now converges in 11
iterations (dual infeasibility `9.3e-9`, constraint violation `6.5e-19`). The
topology guard alone is inert on all three arms, since `solve_parametric` has no
production caller.

## Tests

| test | bites? |
|---|---|
| `ineligible_parametric_reuses_the_working_set` | **yes** — fails on the parent at 7 working-set changes against cold's 7. Asserts the *change count*, not just the answer: the old behaviour returned the right result and would pass any answer-only assertion, which is the #544 shape |
| `declined_parametric_survives_an_equality_becoming_a_range` | **yes** — the originally reported bug |
| `declined_parametric_survives_a_fixed_variable_being_freed` | **yes** — the `Fixed` variant |
| `traced_path_survives_an_equality_becoming_a_range` | **yes** — the review finding, across both `use_homotopy` settings |
| `traced_path_survives_a_fixed_variable_being_freed` | **yes** — wrong on *both* settings pre-fix |
| `solve_with_working_set_reconciles_a_stale_hint` | **yes** — the public-API route, hint passed straight in as an external caller would |
| `declined_parametric_survives_a_range_becoming_an_equality` | **no** — passes either way (a stale `AtUpper` pins to `bu`, the same point when `bl == bu`) |
| `declined_parametric_survives_a_bound_side_disappearing` | **no** — passes either way, because `AtLower` is *droppable*: the dual test removes it, so the same bad pin costs iterations rather than correctness. Only `Equality`/`Fixed`, which no drop test can reconsider, turn it into a wrong answer |
| 4 new cases in `tests::homotopy_unit` | pin the bound events' selection rule; they cannot compile on the parent, since the enum variants are new |
| `tests::hint_signal::hint_signal_sweep` | `#[ignore]`d — a measurement, not a property. Nothing here should gate CI |

The two that do not bite say so in their doc comments rather than implying
coverage they do not have.

**Deliberately not pinned:** the warm-start, Maros-Mészáros and fixture-sweep
numbers above. They need built Python bindings, a corpus and a baseline binary,
so they are recorded in the dev-note and the linked comment and reproducible
from those rather than asserted in CI.

---

- [x] Tests fail on the parent commit for the stated reason — for the six
      marked "yes"; the other three are labelled above rather than claimed
- [x] Book page under `docs/src/` updated — the design note's §5 trait sketch
      now matches the shipped guard, including the topology condition
      (`pounce-qp`, `pounce-rs`, `pounce-convex`, `pounce-algorithm`; clippy at
      CI strength `-D clippy::correctness -D clippy::suspicious`);
      `check-docs-consistency.sh` OK. Re-verified after the merge with `main`
- [x] Every claim in this body is true of the code as it stands — including the
      three declined experiments, which are described as declined rather than
      as features, and the review round, which corrected a "bit-identical"
      claim in a code comment that the measurement did not support

### Review checklist beyond the usual

1. ~~Run the Maros-Mészáros homotopy sweep~~ — **done**, passes on correctness;
   see the note at the top. Worth recording that the prediction attached to this
   item was wrong: two earlier verdicts in this PR flipped once the measurement
   widened, so this one was expected to change the picture too. It did not — the
   correctness gate came back clean and the cost it surfaced was already
   attributable.
2. ~~Land the design-note correction~~ — done.
3. ~~`solve_with_working_set` applies no reconciliation of its own~~ — done in
   this PR after review, with the A/B that had been deferred.
4. `CVXQP2_M` (1768 → 9575 iterations, same status and objective) is filed as
   #615 rather than left in a commit message, per the gh#544 lesson in
   `CLAUDE.md`. Not a blocker: the ratio test causing it is the correctness fix,
   and the homotopy was already losing 3.4× on that instance before this PR.

https://claude.ai/code/session_01QMxWPhrmhNUXMPjLxNacCT


jkitchin:
@GermanHeim — this is #602 landed, and since three of the four pieces came out of your report or your review, a look from you would be worth more than most. (GitHub won't let me add you as a formal reviewer — not a repo collaborator — hence the mention.)

The parts most worth your time, roughly in order:

**1. Your bug fix, and whether the rule is right.** `WorkingSet::reconciled_with` in `crates/pounce-qp/src/working_set.rs`. It re-derives `Equality`/`Fixed` from the new problem and drops any status naming a bound the new problem doesn't have. The specific thing to check is the choice for a row that *was* an equality and is now a range: I set it `Inactive`, on the reasoning that it tells you the row was tight but not which side of the new range it belongs on. Mapping it to `AtLower`/`AtUpper` by whichever side is finite was the alternative and I couldn't justify it — but you found the original defect, so if that reasoning has a hole you're likely to see it.

The one derived finding worth flagging: the dividing line isn't staleness, it's **droppability**. A stale `AtLower` pins to the same `-1e20` sentinel and still returns the right answer, because the dual ratio test removes it. Only `Equality` and `Fixed`, which no drop test can reconsider, turn a bad pin into a wrong answer. That's why two of the four regression tests bite and two don't, and the two that don't say so in their doc comments.

**2. Your bullet 2, which had a deeper cause than either of us said.** The homotopy had no bound-*adding* event at all — bounds could be dropped, never added — so interpolating `xl`/`xu` as the issue floated would have moved a bound the ratio test still couldn't see. Added `AddBoundLower`/`AddBoundUpper` and the ratio test over `j in 0..n`. Note this is a defect on the **cold** arm too, not just the parametric one.

**3. Your bullets 1 and 3 were declined, and you should push back if you disagree.** I implemented the guard you asked for, measured it, and backed it out: better or equal in 14 of 14 rows at `n = 30`, worse in 9 of 14 at `n = 20`, same family — `A` perturbed 10% going from 2 working-set changes to 34. The argument I'd made for it (the path doesn't model `A`, so tracing it is unjustified extrapolation) is true and survives the data; what doesn't follow is that *declining* is better, because rejection isn't a return to correctness, it's a switch to the working-set heuristic with its own failure mode. Full tables in `dev-notes/issue-602-parametric-eligibility.md` under "Step 2, declined". If you think the measurement is the wrong shape — different family, different metric — that's a fair challenge and I'd rather hear it now.

**4. The generalisation of your finding, which I did *not* fix.** `solve_with_working_set` is public API and applies no reconciliation of its own, so a caller handing it a topology-mismatched working set hits the `x = -1e19` behaviour directly, with no `solve_parametric` involved. What my change did was make `solve_parametric` reach that hazard on its own from two well-formed problems — that made it my regression to fix, but it isn't the whole defect. The SQP driver is safe only *by construction* (an equality row stays an equality across linearizations), which is a precondition holding by accident of the caller rather than by contract. Making it reconcile internally is a trajectory change for that driver and wants its own A/B, so it's recorded rather than done. Probably its own issue.

Two things gate merge regardless, both in the PR body: the Maros-Mészáros homotopy sweep (`pounce-convex` is the only shipped path running the tracer by default, and it's the one path I couldn't measure), and a six-line doc edit that couldn't be pushed.

---

GermanHeim:
As far as I can see, there are (possibly) two issues:

- `solve_parametric_scoped` enters the homotopy before calling `reconciled_with`, then the tracer clones the (stale) working set.  With identical H, changing x == 1 to x <= 2 would still be fine? I believe fixed var to free would give the same wrong result. I think we could add a topology-like guard that routes these cases through the reconciled fallback, as well as identical-H regressions. Tests would probably be nice.
- `solve_with_working_set` still accepts `Equality`/`Fixed` statuses that are not possible. The entry point passes the supplied set without reconciliation.

I haven't actually tested it because I am working on other stuff, but as far as I can see, those would fail.

jkitchin:
@GermanHeim Both right, both fixed in f6b0a3a. Point 1 was worse than you stated — thanks for catching it.

## Point 1: the traced path

Reproduced before touching anything, `H` bit-identical so the guard admits:

| case | `use_homotopy` | parametric | cold |
|---|---|---|---|
| equality → range | false | **`Optimal`, x = −1e19** | x = 0 |
| equality → range | true | x = 0 | x = 0 |
| fixed → free | false | **`Optimal`, x = −1e19** | x = 0 |
| fixed → free | true | **`Optimal`, x = −1e19** | x = 0 |

So three of four, not the one you predicted — your "identical H, x == 1 to x <= 2 would still be fine?" was right for `use_homotopy=true` and wrong for `false`, and the `Fixed` case is wrong on both. The reason my first round of tests missed all of it: every one of them perturbed `H`, so they only ever exercised the fallback branch. One reproduction picks one route, and I fixed the route rather than the hazard.

**On your suggestion of a topology-like guard vs. reconciling the tracer's set** — I went with the guard, because the path can't be patched into handling this. The row *type* doesn't interpolate: with the new lower bound infinite, `bound_rate` is zero for it, so the interpolated row is `[1, 1+t]`. `Equality` is correct at exactly `t = 0` and wrong at every `t > 0`, the tracer has no mechanism to re-type it mid-path, and no drop test can either. That puts it in the same category as `H` — a change the path genuinely cannot model — so declining is the honest answer, and declining is now cheap because the fallback reconciles.

`same_topology` requires equality-ness per row and fixed-ness per variable to agree, using the solver's own predicates (exact `bl == bu`; both-finite-within-`feas_tol` for the box). Inert on the CLI fixture sweep across all three arms, since `solve_parametric` has no production caller.

Worth flagging explicitly since I declined a guard two commits ago and am now adding one: **this is a correctness guard, not a cost guard.** The `A`/box guard was declined because it swapped one heuristic for another with no reliable winner. This one prevents a wrong answer. Different question, different bar.

## Point 2: `solve_with_working_set`

You asked, so it got the A/B I'd said it needed rather than the deferral:

| | base → new |
|---|---|
| `cold-sqp`, `warm-sqp`, all IPM arms | **bit-identical** |
| `cold-sqp-hom` | 28487 → **27828** |
| `warm-sqp-hom` | 2005 → **1755** (−12%) |

Solved counts unchanged, every per-family move an improvement. On the CLI sweep five lines move, all on fixtures already failing — and one stops failing: `jit1` on the homotopy arm now converges in 11 iterations, dual infeasibility `9.3e-9`, constraint violation `6.5e-19`, objective 173346.0967 against the IPM path's 173345.3768 (two independently converged KKT points, 4e-6 apart).

So my prediction that it would move the SQP trajectory was right, and my inference that this made it too risky to include was wrong — it moves in the right direction, and the arms where the driver actually lives are bit-identical. The residual difference traces to hints where a variable's bounds sit within `feas_tol` of each other without being exactly equal, which reconciliation promotes to `Fixed`.

## Tests

Three, all verified failing on the pre-fix code, two across both `use_homotopy` settings:

- `traced_path_survives_an_equality_becoming_a_range`
- `traced_path_survives_a_fixed_variable_being_freed`
- `solve_with_working_set_reconciles_a_stale_hint` — passes the stale hint straight into the public entry point, as an external caller would

You said you hadn't tested it because you're on other things — you didn't need to; the reasoning was enough to find it. The gating item on this PR is still the Maros-Mészáros sweep, unchanged.

---

jkitchin:
## The gating Maros-Mészáros sweep: run, and it passes on correctness

**Correctness is unchanged and there are no wrong answers on either arm.** That
is the question the gate exists to ask.

| | `main` (5d8ad360) | this PR (f6b0a3a + merge) |
|---|---|---|
| correct vs published optimum | **57 / 138** | **57 / 138** |
| solved-but-wrong | **0** | **0** |
| correct on one arm only | — | **none, in either direction** |

Same 57 problems, not just the same count.

### The instrument had to be substituted, and that is worth stating plainly

`cargo run -p pounce-convex --release --example homotopy_sweep` **cannot be run
here.** It wants a flat `.qp` corpus, and per
`dev-notes/issue-434-homotopy-cost.md:280-282` the sweep driver and the `.mat`
→ flat-text converter were never committed. No `.mat` or `.qp` files exist on
this machine.

Substitute, reaching the same code: the 138-problem `.nl` corpus through the
CLI with `solver_selection=qp-active-set`, which enters
`pounce_convex::active_set` and its `use_homotopy: true` (`active_set.rs:518`)
— the surface the PR body names as unmeasured. Confirmed the tracer actually
runs rather than being bypassed:

```
[hom] box relaxation: Optimal obj=-5.125000e0
[hom] summary exit=complete steps=2 t=1 changes=3 refactor=2 stall=0
[hom] reached t=1 after 3 working-set changes; max target violation 2.220e-16
```

120 s cap, matching the cap behind the `71/138 vs 58/138` figure already
recorded at `active_set.rs:505`. The machine was shared with an unrelated
long-running job for most of the sweep, so **every problem ran on both binaries
back-to-back on the same worker** — arm-then-arm would have let an hour of load
drift land on one arm and read as a timeout difference. Wall clock is therefore
not a reported metric; status, objective-vs-published and iteration count are
load-independent. 63/60 timeouts (`main`/PR) are load-inflated against the
71/138 on record and should be read as censored, not as regressions.

## Effort moved, and the ablations say exactly which commit moved it

Over the 57 problems both arms solve: **103,810 → 113,623 iterations (+9.5%)**.
18 instances move, 13 worse and 5 better. One instance is most of it —
**excluding `CVXQP2_M` the aggregate is +2.0%.**

| instance | `main` | PR | ratio |
|---|---|---|---|
| `CVXQP2_M` | 1768 | **9575** | ×5.42 |
| `HS76` | 2 | 13 | ×6.50 |
| `QISRAEL` | 637 | 1007 | ×1.58 |
| `GOULDQP3` | 1687 | 2649 | ×1.57 |
| `GOULDQP2` | 1655 | 2193 | ×1.33 |
| `MOSARQP2` | 1949 | 2182 | ×1.12 |
| `LOTSCHD` | 37 | **20** | ×0.54 |
| `DPKLO1` | 400 | **325** | ×0.81 |
| `CVXQP1_S` | 334 | **294** | ×0.88 |
| `CVXQP2_S` | 173 | **153** | ×0.88 |

Two ablation binaries off the PR head, one change disabled in each:

| instance | `main` | PR head | reconciliation OFF | bound ratio test OFF |
|---|---|---|---|---|
| `CVXQP2_M` | 1768 | 9575 | 9575 | **1768** |
| `GOULDQP3` | 1687 | 2649 | 2649 | **1687** |
| `GOULDQP2` | 1655 | 2193 | — | **1655** |
| `QISRAEL` | 637 | 1007 | 1007 | **637** |
| `LOTSCHD` | 37 | 20 | — | 20 |
| `DPKLO1` | 400 | 325 | — | 325 |

Clean split, no overlap:

- **Every loss is the bound-adding ratio test** (9befbffa). Disabling it returns
  each losing instance to `main`'s exact iteration count.
- **Every gain is the reconciliation** in `solve_with_working_set`. Disabling
  it leaves the losses untouched; the bound-ratio ablation leaves the gains
  untouched.

That is the honest shape of the trade and it matches the change: a path that
could previously walk out of a stationary box now stops at each crossing, so it
takes more events and more refactorizations to reach `t = 1`. The extra
iterations are the cost of the path no longer being silently wrong about the
box — not a tuning regression.

Cross-check that it is the tracer and nothing else: with
`sqp_qp_use_homotopy=no`, both binaries give **525 iterations, `Optimal`,
identical objective** on `CVXQP2_M`. The whole difference lives on the homotopy
arm. (Also worth noting for its own sake: this instance solves *better* with
the homotopy off — 525/`Optimal` against 1768/`SolvedToAcceptableLevel` — which
is the trade `active_set.rs:505` already documents.)

### `CVXQP2_M` needs an issue and an owner before this merges

Not a merge blocker on correctness, but ×5.42 on one instance is exactly the
shape `CLAUDE.md` says must not be recorded as "an accepted cost of the fix" and
left in a commit message. gh#544 went 206 → 812 and shipped, because the suite
asserts status and objective and this instance keeps both
(`SolvedToAcceptableLevel`, objective 820155.431 on both arms). 1768 → 9575 at
8.6 s → 63.6 s is the same signature. Attribution is already known, which is the
expensive half of such an issue; what it needs is a name against it.

## Failure-mode changes: 11 instances, none of them a solve either way

| instance | `main` | PR |
|---|---|---|
| `QCAPRI` | `MaxIter`, 6207 it, 102 s | `InternalError`, 7 s |
| `QE226` | `MaxIter`, 4788 it | `InternalError` |
| `QGROW7` | `MaxIter`, 4799 it, obj 4.55e17 | `InternalError` |
| `QSHARE1B` | `MaxIter`, 3514 it | `InternalError` |
| `QBANDM`, `QFORPLAN`, `QSCAGR25`, `QSCSD6`, `QSIERRA` | `TIMEOUT` (120 s) | `InternalError`, faster |
| `QPILOTNO`, `QSTAIR` | `InternalError` | `TIMEOUT` |

None of these solves correctly on either arm, so none affects the 57/57 above.
Ablation attributes them to the **reconciliation**: with it disabled `QCAPRI`
returns to `MaxIter` (6957 it, 137 s).

Mostly this is failing faster and more honestly — `QCAPRI` gives up in 7 s
instead of spending 102 s to return `MaxIter` at an objective of 6.4e11 against
a published 6.68e7, and `QGROW7`'s `MaxIter` point is 4.55e17. But it is a
**status-class change on four instances** (`MaxIter` → `InternalError`, i.e. a
returned iterate becomes "numerical failure, no verified KKT point"), and
callers that branch on status will see it. Worth a CHANGELOG line even though
no correct answer moves.

## Code review

Three findings; the third is now closed by f6b0a3a.

**1. `max_violation`'s doc comment is captured by `HintPinQuality`**
(`solver.rs:4094`). The struct landed between an existing doc comment and the
`fn` it documented, so "The magnitude behind `point_is_feasible`'s boolean…"
now documents the struct and `max_violation` is undocumented. Fixed in 960935a.

**2. A box violation prints as a row that does not exist** (`homotopy.rs:961`).
`worst_path_violation` now returns `n_rows + j` for a variable — its own doc
says so — but the only caller formats every index as `row {i}`, so on a
20,200-variable problem a bound violation prints as `row 25000`. Since this
trace is *the* instrument this PR adds for path feasibility on the box,
misreporting it defeats the change. Fixed in 960935a: decodes to `bound on x[j]`.

**3. `solve_with_working_set` reconciliation — closed.** This was going to be my
recommendation (`O(n+m)` against a factorization, and the failure it prevents is
`Optimal` at 1e38); f6b0a3a does it, with the A/B rather than an issue. Checked
that the placement is safe: `solve_with_working_set_scoped` has exactly one
caller, the public entry point, so reconciliation happens once at the boundary
and `pin_working_set`'s rank repair runs downstream of it — there is no
re-entry path where reconciliation could re-add a status the rank repair just
pruned.

**On `same_topology`**: predicates match the solver's own (`solver.rs:519`,
`:790`) and `reconciled_with`, `qp_prev.m == qp_new.m` is the first conjunct so
the per-row loop cannot index out of bounds, and the argument for guarding
rather than patching the path holds — the row type does not interpolate, so
`Equality` is right at exactly `t = 0` and wrong at every `t > 0`.

**On the `Equality`-becomes-a-range → `Inactive` choice** (the specific question
put to review): `Inactive` is right, and the reason is the same droppability
argument the PR already makes. `Inactive` is recoverable by the *add* ratio
test; a wrongly-guessed `AtLower` is recoverable only by the *drop* test. The
guess buys nothing and risks a pin.

**Cosmetic**: the declined branch calls `reconciled_with` explicitly and then
calls `solve_with_working_set`, which now reconciles again. Idempotent, so
correct — just duplicated.

## Housekeeping

- **Merge conflict resolved and pushed** (`CHANGELOG.md` only, both
  `[Unreleased]` sections kept, zero markers left) — 4d87061. The two fixes
  above are 960935a. The branch should read `MERGEABLE` again.
- **`gh pr checks 614` reports no checks at all on the branch** — nothing has
  ever run on it. Stand-ins: `cargo fmt --check` clean, `cargo test -p pounce-qp
  --release` green (13 tests in `tests/homotopy.rs`, including the three new
  ones), `scripts/check-docs-consistency.sh` OK at 44 pages, and
  `cargo test --workspace --exclude pounce-hsl --release` green — 223 targets,
  2917 tests, 0 failed.
- The outstanding doc edit is **done upstream** in f6b0a3a, and its version is
  better than the one drafted in the PR body — it names the topology condition.

---

## TL;DR

**Good to go on correctness; one thing to name first.** The gate passes —
57/138 correct on both arms, the same 57, zero solved-but-wrong. The cost is
+9.5% iterations, and the ablation says every bit of it is the bound-adding
ratio test, which is a correctness fix paying for itself. The one item I would
not merge past silently is `CVXQP2_M` at 1768 → 9575 (×5.42, same status, same
objective) — that is the gh#544 signature exactly, and `CLAUDE.md` says it needs
an issue and an owner rather than a line in a commit message. Attribution is
already done here; it needs a number against it.

_Reviewed with Claude Code._

jkitchin:
Filed the `CVXQP2_M` regression as #615 — 1768 → 9575 iterations (×5.42, 8.6 s
→ 63.6 s), same status and same objective on both arms, attributed by ablation
to the bound-adding ratio test in 9befbffa.

That closes the one condition I raised above. It is **not** a blocker for this
PR: the ratio test is the correctness fix, the gate passes 57/57 with zero
wrong answers, and the point of #615 is that the number has a name against it
rather than living in a commit message — the gh#544 lesson from `CLAUDE.md`.

Worth carrying into #615 from the sweep: with `sqp_qp_use_homotopy=no` this
instance solves in 525 iterations to `Optimal`, against 1768 to
`SolvedToAcceptableLevel` on `main`. The homotopy was already losing here by
3.4× before this PR; the ratio test widened a gap it did not create, which
suggests the admission test is a likelier place to fix it than the ratio test.

Issue #618, opened 2026-08-16
Warm start: mildly stale seeds regress — the μ escalation band has a hole in the middle


jkitchin:
Follow-up to #606, requested by that PR's own measurements. Filed so the regression has an owner rather than living only in a PR body — per `CLAUDE.md`. **This issue is unassigned and needs an owner.**

## What was measured

Baseline `70bf53ded3d893cfa2da6ead5195fda5ac096f68`, `benchmarks/warmstart`. Seed each family from the far end of its own 4×-scale parameter path, with a full seed:

| | iterations, summed over 9 families |
|---|---|
| before #606 | 118 |
| after #606 | **127** |

Per family: `redundant_rows` 3 → 7 (worst), `degenerate_corner` 9 → 11, `hanging_chain` 8 → 10, `moving_bound_qp` 11 → 12; the rest unchanged.

## Why it happens

#606 raises `μ` to the point's measured average complementarity only when that overshoots `mu_init` by more than **10×**. That trigger is deliberately conservative: an earlier revision that escalated whenever the measurement disagreed moved `cresc4` from 85 to 206 iterations on a half-order-of-magnitude change — the gh#544 shape exactly.

The consequence is a hole in the middle of the staleness range. These points are stale enough that dual reconstruction fires and reroutes the trajectory, but not stale enough to trip the escalation that is supposed to pay for it. So they get the cost without the benefit.

Both ends of the range are fine:

- **Fresh** — exact same-model restart improves (6 → 2 iterations summed, full seed).
- **Strongly stale** — HS071 warm-started from the far corner of its box goes from `RestorationFailed` in 36 iterations to `SolveSucceeded` in 43. That case *fails* before #606.

## What to do

A second, gentler escalation band between "untouched" and "10× overshoot" — raising `μ` partway rather than all the way, or scaling the reconstruction's aggressiveness by measured staleness instead of gating it on a single threshold.

The constraint is `cresc4`: any proposal has to leave it bit-identical, or justify moving it. It is nonconvex, enters restoration on iteration 2, and compounds perturbations, which makes it the sensitive canary for anything that touches `μ`.

## Reproduce

`benchmarks/warmstart`, stale-seed mode at 4× scale, against `70bf53d` and the #606 branch head. `warm_start_recentering=none` restores the pre-#606 numbers exactly and is the attribution control.

jkitchin:
One more data point for this band, from the review fixes on #620 (`96ee510`).

[...]

Issue #617, opened 2026-08-16
Warm start: reject corrupted dual seeds instead of reconstructing off them


jkitchin:
Follow-up to #606, requested by that PR's own measurements. Filed so the regression has an owner rather than living only in a PR body — per `CLAUDE.md`. **This issue is unassigned and needs an owner.**

## What was measured

Baseline `70bf53ded3d893cfa2da6ead5195fda5ac096f68`, `benchmarks/warmstart`, 9 families. Seed the exact primal solution together with multipliers corrupted by `N(0, 1e4)` noise:

| | iterations, summed over 9 families |
|---|---|
| before #606 | 18 |
| after #606 | **71** |
| cold (no seed) | 81 |

Worst single family: `hanging_chain` 3 → 13. Every case still converges with status `SolveSucceeded` to the same objective, and stays within ~1.4× of the cold solve — so this is a slowdown, not a correctness failure.

## Why it happens

#606 reconstructs the slack-bound multiplier block (`v_L` / `v_U`) — which no frontend can seed — from the supplied `y` via the stationarity identity. When the supplied `y` is garbage, the reconstruction faithfully propagates the garbage into a block that previously got a constant floor.

The pre-#606 number was not skill. Filling that block with `warm_start_mult_bound_push` regardless of input happened to discard the corruption for free. #606 trades that accidental robustness for a genuine reconstruction that pays off on good seeds (`hanging_chain` full restart 49 → 14) and costs on bad ones.

## What to do

Reject a seeded dual block whose implied complementarity is orders of magnitude away from the rest of the point, and fall back to the pre-#606 constant fill for that block — rather than reconstructing off it. The measurement machinery to detect this already exists in the new initializer: it computes `avrg_compl` and the per-block residuals. What is missing is a consistency test between the seeded `y` and the primal point it is supposed to accompany.

Care needed on the threshold: #606's `μ` escalation deliberately fires only at 10× because a half-order move on `cresc4` cost 85 → 206 iterations. A seed-rejection test wants to be similarly conservative, and it should be measured on the same corpus in all four seed modes, not just the corrupted one.

## Reproduce

`benchmarks/warmstart`, corrupted-seed mode, against `70bf53d` and the #606 branch head. Setting `warm_start_recentering=none` restores the pre-#606 numbers exactly and is the attribution control.

PR #620, opened 2026-08-16
Add residual-adaptive, complete NLP warm-start recentering (#606)

### LINKED ISSUE #606: Add residual-adaptive, complete NLP warm-start recentering
## Summary

Warm starting should adapt to the quality of the supplied iterate and should reconstruct missing state when possible. Today the high-level helper uses fixed push values and a clipped previous barrier value, while the core fills missing duals with constants/zeros and two advertised full-iterate flags are parsed but not meaningfully consumed.

## Current behavior

- `WarmStart` defaults to a fixed `bound_push=1e-9` and derives all warm-start push options from it ([_warm_start.py#L58-L76](https://github.com/jkitchin/pounce/blob/5d8ad360feb13e49bf6aefe66e78a14ddd054229/python/pounce/_warm_start.py#L58-L76), [_warm_start.py#L160-L178](https://github.com/jkitchin/pounce/blob/5d8ad360feb13e49bf6aefe66e78a14ddd054229/python/pounce/_warm_start.py#L160-L178)).
- Missing equality/bound dual entries are filled with zero/default values in the core warm initializer rather than estimated from stationarity ([warm_start.rs#L80-L130](https://github.com/jkitchin/pounce/blob/5d8ad360feb13e49bf6aefe66e78a14ddd054229/crates/pounce-algorithm/src/init/warm_start.rs#L80-L130)).
- `warm_start_same_structure` and `warm_start_entire_iterate` are parsed in [application.rs#L3412-L3444](https://github.com/jkitchin/pounce/blob/5d8ad360feb13e49bf6aefe66e78a14ddd054229/crates/pounce-algorithm/src/application.rs#L3412-L3444), but their promised reuse semantics are incomplete.

## Proposed scope

1. Measure primal, dual, and complementarity residuals of the supplied warm point.
2. Choose `mu`, bound/slack pushes, and multiplier floors from those residuals and scaling—not fixed universal constants.
3. Reconstruct missing duals with a sparse regularized stationarity least-squares solve, then clip/recenter for positivity.
4. Support a genuinely complete iterate transfer (primal, equality duals, bound duals, slacks, barrier state, and reusable structure/factorization where valid).
5. Either implement `warm_start_same_structure` / `warm_start_entire_iterate` fully or reject them clearly until supported.
6. Expose diagnostics describing what was accepted, reconstructed, or discarded.

## Acceptance criteria

- Exact same-model restarts converge without degrading a supplied KKT-quality point.
- Partial warm starts outperform zero-filled duals on a repeated-solve benchmark.
- Bad/stale warm points safely fall back to stronger recentering or cold initialization.
- Option semantics are covered end-to-end through all public frontends.

Related: #86, #121, #484.

jkitchin:
Fixes #606

## Problem

Two defects in the interior-point warm start, both invisible from the outside, both measured on the parent commit `70bf53de` before anything was changed.

**1. You cannot seed every multiplier block, and the block you cannot seed was filled with a constant.** `TNLP::get_starting_point` — what `lagrange` / `zl` / `zu` reach on Python, the C ABI and the `.nl` path alike — carries the constraint multipliers and the *variable*-bound multipliers. The IPM also needs a multiplier for each inequality row's slack (`v_L` / `v_U` internally), and no frontend has a field for one. On every warm start POUNCE has ever run, that block arrived as zero and was floored at `warm_start_mult_bound_push` — a number chosen with no reference to the slacks it is paired against. The "warm" point handed to the solver was therefore not a stationary point of anything.

On HS071, restarting from its own converged solution with every multiplier the TNLP surface can carry:

| | iters | measured `‖∇ₓL‖∞` after init |
|---|---|---|
| before | 2 | — (not measured) |
| after | 1 | 3.5e-10 |
| cold, same model | 8 | |

**2. `mu_init` was taken on trust.** A restart at the optimum and a stale point from a different parameter both started on exactly the barrier the caller named. On HS071 warm-started from the far corner of its box with the converged duals attached — the documented recipe from `docs/src/initialization.md`, applied to a point that has moved:

| | status | iters |
|---|---|---|
| before | **`RestorationFailed`** | 36 |
| after | `SolveSucceeded` | 43 |
| cold, same model | `SolveSucceeded` | 8 |

That is not a slowdown. The pre-#606 path *fails* on it.

**3. Two advertised options did nothing.** `warm_start_same_structure` and `warm_start_entire_iterate` parsed, set a field on `WarmStartOptions`, and were never read. Verified bit-for-bit on `70bf53de` before touching anything — a warm re-solve of `simplex_proj` with each flag, both, or neither:

| `70bf53de` | status | iters | objective | KKT error |
|---|---|---|---|---|
| neither set | 0 | 2 | 0.590398708046706 | 2.5059430435806254e-09 |
| `same_structure=yes` | 0 | 2 | 0.590398708046706 | 2.5059430435806254e-09 |
| `entire_iterate=yes` | 0 | 2 | 0.590398708046706 | 2.5059430435806254e-09 |
| both `yes` | 0 | 2 | 0.590398708046706 | 2.5059430435806254e-09 |

Identical to the last digit. The issue's claim that their "promised reuse semantics are incomplete" is generous: they were complete no-ops.

## Cause

The warm initializer (`crates/pounce-algorithm/src/init/warm_start.rs`) received `cq` and `aug_solver` and used neither — both were bound as `_cq` and `_aug_solver`. Every decision it made was a function of the options rather than of the iterate it was handed: fixed pushes, a fixed multiplier floor, a fixed `mu`. There was no measurement anywhere in it, so there was nothing for it to adapt to.

The two dead flags name Ipopt's `TNLP::GetWarmStartIterate` surface, which POUNCE does not expose at all.

## Fix

`warm_start_recentering` (new, `residual` | `none`, default `residual`) turns the initializer into a pass over the supplied point.

1. **Measure.** `inf_pr` first — it is the one residual that does not depend on the duals, so it is meaningful even when every multiplier block is missing.
2. **Reconstruct the bound multipliers.** An entry that arrives as exactly `0` or `NaN` is not a legal barrier multiplier (the barrier needs `z > 0`), so it was never a seed — and it is already special-cased today, being floored at `warm_start_mult_bound_push`. What changes is the value it takes: the stationarity identity `P_L z_L − P_U z_U = ∇f + J_c^T y_c + J_d^T y_d`, and its slack-block twin `P_L v_L − P_U v_U = −y_d`, split by sign and floored at `μ / slack` so an inactive bound still gets what complementarity implies.
3. **Reconstruct an identically-zero `y` block** through the same regularized least-squares augmented solve the cold path uses (`LeastSquareMults`), now with real bound multipliers in its right-hand side.
4. **Raise `μ`** to the point's measured average complementarity when that overshoots `mu_init` by more than 10×.

`warm_start_target_mu` still pins `μ` outright. `warm_start_recentering=none` restores pre-#606 behaviour exactly.

### Four things tried and rejected, each on a number

A future maintainer retuning this will otherwise repeat these experiments. Each is documented at its definition in the source.

**Rejected: `μ ≥ κ·max(inf_pr, inf_du)`.** This is the obvious reading of the issue's "choose `mu` … from those residuals", and it is wrong. It cost **715 → 1129 iterations** across the 27 parametric paths of `benchmarks/warmstart`. On `simplex_proj/tiny`, a re-solve that needed one iteration measured `inf_du = 2e-3`, took `μ = 2e-3`, and needed five. A warm point at a moved parameter carries a primal and a dual residual of order Δθ *by construction* — that is the premise of warm starting — and raising `μ` to meet them throws the warm start away to pay for a Newton step that was about to happen anyway. Of the three KKT residuals, complementarity is the one `μ` *is*; the other two are what the Newton step is for. `avrg_compl` on that same point read `2.6e-9`, the converged barrier, correctly recognised.

**Rejected: filling `z = μ̂/slack` with a residual-inflated `μ̂`.** The slacks reaching the initializer have already been pushed into the bound interior by `warm_start_slack_bound_push`, so on a converged point — where the true slack sits at `μ/z` and the push dominates it — `μ̂/slack` is off by exactly the push's inflation. On HS071 an 8×-inflated `μ̂` put the reconstructed slack multiplier 8× high and took an exact restart from 1 iteration to 5. The fill barrier is `mu_init`; the stationarity split, which never looks at a slack, is immune to it entirely.

**Rejected: overriding `μ` whenever the measurement disagrees.** Moving `μ` reroutes the whole trajectory, so it has to be earned. On `cresc4` (nonconvex, enters restoration on iteration 2, where perturbations compound) a measurement of `5e-7` against a `mu_init` of `1e-7` moved `μ` by half an order and the solve from **85 iterations to 206** — same status, same objective, 2.4× the work. This is the gh#544 shape exactly. Hence the 10× trigger; with it that fixture is bit-identical again, and the cases this exists for (stale seeds miss by three to six orders) still fire.

**Rejected: reconstructing from a seed with no duals at all.** Completing a *partial* warm start is well-posed — the supplied blocks pin the missing ones through stationarity. Manufacturing a whole one from a primal point is not: what comes out is the cold path's estimate wearing the warm path's barrier, and it cost **1102 → 1211 iterations** on the same corpus (`degenerate_corner` 17 → 38 at every step size). Such a seed is now reported `unseeded` and left exactly as before. The same reasoning capped the least-squares `y` estimate at `constr_mult_init_max` (1e3, the cold path's cap) rather than `warm_start_mult_init_max` (1e6): on `redundant_rows`, whose duplicated equality rows make the system singular, the looser cap let an arbitrary estimate through and cost 7 → 25 iterations.

### Scope: what was and was not done

The issue lists six items. Items 1, 2, 3, 5 and 6 are implemented.

- **Item 4** — "a genuinely complete iterate transfer (… and reusable structure/factorization where valid)" — is **not** done, and is the same missing capability as item 5's refused half: it needs the `GetWarmStartIterate` TNLP surface POUNCE does not expose, plus a factorization-reuse path that does not exist. Building it is a feature, not a line; refusing the two options that advertise it is the honest half, and is what shipped.
- **Item 2's "bound/slack pushes"** — `warm_start_bound_push` and friends are still fixed constants. Only the multiplier *floors* and `μ` became residual-derived. Making the primal push adaptive moves the primal point itself, which is a much larger trajectory risk than anything here, and nothing in the measurements pointed at it.
- The **GAMS native C link** and the **C ABI** were not touched. They reach the same core initializer, so they get the behaviour change; they get no new frontend surface for the diagnostics.

## Blast radius

Baseline for every number below: **`70bf53ded3d893cfa2da6ead5195fda5ac096f68`** (`main`, the parent of this branch). Same machine, same binaries, both directions measured with the identical script.

### Fixture sweep — three regimes

`scripts/sweep-fixtures.sh` with default options never reaches this code: `warm_start_init_point` defaults to `no`, so the warm initializer is not even constructed. That sweep is bit-identical, and it proves almost nothing. It was run anyway, and then twice more in regimes that *do* exercise the change:

| sweep | options | result |
|---|---|---|
| cold (the standard one) | defaults | **bit-identical across all 57 fixtures** |
| warm, default pushes | `warm_start_init_point=yes` | **bit-identical across all 57 fixtures** |
| warm, tightened recipe | `warm_start_init_point=yes mu_init=1e-7` + all five `warm_start_*_push`/`_frac` at `1e-9` | **3 of 57 move** |

The middle sweep is the load-bearing one: it differs from the cold sweep on 17 of 57 fixtures, so the warm initializer demonstrably runs, and the change is still a no-op there. The mechanism is exact — with no dual seeds in the `.nl` files, `any_dual_seeded` is false, so no reconstruction runs; and with the default `warm_start_mult_bound_push` of `1e-3` against `O(1)` slacks the measured complementarity is `~1e-3`, below the `mu_init` default of `0.1`, so `μ` is untouched.

The third sweep is the documented tight-push recipe, i.e. the regime the change is *for*. Every line that moves:

| fixture | before | after | status / objective |
|---|---|---|---|
| `hs13_bigstart` | it=30, obj=0.9891769904 | it=29, obj=0.9883552912 | `SolveSucceeded` both; improvement |
| `jit1_boxed` | it=16 | it=17 | `SolveSucceeded` both, obj=173345.3768 to 10 digits |
| `jit1_node` | it=19 | it=21 | `SolveSucceeded` both, obj=173345.3768 to 10 digits |

Each was attributed by running it back with `warm_start_recentering=none`, which reproduces the "before" column exactly on all three — so the mechanism is this change and nothing else, and the kill switch works. `hs13` is the known MFCQ-failure model, whose objective is tolerance-legal anywhere near 0.988–0.989; both runs sit in that band. `cresc4` moved 85 → 206 in an earlier revision and is bit-identical now (see "Rejected" above).

### Repeated-solve benchmark

`benchmarks/warmstart`, 9 families × 3 step scales = 27 parametric paths, 7 warm re-solves per path (189 warm solves per mode). Every mode sees the *same* per-step seed, so the only variable is how much of the dual state is handed back.

| seed | | iters | obj evals | jac evals |
|---|---|---|---|---|
| **full** (x, λ, z_L, z_U, μ) | before | 715 | 1207 | 994 |
| | **after** | **677** | **1068** | **950** |
| | | −5.3% | −11.5% | −4.4% |
| **partial** (bound mults kept, `lagrange` dropped) | before | 718 | 1210 | 997 |
| | **after** | **709** | **1167** | **988** |
| | | −1.3% | −3.6% | −0.9% |
| **primal only** (x and μ) | before | 1102 | 2421 | 1492 |
| | **after** | 1102 | 2421 | 1492 |
| | | *bit-identical by construction* | | |
| **cold** (reference, no seed) | before/after | 2036 | 2619 | 2199 |

`partial` is the issue's own acceptance criterion — "partial warm starts outperform zero-filled duals on a repeated-solve benchmark" — and it is met, though modestly. `full` is what `pounce.WarmStart` produces and what the docs recommend, and is where the change pays.

Only 8 of the 108 (path × mode) cells move at all. Every one:

| path | mode | before | after |
|---|---|---|---|
| `hanging_chain/tiny` | full | 49 | **14** |
| `hanging_chain/small` | full | 65 | **57** |
| `hanging_chain/large` | full | 68 | **47** |
| `rosenbrock_ring/small` | partial | 41 | **35** |
| `rosenbrock_ring/large` | partial | 40 | **37** |
| `rosenbrock_ring/tiny` | full | 28 | 37 |
| `rosenbrock_ring/small` | full | 42 | 45 |
| `rosenbrock_ring/large` | full | 36 | 50 |

`hanging_chain` is the flipping-contact family — 15 inequality rows whose slack multipliers were exactly the block nobody could seed, so it is where reconstructing them pays most. `rosenbrock_ring` is nonconvex with a single activation switch crossed along the path; the reconstructed slack multipliers reroute it, better in `partial` and worse in `full`. Net over the family: `full` +23, `partial` −9.

### Exact same-model restarts

Solve, then re-solve the identical model from its own answer. Summed over the 9 families:

| seed | before | after |
|---|---|---|
| full | 6 | **2** |
| partial | 9 | **7** |
| primal only | 17 | 17 |

No family regresses. `hanging_chain` and `rosenbrock_ring` go from 2 iterations to **0** on a full restart — the supplied point is now recognised as the KKT point it is.

### Regressions, named — now tracked as #617 and #618

Two, both on deliberately-bad seeds, both accepted here as the cost of the fix. Neither is closed by this PR.

**(a) Corrupted dual seeds cost more than they used to — #617.** Seeding the exact primal solution together with multipliers corrupted by `N(0,1e4)` noise, summed over 9 families: **18 → 71 iterations**. Per family the worst is `hanging_chain` 3 → 13; every case still converges with status `SolveSucceeded` to the same objective, and lands within ~1.4× of the cold solve (cold sum 81). Mechanism: the reconstructed slack multipliers are derived from the seeded `y`, so they inherit its corruption, where before they were discarded in favour of a constant. The pre-#606 "2 iterations" was not skill — it was throwing the bad duals away. Follow-up: reject a seeded dual block whose implied complementarity is orders away from the rest of the point, rather than reconstructing off it.

**(b) Mildly stale seeds regress slightly — #618.** Seeding each family from the far end of its own 4×-scale path: **118 → 127** iterations summed over 9 families with a full seed (`redundant_rows` 3 → 7 is the worst, `degenerate_corner` 9 → 11, `hanging_chain` 8 → 10, `moving_bound_qp` 11 → 12; the rest unchanged). These points are stale enough for the reconstruction to fire but not stale enough to trip the 10× `μ` escalation, so they get the reconstruction's trajectory change without the recentering that is supposed to pay for it. The *strongly* stale case is the opposite — HS071 from the far corner goes from `RestorationFailed` to `SolveSucceeded` — so the gap is in the middle of the staleness range. Follow-up: a second, gentler escalation band.

## Tests

`crates/pounce-algorithm/tests/warm_start_recentering.rs`, 7 tests over an HS071 whose starting point *and* dual seeds are both settable.

Four of them fail on the parent commit `70bf53de` for behavioural reasons. Verified in a `git worktree` at that commit with the assertions that need new API (`BlockVerdict`, `warm_start_diagnostics()`, the `warm_start_recentering` option itself) stripped, so the failures are behavioural and not compile errors. Exactly what they printed:

```
exact restart: none=2 residual=2 cold=8
  panicked: residual=2 vs none=2                 (assert residual < none)
partial: none=3 residual=3 cold=8
  panicked: residual=3 vs none=3                 (assert residual < none)
stale: status=RestorationFailed iters=36 cold=8
  panicked                                       (assert SolveSucceeded)
warm_start_same_structure=yes -> refusal None
  panicked                                       (assert is_some)
```

On this branch the same four read `1 vs 2`, `2 vs 3`, `SolveSucceeded in 43`, and a refusal naming `GetWarmStartIterate` and issue 606.

The remaining three pin the kill switch (`recentering=none` leaves every block alone and does not move `μ`), the deliberate non-behaviour on a primal-only seed, and a registry invariant.

**Deliberately not a regression test:** `every_registered_warm_start_option_is_consumed_or_refused` — it mirrors gh#604's `Initialization` invariant one category over, but on the parent commit the two dead flags *are* read (into fields nothing consumes), so it would have passed there. It guards the next warm-start option, not this one.

**Deliberately not pinned:** the corpus-level iteration counts above. They come from `benchmarks/warmstart`, which is a benchmark suite rather than a test, needs the Python extension built, and takes minutes. The numbers in this PR are the record; nothing in CI will notice if they drift. That gap is the same one the fixture sweep exists to cover for the cold path, and the third sweep regime above is the closest CI-shaped proxy for it.

Also run clean: `cargo fmt --all -- --check`, `cargo clippy` (no new warning categories in the changed files — 6 `unwrap_used` warnings against the baseline's 4, both new ones in `scatter`, matching the type-tested idiom already in `resolve_nan_seeds` beside it), `cargo test --workspace --exclude pounce-hsl`, the Python suite (777 passed, 34 skipped), `scripts/check-release-consistency.sh`, `scripts/check-docs-consistency.sh`.

---

- [x] Book page under `docs/src/` updated (`initialization.md`; already in `SUMMARY.md`)

---

### DIFF
--- a/crates/pounce-algorithm/src/init/warm_start.rs
+++ b/crates/pounce-algorithm/src/init/warm_start.rs
@@ -261,15 +261,26 @@ impl IterateInitializer for WarmStartIterateInitializer {
             // circular, and measurably so (see the fn docs). Only a
             // caller-supplied `y` earns it.
             if diag.eq_duals != BlockVerdict::Reconstructed {
-                refine_bound_duals_from_stationarity(data, cq, nlp, &self.opts, mu_hat, &unseeded);
-                diag.stationarity_split = true;
+                // Reported from the function's own return value, not
+                // from the fact that it was called: it returns early
+                // with nothing to do when every bound multiplier
+                // arrived seeded, and claiming the split ran there
+                // made `info["warm_start"]` say something untrue
+                // (gh#606 review).
+                diag.stationarity_split = refine_bound_duals_from_stationarity(
+                    data, cq, nlp, &self.opts, mu_hat, &unseeded,
+                );
             }
             Some(mu_hat)
         } else if self.opts.recentering == WarmStartRecentering::Residual {
             diag.primal_residual = cq.borrow().curr_primal_infeasibility_max();
             diag.bound_duals = BlockVerdict::Unseeded;
             diag.eq_duals = BlockVerdict::Unseeded;
-            Some(safe_mu(data.borrow().curr_mu))
+            // Only the `is_some()` is read on this path — `final_mu`
+            // re-reads `curr_mu` itself — so this carried a clamp that
+            // was computed and thrown away. Left as the plain value so
+            // it does not read as a policy it never was.
+            Some(data.borrow().curr_mu)
         } else {
             None
         };
@@ -538,13 +549,15 @@ fn refine_bound_duals_from_stationarity(
     opts: &WarmStartOptions,
     mu_hat: Number,
     unseeded: &[Vec<bool>; 4],
-) {
+) -> bool {
+    // `true` only when the split actually rewrote a multiplier, so the
+    // caller can report `stationarity_split` honestly (gh#606 review).
     if unseeded.iter().all(|m| m.is_empty()) {
-        return;
+        return false;
     }
     let curr = match data.borrow().curr.clone() {
         Some(c) => c,
-        None => return,
+        None => return false,
     };
 
     // r_x = ∇f + J_cᵀ y_c + J_dᵀ y_d, and r_s = −y_d.
@@ -621,7 +634,7 @@ fn refine_bound_duals_from_stationarity(
     }
 
     if rebuilt.iter().all(|r| r.is_none()) {
-        return;
+        return false;
     }
     let pick = |i: usize, orig: &Rc<dyn Vector>| -> Rc<dyn Vector> {
         rebuilt[i].clone().unwrap_or_else(|| Rc::clone(orig))
@@ -637,6 +650,7 @@ fn refine_bound_duals_from_stationarity(
         pick(3, &curr.v_u),
     );
     data.borrow_mut().set_curr(new_curr);
+    true
 }
 
 /// `sign · Pᵀ r`, as a plain slice of length `n_out`. `P` is one of the
@@ -782,11 +796,24 @@ fn final_mu(data: &IpoptDataHandle, cq: &IpoptCqHandle, diag: &mut WarmStartDiag
     diag.dual_residual = inf_du;
 
     let mu_in = data.borrow().curr_mu;
-    let mut mu = mu_in;
-    if compl.is_finite() && compl > MU_ESCALATION_TRIGGER * mu_in {
-        mu = compl;
+    // A non-finite or non-positive barrier is not a setting, it is a
+    // broken one, and the band's fallback still applies to it.
+    if !mu_in.is_finite() || mu_in <= 0.0 {
+        return MU_CEILING;
     }
-    safe_mu(mu).max(mu_in.min(MU_CEILING))
+    // Clamp the *measurement*, never the caller's setting (gh#606
+    // review). The previous form applied `safe_mu` to the pass-through
+    // value too, so an explicit `mu_init = 1.0` silently started the
+    // solve at `MU_CEILING` — on every warm start, escalation or not,
+    // and without printing anything, since the `wmu` token only fires
+    // when μ goes up. `.max(mu_in)` keeps the floor semantics this
+    // function documents without capping a barrier the caller chose on
+    // purpose.
+    if compl.is_finite() && compl > MU_ESCALATION_TRIGGER * mu_in {
+        safe_mu(compl).max(mu_in)
+    } else {
+        mu_in
+    }
 }
 
 /// Clamp a candidate μ into the band a warm start may use, and reject
@@ -847,9 +874,17 @@ fn scatter(v: &mut dyn Vector, src: &[Number]) -> bool {
 /// equality-multiplier block that no caller seeded. An uninitialized
 /// block counts as zero: that is what the clamp step below fills it
 /// with.
+///
+/// An **empty** block is zero vacuously, and saying so is what makes
+/// [`reconstruct_eq_duals`] work on a model that has only equality
+/// rows or only inequality rows. Returning `false` there — "this
+/// zero-dimension block carries a seed" — short-circuited the
+/// reconstruction on every single-block model and then reported the
+/// result as `Accepted` (gh#606 review). The two [`any_dual_seeded`]
+/// call sites already guard on `dim() > 0`, so they are unaffected.
 fn is_identically_zero(v: &Rc<dyn Vector>) -> bool {
     if v.dim() == 0 {
-        return false;
+        return true;
     }
     match flatten(&**v) {
         Some(vals) => vals.iter().all(|e| *e == 0.0),
--- a/crates/pounce-algorithm/src/init/warm_start.rs
+++ b/crates/pounce-algorithm/src/init/warm_start.rs
@@ -22,12 +22,57 @@
 //! `mult_init_max`, `target_mu`. `mult_bound_push` floors the four
 //! bound-multiplier blocks (mirroring upstream's `ElementWiseMax`
 //! with `warm_start_mult_bound_push`): a user-seeded `z = 0` would
-//! otherwise start the barrier on its boundary. The remaining knobs
-//! (`entire_iterate`, `same_structure`) are parsed for Ipopt option
-//! compatibility but not yet consumed — they require the
-//! `GetWarmStartIterate` TNLP surface, which pounce does not expose.
+//! otherwise start the barrier on its boundary.
+//!
+//! # Residual-adaptive recentering (gh#606)
+//!
+//! `warm_start_recentering=residual` (the default) adds a pass over
+//! the *supplied* point before the clamps: measure what was actually
+//! handed in, reconstruct what is missing, and choose μ from the
+//! measurement rather than from a universal constant.
+//!
+//! 1. **Measure.** `inf_pr` comes first because it is the one residual
+//!    that does not depend on the duals, so it is meaningful even when
+//!    every multiplier block is absent. It is reported rather than
+//!    acted on (step 4 explains why).
+//! 2. **Reconstruct bound multipliers.** An entry that arrives as
+//!    exactly `0` (or NaN) is not a legal barrier multiplier — the
+//!    barrier needs `z > 0` — so it was never a seed. Upstream floors
+//!    it at the constant `warm_start_mult_bound_push`; here it takes
+//!    `μ̂ / slack` instead, the same complementarity relation the
+//!    solver is about to enforce. Seeded (strictly positive) entries
+//!    are left alone.
+//! 3. **Reconstruct equality multipliers.** A `y` block that is
+//!    identically zero is likewise unseeded, and is re-derived from
+//!    stationarity by the same regularized least-squares augmented
+//!    solve the cold path uses ([`LeastSquareMults`]) — now with the
+//!    reconstructed `z` in its right-hand side, so the estimate is not
+//!    forced to absorb the bound multipliers.
+//! 4. **Choose μ.** With the point complete, μ is raised to the
+//!    measured `avrg_compl` when that overshoots what `mu_init` asked
+//!    for by more than [`MU_ESCALATION_TRIGGER`], clamped to
+//!    `[MU_FLOOR, MU_CEILING]`. A KKT-quality point
+//!    measures its own converged complementarity and keeps it; a stale
+//!    one, whose multipliers and slacks no longer pair up, measures a
+//!    large one and gets a correspondingly loose barrier — the "safely
+//!    fall back to stronger recentering" half. The primal and dual
+//!    residuals are deliberately *not* in that max; see [`final_mu`].
+//!    `warm_start_target_mu` still wins outright when set.
+//!
+//! `warm_start_recentering=none` restores the pre-gh#606 behaviour
+//! exactly: constant floor, zero-filled `y`, μ untouched. It is the
+//! kill switch for this whole block.
+//!
+//! Every branch above records what it did on
+//! [`WarmStartDiagnostics`], which lands on `IpoptData` and is
+//! readable afterwards through
+//! `IpoptApplication::warm_start_diagnostics()`.
+//!
+//! [`LeastSquareMults`]: crate::eq_mult::least_square::LeastSquareMults
 
-use crate::alg_builder::WarmStartOptions;
+use crate::alg_builder::{WarmStartOptions, WarmStartRecentering};
+use crate::eq_mult::least_square::LeastSquareMults;
+use crate::eq_mult::r#trait::EqMultCalculator;
 use crate::init::default::push_x_into_interior;
 use crate::init::r#trait::IterateInitializer;
 use crate::ipopt_cq::IpoptCqHandle;
@@ -35,11 +80,109 @@ use crate::ipopt_data::IpoptDataHandle;
 use crate::ipopt_nlp::IpoptNlp;
 use crate::iterates_vector::IteratesVector;
 use crate::kkt::aug_system_solver::AugSystemSolver;
+use pounce_common::types::Number;
 use pounce_linalg::Vector;
 use pounce_linalg::compound_vector::CompoundVector;
 use pounce_linalg::dense_vector::{DenseVector, DenseVectorSpace};
 use std::cell::RefCell;
 use std::rc::Rc;
+
+/// Hard floor on the residual-derived μ. Below this the barrier term
+/// is at the noise level of a `tol=1e-8` solve's complementarity and
+/// carries no information; letting a measurement drive μ further down
+/// would start the solve on a boundary the line search then has to
+/// climb back off.
+const MU_FLOOR: Number = 1e-11;
+
+/// Hard ceiling on the residual-derived μ: upstream's registered
+/// `mu_init` default. A warm start that measures badly should degrade
+/// *to* a cold start, never past one.
+const MU_CEILING: Number = 0.1;
+
+/// How far the measured complementarity has to exceed `mu_init` before
+/// it is allowed to override it.
+///
+/// Moving μ reroutes the entire trajectory, so the move has to be
+/// worth it. The cases this exists for — a stale seed, a caller who
+/// kept only the primal point — miss by three to six orders of
+/// magnitude; a good seed misses by a factor of two, and overriding
+/// there buys nothing and costs whatever the reroute costs. On
+/// `cresc4` (a nonconvex model that enters restoration on iteration 2,
+/// where small perturbations compound) a measurement of `5e-7` against
+/// a `mu_init` of `1e-7` moved μ by half an order and the solve from
+/// 85 iterations to 206 — same status, same objective, 2.4x the work.
+/// With this gate that model is bit-identical again, and the
+/// three-order cases still fire.
+const MU_ESCALATION_TRIGGER: Number = 10.0;
+
+/// What happened to one multiplier block of the supplied warm point.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum BlockVerdict {
+    /// The block has dimension zero — the model has no such block.
+    #[default]
+    Absent,
+    /// Every entry arrived seeded and was kept (modulo the clamps).
+    Accepted,
+    /// Some or all entries arrived unseeded and were rebuilt.
+    Reconstructed,
+    /// A reconstruction ran and was thrown away (it exceeded
+    /// `constr_mult_init_max`, or the augmented solve failed); the
+    /// block fell back to the pre-gh#606 constant fill.
+    Discarded,
+    /// The caller supplied no dual information *at all*, so there was
+    /// nothing to reconstruct this block from and it kept the
+    /// pre-gh#606 constant fill. See [`any_dual_seeded`].
+    Unseeded,
+}
+
+/// What the warm-start initializer accepted, reconstructed, or
+/// discarded, and the residuals it based those calls on (gh#606).
+///
+/// Written once per solve, at initialization. `mu_in` is what
+/// `mu_init` / `warm_start_target_mu` asked for; `mu_out` is what the
+/// iterate actually starts at.
+#[derive(Debug, Clone)]
+pub struct WarmStartDiagnostics {
+    /// `‖c(x)‖_∞` of the supplied primal point, measured before any
+    /// dual reconstruction (it does not depend on the duals).
+    pub primal_residual: Number,
+    /// `‖∇_x L‖_∞` after reconstruction.
+    pub dual_residual: Number,
+    /// Average complementarity after reconstruction.
+    pub complementarity: Number,
+    pub mu_in: Number,
+    pub mu_out: Number,
+    pub bound_duals: BlockVerdict,
+    pub eq_duals: BlockVerdict,
+    /// Bound-multiplier entries that arrived unseeded and were filled
+    /// from `μ̂ / slack` rather than from the constant floor.
+    pub bound_duals_reconstructed: usize,
+    /// `true` when the unseeded bound multipliers were re-derived from
+    /// the stationarity identity rather than left at `μ̂ / slack`. That
+    /// needs a caller-supplied `y`; see
+    /// [`refine_bound_duals_from_stationarity`].
+    pub stationarity_split: bool,
+    /// `true` when `warm_start_recentering=none` turned all of the
+    /// above off and the fields are the legacy constants.
+    pub recentering_disabled: bool,
+}
+
+impl Default for WarmStartDiagnostics {
+    fn default() -> Self {
+        Self {
+            primal_residual: Number::NAN,
+            dual_residual: Number::NAN,
+            complementarity: Number::NAN,
+            mu_in: Number::NAN,
+            mu_out: Number::NAN,
+            bound_duals: BlockVerdict::Absent,
+            eq_duals: BlockVerdict::Absent,
+            bound_duals_reconstructed: 0,
+            stationarity_split: false,
+            recentering_disabled: false,
+        }
+    }
+}
 
 pub struct WarmStartIterateInitializer {
     opts: WarmStartOptions,
@@ -67,9 +210,9 @@ impl IterateInitializer for WarmStartIterateInitializer {
     fn set_initial_iterates(
         &mut self,
         data: &IpoptDataHandle,
-        _cq: &IpoptCqHandle,
+        cq: &IpoptCqHandle,
         nlp: &Rc<RefCell<dyn IpoptNlp>>,
-        _aug_solver: &mut dyn AugSystemSolver,
+        aug_solver: &mut dyn AugSystemSolver,
     ) -> bool {
         // Two entry points share this initializer: the re-optimize path
         // (curr.x carries values from the prior solve) and the first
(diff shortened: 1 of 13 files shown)


crates/pounce-algorithm/src/init/warm_start.rs
```diff
@@ -129,14 +317,547 @@ impl IterateInitializer for WarmStartIterateInitializer {
...
+        return;
+    }
+    // Upstream's own guard from the cold path: with as many equalities
+    // as variables the least-squares system is square and the estimate
+    // is not a projection of anything.
+    if n_yc == curr.x.dim() {
+        diag.eq_duals = BlockVerdict::Accepted;
+        return;
+    }
+    let seeded = !is_identically_zero(&curr.y_c) || !is_identically_zero(&curr.y_d);
```
jkitchin, line 712: **Step 3 never runs on a model that has only one of the two constraint blocks.**

`is_identically_zero` returns `false` for a zero-dimension vector (line 850-852), so on a model with no inequality rows (`y_d.dim() == 0`) this reads `!true || !false` → `seeded = true`, and symmetrically for a model with no equality rows. The least-squares reconstruction is skipped, and `diag.eq_duals` is set to `Accepted` — "every entry arrived seeded and was kept" — for a block nothing seeded.

Equality-only and inequality-only models are a large share of real NLPs, so this is most of the feature's reach. Reproduced against this branch's head on `min (x0-3)² + (x1-3)²  s.t.  x0 + x1 = 2, 0 ≤ x ≤ 10` (one equality row, no inequality rows), warm-started with `z_l`/`z_u` seeded and `lagrange` left out:

```
PROBE eq-only: WarmStartDiagnostics { primal_residual: 0.0, dual_residual: 4.0,
  complementarity: 2.5, mu_in: 1e-7, mu_out: 0.1,
  bound_duals: Accepted, eq_duals: Accepted,
  bound_duals_reconstructed: 0, stationarity_split: true, ... }
```

`eq_duals` should be `Reconstructed`. `dual_residual: 4.0` after initialization is the stationarity residual the reconstruction exists to remove.

`warm_start_recentering.rs` cannot see this: its HS071 has `m = 2` with one equality row and one inequality row, so both blocks are non-empty in every test.

Suggested fix — guard on the dimension the way `any_dual_seeded` already does:

```rust
let seeded = (curr.y_c.dim() > 0 && !is_identically_zero(&curr.y_c))
    || (curr.y_d.dim() > 0 && !is_identically_zero(&curr.y_d));
```

or make `is_identically_zero` return `true` for an empty block, which is both the vacuous reading and what its own doc comment says ("`true` when every entry is exactly `0`"); the `any_dual_seeded` call sites already guard on `dim() > 0`, so nothing else changes.

Worth a regression test on a single-block model either way.

---

crates/pounce-algorithm/src/init/warm_start.rs
```diff
@@ -129,14 +317,547 @@ impl IterateInitializer for WarmStartIterateInitializer {
...
+    // Recorded but deliberately not fed into μ — see above. It is the
+    // number that says whether the reconstruction worked.
+    diag.dual_residual = inf_du;
+
+    let mu_in = data.borrow().curr_mu;
+    let mut mu = mu_in;
+    if compl.is_finite() && compl > MU_ESCALATION_TRIGGER * mu_in {
+        mu = compl;
+    }
+    safe_mu(mu).max(mu_in.min(MU_CEILING))
```
jkitchin: **An explicit `mu_init` outside `[MU_FLOOR, MU_CEILING]` is silently overridden on every warm start, escalation or not.**

`safe_mu` clamps into `[1e-11, 0.1]`, and it is applied to the pass-through value as well as to the escalated one. When no escalation fires, `mu = mu_in`, and the result is `min(mu_in, 0.1)` — so a caller who deliberately set `mu_init = 1.0` starts the solve at `0.1`. Measured on this branch's head, same probe model, `mu_init = 1.0` with `warm_start_init_point=yes` and no escalation triggered (`compl = 2.5 < 10 · mu_in`):

```
PROBE mu: mu_in=1 mu_out=0.1
```

Three reasons to treat that as a defect rather than a policy:

- It contradicts this function's own doc comment eight lines up — "`mu_in` is a floor, never a ceiling" — and the registered option help, which scopes the clamp to the measured value ("raises mu to the measured average complementarity (clamped to [1e-11, 0.1])").
- It is silent. The `wmu` info-string token only fires when μ goes *up* (`mu_out > mu_in * 10.0`), so a μ that was quietly lowered by an order of magnitude prints nothing on the iteration line. That is the gh#544 shape the PR body opens with.
- It applies even to a primal-only seed, where `mu_hat` is `Some` but nothing was reconstructed — so the "primal only … bit-identical by construction" row of the benchmark table holds only because those `mu` values happened to land inside the band.

Suggested fix — clamp the measurement, not the caller's setting:

```rust
if compl.is_finite() && compl > MU_ESCALATION_TRIGGER * mu_in {
    safe_mu(compl).max(mu_in)
} else {
    mu_in
}
```

(`.max(mu_in)` keeps the documented floor semantics without capping an `mu_init` the caller chose on purpose.)

---

crates/pounce-algorithm/src/init/warm_start.rs
```diff
@@ -89,6 +232,48 @@ impl IterateInitializer for WarmStartIterateInitializer {
...
+        let mu_hat = if dual_info {
+            let (mu_hat, unseeded) = recenter_from_residuals(data, cq, &self.opts, &mut diag);
+            reconstruct_eq_duals(data, cq, nlp, aug_solver, &self.opts, &mut diag);
+            // The split below reads `y`; running it against a `y` this
+            // same pass just derived *from* the provisional `z` is
+            // circular, and measurably so (see the fn docs). Only a
+            // caller-supplied `y` earns it.
+            if diag.eq_duals != BlockVerdict::Reconstructed {
+                refine_bound_duals_from_stationarity(data, cq, nlp, &self.opts, mu_hat, &unseeded);
+                diag.stationarity_split = true;
```
jkitchin: Minor, but it makes the diagnostics say something untrue: `stationarity_split` is set unconditionally after the call, while `refine_bound_duals_from_stationarity` returns immediately when `unseeded.iter().all(|m| m.is_empty())` — i.e. when every bound multiplier arrived seeded and there was nothing to refine. The probe run above shows the mismatch directly: `bound_duals: Accepted, bound_duals_reconstructed: 0, stationarity_split: true`.

Setting the flag from inside the function (or gating it on `diag.bound_duals_reconstructed > 0`) keeps `info["warm_start"]` honest about what actually ran.

---

jkitchin (commented):
Reviewed alongside #619 and #623. The design here is well argued and the "four things tried and rejected, each on a number" section is exactly the record the next maintainer needs — the `μ ≥ κ·max(inf_pr, inf_du)` rejection in particular. The benchmark and the three-regime fixture sweep answer the CLAUDE.md trajectory requirement, and #617/#618 track the two named regressions.

Holding the merge on two defects found by reading and then reproduced against this branch's head (`4f06973`, `cargo test -p pounce-algorithm`, probe model in the inline comments):

1. **`reconstruct_eq_duals` is a no-op on any model that has only equality rows or only inequality rows.** `is_identically_zero` returns `false` on a zero-dimension block, so `seeded` short-circuits to `true` and the least-squares reconstruction — the PR's step 3 — never runs. The diagnostics then report `eq_duals: Accepted` for a block nothing seeded. The HS071 test model has both blocks non-empty, so nothing in the suite covers it.

2. **`final_mu` clamps the caller's `mu_init` into `[1e-11, 0.1]` even when no escalation fires**, so `mu_init = 1.0` silently becomes `0.1` on every warm start under the new default. That contradicts the function's own "`mu_in` is a floor, never a ceiling" and the registered option help, and it prints nothing (the `wmu` tag only fires on an increase).

Plus one cosmetic: `stationarity_split` is reported `true` when the refinement returned early with nothing to do.

Neither is a wrong-answer bug — (1) is a silently missing feature with misreporting attached, (2) is a silent override of an explicit user option — but both are the class of thing this PR's own opening argument is about, so they seem worth fixing before this lands rather than after.

Two smaller notes, no action needed unless you want it:

- `docs/src/initialization.md` says μ "is raised to the point's measured average complementarity when that exceeds `mu_init`", which omits the 10× `MU_ESCALATION_TRIGGER` gate that the option help does state. Worth one clause so the two pages agree.
- Once (1) is fixed, the `benchmarks/warmstart` numbers in the PR body will move — the corpus almost certainly contains single-block families that were not reaching the reconstruction.

#619 and #623 are being merged; this branch will need a `main` merge for the `CHANGELOG.md` `[Unreleased]` anchor (and, from #619, `application.rs` — the two hunks are disjoint and auto-merge locally) whichever way the fixes go.

---

jkitchin:
All three fixed in `96ee510`, plus the doc clause, and the branch is merged up to `0b09ca4` (#619, #614 and #623 all landed while this was in review). Both defects reproduced first, exactly as you described them.

## 1. `reconstruct_eq_duals` on single-block models

Fixed by making `is_identically_zero` return `true` for an empty block — your second suggestion, and the one its own doc comment already claims. Both `any_dual_seeded` call sites guard on `dim() > 0`, so nothing else moves, and `reconstruct_eq_duals` already returns early on `n_yc + n_yd == 0`.

**Your "almost certainly contains single-block families" was an understatement: it is 12 of the 14 default-tier families in `benchmarks/warmstart`.** Only `redundant_rows` has both blocks; `double_well_chain` has no constraints at all.

| family | n_eq | n_ineq | single-block |
|---|---|---|---|
| `simplex_proj` | 1 | 0 | yes |
| `moving_bound_qp` | 3 | 0 | yes |
| `degenerate_corner` | 0 | 3 | yes |
| `degenerate_vertex` | 0 | 12 | yes |
| `hanging_chain` | 0 | 15 | yes |
| `rosenbrock_ring`, `..._cycle` | 0 | 1 | yes |
| `nmpc_vanderpol` | 32 | 0 | yes |
| `mpc_horizon_{10,20,40,80}` | 22–162 | 0 | yes |
| `redundant_rows` | 2 | 3 | **no** |
| `double_well_chain` | 0 | 0 | n/a |

Verdicts before and after, same binary pair, partial seed (`lagrange` dropped):

| family | before | after |
|---|---|---|
| `simplex_proj` | `eq_duals: accepted`, `split: true` | `reconstructed`, `split: false` |
| `mpc_horizon_10` | `accepted`, `split: true` | `reconstructed`, `split: false` |
| `nmpc_vanderpol` | `accepted`, `split: true` | `reconstructed`, `split: false` |
| `hanging_chain` | `unseeded` | `unseeded` (no bounds to seed from) |
| `redundant_rows` | `reconstructed` | `reconstructed` |

## 2. `mu_init` clamped on the pass-through path

Fixed with your suggested form — clamp the measurement, `.max(mu_in)` for the floor — plus an early return keeping the old fallback for a non-finite or non-positive `mu_in`, which `safe_mu` used to absorb. Your `PROBE mu: mu_in=1 mu_out=0.1` reproduces verbatim as a test failure on the parent.

You were also right that it reached the primal-only path: `Some(safe_mu(data.borrow().curr_mu))` at the `dual_info == false` branch computed a clamp that was then **discarded** — only its `is_some()` is read, since `final_mu` re-reads `curr_mu` itself. So that row of the benchmark table was safe for a different reason than it looked. It is now the plain value, and the clamp it implied is gone.

## 3. `stationarity_split`

Now returned from `refine_bound_duals_from_stationarity` rather than set at the call site.

Worth noting for the test: this one is **unreachable on HS071**, which is why it was accidentally right there — the slack-bound block is never seedable, so something is always rebuilt. Pinning it needed a model with *no* bound multipliers at all (free variables) and a seeded `y`, so the split is still called and still has nothing to do.

## Tests

Three new tests, all failing on `4f06973` for the stated reason:

```
a_single_constraint_block_still_reaches_the_reconstruction
  OneBlock eq-only: eq_duals=Accepted reconstructed=0 inf_du=4.000000001505904e0 split=true
  assertion `left == right` failed: left: Accepted  right: Reconstructed

an_explicit_mu_init_above_the_ceiling_survives_when_nothing_escalates
  OneBlock mu pass-through: mu 1e0 -> 1e-1 compl=5.75e-9
  assertion `left == right` failed: left: 0.1  right: 1.0

stationarity_split_is_reported_only_when_it_runs
  OneBlock free+seeded-y: bound_duals=Absent eq_duals=Accepted reconstructed=0 split=true
  the flag must track the work, not the call site
```

Your `dual_residual: 4.0` probe reproduces exactly.

## What it is worth, and what it costs

You predicted the benchmark numbers would move. They move less than either of us expected, and the reason is worth recording.

**On the residual the warm point is handed to the solver with, the repair is enormous:**

| family | `inf_du` after init, before | after |
|---|---|---|
| `mpc_horizon_10` | 2.27e+1 | **3.74e-6** |
| `nmpc_vanderpol` | 9.73e+0 | **4.27e-6** |
| `simplex_proj` | 4.11e-1 | **2.00e-2** |

**In iterations it is close to a wash.** Partial seed, 42 paths, `warm-ipm`: **3052 → 3050**, 8 paths moving, 0 correctness changes. Best `rosenbrock_ring` at `tiny` 49 → 43; worst **`nmpc_vanderpol` at `large` 195 → 209 (+14)**. A full-seed run does not move at all (2936 → 2936), which is expected — a supplied `lagrange` was never the block at issue.

I have added that +14 to #618 rather than filing a near-duplicate: it looks like the same "reconstruction fires, μ escalation does not" band that issue already describes. Flagging it explicitly because it is a real cost of this fix and I would rather it be visible than averaged away in a −2 total.

**The PR body's `677 / 715 / 1068` figures I could not re-derive**, and that is a gap rather than a reassurance: they came from a seed-mode harness that was never committed, so nothing in the tree reproduces them. The numbers above are from the committed `benchmarks/warmstart` arms plus a 20-line `PartialSeedAdapter` subclass that drops `mult_g`. Making that a real arm looks like the right fix and I have not done it here — say the word and I will, or file it.

## Sweep and checks

Re-swept against **`0b09ca4`** (the current merge base, not the `70bf53de` the original numbers were taken on): cold **bit-identical across all 57**, warm default-push **bit-identical**, tightened-push moves **the same three lines and no others** — `hs13_bigstart` 30→29, `jit1_boxed` 16→17, `jit1_node` 19→21. Same result against `369f13e` and `70bf53de`, so the claim survived both intervening merges.

`cargo fmt` clean, clippy clean, `cargo test --workspace --exclude pounce-hsl` 2949 passed, Python **815 passed / 36 skipped** (the union with #623's 34), both consistency scripts OK.

One failure to disclose: `optimize_hs71::hs071_max_cpu_time_terminates` fails intermittently (1 of 6 runs here). It sets `max_cpu_time = 1e-12` and expects `MaximumCpuTimeExceeded`, returning `SolveSucceeded` when the solve beats the timer's granularity. It fails at the same rate on unmodified `main` and touches nothing here.

## Merge with #623

Merged up cleanly. `_warm_start.py` conflicted in exactly the three places #623's body predicted, resolved as it prescribed (keep both, `recentering` first); `problem.rs`, `initialization.md` and `CHANGELOG.md` auto-merged. The interaction that mattered, on the merged tree:

```
#606 recentering key present: ['warm_start_recentering']
option list restored exactly: True
warm_start_recentering left behind: False
pristine iters=17   after-warm iters=17
```

`warm_start_recentering` is scoped by #623's overlay with no change on either side, because the overlay snapshots the option list rather than naming keys.

---

PR #626, opened 2026-08-16
CasADi nlpsol plugin, nonlinear-variable masks, and a C-API abort fix

### LINKED ISSUE #624: [feature-request] Support Ipopt-compatible nonlinear-variable masks for limited-memory Hessians
> **AI assistance disclosure:** I used an AI assistant to help research and draft this feature request. I reviewed and edited the content before posting it. Apologies in advance if POUNCE’s contribution policy prohibits AI-assisted issue text; please let me know and I will revise or withdraw it.

## Feature request

I would like to request support for Ipopt-compatible nonlinear-variable subsets in the limited-memory Hessian path.

POUNCE already includes the corresponding `TNLP` methods:

- `get_number_of_nonlinear_variables`
- `get_list_of_nonlinear_variables`

These currently appear to default to treating all variables as nonlinear and are not yet used by the IPM L-BFGS implementation.

In Ipopt, these callbacks allow L-BFGS updates to be restricted to variables that occur nonlinearly. CasADi exposes this through its `pass_nonlinear_variables` option and passes the resulting mask through its Ipopt `TNLP` implementation. For reference:

- [CasADi: forwarding the nonlinear-variable count and indices through `IpoptUserClass`](https://github.com/casadi/casadi/blob/main/casadi/interfaces/ipopt/ipopt_nlp.cpp#L334-L345)
- [CasADi: deriving and returning the nonlinear-variable list](https://github.com/casadi/casadi/blob/main/casadi/interfaces/ipopt/ipopt_interface.cpp#L659-L684)

This can improve performance for large NLPs containing many variables that enter only linearly.

The eventual use case is a POUNCE CasADi `Nlpsol` plugin that can forward CasADi’s nonlinear-variable mask to POUNCE while preserving behavior comparable to the Ipopt plugin.

It would be useful for this to support:

- arbitrary, noncontiguous nonlinear-variable subsets;
- the existing `TNLP` callbacks, with the default continuing to treat all variables as nonlinear;
- limited-memory Hessian updates restricted to the declared subset;
- correct handling through fixed-variable removal and other variable transformations;
- an additive C API entry point through which a CasADi plugin can provide the mask;
- `num_linear_variables` as the Ipopt-compatible contiguous-prefix fallback, if appropriate;
- no change to exact-Hessian behavior.

A possible C API shape might be:

```c
Bool IpoptSetNonlinearVariables(
    IpoptProblem problem,
    const Bool* is_nonlinear
);
```

The specific API is not important to this request; the main functional requirement is that an external CasADi plugin can pass its nonlinear-variable mask into POUNCE.

Expected compatibility would be:

- no supplied mask preserves current behavior;
- a supplied mask affects only limited-memory Hessian approximation;
- callback-provided nonlinear-variable information takes precedence over `num_linear_variables`, as in Ipopt;
- masked and full-space modes solve the same NLP and return equivalent KKT solutions.

This would close an Ipopt feature-parity gap and provide an important performance enabler for using POUNCE through CasADi.

jkitchin:
Fixes #624.

## Problem

#624 asks for Ipopt-compatible nonlinear-variable subsets in the limited-memory
Hessian, with the stated end goal of "a POUNCE CasADi `Nlpsol` plugin that can
forward CasADi's nonlinear-variable mask". Three things stood between us and
that:

1. `TNLP::get_number_of_nonlinear_variables` / `get_list_of_nonlinear_variables`
   were declared and forwarded through every decorator, but read nowhere. So was
   `num_linear_variables`, which sat on the unimplemented-options list.
2. There was no CasADi interface at all, and no analysis of which shape one
   should take.
3. `GetIpoptCurrentIterate` **aborted the process** when used as documented —
   found while wiring the plugin's `iteration_callback`:

   ```
   thread '' panicked at crates/pounce-algorithm/src/ipopt_cq.rs:502:
   RefCell already borrowed
   ...
   thread caused non-unwinding panic. aborting.
   ```

Building the interface then turned up a fourth, worse than any of them: with
default options **the NMPC feedback gain comes back silently zero**, in POUNCE
and in CasADi's own Ipopt plugin alike. Details under *Fix*.

## Cause

**The abort.** `GetIpoptCurrentIterate` held a shared borrow of the
algorithm-side `IpoptNlp` across `IpoptCq::curr_c` / `curr_d`, which take
`nlp.borrow_mut()` internally. Any call with a non-NULL `g` and `m > 0` hit it,
and because the panic crosses `extern "C"` it aborts instead of returning
`FALSE`. `GetIpoptCurrentViolations` had the same defect on its `grad_lag_x`
branch (via `curr_grad_lag_x` → `curr_grad_f`). This is not CasADi-specific: any
C consumer that logs its iterates is on this path, the GAMS C link included.

**The interface shape.** Four routes were measured before picking one:

| Route | Verdict |
|---|---|
| CasADi's `ampl` plugin + POUNCE as the solver binary | Rejected. `AmplInterface::init` asserts `is_a("SXFunction")` (no MX, so no `Opti`) and `p.is_empty()` (no parameters), and sets bound multipliers to NaN. |
| Python `casadi.Callback` wrapper | Rejected. `Opti` only accepts an `nlpsol` **plugin name**, and solution-map derivatives would have to be hand-rolled instead of inherited. |
| Out-of-tree `nlpsol` plugin | **Chosen.** |
| Drop-in `libipopt` | Rejected — not for missing C API (the whole plugin runs on it) but for a missing **C++ ABI**: CasADi's ipopt plugin imports constructors, a non-virtual member and three vtables from the `libipopt.so.3` bundled under `$ORIGIN` in the casadi wheel, so a drop-in means re-implementing Ipopt's object model in C++ *and* overwriting a file inside someone else's wheel — and it would cost side-by-side Ipopt, which is how this integration is validated. |

Full analysis, including the upstream in-tree option and the MadNLP precedent
(a Julia solver bound into CasADi through a C API — structurally our situation):
`dev-notes/casadi-interface-options.md`.

## Fix

**`casadi/`** (mirroring `gams/`) — a `casadi::Nlpsol` subclass wiring CasADi's
oracle functions into POUNCE through the existing Ipopt-compatible C API, with a
Makefile (`fetch-src` / `install` / `test` / `examples`), 50 parity checks
against CasADi's bundled Ipopt, and eight examples.

Measured against `ipopt` on the same model objects, casadi 3.7.2:

| | POUNCE | Ipopt |
|---|---|---|
| MX + parameters | `x = [0.907234, 0.822755]` | identical to 6 digits |
| `lam_x` at an active bound | `[-80.953, 0]` | identical |
| `lam_p` (parameter sensitivity) | `[-2.198878, -1.130540]` | identical, and pinned against a finite difference of `f*` |
| `jacobian(sol['x'], p)` | `[0.208074, 0.378275]` | identical |
| `convexify_strategy=eigen-reflect` | `f = -1.922746295` | identical — the same *different* local minimum |
| `Opti` | `Solve_Succeeded` | — |
| `iteration_callback` with live iterates | works | *"disfunctional in your installation"* on a stock build |

**Install** is either `make -C casadi install` (copies the plugin into CasADi's
own package directory — the first place its loader looks, so no environment
variable and no `sudo`), or a wheel: `casadi/wheel/build.sh` packages
`pounce-casadi`, and `import pounce_casadi` then registers the plugin
**in-process** by calling its `casadi_load_nlpsol_pounce` hook — the same entry
point CasADi's loader calls after finding a plugin on its search path. That
leaves CasADi's installation untouched and its bundled Ipopt loadable alongside.
Verified in a clean virtualenv holding nothing but `casadi==3.7.2` and the built
wheel. Nothing is published to PyPI; what a release adds is the manylinux /
macOS / Windows matrix, one entry per CasADi minor version
(`casadi/wheel/README.md`).

### Silently-zero sensitivities on bounded variables

An interior-point solve leaves a residual ~1e-12 multiplier on bounds it never
touched. CasADi's solution-map derivative reads *any* nonzero bound multiplier
as an active constraint and holds that variable fixed, so one such residual
turns the variable's whole sensitivity row into zeros. On a cart-pole NMPC with
bounded controls, `jacobian(u0, x0)` — the feedback gain — reads `0.0` where a
re-solve says `-9.109838`. Every bounded variable in the model is affected, and
CasADi's own Ipopt plugin does the same at its defaults; it ships
`clip_inactive_lam` for exactly this and leaves it **off**.

The plugin clips demonstrably inactive multipliers by default, testing *primal
distance to the bound* rather than multiplier magnitude — the same rule, option
name and margin as the ipopt plugin, with the default flipped.
`clip_inactive_lam=False` restores Ipopt-identical behaviour. Both directions
are pinned against a re-solve, which cannot be fooled the same way.

### Nonlinear-variable masks

`TNLPAdapter::quasi_newton_nonlinear_vars` ports upstream's
`GetQuasiNewtonApproxSpaces`: an arbitrary noncontiguous subset, mapped through
fixed-variable removal into the compressed space, in either index style, with
callback information taking precedence over `num_linear_variables` (which is now
implemented and off the unimplemented list). The L-BFGS updater projects
curvature pairs onto the subspace and publishes `B = σI + P(VVᵀ − UUᵀ)Pᵀ`. New C
API: `IpoptSetNonlinearVariables` / `IpoptClearNonlinearVariables`.

Two deliberate divergences, both because the obvious thing is worse:

*Count plus index list, not the `Bool` mask the issue proposed.* `Bool` was a
C99 `bool` in `pounce.h` (Ipopt 3.14 parity) but `c_int` in the Rust
implementation — an *array* of those is a data-layout bug, not a benign one.
That mismatch has since been fixed on its own in #627, which this branch is
rebased onto; the index-list API stands either way, being the shape Ipopt's own
`get_list_of_nonlinear_variables` uses.

*A curvature floor instead of upstream's exact zero.* Ipopt builds the space with
`reduced_diag = true`, so `W` is exactly zero on the linearly-entering variables.
Truthful, and a trap: those rows of the `(1,1)` block are then carried by `Σ_x`
alone — ~0 for a variable far from its bounds — and the factorization pays for a
near-singular diagonal on every one. Ported faithfully, 2 nonlinear + 2000 linear
variables went **0.85 s → 4.7 s**. CasADi's Ipopt on the same model goes
**0.40 s → 399 s**, which is what identified the flag rather than the port as the
problem. POUNCE keeps `limited_memory_init_val_min` (1e-8, registered with a
strict positive lower bound) there instead:

| linear variables | full space | masked, exact zero | masked, floor |
|---|---|---|---|
| 2 000 | 0.86 s, 25 it | 4.7 s, 28 it | 1.13 s, 28 it |
| 10 000 | 8.3 s, 31 it | — | **7.1 s, 27 it** |

Filling the whole diagonal with σ instead is equally fast and measurably worse:
it injects a proximal term at the problem's own curvature scale into coordinates
whose curvature is zero, and the masked update has no columns there to learn it
back down (the unmasked path does, which is why it never had this problem). That
variant stalls the 6-variable fixture at `Solved_To_Acceptable_Level` where
`tol=1e-9` was reached before — that experiment is why the floor is 1e-8 and not
σ.

### Carrying the working set between calls

`algorithm=active-set-sqp` already reached POUNCE's active-set driver through
the option dict (checked against the interior-point default). What could not
cross `nlpsol`'s fixed input signature is the **working set** — which bounds and
constraints were active — so `warm_start_from_previous` (default off) carries it
in the plugin's memory object from one call of a solver to the next.

Measured on a cart-pole MPC whose force limits saturate, 30-step loop, per solve:

| | mean | max | iters |
|---|---|---|---|
| active-set SQP, no working set | 233 ms | 375 ms | 2.86 |
| active-set SQP, `warm_start_from_previous` | **18 ms** | **23 ms** | 2.79 |
| interior point, warm started (reference) | 27 ms | 38 ms | 8.31 |

An order of magnitude, and it flips the ranking: without a carried set the SQP
is the worst option on that problem, with one it is the best. Iteration counts
barely move because the saving is inside each QP, not in fewer of them. Control
trajectory identical to 3e-11 — a working set is a starting guess, not a
constraint on the answer. On a loop whose bounds stay inactive there is little
to reuse (11.0 ms against 10.1 ms, near noise).

Off by default because it makes the function stateful: call *k+1* starts from
what call *k* found. A stale set is validated against the model and refused
rather than obeyed, and `stats()["warm_started_working_set"]` reports which
happened. Inert under the interior-point default, which produces no working set.

### An option surface that is a superset of the ipopt plugin's

Swapping `ipopt` for `pounce` in an existing script should not need the script
edited. Four options the ipopt plugin has that this one rejected or lacked, plus
one it read and ignored:

- **`grad_f` / `jac_g` / `hess_lag`** — user-supplied derivative functions
  (hand-derived, codegen'd, anything cheaper than differentiating the graph).
  Same signatures as the ipopt plugin, with the arity checked when the solver is
  built rather than misread at the first evaluation.
- **`convexify_strategy` / `convexify_margin` / `max_iter_eig`** — CasADi's own
  `Convexify`, literally the code its ipopt plugin calls, so a model tuned there
  behaves the same here. Pinned against that plugin on a nonconvex model where
  the strategies disagree with each other: both reach `f = -1.664033` plain and
  `-1.922746` under `eigen-reflect` — the same *different* local minimum.
  Exact-Hessian path only; a `limited-memory` solve that asks for it is warned,
  not silently ignored.
- **`save` / `load`** — serialization, as CasADi's own plugins support; the
  reloaded solver solves bit-identically. Configuration crosses, per-solve state
  does not, so a reloaded solver is cold and a carried working set stays in the
  memory object. Reading needs the plugin loadable in that process — the rule
  for every out-of-tree CasADi plugin — and fails cleanly if it is not.
- **`var_*_md` / `con_*_md`** — previously `Unknown option: var_string_md`,
  which is a porting papercut. Now accepted and echoed back through `stats()`.
  Nothing is forwarded (POUNCE has no metadata channel) and the docs no longer
  imply otherwise; the earlier claim that they were "accepted and dropped" was
  simply wrong, and is corrected.
- **`iteration_callback_step`** was read by CasADi and ignored here, so a
  callback throttled to every tenth iteration still ran on all of them. It now
  throttles — but only the callback. The ipopt plugin's early return also thins
  `stats()['iterations']`; throttling an expensive callback and losing the
  convergence history are unrelated wishes.

Callback time is now visible as `t_wall_callback_fun`, failed evaluations as
`stats()['n_eval_errors']`.

### Code generation of the whole solve

`solver.generate("mpc_step.c")` now emits the model **and** the solve: the
oracle functions, the option calls, and the loop that drives them. The file
compiles against `pounce.h` and links `libpounce_cinterface` — no CasADi, no
Python, no plugin at run time, which is the deployment story for firmware, a
ROS node, or a real-time target. Linked codegen, not freestanding C, so it does
not reach the smallest microcontrollers.

This one started as a wrong claim in this PR: the "not supported" list said
codegen was unavailable *"the same as for CasADi's Ipopt plugin"*. Ipopt's
plugin has generated all along — `IpoptInterface::codegen_body` emits C that
includes `<coin-or/IpStdCInterface.h>`. What settled it was not more reading but
taking the file CasADi generates for an ipopt solver, swapping that one include
for `pounce.h`, and compiling it: clean first try, because `pounce.h` **is**
that API. It failed only at run time, on the `linear_solver=mumps` CasADi bakes
in from Ipopt's option registry. So POUNCE's codegen types options off CasADi's
own `GenericType` instead — the way the interpreted path already did.

The contract is that the generated solve is **bit-identical** to the interpreted
one, `x` / `f` / `lam_x` / `lam_g`. That is stronger than it sounds:
`clip_inactive_lam` is a plugin-side correction, so the emitted runtime
(`casadi/pounce_runtime.hpp`, the counterpart of CasADi's `ipopt_runtime.hpp`)
has to reproduce it or the bound multipliers quietly differ between the two
paths. The L-BFGS nonlinear-variable subset crosses too, as a static index
array. `iteration_callback`, `warm_start_from_previous` and `convexify_strategy`
cannot cross and are refused **by name** at `generate()` rather than dropped.

Writing the test found a bug reachable only through generated code: the emitted
runtime carves `z_L` / `z_U` out of CasADi's `w` scratch, which the plugin had
never reserved — so generated code wrote past the caller's buffer. Heap
corruption, no error path, invisible from the interpreted plugin.
`alloc_w(2 * nx_, true)`.

## Blast radius

**Bit-identical.** `scripts/sweep-fixtures.sh` over all 57 CLI fixtures, run
twice: against this branch's original parent `0b09ca4`, and again after merging
`origin/main` (`cfc1121`, which brought #620's warm-start recentering into the
same `alg_builder.rs` / `application.rs` this branch edits). Empty diff both
times — same status, same objective, same iteration count on every model, no
`NO_JSON`. The branch has since been rebased onto `493cd3c` (#625's continuation
work and #627's `Bool` fix), neither of which touches the algorithm; the only
conflicts were `CHANGELOG.md` and the `num_linear_variables`
unimplemented-options entry this branch deletes.

That is the expected result: with no mask declared the updater builds the same
space it always did, and the mask is unreachable except through the new C API or
a TNLP that implements the callbacks. Exact-Hessian solves never consult it, and
the restoration sub-IPM clears it (its primal is the 5-block compound, where
those indices mean something else).

Everything else here is plugin-side and reaches no other caller: the new plugin
options live in `casadi/`, and the C-API borrow fix changes only failure
behaviour — a call that aborted the process now returns data.

## Tests

- `crates/pounce-cinterface/tests/current_iterate_inspectors.rs` — calls both
  inspectors from inside an intermediate callback with every buffer non-NULL,
  and checks the reported `g` against `g(x)` at the reported iterate. Confirmed
  on the parent commit: it does not fail, it takes the test process down with
  SIGABRT at `ipopt_cq.rs:502`, which is the defect.
- `crates/pounce-cinterface/tests/nonlinear_variables_mask.rs` — five tests:
  masked == unmasked KKT point, masked == exact-Hessian, declaring the subset
  leaves the exact-Hessian path bit-identical, declaring everything == declaring
  nothing, and argument validation refusing bad declarations without disturbing
  an accepted one. All five fail to compile on the parent commit — the API does
  not exist there.
- `crates/pounce-nlp` — nine adapter tests: fixed-variable remapping, Fortran
  index style, dedup/sort, the `num_linear_variables` prefix, callback
  precedence, and both error paths.
- `crates/pounce-algorithm` — five updater tests: default is no mask, a mask
  spanning everything is the identity, a mask that does not fit the space is
  ignored rather than misindexed, projection gathers correctly, and the history
  holds reduced-space pairs.
- `casadi/test_parity.py` (`make -C casadi test`) — **50 checks** against
  CasADi's bundled Ipopt: the NMPC feedback gain against a re-solve (both with
  clipping and with it disabled), `lam_p` against Ipopt and against a finite
  difference, a threaded `map(N, "thread")` fan-out matching serial solves
  bit-for-bit, a raising `Callback` failing the evaluation instead of aborting
  the process, `active-set-sqp` agreeing with the IPM, the working set carrying
  between calls without changing the trajectory, user-supplied
  `grad_f`/`jac_g`/`hess_lag` (including a mis-shaped one being refused), both
  convexify strategies against the ipopt plugin, a `save`/`load` round trip, the
  metadata options being accepted, `iteration_callback_step` throttling the
  callback while leaving the trace whole, and **code generation** — a generated
  file compiled with `cc`, loaded back through `ca.external`, and required to
  match the interpreted solve bit for bit, including the clipped multipliers and
  the L-BFGS mask.
- **CI runs it.** A `CasADi plugin parity` job in `.github/workflows/ci.yml`
  installs pip CasADi, fetches the matching source tree for headers, builds the
  plugin, runs all 50 checks and all eight examples, then builds the wheel and
  imports it in a clean virtualenv — on every PR.

`cargo test` is clean for `pounce-nlp`, `pounce-algorithm`, `pounce-cinterface`,
`pounce-restoration`, `pounce-presolve`, including #620's `warm_start_recentering`
suite and #627's `header_abi` tests on the rebased tree. `cargo test
--workspace` cannot complete in my environment for an unrelated reason:
`pounce-hsl`'s `t_sym_solver_ma57` fails to *link* without HSL's Fortran library
(`undefined symbol: ma57id_`), which stops the run before other crates execute.
`--exclude pounce-hsl` is clean; CI has HSL and covers it.

## Docs

- `docs/src/casadi.md`, linked from `SUMMARY.md` — install (both routes), the
  full option table, stats, callbacks and their throttle, warm starting
  including the working set, the bounded-variable sensitivity trap, the mask
  with its measurements, saving and reloading, code generation, what is not
  supported, and the AMPL fallback.
- `python/notebooks/35_casadi.ipynb` — executed with saved outputs like the rest
  of the series: first solve and convergence plot, `Opti`, warm-started MPC,
  differentiating through a solve plus a bilevel problem, the feedback-gain
  trap, the L-BFGS mask at two sizes, a performance section (POUNCE vs Ipopt,
  interior point vs active-set SQP, cold vs warm), and a cart-pole worked
  example — swing-up by direct collocation, then balancing against a control
  deadline at 25 and 50 Hz, then the working-set reuse. Indexed from the
  notebook README and the top-level README's integrations table.
- `casadi/README.md` — build, install, and the three ABI constraints that bite:
  the pip wheel ships no internal headers, it is built `_GLIBCXX_USE_CXX11_ABI=0`,
  and CasADi's loader performs **no** version handshake.
- `casadi/wheel/README.md`, `dev-notes/casadi-interface-options.md` — packaging
  and the decision record.

## Not in this PR

- A **published** `pounce-casadi` wheel. The wheel builds and works and is
  smoke-tested in CI; publishing needs the per-CasADi-minor build matrix. That
  is the decision left open.
- **Verification off Linux / off casadi 3.7.2.** Everything measured here is one
  platform and one CasADi minor version. The macOS and Windows paths are written
  but unexercised, and the plugin must be rebuilt per CasADi minor either way.
- Integer variables — POUNCE is a continuous local NLP solver, so `discrete` is
  refused by CasADi's base class rather than quietly relaxed. The useful role
  for POUNCE in a mixed-integer problem is as the NLP subsolver a branch-and-bound
  framework calls (which is exactly what Ipopt is inside Bonmin), reachable today
  through the `.nl` and GAMS routes.
- `convexify_strategy` inside generated code. It works interpreted; the emitted
  runtime would need `convexify_eval` wired in, and `generate()` refuses it by
  name meanwhile.
- A profile of why the exact-zero diagonal is so expensive in the KKT solve. The
  floor sidesteps it; the underlying question is still interesting.
- POUNCE's own parametric-sensitivity session (factor-once / solve-many) behind
  CasADi's derivative interface. The base class's generic KKT linearization
  costs only 8–12% on top of the solve at nx = 62…902, so the case for
  overriding it is weak once the clipping above is in place.


jkitchin:
## Reviewed and tested on macOS

CI runs the `casadi-plugin` job on `ubuntu-latest` only, so everything here was
unverified on macOS. Ran the whole thing on **macOS 26.5.2, arm64 (Apple
Silicon), Apple clang 21.0.0, CasADi 3.7.2 from the pip wheel, Python 3.12**,
in a throwaway venv so nothing leaked into a working environment.

### Everything passes

| | result |
|---|---|
| `make` (Darwin branch, `-Wall`) | builds clean, **zero warnings**, produces `libcasadi_nlpsol_pounce.dylib` |
| `make test` — parity suite | **50/50 PASS**, including codegen compiled with `cc` and matching the interpreted solve bit for bit |
| `make examples` | **8/8** |
| `casadi/wheel/build.sh` → install → solve | wheel builds, installs into a clean venv, solves Rosenbrock |
| `cargo test -p pounce-cinterface` | **63/63** (the new `nonlinear_variables_mask` and `current_iterate_inspectors` suites included) |

The Darwin branch of the Makefile, the `sys.platform == "darwin"` arm of
`pounce_casadi/__init__.py`, and the `.dylib` naming all work as written. The
codegen test builds a `.so` on macOS too — that is fine, dyld loads a `-shared`
bundle regardless of extension.

### One real macOS-only defect — fixed in 91f20d5b

Rust stamps a cdylib's `LC_ID_DYLIB` with its **absolute build path**. `otool -L`
on the freshly-built plugin:

```
@rpath/libcasadi.3.7.dylib
/abs/path/to/target/release/deps/libpounce_cinterface.dylib   <-- absolute
```

dyld resolves an absolute install name directly and **never consults the rpath
list**, so both existing `-rpath` entries were inert for it. The plugin loaded
only while that exact build tree survived at that exact path. Linux is unaffected:
the reference there is the soname, which the rpath already resolves.

Two consequences, both reproduced:

1. **`make install` / `CASADIPATH` breaks** once the build tree moves — and the
   error is actively misleading, because CasADi's loader swallows the dlopen
   failure:
   ```
   RuntimeError: plugin_interface.hpp:317: Plugin 'pounce' is not found.
   ```
   Nothing in that message mentions a missing dylib. (`DYLD_LIBRARY_PATH` does
   rescue it, if you already know that's the problem.)

2. **The wheel worked, but not for the reason the code gives.** `__init__.py`
   says the solver library is preloaded "rather than relying on an rpath, so the
   wheel is relocatable." On macOS what actually saved it is that `cp` preserves
   `LC_ID_DYLIB`, so the preloaded copy registers under the same absolute install
   name the plugin asks for and dyld matches it. Delete the co-located copy and
   the load fails naming the build path:
   ```
   OSError: dlopen(...libcasadi_nlpsol_pounce.dylib, 0x000A):
     Library not loaded: /abs/path/to/target/release/deps/libpounce_cinterface.dylib
   ```
   That is an accident, not relocatability, and one `install_name_tool` run from
   breaking.

**Fix:** rewrite the reference to `@rpath` after linking (reading the install
name back with `otool -D` rather than hard-coding it) and add `@loader_path` to
the rpath list so the wheel's side-by-side layout resolves. Entirely inside the
`ifeq ($(UNAME_S),Darwin)` branch — the Linux link line is byte-identical.

A/B on the identical relocated layout — plugin + solver library in a fresh
directory, build tree **removed entirely**:

| | result |
|---|---|
| before | `RuntimeError: Plugin 'pounce' is not found` |
| after | solves, `x* = [1. 1.]` |

`otool -L` now reads `@rpath/libpounce_cinterface.dylib`, and the parity suite is
still 50/50 after the change.

### Also fixed

`build.sh`'s `libpounce_cinterface.*` glob matched every cargo artefact sharing
the stem, so each wheel shipped the `.d` depfile and a **1.3 MB `.rlib`**. Now
copies the shared library only — staged tree went from 4 files to 2.

### Flagged, not fixed

**`build.sh` produces a `py3-none-any` wheel while staging a platform-specific
`.dylib` into it.** As written, the published artefact would install on Linux and
Windows and fail at import. It needs a platform tag — but that is a decision
across the whole (CasADi minor × platform) matrix, not something one macOS run
should settle unilaterally.

Two cosmetics, left alone: `casadi/README.md:3` and `:61` say the build produces
`libcasadi_nlpsol_pounce.so` and that `make install` "copies the .so" — the
Makefile handles the extension correctly, only the prose is Linux-flavoured.

### Worth considering

CI still proves none of this. A macOS leg on the `casadi-plugin` job would have
caught the install-name defect on the commit that introduced it; without one,
`casadi/wheel/README.md`'s claim of macOS x86_64 + arm64 support rests on
someone remembering to check by hand. The `otool -L` assertion is a one-liner —
the line must read `@rpath/libpounce_cinterface.dylib`, never an absolute path —
and I've noted it in the wheel README's macOS bullet for now.

jkitchin:
Thanks — merged your `91f20d5` and confirmed the Darwin branch leaves Linux untouched (rebuilt, parity 50/50).

Acted on two of the three items, in `d817f21`.

**macOS is in the CI matrix now.** You're right that the defect is less interesting than CI having been green on the commit that introduced it, so the job runs on `macos-latest` as well as `ubuntu-latest` (`fail-fast: false`), with two assertions aimed at this specific failure mode:

- `otool -L` on the built plugin must read `@rpath/libpounce_cinterface.dylib`. This is your one-liner. Worth stating why it needs to be an assertion and not a note: an absolute reference passes every other test on the runner that produced it, because the build tree is right there.
- the wheel must still solve **with the build tree hidden**. That is the half your review showed the `otool` check alone would not have caught — pre-fix, the wheel worked here too, because `cp` preserves `LC_ID_DYLIB` and the co-located copy answered to the absolute path. The step renames `target/release/libpounce_cinterface.dylib` aside, re-imports through the installed wheel, and restores it on exit. It's the A/B you ran by hand.

**The two README lines** are fixed — `.dylib` on macOS at `:3`, "copies the plugin" at `:61`. The Makefile always had it right; only the prose was Linux-flavoured. I also rewrote the macOS bullet in `casadi/wheel/README.md` to point at the CI assertion rather than asking someone to remember `otool` before shipping.

**The `py3-none-any` wheel tag I have left alone**, and I agree with your reasoning for not settling it in a macOS run. Recording where it stands so it isn't rediscovered: nothing that ships today is wrong, because nothing ships — the wheel is built and smoke-tested in CI and never uploaded. It becomes a real defect at the moment of the first `twine upload`, and it is a subset of the build-matrix decision already listed under *Not in this PR*. If it helps to have a starting proposal: tag each wheel `cp3x-none-<platform>` and put the CasADi minor in the version's local segment (`0.10.0+casadi3.7`), so pip resolves platform correctly and the CasADi pairing stays legible — but that is a release-process call, not a review comment, and it does not block this PR.

One consequence worth flagging: `macos-latest` is arm64, so the leg proves Apple Silicon only. `casadi/wheel/README.md` claims x86_64 as well, and that claim is still resting on nothing. Adding `macos-13` would cover it; I have not, since it doubles the leg for a platform the matrix decision above may drop anyway.

---

Issue #615, opened 2026-08-16
QP homotopy: CVXQP2_M takes 5.4x more iterations after the bound-adding ratio test (#614)


jkitchin:
## Summary

PR #614 makes the QP homotopy stop at every bound crossing on the path
(the new bound-adding ratio test in `crates/pounce-qp/src/homotopy.rs`,
9befbffa). That is a correctness fix — the path could previously walk out of
a stationary box — and it is worth its cost. But on one Maros-Mészáros
instance the cost is large enough that it should not sit unnamed in a commit
message.

**`CVXQP2_M`: 1768 → 9575 iterations (×5.42), 8.6 s → 63.6 s.**

Status and objective are **identical on both arms** —
`SolvedToAcceptableLevel`, objective `820155.431`. That is precisely why the
test suite cannot see this: the fixtures assert status and objective, not
iteration count. This is the gh#544 signature (206 → 812 on `pooling_rt2stp`,
shipped in 0.10.0, surfaced later as a CI wall-clock timeout, misattributed,
and came back as gh#592). Per `CLAUDE.md`, a measured regression accepted as
the cost of a fix needs an issue and an owner, or it is indistinguishable
from noise to the next reader.

Filing this so the number has a name against it. **Not a merge blocker for
#614** — the gate that PR exists to pass, it passes.

## Measurement

Paired sweep, `main` (5d8ad360) vs the #614 head, 138 Maros-Mészáros `.nl`
problems through the CLI with `solver_selection=qp-active-set` (which enters
`pounce_convex::active_set` and its `use_homotopy: true`,
`active_set.rs:518`). 120 s cap. Every problem ran on both binaries
back-to-back on the same worker, because the machine was shared with an
unrelated long job — arm-then-arm would have let load drift land on one arm.

Correctness is unmoved: **57/138 correct on both arms — the same 57 problems
— and zero solved-but-wrong on either arm.**

Aggregate effort over those 57: **103,810 → 113,623 iterations (+9.5%)**.
18 instances move, 13 worse and 5 better. **Excluding `CVXQP2_M`, the
aggregate is +2.0%** — this one instance is most of the total.

## Attribution — already done

Two ablation binaries off the #614 head, one change disabled in each:

| instance | `main` | #614 head | reconciliation OFF | bound ratio test OFF |
|---|---|---|---|---|
| `CVXQP2_M` | 1768 | 9575 | 9575 | **1768** |
[...]

Issue #608, opened 2026-08-15
Expose predictor-corrector NLP continuation across non-AD frontends


jkitchin:
## Summary

#90 delivered a numerical-continuation engine around the held KKT factor in the differentiable frontend. The same predictor-corrector capability is not exposed as a reusable repeated-NLP initialization path for the generic `Problem`, Pyomo, CLI, or GAMS frontends.

This is an integration issue, not a request to reimplement #90.

## Proposed scope

Add a frontend-neutral continuation driver that:

- accepts a parameter path `theta(s)` and a callback/model update;
- transfers the complete previous iterate;
- uses sensitivity/KKT solves for a tangent predictor when available;
- applies an explicit user transfer/prolongation map for horizon or mesh changes;
- performs a warm-started corrector;
- adapts step size from residuals and corrector work;
- detects active-set events and reanchors;
- optionally supports pseudo-arclength near folds;
- falls back to zero-order warm transfer when sensitivities are unavailable.

Expose thin adapters for generic Python `Problem` and Pyomo first; CLI/GAMS can follow using a path manifest or repeated-solve protocol.

## Acceptance criteria

- A parametric NLP sequence can be traced through the generic `Problem` API without rebuilding orchestration in user code.
- A Pyomo MPC/horizon-shift example supplies a transfer map and reuses primal/dual/barrier state.
- The driver reports predictor residual, corrections, step rejections, active-set events, and total evaluations.
- Benchmarks compare cold solve, previous-solution warm start, linear predictor, and predictor-corrector.

Related: #90 (completed engine), #602 (QP homotopy eligibility), #86.

PR #629, opened 2026-08-16
Strengthen Pyomo initialization: scaling, option propagation, and block-DAG recovery (#609)

### LINKED ISSUE #609: Strengthen Pyomo initialization: scaling, option propagation, and block-DAG recovery
## Summary

The Pyomo initialization stack has three remaining robustness gaps: an unscaled projection merit, inconsistent option propagation, and failure handling that abandons independent downstream work.

## Current behavior

- Nonlinear projection minimizes an unscaled sum of squared residuals, so large-magnitude rows dominate small but important constraints ([repair.py#L42-L105](https://github.com/jkitchin/pounce/blob/5d8ad360feb13e49bf6aefe66e78a14ddd054229/pyomo-pounce/pyomo_pounce/repair.py#L42-L105)).
- `initialize(..., options=...)` passes options to projection but not consistently to block initialization ([repair.py#L268-L295](https://github.com/jkitchin/pounce/blob/5d8ad360feb13e49bf6aefe66e78a14ddd054229/pyomo-pounce/pyomo_pounce/repair.py#L268-L295)).
- A failed block stops the block traversal rather than invalidating only its descendants; independent branches cannot continue ([block_init.py#L947-L986](https://github.com/jkitchin/pounce/blob/5d8ad360feb13e49bf6aefe66e78a14ddd054229/pyomo-pounce/pyomo_pounce/block_init.py#L947-L986)).
- Structural matching establishes incidence solvability but not numerical conditioning.

## Proposed scope

1. Add row/variable scaling to the projection merit, with user scaling suffix support and robust automatic defaults.
2. Define one typed initialization-options object and propagate it through repair, block solve, projection, and fallback paths.
3. Track the block graph explicitly; on failure, skip dependent descendants but continue independent branches.
4. Add numerical rank/conditioning checks after structural matching and route weak blocks to regularized least squares or a coupled fallback.
5. Return a structured report of initialized, skipped, failed, and fallback blocks.

## Acceptance criteria

- Initialization is invariant within tolerance under equivalent row rescaling.
- The same options reach every selected initialization stage.
- A failure in one branch does not prevent initialization of independent branches.
- Near-singular structurally square blocks produce a diagnostic/fallback rather than unstable values.
- Existing incidence-plan caching from #444 is preserved.

jkitchin:
Fixes #609

Baseline for every measurement below: **`cfc11218d75b1751d4bb2730be30bf35814a4016`** (tip of `main` when this branch was cut). No Rust changed.

> **Merged up before opening.** `main` has since advanced to `a44f4e8` (#608/#625, #615/#628, #627). The two conflicts this branch predicted — `pyomo_pounce/__init__.py` and `CHANGELOG.md`, both "keep both lines" — were the only ones, and are resolved in `d0c2509`. Re-verified on the merged tree: `test_issue_609_init_hardening.py` **29 passed**, full `pyomo-pounce/tests` **377 passed / 4 skipped**, and the new tests still fail on the parent. The measurements below were taken on `cfc11218` and are not re-derived; the behaviours they describe are pinned by tests that pass on the merged tree.

## Problem

The Pyomo initialization stack had four measurable gaps. Each was reproduced on the baseline before anything was written; the harness that produced these numbers is the same script run against both trees, with `PYTHONPATH` selecting which `pyomo_pounce` is imported and the same `pounce` binary (built from `cfc11218`) serving both.

### 1. Options stopped at the projection

`initialize(model, options={"max_iter": 137, "tol": 1e-9})`, with the solver's `solve` wrapped to record what each stage was handed:

| stage | baseline `cfc11218` | after |
|---|---|---|
| projection | `{max_iter: 137, tol: 1e-9}` | `{max_iter: 137, tol: 1e-9, nlp_scaling_method: user-scaling}` |
| block subsystem solve | `{}` | `{max_iter: 137, tol: 1e-9}` |
| stages dropping the option | `['block_subsystem']` | `[]` |

`block_initialize` had no `options` argument at all, so the stage that produces the starting point you actually get ran on solver defaults.

### 2. A failed block abandoned independent branches

Two branches sharing no variable: a 2x2 whose solution lies outside `p`'s bounds, and an unrelated `q -> r` chain.

| | baseline | after |
|---|---|---|
| `q` (exact 7.0) | **0.0** (seed) | 7.0 |
| `r` (exact 8.0) | **0.0** (seed) | 8.0 |
| `n_vars_initialized` | 0 | 2 |
| failed block's own seeds | restored | restored |

### 3. A near-singular square block returned wrong values, silently

`u + v == 2`, `u + (1+eps)v == 2 + eps` (Jacobian condition number ~`4/eps`), feeding `w == u + 2v`. Exact answer `u = v = 1`, `w = 3`.

| eps | baseline values | baseline error | after error | diagnostic |
|---|---|---|---|---|
| 1e-10 | `u=2, v=0, w=2` | **1.0** | 7.54e-09 | rcond 2.50e-11, routed to fallback |
| 1e-14 | `u=2, v=0, w=2` | **1.0** | 7.54e-09 | rcond 2.49e-15, routed to fallback |

The baseline reported `report.ok == True` and said nothing.

### 4. The projection merit was unscaled

**Rows.** A 1e6-unit energy balance beside a 1e-6-unit trace balance, both true at `a = b = 1`:

| | baseline | after |
|---|---|---|
| relative residual, energy row | 1.48e-16 | 0.0 |
| relative residual, trace row | **1.18e-08** | 0.0 |
| error vs exact | 7.87e-09 | **0.0** |

**Variables.** A pressure at 1e6 Pa and a mole fraction at 1e-4, coupled so each contributes 1.0 in its own units, with a 10% deficit:

| | baseline | after |
|---|---|---|
| relative move, pressure | 1.75e-12 | 0.100 |
| relative move, mole fraction | 0.200 | 0.100 |
| ratio | **1.14e+11** | **1.000** |

**Row-rescale invariance** (acceptance criterion 1) was *already met* on the baseline and is reported honestly as such — see Blast radius.

## Cause

1. `repair.initialize` called `block_initialize(model, solver=..., tee=..., repair="off", igraph=...)` — no `options`, because the parameter did not exist. There was also nowhere to put a knob that is not a solver option, so the scaling/conditioning/recovery policies did not exist.
2. The block loop `break`s on the first failed block. The comment said stopping was "the conservative choice" for blocks that "typically feed on the failed one" — but the loop had no representation of which blocks actually did, so it could not distinguish them.
3. The Dulmage–Mendelsohn partition establishes *structural* solvability. Nothing checked numerical rank, so a square block with a near-null direction went to `calculate_variable_from_constraint` / the subsystem solve and landed wherever that direction took it.
4. The merit was `sum((v - v0)**2)` — an absolute distance, so across mixed units a repair goes to whichever variable the constraint gradient favours regardless of physical meaning. And POUNCE's default gradient-based row scaling follows Ipopt's `min(1, g_max/||grad c||)`, which scales a large row **down** and leaves a small row alone; an absolute convergence test then enforces a 1e-6-unit row eight orders of magnitude more loosely than a 1e6-unit one.

## Fix

**`pyomo_pounce/init_options.py`** — `InitOptions`, frozen, validated, carrying `solver_options`, `tee`, `scaling`, `conditioning`, `cond_tol`, `fallback`, `regularization`, `on_block_failure`, `max_list`. One is built per `initialize()` call and threaded through the projection, every block solve, and every fallback solve. `InitOptions.coerce` keeps a bare mapping meaning solver options and **never** reinterprets it as policy — POUNCE has a solver option called `scaling`, so guessing would silently change a solve.

**`pyomo_pounce/init_scaling.py`** — two-sided row normalisation (`1/||grad c||_inf`) and per-anchor variable weights (`w = 1/|v0|`, with a floor so an anchor at zero stays free to move and a cap on the weight spread). Row factors are delivered through the model's own `scaling_factor` Suffix plus `nlp_scaling_method=user-scaling` — the route the NL writer already emits and `pyomo_pounce.scaling` already reads — so the solver learns nothing new, a user's own entry simply wins, and the Suffix is restored entry-by-entry afterwards (a model that declared none does not acquire one).

**`block_init.py`** — `BlockAnalysisReport.block_dependencies` is the block DAG, read off the incidence graph already in hand; a failure skips its transitive descendants via reverse reachability and other branches continue. Each block is rank-checked before solving, on a *scaled* Jacobian, and a weak one is routed to a regularized least-squares fallback. `BlockOutcome` records index/size/constraint/status/rcond/depends_on per block.

**Rejected along the way**, so nobody repeats them:

* *Rounding row factors to powers of two* to make the scaling exact in binary. Implemented, measured, removed: it did not move the residual 1.2e-9 row-rescale spread at all (1.2496e-9 → 1.2200e-9), which disproved the hypothesis that the spread came from factor rounding. It is a solver stopping-point effect — tightening `tol` to 1e-12 collapses the spread to 2.2e-16.
* *Column-normalising the Jacobian* for the conditioning check. It drives every 1x1 block to exactly rcond 1.0, retiring the check on most of a typical calculation order. Columns are scaled by their variable's magnitude instead, which keeps 1x1 blocks informative.
* *Ridging the whole merged group* in the coupled fallback. Measured: it picked `u = 2, v = 0` — precisely the arbitrary end of the near-null direction the fallback exists to avoid, because the downstream block's pull toward its own seed decided the weak block's answer. The ridge now covers only the weak block's own variables.
* *Raising the default `regularization`.* On the regularized path the error **is** the ridge bias and falls linearly with it (7.5e-9 at 1e-8, 7.5e-5 at 1e-4), so the default stays small. The coupled path is limited by the ridge gradient against the solver tolerance instead and wants a larger value (7.4e-3 at 1e-8, 7.6e-5 at 1e-6); that is documented rather than forced into one default.

## Blast radius

**Fixture sweep: empty diff, 57/57 fixtures bit-identical.** `scripts/sweep-fixtures.sh` run against a release binary built from the untouched parent `cfc11218` in a separate worktree, and against a release binary built from this branch; `diff` of the two outputs is empty. This is the expected result — the change is entirely Python-side under `pyomo-pounce/` and no Rust file, `Cargo.toml`, or `Cargo.lock` is touched (`git diff --stat -- '*.rs' Cargo.toml Cargo.lock crates/` is empty). The sweep was run anyway because a scaled projection merit changes the point later solves start from, which is exactly the trajectory class `CLAUDE.md` warns about; the corpus sweep drives the CLI on `.nl` fixtures and does not exercise `pyomo-pounce`, so it bounds the Rust side only. The Python-side trajectory change is bounded by the before/after tables above instead, which is where it is visible.

**Row-rescale invariance is preserved, and this is the one number that moved the wrong way.** Multiplying a row by a constant does not change the feasible set, so the projection must land in the same place:

| rescale factor k | baseline max deviation | after (default tol) | after (`tol=1e-12`) |
|---|---|---|---|
| 1e6 | 0.0 | 1.25e-09 | 5.6e-17 |
| 1e-6 | 0.0 | 0.0 | 0.0 |
| 1e10 | 0.0 | 1.25e-09 | 2.2e-16 |
| 1e-10 | 0.0 | 2.2e-16 | 2.2e-16 |
| nonlinear rows, worst over k ∈ {1e±6, 1e10} | 2.2e-16 | 2.2e-16 | — |

The baseline reached exact invariance; this branch reaches 1.25e-9 at the default tolerance. It is a *tolerance* effect, not a scaling defect, and the evidence is that tightening `tol` to 1e-12 collapses it to 2.2e-16 — with normalised rows the solver stops as soon as the scaled residual is inside `tol`, and where that happens depends on the iterate path. It also is not a quality regression: on the same model the baseline's "invariant" answer sits 2.8e-11 from the true projection while this branch's is exact to machine precision at k=1. The criterion says "invariant within tolerance", and 1.25e-9 is inside the default 1e-8. Recorded here rather than left for the next reader to rediscover.

**Behaviour changes for existing callers**, all opt-out:

* The projection now installs a temporary `scaling_factor` Suffix and turns on `nlp_scaling_method=user-scaling`, unless the caller set `nlp_scaling_method` themselves, the model carries a *non*-export `scaling_factor` Suffix (left strictly alone — flipping it to EXPORT would ship values the user deliberately kept local), or there is nothing to scale. `scaling="none"` restores the old merit exactly.
* The merit is weighted, so the projected point moves on any model whose anchors span magnitudes. On models whose anchors are all the same magnitude the weights are equal and the result is unchanged — the existing `test_project_repairs_inconsistent_fill` still gets 1/3 each.
* A failed block no longer stops the traversal. `on_block_failure="stop"` restores it. Dependent blocks are still skipped, so `test_failure_stops_before_downstream_blocks` is unaffected.
* Weak blocks are rerouted. `conditioning="off"` skips the check; `fallback="off"` diagnoses without rerouting.

**Conflict note.** `claude/pounce-issue-608-loop` (PR #625) conflicts with this branch in **two files**, both trivially:

* `pyomo-pounce/pyomo_pounce/__init__.py` — one hunk, adjacent import lines directly after the `block_init` import block. #608 adds `from pyomo_pounce.continuation import continuation, shift_map`; this branch adds `from pyomo_pounce.init_options import InitOptions`. Resolution is to keep both lines. The `__all__` additions (`continuation`/`shift_map` vs `BlockOutcome`/`InitOptions`) land in different parts of the list and merge cleanly.
* `CHANGELOG.md` — both add an entry at the top of `[Unreleased]`. Resolution is to keep both entries.

`repair.py` and `block_init.py` are untouched by #608, and `continuation.py` is untouched by this branch, so there is no semantic overlap — only the two textual hunks above. Verified with `git merge-tree --write-tree HEAD origin/claude/pounce-issue-608-loop`. **#625 has since merged and the resolution above was applied as described in `d0c2509`.**

## Tests

`pyomo-pounce/tests/test_issue_609_init_hardening.py`, 29 tests. Every number in their comments is a measurement on `cfc11218`.

**How they bite.** As written the module does not import on the parent at all — `InitOptions` does not exist — so collection fails there and that alone is not evidence. Re-run against the parent with the `InitOptions` import and the new-attribute assertions stripped, so only parent-era API is used, **nine fail behaviourally and for the right reason**:

| test | parent failure |
|---|---|
| `small_magnitude_row_is_enforced_like_a_large_one` | `0.9999999921 != 1.0 ± 1e-11` |
| `merit_shares_repair_by_relative_magnitude` | pressure moved `1.75e-12`, expected `0.2` |
| `user_scaling_suffix_entries_win_over_the_automatic_ones` | `0.9999999921 != 1.0 ± 1e-11` |
| `options_reach_every_stage` | `('block_subsystem', {})` — option dropped |
| `independent_branch_survives_a_failure` | `q = 0.0`, expected `7.0` |
| `near_singular_block_...[1e-10]`, `[1e-14]` | `u = 2.0`, expected `1.0` |
| `healthy_blocks_are_not_diagnosed` | `AttributeError: no attribute 'diagnostics'` |
| `block_initialize_takes_options_directly` | `TypeError: unexpected keyword argument 'options'` |

The last two are **API absence, not a behavioural bite**, and are listed as such rather than dressed up: there is no parent-era way to ask "was a diagnostic emitted" when the field does not exist.

**Two tests pass on the parent by design** — they are preservation guards, not regression tests:

* `test_projection_leaves_no_scaling_suffix_behind` holds the Suffix restoration.
* `test_still_one_incidence_walk_per_initialize` holds gh #444's single whole-model incidence walk, which is an acceptance criterion here.

Both bite when broken. Making `_block_dependencies` construct an `IncidenceGraphInterface` instead of reading the one already in hand fails **four** cache guards: the two in this file and `test_shared_graph_matches_fresh_analysis` / `test_shared_graph_zero_decision_falls_back_to_fresh` in `test_repair.py`. That was run and reverted, not assumed.

**Suite results** (Pyomo 6.10.1, so the `pyomo_pounce.v2` floor is met; the CLI binary was built and staged into `python/pounce/bin/` so the plugin resolves this checkout's build rather than falling back to PATH):

| suite | result |
|---|---|
| `pyomo-pounce/tests` | **372 passed, 0 skipped** (343 before this branch + 29 new) |
| `python/tests` | 815 passed, 36 skipped |
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --workspace --exclude pounce-hsl --all-targets` | exit 0 (pre-existing warnings only; no Rust touched) |
| `cargo test --workspace --exclude pounce-hsl` | exit 0, 226 suites green |
| `scripts/check-release-consistency.sh` | OK |
| `scripts/check-docs-consistency.sh` | OK — 44 pages, no dead links |

On the merged tree (`d0c2509`, after taking in #608/#615/#627) the same suites read **`pyomo-pounce/tests` 377 passed / 4 skipped** and `test_issue_609_init_hardening.py` **29 passed**; the extra tests and skips are #608's, not this branch's.

**Not pinned, deliberately.** The row-rescale invariance test asserts at `tol=1e-12` rather than the default, because at the default tolerance the quantity it measures is the solver's stopping behaviour and not the scaling. Pinning 1.25e-9 at the default would pin solver noise.

---

## Acceptance criteria

| # | criterion | verdict |
|---|---|---|
| 1 | Initialization is invariant within tolerance under equivalent row rescaling | **met** — 2.2e-16 at `tol=1e-12`; 1.25e-9 at the default, inside it. Already held on the baseline; kept, and now pinned by `test_projection_is_invariant_under_row_rescaling`. The deviation at default tolerance is recorded above rather than hidden. |
| 2 | The same options reach every selected initialization stage | **met** — stages dropping the option: `['block_subsystem']` → `[]` |
| 3 | A failure in one branch does not prevent initialization of independent branches | **met** — `n_vars_initialized` 0 → 2; `q, r` reach 7.0, 8.0 |
| 4 | Near-singular structurally square blocks produce a diagnostic/fallback rather than unstable values | **met** — error 1.0 → 7.5e-9, with `rcond` and a `diagnostics` entry |
| 5 | Existing incidence-plan caching from #444 is preserved | **met** — one walk per call, on the success *and* the failure path; four guards fail if broken |

## Scope items

| # | item | verdict |
|---|---|---|
| 1 | Row/variable scaling in the projection merit, user scaling suffix support, robust automatic defaults | **met** — `init_scaling.py`; suffix entries win; floor + spread cap as the robustness guards |
| 2 | One typed initialization-options object, propagated through repair, block solve, projection, and fallback | **met** — `InitOptions`, built once per call and threaded through all four |
| 3 | Track the block graph explicitly; on failure skip dependent descendants but continue independent branches | **met** — `BlockAnalysisReport.block_dependencies` + reverse reachability |
| 4 | Numerical rank/conditioning checks after structural matching; route weak blocks to regularized least squares or a coupled fallback | **met** — scaled-Jacobian rcond vs `cond_tol`; both `fallback="regularized"` and `fallback="coupled"` implemented and tested |
| 5 | Structured report of initialized, skipped, failed, and fallback blocks | **met** — `BlockOutcome` + `report.blocks` and the four bucket properties |

Nothing deferred; no follow-up issues filed.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,62 @@ changes.
 
 ## [Unreleased]
 
+- **Pyomo initialization: scaled projection, options that reach every
+  stage, and a failed block that no longer takes independent branches
+  down with it** (#609). Four measurements on the parent commit
+  `cfc11218`, each a defect the issue names:
+
+  *Options were dropped after the projection.* `initialize(...,
+  options={"max_iter": 137})` handed the dict to `project_to_feasible`
+  and to nothing else — `block_initialize` had no `options` argument at
+  all — so a tolerance tuned for the model reached the repair stage and
+  was silently dropped by every block solve that followed, which is the
+  stage that produces the starting point you actually get. There is now
+  one typed `InitOptions` object, built once per call and threaded
+  through the projection, every block solve, and every fallback solve. A
+  bare dict still means solver options and is never reinterpreted as
+  policy.
+
+  *A failed block abandoned the whole traversal.* On a model with two
+  independent branches, failing one left the other at its seeds: `q` and
+  `r` stayed at `0.0` instead of reaching `7.0` and `8.0`, and
+  `n_vars_initialized` was `0`. The block order is a DAG, and it is now
+  tracked as one (`block_analyze(...).block_dependencies`): a failure
+  skips its **descendants** and every independent branch still
+  initializes — `n_vars_initialized` `0` → `2` on that model.
+  `on_block_failure="stop"` restores the old behaviour.
+
+  *A structurally square, numerically near-singular block returned wrong
+  values and reported success.* On a 2x2 block with condition number
+  `4/eps`, initialization wrote `u = 2, v = 0, w = 2` where the exact
+  answer is `u = v = 1, w = 3` — an error of **1.0**, `report.ok` true,
+  nothing said. Blocks are now rank-checked on their *scaled* Jacobian
+  before being solved (so a mixed-units block is not mistaken for a weak
+  one), and a weak block is diagnosed and routed to a regularized
+  least-squares fallback that selects the minimum-norm solution: error
+  **1.0 → 7.5e-9**. `fallback="coupled"` merges the weak block with its
+  dependents instead; `conditioning="off"` skips the check.
+
+  *The projection merit was unscaled.* Rows: on a model pairing a
+  `1e6`-unit energy balance with a `1e-6`-unit trace balance, the trace
+  row's relative residual was `1.2e-8` against the energy row's
+  `1.5e-16`, because the solver's gradient-based rule only ever scales a
+  row *down*. Rows are now normalised two-sided through the model's own
+  `scaling_factor` Suffix — error **7.9e-9 → 0**. Variables: the merit
+  measured absolute distance, so a repair shared between a pressure at
+  `1e6` and a mole fraction at `1e-4` moved the mole fraction 20% and
+  the pressure `1.8e-12` relative, a ratio of **1.1e11**; weighting each
+  anchor by its own magnitude makes that ratio **1.000**. Your own
+  `scaling_factor` entries win over the automatic ones, and the Suffix
+  is restored exactly. `scaling="none"` restores the old merit.
+
+  `report.blocks` is the new per-block record (initialized / fallback /
+  failed / skipped, with `rcond` and `depends_on`). The single
+  whole-model incidence walk per `initialize()` call from #444 is
+  unchanged: the block DAG is read off the graph already built, and the
+  conditioning check differentiates per block rather than rebuilding it.
+  No Rust changed; the CLI fixture sweep is bit-identical.
+
 - **`WarmStart` artifacts carry the model they belong to, and warm-start
   options no longer outlive the call that asked for them** (#607). Two
   independent silences in `pounce.WarmStart`, both of the shape gh#544
--- a/docs/src/pyomo.md
+++ b/docs/src/pyomo.md
@@ -286,15 +286,18 @@ rep = pyomo_pounce.block_initialize(            # solve the equality
 `initialize_missing_values` fills each variable independently, so the
 fill can be internally inconsistent (mole fractions that do not sum to
 one); `project_to_feasible` repairs that by minimizing
-`sum((v - v0)**2)` subject to the model's active constraints and
-bounds — the full nonlinear projection, solved with POUNCE, with the
-original objective restored afterwards.
+`sum(w**2 * (v - v0)**2)` subject to the model's active constraints
+and bounds — the full nonlinear projection, solved with POUNCE, with
+the original objective restored afterwards. The weights `w` and the
+row scaling that goes with them are described under [Scaling the
+projection](#scaling-the-projection) below.
 
 Both stages guarantee that a failed solve leaves variable values
 exactly as they were: a diverged projection restores the
 pre-projection point, and a failed block solve restores that block's
-seeds and stops, so initialization can never make your starting point
-worse than it found it.
+seeds, so initialization can never make your starting point worse than
+it found it. A failed block does **not** stop the traversal — see
+[When a block fails](#when-a-block-fails).
 
 `block_initialize` is IDAES-flavored initialization without
 hand-written routines. `decisions=` holds the listed variables at
@@ -332,6 +335,116 @@ block-triangular calculation order. Use it to diagnose a large model's
 specification, or as the structural front end for tooling that decides
 *what* to specify before calling `initialize` /
 `block_initialize` to do the work.
+
+## Tuning initialization: `InitOptions`
+
+Every stage of the pipeline takes its solver options and its policy
+from one object:
+
+```python
+from pyomo_pounce import InitOptions
+
+rep = pyomo_pounce.initialize(
+    model,
+    decisions=[m.feed, m.reflux],
+    options=InitOptions(
+        solver_options={"tol": 1e-10},  # reaches EVERY solve below
+        scaling="auto",                 # projection merit + row scaling
+        cond_tol=1e-8,                  # what counts as a weak block
+        fallback="regularized",         # where a weak block is routed
+        on_block_failure="skip-dependents",
+    ),
+)
+```
+
+`solver_options` reaches the projection, every block subsystem solve,
+and any fallback solve. A bare dict still works and still means solver
+options — `options={"tol": 1e-8}` is unchanged — and is *never*
+reinterpreted as policy, because POUNCE has solver options whose names
+collide with these fields.
+
+### Scaling the projection
+
+The projection merit is `sum(w**2 * (v - v0)**2)`. With the default
+`scaling="auto"`:
+
+* **Variables** are weighted by their own magnitude (`w = 1/|v0|`), so
+  the merit measures *relative* movement. An unweighted merit measures
+  absolute distance, which across mixed units is not a distance anyone
+  wants: with a pressure at `1e6` beside a mole fraction at `1e-4`, it
+  dumps the whole repair on the mole fraction. An anchor at (or very
+  near) zero has no relative scale, so it keeps `w = 1` and stays free
+  to move, and the spread between the smallest and largest weight is
+  capped.
+* **Rows** are normalised two-sided (`1/||grad c||_inf`) through the
+  model's own `scaling_factor` Suffix. The solver's default
+  gradient-based rule only ever scales a row *down*, so a row in units
+  of `1e-6` keeps its magnitude and an absolute convergence test
+  enforces it far more loosely than a row in units of `1e6`.
+
+Entries the model already carries in its `scaling_factor` Suffix
+**win** over the automatic ones, and the Suffix is restored exactly
+afterwards (a model that declared none does not acquire one). See
+[User scaling with the `scaling_factor` Suffix](#user-scaling-with-the-scaling_factor-suffix)
+for the Suffix itself. `scaling="user"` uses only your entries;
+`scaling="none"` restores the old unweighted merit.
+
+Rescaling a row by a constant does not change the feasible set, and it
+does not move where the projection lands — to within the solver's own
+convergence tolerance, which is what sets the floor (measured: `1.2e-9`
+at the default tolerance, `2.2e-16` at `tol=1e-12`).
+
+### When a block fails
+
+`block_initialize` walks the equality blocks in calculation order, and
+that order is a DAG: `block_analyze(model).block_dependencies[k]` lists
+the blocks block `k` consumes values from. When a block fails, only its
+**descendants** are skipped; branches that do not depend on it are
+still initialized:
+
+```python
+rep = pyomo_pounce.block_initialize(model)
+rep.initialized_blocks   # solved as square systems, values kept
+rep.fallback_blocks      # weak blocks routed to a fallback
+rep.failed_blocks        # solve failed; seed values restored
+rep.skipped_blocks       # descendants of a failure (with the reason)
+```
+
+Each entry is a `BlockOutcome` carrying the block's index, size,
+leading constraint name, status, `rcond`, and `depends_on`. Pass
+`on_block_failure="stop"` to go back to abandoning the traversal at the
+first failure.
+
+### Weak blocks: structurally square is not numerically solvable
+
+Structural matching proves a block is solvable *in principle*. Whether
+its Jacobian has usable rank at the current point is a separate,
+numerical question, and a square-but-near-singular block handed to a
+Newton step lands wherever the near-null direction takes it, silently.
+
+Each block is therefore rank-checked before it is solved. The check is
+run on the **scaled** Jacobian — rows divided by their own gradient
+norm, columns multiplied by their variable's magnitude — so it answers
+a numerical question and not a units one; a mixed-units block is not
+weak. A block whose reciprocal condition number falls below `cond_tol`
+is recorded in `rep.diagnostics` and routed to `fallback`:
+
+* `"regularized"` (default) minimises the block's scaled squared
+  residuals plus `regularization` times the scaled squared step from
+  the seed, which selects the minimum-norm solution instead of an
+  arbitrary point on the near-null direction. The residual it leaves is
+  the ridge bias, so it falls linearly with `regularization` — which is
+  why the default is small (`1e-8`).
+* `"coupled"` merges the weak block with the blocks that depend
+  directly on it and regularizes the union, for a deficiency that only
+  downstream rows resolve. Its precision on the near-null direction is
+  set by `regularization` against the solver tolerance rather than by
+  the ridge bias, so a *larger* `regularization` (around `1e-6`) is the
+  better choice there.
+* `"off"` reports the diagnostic and solves the block squarely anyway.
+
+`conditioning="off"` skips the check entirely.
+
 
 ## Repairing a bad specification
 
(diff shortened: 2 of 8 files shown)


jkitchin:
## Review — read the code, reproduced the claims independently on macOS, found and fixed one regression

Reviewed at `d0c2509`, tested on macOS (arm64, Python 3.12, Pyomo 6.10.0), CLI built from `a44f4e8b`. CI covers `ubuntu-latest` only, so this is the second platform.

### The headline claims reproduce, independently

Not by re-running the PR's own tests — by writing parent-era-API scripts and running them against `main` and against this branch with `PYTHONPATH` selecting which `pyomo_pounce` is imported.

**Independent branch survives a failure** (criterion 3), two branches sharing no variable, one failing on its bounds:

| | main | this branch |
|---|---|---|
| `q`, `r` | `0.0`, `0.0` | **7.0, 8.0** |
| `n_vars_initialized` | 0 | **2** |

**Near-singular square block** (criterion 4), exact answer `u = v = 1`, `w = 3`:

| eps | main | this branch | diagnostic |
|---|---|---|---|
| 1e-10 | `u=2, v=0, w=2`, err **1.0**, `ok=True` | err **7.54e-09** | rcond 2.50e-11, routed to the regularized fallback |
| 1e-14 | `u=2, v=0, w=2`, err **1.0**, `ok=True` | err **7.54e-09** | rcond 2.51e-15, routed to the regularized fallback |

Both match the body's numbers. The `ok=True` on the parent is the part worth keeping in view: it was wrong by 1.0 and said nothing.

**Cost of the always-on conditioning check.** `conditioning="auto"` is the default, so every caller pays it, and the body does not measure it — so I did. A 400-block 1x1 calculation chain: main 0.177 s, this branch 0.144 s. No measurable overhead (the difference is noise). Concern retired.

**Suite on macOS**: `pyomo-pounce/tests` **346 passed, 8 skipped**; `test_issue_609_init_hardening.py` **29 passed**. The 2 failures and 1 collection error here are pre-existing on `main` in this environment (Pyomo 6.10.0 against `pyomo_pounce.v2`'s 6.10.1 floor; no `pounce_v2` on PATH) — verified by running the same three against `main`'s tree.

### One regression, found and fixed in `ea15387`

`_install_projection_scaling` creates a `scaling_factor` Suffix when the model has none. A model that already uses that name for a **Param, Var or Block** makes `add_component` raise — and the call sat after the original objectives were deactivated and the projection objective attached, but *before* the `try` whose `finally` undoes both. The exception escaped and left the model permanently broken:

```
# model carrying  m.scaling_factor = Param(initialize=1.0)
main:        cond = optimal      obj active? True    projection obj left? False
this branch: RuntimeError        obj active? False   projection obj left? True
```

That is a hard failure plus a mutation leak on a model that worked before #609, and it contradicts the docstring this PR itself added ("objectives/constraints/**Suffixes** are restored exactly").

Fixed narrowly:

* the collision **degrades to an unscaled projection** rather than raising — the name is the user's, and running unscaled is exactly the pre-#609 behaviour, the same way a non-export `scaling_factor` Suffix is already left strictly alone;
* `_install_projection_scaling` moved **inside** the `try`, so a failure while installing scaling is undone by the same `finally` as a failed solve;
* the helper's two unused parameters (`anchored`, `variables`) dropped.

`test_a_taken_scaling_factor_name_degrades_instead_of_raising` pins it, and **fails on the unpatched branch** with the RuntimeError above — checked by reverting `repair.py` alone and re-running. Suite after the fix: **347 passed, 8 skipped**.

### On the rest

The design decisions are argued in the code rather than asserted, and the ones that were measured and *rejected* are written down (power-of-two row factors, column-normalising the conditioning Jacobian, ridging the whole merged group, a larger default `regularization`) — that is the part that saves the next reader from repeating them. The row-rescale invariance number that moved the wrong way (0 → 1.25e-9 at default tolerance) is disclosed rather than buried, with the evidence that it is a stopping-point effect (2.2e-16 at `tol=1e-12`), and the test deliberately pins the tightened tolerance instead of pinning solver noise. `InitOptions.coerce` never reading a mapping as policy is the right call given POUNCE has a solver option named `scaling`.

No Rust changed, so the empty fixture sweep is expected rather than evidence; the Python-side trajectory change is bounded by the before/after tables, which is where it is visible.

Merging.

PR #631, opened 2026-08-16
sens: estimate(mode="path") and the record of active-set changes (roadmap item 2)

devin-griff:
## What this adds

`estimate()` gains a third mode. `mode="path"` applies the perturbation a little at a time: it takes the largest fraction of the change the current active set allows, applies the one change that happens at that fraction, and continues under the updated set. The prediction is piecewise linear in the parameter. For a QP that is the exact solution path, since a QP's solution is piecewise affine in the parameter. For an NLP the one error left is the linearization at the base point, because nothing is re-linearized between breakpoints.

Three kinds of breakpoint end a segment:

1. A free variable reaches a bound and is held there through a Schur row, without re-factoring.
2. A bound active at the base point has its multiplier fall to zero and the variable leaves it. The release re-factors without that bound's sigma, on #600's machinery.
3. A hold the path itself added stops binding under a later direction and the variable leaves that bound again. The hold's Schur-row multiplier starts at zero at the crossing, takes a sign on the following segment, and the hold drops where it crosses zero again. This is the constraint add and drop the roadmap item names, and it is what no decision at the base point can represent.

`active_set_changes()` returns the record, which is the part no other mode produces: one entry per change with the fraction of the perturbation, the variable, which bound, and whether the variable reaches it or leaves it. The first entry's fraction is how much measurement discrepancy the held solve's active set survives unchanged. `max_iter` is the same knob it is under `fix_relax`. It caps the changes applied, and past the cap the rest of the perturbation is taken in one step under the active set reached, with a warning.

## Four defects found while building it

Each was isolated on a QP small enough to check by hand, and each fix carries its measurement in the commit.

1. **The barrier correction entered released equations at order one.** The parametric right-hand side puts minus mu on every bound-multiplier row. Eliminating a released row folds its right-hand side into the variable's equation divided by the slack, and at an active bound mu over the slack is the multiplier itself. The released direction came back [1.154, 0.194] against the analytic [1.227, 0.454], digit for digit the injected system's answer. `solve_released` now zeroes released rows before the fold. This shipped with #600 but nothing could hit it: `fix_relax`'s shifted call zeroes the row as part of moving the base multiplier, and until this PR there was no unshifted caller. `released_sigma_x` also rebuilds a released diagonal entry from the bounds that stay active instead of differencing two numbers of order z/s.
2. **The repeat guard barred a whole segment.** A row that changed at one fraction was barred from changing back for the entire next segment, so a released variable's next bound crossing went unrecorded. The guard now clears as soon as the path advances.
3. **A bound the factorization still enforces could be held again.** The reach scan now skips factor-held bounds, and a release bars both the multiplier row and the variable row at that one fraction, so a just-released variable is not re-held where it was freed.
4. **Base-bound activity drifted mid-path.** The release test compared the accumulated multiplier against the current slack, so a near-bound inactive multiplier could be reclassified active while its variable sat on that bound under a path hold, which put a departure in the record for a variable that was never on the bound. Activity of a multiplier row is now the complementarity split at the base point, where the factorization's sigma was frozen. The test that caught this is the added-then-dropped QP in `test_path.py`.

## Measurements

- A 400-case random QP scan ends every path at the direct target solve to 1e-10, leaves no variable held that the target solve frees, and produces records with a bound added and later dropped where re-solves require them.
- On the notebook's Hicks CSTR (100 one-minute intervals, 13 with the coolant valve saturated): where the two modes settle the same active set they agree to 1e-6, both at 0.199 worst relative miss. At the full change to the steady state, `fix_relax` decides all 13 releases at once from base-point multipliers and lands below even `mode="linear"` (0.950 against 0.833) while `mode="path"` lands at 0.626, the best of the three.
- Cost grows about linearly with changes applied, about 50 ms at a cap of 16 against re-solves of 0.8 to 0.9 s.
- The record on the CSTR is 13 departures, from the last saturated interval backward, the first at 1.3% of the change. The notebook's interval table is now read off the record's components and matches the ordering previously inferred from consecutive KKT rows.

## Also in this PR

- The refinement cap is named `max_iter` and the comments and locals say clamp throughout, per the item plan of doing both in one PR.
- The record names variables: `PathSegment` carries `var_row` and `lower` for every kind of change, and the binding returns `(fraction, var_row, lower, pinned)`.
- The fallback warning under `mode="path"` no longer claims `fix_relax` pinned something.
- A `mode="path"` section in `docs/src/sensitivity.md`, the CHANGELOG entry, and notebook `35_active_set_parametric_sensitivity.ipynb`, the worked example behind the docs numbers, with `max_iter` sweeps of both modes against re-solve wall time.

## Testing

- `released_direction.rs`: the held and released directions against hand algebra, the released direction after a prior plain solve (the exact call sequence the path makes), and the path through a release and a hold to the exact endpoint.
- `test_path.py`, 11 tests: the crossing against a re-solve, releases on both sides, agreement with `fix_relax` where they settle the same set, the first record fraction against `estimate_report().alpha`, the cap falling back to the clamp with a warning, and the QP whose record holds a bound the path added and later dropped.
- Full suites green: 12 rust sensitivity test files, 350 pyomo-pounce tests, and the 842-test python suite with only its 3 preexisting environmental failures (numpy 1.26 lacking `trapezoid`, a cp1252 decode), unrelated to this PR.


jkitchin:
Reviewed, verified locally, and merged `main` into the branch to clear the conflict. Two small fixes pushed on top (`a8710f05`); the numerics are yours and untouched.

## What I checked, and how

**The three defects the PR says it found are real, and the fixes are right.**

The `solve_released` one is the one that matters. A released bound's multiplier is fixed at zero, so its row carries no equation and the right-hand side there is meaningless — but it was being folded into the variable's equation as if it did. By complementarity that injection is `-mu / s = -z`, which is order one, not order `mu`: it was corrupting the direction, not perturbing it. Zeroing the released rows before the fold is the correct fix, and the reason it shipped with #600 unexercised is exactly as the CHANGELOG says — `fix_relax`'s shifted call zeroes the row on the way in, so nothing reached the bad path until the walk did.

`released_sigma_x` I checked against the definition rather than the diff. `curr_sigma_x` is `P_l S_l^-1 Z_l + P_u S_u^-1 Z_u` (`ipopt_cq.rs:729`) — purely the two bound terms, nothing else. So rebuilding the entry from the variable's surviving non-released bound rows is not an approximation of the subtraction, it is exactly the same number, computed without differencing two quantities of order `z / s`. Strictly better conditioned.

The activity-at-the-base fix is the subtle one and the comment earns its length. Deciding activity from accumulated values let a near-bound inactive multiplier drift past its shrinking slack mid-path and report a departure from a bound the variable was never on — a wrong entry in the record, which is worse than a missing one, because the record is the thing a user reads.

**Tests.** 112 in `pounce-sensitivity`, including all 16 in the three new files:

    parametric_cpp.rs      10 passed
    path_following.rs       2 passed
    released_direction.rs   4 passed

and 392 passed / 6 skipped across `pyomo-pounce/tests` against a from-source build of the extension, `test_path.py`'s 11 included. `cargo fmt` clean; `pounce-sensitivity` and `pounce-py` add no clippy warnings.

`test_a_variable_reached_partway_can_leave_again` is the test that justifies the whole mode. A variable reaching a bound partway through the perturbation and leaving it again is not representable by any decision taken at the base point, so it is precisely what `fix_relax` cannot do, and the endpoint matching a re-solve to `abs=1e-8` is the right way to assert it.

**`max_passes` → `max_iter` is not a breaking change.** I checked before assuming: `## [0.10.0]` starts at CHANGELOG line 1453, and every entry mentioning `max_passes` sits above it, in `[Unreleased]`. The name was never released, so renaming it costs no one, and editing those entries to say `max_iter` is copy-editing unreleased text rather than rewriting history. Presolve's unrelated `max_passes` is correctly left alone.

**No fixture sweep.** Per CLAUDE.md I considered whether this is a trajectory change and concluded it is not: `solve_released`, `released_sigma_x` and `step_along_path` all run after convergence, on the frozen factorization. `pounce-cli` depends on `pounce-sensitivity`, but only for the post-solve `--sens` path — nothing here can reorder or rescale a barrier iteration.

## The two fixes I pushed

Neither changes behaviour.

1. The augmented-solve failure message in `path_direction` carried a 14-space run left over from a reflow — user-visible, so worth the one-line fix.

2. `factor_holds` recomputed `slack_base` inside the `mult_nat` search, once per candidate multiplier row, though it depends only on the variable and the side. Hoisted out. The `is_finite` guard was ANDed inside the `any`, so a non-finite slack already made the search false for every row; the early return now says that directly.

## One thing left, not a blocker

The base-point activity test — *the multiplier dominates the slack* — now appears twice: in `factor_holds` and in the release scan below it. They are the same predicate, and only the second one carries the paragraph explaining why it is evaluated at the base. I added a comment on the first pointing at the second, but a shared helper would be better, since a future fix to one is a fix the other silently misses. I left that to you rather than refactoring the subtle part of someone else's PR at merge time.

Separately, worth knowing rather than acting on: `factor_holds` is `O(n_x x n_bound_rows x n_mult)` per path iteration. Fine at the sizes the tests and the notebook exercise, and the path is a post-solve analysis rather than an inner loop, but it is the first thing that will bite on a large active set.

Merging.

Issue #483, opened 2026-08-05
User scaling from Pyomo is a silent no-op, and the core silently discards x_scaling


devin-griff:
**What happens.** A Pyomo user who tags the standard `scaling_factor` Suffix and sets `nlp_scaling_method=user-scaling` — the workflow that works with Ipopt through ASL — gets no scaling at all, and no message saying so. pyomo_pounce contains no scaling code: nothing populates the bridge's `user_scaling` state (`tnlp_bridge.rs:202`), so the solver's `UserScaling` arm (`orig_ipopt_nlp.rs:868`) asks, is declined, and falls back to no scaling. The option is accepted and means "none."

One layer down, the raw API has the same trap for variable factors specifically: `Problem.set_problem_scaling(obj, x_scaling=, g_scaling=)` accepts all three, but `scale_user_supplied` ends with

```rust
// `use_x_scaling`: silently ignored (not modeled — see doc).
let _ = use_x_scaling;
```

so even a raw pounce-py caller loses variable scaling silently. Objective and constraint factors are fully modeled (combined with `obj_scaling_factor`, split into the c/d vectors, d-bounds brought into scaled space); variable factors are plumbed end to end and used nowhere.

**What this is and is not.** Solutions come back correct in model units, and the sensitivity accessors are unaffected today (their natural-units translation covers every scaling the core can currently produce). What a user loses is the conditioning they explicitly asked for, invisibly: a badly scaled model converges poorly or fails with no hint the suffix was ignored.

**Proposed fix**, in three stages so nothing is silent at any intermediate point:

1. **Objective and constraint scaling end to end.** pyomo_pounce reads the EXPORT Suffix (ConstraintData entries in row order, the active Objective's entry; entries on fixed variables or inactive constraints skipped; unlisted components at 1.0; no suffix leaves `user_scaling` unset) and calls the existing `set_problem_scaling` on the in-process path; the CLI wires the already-parsed `.nl` suffix vectors through for subprocess solves. Variable entries in the Suffix, and `x_scaling` in the raw API, raise with a clear message instead of being discarded. Sensitivity needs no code here — `natural_units_conj` already translates `df`/`dc`/`dd` from any method — and a test axis proves rather than assumes it: every accessor (`covariance`, `information`, `gradient`, `estimate`, the classifier's statuses, `wrt=` blocks, retain-only) against unscaled ground truth with user scaling engaged, plus the fixed-variable composition.

[...]

PR #632, opened 2026-08-16
init: measure the 1x1 convergence test on the row's stated scale

devin-griff:
## What happens

A cold start on the IDAES saponification CSTR in raw SI units (energy holdups near 3e7 J) runs `pyomo_pounce.initialize`, which reaches `block_initialize`. The 1x1 block solving the energy holdup equation for a temperature fails with Pyomo's "Linesearch iteration limit reached" at a remaining residual of 7.45e-8. The block's seeds are restored, 827 of the 1001 variables keep their seeds, every reaction rate starts the NLP at zero, and the downstream dynamic optimization takes 137 to 225 iterations where a fully initialized start takes 19.

The model carries a `scaling_factor` Suffix stating that equation's factor as 1e-7, and the 1x1 path is the one place in the package that does not read it: the gh #483 reader collects the Suffix for full-model solves, and `repair.py` delivers it to the projection under `user-scaling`.

## Why

The 1x1 path calls `calculate_variable_from_constraint` with its defaults, and that function's convergence test is `eps=1e-8`, absolute. An equation whose terms sit near 3e7 has a double-precision evaluation floor near `3e7 * 2.2e-16` times a few operations, which lands at 1e-8 to 1e-7. The residual 7.45e-8 is converged by any measure relative to the equation's own magnitude, and the linesearch can never reduce it further. Historically callers worked around this by initializing a `core.scale_model` clone, which brought every equation to order one. A caller that has removed those clones has the Suffix as its only channel.

## The change

`block_initialize` reads the model's constraint factors once through `read_scaling`, the same gh #483 reader every full-model solve uses, and the 1x1 path passes `eps = 1e-8 / f` for a tagged constraint. That is the identical test measured on the scaled row `f*g(x)`: for the energy equation at factor 1e-7 it accepts residuals below 1e-1 in raw units, which is 1e-8 on the row's own scale, and 7.45e-8 passes by six orders. Variable factors change nothing on this path, since one-dimensional Newton is invariant to rescaling the variable, so the whole effect is the residual test.

An untagged constraint makes the exact call the current release makes. A model without an export-enabled Suffix never reaches the lookup, matching the #483 policy that a `direction=LOCAL` Suffix reaches no solver and so reaches this test no differently.

## Tests

The test row is `1e10 * x**2 == 2e10`, chosen because its residual floor is deterministic in IEEE doubles: the root is the square root of 2, no double near it evaluates `x**2` to exactly 2, so the raw residual bottoms out near 4.4e-6, unreachable by the absolute test and six orders inside the scaled one. Six tests in `test_block_init_row_scale.py`:

- the tagged row initializes to the root,
- the same row untagged still fails with its seed restored, which also proves the tagged case isn't vacuous,
- a `direction=LOCAL` Suffix is not read,
- an untagged row keeps the default next to a tagged one,
- a factor on the indexed container covers each member row,
- a factor of 1.0 gives the identical outcome to the untagged path.

Full pyomo-pounce suite: 384 passed, main's 378 plus these 6.

## Verified downstream

On the CSTR cold start above, with the factors written to the Suffix before the run, the report goes from FAILED with 827 variables at their seeds to all 1001 initialized, matching what the scaled-clone route used to produce, and the first dynamic optimization returns to the iteration count of the fully initialized start. A solvent extraction cold start over 16,217 variables, equations spanning 1e-7 to 1e7, has identical block outcomes with and without the change.


jkitchin:
Reviewed, verified against the code rather than the description, and pushed two additions plus the `main` merge (CHANGELOG conflicted with the casadi wheel entry; both kept).

**The diagnosis holds.** `calculate_variable_from_constraint`'s test really is `while abs(fk) > eps`, absolute, and the linesearch's `c1 = 0.999` sufficient-decrease with `alpha_min = 1e-8` is what raises once the residual is at its evaluation floor — so an unreachable `eps` surfaces as *"Linesearch iteration limit reached"*, exactly as reported. `read_scaling` returns `(obj, {ConstraintData: f}, vars)`, so `scaled[1]` keyed by `cblk[0]` is right, and the LOCAL-Suffix and container-expansion tests are testing real behaviour of `_suffixes` / `_iter_data`, not incidental behaviour. Reproduced the six tests locally: 6 passed.

**What was missing: the other direction, which is the silent one.**

`eps = 1e-8 / f` moves both ways, and the PR measures, tests, and documents only `f < 1`. The `f > 1` half is not cosmetic:

```
1e-8 * x**2 == 2e-8        # same root, row in small units
```

The raw residual at the seed `x = 1` is 1e-8, so the absolute test accepts it **before taking a step**. Measured on `main`:

| | status | x | rel. error |
|---|---|---|---|
| untagged (today) | `initialized` | 1.00000001 | **29%** |
| tagged `1e8` (this PR) | `initialized` | 1.4142135623746899 | 1.1e-12 |

Nothing fails and nothing warns — the block is *reported initialized* and the value handed downstream is the seed. That is the same defect as the IDAES failure with the sign flipped, and it is the worse of the two, because the loud version at least stops. Added `test_an_untagged_cold_row_is_declared_solved_at_its_seed` and `test_a_tagged_cold_row_is_solved_to_the_root` to pin both halves, and the CHANGELOG now says so. This is an argument *for* the change that the PR was leaving on the table.

**The same half has a regression surface, and now has an opt-out.** A row genuinely of order one, tagged `1e8`, gets `eps = 1e-16` and exhausts the linesearch — measured, `x**2 == 2` goes from `initialized` to `failed` at factors ≥ 1e8. That is correct under the PR's own principle (the factor is the user's statement about the row) but it is a behaviour change with no way out. `scaling="none"` is already documented as *"no row scaling"*, so it now reaches this test too. "None means no row scaling except here" was the reading the code had; it isn't one worth shipping.

**One thing I did not change, worth a thought.** `pyomo_pounce.init_scaling.row_factors` computes `1 / ||grad c||_inf` per row with no Suffix at all, and `block_init.py` already calls it for the regularized fallback and the rank check. On the very row `test_the_same_row_untagged_still_fails` pins as failing, it returns **5e-11** — within a factor of 2 of the `1e-10` the test hand-writes:

```
auto row_factor on the untagged hot row: 5e-11
  default 1e-8  -> IterationLimitError: Linesearch iteration limit reached
  auto-scaled   -> x = 1.4142135623746899
```

So the untagged failure is a policy choice, not a limitation, and `InitOptions.scaling="auto"` — the default, documented as *"automatic row factors ... the model's own Suffix entries win"* — is the policy that would already cover it. The conservative reading is that a gradient at a bad seed is a worse authority than the user's Suffix, and I think that reading is defensible enough to ship this as-is. But it means the framing *"a caller that has removed those clones has the Suffix as its only channel"* is not quite true, and the IDAES model would not have needed a hand-written Suffix under `scaling="auto"` if the 1x1 path read the automatic factors. Worth deciding deliberately rather than by omission.

Local run: 9 passed in the new file; 356 passed in `pyomo-pounce/tests` (two `every solver route` failures are the known worktree-has-no-bundled-CLI environment issue, green in CI; one collection error is my local Pyomo being older than 6.10.1).

Merging once CI is green.

Issue #605, opened 2026-08-15
Make feasibility projection sparse and safeguard the linearized normal step


jkitchin:
## Summary

`pounce.project_to_feasible` currently materializes dense Jacobian/projection matrices and accepts the linearized least-squares point without checking whether it improves the true nonlinear constraints. That makes the helper unsuitable as a large-scale NLP initializer and fragile when the linearization is poor.

## Current behavior

The standalone projection allocates dense `J` and `P = I` and builds dense constraint rows in [_starts.py#L183-L274](https://github.com/jkitchin/pounce/blob/5d8ad360feb13e49bf6aefe66e78a14ddd054229/python/pounce/_starts.py#L183-L274).

The core `least_square_init_primal` path similarly computes a least-squares correction and copies it into the iterate without a trust region, fraction-to-boundary control, or actual-versus-predicted reduction test ([default.rs#L260-L300](https://github.com/jkitchin/pounce/blob/5d8ad360feb13e49bf6aefe66e78a14ddd054229/crates/pounce-algorithm/src/init/default.rs#L260-L300)).

A linearized solution is a local model step, not automatically a better NLP starting point.

## Proposed design

Implement a sparse elastic normal-step initializer:

```
min_d,s  1/2 ||W(c(x)+Jd)-s||^2 + rho ||s||_1 + sigma/2 ||D d||^2
s.t.     l_x + margin <= x+d <= u_x - margin
         ||D d|| <= Delta
```

- Use sparse Jacobian/operator products and reuse the existing sparse QP/KKT infrastructure.
- Scale residual rows and variables.
- Globalize the step using actual/predicted feasibility reduction; shrink `Delta` or backtrack if the true nonlinear violation does not improve.
- Keep the best accepted point and return diagnostics (initial/final violation, step norm, rejected trials, termination reason).
- Handle inconsistent linearizations through elastic variables rather than an all-or-nothing least-squares solve.

## Acceptance criteria

- No `O(n^2)` identity or dense `m x n` Jacobian allocation on sparse problems.
- Bound interiority is preserved.
- Accepted steps monotonically improve a documented nonlinear violation merit; otherwise the original point is returned.
- Tests cover rank-deficient Jacobians, inconsistent rows, poor linearizations, and a large sparse model.
- Benchmark against the current projection on feasibility gained per function/Jacobian evaluation.

PR #633, opened 2026-08-16
Safeguarded LS init: the two downgrades are not tunable away, and why (#616)

### LINKED ISSUE #616: Safeguarded LS init: two fixtures downgrade to SolvedToAcceptableLevel under least_square_init_primal=yes
Follow-up to #605. Filed so the regression measured there has an owner rather than sitting in a PR body — per `CLAUDE.md`, a regression accepted as the cost of a fix and recorded nowhere else is indistinguishable from noise to the next reader. **This issue is unassigned and needs an owner.**

## What was measured

Sweeping the 57 CLI fixtures with `least_square_init_primal=yes` against baseline `70bf53ded3d893cfa2da6ead5195fda5ac096f68`, the #605 branch is a large net win — 2067 → 1687 total iterations (−18%), ten fixtures faster, several dramatically (`deb7` 479→202, `pooling_rt2stp` 134→81, `csfi2` 53→35).

It also costs three things:

| fixture | baseline | with #605 | |
|---|---|---|---|
| `csfi2` | `SolveSucceeded` | `SolvedToAcceptableLevel` | obj bit-identical at 55.0176045 |
| `eigenb2` | `SolveSucceeded` | `SolvedToAcceptableLevel` | 1.6 → 1.599999991 |
| `pooling_rt2stp` | obj −4391.83 | obj −3273.95 | worse local optimum |
| `unbounded_cubic` | 91 it → `DivergingIterates` | 290 it → `DivergingIterates` | diverges either way |

`SolveSucceeded` count goes 46 → 44. The solved-or-acceptable set is unchanged at 46, and the downgrades are deterministic across repeated runs.

## Why it happens

#605 makes the initializer decline a linearized normal step that would worsen the true nonlinear violation, and accept a backtracked one otherwise. That guarantee is about **the starting point's violation** — the quantity it measures. It says nothing about the trajectory that follows. A different (more feasible) starting point is entitled to reach a different local minimum on a nonconvex model, and to converge to a different tolerance band.

`pooling_rt2stp` and `deb7` move in opposite directions on the same mechanism (`deb7` improves, 249.75 → 97.56), which is the signature of local-minimum selection rather than a defect in the safeguard.

## Why it is not blocking #605

- Default options are **bit-identical across all 57 fixtures**; `least_square_init_primal` defaults to `no`.
- Under `mehrotra_algorithm=yes` the picture is neutral: same 27 solved, same objectives, 292 iterations on both sides.
- The option only became settable at all in #613, so no existing user configuration reaches this.

## What to do

Decide whether the tolerance downgrade on `csfi2` / `eigenb2` is acceptable for a route that is off by default, or whether the accept test needs a tighter criterion than "violation did not get worse" — e.g. also requiring that the accepted point not degrade the dual residual, or an acceptance band that prefers the untouched point when the improvement is marginal.

Worth checking first whether the two downgrades share a mechanism with `unbounded_cubic`'s 91 → 290, or are independent.

## Reproduce

```sh
scripts/sweep-fixtures.sh <binary> <out> least_square_init_primal=yes
```

against `70bf53d` and the #605 branch head.

jkitchin:
Fixes #616

> **Merged up before opening.** Baseline for every measurement is `a44f4e8b`; `main` has since moved to `66cc1d4`. The only conflict was `CHANGELOG.md` at the top of `[Unreleased]` — exactly as this branch's conflict note predicted — resolved by keeping both entries in `ad5c79c`. Re-verified on the merged tree: the 12 new tests pass (5 + 7), `cargo test --workspace --exclude pounce-hsl` **2985 passed / 0 failed**, `cargo fmt` clean, both consistency scripts OK.

## Problem

gh#605 gave the least-square primal initializer a backtracking safeguard: the min-norm solution of the *linearized* constraints is accepted only when it actually reduces the true nonlinear violation. Under `least_square_init_primal=yes` — off by default — that costs two tolerance downgrades, and #616 was filed so the cost had an owner instead of sitting in a PR body.

**The issue's numbers were measured against `70bf53de`. They reproduce exactly on current `main` (`a44f4e8b`), after #613, #619, #614, #623, #620, #628 and #627.** Measured by building a binary with the safeguard bypassed (pre-#605 semantics) from the same commit, so the comparison isolates #605 rather than main's drift:

| fixture | issue's number (70bf53d) | reproduced: unsafeguarded on `a44f4e8b` | reproduced: `main` today |
|---|---|---|---|
| `csfi2` | `SolveSucceeded` → `SolvedToAcceptableLevel`, 53 → 35 it | `SolveSucceeded`, 53 it, obj 55.01760453 | `SolvedToAcceptableLevel`, 35 it, obj 55.01760453 |
| `eigenb2` | `SolveSucceeded` → `SolvedToAcceptableLevel`, 1.6 → 1.599999991 | `SolveSucceeded`, 55 it, obj 1.6 | `SolvedToAcceptableLevel`, 57 it, obj 1.599999991 |
| `pooling_rt2stp` | obj −4391.83 → −3273.95, 134 → 81 it | 134 it, obj −4391.826001 | 81 it, obj −3273.954992 |
| `deb7` | 479 → 202 it, 249.75 → 97.56 | 479 it, obj 249.7459752 | 202 it, obj 97.55993789 |
| `unbounded_cubic` | 91 → 290 it, `DivergingIterates` both | `DivergingIterates`, 91 it | `DivergingIterates`, 290 it |
| `SolveSucceeded` count | 46 → 44 | 46 | 44 |
| solved-or-acceptable | unchanged at 46 | 46 | 46 |

Every row reproduces. Fourteen fixtures move in total.

The one figure that does **not** reproduce is the total-iteration headline: the issue reports 2067 → 1687 (−18%), and on `a44f4e8b` the same sweep reads **5426 → 5245 (−3.3%)**. That is corpus drift, not a changed conclusion — `issue_508_infeasible_gap_1em4` (3000 iterations) and `cresc4` (742) dominate the current total and are identical on both sides. The set of fixtures that move is exactly the set the issue describes.

## Cause

The issue asked, first, whether the `csfi2`/`eigenb2` downgrades share a mechanism with `unbounded_cubic`'s 91 → 290. **They do not — and the two downgrades do not share one with each other either.** Attributing all fourteen moving fixtures through `LeastSquareInitReport` puts them in three different arms of the safeguard:

| arm | fixtures | what happens |
|---|---|---|
| `θ₀ = 0`, short-circuit | `unbounded_cubic`, `unbounded_exp`, `boxed_qp_fixed_var` | already feasible; no direction is ever computed |
| declined | `csfi2`, `deb7`, `pooling_rt2stp`, `linear_eq_aggregation`, `linear_eq_aggregation_row_constant`, `issue_372_infeasible_bounds` | all four trials worse than `θ₀`; the user's `x` is kept |
| backtracked accept | `eigena2`, `eigenb2`, `hs71_obj1e8`, `user_scaling_suffix`, `user_scaling_var_suffix` | accepted at `α < 1` |

`csfi2` is in the *declined* group (`θ₀ = 1508.554`, 4 rejected trials). `eigenb2` is in the *accepted* group (`θ₀ = 1.0 → θ = 0.25` at `α = 0.5`). `unbounded_cubic` never reaches either test. Three arms, three mechanisms, independent.

That answer settles the design question, because it is what makes both of the issue's proposed remedies unreachable:

- **`csfi2` cannot be recovered by a tighter test.** The safeguard already declines there. Its old `SolveSucceeded` came from taking a step that *raises* the true violation above 1508.554 — the one thing gh#605 exists to refuse. A stricter criterion still declines. Only removing the safeguard gets it back.
- **`eigenb2` cannot be separated from a win.** `eigena2` hands the safeguard **bit-identical** numbers — `θ₀ = 1.0`, `θ = 0.2500000062500001`, `α = 0.5`, one rejected trial, step norm 3.2596011939729705 — and improves (78 → 65 iterations, same objective) where `eigenb2` drops a band. Any criterion computed from the safeguard's own inputs treats the two alike.

## Fix

Three candidate criteria were measured, including both the issue named. All three fail:

| candidate | result | why it fails |
|---|---|---|
| Retune `least_square_init_accept_ratio` | rejected | Acceptance is `θ₀ − θ ≥ η·α·θ₀`, so `eigenb2`'s trial survives every `η ≤ 1.5`; `η > 1` is meaningless (it demands a negative violation at `α = 1`). No reachable setting rejects it. |
| Acceptance band preferring the untouched point when the improvement is marginal | rejected | `eigenb2`'s step is not marginal — it cuts the violation 4×, the **median** of the sixteen accepted steps in the corpus, the same ratio as `airport` (103.6 → 25.96), `cresc4` (1715.3 → 437.2) and both `issue_508_infeasible_gap_*`, all wins. A band tight enough to exclude it excludes those too. |
| Require the accepted point not to degrade the dual residual | rejected | Measured: iteration-0 `inf_du` *improves* on both, 100 → 13.9 on `eigena2` and 100 → 47.7 on `eigenb2`. The gate accepts the step. |

So the decision is: **the downgrades stay.** They are tolerance-legal answers at the right objective, on a route that is off by default, and recovering them means giving back `deb7` 479 → 202 and `pooling_rt2stp` 134 → 81 — which, per the issue's own framing, is not a fix. The PR is documentation plus tests that pin current behaviour.

Two supporting changes were made, neither touching the solve:

1. **The safeguard's decision is emitted through `tracing`**, once per solve at `debug` on `pounce::algorithm`. `least_square_init_report()` is unreachable from the CLI, and the fixture sweep runs the CLI, so taking this measurement meant patching `init/default.rs` to print the report and rebuilding the workspace — once per hypothesis. The next investigator runs `RUST_LOG=pounce::algorithm=debug`.
2. **The accept test is a named predicate**, `DefaultIterateInitializer::accepts_trial`, so the argument above is pinned as executable properties rather than prose.

### Also found and documented: a declined step is not a no-op

Not in the issue, found while attributing it, and it corrects what the docs implied. Declining restores the user's `x` exactly, but not the solver's state: `calculate_least_square_primals` has by then driven the first factorization through the augmented-system solver, on the `W = 0` least-square matrix rather than the first real KKT matrix.

Isolated by forcing a decline on either side of that call — declining *before* it is bit-identical to `least_square_init_primal=no` on every fixture, declining *after* it is bit-identical to the real safeguard. So the carrier is that one solve; the staging and the trial evaluations are free.

It shows on two of the eight declining fixtures — `pooling_rt2stp` (298 iterations with the option off, 81 declined, same objective and status) and `deb7` (154 against 202). The other six agree with `=no` exactly.

**Not changed**: making the decline a true no-op needs a separate augmented-system solver for the initializer and costs `pooling_rt2stp` 81 → 298 iterations to buy a tidier contract on an off-by-default option. That is a trajectory regression of exactly the shape CLAUDE.md warns about, so it is described rather than done.

## Blast radius

Baseline: parent commit **`a44f4e8bcb67e577fb27c97a628efaff9822e494`**, release binary built from a `git worktree` of that commit, kept outside the repo. `scripts/sweep-fixtures.sh`, all 57 fixtures, three regimes:

| regime | result |
|---|---|
| default options | **bit-identical** to the parent |
| `least_square_init_primal=yes` (the regime that reaches the code) | **bit-identical** to the parent |
| `mehrotra_algorithm=yes` | **bit-identical** to the parent |

Nothing moves, which is the intended result: this branch adds a `tracing::debug!` line, extracts a predicate without changing it, and writes documentation and tests.

For completeness, the **`mehrotra_algorithm=yes` neutrality claim in the issue was re-checked** against the unsafeguarded binary, since it was measured at `70bf53d`. It still holds in the sense that matters — same 27 `SolveSucceeded`, same objectives — though the "292 iterations on both sides" figure is stale and now reads 2475 → 2463. Twelve fixtures move there; ten fail on both sides, so only the failure label and its meaningless objective change, and the other two are `eigena2` and `eigenb2`, which solve to the same objectives either way and differ by one iteration.

Across the 57 fixtures the safeguard engages on 29 (16 accept, 8 decline, 5 short-circuit). It is inert on the other 28: 26 are LP or convex-QP models the CLI dispatches to `pounce-convex`, which does not run this initializer, and 2 have no constraints for the step to act on.

## Tests

Two new files, 12 tests.

`crates/pounce-cli/tests/issue_616_ls_init_downgrades.rs` (7) pins the corpus measurement, so the decision is visible to the suite rather than only to a PR body — the failure mode CLAUDE.md names, and the one that let gh#544's 206 → 812 ship. Status and objective throughout, plus three iteration-count *relations* that each carry a mechanism claim; no absolute iteration count is asserted.

`crates/pounce-algorithm/tests/issue_616_ls_init_accept_test.rs` (5) pins the algebra that makes the corpus result general rather than a coincidence of two models.

**How they bite, stated precisely:**

- `the_safeguard_decision_is_attributable_from_the_cli` **fails on the parent commit for the right reason** — the debug line does not exist there, and the assertion says so: *"RUST_LOG=pounce::algorithm=debug did not surface the safeguard's decision"*. This is the one behavioural change on the branch, and this test is its regression test.
- **The other six CLI tests pass on the parent, and are meant to.** They pin behaviour this branch deliberately leaves alone. That is inherent to `VERDICT: no-improvement` and is not dressed up as anything else.
- **The five algorithm tests do not compile on the parent**, because `accepts_trial` is new there. That is a missing-API failure, not a behavioural one, and it is not offered as evidence that they bite.

Verified by copying both files into a `git worktree` at `a44f4e8b` and running them.

**Full suite** (branch, before merge-up): `cargo fmt --all -- --check` clean; `cargo clippy --workspace --exclude pounce-hsl --all-targets` no errors (the two `expect_used` warnings on the new CLI test match the surrounding test files); `cargo test --workspace --exclude pounce-hsl` 229 test binaries, 0 failures; `python/tests` 1114 passed, 7 skipped (plotly, ipopt, gamsapi, two missing `.nl` fixtures — all environmental and all reproduce on the parent); `pyomo-pounce/tests` 352 passed; `scripts/check-docs-consistency.sh` OK; `scripts/check-release-consistency.sh` OK.

**After merging `66cc1d4` up** (`ad5c79c`): the 12 new tests pass (5 + 7), `cargo test --workspace --exclude pounce-hsl` **2985 passed / 0 failed**, `cargo fmt` clean, both consistency scripts OK.

## Conflict note

Test-merged both sibling branches into this one:

| branch | PR | conflict |
|---|---|---|
| `claude/pounce-issue-609-loop` | #629 | **`CHANGELOG.md`** — both add at the top of `[Unreleased]`. Nothing else. *(#629 has since merged; resolved in `ad5c79c`.)* |
| `claude/pounce-issue-610-loop` | #630 | **`CHANGELOG.md`** — same. Nothing else. *(Still open.)* |

One collision each, textual, at the top of `[Unreleased]`; resolution is to keep both entries.

Worth flagging: **#630 also edits `docs/src/initialization.md`**, which the issue brief did not list. It does **not** conflict — #630 adds at the "No good starting point at all?" section, this branch edits the "Safeguarding the least-square start" section — but both land in the same file, so whichever merges second should re-read the page rather than trust a clean auto-merge.

## Why `VERDICT: no-improvement` is the right label

The issue asked for a decision, and the decision is that neither remedy it proposed is reachable — established by measuring three candidate criteria and refuting each with a number, not by declining to try. What ships is the attribution mechanism that made the argument possible (`RUST_LOG=pounce::algorithm=debug`), the predicate that makes it executable, tests that pin the corpus cost so it cannot silently drift, and documentation of both the decision and the declined-step-is-not-a-no-op finding that turned up alongside it.

No follow-up issues filed.

---


jkitchin:
## Review: approved

Reviewed the whole diff and re-ran the new tests on macOS (arm64, release).

**The Rust change is behaviour-neutral, and I checked rather than assumed.** The
extracted predicate

```rust
theta.is_finite() && actual >= eta * predicted && theta < theta_0
```

is term-for-term the inlined original, same operand order, same short-circuit
order, so it cannot differ even on the NaN/inf paths. The only other executable
line added is one `tracing::debug!` per solve on `pounce::algorithm`, which is
silent at the default level. Everything else in `default.rs` is doc comment.

**New tests, run here:**

```
issue_616_ls_init_accept_test:  5 passed
issue_616_ls_init_downgrades:   7 passed
```

That matters more than usual, because seven of those assertions are on solver
*status* and two are iteration-count relations — the numbers most likely to be
platform-specific. `eigena2` still `SolveSucceeded` and `eigenb2` still
`SolvedToAcceptableLevel` on macOS, so the pair that gh#616's whole argument
rests on disagrees here too, and `pooling_rt2stp`'s declined-vs-`=no` iteration
counts still differ. The choice to assert relations (`==`, `!=`) rather than
absolute counts is the right one and is why this survives a platform change.

**I re-measured the census independently**, driving all 57 CLI fixtures through
the new debug channel and bucketing on `termination`:

```
29 engage / 28 inert
16 accepted | 8 declined | 5 "x0 already feasible"
```

Exactly the numbers in `docs/src/initialization.md`. The declined eight are
`csfi2`, `deb7`, `pooling_rt2stp`, `linear_eq_aggregation`,
`linear_eq_aggregation_row_constant`, `issue_372_infeasible_bounds`, `jit1`,
`jit1_boxed`; the five feasible starts are `autocorr_bern55-06`,
`convex_qp_sens`, `unbounded_cubic`, `unbounded_exp`, `boxed_qp_fixed_var`.
Incidentally this is a good demonstration of the debug line's value — taking
that census was one shell loop against a stock binary, which is precisely the
thing the issue says used to require patching the source.

The `csfi2` argument is the strongest part of the writeup: it *declines*, so no
tightening of the accept test can reach it and only removing the safeguard can,
which turns "should we tune this?" from a judgement call into a closed
question. The `eigena2`/`eigenb2` pair closes the other half — bit-identical
inputs, opposite outcomes, so no function of those inputs separates them — and
`the_eigenb2_step_is_a_median_sized_reduction_not_a_marginal_one` correctly
pre-empts the obvious follow-up proposal by showing the corpus tail a
rejection band would take with it.

The "a declined step is not the same as never asking" finding is the one I did
not expect and is worth the page it gets. Isolating it by forcing a decline on
either side of the augmented-system solve is a clean experiment, and stating
plainly that the tidier contract was *not* bought because it would cost
`pooling_rt2stp` 81 → 298 is exactly the right call and exactly the right way
to record it.

### One nit, not blocking

The arm breakdown of the *moving* fixtures does not add up the same way in both
places. `CHANGELOG.md` says "three start already feasible ... six decline every
trial ... five accept a backtracked step" — 3 + 6 + 5 = 14, matching "fourteen
fixtures move". `docs/src/initialization.md`'s bullets name 3 + 5 + 5 = 13, so
the declined bullet is one name short of its own count. From the census above
the missing one is among `linear_eq_aggregation_row_constant`, `jit1` and
`jit1_boxed`; I can't tell which without an unsafeguarded binary, so I have not
guessed. Worth a one-line fix on a later pass — in a PR whose subject is
recorded numbers staying honest, that is the kind of thing worth getting right.

Note for after merge: #630 also edits `docs/src/initialization.md`, in the
racing-starts section, and merges first. Different section, no textual conflict,
but per this PR's own body the second one in should re-read the page — I'll do
that.

Merging.

jkitchin:
Nit fixed in **`486d445`** — the missing name is **`linear_eq_aggregation_row_constant`**, one of the three you shortlisted.

No unsafeguarded binary needed to resolve it: the PR body's own attribution table already names all six declined-and-moving fixtures, and `docs/src/initialization.md` reproduced five of them. So it was a transcription drop from this PR's own measurement, not an open question — `csfi2`, `deb7`, `pooling_rt2stp`, `linear_eq_aggregation`, **`linear_eq_aggregation_row_constant`**, `issue_372_infeasible_bounds`. The arms now sum 3 + 6 + 5 = 14, matching both the "fourteen fixtures move" sentence in the same section and the CHANGELOG.

`jit1` / `jit1_boxed` are the other two of your declining eight; they decline but do not *move* against the unsafeguarded binary, which is why they are absent from a breakdown of the fourteen. Two different sets — the census counts what the safeguard does across all 57, the arm breakdown only what moves — and the docs could say that more plainly than they do.

`check-docs-consistency.sh` OK. Docs-only, so the 12 tests and the sweep are unaffected; CI is re-running on the new head.

Thank you for the second-platform run. The status assertions surviving macOS is the part I could not have checked from here, and `eigena2`/`eigenb2` disagreeing the same way on arm64 is what makes the "bit-identical inputs, opposite outcomes" argument portable rather than an x86 coincidence. Your independent census reproducing 29/28 and 16/8/5 off the new debug channel is also the first outside use of that channel, which is a better argument for it than anything in the writeup.

---

PR #636, opened 2026-08-17
Make a transferred warm start pay, and stop taxing one that carries only a point (gh #622)

### LINKED ISSUE #622: A transferred (mapped) warm start is slower than a cold solve, and gets worse with horizon
Split out of #607 (PR to follow), which adds `WarmStart.transfer(prob, mapper)` — the horizon-shift / changed-discretization hook the issue asked for. It is correct, and it is **slower than not using it**. Recording the numbers so the next reader does not re-run the experiment, and so the shortfall has an owner rather than living only in a "known limitations" section.

## Measured

Receding-horizon fixture, transferred point vs. cold solve, iterations to `Solve_Succeeded`:

| horizon | transferred | cold |
|---|---|---|
| slew-limited fixture | 12 | 7 |
| 5 (sinusoidal track) | 12 | 9 |
| 10 | 15 | 11 |
| 20 | 22 | 9 |
| 40 | 30 | 10 |

The gap widens with the horizon. Dropping the carried `mu`, loosening it, and loosening `bound_push` change none of it — six variants on the test fixture all land in 11–13 iterations.

## Why

This is the barrier/active-set limit `docs/src/initialization.md` already describes: a point that is optimal for the previous parameter is interior-infeasible or on the wrong active set for the new one, and the filter spends its first iterations undoing the transfer. The prolongated tail of a shifted horizon is the worst case — those variables have no meaningful history at all.

#607's horizon test therefore asserts the answer and the `mapped` labelling and explicitly does **not** assert an iteration win, with these numbers in a comment.

## What might fix it

- #620 (issue #606) attacks exactly this on the Rust side with residual-adaptive recentering; the transferred point is the case it should help most. **Re-measure this table once #620 lands** — it may close some or all of the gap on its own.
- Seeding the prolongated tail from the model rather than from the mapper's `fill_x` (currently unmatched multiplier entries are seeded `NaN`, the native "unseeded" contract).
- A tangent predictor rather than a zero-order transfer — which is #608's continuation driver, where this shortfall is the motivating measurement.

## Acceptance

A transferred warm start beats a cold solve on the receding-horizon fixtures, at every horizon in the table above — or the API documents that it is a correctness/provenance mechanism only and stops implying otherwise.

Related: #607, #620, #608, #90.

jkitchin:
Fixes #622

## TL;DR

Everything below is measured on the parent commit `66cc1d4` against this branch. Lower is better; **bold** is this branch.

**Transferred (mapped) warm start — the defect #622 filed:**

| | before | after |
|---|---|---|
| slew fixture, one hand-off | 11 (cold 7–9) | **8** |
| slew fixture, 3-step closed loop | 27 (cold 22) | **21** |
| sine family, 8-step loop, horizon 40 | 53 (cold 76) | **46** |

**Warm start carrying only a point — the defect found while measuring it:**

| | before | after |
|---|---|---|
| HS071 from a transferred point | 11 (plain solve from same x: 6) | **7** |
| HS071 restarted from its own solution | 3 | **2** |
| `benchmarks/warmstart` corpus, new `values-ipm` arm | 5490 | **4385** (39 of 42 rows) |
| `moving_bound_qp` alone | 1040 | **428** |
| tightening the five `warm_start_*` pushes | +4 iterations | **no effect** |

**Unchanged, deliberately:** `cold-ipm` 10288, `warm-ipm` 3404, `pred-ipm` 1258, `predcorr-ipm` 1242 — bit-identical, 0 of 42 rows. CLI fixture sweep identical across all 57 models.

And the headline the issue asked for — re-measure the table once #620 lands. It inverted. Transferred against cold, 8-step loop:

| horizon | transferred | cold |
|---|---|---|
| 5 | 45 | 67 |
| 10 | 50 | 75 |
| 20 | 54 | 77 |
| 40 | 46 | 76 |

Most of that turn is #620's residual-adaptive recentering, measured here for the first time since it landed. The rest is this PR.

## Problem

Four things, each found by measuring the one before it.

1. **The prolongated stage was filled with a number chosen without reference to anything.** `WarmStart.reindex` gave variables the target has and the source does not `0` clipped into the box. On a chain with a slew limit that enters the new problem **2.25 away from feasible** — new variable at `0` beside a neighbour at `2.75`, limit `0.5` — and the filter's first iterations go on undoing the transfer: 11 iterations against 7 cold.
2. **`warm_start_init_point=yes` taxed a caller who had only a point.** Raised by @devin-griff with a controlled ipopt-vs-pounce table; reproduced here, smaller — 11 iterations against 6 for a plain solve from the same `x`, and *worse* the tighter `warm_start_mult_bound_push` was set.
3. **`transfer()` silently replayed stale duals.** A mapper that returns `x` and omits the multipliers gets them carried over unchanged — right for a mapper that nudges a point, wrong when the identifiers say the entries moved. Same length, so `_check_lengths` passes.
4. **The warm-start suite could not see any of this.** Every arm seeds duals, so the corpus was bit-identical across the fix for defect 2.

## Cause

1. A fill policy with no information in it. `0` is not "no answer", it is an answer, and on a coupled model it is usually a violated constraint.
2. `TNLP::get_starting_point` is asked for `init_z` and may leave the blocks untouched — exactly what a values-only caller does — and the buffer behind them was pre-filled with **zero**. Zero is a *legal multiplier value*, so it sailed past the "was this seeded?" resolution performed on NaN and was merely floored at `warm_start_mult_bound_push`: 1e-3 by default, **1e-9 under the tightened pushes `pounce.WarmStart` itself ships**. `z = 1e-9` declares every bound inactive and breaks complementarity against μ before iteration one. Finding it took three attempts — the pre-fill I corrected first was overwritten in `initialize_starting_point`, then again by `fetch_warm_start_snapshot`; an `eprintln` settled in one run what code reading got wrong twice.
3. Carry-over is the documented default and there was no check that the axis had not moved.

## Fix

**`fill_x` is a policy.** Default `"prolong"` repeats the last stage, reading the stage's *period* off the identifier map rather than assuming a layout: when the map is a pure shift, each unmatched entry takes the value one period behind it, clipped into its box. `(p, v, u)` interleaved lands in the same *kind* of slot. Not a shift → degrades to the old behaviour, still available as `fill_x="zero"`.

*Trade-off, stated because it is real:* `"zero"` runs the sine loop in 42/47/55/53 against `"prolong"`'s 45/50/54/46 — ahead at the two short horizons, behind at the two long ones, and 27 against 21 on the slew fixture. The default is not the arm that wins every row; it is the one that never hands the solver a point the model rejects.

**The bound-multiplier buffers carry the unseeded marker**, so an absent block and a NaN block are the same statement. That exposed a second half: #606's reconstruction was gated wholesale on some dual having been supplied, but only half of it needs one — `z = μ / slack` needs the slacks the point already determines, while the least-squares `y` needs a dual to complete and from a bare point is the cold path's estimate wearing the warm path's barrier (the 1102 → 1211 measurement behind the gate). **The gate is now drawn between those two.** μ escalation stays on the gated side: with the blocks filled to `μ / slack` the measured complementarity *is* μ, and escalating off it is the barrier arguing with itself.

**`transfer()` refuses to carry a dual across an axis that moved**, per axis, only on a non-identity map, silent when identifiers are absent. `pounce.Continuation` passes none and is unaffected. I walked into this myself: a mapper returning only `x` produced numbers that looked like a predictor performing badly and were actually stale duals in the wrong slots.

**`values-ipm`**, a new suite arm seeding the previous point alone. Not synthetic — it is the default on every frontend carrying variable levels but no duals (GAMS `x.L`, a Pyomo model whose `dual` Suffix was never loaded, a `.nl` without dual guesses).

**Tried and rejected, all measured:**

* *Drop the transferred multipliers* — close to a wash once the values-only path is fixed (44/55/55/47 against the carried 45/50/54/46). What the carried block buys is the *equality* multipliers, the half not reconstructible from a bare point.
* *Restart the barrier* — a wash at `mu_init=1e-6`, losing above it; by `1e-1` the four horizons run 72/79/74/72, at or above cold's 67/75/77/76.
* *Cap the unseeded fill at `bound_mult_init_val`* — tried against `degenerate_vertex` (below). Costs 341 there instead of 396, but regresses `redundant_rows` 162 → 292 and the corpus total 4385 → 4482. Dropped.
* *A secant predictor* — worse than zero-order everywhere (59/75/66/60), at horizon 10 no better than cold.

## Blast radius

Baseline `66cc1d4`.

* **CLI fixture sweep: identical across all 57 models.**
* **Corpus: bit-identical** on `cold-ipm` / `warm-ipm` / `pred-ipm` / `predcorr-ipm`. Anything that seeds duals is untouched by construction — the changed branch is reached only when `any_dual_seeded` is false. Pinned #606 behaviours (exact restart 2/1, partial 3/2, stale 43/36) unmoved.
* **One family pays, and the new arm is what priced it: `degenerate_vertex` 220 → 396.** Twelve rows tight in 4 variables means the true multipliers are a mass of ties near zero, and the old fill — a push small enough to read as "every bound inactive" — was accidentally right about them. Every honest fill loses there; the cap above was the attempt and it cost more elsewhere. This is the price of the 2.4× on `moving_bound_qp`, and the arm will notice if it moves again.
* **`warm_start_recentering=none` on HS071 restarted from its own solution: 3 → 4.** The kill switch keeps the constant fill and the constant is now `bound_mult_init_val`. Under the default the same restart is 3 → 2.
* Python `reindex` default changes what lands in a prolongated stage; `transfer()` with an explicit mapper is unaffected by the policy, and newly refuses the stale-dual case above.

## What a better transfer would be worth (and why two routes fail)

Recorded in `docs/src/initialization.md` so it is not re-derived. Oracles no transfer can beat, seeding each window with the *next* window's converged answer:

| horizon | cold | shipped | perfect primal | perfect primal+dual |
|---|---|---|---|---|
| 5 | 67 | 45 | 21 | 9 |
| 10 | 75 | 50 | 24 | 12 |
| 20 | 77 | 54 | 26 | 18 |
| 40 | 76 | 46 | 23 | 18 |

So the barrier's floor is ~1 iteration per warm step and about half of what a transfer spends is the zero-order prediction. Neither obvious route collects it. The secant is worse than zero-order. The KKT tangent (`Solver.parametric_step`, the machinery behind `pred-ipm`) **cannot be pointed at a horizon shift at all**: on the stages two windows share, θ does not move — what changes is which stages exist — so it is handed a delta vector of exact zeros and correctly returns a zero step, bit-identical to zero-order. Measured on a fixture built to route θ through pin-equality rows. Given a parameter that genuinely moves (an MPC initial-condition pin) the tangent degrades with horizon: 52/87/95/137 against 54/78/81/93. What is left is the freshly-entered stage, which has no history to extrapolate from — a model rollout, which `transfer()`'s mapper already accepts.

## Tests

Rust, `crates/pounce-algorithm/tests/warm_start_recentering.rs`:

* `a_values_only_warm_start_is_not_taxed_by_the_tightened_pushes` — new. **Fails on the parent**: 11 iterations there against a cold 6.
* `a_primal_only_seed_gets_its_bound_blocks_and_nothing_derived` — rewritten from `a_primal_only_seed_is_reported_unseeded_and_left_alone`, whose invariant this deliberately changes. **Fails on the parent** on `bound_duals == Reconstructed`.

Python, `python/tests/test_warm_start_schema.py`:

* `test_prolong_repeats_the_stage_a_stride_back`, `..._stays_in_the_box_and_yields_to_an_explicit_fill`, `..._falls_back_when_the_map_is_not_a_shift` — new, all **fail on the parent** (no fill policy there; `fill_x="zero"` parses as an array and raises).
* `test_a_mapper_that_moves_the_point_must_say_what_happens_to_the_duals` and `test_a_mapper_may_still_carry_the_duals_when_nothing_moved` — new, the guard and its negative case. The first **fails on the parent**, which accepts the mapper silently.
* `test_a_mapped_replay_carries_the_equality_multipliers_a_bare_point_cannot` — **fails on the parent** (`bound_duals: unseeded` for a bare point).
* `test_horizon_shift_transfers_and_replays` — new `moved.x[-1] == x0[-1]` assertion **fails on the parent** (fills `0.0`).
* `test_a_transferred_start_beats_a_cold_solve_over_the_horizon` — **passes on the parent too**, included deliberately: it pins #622's acceptance criterion at every horizon of its table, not this diff.

Deliberately not pinned: the single hand-off on the slew fixture, at 8 iterations against a cold solve's 7–9 — inside the cold arm's own spread, so there is no honest inequality to assert. `docs/src/initialization.md` says so in those words.

Full runs: **2974 Rust tests** (workspace, `--exclude pounce-hsl`, which fails to link against absent proprietary HSL libraries on the parent too), **853 Python tests**, `warmstart.selftest` clean.

---

- [x] Tests fail on the parent commit for the stated reason — with the one exception named above
- [x] Book pages updated (`docs/src/initialization.md`, `docs/src/warm-start-benchmark.md`, `benchmarks/warmstart/README.md`; no new page, so `SUMMARY.md` is unchanged)
- [x] `cargo fmt --all -- --check` clean; `cargo clippy` 0 errors and no new warnings at the changed lines; `cargo test` clean

https://claude.ai/code/session_015oMY7jtPmkRYFFhMAEN9HQ


jkitchin:
Reviewed. Approving to merge, with three notes — one number to correct in the body, one gap that predates this PR, and one thing I checked specifically because the diff invited it.

## What I verified rather than took on faith

**The NaN pre-fill cannot poison an iterate.** This is the change that looks riskiest — writing `Number::NAN` into `z_l`/`z_u` in `StartingPointSnapshot` (`crates/pounce-nlp/src/orig_ipopt_nlp.rs`) — so I traced the resolution rather than trusting the description. `resolve_nan_seeds` is pre-existing, not added here; it handles `DenseVector` and `CompoundVector` recursively and `debug_assert!`s on a third impl; and it is reached only through `clone_clamped`, which is applied **unconditionally** at `init/warm_start.rs:370-375` to all six blocks with `z_nan = self.opts.bound_mult_init_val`. There is no path on which a NaN survives into the iterate. `lambda` correctly keeps its zero pre-fill: its unseeded resolution is also zero, and `any_dual_seeded` already reads an all-zero `y` as unseeded, so a NaN there would buy nothing and cost a branch.

**"CLI fixture sweep identical across all 57 models" — true by construction, so I did not re-run it.** `WarmStartIterateInitializer` is reachable only through the `if self.warm_start_init_point` arm at `alg_builder.rs:1203`; `final_mu` has exactly one caller, inside it. Every Rust line this PR changes is unreachable from a cold CLI solve. That is a stronger statement than a passing sweep and worth saying in the body next time, since it is checkable by reading.

**The gate split is drawn in the right place.** `z = μ/slack` needs only the slacks a bare point already determines; the least-squares `y` needs a dual to complete, and ungated it is the cold estimate wearing the warm barrier — the 1102 → 1211 measurement. Routing μ escalation to the gated side via `mu_may_move` is the same argument: with the blocks filled to `μ/slack` the measured complementarity *is* μ, so escalating off it is circular. Correct.

**Local runs on this merge commit (macOS arm64):** workspace `cargo test --workspace --exclude pounce-hsl` → 231 binaries, **2986 passed, 0 failed**; `cargo fmt --check` clean; `python/tests` → **902 passed, 38 skipped, 1 failed**. The one failure is `test_continuation.py::test_step_controller_is_what_the_jax_follower_uses`, a bare `ModuleNotFoundError: No module named 'jax'` with no skip guard — an environment gap #638 documents independently, not this PR.

## 1. The `7` in the headline table reads `8` here

> HS071 from a transferred point | 11 (plain solve from same x: 6) | **7**

The test prints, on this merge commit:

```
HS071 values-only warm start: cold=6 switch+tight=8 switch+default=8 switch+tight, NaN-seeded=8
```

I checked whether the merge moved it: **the pre-merge head `63e47d96` prints the same `8`.** So this is a platform difference between the Linux machine the PR was measured on and macOS arm64, not something the merge introduced, and it does not block.

Two things follow. Every qualitative claim still holds exactly here — `it_tight == it_loose == it_nan`, `mu_in == mu_out`, `bound_duals == Reconstructed` — and the neighbouring numbers reproduce (primal-only gives `none=4 residual=2`, matching the stated 3 → 2 default and 3 → 4 under `recentering=none`). But `assert!(it_tight <= it_cold + 2)` is satisfied here **exactly at its boundary**, 8 against 6+2. That is the right way for it to fail — one more iteration on any platform is a red test, not silent drift — so I would leave the assertion alone and instead make the table say `7 (8 on macOS arm64)`, so the next reader who measures 8 does not go looking for a regression the way I just did.

## 2. `degenerate_vertex` 220 → 396 is accepted, and nothing automated will notice if it moves again

The body says "the arm will notice if it moves again". It will not, unattended: `grep -rn warmstart .github/workflows/ Makefile` returns nothing. No workflow and no make target runs `benchmarks/warmstart`, so the new `values-ipm` arm only notices when someone runs it by hand.

That gap predates this PR and I am not asking it to close it. But CLAUDE.md is specific that a measured regression recorded as the cost of a fix needs an issue and an owner, and the reason is exactly this — 206 → 812 sat in a commit message through a release. What this PR does have going for it is far better than that precedent: `docs/src/warm-start-benchmark.md:79` records the mechanism (twelve rows tight in 4 variables, so the true multipliers are ties near zero and the old under-fill was accidentally right), names the rejected alternative and what it cost elsewhere (`redundant_rows` 162 → 292, corpus 4385 → 4482), and prices it against the 2.4× on `moving_bound_qp`. That is a documented trade, not a buried one, and it is enough for me. Flagging the CI gap so it is a known unowned thing rather than an assumed-covered one.

## 3. Things I looked at because the diff invited it

* `_prolong` fills in the direction that reads only already-filled entries — `todo[::-1]` when `stride < 0`. Right, and worth the one-line comment it has.
* `_check_carried_blocks` raises per axis and only on a non-identity map, so `pounce.Continuation` (which passes no identifiers) is genuinely unaffected. Confirmed by reading, not just by the test.
* `fill_x="prolong"` losing to `"zero"` at the two short horizons is in the body rather than hidden, along with the reason the default is not the arm that wins every row. That is the right call and the right way to write it up.

Merging.

PR #637, opened 2026-08-17
casadi: solver diagnostics through stats(), and two defects found exposing them

jkitchin:
Refs #634. **Deliberately not a closing keyword** — this is items 1–4 of that issue. Native non-Python builds and prebuilt plugin archives are declined/deferred (below), and they are the reporter's asks, so the issue should stay open rather than auto-close on merge.

## Problem

gh#634 asks the CasADi plugin for six things. Four of them were reachable because POUNCE already collected the data and simply had no channel to report it through CasADi; two are a build/packaging commitment.

Two of the four turned out to be defects rather than missing features:

| | before | after |
|---|---|---|
| 3 solves on one solver object, `(iter_count, len(iterations['inf_pr']))` | `(12,13) (12,26) (13,40)` | `(12,13) (12,13) (13,14)` |
| `tol=1` (Python `int`) vs `tol=1.0`, iterations | 12 vs 9 | 9 vs 9 |
| `tol=1.0` vs `tol=1e-12`, iterations (control) | 9 vs 16 | 9 vs 16 |

Row 1: the per-iteration trace accumulated across solves, so `stats()["iterations"]` described every solve so far while `iter_count` beside it described only the last. Every receding-horizon loop hits this — the reporter's own use case.

Row 2: `{"tol": 1}` is an `int` in Python and a number to POUNCE. It was forwarded with `AddIpoptIntOption`, which refuses it, so `tol` silently kept its default while the script looked like it had set it. The 12 in that cell is the default `tol=1e-8`, not the requested `1`.

## Cause

**Trace.** CasADi's ipopt plugin clears its trace vectors at the top of `solve()` (`ipopt_interface.cpp:431`); this plugin never did, and a CasADi memory object is reused across calls.

**Option typing.** Forwarding dispatched on the `GenericType` the value happened to carry, i.e. on how the user typed the literal, rather than on the type POUNCE registers for that keyword.

**Everything else** was a reporting gap, not a computation gap: `GetIpoptPrimalInf`/`DualInf`/`ComplInf` already existed, `LinearSolverSummary` already fed the CLI's solve report, and `GetIpoptCurrentViolations` already worked live under the algorithm's intermediate-callback context.

## Fix

Three additive C entry points expose data POUNCE already collects — `GetPounceLinearSolverStats`, `GetPounceRestorationStats`, `GetPounceOptionType`. No new instrumentation, no existing signature changed.

`stats()` gains `final_inf_pr` / `final_inf_du` / `final_compl_inf`, `linear_solver` (the backend that *actually ran*, factorization counts, pattern reuse, fill ratio, pivot range, final inertia) and `restoration` (calls, inner/outer iters, wall seconds). These are harvested before `FreeIpoptProblem`, which takes the accessors with it; `m->prob` is nulled there so it doubles as the "solve in flight" flag rather than becoming a dangling pointer.

**Live diagnostics needed no new mechanism, and I rejected inventing one.** CasADi fixes the callback signature at `(x, f, g, lam_x, lam_g)`, so the first instinct is a plugin-specific accessor. Unnecessary: `stats()` is already callable from inside the callback, and the plugin records the current iteration into the trace *before* handing over control, so the last entry of each trace is the iteration you are in. Only the violation vectors were missing, and they are fetched on demand — present only while a solve is in flight, so a `stats()` call after the solve pays nothing. No key is served stale: live keys are absent afterwards, final keys absent during.

Option typing now asks POUNCE's registry. `GetPounceOptionType` accepts a NULL problem handle because the code generator has none, and that keeps the generated and interpreted paths from disagreeing about an option's type.

### Rejected

**Per-iteration restoration flag (`alg_mod`).** Requested, and it looks like a one-line change: the value reaches the plugin as the callback's leading argument. It is a constant. `IpoptAlgorithm::build_iter_stats` hardcodes `mode: AlgorithmMode::RegularMode` and the callback fires only from the outer loop — restoration's inner solve does not drive it. Publishing it would ship a column that is always `0` and reads as working restoration detection. Solve-level `stats()["restoration"]` ships instead, verified non-zero on a solve that enters restoration and zero on one that does not. Making the flag real means firing the user's callback from the restoration inner solve, which changes how often every existing callback fires — its own issue.

**Linear-solver phase timings.** Requested, with the reporter's own caveat that it was "not a request to add instrumentation that does not otherwise exist". POUNCE does not instrument analyse/factor/back-solve separately, so they are **absent** from the dict rather than reported as zero, which would read as "instantaneous" instead of "unmeasured". Pinned by a check that the dict contains no timing keys.

**Native CMake/GCC builds, prebuilt archives, Windows.** Windows is declined — not supported by this project. Worth recording that it would not have been one build anyway: the wheel path targets MSVC to match CasADi's own wheels while the issue asks for MinGW to match a native distribution, which are two incompatible binaries for one platform. The rest is written up with what it would cost in `dev-notes/casadi-diagnostics-and-native-builds.md`, including that most of it evaporates if the interface goes upstream — which the reporter offered to help broker.

## Blast radius

No trajectory change: nothing here touches step selection, scaling, or which correction the solver reaches for, so no fixture sweep applies. The one behavioural change to an existing path is option typing — an option that previously reached POUNCE now still reaches it with the same value, except for the case that was previously *refused*, which now takes effect. That is the fix.

C ABI is additive only; existing cyipopt / JuMP / AMPL / GAMS callers are unaffected. `stats()` gains keys and removes none.

Also fixes a stale doc block in `pounce.h` claiming `GetIpoptCurrent*` return false "until pounce's algorithm core invokes the intermediate callback (currently a follow-up)". They have been live since the context guard landed; the comment told native callers the API was a stub.

## Tests

23 new checks in the CasADi parity suite (73 total), 3 new unit tests in `pounce-cinterface`.

**Verified to bite.** I built the parent commit (`eeca47f`) in a worktree and ran the new suite against it. All six new test functions fail there, each for its own reason — `final_inf_pr` raises `KeyError`, the trace check reports `(12,13) (12,26) (13,40)`, the violation and linear-solver checks fail on absent keys.

That exercise caught a bad test of my own, fixed in bb797e8: "`tol=1` takes fewer iterations than `tol=1e-12`" **passes pre-fix**, because the default tolerance the refused option left in place is also looser than `1e-12`. The check that actually distinguishes the two builds is that `1` and `1.0` must produce the same solve — 12 vs 9 pre-fix, 9 vs 9 post-fix.

Deliberately not pinned: the exact iteration counts above are from casadi 3.7.2 on Linux x86-64 and the checks assert relations between them, not the numbers.

Locally green against casadi 3.7.2: 73/73 parity checks, `cargo test -p pounce-cinterface` (52 lib + integration), `cargo fmt --all -- --check`, and `cargo clippy` with no new warnings (the one `pounce-cinterface` lib warning is pre-existing at `HEAD~1`). All eight `casadi/examples/` scripts run.

---

- [x] Book page under `docs/src/` updated, and linked from `SUMMARY.md` if new — `docs/src/casadi.md`, already linked

https://claude.ai/code/session_01FUXrfbxsYqpUuxpLmrBsSi

---
_Generated by [Claude Code](https://claude.ai/code/session_01FUXrfbxsYqpUuxpLmrBsSi)_


srikanth-gm:
prof. @jkitchin thank you for the PR for getting more detailed diagnostics for POUNCE. 

I understand Windows builds or maintaining native plugin artifacts are not in scope of this project. 

Is it however possible to have some docs or a CMake recipe in the repo on how to build the `nlpsol` plugin as native binaries for CasADi ? this way, we can pull in POUNCE into our CI and use the prescribed CMake recipe to build and link the plugin against CasADi native. We are building from source in our CI since they only release binaries for the python version of CasADi.

jkitchin:
Yes — and it turns out to need less than a CMake recipe, because the existing Makefile already builds without Python. Its CasADi inputs are all `?=`, so the values derived from `import casadi` are only defaults. Pushed in 6562d1f: the recipe is now documented rather than newly built.

```bash
cargo build --release -p pounce-cinterface

make -C casadi \
  CASADI_LIB=/opt/casadi/lib \      # holds libcasadi.so
  CASADI_INC=/opt/casadi/include \  # public headers
  CASADI_SRC=/src/casadi \          # include root holding casadi/core/ — see below
  CASADI_VER=3.7.2 \
  CXX11_ABI=1
```

Verified by building with `PYTHON=/nonexistent` and explicit paths; the resulting plugin passes all 73 parity checks.

Two things bite on that path, and both used to fail in ways that named neither the option nor the cause. `check-env` now catches them up front:

- **`CASADI_SRC` must be the directory *containing* `casadi/core/nlpsol_impl.hpp`.** CasADi's `INSTALL_INTERNAL_HEADERS` defaults to `OFF`, so a source-built CasADi's `make install` does **not** ship the internal headers a plugin subclasses. Either point `CASADI_SRC` at your CasADi source checkout's root, or configure CasADi with `-DINSTALL_INTERNAL_HEADERS=ON` and use `<prefix>/include` for both `CASADI_SRC` and `CASADI_INC`. That second form is probably what you want in CI, since you are building CasADi anyway.
- **`CASADI_VER` has to be passed explicitly.** With no Python to read `casadi.__version__` from it is empty, the version macros get defined as nothing, and the build previously died inside the plugin source on `expected primary-expression before ';'`.

Also worth knowing for a from-source CasADi: `CXX11_ABI=1` (the default `0` matches the pip wheels' pre-C++11 libstdc++ string ABI), and `make abi-flags` prints the `-D` set the CasADi you are pointing at was actually built with. CasADi's loader does no version handshake, so an ABI mismatch surfaces as an undefined symbol at load time rather than a clear error.

`make install` targets the Python package directory and so does not apply — put the plugin on `CASADIPATH` or beside the CasADi runtime, and make `libpounce_cinterface` visible to the loader.

Full recipe with failure signatures: [`casadi/README.md`](https://github.com/jkitchin/pounce/blob/claude/issue-634-yxwa95/casadi/README.md#against-a-casadi-you-built-yourself-with-no-python).

Two caveats on scope, so nothing is over-promised. This path is verified by hand (against casadi 3.7.2, Linux x86-64), not in CI — CI builds and tests the pip path, since `make test` drives the plugin through CasADi's Python bindings. And there is still no CMake package: a `find_package`-style build would be a second build system that has to stay ABI-identical to the Makefile, which is a real ongoing cost. If the documented path proves awkward in your native CI, say what breaks and that trade-off is worth revisiting.

---

PR #638, opened 2026-08-17
Refuse a dual seed that cannot belong to its primal point (gh #617, gh #618)

### LINKED ISSUE #617: Warm start: reject corrupted dual seeds instead of reconstructing off them

### LINKED ISSUE #618: Warm start: mildly stale seeds regress — the μ escalation band has a hole in the middle

jkitchin:
Fixes #617
Fixes #618

Two #606 follow-ups in the same function region of
`crates/pounce-algorithm/src/init/warm_start.rs`, measured on the same
corpus in overlapping seed modes. Dispatching them separately would have
guaranteed a conflict and a re-measure, so they are one branch and one PR.

All numbers in this body were taken on the parent commit
**`b4c4d32e`** (`Merge pull request #633`) and on this branch, on the same
machine, in the same session. `#630` and `#633` both merged to `main`
while this work was in flight; the branch was merged forward onto each
and every measurement re-taken afterwards.

## Problem

Both issues' numbers are stale — they were filed against
`70bf53ded3d893cfa2da6ead5195fda5ac096f68`, and `main` has since taken
#613, #619, #614, #623, #620, #628, #627, #629, #626, #630 and #633.
Both were reproduced on current `main` before anything was designed,
using the attribution control both issues name: `warm_start_recentering=none`
restores pre-#606 behaviour exactly, so #606 is isolated at a single
commit rather than across `main`'s drift.

Neither regression has gone away. Both are **larger** than filed.

`benchmarks/warmstart`, 14 default-tier families, `large` scale (the 4x
path both issues specify), summed iterations over the corpus:

| seed | pre-#606 (`none`) | parent `b4c4d32e` | this branch |
|---|--:|--:|--:|
| exact restart, full duals | 13 | 9 | **9** |
| exact restart, `lagrange` only | 24 | 11 | **11** |
| exact restart, primal only | 24 | 24 | **24** |
| **corrupted duals, full** | 33 | **164** | **25** |
| **corrupted duals, `lagrange` only** | 27 | **115** | **20** |
| corrupted duals, primal only | 24 | 24 | **24** |
| stale seed, full | 344 | 351 | **354** |
| **stale seed, `lagrange` only** | 273 | **339** | **279** |
| stale seed, primal only | 255 | 255 | **255** |
| cold (no seed) | 166 | 166 | 166 |
| **total, warm cells** | **1017** | **1292** | **1001** |

### #617 as filed vs. as reproduced

| | filed (9 families, `70bf53d`) | reproduced (14 families, `b4c4d32e`) |
|---|--:|--:|
| before #606 | 18 | 33 |
| after #606 | 71 | **164** |
| cold (no seed) | 81 | 166 |

Worst single family as filed was `hanging_chain` 3 → 13; reproduced it is
`hanging_chain` 4 → 17, with `simplex_proj` 2 → 19 and `mpc_horizon_80`
3 → 23 worse still. Status is `SolveSucceeded` throughout, at the same
objective — a slowdown, not a correctness failure, exactly as filed.

### #618 as filed vs. as reproduced

| | filed (9 families, `70bf53d`) | reproduced (14 families, `b4c4d32e`) |
|---|--:|--:|
| before #606 | 118 | 344 |
| after #606 | 127 | **351** |

**The per-family regressions #618 names have all gone.** On current
`main`, `redundant_rows` is 4 → 4, `degenerate_corner` 9 → 9,
`moving_bound_qp` 29 → 29, and `hanging_chain` 18 → **14** — it now
*improves*. The +7 that remains is a different set of families, and it is
dominated by one: `degenerate_vertex` 2 → 12.

The larger instance of #618's phenomenon is in a cell #618 did not
measure — a `lagrange`-only stale seed, +66 (273 → 339), concentrated in
`mpc_horizon_40` (45 → 81) and `mpc_horizon_80` (53 → 77). This is
#620's single-block fix reaching models with one constraint block, the
same mechanism as the `nmpc_vanderpol @ large` 195 → 209 datapoint added
to #618 in comment. That family is in the corpus above and its stale
`lagrange`-only seed moves 42 → 36 on the parent (a #606 *win*).

### #617 — the escalation, not the reconstruction

#617 attributes the regression to `v_L`/`v_U` being reconstructed off a
corrupted `y`. The diagnostics say otherwise, and say it unanimously:
**every** corrupted regression in the corpus ends at `mu_out = 0.1`.

The corrupted multipliers read back as an average complementarity of
`2e-1` to `4.4e+2`. `final_mu`'s escalation takes that at face value and
moves μ eight orders — from the converged barrier to `MU_CEILING` —
while `x` stays at the exact solution. That is the most off-centre a
point can be for the barrier it has just been given: the slacks are
sized for `μ ≈ 2.5e-9` and the barrier now claims `0.1`. Hence the
result costing about what a cold solve costs, and on `simplex_proj`
more (19 against a cold 11).

On families whose only bound multipliers are seedable variable bounds
(`mpc_horizon_*`, `nmpc_vanderpol` at a full seed) *nothing at all* is
reconstructed — `bound_duals: accepted`, `bound_duals_reconstructed: 0`
— and they still regress 3 → 23. The reconstruction is not the
mechanism there; the barrier move is the whole of it.

### #618 — the reconstruction reads the seed's active set

Both halves of the bound reconstruction treat a small slack as "this
bound is active": `μ̂ / slack` sizes the multiplier from it, and the
stationarity split decides which bound carries the residual from it. On a
seed carried across a parameter move that is a claim about a bound that
may since have released. On `mpc_horizon_40`, `nrec = 244` — every
variable bound — with `mu_in = mu_out = 2.51e-9`: μ never moves, and the
entire 45 → 81 is the reconstruction rerouting the trajectory.

That is #618's diagnosis exactly ("stale enough that dual reconstruction
fires and reroutes the trajectory, but not stale enough to trip the
escalation that is supposed to pay for it") — but located in the
reconstruction rather than in μ.

## Fix

Three guards for #617 and one for #618, all at the same 10x conservatism
`MU_ESCALATION_TRIGGER` already uses, because refusing a seed reroutes
the trajectory exactly as much as trusting a bad one does.

1. **A seeded bound-multiplier block is refused when its implied
   complementarity cannot belong to its primal point.** Measured on
   `|z_i| · s_i` over the entries the caller actually seeded — the
   quantity the barrier *is*, and the machinery #617 correctly says
   already exists — and refused when that reads ten times above **both**
   `mu_init` and the point's own `inf_pr`. `inf_pr` is in that max
   because a point that misses feasibility by that much is *allowed*
   complementarity of that order; it is what keeps the strongly-stale
   case out of the test. A refused block takes the pre-#606 constant
   fill and stops being an input to anything, which drops `avrg_compl`
   back to the barrier and takes the escalation with it — no separate μ
   rule was needed. `|z_i|` and not `z_i`: a corrupted block is roughly
   half negative and a signed average cancels to nothing.
2. **A seeded `y` that does not belong to the point stops feeding the
   split.** Refused when its stationarity residual dwarfs `∇f` and the
   bound multipliers *the caller supplied*. The block itself is left
   where the caller put it — that is what the pre-#606 path does with a
   supplied `y`, and there is no constant to fall back to — but nothing
   is derived from it.
3. **The split may not raise a multiplier by orders.** It may raise it
   above the complementarity floor — the slack push's inflation put
   #606's reconstructed HS071 multiplier 5.5x low, so the margin is real
   — but it is now capped at ten times that floor.
4. **(#618) A slack the point's own primal residual swamps is not a
   measurement.** Those entries keep the pre-#606 constant. Per-entry,
   not a per-point gate, so a partly-stale seed keeps its reconstruction
   wherever the slacks still outrun the residual — the reconstruction's
   reach scales down with the measurement instead of switching off at a
   threshold, which is the second remedy #618 offers.

* **#618's first proposed remedy — a gentler μ escalation band below the
  10x trigger — is inert, and was measured rather than assumed.** Every
  stale regression remaining on current `main` has a measured
  complementarity within **1.2x to 3x** of `mu_init`
  (`degenerate_vertex` 2.986e-9 against 2.506e-9 — a 1.19x overshoot;
  `mpc_horizon_40` 7.54e-9 against 2.51e-9). A band starting at 1x has
  nothing to raise μ *to*.
  Separately, it cannot be built without moving `cresc4`, which sits at
  5x — the band would have to exclude the very ratio that motivates it.
* **Scaling the complementarity fill by `μ̂ / max(slack, inf_pr)`**
  instead of falling back to the constant: measured, and worth −1 on
  stale/full and +11 on stale/`lagrange`-only. Rejected; the constant
  fallback is both simpler and better.
* **Comparing slacks against `inf_pr` unguarded**: fires on *converged*
  points, whose `inf_pr` sits at solver tolerance and routinely exceeds
  the slacks `warm_start_slack_bound_push` has shoved to 1e-9. It called
  an exact restart's active bounds swamped and cost the exact partial
  restart 11 → 15. Guarded by requiring `inf_pr` to clear the barrier
  tenfold first.
* **Scaling the `y` test by all four bound-multiplier blocks** including
  the ones this same pass reconstructed: a `μ̂ / slack` fill at a tight
  bound is large by construction, and on `nmpc_vanderpol` it was
  vouching for the very seed it had been derived from — 12 iterations
  against a cold solve's 12. The scale now uses only caller-supplied
  multipliers.
* **A sign test** (`z_i < 0` proves corruption) separates this corpus
  perfectly — 0.00 negative fraction on every exact and stale seed,
  0.48–0.88 on every corrupted one. Rejected as fitting the fixture:
  the corruption model here is symmetric Gaussian, and an all-positive
  corruption would defeat it while leaving the magnitude problem intact.

### A caveat stated rather than hidden

Guard 4 compares a slack against `inf_pr`. For the slack-bound blocks
those are the same units. For *variable*-bound blocks they are not — the
link from constraint-space error to variable-space position error goes
through the Jacobian, so this is an order-of-magnitude argument with
`‖J‖ ~ 1` assumed, in the same spirit as the rest of the module. It is
the case guard 4 helps most (`mpc_horizon_*`), and it is worth knowing
that the reasoning there is scale-approximate.

### #617 — every family, at or below pre-#606

Corrupted seed, full duals, per family (pre-#606 / parent / branch):

| family | pre-#606 | parent | branch |
|---|--:|--:|--:|
| `simplex_proj` | 2 | 19 | **2** |
| `moving_bound_qp` | 3 | 12 | **3** |
| `degenerate_corner` | 2 | 12 | **2** |
| `redundant_rows` | 3 | 7 | **2** |
| `degenerate_vertex` | 1 | 5 | **1** |
| `hanging_chain` | 4 | 17 | **3** |
| `rosenbrock_ring` | 2 | 5 | **1** |
| `rosenbrock_ring_cycle` | 2 | 5 | **1** |
| `double_well_chain` | 0 | 0 | 0 |
| `nmpc_vanderpol` | 3 | 14 | **2** |
| `mpc_horizon_10` | 2 | 14 | **2** |
| `mpc_horizon_20` | 3 | 18 | **2** |
| `mpc_horizon_40` | 3 | 13 | **2** |
| `mpc_horizon_80` | 3 | 23 | **2** |
| **total** | **33** | **164** | **25** |

Not one family is left above its pre-#606 number, and eight are below
it — the refused seed still leaves the exact primal point, which the
pre-#606 constants also had but paired with a less useful `z`.

On HS071 with the same corruption: **9 → 3** iterations (cold 8,
pre-#606 3), `mu_out` back at `mu_in`. With `lagrange` corrupted and the
bound multipliers left out: **7 → 3**.

### #618 — the reconstruction cost recovered

Stale seed, `lagrange` only, the families that move:

| family | pre-#606 | parent | branch |
|---|--:|--:|--:|
| `mpc_horizon_40` | 45 | 81 | **45** |
| `mpc_horizon_80` | 53 | 77 | **53** |
| `mpc_horizon_20` | 23 | 30 | **23** |
| `mpc_horizon_10` | 17 | 24 | **17** |
| `degenerate_corner` | 10 | 12 | **6** |
| `degenerate_vertex` | 2 | 12 | 12 |
| `rosenbrock_ring` | 17 | 18 | **17** |
| `simplex_proj` | 11 | 10 | 11 |
| `moving_bound_qp` | 22 | 12 | 22 |
| `hanging_chain` | 18 | 14 | 18 |
| `nmpc_vanderpol` | 42 | 36 | 42 |
| **total** | **273** | **339** | **279** |

The +66 regression becomes +6. `degenerate_corner` lands 4 *below*
pre-#606. The cost is #606's stale-seed wins on `moving_bound_qp`,
`hanging_chain` and `nmpc_vanderpol` (−10, −4, −6), given back to buy
−74 on the `mpc_horizon` column.

`stale`/full moves 351 → 354, i.e. from +7 to +10 over pre-#606: the
same `hanging_chain` win (14 → 18) is given back there too, and the
`rosenbrock_ring` +1 is recovered.

### What is not fixed — `degenerate_vertex`

One family, 2 → 12, in both stale seed modes, unchanged by this branch.
Its diagnostics:

```
inf_pr = 0.0        avrg_compl = 2.986e-09  mu_in = mu_out = 2.506e-09
inf_du = 5.76       bound_duals = reconstructed (12)   split = true
```

θ enters this family through the objective, so a stale `x` is still
exactly feasible: `inf_pr` is **0**, not small. It is centred at exactly
the barrier it asked for, μ never moves, and only its `y` has moved.
Guard 1 cannot see it (complementarity is *right*); guard 4 cannot see
it (`inf_pr` is zero, so no slack can be swamped); guard 2 does not fire
(the stationarity residual is a small multiple of the scale it is
measured against, nowhere near 10x); a gentler μ band cannot see it
(1.19x overshoot). It is #618's
"hole in the middle" in its purest form. No measurement available to the
initializer separates it from a good seed, and lowering any trigger far
enough to catch it is precisely the un-conservative move both issues
warn against — #617 says "similarly conservative", #618 says `cresc4`
"has to be left bit-identical".

### Cross-check at a second scale

Same corpus at `small` (1.0x rather than 4.0x), against the pre-#606
control: exact 7/8 (control 10/22), corrupted 25/18 (control 32/28),
stale 204/193 (control 201/189). Same shape — corrupted below pre-#606,
exact well below, stale within a few.

## `cresc4` — the canary

**Bit-identical, 85 iterations, `obj = 0.87189753`,** on parent and
branch, in the tightened-push regime where #606 measured it. No
justification needed because nothing moved.

For the record, the attribution control at the parent shows what #606
itself still moves on the CLI corpus in that regime — and this branch
leaves all three exactly where #606 put them:

| fixture | `recentering=none` | `recentering=residual` (parent = branch) |
|---|--:|--:|
| `hs13_bigstart` | 30 | 29 |
| `jit1_boxed` | 16 | 17 |
| `jit1_node` | 19 | 21 |
| `cresc4` | 85 | 85 |

## Blast radius

`scripts/sweep-fixtures.sh`, 57 fixtures, against a release binary built
from the parent commit `b4c4d32e` in a `git worktree`, in all three
regimes #606 established reach this code:

| regime | result |
|---|---|
| default options (`warm_start_init_point` defaults to `no`) | **bit-identical**, 57/57 |
| `warm_start_init_point=yes`, default pushes | **bit-identical**, 57/57 |
| tightened recipe (`mu_init=1e-7`, all five `warm_start_*_push`/`_frac` at `1e-9`) | **bit-identical**, 57/57 |

Zero moving lines, so there is nothing to explain. The middle regime is
load-bearing evidence that the initializer actually runs: it differs from
the cold sweep on **17 of 57** fixtures, so a bit-identical result there
is a real negative, not a no-op.

The reason the corpus does not move is that the guards fire only on
seeds it does not contain. The `.nl` fixtures carry either no dual
segment or a coherent one, and none is a stale replay.

`warm_start_recentering=none` — the kill switch — is **bit-identical to
the parent across the entire seed-mode corpus**, every family, every
staleness, every content mode. Nothing here escapes the switch.

Beyond the solver: `info["warm_start"]` gains `bound_duals_rejected`
(int) and `eq_duals_rejected` (bool), `bound_duals` gains a `rejected`
verdict string, and the `print_level=5` iteration line gains `wz!` and
`wy!`. All additive.

## Tests

`crates/pounce-algorithm/tests/warm_start_recentering.rs`, +6 tests.

**Behavioural, and verified failing on the parent for the right reason.**
Both were lifted into a tree at `b4c4d32e` and run there:

* `a_corrupted_dual_seed_does_not_escalate_mu_to_the_ceiling` — parent
  output: `residual=9 (SolveSucceeded) none=3 cold=8 mu 2.5059e-9 -> 1e-1`.
  Fails on `mu_out <= mu_in`. The warm solve costs *more than a cold
  one* there (9 against 8). Branch: `residual=3 ... mu 2.5059e-9 ->
  2.5059e-9`.
* `a_corrupted_lagrange_seed_is_not_split_into_bound_multipliers` —
  parent output: `residual=7 none=3 mu 2.5059e-9 -> 1e-1 compl=9.13e+1
  split=true`. Same failure. Branch: `residual=3 ... split=false`.

**API-shaped, and stated as such.** These reference `BlockVerdict::Rejected`
and the two new counters, so they cannot be compiled against the parent
at all. They pin the diagnostics, not a behaviour, and are not counted
as a behavioural bite:

* `a_refused_dual_seed_says_so_in_the_diagnostics` — also pins that the
  verdicts are per block: HS071 refuses the two seeded variable-bound
  blocks *and* still reconstructs the slack-bound blocks no frontend can
  seed, and both counters stay honest about it.
* `an_exact_restart_is_unchanged_by_the_seed_guards` — nothing refused,
  reconstruction and split both still run, still beats the constants.
* `a_strongly_stale_point_still_gets_its_looser_barrier` — the case #618
  cites as "both ends of the range are fine". Still escalates, still is
  not mistaken for a corrupted seed. Live output: `residual=37
  (SolveSucceeded) none=36 (RestorationFailed)` — the case that *fails*
  before #606 still passes.

Full suites, on this branch:

* `cargo fmt --all -- --check` — clean.
* `cargo clippy --workspace --exclude pounce-hsl --all-targets` — no new
  warning. `init/warm_start.rs` emits the same 7 warnings (6
  `unwrap_used`, 1 collapsible `if`) on this branch as on the parent,
  verified by running clippy in both trees and diffing. The
  `too_many_arguments` this work first introduced was fixed by grouping
  the three measurements into `SeedMeasurement`, not by an `allow`.
* `cargo test --workspace --exclude pounce-hsl --release` — **231 test
  binaries, all ok, 0 failures**.
* `python/tests` — **1082 passed, 23 skipped**. One failure before `jax`
  was installed (`test_step_controller_is_what_the_jax_follower_uses`,
  a bare `ImportError` from `pounce.jax` with no skip guard) — an
  environment gap, unrelated, and it passes with `jax` present. `torch`
  is not installed and its modules skip.
* `pyomo-pounce/tests` — **382 passed** (`pyomo 6.10.1`, `networkx`,
  `scipy`, `pandas`, CLI staged into `python/pounce/bin/`).
* `scripts/check-docs-consistency.sh` — OK, 46 pages.
* `scripts/check-release-consistency.sh` — OK.

## Conflict note

Both siblings **merged into `main` during this work**, so these are
resolved conflicts, not predictions.

* **#630** (`claude/pounce-issue-610-loop`) merged as `7d406cfd`. Two
  overlaps, both auto-merged cleanly:
  - `CHANGELOG.md`, top of `[Unreleased]` — its `race_starts` entry
    versus this branch's warm-start entry. Resolved with this entry
    first and #630's existing order (`Pyomo` → `race_starts` → `Bool`)
    preserved beneath, unchanged.
  - `crates/pounce-py/src/problem.rs`, `build_info_dict` — #630 replaced
    twelve scalar parameters with `stats: &SolveStatistics` and dropped
    the `#[allow(clippy::too_many_arguments)]`; this branch adds two
    `set_item` calls to the `warm_start` sub-dict and one arm to
    `verdict_name`. Different hunks of the same function; no textual
    conflict. Verified by rebuilding and re-running the full seed-mode
    corpus after the merge — identical in every cell.
  - `docs/src/initialization.md` — #630 added `### Racing starts: the
    successive-halving ladder` at the end of the page; this branch adds
    `### A seed the solver will not believe` inside the warm-start
    section. No overlap.
* **#633** (`claude/pounce-issue-616-loop`) merged as `b4c4d32e`. Also
  auto-merged cleanly:
  - `CHANGELOG.md` — its least-square-init entry slotted below, ordering
    preserved.
  - `docs/src/initialization.md` — #633 added `### What the safeguard
    costs, and why it is not tuned away` and `### A declined step is not
    the same as never asking` to the **cold-start** half (lines 152 and
    242); this branch's section is at line 385 in the **warm-start**
    half. Disjoint.
  - `crates/pounce-algorithm/src/init/default.rs` — #633's only solver
    edit, and it is a pure refactor (`accepts_trial` extracted with
    identical logic) plus a `tracing::debug!` line. No overlap with
    `init/warm_start.rs`. Confirmed non-interacting by re-running the
    three-regime fixture sweep against a baseline built at `b4c4d32e`
    after the merge: still bit-identical.

A third merge-up, onto `eeca47f9` (#635, the `pounce-casadi` wheel
platform tag), was taken when this PR was opened. It touches only
`casadi/`, `.github/workflows/ci.yml` and `CHANGELOG.md`; the only
conflict was the `[Unreleased]` head, resolved with `main`'s entry
first. No solver code moved, so the measurements above stand unchanged.

## The decisions, and what each cost

**#617: refuse the seed.** Take the remedy the issue asks for, but
locate it where the measurement says the damage is — the μ escalation
reading a corrupted `z`, not the reconstruction reading a corrupted `y`
— and add the split cap the second reading exposed. **Cost: nothing
measurable.** Exact restarts are bit-identical to #606's, the fixture
corpus is bit-identical, the kill switch is bit-identical, and every
corrupted family lands at or below pre-#606. The only thing spent is a
trigger constant that a future retuner has to keep conservative.

**#618: scale the reconstruction, don't touch μ.** Take the issue's
*second* proposed remedy after measuring the first and finding it inert.
**Cost: three of #606's stale-seed wins** — `moving_bound_qp` 12 → 22,
`nmpc_vanderpol` 36 → 42, `hanging_chain` 14 → 18, and the same
`hanging_chain` win again in the full-seed column. That is 20 iterations
of #606's gains given back in the `lagrange`-only column, and 4 more in
the full column, traded for 74 of its losses — plus a
`degenerate_corner` improvement of 4 that neither issue predicted.
`stale`/full ends 3 worse than the parent (351 → 354) while
`stale`/`lagrange`-only ends 60 better (339 → 279).

**Left undone, deliberately:** `degenerate_vertex`, 2 → 12. Nothing
available to the initializer distinguishes it from a good seed, and
inventing a trigger that catches it is the move both issues were filed
to prevent.

---

Verified rather than taken on report:

* The two behavioural tests were re-run against the parent's
  `warm_start.rs` in this tree. Both fail there, with the claimed cause:
  `mu_in=2.5059035596800634e-9 mu_out=1e-1`. Parent iteration counts
  reproduce exactly — `residual=9 none=3 cold=8 compl=7.44e1` for the
  full corrupted seed, and `residual=7 none=3 compl=9.13e1 split=true`
  for the corrupted `lagrange`. Both branch numbers (3 and 3) reproduce
  too.
* `cargo test --release -p pounce-algorithm` — exit 0, 50 suites, zero
  failures, including the 15 in `warm_start_recentering`.
* `cargo fmt --all -- --check` — clean.
* Two stale figures in the changelog entry were corrected in
  `354670fa`: it named `66cc1d41` (the pre-merge commit) as the parent
  the numbers were taken on, where the numbers themselves are the
  post-merge ones and match this body cell for cell; and it described
  `degenerate_vertex` as `inf_pr = 8.4e-9` where the re-taken
  diagnostics read 0. The extension module in this tree predates the
  guards, so rather than restate a figure that could not be re-measured
  here, the entry now states the property both runs agree on.

---


jkitchin:
Reviewed. This is ready to merge once #636 lands ahead of it — see §4 for why the order matters and what to re-check after the merge-forward. Two substantive notes below, one of which I could not settle by measurement and am flagging as analysis.

## What I checked rather than took from the body

**The split cap is arithmetically right, including on a negative target.** `target[k].min(SEED_REJECTION_TRIGGER * compl_floor)` is followed by `.max(compl_floor).max(hard_floor).min(cap)`, so a `target[k]` that comes out negative — which the sign split can produce — is raised back to the complementarity floor rather than clamped to a negative multiplier. The net effect is `vals[k] ∈ [max(compl_floor, hard_floor), min(10·compl_floor, cap)]`. That is the interval the comment describes.

**`seeded_z_amax` really does exclude what this pass built.** This is the load-bearing part of guard 2 and the thing the `nmpc_vanderpol` 2 → 12 measurement is about, so I traced it rather than trusting the doc comment. A rejected block `continue`s before the accumulation loop is reached, and inside that loop the accumulation sits on the `!(v.is_nan() || *v == 0.0)` arm — the *unseeded* arm is the one that falls through to the `μ̂ / slack` fill. So the scale is caller-supplied multipliers from surviving blocks only, exactly as documented. Good.

**The `swamping_residual` guard is what makes guard 4 safe on a converged point.** Requiring `inf_pr > 10·μ̂` before `inf_pr` is allowed to swamp anything is the whole reason an exact restart's `warm_start_slack_bound_push`-shoved 1e-9 slacks are not called swamped. Removing that guard is the 11 → 15 the body reports, and reading the code that is exactly what would happen. The per-entry (not per-point) shape is right too, and worth the paragraph it gets.

**Local runs (macOS arm64):** `cargo test --workspace --exclude pounce-hsl` → 231 binaries, **0 failures**, exit 0. `cargo fmt --all -- --check` clean.

## 1. Guard 2's scale assumes ‖z‖ ≲ ‖∇f‖, and that is a property of the model

This is **derived, not measured** — I am flagging the reasoning, not reporting a reproduction.

`eq_seed_is_incoherent` rejects when `‖r_x‖∞ > 10·max(‖∇f‖∞, seeded_z_amax, μ̂)`. On a **`lagrange`-only** seed `seeded_z_amax` is zero by construction, so the scale collapses to `‖∇f‖∞`. But at a stationary point `r_x = ∇f + Jᵀy` *is* the true bound-multiplier vector, so the test is really

> reject when `‖z_true‖∞ > 10‖∇f‖∞`

and with the basis/non-basis split `z_N = ∇f_N − J_Nᵀ J_B^{-ᵀ} ∇f_B`, that is a statement about `‖J_Nᵀ J_B^{-ᵀ}‖` — the conditioning of the basis at the seeded point — not about whether the `y` is any good. A perfectly correct `lagrange`-only seed on a model whose basis is poorly conditioned relative to its non-basic columns is refused.

Three things keep this from being a blocker, and I want to be as clear about them as about the gap:

* **The consequence is bounded.** The `y` is left where the caller put it, the split does not run, and the `μ̂ / slack` fill stands — i.e. #606-without-the-split, still ahead of pre-#606 in most cells. No wrong answer, no escalation.
* **Default `nlp_scaling_method=gradient-based` normally keeps the ratio near one**, which is presumably why 14 families across three staleness modes never trip it, and why `exact restart, lagrange only` holds at 11 on both parent and branch.
* **The obvious "fix" is worse.** Putting `‖Jᵀy‖∞` into the scale defeats the test entirely — for a corrupted `y`, `r_x ≈ Jᵀy`, so the ratio goes to one and nothing is ever refused. The body already reports measuring a version of this on `nmpc_vanderpol`. If this is ever revisited, the better-founded test is probably *cancellation* (`‖∇f + Jᵀy‖ ≪ max(‖∇f‖, ‖Jᵀy‖)` holds for a real `y` and fails for a corrupted one) rather than absolute magnitude — but that is a redesign that would invalidate every number in this body, and I am not asking for it now.

What I would ask for is one sentence in `eq_seed_is_incoherent`'s doc comment recording the assumption, in the same voice as the caveat already written for guard 4. The module is unusually good at naming what its order-of-magnitude arguments rest on, and this one is currently unnamed.

## 2. `degenerate_vertex` 2 → 12, and closing #618

Leaving it is the right call and the reasoning is sound — I checked each of the four "cannot see it" claims against the guards as written and they all hold (`inf_pr = 0` really does make `swamping_residual` return 0, so `slack_is_swamped` is false for every slack; a 1.19× overshoot really is below any trigger in the file). Inventing a trigger that catches it is the move both issues were filed to prevent.

The thing I want to flag is the bookkeeping, not the decision. `Fixes #618` closes the issue while its named phenomenon survives in one family. What makes that acceptable here is that the residue is written into `CHANGELOG.md` under *What is not fixed*, with the mechanism and the diagnostics — so it is a recorded limitation rather than a lost one. That is the standard CLAUDE.md sets after gh#544, and this clears it.

One gap that is **not** this PR's to close but is worth stating out loud, because both this PR and #636 lean on it: `grep -rn warmstart .github/workflows/ Makefile` returns nothing. Nothing automated runs `benchmarks/warmstart`, so the corpus that priced every number above — and that would notice `degenerate_vertex` moving from 12 — only runs when someone runs it. #640 adds `make -C benchmarks warmstart-all` but does not wire it into CI either.

## 3. Smaller things, all fine

* `BlockVerdict::Rejected` winning the block summary over `Reconstructed` is the right precedence, and `verdict_name` + the two new `info["warm_start"]` keys are additive with no strict key-set assertion anywhere to break.
* `wz!` / `wy!` in the `print_level=5` line read unambiguously against the existing `wz` / `wy` / `wy0` / `wmu`.
* `docs/src/initialization.md:431` documents both new keys with a worked `info` dict. Good.
* `seed_is_incoherent` returning `true` on a non-finite seed entry, and `false` when `n == 0`, are both the right defaults.

## 4. Merge order and the one thing to re-check afterwards

**#636 must land first** — it rewrites `recenter_from_residuals`'s signature and the `dual_info` gate in the same region this PR touches, and the two conflict textually.

After merging `main` forward, one interaction needs re-measuring rather than re-reading. #636 adds a `bound_fill_only` path that calls `recenter_from_residuals` for a caller who supplied **no duals at all** — a path that did not exist when this branch's numbers were taken. That path now runs this PR's `slack_is_swamped` fill. A transferred bare point can carry real primal infeasibility, so `swamping_residual` can be non-zero there and entries that #636 wants filled at `μ̂ / slack` would take `mult_bound_push` instead — which is precisely the constant #636 exists to stop those callers being given.

Concretely: after the merge-forward, re-run

```
cargo test -p pounce-algorithm --test warm_start_recentering -- --nocapture
```

and confirm the values-only line still reads `cold=6 switch+tight=8` (on aarch64 macOS; 7 on x86-64 Linux). #636's `it_tight <= it_cold + 2` assertion sits exactly on its boundary, so this interaction would show up as a red test rather than as drift — but it is worth looking at the printed numbers, not just the green.

Merging after #636.

jkitchin:
## Merged `main` forward — and one of the three conflicts was a real defect

I said in the review above that after the merge-forward I'd re-run
`warm_start_recentering` and confirm the values-only line still read
`cold=6 switch+tight=8`. It did not. It read:

```
cold=6  switch+tight=11  switch+default=7
```

That's a failure of #622's own governing assertion — "with no multipliers
seeded there is nothing for the multiplier push to push, so tightening it
must not reroute the solve" — and it appears only in the merge. Both
branches are individually green.

### What happens

#622 added a `bound_fill_only` path so that a seed carrying only a point
still gets its bound blocks: `z = mu / slack` needs no dual to derive it.
This PR's `slack_is_swamped` then ran on that path too. On #622's own HS071
transfer fixture, `inf_pr = 1.576` against `mu_hat = 1e-1`, so the swamping
threshold is 1.576 — and every slack (5.0e-2, 3.5e-1, 4.9e-1, 1.10, 2.5e-8)
is under it. All four blocks fall to the constant fill `opts.mult_bound_push`,
which under the tightened pushes is 1e-9, i.e. "this bound is inactive" for
every bound in the problem. That is exactly the push sensitivity #622 exists
to remove, and worse than where it started.

### The fix, and why it isn't a threshold

The guard is now armed only for a seed that carried multipliers
(`seed_implies_barrier`). Its question is whether the point's infeasibility
outruns *the barrier the seed implies* — and a values-only seed implies
none. The `mu_hat` it would be measured against on that path is
`safe_mu(curr_mu)`, i.e. the caller's `mu_init`: an option, not a
measurement of the seed. Comparing against it measures nothing.

No threshold fixes this, because the push-dependence enters through the
guard's *input*: `slack_is_swamped` reads a slack, and the warm-start pushes
move slacks. Anywhere the guard is armed, the push can change which entries
fire. Two alternative fallbacks were measured before settling on the gate:

| variant | tight | default | verdict |
|---|--:|--:|---|
| armed, fall back to `mult_bound_push` (naive merge) | 11 | 7 | push-sensitive |
| armed, fall back to `bound_mult_init_val` | 10 | 10 | push-insensitive, but over #622's `cold + 2` ceiling of 8; also took the #617 corrupted seed from `residual=3` to `residual=8` |
| armed, fall back to `mu_hat / swamping` | 8 | 7 | still splits — the *partition* moved with the push even though the fill did not |
| **gated (shipped)** | **8** | **8** | matches #622 |

The third row is the one that settles it: making the fill itself
push-independent is not enough.

### What the gate costs

Measured with #638's own `seedmodes` harness, 14 families at 4x scale,
`--contents primal,cold`, all three staleness modes, against the pre-#617
`main`:

- **gated (shipped) vs main: 0 differing cells out of 84.** Iteration counts
  and statuses both.
- armed everywhere vs main: 7 differing cells; stale primal-only seeds total
  262 against the gated 334.

So the guard, if armed on the primal-only path, would be a real improvement
on stale seeds — five families better, two worse — and taking it would mean
giving up #622's push-insensitivity. This is an improvement declined, not a
cost incurred: nothing regresses against `main`. Both numbers and the seven
family names are in the CHANGELOG so the next reader doesn't have to
rediscover the trade.

### Conflict resolution, for the record

The other two were textual. `BlockVerdict::Unseeded` keeps #622's newer
doc comment alongside this PR's new `Rejected` variant. The test file is
main's 1085 lines plus this branch's 241 appended — git split it
mid-signature because both sides append after the same line and the two new
helpers share a trailing fragment; verified by checking that HEAD's first
933 lines are byte-identical to the merge base.

### Gates

- `warm_start_recentering`: 16 passed, 0 failed. Values-only line reads
  `cold=6 switch+tight=8 switch+default=8 switch+tight, NaN-seeded=8`.
  Every #617/#618 number unchanged: exact restart `residual=1`, corrupted
  seed `residual=3` with `mu_out = mu_in`, corrupted `lagrange` `residual=3`,
  stale `37`.
- `cargo test --workspace --exclude pounce-hsl`: exit 0, 233 binaries ok, 0 failures.
- `cargo fmt --check`: clean.

Merging once CI is green.

PR #639, opened 2026-08-17
Refuse a reordered warm start without the caller naming the ordering (gh #621)

### LINKED ISSUE #621: A reordered warm start is not refused unless the caller supplies var_ids
Split out of #607 (PR to follow). #607's acceptance criteria say "replaying an incompatible artifact fails with a useful mismatch report before entering the solver". The `ProblemSignature` landed there catches changed bounds, sparsity, scaling, algorithm, dimensions and model-defining options — but **not a pure permutation of the variables**, unless the caller supplies `var_ids` on both the capture and the replay.

## Measured

Permuting HS071 (uniform box `[1, 5]^4`, dense jacobian) leaves the bound digest and the sparsity digest bit-identical, so `ws.check_compatible(prob)` reports no mismatch. On the parent commit the reordered replay solved to `Error_In_Step_Computation` with objective **16.3801074993** against the true **17.0140171452** — `x` off by 0.257 — in 44 iterations. With #607 merged and no `var_ids`, that replay is still attempted; it is caught only when both sides carry stable identifiers.

## Why it was left this way in #607

Ordering is knowledge only the caller has: a live `Problem` does not know what its variables are called. Treating a missing target `var_ids` as an unverifiable-facet mismatch would make signing an artifact *with* IDs strictly worse than signing it without, since the strict check would then refuse every replay against a plain `Problem`. That tradeoff was judged wrong for the default path, and the gap was documented rather than closed.

## What would close it

Options, none of them free, all needing measurement before adoption:

- A cheap order-sensitive probe of the model itself — e.g. digesting `grad_f` / `jac_g` evaluated at a fixed deterministic point — which distinguishes a permutation without needing names. Costs one extra evaluation at capture and at replay, and needs a decision about tolerance and about models whose evaluation is not bitwise reproducible.
- Deriving `var_ids` automatically on frontends that *do* have names (Pyomo, and the AMPL/`.nl` path), so the common structured-model case is covered without the user doing anything.
- Leaving it as-is and making the limitation louder — e.g. a one-line note in the mismatch report when a signature carries no `var_ids` on either side.

## Acceptance

A reordered replay is refused, or accepted deliberately, without the user having to know to pass `var_ids` — and whatever it costs is measured on the fixture corpus, not asserted.

Related: #607, #620.

jkitchin:
Fixes #621

`ProblemSignature` gains an order-sensitive **model probe** facet, so a
pure permutation of the variables is refused before the solver is
entered without the caller having to know to pass `var_ids`.

**No Rust changed** — `git diff origin/main -- crates/ Cargo.toml Cargo.lock`
is empty.

## 1. The reproduction, on unmodified `main`

Permuted HS071 (`perm = [2, 0, 3, 1]`, uniform box `[1,5]^4`, dense
jacobian), captured with **no `var_ids`** and replayed on the permuted
model. Every facet of the two signatures came out identical:

| facet | canonical | permuted | |
|---|---|---|---|
| n / m | 4 / 2 | 4 / 2 | SAME |
| var_ids / con_ids | None | None | SAME |
| bounds | `51e5c8cd33c97b92` | `51e5c8cd33c97b92` | SAME |
| sparsity | `77ecf28c2ff3692a` | `77ecf28c2ff3692a` | SAME |
| scaling | `12ac3016134fa70f` | `12ac3016134fa70f` | SAME |
| algorithm | `4a8839771719ec68` | `4a8839771719ec68` | SAME |
| model | `4a8839771719ec68` | `4a8839771719ec68` | SAME |

`check_compatible` returned `[]` — no mismatch — and the replay damaged
the solve:

| | value |
|---|---|
| true objective (cold, either ordering) | **17.0140171452**, 11 iterations, `Solve_Succeeded` |
| reordered warm replay | **16.0909032757**, 44 iterations, `-3 Error_In_Step_Computation` |
| `‖x − x_true‖_∞` (un-permuted) | 0.2205 |

**Difference from the issue's numbers, stated plainly.** The issue
reports objective 16.3801074993 and `x` off by 0.257, measured on the
pre-#607 parent. This is 16.0909032757 and 0.2205 on current `main` —
same status, same iteration count, same character of failure, but these
are measured numbers on this commit, not the issue's. The drift is most
likely #606's residual recentering landing in between.

This reproduction is the positive control: the detector was watched
firing on it after the fix, not only on synthetic cases.

## 2. Why the probe, and not the other two options

- **Auto-derived `var_ids` on Pyomo / `.nl`** cannot close the acceptance
  criterion: the reproduction *is* a plain `pounce.Problem` built from a
  `problem_obj`, which has no names to derive. It would fix the
  structured-model case and leave the measured failure standing.
- **Making the limitation louder** does not refuse anything. (Folded in
  anyway — it is nearly free; see §6.)
- **The probe** needs no names, so it covers the raw-`Problem` path where
  the failure actually lives.

### The mechanism

The model is evaluated **once** at a fixed point inside the bounds,
recorded as a small vector of order-weighted projections.

**The probe point** (`_probe_point`) is `centre + (t_i − ½)·min(span, 1)`
with `t_i = 0.25 + 0.5·frac((i+1)·φ)`:

- **Deterministic** — no RNG. `*` and `floor` are exact IEEE-754, so it is
  bit-identical on every platform; nothing routes through libm (`cos`,
  `exp`), which is *not* cross-platform stable.
- **Strictly inside the bounds**, because a model is entitled to be
  undefined outside its own box (`log(x)` with `lb=0`). Infinite bounds
  and fixed variables (`lb == ub`) are handled explicitly and tested.
- **Varies with the index**, deliberately: a point that did not (all-ones)
  is a fixed point of every permutation.
- The span cap keeps a `[-1e18, 1e18]` box from being probed at `1e17`.

**Nothing is digested, and that is the central design decision.** Four
blocks (`objective`, `gradient`, `constraints`, `jacobian`) are each
reduced to 4 order-weighted projections plus an L1 scale and compared at
a **relative tolerance** `PROBE_RTOL = 1e-9`. A hash cannot be compared
approximately, and refusing a valid replay for reproducing itself to 15
digits instead of 17 is a worse failure than the one being fixed.

## 3. Does it fire?

Same reproduction, on this branch:

```
warm start is not compatible with this problem (1 mismatch,
exact-structure replay, schema v2):
  - probe: this problem's model does not evaluate to the same numbers as
    the one the warm start was captured against (a reordering of the
    variables looks exactly like this; so does a different model of the
    same shape)
```

One measured detail justifying four blocks rather than one: on HS071 the
**constraints and jacobian blocks are permutation-invariant** (`prod` /
`dot` are symmetric, jacobian entries are per-variable so the permutation
cancels) — they agreed to the last ulp. Only **objective and gradient**
caught it. A single-block probe would have missed this fixture.

## 4. Reproducibility — measured, not assumed

| test | result |
|---|---|
| **A.** Pure numpy model, 20 repeated captures | **bit-identical** across all 20 |
| **C.** Same maths, internal summation order reversed (what a different BLAS / thread count / SIMD width does) | not bit-identical; worst **4.441e-16** absolute, **4.764e-18** relative → **accepted** |
| **D.** jax 0.10.2, `jit`+`grad`, x64, two freshly compiled models | **bit-identical** → accepted |
| **D.** torch | **not installed — not measured.** Reported as a gap, not a pass. |

**Where is the refusal threshold?** Injected relative jitter, 40 trials each:

| jitter | accepted | verdict |
|---:|---:|---|
| 0 → 1e-9 | 40/40 | accepted |
| 1e-8 | 23/40 | borderline |
| 1e-6 | 0/40 | REFUSED |

Real evaluation noise (≈5e-18 relative, measurement C) sits **nine orders
below** the tolerance. The tolerance sits in a wide gap, which is what
makes `1e-9` defensible rather than tuned.

## 5. Cost — measured, not asserted

Capture side:

| n | nnz | no probe | + probe | **probe cost** | cold solve | **% of solve** |
|---:|---:|---:|---:|---:|---:|---:|
| 4 | 8 | 0.0445 ms | 0.1942 ms | **0.1497 ms** | 12.78 ms | **1.171%** |
| 100 | 200 | 0.0556 ms | 0.2344 ms | **0.1788 ms** | 44.96 ms | **0.398%** |
| 1 000 | 2 000 | 0.1370 ms | 0.4748 ms | **0.3378 ms** | 400.4 ms | **0.084%** |
| 10 000 | 20 000 | 0.8794 ms | 2.6092 ms | **1.7299 ms** | 86 986 ms | **0.002%** |

Artifact growth: **1 640 B**, a fixed 20 floats regardless of problem size.

### The regression the cost measurement caught

`test_starts_racing.py::test_the_ladder_beats_the_pre_610_cost` failed —
real, not a threshold artefact. `_starts.py` re-checks its warm start once
per multistart rung, and on that path the model is the **caller's own
counted callable**, so probing per rung spent four of the user's
evaluations per resume:

| | user-callable evaluations |
|---|---:|
| `origin/main` baseline | **2592** |
| probe enabled on that path | **2920** |
| after the fix (`probe=False` there) | **2592** — objectives bit-identical |

Pinned by a test asserting the ladder never evaluates the model **at the
probe point**, confirmed failing when the probe is reinstated (12
probe-point evaluations).

### Fixture sweep

Baseline binary built from `origin/main` in a separate worktree, all 57
CLI fixtures swept both ways: **empty diff, 57/57, 0 NO_JSON**.

Stated honestly: that is *necessary but not sufficient* here, because no
Rust changed and the sweep drives the Rust CLI, which never touches the
Python warm-start path. The load-bearing trajectory evidence is the
Python-side analog — cold solve, warm replay, warm replay **from file**
(exercising the signature round-trip), and the multistart ladder, at
n = 4 / 10 / 50 / 200, recording status, objective **and iteration
count**. **Empty diff.**

```
hs071/cold    status=0 it=11 obj=17.0140171452     hs071/warm   status=0 it=1 obj=17.0140171420
quad10/cold   status=0 it=10 obj=0.0000000000      quad10/warm  status=0 it=1 obj=0.0000000000
quad50/cold   status=0 it=12 obj=0.0000000000      quad50/warm  status=0 it=1 obj=0.0000000000
quad200/cold  status=0 it=14 obj=0.0000000000      quad200/warm status=0 it=1 obj=0.0000000000
```

## 6. The #607 non-regression

#607 deliberately rejected any design where signing an artifact *with* IDs
becomes strictly worse than signing it without. The probe is symmetric —
computed the same way on both sides — and
`test_signing_with_ids_is_never_worse_than_signing_without` pins it across
five targets (same model, reordered, moved bounds, extra nonzero, changed
scaling). Confirmed to detect a break: making `var_ids` a required facet
(the design #607 rejected) fails it with

```
same model: IDs changed the verdict against a plain Problem
(['var_ids', 'con_ids'] vs []) — signing with IDs must not cost anything
```

**`var_ids` remain the rigorous answer, and the docs say so.** The probe
infers ordering from arithmetic, so a model genuinely symmetric under the
permutation is invisible to it —
`test_ids_still_beat_the_probe_where_the_probe_is_blind` builds exactly
such a model (`sum x²`, `sum x`, `dot(x,x)`), shows the probe accepts the
permutation, and shows the IDs still catch it. Only IDs let `reindex`
**repair** a reordering rather than only refuse it.

## 7. Acceptance criterion

> *A reordered replay is refused, or accepted deliberately, without the
> user having to know to pass `var_ids` — and whatever it costs is
> measured on the fixture corpus, not asserted.*

**Met, with one qualification.**

- *Refused without `var_ids`* — **yes**, for the issue's own fixture and in
  general for any model not permutation-symmetric at the probe point.
- *Or accepted deliberately* — **yes**: `probe=False` at capture,
  `compat="warn"`/`"unsafe"` at replay, `migrate()` to re-sign.
- *Cost measured* — **yes**: §5, including a real regression the
  measurement caught and fixed, an empty 57-fixture sweep, and an empty
  Python-side trajectory diff.
- **Qualification:** the guarantee is "refused unless the model is
  symmetric under the permutation", not "always refused". That residual is
  inherent to inferring ordering from arithmetic instead of names; it is
  tested and documented rather than glossed.

## 8. What is NOT done

- **`torch` is not installed in this container, so measurement D covers
  jax only.** No measurement for torch, and none is claimed. A threaded
  torch backend is the most plausible candidate for evaluation noise above
  1e-9, and the jitter sweep says that would be refused. **This is the
  weakest point in the change.**
- **The probe cannot see a permutation that is a genuine symmetry of the
  model** at the probe point — demonstrated, not hidden. Mostly benign (a
  fully symmetric model's reordering *is* the same model), but a model
  symmetric only *at the probe point* is a real, narrow false-negative
  window. IDs close it.
- **Option 2 (auto-derived `var_ids` on Pyomo / `.nl`) was not
  implemented.** It is an alternative to the probe, not a gap in it, and
  the probe covers the frontends it would have plus the raw-`Problem` path
  it could not. It would still be a real improvement — it enables
  `reindex` (repair) where the probe only refuses.
- **`compat="strict"` gets stricter for artifacts captured on this build.**
  That is the point of the issue, but it is a behaviour change: a replay
  that used to be silently attempted now raises. `compat="warn"` /
  `"unsafe"` / `probe=False` / `migrate()` are the outs, all documented.
- **Forward-compatibility footgun, inherited from #607.** A signature
  written by this build carries a facet older builds do not know, and
  `from_json` **raises** on unknown facets by design. So a new-build
  artifact is unreadable by an old build. That policy is #607's deliberate
  choice and was not changed, but it is called out in the CHANGELOG.
- **The probe also fires whenever the bounds move**, because the probe
  point is derived from the bounds. Harmless (the `bounds` facet already
  refuses those) but it makes the report slightly redundant there.
- **No benchmark-suite run.** The §5 cost numbers are from a timing
  harness at four sizes, not from `make benchmark`.

## Tests

| command | result |
|---|---|
| `pytest python/tests -q -p no:randomly` | **1095 passed, 23 skipped** (517 s) |
| `pytest python/tests/test_warm_start_schema.py -q` | **48 passed** |
| `cargo fmt --all` | clean |
| `bash scripts/check-release-consistency.sh` | **OK** |

**The 23 skips, honestly** — none are passes: 9 × no `torch`, 4 × no
`yaml`, 3 × missing `.nl` fixtures, 2 × no `pyomo`, 1 × no `sympy`, 1 × no
`plotly`, 1 × needs `gamsapi[core]`. `jax` was installed specifically so
measurement D would be real rather than skipped.

`cargo test --workspace` was **not** run — no Rust changed, and the
`crates/` diff against `origin/main` is empty. Flagged rather than implied.

---

**A defect found on review and fixed here, in `afe73810`.**

`transfer()` re-signs the target problem, and it did so with
`probe=True` **unconditionally** — so an artifact captured with
`probe=False` still paid four of the caller's model evaluations on every
`transfer`, for a facet it had explicitly opted out of. The report's own
claim that "an artifact captured with `probe=False` costs nothing extra,
by construction" holds for `check_compatible` but did **not** hold for
`transfer`.

This is the same defect the multistart ladder hit (§5), one call site
over, and it matters more here: `transfer` is the per-step call in a
receding-horizon loop (#622), so the cost was four evaluations *per step*
rather than once. Measured on HS071 with counted callbacks:

| capture | evaluations inside `transfer()` — before | after |
|---|--:|--:|
| `probe=False` | 4 | **0** |
| `probe=True` | 4 | 4 (retained — the mapped result stays protected) |

The fix is the one `_target_signature` already documents: probe the target
only when this state carries a probe to compare against. New test
`test_transfer_does_not_pay_for_a_probe_the_caller_declined` fails without
it with exactly that count (`transfer evaluated the model 4 times for a
probe the caller declined`).

**Verified rather than taken on report:**

* `pytest python/tests/test_warm_start_schema.py python/tests/test_starts_racing.py python/tests/test_warm_start.py` — **84 passed**, including the racing suite the §5 regression came from.
* The new test was confirmed to bite: reverted the fix, it fails; restored, it passes.
* `git diff origin/main -- crates/ Cargo.toml Cargo.lock` is empty, confirming the no-Rust claim.
* These runs set `POUNCE_SKIP_EXT_STALE_CHECK=1`: the compiled extension in the collecting container predates recent `crates/` changes. That is sound for a pure-Python change but means the full `python/tests` count above is the child's, not re-verified here — CI is the gate on it.

**Sibling branch — `claude/pounce-issue-622-vl5h04` (issue #622, another session, unmerged).**

It touches four of this branch's six files: `python/pounce/_warm_start.py`,
`python/tests/test_warm_start_schema.py`, `docs/src/initialization.md`,
`CHANGELOG.md`. A trial merge of the two branches conflicts in
**`CHANGELOG.md` only** (the `[Unreleased]` head, as usual); the other
three auto-merge.

Semantically they are independent but they meet inside `transfer()`: #622
changes *what the mapper fills* (`fill_x="prolong"`, prolongating the
transferred stage instead of zero-filling it), while this branch changes
*whether the target gets probed* when it is re-signed. Neither reads the
other's state. The one interaction worth naming for whoever merges second:
a receding-horizon loop that captures with `probe=True` pays four model
evaluations per `transfer` — which is exactly the cost axis #622 is
measuring on. With the fix above, `probe=False` makes that zero.

---


jkitchin:
Reviewed. Ready to merge after #636 (they both rewrite `_warm_start.py` and `test_warm_start_schema.py`). One measured finding below that bounds how far the new protection reaches — it does not make anything worse than `main`, so it is not blocking, but it belongs in §8 rather than being discovered later.

## Verified locally (macOS arm64)

Run against this branch's source tree with the extension module dropped in, so the pure-Python changes are the ones under test:

* `tests/test_warm_start_schema.py` + `tests/test_starts_racing.py` → **79 passed**.
* Full `python/tests` → **903 passed, 41 skipped, 1 failed**. The failure is `test_continuation.py::test_step_controller_is_what_the_jax_follower_uses`, the bare `ModuleNotFoundError: No module named 'jax'` with no skip guard that #638 documents independently. Not this branch.
* `git diff origin/main...HEAD -- '*.rs' Cargo.toml Cargo.lock` is empty, confirming the "No Rust changed" claim.

I also read the four probe functions rather than reasoning from the body. Three things I specifically wanted to be sure of, all correct:

* **A model that will not evaluate cannot cause a refusal.** `_model_probe` returns `None` on `AttributeError`, on any exception from the callables, and on a non-finite result — and `None` is an unrecorded facet, not a failed one. So a model undefined at an interior point, or one that throws, degrades to today's behaviour instead of blocking a valid replay. This is the single most important property of the design and it is implemented the way the docstring says.
* **`_probe_point` stays strictly inside the box** in all four bound configurations (both / lower-only / upper-only / free), `t ∈ [0.25, 0.75)` keeps it off the bounds, and `0.5*lb + 0.5*ub` rather than `0.5*(lb+ub)` is the right way to avoid overflow on a wide box. The span cap is doing real work.
* **Cost and side effects are already documented**, at `docs/src/initialization.md:482` — one evaluation at capture, `probe=False` named explicitly for "a model whose evaluation is expensive or has side effects". I had this on my list to raise and it is already there; the `_starts.py` regression you caught and pinned is the same concern in its most easily-missed form.

## The finding: the tolerance floor makes the probe blind to a small block

`_probe_agrees` computes `floor = max(scales)` over all four blocks and then compares block `b` at `tol = PROBE_RTOL * max(scales[b], floor)`. Since `floor >= scales[b]` always, that simplifies to **every block being compared at `PROBE_RTOL * (largest block's L1 scale)`** — an absolute tolerance set by the biggest block, not a relative one per block.

The docstring justifies this for an identically-zero gradient, and that reasoning is right. The cost is that a block whose magnitude sits far enough below the largest block stops being able to disagree at all. Measured directly against the shipped `_sketch` / `_probe_agrees`, with a 20 000-entry jacobian (L1 ≈ 1.6e4) and matched-scale objective and gradient blocks:

| objective / gradient scale | doubling the objective detected? |
|---|---|
| 1e-2 | yes |
| 1e-4 | yes |
| **1e-5** | **no** |
| 1e-6 | no |
| 1e-7 | no |

and the same for an outright permutation, which is the thing being defended against:

| gradient scale | **reversing** the gradient block detected? |
|---|---|
| 1e-2 | yes |
| 1e-4 | yes |
| **1e-6** | **no** |

So the guarantee is bounded not only by "the model is symmetric under the permutation" (§8, documented) but also by "no block is more than about nine orders below the largest" — and the objective and gradient blocks are the two that caught the HS071 reproduction, precisely because the constraint and jacobian blocks were permutation-invariant there.

This needs a badly scaled or near-flat-objective model to bite, and when it bites the result is that the probe silently reverts to the no-protection behaviour of `main` — a false negative, never a false refusal. So: not blocking, and I am not asking for a code change here, because changing the rule changes what gets refused and would invalidate the jitter sweep in §4.

Two options, your call:

1. **Document it** — one row in §8 and one sentence in `_probe_agrees`'s docstring, next to the zero-gradient justification that motivates the floor.
2. **Fix it and re-measure** — a per-block floor of `1e-6 * floor` instead of `floor` keeps the zero-gradient block from being held to bit equality while restoring about six orders of detection. That needs the §4 jitter sweep re-run to confirm the acceptance side did not move; I can do that if you want it.

I would take (1) now and (2) only if someone hits it, since the false-negative direction is the safe one.

## Smaller notes, all positive

* **`test_signing_with_ids_is_never_worse_than_signing_without`** is the right test to have written, and confirming it fails when `var_ids` is made required is what makes it evidence rather than decoration. The #607 constraint is genuinely preserved.
* **`test_ids_still_beat_the_probe_where_the_probe_is_blind`** builds the symmetric model, shows the probe accepting and IDs catching. Documenting the limitation *by testing it* is the right form.
* **The `_starts.py` regression** — 2592 → 2920 → 2592 user-callable evaluations, objectives bit-identical, pinned by asserting the ladder never evaluates at the probe point — is the strongest single thing in this PR. A cost measurement that finds a real defect is worth more than the cost number it was taken for.
* **Reporting torch as unmeasured rather than assumed** is right, and §8 naming it "the weakest point in the change" is the correct weight: a threaded backend is the plausible source of >1e-9 jitter, and the sweep says that would be refused.
* The forward-compatibility footgun (a new-build artifact is unreadable by an older build, because `from_json` raises on unknown facets) is #607's policy, unchanged here, and called out in the CHANGELOG. Noted and accepted.

Merging after #636.

jkitchin:
Merged `main` forward (`5e1ff86f`) now that #622, #617 and #618 are in.

One conflict, in `CHANGELOG.md`, and it was the trivial kind: both sides
append a new bullet to the same Unreleased section. Kept both.

Everything else auto-merged, including the three files the other warm-start
PRs also touched — `python/pounce/_warm_start.py`,
`python/tests/test_warm_start_schema.py`, `docs/src/initialization.md`.
I checked that rather than trusting it, since #622 and #638 turned out to
interact where neither branch conflicted: the merged branch's diff against
`origin/main` is five files and **zero lines of Rust**, so there is no path
by which this can move a solver trajectory. The probe facet is entirely a
capture/replay-time gate in Python.

`python/tests` against the merged head, extension rebuilt from this commit:
**916 passed, 38 skipped, 0 failed.**

(`test_continuation.py::test_step_controller_is_what_the_jax_follower_uses`
deselected locally — it imports `jax` with no skip guard, so it fails
identically on `main` in any jax-free environment. Pre-existing and
unrelated; CI has jax.)

Merging once CI is green.

PR #647, opened 2026-08-17
Crossover: identify an exact active set after the IPM converges (#612)

### LINKED ISSUE #612: Crossover: hand a converged IPM iterate to the active-set path to identify an exact active set
## Summary

Add an opt-in **crossover** phase: after the NLP interior-point solve converges, estimate the active set, pivot to the active-set path, and return a solution at which a *linearly independent* set of constraints is satisfied to near equality with multipliers that certify stationarity against that set.

`pounce-convex` already has this for LPs (`crates/pounce-convex/src/crossover.rs`). The NLP path has no equivalent.

Source: Byrd, Nocedal & Waltz, "KNITRO: An Integrated Package for Nonlinear Optimization" (2006), §7. Full gap analysis of that paper in `dev-notes/research/knitro-2006-gap-analysis.md` (branch `claude/knitro-pounce-comparison-75rlpt`).

## Why — three places already paying for an approximate active set

This is **not** a robustness or speed proposal. Mittelmann is 47/47 for POUNCE against 41/47 for IPOPT+MA57, and the residual measured gap there is per-iteration linear algebra on collocation models, which crossover does not touch. The objective value is already right on the problems below. What is missing is *which constraints are active*, and three subsystems work around that today:

1. **`covariance()` has a whole taxonomy for "we cannot tell."** Activity is classified from the barrier geometry into STRONGLY ACTIVE / WEAKLY ACTIVE / **AMBIGUOUS (loosely converged)** / UNIDENTIFIED, each with its own warning path (`docs/src/sensitivity.md`). The AMBIGUOUS class exists precisely because an interior solve cannot answer the question. Crossover collapses it.

2. **Degeneracy keeps returning as an inertia problem.** #540, #541, #544, #592 and the `feral_singular_pivot_floor` knob (`docs/src/troubleshooting.md`) are the same underlying situation — the reduced Hessian collapsing at a degenerate solution — met each time with a perturbation-side heuristic. Crossover attacks it structurally: identify a linearly independent active set, satisfy it to near equality, certify stationarity against it.

3. **There is no IPM → SQP handoff.** The active-set SQP warm-starts only from a previous *SQP* solve (`docs/src/active-set-sqp.md`). A warm-started sequence whose first solve wants the IPM cannot hand off to the SQP for the sweep, where the benchmark measures 0.02–0.17×. Crossover is the missing edge, and it is what makes the MPC / homotopy story end-to-end rather than SQP-only. Related: #611.

## What it does (KNITRO §7)

1. IPM terminates at `(x, s, y, z)` with tolerance `E_tol`.
2. Estimate the active set `A` by a tolerance test on primal-dual feasibility and complementarity.
3. Take one EQP step over `A` plus a line search on the penalty model. **If that satisfies the stopping tolerances, terminate** — this is the common path, and it solves no LPs.
4. Otherwise start the full active-set algorithm from the interior iterate, with the LP trust region sized to exclude every inactive constraint (their eq. 7.22) and `ν₀` slightly above the largest |multiplier| at the interior solution.

The design point worth preserving: step 3 avoids an LP solve wherever the tolerance test is already right, so on well-behaved problems crossover costs approximately one iteration.

## Why it is cheap here

- Both endpoints exist — the IPM and `pounce-qp`-backed active-set SQP.
- `pounce-convex/src/crossover.rs` is the precedent for the representation bridge and multiplier sign transforms.
- It runs **after** convergence behind an option, so **no default trajectory moves and no baseline fixture sweep is required** (contrast: an initial-point or merit-function change would need one, per `CLAUDE.md`).

Note POUNCE's active-set path is SQP, not KNITRO's SLQP, so step 3's "EQP over a tolerance-based estimate" has to be expressed against `pounce-qp`'s working-set interface rather than a literal EQP phase.

## How to test the effectiveness

**The discriminating property is failure of strict complementarity.** The paper is explicit that where it holds, the tolerance estimate is already correct and crossover is a no-op by design. So the metric is never "did it solve" or "how fast" — it is **active-set correctness, dual uniqueness, and the ambiguity classes**.

### Synthetic families (already in-tree, ground truth known)

The warm-start benchmark already separates the three degeneracy modes, and they are not interchangeable. They are small and analytic, so the true active set is known by construction — the ground truth CUTEst cannot give.

| Family | Degeneracy | What crossover should show |
|---|---|---|
| `degenerate_corner` | strict complementarity fails (zero multiplier) | slack and multiplier both `O(√μ)` at a weakly active bound, so activity is a guess today; crossover forces the decision |
| `redundant_rows` | LICQ fails (duplicated rows) | the IPM splits a multiplier across duplicates by centering; crossover puts it on a linearly independent subset, so duals become unique |
| `degenerate_vertex` | primal degeneracy (12 rows tight at a 4-var vertex) | which maximal independent subset is chosen, and whether it is stable under perturbation |

### Real problem shapes

1. **Shadow prices off a degenerate model.** When the dual optimum is a set rather than a point, an IPM returns its analytic center — one interior point of that set, reported as *the* marginal. In blending / pooling / network models those numbers get read as decisions. NETLIB `gen` / `gen1` is the in-tree demonstration (`docs/src/lp-qp-routing.md`): degenerate, rank-deficient, strict complementarity fails, convex IPM burns its 200-iteration budget without a certificate.
2. **Parameter estimation with a fitted parameter at a bound** — the Pyomo estimation model in `docs/src/sensitivity.md`. Whether `covariance()` projects depends on whether the bound is genuinely active; a weakly active one is currently kept at full variance *with a warning*. Crossover turns the warning into an answer.
3. **Over-modeled engineering models with redundant constraints** — the population `dev-notes/research/future-work-roadmap.md` §4.1 already names for C3: CUTEst DECONVBNE, S365, S365MOD, HIMMELBJ, PFIT\*, the ACOPR family.
4. **MPC with a cold first solve** — the one archetype where the metric *is* wall time. The harness already excludes step 0 as a cold solve; the new arm is IPM-cold → crossover → SQP-warm against SQP-cold → SQP-warm. This measures whether the active set is a *usable seed*, a stronger claim than it being correct.

### Where it should show nothing

Stated up front so a null result is not read as failure: the nondegenerate bulk of Hock–Schittkowski, and the Mittelmann collocation models — strict complementarity typically holds there, and their cost is linear algebra (#588), which crossover does not address.

## Scope

In scope: an opt-in post-convergence phase on the NLP interior-point path; the active-set estimate; the EQP-equivalent step through `pounce-qp`; reporting the crossed-over active set and multipliers; a `var_status` / activity path that consumes the exact set.

Out of scope: SLQP as an algorithm (see the gap analysis §3.5), making crossover a default, and any change to the interior iteration itself.

### LINKED ISSUE #646: Crossover reports residuals in the relaxed-bound frame, inflating complementarity above tol
Split out of #612, where it was measured. Not a defect in crossover — a
defect in which model the end-of-run summary thinks it is describing.

## The mismatch

`bound_relax_factor` widens every bound by `δ ≈ 1e-8` before the interior
solve. Crossover (#612) deliberately runs against the **declared** bounds,
because against the relaxed ones a pivot that lands exactly on a binding
constraint reads as strictly interior and the phase identifies nothing.

The end-of-run summary, though, still measures against the **live**
(relaxed) bounds. The two frames disagree by exactly the relaxation, and
the whole disagreement lands on the complementarity term: at an active
constraint the crossed-over point has slack `δ` in the relaxed frame with
the true multiplier `v` beside it, so the product reads `v·δ` rather than
`~μ`.

## Measurement

HS14 (`crossover_is_a_no_op_on_a_strictly_complementary_solution` in
`crates/pounce-algorithm/tests/crossover_issue612.rs`), `v ≈ 1.85`:

| | `crossover=no` | `crossover=yes` |
|---|---|---|
| Dual infeasibility | `1.9e-12` | `8.9e-16` |
| Constraint violation | `2.9e-13` | `2.2e-16` |
| Complementarity | `2.5e-09` | `1.9e-08` |
| Overall NLP error | `2.5e-09` | `1.9e-08` |

Three to four orders better on stationarity and feasibility; ~7× worse on
a complementarity residual computed against a constraint the model does
not contain. In the declared frame that term is zero — which is why
crossover's own never-regress gate (`check_kkt`, which measures there)
accepts the point.

## Why it matters

- `Overall NLP error` in the summary is the max over the three, so it can
  print **above `tol`** on a solve that legitimately converged. Confusing
  on its own; actively misleading next to `EXIT: Optimal Solution Found`.
- `kkt_fidelity_tol` (#173) is applied *after* crossover in
  `optimize_tnlp`, so a threshold set between the two figures above will
  **downgrade the status** of a crossed-over solve that is strictly better
  on every axis the user cares about.

## The obvious fix and why #612 did not take it

Un-relax the live bounds once crossover has been accepted, so every
downstream residual is computed against the model crossover actually
solved. Contained (the bound `Rc`s are uniquely owned — the same
invariant `relax_bounds` and `adjust_variable_bounds` already assert) and
semantically right.

The hazard is that it makes the slack at an active constraint **exactly
zero**, and the `on_converged` sensitivity hook runs after this point and
builds the barrier diagonal `Σ = V/S`. Worth noting that combination is
already reachable without any change here: the sensitivity path sets
`bound_relax_factor = 0` by design (`docs/src/sensitivity.md`), so
`crossover=yes` under a sensitivity-declared solve produces exact-zero
slacks today. Whether that is already a latent divide-by-zero should be
established before or alongside the fix — it is the same question.

## Scope

- Decide the frame a crossed-over solve reports in, and make the summary,
  `curr_unscaled_nlp_error`, and the `kkt_fidelity_tol` gate agree on it.
- Establish what `Σ = V/S` does at exact-zero slack on the sensitivity
  path, with or without this change.
- Regression test asserting the reported complementarity of a
  crossed-over strictly-complementary solve is not worse than the
  interior one.

Docs currently state the behavior honestly rather than hiding it:
`docs/src/crossover.md`, "The one number that gets worse".

jkitchin:
Fixes #612.
Fixes #646.

Adds an opt-in **crossover** phase: after the NLP interior-point solve
converges, pivot to the active-set path and return a point at which a
linearly independent set of constraints holds with *equality*, with
multipliers certifying stationarity against exactly that set. Follows
Byrd, Nocedal & Waltz, *KNITRO: An Integrated Package for Nonlinear
Optimization* (2006), §7.

## The problem

An interior-point method never puts an iterate *on* a constraint. At
termination, "which constraints are active" is an inference from a
tolerance test, not something the solve established. Usually the
inference is right; where it is not, the case is sharply defined.

The discriminating property is **failure of strict complementarity** — a
constraint active with a zero multiplier. There the barrier's own
geometry parks the iterate `O(√μ)` from the constraint: with `μ ≈ 1e-9`
at termination that is about `1e-5`, four orders of magnitude *further
out* than the `1e-8` the solve reports converging at. No post-hoc
tolerance test recovers it, because the information is not in the
iterate.

Three subsystems were already paying for the approximate set:

- `covariance()`'s **AMBIGUOUS (loosely converged)** activity class
  exists precisely because the interior iterate cannot decide.
- Degeneracy has come back as an inertia problem repeatedly (#540, #541,
  #544, #592, `feral_singular_pivot_floor`) and was met each time on the
  perturbation side. A linearly independent active set attacks it
  structurally.
- The active-set SQP could only warm-start from a previous *SQP* solve,
  so a sequence whose first solve wants the IPM had no handoff (#611).

## What was added

`crossover=yes` runs §7 after convergence:

1. estimate the active set by a tolerance test (`crossover_mult_tol`,
   `crossover_primal_tol`);
2. take one EQP-equivalent step over it through `pounce-qp`'s
   working-set interface, with an ℓ₁ line search seeded at `ν₀` just
   above the largest `|multiplier|`;
3. if that already meets the stopping tolerances — the common case, and
   it solves no LPs — stop;
4. otherwise run the full active-set SQP from the interior iterate for
   at most `crossover_max_iter` iterations.

On acceptance the crossed-over point **replaces** `data.curr` rather
than being reported alongside it, so the residual drain, `on_converged`,
`finalize_via_orig_nlp` and the end summary all describe the returned
point with no second copy of the unscaling path. All eight iterate
components move together, using the barrier problem's own relations
(`s = d(x)`, `v_l − v_u = −y_d`).

`last_sqp_working_set()` now returns the identified set, which is the
IPM → SQP handoff that did not previously exist.

## Departures from the paper

KNITRO's active-set path is SLQP; POUNCE's is a line-search SQP. Step 3
is therefore one `pounce-qp` `solve_with_working_set` against the NLP
linearization — which factorizes the hinted set to recover a primal and
then pivots, i.e. "solve the EQP over A, and fix A where the tolerance
test got it wrong". The paper's property that step 3 avoids an LP is
preserved. Step 4's LP trust region (eq. 7.22) has no analogue in a
line-search SQP and is not implemented; only the `ν₀` half of that setup
is.

## Two defects found on the way

Both would have made the phase report the opposite of the truth.

- **`classify_working_set` had the constraint-row multiplier signs
  inverted.** Rows and variable bounds carry *opposite* conventions
  (`λ_g ≤ 0` at a lower bound, `λ_x > 0` at one), because the bound block
  enters the stationarity condition negated. Nothing caught it because no
  test asserted a working-set *estimate*, only the solutions it
  warm-starts — and to `pounce-qp` a wrong hint is merely a slower
  correct answer. Now pinned by a test that solves a QP and asserts the
  classifier reproduces the engine's own answer, rather than a literal.

- **Bound relaxation was inverting the activity test.**
  `bound_relax_factor` widens every bound by `1e-8` before the solve, so
  a point exactly on a declared bound is `1e-8` *inside* the relaxed
  one. Measured against the live bounds, a pivot that lands precisely on
  the binding constraint reads as strictly interior — crossover would
  run, succeed, and correctly report an empty active set. Worse, the
  pivot itself would stop `1e-8` shy of every constraint, because
  against the relaxed problem that point genuinely is optimal. Crossover
  now runs against the declared bounds, via a new
  `Nlp::declared_x_bounds()` alongside the existing `declared_d_bounds()`.

## Measurements

On `min (x₀−1)² + x₁²  s.t.  x₀+x₁ ≤ 1`, whose solution `(1, 0)` is
weakly active: the interior solve leaves the row slack by more than
`1e-7`; after crossover it holds to better than `1e-10`, and the
published set reports it `AtUpper`. Same result for the bound-block twin
(`min (x₀−1)² + (x₁−½)²  s.t.  0 ≤ x ≤ 1`). Both have a baseline test
alongside them, so a future change that happens to land the IPM on the
bound cannot let the crossover assertion pass vacuously.

Control: on HS14, which is strictly complementary, crossover takes its
one step, finds the point already converged, and returns the same
optimum.

## Reporting frame: #646, fixed here rather than deferred

This was originally filed as a follow-up and is now part of this PR.

`bound_relax_factor` (default `1e-8`) widens every bound by `δ` before the
solve. That is invisible during the interior iteration, which never touches
even the widened bound — but crossover puts the iterate *exactly* on the
declared one, which is `δ` **inside** the relaxed one. Measured in the
relaxed frame the returned point therefore carries slack `δ` at every active
constraint, and complementarity reads `|multiplier|·δ` rather than `~μ`.

For a unit multiplier and the default relaxation that is `1e-8` — which is
`tol`. So a strictly better point printed an `Overall NLP error` at or above
the tolerance the run had converged at, and the opt-in `kkt_fidelity_tol`
gate, applied *after* crossover, downgraded `Solve_Succeeded` on it.

An accepted crossover now reports the complementarity term and the two KKT
aggregates against the declared bounds — the frame it solved in, and the
frame the never-regress gate had already judged the point in. HS14
(`v ≈ 1.85`), every figure improving:

| | `crossover=no` | `crossover=yes` (before) | `crossover=yes` (now) |
|---|---|---|---|
| Dual infeasibility | `1.9e-12` | `8.9e-16` | `8.9e-16` |
| Constraint violation | `2.9e-13` | `2.2e-16` | `2.2e-16` |
| Complementarity | `2.5e-09` | `1.9e-08` | **`3.5e-16`** |
| Overall NLP error | `2.5e-09` | `1.9e-08` | **`8.9e-16`** |

Two things the substitution deliberately does not do:

- **It does not touch stationarity or constraint violation.** Stationarity
  involves no bounds, and the crossed-over point is strictly interior to the
  relaxed box, so its violation is zero under either reading. Only
  complementarity moves, so only complementarity is substituted.
- **It does not use the interior machinery's safe slacks.**
  `CalculateSafeSlack` floors a slack below `eps·min(1,μ)` up to about `μ/z`,
  which is what keeps the barrier's `Σ = V/S` finite while the iteration
  runs. At a purified point the active slacks are exactly zero and that floor
  would put `μ/z ≈ 1e-9` straight back — reintroducing, as a fresh reporting
  artifact, the quantity being removed.

The originally-proposed fix — un-relaxing the live bounds on acceptance — was
**not** taken, for that same reason: it makes the active slacks exactly zero
inside the live iterate, where `on_converged`'s sensitivity hook builds
`Σ = V/S`. Substituting at the reporting boundary has no such hazard. (The
underlying question — what `Σ = V/S` does at exact-zero slack, which
crossover plus a sensitivity-declared solve can already reach today because
that path sets `bound_relax_factor = 0` — is independent of this PR and
remains open.)

This is a reporting change and nothing more. It runs after the exit status is
decided, and it applies solely to a point the never-regress gate already
accepted on its declared-bound residuals, so it cannot dress up a worse
iterate; the reading it replaces is the artifact, not the point.

Mechanically: `CrossoverReport::compl_after` carries the declared-frame
max-norm from the new `complementarity_at`, computed on the
`new_with_declared_bounds` adapter crossover already uses;
`IpoptCq::curr_nlp_error_with_complementarity` re-aggregates with the `s_c`
normalisation intact; `application.rs` substitutes into `SolveStatistics`
after the drain, which is what both the end summary and
`apply_kkt_fidelity_gate` read.

Pinned by `crossover_does_not_inflate_the_reported_complementarity`: it
solves HS14 twice and asserts the crossed-over run's reported figures are no
worse than the interior run's. The assertion is relative rather than a fixed
threshold because the defect is itself a comparison between two runs, and a
constant would drift with `tol`. It also guards the other two terms, so a
future "fix" cannot trade stationarity away for the complementarity number.

## Safety

Never-regress: the crossed-over point replaces the interior one only if
constraint violation, stationarity, and objective all hold up against
it; any failure returns the interior solution untouched. Each residual is
compared against `max(interior residual, its own tolerance)` — crossover
puts the iterate *on* the active constraints, which can move a residual
from `1e-12` to `1e-10` while the point is unambiguously better
identified, and refusing that would reject exactly the cases the phase
exists for. What the gate still refuses is a residual crossing its own
tolerance.

`crossover_report()` distinguishes "never ran" from "ran and declined",
so a consumer reasoning about active-set certainty cannot read a
declined crossover as a confirmed active set.

**No fixture sweep.** The phase runs strictly after convergence and is
off by default, so it moves no interior trajectory — the `CLAUDE.md`
trajectory-change rule does not apply. The one behavior change to an
existing path is the `classify_working_set` sign fix, which affects only
the active-set SQP's warm-start *hint*.

## Docs

New `docs/src/crossover.md`, including "Which bounds the reported residuals
are measured against" for the #646 fix, with cross-references added from
`options.md` (bound relaxation), `sensitivity.md` (the AMBIGUOUS class)
and `active-set-sqp.md` (the cold first solve in a warm-started
sequence).


srikanth-gm:
Would you consider a **pre-convergence IPM → SQP handoff** for the future roadmap?

This would be distinct from the post-convergence crossover proposed here. Instead of using SQP only to refine the active set after IPM convergence, the solver would switch earlier and allow SQP to finish the current solve once the IPM iterate is sufficiently feasible and close to a local solution.

### Potential applications

This may be useful for difficult problems started far from a solution, where an interior-point method provides a more robust global phase, while an active-set SQP method may be more efficient near the solution.

Potential applications include trajectory optimization, direct collocation and other large discretized optimal-control problems, as well as MPC, continuation, and parameter-estimation workflows.

### Potential advantage for POUNCE

This could provide POUNCE with a novel differentiator. Based on our discussions with the KNITRO team, KNITRO does not currently offer this type of adaptive pre-convergence IPM-to-SQP transition either.

If this direction seems acceptable and aligned with the project’s goals, I would be willing to submit a separate feature request so it can be evaluated and tracked as a possible roadmap item.

jkitchin:
Thanks — this is a well-posed distinction, and you've stated it correctly. What's in #647 is strictly post-convergence: the IPM does all the optimization, and crossover only identifies the active set once it's done. What you're describing is a genuinely different thing — an adaptive switch that lets the SQP *finish the solve*, with the IPM acting as a globalization phase. Agreed that it's a plausible and interesting direction, and agreed it isn't in KNITRO.

I'm not ready to commit it to the roadmap, but I'd rather explain the gating than give you a vague "maybe."

The applications you list — collocation, trajectory optimization, MPC — mostly reach POUNCE through CasADi, and the CasADi `nlpsol` plugin is fairly new here and still maturing. That's the natural proving ground for this, and how much traction POUNCE gets there is what would tell me whether the effort is aimed at real workloads or at a benchmark. So: worth revisiting once there's evidence POUNCE is pulling weight on those problems.

The other thing I'd want first is **case studies**. This feature makes a cost claim, not a correctness claim, which makes it much harder to validate than the crossover in this PR — that one is provable pointwise on a two-variable problem, whereas a pre-convergence handoff needs a population of problems and is exactly the kind of change that can look good on a mean and hide a new failure mode. Concretely, what I'd want to see before building an adaptive criterion:

- A forced-switch knob (hand off at a fixed iteration `k`), swept over `k`, to establish the ceiling — how much is available *at best* on a given problem. If the best achievable `k` saves a few percent, the adaptive trigger isn't worth designing.
- The shape of that cost-vs-`k` curve: a broad basin means a trigger will be robust; a sharp cliff means any trigger will occasionally be catastrophic.
- Cost measured as function/gradient/Hessian evaluations and wall clock, not iteration counts — an SQP iteration solves a QP and an IPM iteration doesn't, so iteration counts aren't comparable across the switch.

There's also a real implementation cost worth naming: crossover can always decline and hand back the interior point, because it already holds a converged solution. A pre-convergence handoff has no such fallback — if the SQP stalls you've spent iterations and hold nothing, so it needs a return path back into the barrier (re-inflating slacks off the constraints the SQP just landed on, and picking a μ to resume from). That direction doesn't exist today.

So yes, please do open a separate feature request — it's the right way to track this, and it shouldn't ride on this PR. If you're in a position to contribute a case study or two from your own models, particularly through CasADi, that would carry considerably more weight than the issue on its own.

---

srikanth-gm:
Thank you dr.@jkitchin we'll get POUNCE working in our casadi toolchain first and get baseline convergence parity with ipopt before we put in the feature request, so we can show you case studies and cost/benefit analysis.

To clarify, the pre-convergence crossover criterion we're thinking about is based on reaching a threshold barrier parameter (-5)

jkitchin:
The `Σ = V/S` question left open in this PR is now filed as #653.

Measuring it changed the answer: `Σ` does **not** blow up. It is finite in all eight probed combinations, and the weakly-active case crossover exists for goes to exactly `0` (its multiplier is `0`, so `V = 0` in the numerator). The crossed-over slack bottoms out at `1.82e-12` — the QP-plus-line-search residual — never exactly zero, so the division never degenerates.

What is real is a conditioning change: `crossover=yes` + `bound_relax_factor=0` on a strictly complementary optimum grows `max Σ_x` from `1.60e+09` to `1.10e+12`, ~690x. That matters because `classify_activity()` *requires* `bound_relax_factor=0` and `docs/src/sensitivity.md` recommends crossover for the AMBIGUOUS class, so the docs steer users straight into that combination. Whether the growth moves any number `covariance()` actually returns is the open part.

Not a regression from this PR, which changed reporting only.

Issue #649, opened 2026-08-17
[Proposal] Move to a more Rusty code


GermanHeim:
**I was going create the following issue:**

The high-level `pounce-rs::Nlp` builder solver configuration is stringly typed:

```rust
Nlp::new(problem)
     .option_num("tol", 1e-8)
     .option_int("max_iter", 500)
     .option_str("mu_strategy", "adaptive")
     .option_str("hessian_approximation", "limited-memory")
     .solve();
```
 
This loses several advantages expected from a native Rust API:
- option names are not checked by the compiler,
- enum-like values such as "adaptive" are also unchecked,
- IDE completion cannot discover available options or valid choices,
- documentation is disconnected from the relevant Rust types.

Worse, `Nlp::solve()` currently ignores the Result returned by `OptionsList::set_*_value`. Consequently, a typo such as  `.option_str("mu_stratgey", "adaptive")` is rejected by `OptionsList` but silently discarded by the high-level builder, leaving the default setting in effect.

I suggest switching to typed options, for example:

```rust
let options = NlpOptions::default()
     .tolerance(1e-8)
     .max_iterations(500)
     .mu_strategy(MuStrategy::Adaptive)
     .hessian_approximation(HessianApproximation::LimitedMemory);

let solution = Nlp::new(problem)
     .options(options)
     .try_solve()?;
```

**However, on closer inspection, I found a better approach to change the pounce internals.**

TNLP is intentionally the low-level Ipopt compatibility interface, so has several C++-shaped properties:
- failures use `bool` or `Option` instead of descriptive `Result` errors;
- dimensions and indices use the Ipopt-style `i32` Index;
- callers pass `Rc<RefCell<dyn TNLP>>`;
- mutable callbacks and runtime borrow checking replace the original shared mutable objects.

String options currently in many different internals:
1. `upstream_options.rs` independently registers names, defaults, ranges, choices, and documentation.
2. `OptionsList` stores dynamically keyed values.
3. `application.rs` contains manual string-key lookups and converts them into typed fields.
4. `AlgorithmBuilder` separately defines typed structs, enums, and defaults.
5. Public frontends independently translate their inputs into `OptionsList`.

This, of course, causes many issues (https://github.com/jkitchin/pounce/issues/551).

I suggest adding a typed `NlpConfig`/`SolverConfig` layer as the sole configuration consumed by the solver core. Each supported option should have one authoritative definition covering:
- its Ipopt-compatible external name,
- Rust value type,
- default,
[...]

Issue #551, opened 2026-08-08
Registered-but-unread options: the rest of #191 round 2 (split from #486)


jkitchin:
Split out of #486 so that issue can close on the scaling work it was really about. This is section 3 of #486, carried over verbatim in substance; nothing here has been started.

Context: `upstream_options.rs` registers every name Ipopt registers, so an `ipopt.opt` written for Ipopt parses unchanged — which silently turned ~200 knobs into no-ops, since registering an option says nothing about implementing it. #191 fixed the half where the feature runs and only the read site was missing. #485 refused 34 options naming absent features and wired 6 whose read site was merely missing. What follows is the remainder.

## 1. ~29 more group (b) options — feature runs, read site missing

Filter/line-search constants, the penalty acceptor's `nu_*` / `rho` / `eta_penalty`, corrector selectors, mu/termination (`tau_min`, `fixed_mu_oracle`, `s_max`, `neg_curv_test_*`).

Mechanical but not free: each one needs its read site found in whichever subsystem owns it, **and a test proving the option changes behaviour**. A read site that parses a value and discards it is the same silent no-op this whole line of work exists to kill, and it is indistinguishable from a real fix by inspection. That test is the deliverable, not the line that reads the field.

No design decisions here — this part can be done as soon as someone picks it up, and is the natural first PR.

## 2. The "no field at all" category — needs a decision before any code

`max_resto_iter`, `resto_failure_feasibility_threshold`, `limited_memory_max_skipping`, `limited_memory_special_for_resto` have **no field at all**. The feature runs but the sub-capability does not exist, so honouring them means building it, not adding a line. They are neither refusable (the feature they name *does* run, so refusing would fail solves pounce answers correctly today) nor wireable (there is nothing to point a read site at). `unimplemented_options.rs` already documents this shape and deliberately leaves them out of the refusal table, so they are silent today.

The fork, per option:

- **Build the capability.** `max_resto_iter` means a real iteration cap inside restoration, which brings a new failure mode when the cap fires and a status to report it — new behaviour, not plumbing.
- **Refuse them**, extending the `unimplemented_options.rs` table with a distinct message: not "pounce does not implement this feature" but "pounce implements the feature and not this sub-capability."

Worth deciding per option rather than as a block; `limited_memory_max_skipping` is a much smaller build than `max_resto_iter`.

## 3. A per-value refusal mechanism — blocks `limited_memory_initialization`

[...]

PR #652, opened 2026-08-17
pounce-rs: report rejected solver options instead of discarding them

jkitchin:
## Problem

A misspelled solver option passed to the `pounce-rs` builder was silently
dropped, and the solve then ran with the default still in effect. Nothing in
the output distinguished that from an honoured request.

```rust
Nlp::new(Rosenbrock)
    .x0(&[-1.2, 1.0])
    .option_str("mu_stratgey", "adaptive")   // typo for mu_strategy
    .option_num("tolerence", 1e-8)           // typo for tol
    .option_int("max_iterations", 3)         // wrong name for max_iter
    .solve();
```

| | options applied | status | iters | user-visible signal |
|---|---|---|---|---|
| before | 0 of 3 | `SolveSucceeded` | 51 | none |
| after | — | panics / `Err(InvalidOption)` | — | names the option and the reason |

The failure got worse the more the option mattered. The third line above is
the interesting one: had it been the real `max_iter`, the value `3` would
have capped the solve. Both spellings report a full converged solve, so
"my iteration cap is being honoured" and "my iteration cap is being ignored"
were indistinguishable from the output.

## Cause

`crates/pounce-rs/src/builder.rs` applied user options through
`OptionsList::set_*_value` but dropped the returned `Result`:

```rust
for (k, v) in &self.string { let _ = app.options_mut().set_string_value(k, v, true, true); }
```

The validation was never missing. `OptionsList` checks every option against
the registry — unknown name, wrong value type, out-of-range value,
unregistered choice (`crates/pounce-common/src/options_list.rs:106-127`) —
and correctly returned `Err`. The builder threw the answer away.

Only this crate was affected. The Python frontend
(`crates/pounce-py/src/problem.rs:949-973`), batch
(`crates/pounce-py/src/nlp_batch.rs:98`), and the C interface all already
propagated these errors.

## Fix

A new `NlpError` enum, and a new `Nlp::try_solve` holding what was previously
`solve`'s body — a `Result` yielding `Solution` on success and `NlpError` on
failure. (Written out in prose because GitHub's body sanitizer silently eats
Rust generic syntax, even inside a code fence.)

`NlpError::InvalidOption` carries the option name, the value, and the
registry's own explanation; `UnknownVariableCount` and `Initialize` cover the
other two setup failures.

`Nlp::solve` keeps its signature and delegates, panicking on `Err`. **This is
a behaviour change**: a caller passing an invalid option previously got a
silent default and now gets a panic. That is the point of the fix — solving a
configuration the caller did not ask for is the defect — but it can turn a
working-looking build into a crash on upgrade, so it is called out in the
CHANGELOG. The considered alternative was a journal warning with `solve`
otherwise unchanged; rejected because the silent-default path stays reachable
for everyone who doesn't switch to `try_solve`, which leaves the original
defect in place for the default API.

Two deliberate non-changes:

- The two internal defaults the builder sets for itself
  (`hessian_approximation`, `sqp_hessian`) still use `let _ =`. Both names
  are hardcoded and valid; there is no caller error to report.
- `Ok(false)` is not treated as a failure. It means an earlier no-clobber
  setting won, not that the option was rejected. Only the `Err` arm is
  escalated, matching what the Python frontend does.

## Blast radius

Confined to the `pounce-rs` builder path. The other three frontends already
behaved this way, so nothing converges on a new behaviour.

**Not a trajectory change, and no fixture sweep was run.** This changes only
whether an *invalid* configuration is reported; it cannot alter which
correction a valid configuration reaches for, or the order or scale of its
steps. Every pre-existing test passes unchanged, which is consistent with
that. Flagging the reasoning explicitly rather than burying it, since the
cost of being wrong about this is the gh#544 shape.

## Tests

New `crates/pounce-rs/tests/option_validation.rs` — 10 tests covering all
four rejection modes the registry distinguishes (unknown name across all
three setter types, wrong value type for a real option, unregistered choice,
out-of-range number), the `solve()` panic path, `UnknownVariableCount`, and
that valid options still reach the solver.

**How I know they bite.** Most of that file cannot compile against the parent
commit, because `try_solve` and `NlpError` do not exist there — true of any
new-API test, and not evidence of anything. The load-bearing check is
`solve_panics_on_a_rejected_option`, which uses only the pre-existing
`solve()` API. Extracted to a standalone file and run against the parent
(`dad6748`) it fails, for the stated reason:

```
test solve_panics_on_a_rejected_option - should panic ... FAILED
note: test did not panic as expected at crates/pounce-rs/tests/bite_check.rs:14:4
test result: FAILED. 0 passed; 1 failed
```

The other test that would survive a lazy implementation is
`an_accepted_option_actually_takes_effect`. Every test above would pass
against a builder that rejected *everything*; that one pins the actual defect
by proving the two cases are now distinguishable — `max_iter=3` stops the
solve short (`success == false`, ≤3 iterations) while the same problem
without it converges in more.

Verification, re-run after merging `main` (`450e5b0`) into the branch:
`cargo test -p pounce-rs --all-features` green (10 new, all pre-existing, 12
doctests), `cargo fmt --all -- --check` clean, `cargo clippy --all-targets`
exits 0 with no `correctness` or `suspicious` findings.

---

- [x] Book page under `docs/src/` updated (`rust.md`; no new page, no
      `SUMMARY.md` change needed)
- [x] Every claim in this PR body and in the code comments is true of the
      code as it stands now

## On #649

Refs #649 — deliberately not a closing keyword. This fixes the concrete
defect the issue reports (the ignored `Result`), but #649 also proposes
replacing the string option layer with a typed `NlpConfig` / `SolverConfig`
across the solver core. That part is untouched here and worth its own
discussion, so the issue should stay open.

One correction for that discussion, from reading the code while fixing this:
the issue's claim that the current design "causes #551" does not really hold.
#551 is about registered-but-unread options, and two of its three sections
(missing capability, per-value refusal) are not addressable by typing at all.
Its claim that five places independently define options is also off —
`upstream_options.rs` is auto-generated by `tools/optport/gen.py` from
upstream, and `AlgorithmBuilder` is already a typed layer with real enums.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,35 @@ changes.
 
 ## [Unreleased]
 
+- **`pounce-rs`: a rejected solver option is no longer silently
+  discarded** (#649).
+
+  `Nlp::solve` applied `.option_num` / `.option_int` / `.option_str`
+  through `OptionsList::set_*_value` but dropped the returned `Result`.
+  `OptionsList` validates every option against the registry — unknown
+  name, wrong value type, out-of-range value, unregistered choice — so a
+  typo like `.option_str("mu_stratgey", "adaptive")` was rejected,
+  discarded, and the solve then ran with the default still in effect.
+  Nothing in the output distinguished that from an honoured request, and
+  the failure got worse the more the option mattered: a misspelled
+  `max_iter` reported a full converged solve at the default cap.
+
+  Rejected options are now reported. New `Nlp::try_solve` returns
+  `Result<Solution, NlpError>`, with `NlpError::InvalidOption` naming the
+  option, the value, and the registry's reason; `NlpError::UnknownVariableCount`
+  and `NlpError::Initialize` cover the other two setup failures.
+  `Nlp::solve` keeps its signature and now panics on those conditions
+  rather than solving a configuration the caller did not ask for — a
+  behaviour change for callers who were passing an invalid option and
+  (knowingly or not) getting the default. A solve that runs and fails to
+  converge is unaffected: still `Ok` with `success == false`.
+
+  The two internal defaults the builder sets for itself
+  (`hessian_approximation`, `sqp_hessian`) still ignore their result;
+  both names are hardcoded and valid. Only the interior-point path
+  through `pounce-rs` was affected — the Python, batch, and C interfaces
+  already propagated these errors.
+
 - **CasADi plugin: solver diagnostics, and two defects found exposing
   them** (#634).
 
--- a/crates/pounce-rs/README.md
+++ b/crates/pounce-rs/README.md
@@ -61,7 +61,17 @@ to a limited-memory (L-BFGS) approximation. This keeps simple problems concise
 without sacrificing access to exact derivatives when needed.
 
 Solver options use the same names as upstream Ipopt
-(`option_num`, `option_int`, `option_str`).
+(`option_num`, `option_int`, `option_str`). Names, value types, ranges, and
+choices are validated against the option registry. A rejected option is
+never applied silently: `solve` panics, and `try_solve` returns
+`Err(NlpError::InvalidOption { .. })` naming the option and the reason.
+
+```rust
+let sol = Nlp::new(P)
+    .x0(&[0.0, 0.0])
+    .option_str("mu_strategy", "adaptive")
+    .try_solve()?;                        // Result<Solution, NlpError>
+```
 
 ## Result
 
--- a/crates/pounce-rs/src/builder.rs
+++ b/crates/pounce-rs/src/builder.rs
@@ -109,8 +109,56 @@ pub struct Solution {
     pub stats: SolveStatistics,
 }
 
+/// Why a solve could not be started.
+///
+/// Returned by [`Nlp::try_solve`]; [`Nlp::solve`] panics on the same
+/// conditions. Every variant here is a configuration error detected
+/// *before* the solver runs — a solve that runs and fails to converge is
+/// not an error, it comes back as a [`Solution`] with `success == false`
+/// and the reason in `status`.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum NlpError {
+    /// Neither [`var_bounds`](Nlp::var_bounds) nor [`x0`](Nlp::x0) was
+    /// called, so the number of variables is unknown.
+    UnknownVariableCount,
+    /// An option passed to [`option_num`](Nlp::option_num) /
+    /// [`option_int`](Nlp::option_int) / [`option_str`](Nlp::option_str)
+    /// was rejected by the options registry: unknown name, wrong value
+    /// type for that name, or a value outside the registered range or
+    /// set of choices.
+    InvalidOption {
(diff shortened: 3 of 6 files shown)


GermanHeim:
@jkitchin, a quick correction regarding the last note.

`dead_code` (via `#[warn(dead_code)]`, from clippy) shows on struct fields that are written but never read, and on enum variants that are never constructed or never matched.
So an option that gets parsed into a typed field but is never touched by the solver logic downstream would trigger it, which is what "registered but unread" means in #551.
That's arguably the strongest argument for typing. But we need to change more of the codebase than just that builder.

jkitchin:
@GermanHeim thanks — I checked this against the workspace rather than from memory, and I don't think `dead_code` gives us the signal here. Two things, then the part I agree with.

**1. It doesn't fire on `pub` fields, which is what our config fields are.**

`dead_code` is silent on a `pub` field of a `pub` struct reachable from a library crate's root. I added `pub tau_min_experiment: Number` to the real `MuOptions` (`crates/pounce-algorithm/src/alg_builder.rs:494`), defaulted it, read it nowhere — `cargo build -p pounce-algorithm` and `cargo clippy -p pounce-algorithm` both give zero warnings and zero mentions of the field. That is exactly the "typed field never touched by solver logic" case.

We can't fix that by making the fields private, because configuration crosses crate boundaries in a 20-crate workspace: `PdPerturbationHandler.delta_xs_max` lives in `pounce-common` and is assigned from `pounce-algorithm` (`alg_builder.rs:1016`). Any option whose typed home and read site are in different crates has to be `pub`, which puts it permanently out of the lint's reach.

Two smaller notes: `dead_code` is a rustc built-in, not a clippy lint — so it's already on by default everywhere here, which is itself evidence it isn't going to start reporting these. And "never matched" isn't a trigger: a variant that is *constructed* but never *matched* draws no warning, which is precisely the shape we'd care about. Even for private fields, any single read satisfies it — a getter called once, or the value echoed into an options dump or a journal line.

**2. Against #551's three sections, it reaches at most one.**

- §1 (~29 options, feature runs, read site missing) — the only one it could touch, and only if the field were private *and* single-crate.
- §2 (`max_resto_iter` and friends, no field at all) — typing creates nothing to lint unless you also build the missing capability.
- §3 (per-value refusal, `limited_memory_initialization`) — the value *is* read and mapped to a fallback, so that's a read; the unmapped choices (`scalar3`, `scalar4`, `constant`) have no variant for typing to create.

This lines up with what #551 already says about the enforcement mechanism: "A read site that parses a value and discards it is the same silent no-op … and it is indistinguishable from a real fix by inspection. That test is the deliverable, not the line that reads the field." A per-option behaviour test, derived mechanically from the registry, is the thing that actually catches this. A lint isn't.

So I'd put this as the weakest argument for typing rather than the strongest — but the strong ones in #649 are untouched by any of the above and don't need it: compiler-checked option names, enum-checked values, IDE completion, and one authoritative definition per option all stand on their own.

**And your last sentence is the right one** — more than the builder has to change. That's why #652 was scoped to the ignored `Result` and #649 is still open. Happy to move this thread over there, since it's really about the design rather than this merged PR.

---

PR #670, opened 2026-08-18
casadi: build against both spellings of the convexify helper (#668)

### LINKED ISSUE #668: CasADi plugin does not build against CasADi master (`convexify_eval` renamed)
## Preface

We are now building and running POUNCE through the CasADi `Nlpsol` plugin inside a
proprietary optimisation toolchain, on a real problem, end to end. Getting there
required working around two issues in the plugin, which we currently patch into the
source as part of our build pipeline. Both fixes were produced with AI assistance.

We would like to upstream them so we can drop the local patching, and this write-up
exists to give you the diagnosis rather than just the diff. The attached patch is
included as evidence that the fix works and to make the change concrete — we are
not proposing it be applied as-is. The shape of the fix is yours to decide.

[0002-casadi-accept-both-convexify-helper-names.patch](https://github.com/user-attachments/files/31160481/0002-casadi-accept-both-convexify-helper-names.patch)

## Summary

`casadi_nlpsol_pounce.cpp` calls the CasADi runtime helper `convexify_eval`.
CasADi renamed it to `casadi_convexify_eval` after the 3.7.2 release, so the
plugin no longer compiles against current CasADi master or against any nightly
build made from it.

## Symptom

```
casadi_nlpsol_pounce.cpp: In static member function 'static bool casadi::PounceInterface::cb_h(...)':
casadi_nlpsol_pounce.cpp:564:16: error: 'convexify_eval' was not declared in this scope
  564 |         return convexify_eval(&self->convexify_data_.config, values, values,
      |                ^~~~~~~~~~~~~~
```

The failure is unconditional at compile time: `cb_h` is compiled whether or not
`convexify_strategy` is ever used, so a plugin built against master fails even for
users who never enable convexification.

## Affected versions

The helper is declared in `casadi/core/runtime/casadi_convexify.hpp`. On current
master:

```cpp
// SYMBOL "convexify_eval"
template<typename T1>
int casadi_convexify_eval(const casadi_convexify_config<T1>* c, const T1* Hin, T1* Hout,
                          casadi_int* iw, T1* w) {
```

The codegen symbol name is unchanged — only the C++ function was renamed. So the
break is source-level only, and nothing in the generated-code path is affected.

Reproduced against a CasADi build reporting `CASADI_VERSION_STRING "3.7.2+"`,
`CASADI_VERSION=31`, `CASADI_IS_RELEASE=0`.

## Why the obvious fix is wrong

Renaming the call site to `casadi_convexify_eval` fixes master and breaks 3.7.2,
which is the version the plugin's own Makefile targets. Version macros do not
separate the two either: the nightly reports the same `CASADI_MAJOR/MINOR/PATCH`
as the release, differing only in `CASADI_IS_RELEASE`. Keying on that would then
misfire on the next release, which will carry the new name with
`CASADI_IS_RELEASE=1`.

The distinguishing fact is not a version number, it is which of the two names the
installed CasADi actually declares.

## Proposed fix

Detect the available spelling by overload resolution. Both calls are dependent on
the template parameters, so each is looked up by ADL at instantiation; the overload
naming a helper this CasADi does not declare fails substitution in its trailing
return type — the immediate context — and is silently dropped rather than becoming
a hard error.

```cpp
template <typename Config, typename T>
static auto convexify_eval_compat(int, const Config* c, const T* Hin, T* Hout,
                                  casadi_int* iw, T* w)
    -> decltype(casadi_convexify_eval(c, Hin, Hout, iw, w)) {
  return casadi_convexify_eval(c, Hin, Hout, iw, w);
}
template <typename Config, typename T>
static auto convexify_eval_compat(long, const Config* c, const T* Hin, T* Hout,
                                  casadi_int* iw, T* w)
    -> decltype(convexify_eval(c, Hin, Hout, iw, w)) {
  return convexify_eval(c, Hin, Hout, iw, w);
}
```

The unused `int` / `long` first parameter exists only to rank the two overloads, so
that a CasADi declaring both spellings resolves to the current name. The call site
becomes `convexify_eval_compat(0, ...)`.

This keeps a single source tree building against 3.7.2 and master with no build
flags, no configure probe, and no version test that has to be revised at the next
CasADi release.

## Verification

Compiled against the nightly described above, `-std=c++17`, GCC 15 (MSYS2 UCRT64),
no build-system modifications: the plugin builds clean and
`casadi/pounce-casadi-smoke` passes on both the `feral` and `ma57` backends with
unchanged results (`x = [0.823128, 0.676872]`, `f = 0.0313283`, 23 iterations).

I have **not** compiled this against released 3.7.2 — I do not have that build
available. The fallback overload is the code that is in the tree today, and the
mechanism selecting it is standard SFINAE, but the release path deserves a real
compile before merge.

## Note for anyone patching this downstream

A `sed` that rewrites `return convexify_eval(` to `return casadi_convexify_eval(`
is the natural stopgap, and it interacts badly with the fix above: it also rewrites
the *fallback* overload's body, leaving a function whose trailing return type names
one helper and whose body names the other. That compiles by accident — the
fallback is never instantiated on master — but it is silently wrong. Any such
downstream patch should be dropped once a fix along these lines lands.

## AI disclosure

This report and the accompanying patch were produced with AI assistance (Claude,
via GitHub Copilot), directed and reviewed by a human.

What that means concretely:

* The build failure, the header evidence, and the successful compile of the
  proposed fix against the nightly were all observed on the machine described
  above, not inferred.
* The gap in that evidence is stated in the Verification section and is real: the
  fallback overload — the path that matters for released 3.7.2 — has not been
  compiled. It is unexercised on master by construction, which is exactly the kind
  of thing worth checking rather than taking on trust.
* The SFINAE argument for why the unavailable overload drops out instead of
  erroring is a standards claim, and it is the part of this report most worth
  verifying independently before merge.

We are flagging this because the review burden falls on you either way, and you
should weight the claims accordingly.

jkitchin:
Fixes #668.

Reported, diagnosed and patched by @srikanth-gm, who is building POUNCE through
the `Nlpsol` plugin against a CasADi nightly and has been carrying this as a
local patch. The mechanism below is theirs; what is added here is the test that
keeps it honest.

## Problem

The plugin does not compile against CasADi master, or any nightly built from
it. CasADi renamed the runtime helper `convexify_eval` to
`casadi_convexify_eval` after 3.7.2:

```
casadi_nlpsol_pounce.cpp:564:16: error: 'convexify_eval' was not declared in this scope
  564 |         return convexify_eval(&self->convexify_data_.config, values, values,
```

The failure is unconditional. The call sits in `cb_h`, a non-template member
compiled whether or not `convexify_strategy` is ever set, so the build fails
for users who never convexify anything.

| plugin source | vs casadi 3.7.2 | vs casadi master `973b086` |
|---|---|---|
| before | compiles | **error at :564** |
| after | compiles | compiles |

Not urgent for release users — 3.7.2 is still latest on PyPI, so nothing is red
today. It is not optional either: CI installs `casadi` unpinned and derives the
header checkout from `casadi.__version__`, so the day the next CasADi release
ships this becomes a red `CasADi plugin parity` on both legs, and a failed wheel
build for that minor, with no change on our side.

## Cause

A source-level rename, and nothing else. The codegen SYMBOL name is unchanged
(`// SYMBOL "convexify_eval"` on both), the signature is unchanged, and the
generated-code path never reaches this helper at all — the plugin refuses to
code generate `convexify_strategy` (`casadi_nlpsol_pounce.cpp:1093`).

Verified against upstream rather than taken from the report: `master` declares
only `casadi_convexify_eval`, with no deprecated alias and no macro providing
the old spelling; `3.7.2` declares only `convexify_eval`. **3.6.7 declares the
old name too** — which matters below.

## Fix

`casadi/convexify_compat.hpp` detects which name the installed CasADi declares,
by overload resolution: two `convexify_eval_compat` overloads, each naming one
spelling in its trailing return type, so the one whose helper is not declared
fails substitution in the immediate context and drops out. An unused `int` /
`long` first parameter ranks them, so a CasADi declaring both resolves to the
current name. The call site becomes `convexify_eval_compat(0, ...)`.

**Rejected: renaming the call site.** It fixes master and breaks 3.7.2 *and*
3.6 — the wheel ships a build for each, under `_plugins/{minor}/`, so this is
not a hypothetical older version, it is both shipped artifacts.

**Rejected: a version test.** No macro separates the two. The nightly carrying
the rename reports the same `CASADI_MAJOR/MINOR/PATCH` as 3.7.2, differing only
in `CASADI_IS_RELEASE` — which the next release flips back while keeping the new
name. The Makefile hard-codes `CASADI_IS_RELEASE=1` anyway.

**Rejected: a configure probe in the Makefile.** It works for our own builds,
but this `.cpp` is compiled by users with their own build systems; the detection
belongs in the source, where it needs no cooperation from the caller.

The result is one source tree building against 3.6, 3.7 and master with no build
flags and nothing to revise at the next CasADi release. If CasADi renames it a
third time, the failure is a legible overload-resolution error naming both
candidates rather than silence.

## Blast radius

Bit-identical except that the plugin now compiles. Same helper, same arguments,
same in-place operation on the same widened pattern — **not a trajectory
change**, so no fixture sweep is owed. `convexify_eval` was the only unprefixed
CasADi runtime helper the plugin called (`casadi_copy` was always prefixed), so
the blast radius is one call site. The emitted C is untouched, per the codegen
refusal above.

Nothing else in CasADi master breaks the plugin: it compiles clean against
`973b086` in full, not just past this line.

No Rust changed (0 `.rs` files in the diff).

## Tests

`casadi/tests/convexify_compat/` — compiles **both** overloads against mock
CasADi declarations, with no CasADi installed and nothing linked. It exists
because of a gap that outlives this PR: any one CasADi declares exactly one of
the two names, so *any* build compiles one overload and leaves the other
unchecked — and the unchecked one is what most users get, since the wheel's 3.6
and 3.7 builds both take the fallback while CI builds against the current
release. Mocking is what lets one machine compile every combination.

It asserts which overload ran, that the arguments arrive intact (an overload
that compiled but swapped `Hin`/`Hout` would pass a bare "it built" check and
then convexify against the wrong buffer), and that a CasADi declaring *neither*
name is rejected rather than silently resolved to something else.

`make -C casadi compat`, plus a `CasADi convexify name shim` CI job on Linux,
macOS and **Windows** — the only place in CI the plugin source meets MSVC, which
is the toolchain the Windows wheel is built with and the one compiler nobody had
pointed at this mechanism.

Verified:

| check | result |
|---|---|
| shim: g++ 13 + clang, `-std=c++11/14/17`, 4 declaration states | 24/24 as expected |
| shim: **MSVC** (VS 2026, `cl` 18.8.2), `/std:c++14` and `/std:c++17`, same 4 states | 8/8, in CI on this PR |
| plugin vs real casadi 3.7.2 headers (fallback overload) | compiles, g++ and clang |
| plugin vs real casadi master `973b086` (prefixed overload) | compiles, g++ and clang |
| **pre-fix** source vs master | reproduces the reported error, at the reported line |
| `make -C casadi test` — parity vs bundled Ipopt, incl. convexification | all checks pass |

MSVC was the one leg that had not been run when this PR was opened — the
mechanism is standard SFINAE and MSVC's two-phase lookup is the likeliest of the
three to disagree, so it was the gap worth naming rather than assuming away. It
has now run green on this PR, including the negative case, which is why the
claim above is a measurement rather than an expectation.

On "tests fail on the parent commit": the compat test is a new file and cannot
run on the parent, so the pre-fix check in the table above is what stands in for
it — the parent's source, compiled against master headers, fails with the
reported error at the reported line, and stops failing with this diff applied.

## Rebased onto #669

#669 (gh#667) landed in `casadi/` while this was open — the output-tearing flush
guard in `guarded`, a `test_output_interleaving.cpp` target, and its own
CHANGELOG and `docs/src/casadi.md` sections. Merged with no textual conflicts,
but the convexify call site now sits inside a `guarded` lambda whose scope also
carries that flush guard, so the merged tree was re-verified rather than assumed:
`make -C casadi compat` green, the plugin compiles against real 3.7.2 and real
master on g++ and clang, and the full parity suite passes — including #669's own
`output: POUNCE does not tear embedder lines` check (0/13 lines torn).

---

- [x] Tests fail on the parent commit for the stated reason — see the note above
      on how this was checked for a new-file test
- [x] Book page under `docs/src/` updated (`docs/src/casadi.md`); no
      `SUMMARY.md` change — not a new page
- [x] `cargo fmt --all -- --check` clean; `cargo clippy` / `cargo test` not
      re-run — the diff contains no Rust
      as it stands now — body re-read against the final diff after the #669
      merge, and the MSVC row updated from "not yet run" to the result

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -647,3 +647,47 @@ jobs:
           assert abs(sol[0] - 1) < 1e-6 and abs(sol[1] - 1) < 1e-6, sol
           print("wheel solves with the build tree hidden, x* =", sol)
           PY
+
+  # Compile the convexify name shim against every spelling of CasADi's
+  # runtime helper (gh#668). Cheap and hermetic: no CasADi, no Rust, no
+  # link step — mock declarations and a compiler.
+  #
+  # This job exists because the job above cannot cover it. CasADi renamed
+  # `convexify_eval` to `casadi_convexify_eval` after 3.7.2, and any one
+  # CasADi declares exactly one of the two, so a build against a real
+  # CasADi compiles one overload of the shim and leaves the other
+  # unchecked — and it is the *unchecked* one that most users get, since
+  # the wheel ships plugins for 3.6 and 3.7 and both take the fallback.
+  # The mock is what lets one runner compile both.
+  #
+  # Windows is here and nowhere else in this workflow: the Windows wheel is
+  # built with MSVC, and MSVC's two-phase lookup is the likeliest of the
+  # three to disagree about a dependent name that is not declared. That is
+  # the whole mechanism this shim rests on.
+  convexify-name-shim:
+    name: CasADi convexify name shim (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.11"
+
+      - name: Compile against every spelling
+        if: runner.os != 'Windows'
+        run: python3 casadi/tests/convexify_compat/run.py
+
+      # `cl` is only on PATH inside a developer shell, and vswhere ships with
+      # every runner image, so no extra action is needed to find it.
+      - name: Compile against every spelling (MSVC)
+        if: runner.os == 'Windows'
+        shell: cmd
+        run: |
+          for /f "usebackq tokens=*" %%i in (`"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath`) do set VSPATH=%%i
+          if not defined VSPATH (echo could not locate a Visual Studio install & exit /b 1)
+          call "%VSPATH%\VC\Auxiliary\Build\vcvars64.bat" || exit /b 1
+          python casadi\tests\convexify_compat\run.py --cxx cl
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,41 @@ changes.
 
 ## [Unreleased]
 
+- **The CasADi plugin builds against CasADi master again** (#668).
+
+  CasADi renamed the runtime helper `convexify_eval` to
+  `casadi_convexify_eval` after 3.7.2. The plugin calls it from `cb_h`,
+  which is compiled whether or not `convexify_strategy` is ever used, so
+  the failure was unconditional: a plugin built against master or any
+  nightly made from it did not compile, for users who never convexify
+  anything. Reported with a diagnosis and a working patch by
+  @srikanth-gm.
+
+  Renaming the call site would fix master and break 3.7.2 — and 3.6,
+  which the wheel also ships a build for. No version macro separates the
+  two: the nightly carrying the rename reports the same
+  `CASADI_MAJOR/MINOR/PATCH` as 3.7.2, differing only in
+  `CASADI_IS_RELEASE`, which the next release will flip back while
+  keeping the new name. What distinguishes them is not a version, it is
+  which name the installed CasADi declares, so `convexify_compat.hpp`
+  detects that by overload resolution. One source tree, no build flags,
+  no configure probe, and no version test to revise at the next CasADi
+  release. The codegen SYMBOL name is unchanged and generated code never
+  reaches this helper (the plugin refuses to code generate
+  `convexify_strategy` at all), so nothing about the emitted C moves.
+
+  The shim has two overloads and any one CasADi declares exactly one of
+  the names, so building against a real CasADi compiles one of them and
+  leaves the other unchecked — and the unchecked one is what most users
+  get, since the wheel's 3.6 and 3.7 builds both take the fallback. That
+  is what `casadi/tests/convexify_compat/` is for: it compiles both
+  against mock declarations, with no CasADi installed and nothing linked,
+  asserts which overload ran and that the arguments arrive intact, and
+  checks that a CasADi declaring *neither* name is rejected rather than
+  silently resolved. `make -C casadi compat`, and the `CasADi convexify
+  name shim` CI job — which is also the only place in CI the plugin
+  source meets MSVC, the toolchain the Windows wheel is built with.
+
 - **The restoration divergence guard's waiver now measures a floor
   instead of counting iterations** (#661, #664).
 
--- a/casadi/Makefile
+++ b/casadi/Makefile
@@ -3,6 +3,8 @@
 #   make                 build libcasadi_nlpsol_pounce.so
 #   make test            build, then run the parity checks against ipopt
 #   make examples        run every script in examples/
+#   make compat          compile the convexify name shim against every
+#                        CasADi spelling (needs no CasADi at all)
 #   make clean
 #
 # Two inputs are needed and both are auto-detected where possible:
@@ -89,7 +91,7 @@ CXXFLAGS ?= -O2 -std=c++17 -fPIC -Wall
 INCLUDES  = -I$(CASADI_SRC) -I$(CASADI_INC) -I$(POUNCE_INC)
 LDFLAGS   = -L$(CASADI_LIB) -lcasadi -L$(POUNCE_LIB) -lpounce_cinterface $(RPATH)
 
-.PHONY: all check-env fetch-src abi-flags install uninstall test examples clean
+.PHONY: all check-env fetch-src abi-flags install uninstall test examples compat clean
 
 all: check-env $(PLUGIN)
 
@@ -131,7 +133,7 @@ abi-flags:
 	@echo "installed casadi $(CASADI_VER) at $(CASADI_LIB)"
 	@$(PYTHON) -c "import casadi; print(casadi.CasadiMeta.compiler_flags())"
 
-$(PLUGIN): casadi_nlpsol_pounce.cpp pounce_runtime.hpp
+$(PLUGIN): casadi_nlpsol_pounce.cpp pounce_runtime.hpp convexify_compat.hpp
 	$(CXX) -shared $(CXXFLAGS) $(DEFS) $(INCLUDES) $< -o $@ $(LDFLAGS)
 	@$(FIXUP)
 
@@ -157,6 +159,13 @@ examples: all
 	  echo "=== $$f"; \
 	  CASADIPATH=$(CURDIR) $(PYTHON) $$f || exit 1; \
 	done
+
+## Compile `convexify_compat.hpp` against every spelling of CasADi's
+## convexify runtime helper (gh#668). Needs no CasADi and no solver
+## library -- mock declarations and a compiler -- so it runs anywhere,
+## and it covers the overload the installed CasADi does not exercise.
+compat:
+	$(PYTHON) tests/convexify_compat/run.py
 
 clean:
 	rm -f $(PLUGIN)
--- a/casadi/README.md
+++ b/casadi/README.md
@@ -33,6 +33,11 @@ covers building and installing the plugin.
   `CasADi plugin parity` job.
 - `pounce_runtime.hpp` — the C the plugin emits into generated code, the
   counterpart of CasADi's `ipopt_runtime.hpp`.
+- `convexify_compat.hpp` — one shim, for one CasADi rename. See
+  *Source compatibility across CasADi versions* below.
+- `tests/convexify_compat/` — compiles that shim against every spelling,
+  with no CasADi installed. `make compat`, and the `CasADi convexify name
+  shim` job in CI.
 - `examples/` — eight runnable scripts, from hello-world to embedded codegen.
 
 ## Build
@@ -127,7 +132,38 @@ Uninstall with `make uninstall`.
 ```bash
 make test          # parity checks against ipopt — all should PASS
 make examples      # runs every script in examples/
+make compat        # the convexify name shim, against every CasADi spelling
 ```
+
+`make compat` is the odd one out: it needs neither CasADi nor the solver
+library, only a compiler. See below for what it is checking.
+
+## Source compatibility across CasADi versions
+
+CasADi renamed the runtime helper `convexify_eval` to
+`casadi_convexify_eval` after 3.7.2 (gh#668). The codegen SYMBOL name and
+the signature are unchanged, so the break is source-level only — but it is
+unconditional: the call sits in `cb_h`, which compiles whether or not
+`convexify_strategy` is ever used. A plugin built against current CasADi
+master fails for users who never convexify anything.
+
+Nothing in the version macros separates the two spellings — the nightly
+that first carried the rename reports the same
+`CASADI_MAJOR/MINOR/PATCH` as 3.7.2 — so `convexify_compat.hpp` detects
+which name the installed CasADi declares, by overload resolution. One
+source tree builds against 3.6, 3.7 and master with no build flags and no
+version test to revise at the next CasADi release.
+
+The catch, and the reason `tests/convexify_compat/` exists: a real CasADi
+declares exactly one of the two names, so building against one CasADi
+compiles one overload of the shim and leaves the other unchecked — and it
+is the unchecked one that most users get, since the wheel ships plugins
+for 3.6 and 3.7 and both take the fallback. `make compat` compiles both
+against mock declarations, so neither branch can rot unnoticed. Checked
+locally on g++ and clang; the MSVC leg — the toolchain the Windows wheel
+is built with, and the likeliest of the three to disagree about a
+dependent name — runs in CI. Add a compiler to `run.py` rather than
+trusting the mechanism on a new one.
 
 ## ABI: what has to match, and why
 
--- a/casadi/casadi_nlpsol_pounce.cpp
+++ b/casadi/casadi_nlpsol_pounce.cpp
@@ -20,6 +20,7 @@
 
 #include "casadi/core/nlpsol_impl.hpp"
 #include "casadi/core/convexify.hpp"
+#include "convexify_compat.hpp"
 #include "pounce_runtime.hpp"
 #include <cmath>
 #include <cstring>
@@ -561,8 +562,11 @@ namespace casadi {
       // which is the pattern the solver was given, so `values` is big enough.
       return guarded(m, "Hessian convexification", [&] {
         ScopedTiming tic(m->fstats.at("convexify"));
-        return convexify_eval(&self->convexify_data_.config, values, values,
-                              m->iw, m->w) == 0;
+        // `convexify_eval_compat`, not `convexify_eval`: CasADi renamed the
+        // helper after 3.7.2 and the shim picks whichever name is declared
+        // (gh#668, convexify_compat.hpp).
+        return convexify_eval_compat(0, &self->convexify_data_.config, values,
+                                     values, m->iw, m->w) == 0;
       });
     }
     // upper-triangular CCS == lower-triangular triplet (row/col swap)
--- a/casadi/convexify_compat.hpp
+++ b/casadi/convexify_compat.hpp
@@ -1,0 +1,68 @@
+// Call CasADi's convexify runtime helper under whichever name the
+// installed CasADi declares (gh#668).
+//
+// CasADi renamed the C++ helper `convexify_eval` to `casadi_convexify_eval`
+// after the 3.7.2 release. The codegen SYMBOL name is unchanged, and so are
+// the parameters and their order — the break is source-level only, and the
+// generated-code path is not affected at all (this plugin refuses to code
+// generate `convexify_strategy` in the first place).
+//
+// The distinguishing fact is not a version number, it is which of the two
+// names the installed CasADi declares. Nothing in the version macros
+// separates them: the nightly that first carried the rename reports the same
+// `CASADI_MAJOR/MINOR/PATCH` as 3.7.2, and keying on `CASADI_IS_RELEASE`
+// would then misfire on the next release, which will carry the new name with
+// `CASADI_IS_RELEASE=1`. So detect the spelling directly.
+//
+// Both calls below are dependent on the template parameters, so each is
+// looked up at instantiation; the overload naming a helper this CasADi does
+// not declare fails substitution in its trailing return type — the immediate
+// context — and is dropped rather than becoming a hard error. The unused
+// `int` / `long` first parameter exists only to rank the two, so a CasADi
+// declaring both spellings resolves to the current name. Call it as
+// `convexify_eval_compat(0, ...)`.
+//
+// Which overload survives is decided by the CasADi being built against, so
+// on any one build only one of them is ever compiled. That is what
+// `tests/convexify_compat/` is for: it compiles both against mock
+// declarations, so the branch this machine's CasADi does not exercise cannot
+// rot unnoticed.
+//
+// Note for anyone patching this downstream: a `sed` that rewrites
+// `convexify_eval(` to `casadi_convexify_eval(` also rewrites the *fallback*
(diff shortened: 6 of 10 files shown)


srikanth-gm:
closing the loop here - working now, thank you

- confirming commit `cda80f8faa6d951c81b4008cd3cd8184549746a7` builds in our CI without any patches required 
- POUNCE iterating now with both FERAL and MA57
- iterations are being printed out to terminal
- structured iteration info is now available through casadi and populates our internal convergence logging without issue
- next step for us is getting to baseline convergence parity to IPOPT and figure out why the duals are blowing up with POUNCE + LBFGS but not with IPOPT + LBFGS

image


```
Creating NLP solver with 59939 variables and 58304 constraints
CasADi - 2026-08-17 23:44:52 MESSAGE("nlp_solver::init") [.../casadi/core/function_internal.cpp:576]
CasADi - 2026-08-17 23:44:52 MESSAGE("nlp_solver::create_function nlp_f:[x, p]->[f]") [.../casadi/core/oracle_function.cpp:242]
CasADi - 2026-08-17 23:44:52 MESSAGE("nlp_solver::create_function nlp_g:[x, p]->[g]") [.../casadi/core/oracle_function.cpp:242]
CasADi - 2026-08-17 23:44:52 MESSAGE("nlp_solver::create_function nlp_grad_f:[x, p]->[f, grad:f:x]") [.../casadi/core/oracle_function.cpp:242]
CasADi - 2026-08-17 23:44:52 MESSAGE("nlp_solver::create_function nlp_jac_g:[x, p]->[g, jac:g:x]") [.../casadi/core/oracle_function.cpp:242]
Number of nonzeros in inequality constraint Jacobian.:      203

Total number of variables............................:    59939
Total number of equality constraints.................:    58271
Total number of inequality constraints...............:       33
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:       27
        inequality constraints with only upper bounds:        6

iter      objective   inf_pr   inf_du lg(mu)    ||d|| lg(rg) alpha_du alpha_pr  ls
   0  2.2904024e+02 1.06e+01 1.01e+02   -1.0 0.00e+00      - 0.00e+00 0.00e+00   0
   1  2.1123252e+02 9.94e+00 9.45e+01   -1.0 7.14e+00      - 2.23e-01 6.10e-02f  1
   2  2.1089579e+02 9.93e+00 2.86e+02   -1.0 6.93e+00      - 4.66e-02 1.26e-03f  1
   3  2.0186463e+02 9.59e+00 3.91e+03   -1.0 6.91e+00      - 2.63e-01 3.37e-02f  1
   4  2.0177235e+02 9.59e+00 1.18e+04   -1.0 6.62e+00      - 2.56e-02 3.63e-04h  1
   5  2.0152952e+02 9.58e+00 7.91e+05   -1.0 6.61e+00      - 1.54e-01 9.55e-04f  1
   6  1.9360006e+02 9.28e+00 7.68e+05   -1.0 6.61e+00      - 3.09e-02 3.12e-02f  1
   7  1.9353456e+02 9.28e+00 6.78e+05   -1.0 4.60e+02      - 2.42e-03 2.76e-04h  1
   8  1.9352810e+02 9.28e+00 5.93e+05   -1.0 6.17e+02      - 2.05e-03 2.71e-05h  1
   9  1.9325397e+02 9.27e+00 1.14e+06   -1.0 6.24e+02      - 6.47e-02 1.15e-03f  1
iter      objective   inf_pr   inf_du lg(mu)    ||d|| lg(rg) alpha_du alpha_pr  ls
  10  1.9263327e+02 9.24e+00 3.05e+06   -1.0 6.51e+00      - 2.86e-01 2.62e-03f  1
  11  1.8823557e+02 9.07e+00 6.22e+06   -1.0 6.51e+00      - 6.59e-04 1.86e-02f  1
  12  1.8818166e+02 9.07e+00 7.45e+06   -1.0 6.43e+00      - 2.91e-02 2.37e-04h  1
  13  1.8800833e+02 9.06e+00 1.69e+07   -1.0 6.42e+00      - 6.34e-02 7.62e-04h  1
  14  1.8744460e+02 9.04e+00 8.99e+07   -1.0 6.42e+00      - 7.69e-02 2.48e-03f  1
  15  1.8572355e+02 8.97e+00 8.89e+07   -1.0 1.11e+02      - 1.01e-02 7.60e-03f  1
  16  1.8361161e+02 8.89e+00 8.63e+07   -1.0 6.37e+00      - 1.60e-02 9.42e-03f  1
  17  1.8351997e+02 8.88e+00 8.50e+07   -1.0 7.00e+00      - 1.51e-02 4.14e-04h  1
  18  1.8350423e+02 8.88e+00 8.26e+07   -1.0 7.04e+00      - 4.09e-02 7.13e-05h  1
  19  1.8104653e+02 8.78e+00 1.59e+08   -1.0 6.97e+00      - 8.87e-02 1.11e-02f  1
iter      objective   inf_pr   inf_du lg(mu)    ||d|| lg(rg) alpha_du alpha_pr  ls
  20  1.8060472e+02 8.77e+00 2.91e+08   -1.0 7.68e+00      - 6.56e-02 2.10e-03f  1
  21  1.8015400e+02 8.75e+00 5.94e+08   -1.0 7.81e+00      - 1.44e-01 2.17e-03f  1
  22  1.7939849e+02 8.71e+00 5.93e+08   -1.0 2.56e+02      - 4.06e-03 3.66e-03f  1
  23  1.7669665e+02 8.60e+00 6.44e+08   -1.0 3.57e+01      - 3.77e-02 1.33e-02f  1
  24  1.7632750e+02 8.58e+00 6.41e+08   -1.0 1.24e+02      - 8.39e-03 1.94e-03f  1
  25  1.7516833e+02 8.53e+00 6.41e+08   -1.0 1.13e+02      - 2.27e-03 6.13e-03f  1
  ```

This is IPOPT Baseline for first 25 iters in case you're curious

```
Creating NLP solver with 59939 variables and 58304 constraints
CasADi - 2026-08-11 15:00:59 MESSAGE("nlp_solver::init") [.../casadi/core/function_internal.cpp:576]
CasADi - 2026-08-11 15:00:59 MESSAGE("nlp_solver::create_function nlp_f:[x, p]->[f]") [.../casadi/core/oracle_function.cpp:242]
CasADi - 2026-08-11 15:00:59 MESSAGE("nlp_solver::create_function nlp_g:[x, p]->[g]") [.../casadi/core/oracle_function.cpp:242]
CasADi - 2026-08-11 15:00:59 MESSAGE("nlp_solver::create_function nlp_grad_f:[x, p]->[f, grad:f:x]") [.../casadi/core/oracle_function.cpp:242]
CasADi - 2026-08-11 15:00:59 MESSAGE("nlp_solver::create_function nlp_jac_g:[x, p]->[g, jac:g:x]") [.../casadi/core/oracle_function.cpp:242]
There are 59939 variables and 58304 constraints.
Using limited memory Hessian approximation

******************************************************************************
This program contains Ipopt, a library for large-scale nonlinear optimization.
 Ipopt is released as open source code under the Eclipse Public License (EPL).
         For more information visit https://github.com/coin-or/Ipopt
******************************************************************************

This is Ipopt version 3.14.19, running with linear solver ma57.

Number of nonzeros in inequality constraint Jacobian.:      203

Reallocating memory for MA57: lfact (9926263)
Total number of variables............................:    59939
Total number of equality constraints.................:    58271
Total number of inequality constraints...............:       33
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:       27
        inequality constraints with only upper bounds:        6

iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  2.2904024e+02 1.06e+01 1.01e+02  -1.0 0.00e+00    -  0.00e+00 0.00e+00   0
Reallocating memory for MA57: lfact (10474758)
   1  2.1123252e+02 9.94e+00 9.45e+01  -1.0 7.14e+00    -  2.23e-01 6.10e-02f  1
   2  1.6494335e+02 8.36e+00 7.36e+01  -1.5 7.16e+00    -  2.87e-01 2.21e-01f  1
   3  1.6264645e+02 8.15e+00 2.27e+02  -1.5 7.89e+00    -  2.39e-01 2.54e-02h  1
   4  1.4054380e+02 6.85e+00 5.25e+02  -1.5 7.62e+00    -  3.00e-01 2.35e-01f  1
   5  1.3681753e+02 6.44e+00 4.92e+02  -1.5 6.43e+00    -  2.17e-01 6.04e-02f  1
   6  1.3312028e+02 5.85e+00 3.11e+03  -1.5 7.06e+00    -  1.76e-01 9.90e-02f  1
   7  1.3086756e+02 5.69e+00 2.86e+03  -1.5 8.15e+00    -  1.21e-01 9.35e-02h  1
   8  1.2988951e+02 5.31e+00 2.41e+03  -1.5 9.04e+00    -  6.49e-02 6.77e-02h  1
   9  1.2938087e+02 5.00e+00 2.26e+03  -1.5 8.68e+00    -  1.17e-01 5.76e-02h  1
iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
  10  1.2895411e+02 4.52e+00 2.35e+03  -1.5 8.16e+00    -  4.97e-02 9.39e-02h  1
  11  1.2891767e+02 4.14e+00 2.20e+03  -1.5 7.29e+00    -  2.52e-01 8.42e-02h  1
  12  1.2925531e+02 3.40e+00 4.80e+02  -1.5 8.25e+00    -  9.08e-02 1.80e-01h  1
  13  1.2944305e+02 2.84e+00 1.32e+03  -1.5 5.91e+00    -  6.68e-02 1.63e-01h  1
  14  1.2931200e+02 2.37e+00 1.18e+03  -1.5 4.54e+00    -  3.70e-01 1.69e-01h  1
  15  1.2846873e+02 2.42e+00 8.11e+02  -1.5 3.63e+00    -  4.12e-01 4.13e-01h  1
  16  1.2599839e+02 2.15e+00 2.97e+02  -1.5 2.43e+00    -  1.50e-01 5.72e-01h  1
  17  1.0714134e+02 6.15e+00 2.06e+02  -1.5 4.87e+00    -  2.07e-01 4.09e-01f  1
  18  1.0356563e+02 4.98e+00 1.64e+02  -1.5 3.54e+00    -  3.76e-01 1.91e-01f  1
  19  9.4772965e+01 4.39e+00 9.96e+01  -1.5 5.47e+00    -  3.40e-01 3.78e-01f  1
iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
  20  9.2629359e+01 4.68e+00 6.85e+01  -1.5 3.82e+00    -  3.17e-01 2.73e-01f  1
  21  8.9769607e+01 4.85e+00 1.04e+02  -1.5 1.04e+01    -  1.95e-01 2.91e-01f  1
  22  8.8513636e+01 2.46e+01 2.43e+02  -1.5 7.12e+00    -  4.33e-01 3.69e-01h  1
  23  8.8161198e+01 1.71e+01 2.16e+02  -1.5 8.39e+00    -  1.94e-01 1.70e-01h  1
  24  8.9276443e+01 9.18e+00 4.12e+02  -1.5 1.46e+00    -  6.55e-02 4.59e-01h  1
  25  9.0621979e+01 1.14e+00 1.40e+02  -1.5 9.44e-01    -  6.25e-02 1.00e+00h  1
  ```

jkitchin:
Great to hear the build is clean on `cda80f8` without patches — and thanks for the two logs, they were more useful than you probably expected.

We dug into the dual blow-up and think there is a real bug on our side. Sharing the state of it now rather than waiting for the fix, since there's something cheap you can try in the meantime.

## What we think is going on

Your two logs are identical to the last printed digit at iteration 1 and diverge at iteration 2. That timing is diagnostic: our L-BFGS initial Hessian scalar σ is hard-coded to `1.0` while the curvature history is empty, so iteration 1 cannot differ — the first curvature pair lands at iteration 2, and that is exactly where you lose parity.

At that point we appear to pick a different σ than Ipopt does. Ipopt's `limited_memory_initialization` defaults to `scalar1` (σ = sᵀy/sᵀs). We register that option with the same default, but we never actually read it — the updater is pinned to `scalar2` (σ = yᵀy/sᵀy). Setting the option today does nothing, and doesn't warn either, which is the part we're most annoyed about.

σ_scalar2/σ_scalar1 = (yᵀy·sᵀs)/(sᵀy)², which is ≥ 1 and grows without bound as conditioning degrades. On a 60k-variable collocation model that is unlikely to be a small factor. An over-large σI swamps the six rank-2 corrections, the model collapses toward a large multiple of the identity, and the equality multipliers absorb the mismatch — which reads a lot like your `alpha_pr` ~1e-3 and `inf_du` running 1e2 → 6e8.

This is our current best explanation, not a confirmed root cause. We're measuring the trajectory effect across our fixture corpus now.

## The test, if you have a few minutes

`limited_memory_initialization` is inert, but the clamp on σ *is* wired. Capping it approximates the fix:

```python
opts = {"pounce": {"hessian_approximation": "limited-memory",
                   "limited_memory_init_val_max": 1.0}}   # default is 1e8
S = casadi.nlpsol("S", "pounce", nlp, opts)
```

Worth trying `1.0` and `10.0`. If the first 25 iterations move toward your Ipopt baseline, that confirms σ as the cause and tells us the fix is the right one. If nothing changes, our theory is wrong and we'd rather find that out now.

## One thing we can't resolve from the logs

Could you share the exact options dict from **both** runs?

Your Ipopt log shows `lg(mu)` going -1.0 → -1.5 at iteration 2 and then pinned there for 24 iterations. Monotone from `mu_init=0.1` gives `min(0.2·0.1, 0.1^1.5) = 0.02`, which prints as `-1.7`. `-1.5` is exactly `0.1^1.5` — the superlinear term without the min — which isn't the monotone formula. That suggests your Ipopt run is on `mu_strategy=adaptive` (the pinning would be its fixed-mu mode), while POUNCE defaults to `monotone` like stock Ipopt.

If so, the two runs differ on the barrier strategy as well, independently of the σ issue, and matching them would be part of getting to convergence parity. We'd rather confirm that from your options than keep inferring it from a log column.

We'll follow up here once the fixture sweep is in. Changing this default reroutes every limited-memory solve, so it needs a full trajectory sweep before it lands — no ETA yet, but it's being worked.

---

jkitchin:
Following up on the above — we measured it, and the theory holds. Tracking in #677.

Confirmed defect: `limited_memory_initialization` is registered with Ipopt's default (`scalar1`) but has no read site anywhere, and the updater is built with `..Default::default()`, which pins σ to `scalar2`. So every POUNCE limited-memory solve uses σ = yᵀy/sᵀy while Ipopt uses σ = sᵀy/sᵀs, and the option that would change it is inert.

## Sweep across our fixture corpus, scalar2 → scalar1

Default (exact-Hessian) path is bit-identical across all 57 fixtures — this is confined to `hessian_approximation=limited-memory`. Under forced L-BFGS, 20 of 57 move:

| fixture | today (scalar2) | scalar1 |
|---|---|---|
| `jit1` | MaxIter @3000, obj 177485.8 | **Solved, 35 it**, obj 173345.4 |
| `jit1_boxed` | MaxIter @3000, obj 177011.2 | **Solved, 35 it**, obj 173345.4 |
| `jit1_node` | MaxIter @3000, obj 398830.5 | **Solved, 36 it**, obj 173345.4 |
| `pooling_rt2stp` | RestorationFailed, 14 it | **Solved, 413 it**, obj −28.19 → −4391.83 |
| `autocorr_bern55-06` | 2924 it | **106 it**, better objective |
| `csfi2` | 60 it | 32 it |
| `eigenb2` | SolvedToAcceptableLevel | **SolveSucceeded** |

Not uniformly positive — `cresc4` and `deb7` change failure mode, `eigena2` errors 28 iterations later, and `hs13_bigstart` shifts its objective by ~3e-4. Those need explanations before we change the default, which is why this isn't merged yet.

None of that changes the advice: the `limited_memory_init_val_max` clamp from the previous comment is still the thing to try on your model, and your result either way is useful to us. If capping σ moves your first 25 iterations toward your Ipopt baseline, that's independent confirmation on a model far larger than anything in our corpus.

The request for both options dicts still stands — the `lg(mu)` −1.5 question is unresolved and looks like a second, separate discrepancy.

Thanks again for the logs. The two trajectories being digit-identical at iteration 1 is what localised this; without that we'd have been looking at the dual update instead of the Hessian initialisation.

---

PR #603, opened 2026-08-15
588: quadratic structure exploitation — Q0..Q6, measured per phase (Q7–Q9 deferred, Q10 cut)

jkitchin:
## What this is

The implementation series for the quadratic-structure levers in #588, stacked
into one PR: **one phase per commit, each measured before and after**, with the
before/after table posted as a comment on this PR as the phase lands.

Design and evidence: `dev-notes/quadratic-structure-exploitation.md`, added in
the first commit here. It is a research note, so the numbers in it were measured
rather than estimated; where a figure is a forecast it says so, and §0 carries a
status table that each phase updates.

This PR **references** #588 without a closing keyword. #588 is a five-lever
tracking issue and stays open; the individual phases get their own sub-issues.

## The three findings that shape the work

1. **POUNCE already recognizes these problems and then discards the answer.**
   `dispatch.rs` holds an exact quadratic recognizer and an exact LDLᵀ convexity
   certificate. Both run and both succeed on the Mittelmann `qcqp*` family.
   `classify_problem` then returns `ProblemClass::Nlp` and the solver re-derives
   the same constant matrices through a colored AD tape on every iteration —
   332 times on `qcqp500-3c`.
2. **The family splits into three regimes needing three different levers.** A
   single "constant-Hessian" change does not fix it: `qcqp500-3c`/`qcqp750-2c`
   are ~90% evaluation, `qcqp1000-*`/`qcqp1500-*` are 69–81% factorization, and
   `qssp180`/`nql180` are 92% factorization on trivially sparse rows.
3. **The gap is not to IPOPT.** POUNCE beats IPOPT+MA57 on 8 of the 9 `qcqp`
   instances, twice by timeout. The gap is to COPT and KNITRO, and closing these
   levers is worth roughly **1.7× on overall Mittelmann SGM** — the largest
   concentrated win available, and not a path to KNITRO's 1.17. That bound is
   stated up front so no phase has to oversell itself later.

## Measurement protocol

`scripts/qbench.py` (added in the first commit) drives the instruments that
already exist — `--json-output`, `timing_statistics=yes
print_timing_statistics=yes`, the printed `Problem class:` line, and per-child
peak RSS — the same way every time, so a phase's before/after is a comparison
rather than two differently-taken readings.

Each iteration:

1. `qbench.py run <parent-commit binary> before.json --set <set>`
2. implement the phase
3. `qbench.py run <new binary> after.json --set <set>`
[...]

Issue #588, opened 2026-08-14
Future performance levers


jkitchin:
pounce is the only non-commercial solver at https://plato.asu.edu/ftp/ampl-nlp.html solving all 47 problems!

The Mittelmann benchmark contains enough quadratic and quadratically constrained models that adding:

- exact quadratic-expression recognition;
- constant-Hessian treatment;
- convexity classification;
- a dedicated QP/QCQP barrier path;
- specialized presolve and scaling;

might enhance performance further.  This is a tracking issue to work on this over time. We have some auto-routing to QP already, but no QCQP barrier path. There is also some presolve / scaling work, but presumably there are other things to do.

jkitchin:
Closing this. Every lever named in the opening comment has been either shipped
and measured, or cut on evidence. The durable record is
`dev-notes/quadratic-structure-exploitation.md`; the close-out is §0j, added
in #704.

## What the issue asked for, and what happened to each

| lever from the opening comment | outcome |
|---|---|
| exact quadratic-expression recognition | **shipped** — Q3 (recognizer to `pounce-nl`, iterative, allocation-free; an `o0` chain went `O(k²)` → linear, 0.77 s → 0.020 s at k=16 000) and Q5 (recognized at parse time, tree never built) |
| constant-Hessian treatment | **shipped** — Q4 (`eval_h` on `qcqp1000-2c` 19.58 s → 0.20 s, ~95x) and Q6 (the four constant-derivative hints, now *proved* rather than trusted; `∇f` 1729 → 1161 and `∇g` 2005 → 1573 over 57 fixtures at bit-identical trajectories) |
| convexity classification | **shipped** — Q1/Q2, with the header nonlinearity census kept and deliberately **not** routed on |
| a dedicated QP/QCQP barrier path | **cut** — Q8; see below |
| specialized presolve and scaling | **shipped in part** — Q9b wired `presolve_conic` into `run_convex_socp` (it had been registered, documented and silently ignored on every convex QCQP) and pinned a real `dedup_rows` hazard with a 64-instance differential battery. The equilibration half is #703 |
| (added during the work) Gondzio correctors | **shipped** — Q9a; 52 NETLIB LPs, 688 → 644 iterations with the KKT error improving |
| (added during the work) dense-front factorization | **cut** — Q10, by its own control experiment: feral beats MA57 on 2 of 3, and on `qcqp1500-1c` MA57 cannot finish a problem feral solves in under three minutes |

## Why the QCQP barrier path is cut

[...]

PR #671, opened 2026-08-18
Q3 review copy: move the quadratic recognizer to `pounce-nl` and stop it recursing

jkitchin:
**Review-only PR. Do not merge.** This exists so Q3 of the #588 series can be
read as a single-phase diff. The real series PR is #603; its branch
`feat/588-quadratic-structure` already contains this commit. The base here is
pinned at Q2 (`688ccd45`) so the diff below is *exactly* Q3 — commit
`fd738d46` — and nothing else.

## Problem

Two distinct defects in the degree-≤2 recognizer, both in `dispatch.rs`.

**Quadratic blowup.** The merge-then-prune fix already in the tree covered
`Expr::Sum` only. The identical `O(N²)` was still live on binary `Add` chains —
the exact shape the design note cites as the motivating case. Measured on a
left-deep `o0` chain with an 8 MB stack:

| left-deep `o0` chain | before | after |
|---|---|---|
| k = 16 000 | 0.7697 s | 0.0197 s (**39×**) |
| k = 24 000 | **stack-overflow abort** | 0.0254 s |
| k = 1 000 000 | abort | 1.1030 s |

Before is `O(k²)`; after is linear. Dense `(Σxᵢ)²` row at n = 500: 0.3028 →
0.1270 s. Whole 62-file corpus classification: 4.61–5.06 ms → 3.75–3.96 ms.

**Recursion.** `to_poly` recursed over the `Expr` DAG, so a left-deep `Add` tree
of squares — a legitimate model shape — aborted the process during
classification. Q1 reproduced this and pinned its test at `k = 1000` with a
comment saying the limit was the recursion, not the cost model.

## Cause

`Poly` keyed monomials on `BTreeMap<Vec<usize>, f64>` — one heap allocation per
monomial. Legacy cost is now measured and pinned rather than estimated:
**`n² + 2n` monomial `Vec`s per expansion** (251 000 at n = 500). Merging two
polynomials rescanned the whole accumulator for zeros on every merge, which is
where the quadratic term came from.

## Fix

`Quad2` replaces `Poly`: three fields with inline keys, so degree is a property
of the type rather than something re-derived from a map. `recognize_expr` is
iterative (explicit work and value stacks). The per-merge zero scan is now
proportional to what was merged, folding the smaller side into the larger in
either direction.

The algebra moved to `crates/pounce-nl/src/nl_quadratic.rs` — the crate that
owns the `Expr` DAG. `dispatch` keeps routing and re-exports
`analyze_quadratic{,_full}` / `is_trivially_zero`, so no call site in
`qp_extract.rs` or in the tests moved; only the already-unused `n` argument was
dropped. `header_census` deliberately stayed in `dispatch` — it is a classifier
diagnostic about the `.nl` header, not part of the algebra.

**Three things the design note got wrong, corrected in place:**

1. The `O(N²)` was never fully fixed — only the `Sum` half. The
   `dispatch.rs:597` reference is also stale; the code was at line 800. (The
   "2.32 M `Vec`s" figure *is* corroborated.)
2. `pounce-py/src/nl_problem.rs:654` does **not** implement an iterative `Expr`
   drop. It moves a *recursive* drop onto a 64 MB worker thread plus a
   `MAX_DEPTH = 10 000` refusal. The cited precedent is a bigger stack, not a
   different algorithm.
3. Q1's crash is real but narrower than recorded: on the CLI's `.nl` route the
   **parser** overflows first (k ≈ 6 000, versus the recognizer's k ≈ 24 000 on
   the same 8 MB thread), identically before and after. A deep `.nl` file did
   not load before this change and still does not.

## Blast radius

Bit-identical except for the targeted case.

- `scripts/sweep-fixtures.sh` against baseline `688ccd45`: **empty diff**, 57 models.
- `Problem class:` line over all 62 `.nl` files under `crates/`: **empty diff**.
- Differential oracle: **bitwise identical** on 1 631 expressions from the
  corpus plus a 4 000-sample random battery.

`$POUNCE_BENCH_DATA` is not present on this machine, so no `quad` / `control`
set numbers are claimed here — stated in the commit and the note as well. The
timings above are from the recognizer microbenchmarks and the in-repo fixture
corpus, which are what this phase's claims actually rest on.

## Tests

- 10 unit tests in `nl_quadratic`, including three at **depth 250 000**
  (left-deep, right-deep, and early-bail). These abort on the parent commit, for
  the stated reason — stack overflow in the recursive walk.
- `crates/pounce-cli/tests/quad_recognizer_differential.rs` keeps the pre-Q3
  recognizer verbatim as an oracle and compares bitwise. This is the pattern Q1's
  "2-ulp line" argues for: a coefficient differing by one ulp moved a fixture
  from 17 to 12 conic iterations, and only a differential check sees that.
- Q1's `k = 1000` pin raised to **100 000**.

Deliberately not pinned: the `.nl` **parser's** recursion ceiling (k ≈ 6 000).
It is unchanged by this phase and is now the lowest ceiling on deep models —
recorded as a known gap rather than silently fixed here.

`cargo test --workspace --exclude pounce-hsl`: 2974 passed, 0 failed.
`cargo fmt --check` clean. `cargo clippy --workspace --exclude pounce-hsl
--all-targets` exit 0, no new lint class.

---

- [ ] `CHANGELOG.md` `[Unreleased]` entry, in the user's terms
      as it stands now

The two unticked boxes are deliberate and are properties of the series, not
oversights: Q3 is an internal refactor with no user-visible behaviour change, so
there is nothing to say in `CHANGELOG.md` in the user's terms and no book page
to update. The user-facing entry belongs to the series as a whole and lands with
#603.

Refs #588

---

### DIFF
--- a/crates/pounce-cli/src/dispatch.rs
+++ b/crates/pounce-cli/src/dispatch.rs
@@ -35,9 +35,8 @@
 //! (PSD) test uses a tolerance and routes "inconclusive within
 //! tolerance" to the safe side, never to the convex path.
 
-use crate::nl_reader::{BinOp, Expr, NlProblem, UnaryOp};
+use crate::nl_reader::NlProblem;
 use pounce_common::types::{lower_bound_present, upper_bound_present};
-use std::collections::BTreeMap;
 
 /// Tolerance for the smallest-eigenvalue sign test in the convexity
 /// check. A Hessian eigenvalue below `-PSD_TOL` is treated as a genuine
@@ -381,7 +380,7 @@ fn classify_inner(prob: &NlProblem) -> (ProblemClass, ClassReason) {
     }
 
     // Objective curvature.
-    let obj_quad = match analyze_quadratic(&prob.obj_nonlinear, prob.n) {
+    let obj_quad = match analyze_quadratic(&prob.obj_nonlinear) {
         Some(q) => q,
         // Objective has a non-quadratic nonlinear term ⇒ NLP.
         None => return (ProblemClass::Nlp, ClassReason::ObjectiveNotQuadratic),
@@ -394,7 +393,7 @@ fn classify_inner(prob: &NlProblem) -> (ProblemClass, ClassReason) {
         if is_trivially_zero(c) {
             continue;
         }
-        match analyze_quadratic(c, prob.n) {
+        match analyze_quadratic(c) {
             Some(q) if q.is_empty() => {} // purely linear after all
             Some(_) => any_quadratic_constraint = true,
             None => {
@@ -445,7 +444,7 @@ fn classify_inner(prob: &NlProblem) -> (ProblemClass, ClassReason) {
             if is_trivially_zero(c) {
                 continue;
             }
-            match analyze_quadratic(c, prob.n) {
+            match analyze_quadratic(c) {
                 Some(q) if q.is_empty() => {} // purely linear after all
                 Some(q) => {
                     let lo = prob.g_l[row];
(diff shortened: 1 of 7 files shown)


jkitchin:

   produced.

Follow the series there.

PR #672, opened 2026-08-18
sens: degeneracy="directional" decides a kink for the perturbation's own direction (roadmap item 3)

devin-griff:
## What this adds

A solve can converge with a bound weakly active, the variable on the bound with a multiplier of the same order as the slack, both of order the square root of the barrier parameter. The solution has a kink there, a different one-sided derivative on each side, and no single linear step is right for both. The activity thresholds have no answer at a kink because the two quantities they compare are the same size, and the held factorization carries the bound as `sigma = z / s` on the variable's diagonal, which at a kink is of order one: a bound neither enforced nor absent, wrong for both sides.

`estimate()`, `estimate_report()`, and `active_set_changes()` gain `degeneracy`, with values `"directional"` (the default) and `"one_sided"`. The default decides each weakly active bound for the perturbation's own direction by the directional-derivative QP, the sIPOPT paper's eq. 14, solved as an active-set search over the weakly active rows on the held factorization: the rows are released, removing the order-one sigma, a candidate working set pins their variables through Schur rows, and a candidate is accepted when every variable left out moves into its feasible side and every pin is necessary, its removal alone making its variable violate. The necessity probe stands in for the dual-feasibility sign test, whose Schur sign convention this code never names, and agrees with it wherever the returned direction differs. `"one_sided"` is bit-identical to the current release.

All three modes consume the decision. `linear` takes the QP direction itself. `fix_relax` takes it as the predictor its refinement iterates from. `path` starts with the held rows pinned through Schur holds, droppable later like any other, and the rows the QP leaves forced into its base-activity table, so the release scan frees each at the fraction where its multiplier reaches zero, which the record then carries.

`gradient()` is asked for a derivative without a direction, so at a degenerate base point it warns, naming the variables and bounds, and returns the one-sided value. The check is cached per session.

## The inside-band case, found by measurement during development

Held slightly on the active side of a breakpoint, still classified ambiguous, a bound is genuinely active with a small positive multiplier, and the true solution releases it partway through the step. An earlier revision decided those rows at fraction zero: on the CSTR held at 75% of the record's first breakpoint fraction, that released a sixth of a step early and overshot to 0.0207 where the walk's own release reaches 0.0020. The shipped design hands those rows to the walk's machinery instead, and the fraction-zero decision survives only where it is correct, at an exact kink, as the limit of the multiplier's zero crossing.

## Measured

- On the CSTR held at the record's first breakpoint, a 2% step toward the steady state: under `one_sided` the thresholds classify the kink bound inactive, `linear` and `path` miss by 0.0077 with an empty record, and `fix_relax` reaches 0.0018 through its own sign-based release test, a favorable read at a point where the true multiplier is exactly zero. Under `directional` all three modes land at 0.0018, with fix_relax's number now guaranteed rather than read, and the record carries the departure at fraction 0.008, the zero crossing of the residual multiplier the solve left.
- All thirteen breakpoints swept: the first is the only one whose threshold call lands on the wrong side at this tolerance, the ratio there is 4.2x, and the other degenerate breakpoints tie at 1.0x, so the correction costs nothing where the thresholds happen to be right.
- A clean base point takes the plain call with zero trials, bit-identical under both settings.

## Tests

- `released_direction.rs` gains a fixture made degenerate by construction, `b0 = G * A0` putting a variable's unconstrained minimizer exactly on its bound: detection finds exactly that bound, both one-sided directions match hand algebra, [1.2270, 0.4536] free and [-1.10, 0] held with the pin proven necessary, and a strongly active base point spends zero trials.
- `test_degeneracy.py`, nine tests on the canonical kink, minimize (x minus p) squared with x nonnegative: both one-sided derivatives, 1 and 0, right in every mode, `one_sided` provably wrong on at least one side since a single linear map cannot produce both, the record's essentially-zero departure, the inside-band release at the multiplier's own zero crossing, strictly positive, which discriminates against the stamped zero, the `gradient()` warning firing at the kink and not at a clean point, bit-identity at clean points, the budget fallback with its warning, value validation, and a coupled model verified against re-solves on both sides.
- `estimate_report` keeps its contract of measuring the step `estimate()` takes, and an existing test now demonstrates it from both sides: the one-sided step predicts a variable a quarter past a bound it is weakly held at, and the directional default holds it there with nothing crossing.

## Docs and notebook

`docs/src/sensitivity.md` gains the `degeneracy` section, and the CHANGELOG entry carries the numbers. Notebook 36 gains section 8, case E: the CSTR held at the record's first breakpoint, the classifier's ambiguous label and the gradient warning as evidence, and the two settings side by side on the v1 kink intervals, with the linear line drawn widest and path on top since the two coincide under `one_sided`.

Full suites green: 14 rust sensitivity test files, clippy clean, and the pyomo-pounce suite at 409 passed.


jkitchin (changes requested):
The core idea is right and well argued: the eq. 14 QP is the correct answer to a kink, releasing every weakly active row before the search is the right handling of the order-one `sigma`, and the necessity probe is a clean way to get the dual-feasibility test without naming the Schur sign convention. The "inside-band case, found by measurement during development" section is exactly the kind of self-correction that belongs in a PR body.

Four findings, ordered by severity.

---

## 1. `weakly_active_bounds` mixes full-x and var-x index spaces — wrong variable, not a missed one

`crates/pounce-sensitivity/src/solver.rs:879`

```rust
for (i, &st) in report.var_status.iter().enumerate() {
    ...
    let s_lo = ctx.x_curr[i] - ctx.lo[i];                          // var-x
    if let Some(br) = rows.iter().find(|b| b.var_row == i && ...)  // var-x
```

`report.var_status` is **full-x** — `activity.rs:582` scatters through `nl.var_x_to_full_x` into a `vec![fixed_entry; n_full_x]`. `ctx.x_curr` / `lo` / `hi` are **var-x** (`solver.rs:1011`, `x_curr: state.x[..n_x]`), and `bound_context` says so itself: "the two index spaces diverge from the first fixed variable on."

The doc comment at `solver.rs:854` claims this can only under-report:

> a status past one cannot produce an entry here, only miss one

That does not hold. Concretely — variables `[v0, v1(fixed), v2, v3]`, so `n_full_x = 4`, `n_x = 3`, var-x rows `0 -> v0`, `1 -> v2`, `2 -> v3`. If `v2` is `AMBIGUOUS`, its full-x index is `i = 2`. The `i >= ctx.n_x` guard passes (2 < 3). The lookup then finds the bound row for `var_row == 2`, which is **v3**, and the `lower` side is chosen from `ctx.lo[2]` / `ctx.hi[2]` — also v3's. So a `WeakBound` is emitted for the wrong variable, and that is the one that gets pinned or released. `v3`'s own status at `i = 3` is then skipped by the guard.

This is the failure `Session.primal_row` was written to prevent, in a docstring added after gh#450 (`pyomo-pounce/pyomo_pounce/sens.py:347`):

> one fixed variable shifts every later column and the back-solve quietly returns a NEIGHBOURING variable's sensitivity, a plausible number with nothing wrong-looking about it. Route every factor index through here.

`covariance()` keeps that discipline explicitly (`sens.py:2683`: "Keep the two apart — they differ exactly when the model has a fixed variable"). This path does not.

It is not hypothetical: `declare_sens_param` documents **fixed Vars** as accepted input (`sens.py:109`). The Python wrapper compounds it — `_weakly_active` maps the returned `var_row` through `full_of` as if it were a var-x row, so the `gradient()` warning can name a third variable.

Nothing covers it in tests. `primal_row`'s own docstring notes the two spaces coincide "in every model in the test suite."

Suggested fix: route the classifier index through the `var_x_to_full_x` inverse before using it as a factor row.

## 2. `directional_step` builds `2^n` candidates before consulting `max_iter`

`crates/pounce-sensitivity/src/boundcheck.rs:957`

```rust
let mut candidates: Vec<u32> = (0u32..(1 << n)).collect();
candidates.sort_by_key(|s| (s.count_ones(), *s));
```

`n = weak.len()` is uncapped. The budget check lives inside the loop, so the allocation and sort happen unconditionally — and `max_iter` defaults to 16, so at most 16 subsets can ever be tried however many are enumerated.

I confirmed the overflow behaviour with a runtime-opaque `n`:

- `n = 32`, release: `1u32 << 32 == 1` (the shift is masked) -> one candidate, the empty set. If it is not accepted the result is a spurious `"no sign-consistent working set"`, and the Python layer warns and falls back to one-sided.
- `n = 32`, debug: `attempt to shift left with overflow` panic.
- `n` around 25-31: 128 MB to 4 GB allocated and sorted, in order to try 16 of them.

Generating subsets lazily in size-then-least-index order — or capping `n` and erroring cleanly above it — costs nothing and removes all three.

## 3. Two docstrings describe the revision this PR replaced

`crates/pounce-sensitivity/src/solver.rs:743` and the matching `parametric_step_path_directional` docstring in `crates/pounce-py/src/solver.rs` both say out rows are "recorded as departures at fraction zero, which is where the kink resolves."

That is the earlier behaviour the PR body's "inside-band case" section says was wrong and replaced — the one that overshot to 0.0207 against 0.0020. The inline comment ten lines below the Rust docstring already states the shipped behaviour correctly (forced into the base-activity table, released at the multiplier's zero crossing). Given how carefully the PR body draws that distinction, the stale docstrings are the version a reader is most likely to trust.

## 4. `EPS = 1e-9` is an absolute tolerance on a quantity that scales with the perturbation

`crates/pounce-sensitivity/src/boundcheck.rs:948`. The comment justifies it for pinned rows — "zero movement of a pinned variable is exact up to the augmented solve's roundoff" — which is fair. But the same constant is applied through `feasible` / `violates` to **out** rows, whose `d[var_row]` scales linearly with `deltas` via `parametric_rhs_full`.

The QP's answer should be invariant to the perturbation's magnitude; this test is not. On the `kink()` fixture, `estimate(m, [(m.p, -1.0)])` correctly holds the bound, but `estimate(m, [(m.p, -1e-10)])` gives `d = -1e-10`, which clears `feasible` (`>= -1e-9`) and fails `violates` — so the all-released empty set is accepted and the holding side's derivative reads -1 instead of 0. `clamp=True` hides it in `linear`, but not with `clamp=False`, and not in the modes that consume the direction downstream.

A relative test — against `norm(d, inf)` or the perturbation norm — would make the decision scale-free.

---

**Summary.** #1 is the one worth blocking on: it is silent, it produces a plausible wrong number, and it re-opens a hazard the codebase already paid for once in gh#450. #2 is cheap to fix and removes a debug-panic path. #3 and #4 are smaller, but both touch claims the PR makes about its own behaviour.

devin-griff:
All four addressed. Findings 1 through 3 are in c6810deb, finding 4's downstream half in 7e7be99a.

**1.** The loop now walks the var-x rows and reads each status through `var_x_to_full_x`, the same map the classifier scatters through, and the Python wrapper becomes correct as a consequence since the binding now returns true var-x rows. The new test builds your arrangement, a variable fixed by equal bounds ahead of the kink and a spectator behind it, asserts the spaces actually diverge, and checks the warning names x and the spectator is untouched on both sides. One detail worth recording: the NL writer orders columns by expression appearance, so the fixture puts the fixed term first in the objective, and with it anywhere later the divergence sits behind every live variable and the unmapped index passes by accident, which is how the first version of the test failed to fail.

**2.** Working sets are generated lazily in size-then-least-index order through a plain lexicographic next-combination step. Nothing is materialized, the budget bounds the work, there is no shift to overflow, and a unit test pins the order, which is also the determinism promise.

**3.** Both docstrings now state the shipped semantics, rows the working set leaves forced into the base-activity table and released where each multiplier reaches zero.

**4.** Two parts, since your downstream remark was right. The acceptance tolerance is now relative to the trial direction's infinity norm, probes included. Extending your 1e-10 case across all three modes then caught a second scale defect in path: the barrier correction on a weak row is constant-sized, and with the row carried in the walk's table rather than released, it folded into the variable's equation through the order-one sigma as an offset of order sqrt(mu) that dominated tiny steps. The path wrapper now zeroes the weak rows' rhs entries, the same zeroing `solve_released` applies to released rows. The test pins the holding side exact in every mode and linear and fix_relax exact on the releasing side at 1e-10. For path's releasing side it pins boundedness by the step rather than the step itself, deliberately: the release fraction is a quotient of solve quantities, and at a weakly active bound the multiplier's uncertainty equals its magnitude, so a step below that precision cannot be resolved against it. That is the same degeneracy this PR is about, met at the record's precision floor rather than the derivative's.

Suites: 14 rust sensitivity test files, 411 pyomo-pounce, clippy and flake8 clean. The notebook's section 8 numbers are unchanged by the barrier fix, the offset being invisible at four decimals against a 2% step.

jkitchin (approved):
All four findings verified fixed on `a4c60f90`, checked against the code rather than the summary.

**1.** `weakly_active_bounds` now walks the var-x rows and reads status through `var_x_to_full_x`, so `x_curr`/`lo`/`hi` and the `bound_rows` lookup are all in one space and `WeakBound.var_row` is a genuine factor row; the Python wrapper's inverse map is consistent with it. The new fixture is a real guard, not decoration — traced under the old code, the classifier's `i = 1` for `x` finds the bound row for var-row 1, which is `y`, upper side, so `pytest.warns(match=r"x \(lower\)")` fails. The `block_dims[0] < len(var_names)` assertion is the right addition: it is what stops the fixture from silently not exercising the hazard.

**2.** Lazy `next_combination`, no `1 << n`, nothing materialized, and the budget is consulted before every solve including inside the necessity probe. Empty combo terminates, so `size = 0` is well behaved. The order test pins all 16 subsets for n = 4 and the exact size-2 sequence.

**3.** Both docstrings now describe the shipped semantics.

**4.** Tolerance is relative to the trial direction's infinity norm with a floor, applied to both the acceptance test and the removal probe. The downstream half is the better half: zeroing the weak rows' rhs entries is the same argument `solve_released_inner` already makes for released rows, and the index space matches — `released` indexes that same vector. It is scoped to `parametric_step_path_directional` with a non-empty `weak`, i.e. path mode at a degenerate point, which is code this PR introduces, so it cannot move a pre-existing result.

One nit, not blocking and fine to leave: `for w in &weak { rhs_plain[w.row] = 0.0; }` has no bounds guard, where the sibling zeroing writes `if r < scaled.len()`. `w.row` comes from `bound_rows()` so it should be in range, but the defensive form is there for a reason and this one panics rather than degrading.

Issue #673, opened 2026-08-18
Constant-structure evaluation skips factored quadratics, because expanding them cancels


jkitchin:
Follow-up from #588 Q4 (PR #603).

## What Q4 does

`NlTnlp` now evaluates a recognized degree-≤2 objective or row from its constant matrix (`pounce_nlp::quadratic::QuadraticStructure`) instead of from an AD tape. On the real Mittelmann `qcqp1000-2c` that takes `LagrangianHessianEvaluations` from 19.582 s to 0.209 s.

## Why it is gated, and what the gate costs

The fast path evaluates the **expanded** polynomial `½xᵀHx + aᵀx + c`. The tape evaluates whatever the `.nl` writer wrote — and for a least-squares-shaped model that is the **factored** form. Those are not the same computation:

* `crates/pounce-cli/tests/fixtures/feasible_x0_extreme_row.nl` writes `(x − 500000)²`. The tape squares a small residual; the expansion adds `x² − 10⁶x + 2.5·10¹¹` and cancels. Measured disagreement: **2.4e-5 relative** — five digits, not an ulp.
* `airport.nl` is 84 coordinates around 10³ with every row a squared distance. With the fast path ungated, `mu_strategy=adaptive tol=1e-12` went from stopping at a tiny step in **16 iterations to grinding out the 300-iteration cap** at the same objective to 12 digits. That is #544's failure mode exactly: the right answer, slowly. It was caught by #512's regression test, not by `scripts/sweep-fixtures.sh` (which runs default options).

So Q4 ships `pounce_nl::nl_quadratic::is_expanded_quadratic`: a form is admitted only when it is **already a flat sum of monomials**, so reading it back repeats the additions the writer already wrote and can differ from the tape only by summation order. AMPL emits the whole `qcqp*` family that way, so the target of the phase is unaffected — but `airport.nl` keeps 41 of its 42 rows on the tape, and so does every model written as a sum of squared residuals.

That makes the design note's "pays on all 358 structured problems" (`dev-notes/quadratic-structure-exploitation.md` §3.5) **too strong for Q4 as shipped**: it pays on the structured problems whose quadratic parts arrive expanded.

## What would fix it

Evaluate a factored form stably from stored coefficients — i.e. carry a shift rather than expand to the origin. Sketches, in rough order of cost:

[...]

PR #674, opened 2026-08-18
Q4 review copy: evaluate recognized quadratics from their constant matrices

jkitchin:
**Review-only — do not merge.** This exists so Q4 can be read as one diff against its parent (`fd738d46`, the Q3 commit) instead of as the fourth commit of a stack. The series PR is #603; the commit lands there and nowhere else. Both branches here are throwaway.

## Problem

POUNCE recognizes these rows exactly and then throws the answer away. `classify_problem` says "quadratic"; `NlTnlp` goes on to re-derive the same constant matrices through a coloured AD tape on every iteration. On the real Mittelmann `qcqp1000-2c` that is 353 736 summand tapes, 2 445 552 tape ops and 1 000 Hessian colours, walked 101 times to recover eight matrices that never change.

Measured on `qcqp1000-2c` (**a real Mittelmann instance** — the only one available on this machine; `scripts/qbench.py`, same options both sides):

| | status | objective | iters | wall | peak RSS |
|---|---|---|---|---|---|
| before (`fd738d46`) | SolveSucceeded | 738127.4263171845 | 100 | 131.38 s | 786.6 MB |
| after | SolveSucceeded | 738127.4263173543 | 100 | **103.26 / 97.56 s** | **542.7 / 492.7 MB** |

Two after-runs: the phase was measured once at the first green build and again after the shared-DAG fix below, which does not touch this instance's path. They differ by 5.5% on wall and 5.2% on the *factorization* timer doing identical work — this machine's noise — while the iteration count and objective are identical to the last digit across all three runs.

| timer | before | after | |
|---|---|---|---|
| `LagrangianHessianEvaluations` | 19.582 s | **0.209 / 0.198 s** | ~95× |
| `ConstraintEvaluations` | 4.863 s | 0.444 / 0.429 s | ~11× |
| `ConstraintJacobianEvaluations` | 2.055 s | 0.167 / 0.164 s | ~12× |
| `LinearSystemFactorization` | 97.168 s | 97.363 / 92.295 s | untouched |

The note forecast ≈1.4× here. Evaluation is 21.5% of `OverallAlgorithm` on the before-run, so the Amdahl ceiling against that run is **1.26×** — 1.4× was not reachable from the evaluation side at this machine's factorization cost, and Q4 removed 97% of the evaluation time, which is the whole of it. Read this phase off the timer split and the counters (tape ops 2 445 552 → 5 100, colours 1 000 → 0), not off a wall-clock total.

`qcqp500-3c` (≈6× forecast) and `qcqp1500-1c` (≈1.2×) are **not** on this machine and are **not verified**. A *generated* look-alike of the `qcqp500-3c` shape — which does reproduce the evaluation-bound regime (96.7% evaluation, 3855 MB peak) — goes 136.0 → 18.4 s wall and 120.3 → 8.0 s of `OverallAlgorithm` at an unchanged 75 iterations. **That is a generated-instance number and is not a `qcqp500-3c` number.**

## Cause

`NlTnlp::try_new` splits every nonlinear row into summands and builds one `Tape` per summand, unconditionally. For a quadratic row that tape encodes a matrix that is fully determined at load: `eval_h` then runs a forward sweep plus one directional product per colour per tape, per call, to recover constants, and the colouring is sized by a dense quadratic block whose entries no directional product needs to touch.

## Fix

**`pounce_nlp::quadratic::QuadraticStructure`** keeps the recognizer's answer. Each form is the full symmetric `H` in CSR over its own support, the linear and constant terms AMPL folds into the nonlinear tree, and one precomputed scatter slot per lower-triangle entry — so `∇²L` is a `values[slot] += w·h` loop with no hashing and no search. Every array is flat and shared across forms rather than per-form: `qssp180` has 65 341 quadratic rows of three nonzeros each, and a struct-per-row layout would spend ten allocations on each of them to describe 24 bytes of matrix. It lives in `pounce-nlp` because every other crate already depends on it, so `pounce-py`, `pounce-wasm` and `pounce-rs` inherit the type with no manifest change.

A recognized row **builds no tape** — recognition runs before taping, which is where the memory goes; evaluating a tape faster was never the lever. The colouring is built over the tape entries alone (`h_tape_mask`), because a quadratic block's entries are scattered from stored values and never read out of a directional product: 1000 colours → 0 on the real instance, with the assembled Hessian pattern identical either way.

**The fast path is gated twice, and both gates were found by measurement rather than designed in.**

*Accuracy.* **Rejected: taking every recognized form.** That is what the first version did, and it is wrong. The fast path evaluates the **expanded** polynomial while the tape evaluates whatever the writer wrote, and for a least-squares shape that is the **factored** form:

* `feasible_x0_extreme_row.nl` writes `(x − 500000)²`; the expansion adds `x² − 10⁶x + 2.5·10¹¹` and cancels. Measured **2.4e-5 relative** — five digits, not an ulp.
* `airport.nl` (84 coordinates around 10³, every row a squared distance): `mu_strategy=adaptive tol=1e-12` went from stopping at a tiny step in **16 iterations to the 300-iteration cap** at the same objective to 12 digits. gh#544's failure mode exactly.

**Rejected: a magnitude heuristic** (cap `(|c| + Σ|aᵢ|) / max|Hᵢⱼ|`). It catches `(x − 500000)²` and misses `airport.nl` entirely, because there the cancellation is between the quadratic terms themselves and the folded part is zero. Shipped instead: `nl_quadratic::is_expanded_quadratic`, an **exactness** rule — a form is admitted only when it is already a flat sum of monomials, so the read-out repeats the additions the writer already wrote and can differ from the tape only by summation order. AMPL emits the whole `qcqp*` family that way (all 7 rows and the objective of the real `qcqp1000-2c` take the fast path); `airport.nl` keeps 41 of its 42 rows on the tape.

The cost of that gate is real: a model written in factored form gains nothing from Q4, which makes the design note's "pays on all 358 structured problems" too strong. Filed as #673 with three sketches for fixing it.

*Cost.* The gate — and `recognize_expr` behind it — inlines a `Cse` body at **every reference**, which Q3 documented and deliberately left to Q5 to memoize. Putting both on the `try_new` path made a shared DAG `Θ(2^depth)`: `nl_reader::tests::shared_dag_walks_are_memoized_not_exponential` measured **0.00 s → 172.41 s** in a debug build and **still passed**, exponential-but-finite at depth 30. So a `Cse` body reached a second time ends the walk in a refusal and the model keeps its tape, which handles sharing properly. **Rejected: memoizing on `Arc` identity** — it fixes this walk and leaves `recognize_expr` exponential behind it, so it stays whole in Q5. Measured cost of the refusal on this corpus: none (27 models and 251 quadratic rows either way, byte-identical differential numbers; the `qcqp*` files declare no common expressions at all).

`POUNCE_DBG_NO_QUAD=1` forces the tape everywhere (mirroring `POUNCE_DBG_NO_HYBRID`), and `NlTnlp::try_new_with_quadratic(prob, false)` is the same switch as a parameter — the env var is process-global, which is wrong for a differential test that has to build the same model both ways in one process.

## Blast radius

**Exactly the recognized set, and that is asserted rather than assumed**: a model with nothing recognized is byte-identical on both constructions (same Hessian pattern, bitwise-equal `eval_g`), pinned as its own test.

The design note's list of five sites to change is missing one, and it is the dangerous one: **`hessian_vector_products`**, the matrix-free path, also has to add the quadratic block. A Newton-Krylov solve that silently dropped the quadratic part of `∇²L` would still converge, elsewhere. `eval_f`/`eval_grad_f` are two more, for a quadratic objective.

`scripts/sweep-fixtures.sh` over 57 models against `fd738d46`, **one line moves**:

```
-infeasible_equalities  InfeasibleProblemDetected  it=28  obj=1.179054839
+infeasible_equalities  InfeasibleProblemDetected  it=28  obj=1.201505736
```

`infeasible_equalities.nl` is `min x₀² + x₁²` subject to two contradictory cubic equalities: quadratic objective on the fast path, cubic rows on the tape. Measured on that model directly — `eval_f` and `eval_grad_f` bit-identical, `eval_h` differing by 1–2 ulp, and its Lagrangian diagonal cancelling to ~1e-16 (`0.9·2 + 0.7·3 − 1.3·3 = 0`) so a 1-ulp difference is 100% of what the regularization sees. **Status and iteration count are unchanged (28 → 28)**; what moved is the objective at the point where infeasibility was *declared*, a terminal iterate's value and not a converged quantity. `Problem class:` over all 62 `.nl` files: diff empty.

One more, on a diagnostic configuration. `eigena2` is gh#540's fixture, and the same 1 ulp moves its guard test: with `feral_inertia_pivot_floor=0` (the pre-#540 route) it now converges — SolvedToAcceptableLevel/29 it/dual 1.841e-7 → SolveSucceeded/27 it/dual 5.964e-10 — while **the shipped route does not move at all** (SolveSucceeded, 26 iterations, same objective, dual 2.211e-9 → 2.208e-9). The failure got less reproducible, not the fix less effective, which leaves that guard with nothing to guard; it now runs with `POUNCE_DBG_NO_QUAD=1`, restoring the pre-Q4 arithmetic to every digit rather than being loosened.

## Tests

**`crates/pounce-cli/tests/quad_evaluator_differential.rs`** is the load-bearing artefact, modelled on Q3's recognizer differential. Every `.nl` file in the repository is built **twice** — fast path and `try_new_with_quadratic(…, false)` — and `eval_f`, `eval_grad_f`, `eval_g`, `eval_jac_g`, `eval_h` compared at the model's own starting point and three pseudo-random perturbations. The Jacobian and Hessian are compared as `(row, col) → value` maps, because the two paths need not agree on the *pattern*: a coefficient that cancels to exactly zero is dropped by the recognizer and kept by the tape's structural sparsity. Over 27 models and 251 quadratic rows:

| comparison | result |
|---|---|
| `∇²L`, assembled lower triangle | **26 846 of 26 848 bit-identical**; the 2 that are not differ by exactly **1 ulp** (`eigena2.nl`) |
| `g`, `∇g`, `f`, `∇f` | worst relative deviation **4.13e-15** |
| a model with nothing recognized | **byte-identical** both ways |

This bites: it is what found the `(x − 500000)²` cancellation, and it did so only once `eval_f`/`eval_grad_f` were added — on a model whose *rows* are not quadratic the objective is the only thing the phase touches. It also refutes §5.4's "plausibly bit-identical" for `∇²L` in general (right for the shape the note reasoned about, wrong where a model mixes a taped objective with quadratic rows).

`crates/pounce-nl/src/nl_quadratic.rs` gains unit tests pinning `is_expanded_quadratic` in **both** directions — admitting AMPL's `o54`-of-monomials emission, `xᵢ²`, and a singly-referenced `Cse`; refusing `(x₀ − x₁)²`, `(x₀ + 1)·x₁`, a mixed row where one factored term hides in a long expanded sum, and a 60-level shared-`Cse` DAG (which an exponential walk would not return from, so the test passing *is* the assertion) — plus a 250 000-term depth test on a default-sized libtest thread, because the gate runs on every row of every model including ones that overflow a recursive walk. `crates/pounce-nlp/src/quadratic.rs` gains nine unit tests, including one specifically on the factor-of-two trap (`push_form` takes *Hessian* entries, not polynomial coefficients).

Two existing tests are the ones that caught this phase's two defects, and both fail on an ungated build for the stated reason: gh#512's `an_adaptive_solve_stops_at_a_flagged_tiny_step` (`MaximumIterationsExceeded` at `iteration_count=300` on `airport.nl`) and `nl_reader::tests::shared_dag_walks_are_memoized_not_exponential` (0.00 s → 172.41 s). Neither is reachable by `sweep-fixtures.sh`, which runs default options on the fixture corpus. The gh#540 guard change is described under Blast radius; it keeps its assertions and gains an env var, so it still fails if the reproduction stops reproducing.

**Deliberately not pinned**: the benchmark numbers. `$POUNCE_BENCH_DATA` holds one instance here and the corpus is not vendorable; `scripts/gen-qcqp-nl.py` is checked in so the *shape* can be regenerated, but it is calibrated against the real instance and does **not** fully reproduce it — 50 iterations to the real 100, 61.8% factorization to 75.6% — so it overstates the aggregate win (1.57× generated vs 1.27× real) while the per-timer ratios transfer (80.9× vs 93.7× on `eval_h`). Every generated number in this PR, the commit and the note is labelled as generated.

---

- [ ] `CHANGELOG.md` `[Unreleased]` entry, in the user's terms — deferred to the series PR #603, which is where the phases land as one user-visible change
- [ ] Book page under `docs/src/` updated — n/a; no user-facing option or interface changed (`POUNCE_DBG_NO_QUAD` is a diagnostic, like `POUNCE_DBG_NO_HYBRID`)
- [x] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean — 2988 passed, 0 failed; clippy exit 0 with no new lint class

### DIFF
--- a/crates/pounce-cli/tests/issue_540_eigena2_superlinear_tail.rs
+++ b/crates/pounce-cli/tests/issue_540_eigena2_superlinear_tail.rs
@@ -73,6 +73,10 @@ fn tmp_path(suffix: &str) -> PathBuf {
 }
 
 fn solve(extra: &[&str]) -> SolveReport {
+    solve_with_env(extra, &[])
+}
+
+fn solve_with_env(extra: &[&str], env: &[(&str, &str)]) -> SolveReport {
     let json_path = tmp_path("eigena2.json");
     let sol_path = tmp_path("eigena2.sol");
     let mut cmd = Command::new(pounce_exe());
@@ -82,6 +86,9 @@ fn solve(extra: &[&str]) -> SolveReport {
         .arg(&json_path);
     for o in extra {
         cmd.arg(o);
+    }
+    for (k, v) in env {
+        cmd.env(k, v);
     }
     let _ = cmd.status().expect("spawn pounce");
     let text = std::fs::read_to_string(&json_path).expect("read json report");
@@ -124,9 +131,27 @@ fn eigena2_dual_infeasibility_clears_the_strict_tolerance() {
 /// must still reproduce the reported failure. If a later change fixes
 /// `eigena2` by some other route, this test fails and says so rather than
 /// letting the two tests above pass for a reason they do not describe.
+///
+/// **`POUNCE_DBG_NO_QUAD=1` is part of the reproduction, as of gh #588's Q4.**
+/// The pre-#540 route is a knife edge — it turns on an inertia count read off
+/// a factor whose smallest pivot is ~1e-16 — and Q4 moved `∇²L[8, 8]` by
+/// **one ulp** on this model. `eigena2` has 55 quadratic rows and a
+/// non-quadratic objective, so that entry is now assembled as a constant
+/// scatter plus a decoded tape share rather than as one sum, and the two
+/// associate differently. With the fast path on, `feral_inertia_pivot_floor=0`
+/// converges (27 iterations, dual 6.0e-10) instead of stalling, and this
+/// guard would pass vacuously in the other direction — by having nothing left
+/// to reproduce.
+///
+/// The env var restores the pre-Q4 arithmetic exactly: 29 iterations, dual
+/// 1.841e-7, `SolvedToAcceptableLevel`, the same numbers as before that
+/// commit. **The shipped route is untouched** — with the trigger on,
+/// `eigena2` takes 26 iterations to `SolveSucceeded` at dual 2.211e-9 before
+/// Q4 and 2.208e-9 after, on the same objective. So what changed is the
+/// reproducibility of a failure, not the fix and not the answer.
 #[test]
 fn disabling_the_trigger_reproduces_the_reported_failure() {
-    let r = solve(&NO_TRIGGER);
+    let r = solve_with_env(&NO_TRIGGER, &[("POUNCE_DBG_NO_QUAD", "1")]);
     assert_eq!(
         r.solution.status,
         ApplicationReturnStatus::SolvedToAcceptableLevel,
--- a/crates/pounce-cli/tests/quad_evaluator_differential.rs
+++ b/crates/pounce-cli/tests/quad_evaluator_differential.rs
@@ -1,0 +1,525 @@
+//! The Q4 constant-structure evaluator against the AD tape it replaces
+//! (gh #588).
+//!
+//! Q4 stops building a tape for every objective and row the recognizer
+//! proves is degree ≤ 2, and evaluates `g`, `∇g` and `∇²L` from the constant
+//! matrix instead. If those two ways of computing a derivative ever disagree,
+//! POUNCE hands the algorithm a wrong Hessian and converges somewhere else —
+//! or fails to — on a model no fixture solves. The fixture sweep cannot see
+//! that: it only reaches rows a *solve* depends on, and it reports one number
+//! per model rather than one per matrix entry.
+//!
+//! So this is the differential check, modelled on Q3's
+//! `quad_recognizer_differential.rs`: every `.nl` file in the repository is
+//! loaded **twice** — once with the fast path, once with
+//! `NlTnlp::try_new_with_quadratic(prob, false)`, which is what
+//! `POUNCE_DBG_NO_QUAD=1` selects — and the two are compared entry by entry
+//! at several points, on:
+//!
+//! * `eval_g` — the line search's inner loop;
+//! * `eval_jac_g` — compared as a `(row, col) -> value` map, because the two
+//!   paths need not agree on the *pattern*: a coefficient that cancels to
+//!   exactly zero is dropped by the recognizer and kept by the tape's
+//!   structural sparsity;
+//! * `eval_h` — same, over the assembled lower triangle.
+//!
+//! ### Which comparisons are bitwise
+//!
+//! `∇²L` is: the tape's entry for `0.5·((c·xᵢ)·xⱼ)` is the product of
+//! constants `0.5·c` and the decode adds `w·(0.5·c)`, while the scatter
+//! computes `w·q_val` with the same `q_val`. Nothing about that is an
+//! *approximation* of the other. **Measured, that is right for the shape the
+//! note reasoned about and wrong in general**: over this corpus 27 174 of
+//! 27 176 assembled Hessian entries are bit-identical, and the two that are
+//! not (`eigena2.nl`, entry (8, 8)) differ by exactly **one ulp**.
+//!
+//! The mechanism is worth stating because it decides where else it can
+//! happen: **an entry written by both paths at once**. `eigena2` has 55
+//! quadratic rows and a non-quadratic objective, so `∇²L[8, 8]` takes a
+//! constant contribution from a row and an `x`-dependent one from the
+//! objective's tape. On the tape path both land in the same compressed
+//! column pass and reach `values` as one add; on the fast path the row's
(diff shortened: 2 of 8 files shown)


jkitchin:

   produced.

Follow the series there.

PR #675, opened 2026-08-18
Q5 review copy: parse-time quadratic recognition (never build the Expr tree)

jkitchin:
**Review-only. Do not merge.** This PR exists so Q5 of the #588 series can be read as a self-contained diff against Q4; the series PR is **#603** and that is where the work actually lands. Base branch `review/588-q4-base` is Q4's commit `947b3aa3`, so the diff here is exactly one phase.

## Problem

Peak RSS on a quadratic model is set during *parsing*, by the `.nl` text and the `Expr` DAG. Q4 removed what is built from that DAG — the AD tapes — and recorded that the DAG itself was Q5's to remove:

| generated `qcqp500-3c` shape, full solve | peak RSS |
|---|---|
| before Q4 | 3855 MB |
| after Q4 | 1253 MB |
| **after Q5 (this PR)** | **586 MB** |

The CLI also recognized every row twice — once in `classify_problem`, once in `NlTnlp::try_new` — and `recognize_expr` was still `Θ(2^depth)` on a shared `Cse` DAG, which Q4 had to work around by refusing a re-referenced body outright.

## Cause

`Parser::parse_expr` allocates one `Expr` node per token, and `NlProblem` held the result for the life of the solve. For an AMPL-emitted quadratic row that is `n²` nodes describing `n²/2` distinct coefficients — the recognizer then walks all of them to recover a matrix the token stream already spelled out.

## Fix

`Parser::parse_expr_quadratic` reads a degree-2 body off the token stream as a `Quad2` and never builds the tree. The parser is line-cursor based, so a body it cannot prove costs one assignment to rewind (`self.pos = saved`) and `parse_expr` then produces exactly the tree it always produced — the same trick `parse_funcall_arg` already used for Hollerith literals.

- **`NlBody::{Tree, Quad}`** on `NlProblem` is most of the diff. Nine consumers read these bodies and several would read a *missing* tree as "this row is linear" — a silent wrong answer, not a crash. An enum makes each one a compile error until it says which reading it wants.
- **The tree is still available, and it is *the* tree.** A recognized body keeps the byte range it came from; `NlProblem::con_expr`/`obj_expr` re-parse those bytes with the same parser and the same `V`-segment `Arc`s. Rebuilding from the stored *coefficients* was rejected: that is a different tree, taped differently, and the `POUNCE_DBG_NO_QUAD` reference and gh#540's guard depend on the tape path being byte-for-byte what it was.
- **The exactness rule is structural, not a second copy of `is_expanded_quadratic`.** `+`/`-`/`o54` are the sum spine; everything under `*`, `/` or `^` is read in monomial mode; a `+` in monomial mode gives up; a `^` whose base token is not `n` / `v<i (i < n)` / `o16` gives up. Getting this wrong would put a *factored* form on Q4's constant-matrix path — the gh#544 failure mode that cost `airport.nl` 16 → 300 iterations.
- **`recognize_expr` and `is_expanded_quadratic` are memoized on `Arc` identity**, so Q4's `Cse` refusal is dropped. In the gate that is one line — a repeat visit `continue`s instead of returning `false`, sound for the reason `validate_expr` already documents — with the memo keyed on *(body, mode)*, because a body legal on the sum spine is not thereby legal inside a monomial.

A per-line byte-offset table was tried first for the byte ranges and rejected by measurement: on the generated model it cost **136 MB** (845.7 vs 698.8 MB) to save memory. It is one base-pointer subtraction now.

## Blast radius

**Bit-identical except for memory.** Both instances available here solve to every digit:

| | Q4 | Q5 |
|---|---|---|
| `qcqp1000-2c` (**real**) iterations / objective | 100 / 738127.4263173543 | 100 / 738127.4263173543 |
| …final dual inf / constraint violation | 2.467359649926948e-12 / 1.0186340659856796e-10 | identical |
| …load peak RSS, wall | 183.8 MB, 0.98 s | **102.2 MB, 0.71 s** |
| …full-solve peak RSS | 576.9 MB | **444.5 MB** |
| generated `qcqp500-3c` iterations / objective | 75 / 24779.69029930317 | 75 / 24779.69029930317 |
| …full-solve peak RSS | 1253.4 MB | **586.2 MB** (2.14×) |

- `scripts/sweep-fixtures.sh`, 57 models, against `947b3aa3`: **diff empty**.
- `Problem class:` over all 61 `.nl` files: **diff empty**.
- `quad_evaluator_differential` (Q4's): identical 27 models, 251 quadratic rows, 26 846 of 26 848 Hessian entries bit-identical, worst deviation 4.13e-15 — so dropping the `Cse` refusal cost no reach and gained none on this corpus.
- `POUNCE_DBG_NO_QUAD=1` now disables recognition at the parse as well as at the tape, and gives 487.7 MB from *both* binaries.
- Incidental: a left-deep `o0` chain of 20 000 or 100 000 monomials now loads and solves where it aborted with a stack overflow, because the streaming recognizer is iterative where `parse_expr` recurses.

## Tests

`crates/pounce-cli/tests/quad_parse_differential.rs` parses every `.nl` file in the repository twice, recognition on and off — 61 files, 1 631 bodies, 260 recognized at parse time — and asserts, on bit patterns and never a tolerance:

- every recognized form is bit-identical to `recognize_expr` on the tree;
- the recognized *set* is exactly `is_expanded_quadratic ∧ degree 2`, with a named failure for each direction of disagreement;
- every rebuilt tree is structurally identical to the non-recognizing parse's, constants compared as bits;
- rebuilt `Cse` references resolve to this parse's own `Arc`s;
- the stored variable support matches `collect_vars`, including variables whose coefficient cancelled to zero;
- the bounds the gh#492 constant-row fold writes are bit-identical.

Three hand-written models pin what the corpus does not reach: a twice-referenced `V`-segment quadratic (which Q4 refused and this admits — no fixture in the repo has a recognized body that references a CSE at all), a factored defined variable that must *not* be recognized, and `a_sumlist_folds_in_the_order_the_tree_walk_folds`. That last one exists because `recognize_expr` folds an `o54` sumlist **last operand first** while a streaming parser naturally accumulates front to back; with `[1e16, 1, 1]` on the same `x₀²` the two orders give `1e16` and `1e16 + 2`. It was found by writing the test, and it fails on the natural implementation.

`a_shared_cse_body_is_refused_rather_than_walked_exponentially` is replaced by two tests asserting the opposite verdict at depth 60 and the per-context memo key; it fails on the parent commit, for the stated reason — that is the behaviour change.

Deliberately not pinned: the memory numbers themselves. Peak RSS is measured, reported in the note (§0f) and reproducible with `scripts/gen-qcqp-nl.py`, but a test that asserts an RSS figure would be a flake on any other machine.

---

- [ ] `CHANGELOG.md` `[Unreleased]` entry — the series PR #603 carries it; this branch is review-only
- [ ] Book page under `docs/src/` — no user-facing surface changes
- [x] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean (2995 passed, 0 failed; 2988 at Q4)
- [x] Every claim in this body is true of the code as it stands now

---

### DIFF
--- a/crates/pounce-cli/src/dispatch.rs
+++ b/crates/pounce-cli/src/dispatch.rs
@@ -336,9 +336,9 @@ fn header_census(prob: &NlProblem) -> String {
     let tree_rows = prob
         .con_nonlinear
         .iter()
-        .filter(|e| !is_trivially_zero(e))
+        .filter(|b| !b.is_trivially_zero())
         .count();
-    let tree_obj = usize::from(!is_trivially_zero(&prob.obj_nonlinear));
+    let tree_obj = usize::from(!prob.obj_nonlinear.is_trivially_zero());
     let Some(c) = prob.nl_counts else {
         return format!("no .nl header census; trees: nl_rows={tree_rows} nl_obj={tree_obj}");
     };
@@ -373,14 +373,14 @@ fn classify_inner(prob: &NlProblem) -> (ProblemClass, ClassReason) {
     // claim would route a model to the LP solver on the strength of a
     // header field, which is not a trade this classifier makes anywhere
     // else. The header is logged beside the verdict instead.
-    let obj_nl = !is_trivially_zero(&prob.obj_nonlinear);
-    let cons_nl = prob.con_nonlinear.iter().any(|e| !is_trivially_zero(e));
+    let obj_nl = !prob.obj_nonlinear.is_trivially_zero();
+    let cons_nl = prob.con_nonlinear.iter().any(|b| !b.is_trivially_zero());
     if !obj_nl && !cons_nl {
         return (ProblemClass::Lp, ClassReason::NoNonlinearParts);
     }
 
     // Objective curvature.
-    let obj_quad = match analyze_quadratic(&prob.obj_nonlinear) {
+    let obj_quad = match prob.obj_nonlinear.analyze_quadratic() {
         Some(q) => q,
         // Objective has a non-quadratic nonlinear term ⇒ NLP.
         None => return (ProblemClass::Nlp, ClassReason::ObjectiveNotQuadratic),
@@ -390,10 +390,10 @@ fn classify_inner(prob: &NlProblem) -> (ProblemClass, ClassReason) {
     // any non-quadratic constraint term makes the whole problem NLP.
     let mut any_quadratic_constraint = false;
     for (row, c) in prob.con_nonlinear.iter().enumerate() {
-        if is_trivially_zero(c) {
+        if c.is_trivially_zero() {
             continue;
         }
-        match analyze_quadratic(c) {
+        match c.analyze_quadratic() {
             Some(q) if q.is_empty() => {} // purely linear after all
             Some(_) => any_quadratic_constraint = true,
             None => {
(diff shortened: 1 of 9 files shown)


jkitchin:

   produced.

Follow the series there.

PR #676, opened 2026-08-18
Q6 review copy: prove the four constant-derivative hints instead of trusting them

jkitchin:
**Review-only. Do not merge.** This PR exists so phase Q6 of the #588 series can be read as a single diff against its parent (Q5, `19340e6c`). The series PR is **#603**; the commit lands there. Mirrors #671 / #674 / #675.

## Problem

Ipopt registers four caching hints — `grad_f_constant`, `hessian_constant`, `jac_c_constant`, `jac_d_constant` — as **unchecked user assertions**. Setting one makes the solver evaluate that derivative once and reuse it for the rest of the solve; if the assertion is false, the answer is silently wrong. A Hessian frozen at the starting point is still *a* matrix, and the algorithm converges to something with it.

POUNCE's position until now was the mirror image, and equally unhelpful: the four names sat in `unimplemented_options::UNEXPLOITED_HINTS`, setting one printed "pounce does not exploit this", and every derivative was re-evaluated every iteration.

Neither is right, and the model usually knows which. Over this repo's 60-model `.nl` corpus, POUNCE can prove `∇²L` constant on **29** of them and prove it *not* constant on **19** more, without asking anyone.

## Cause

The recognizer that Q3–Q5 built already answers the question exactly — it just was not wired to it. What was missing was the reconciliation: what to do when the model's answer and the user's option disagree, and what to do when the model has no answer at all.

## Fix

`pounce_nlp::constant_derivatives`. `DerivativeProof` is **three-valued**, and that is the substance of the phase — the design note describes two states, and two states is a bug in each direction.

| proof | user asserted | POUNCE | why |
|---|---|---|---|
| `Constant` | either | **reuses** | the option is redundant; the model proved it |
| `Varying` | no | evaluates each iterate | nothing to do |
| `Varying` | **yes** | **warns and ignores** | the divergence — Ipopt honours this and returns a wrong Hessian |
| `Unknown` | no | evaluates each iterate | nothing to do |
| `Unknown` | **yes** | **honours on trust** | upstream's contract |

The last row is the one that had to be got right. A callback TNLP — the C interface, the Python bridge, **both GAMS links** — hands POUNCE numbers, not algebra, so *every* proof comes back `Unknown`. Refusing a hint there, on the grounds that POUNCE could not confirm it, would be the same class of silent wrong answer as honouring a false one, pointed the other way. `Unknown` means *not established*, never *varies*.

**The accessor Q5's handoff named is the wrong one, and this is the largest decision in the diff.** Q5 pointed at `NlBody::admitted_quad_form`. That sits behind Q4's *exactness* gate (`is_expanded_quadratic`), which exists because **evaluating** a factored quadratic from stored coefficients cancels — `(x − 500000)²` loses five digits, gh#544. Q6 evaluates nothing from coefficients; it hands back the number the model's own tape produced at the first call. The applicable question is only the body's **degree**, which a factored `(x − a)²` answers as well as an expanded one — so the proof reads a new `NlBody::provably_affine`, same recognizer, no gate.

Measured cost of the gate had it been used (corpus census, run both ways):

| | ungated (shipped) | gated on `is_expanded_quadratic` |
|---|---|---|
| models with `∇²L` proved constant | **29** | 11 |
| models where a false `grad_f_constant` is detectable | **34** | 9 |
| rows proved varying | **296** | 251 |
| rows proved constant | 787 | 787 |

`airport.nl` — the model that forced Q4's gate into existence — is one of the models Q6 reaches. Row-level *constancy* is identical either way; everything the gate costs is on the factored side, which is exactly where it was designed to bite.

**Reuse mechanism.** Upstream's own: the derivative goes into the cache it already had, under an **empty dependency list**, so the lookup that runs on every call finds it at every point. Two deliberate deviations, both cheap:

* `∇²L` keeps `obj_factor` as a scalar dependency. The restoration phase calls `orig.eval_h(x, 0.0, ..)`; upstream drops the scalar and is right only because the main algorithm always passes 1.0.
* A NaN-filled result from a failed user evaluation is **never** stored as *the* constant answer. Otherwise one bad trial point poisons the rest of the solve instead of being backtracked over.

**Two translations in `OrigIpoptNlp`**, because it is the only layer that knows both row orders:

* the model proves things about *its* rows, the algorithm evaluates a c-block and a d-block, so `c_map`/`d_map` do the split — deriving it a second time from `g_l == g_u` inside the model would be a second rule that could disagree with the first;
* under `fixed_variable_treatment = make_parameter` a `Varying` proof is weakened to `Unknown`. A row `x·y` with `y` fixed has a *constant* gradient in the space the algorithm differentiates in. Constancy survives fixing; a proof of variation does not. Backwards, that guard would make POUNCE refuse a hint that is **true**.

**Both GAMS links now assert what GMO already told them.** When no Jacobian nonzero carries GMO's nonlinear flag, `gams/gams_pounce.c` and `python/pounce/gams/link.py` set `jac_c_constant=yes` / `jac_d_constant=yes` before reading the option file, so a `pounce.opt` can still turn it off. This is not a new claim: both links *already* rely on it per row, copying cached coefficients for a linear row rather than calling the GMO evaluator. If it were false they would already be returning wrong Jacobians.

`UNEXPLOITED_HINTS` is now empty. The table and `hint_warnings` stay — the shape is right for the next registered-but-unexploited hint, and the two membership tests are what keeps such an option from going silent again.

## Blast radius

**Sweep: `scripts/sweep-fixtures.sh` is diff-empty across all 57 models** against the parent commit `19340e6c` — status, objective *and iteration count* identical everywhere. That is the assertion that matters for a phase which changes how often derivatives are evaluated.

What does move is evaluation counts, over the same 57 models:

| counter | Q5 | Q6 | |
|---|---|---|---|
| iterations | 2082 | 2082 | unchanged, by construction |
| `eval_grad_f` | 1729 | **1161** | −568 (−33%) |
| `eval_jac_g` | 2005 | **1573** | −432 (−22%) |
| `eval_h` | 1880 | **1873** | −7 (−0.4%) |

13 of 57 models move: `pooling_rt2stp` `∇f` 304 → 1 over 298 iterations, `unbounded_cubic` `∇g` 291 → 1, `deb7` `∇f` 157 → 1, `autocorr_bern55-06` 73 → 1, the three `jit1` variants `∇g` 26 → 1, `nonconvex_qp` `∇g` and `∇²L` both 7 → 1.

**And on none of them does a timer move.** `pooling_rt2stp`: `ObjectiveGradientEvaluations` 0.006 s either way, `OverallAlgorithm` 0.641 vs 0.637 s. A linear objective's gradient is the cheapest thing in the model, so removing 300 of them is below timer resolution. This is a correctness change with a measurable evaluation-count effect and **no measurable time effect**, and it should be read that way.

**Real `qcqp1000-2c` (the one Mittelmann instance available) — the control, and it does not move**, exactly as the ≈0% forecast said. Proofs: `grad_f`/`hessian`/`jac_d` **varying**, `jac_c` **constant** (its 100 equalities are linear; the 7 quadratic rows are all inequalities). 100 iterations both ways, objective `738127.4263173543` to every digit, every reported eval counter identical (357/357/105/101/101), `OverallAlgorithm` 105.049 → 106.326 s — 1.2% against a ~1.3× noise floor, i.e. nothing.

**Forecasts that were wrong**, recorded in note §0g:

* *"single-digit % on true QPs forced down the NLP path"* — capped at **2.7% by Q4**. On a **generated** dense 500-variable QP (`gen-qcqp-nl.py --quad-rows 0`, `solver_selection=nlp`) `num_hess_evals` goes 18 → 1 at 17 unchanged iterations and `TotalFunctionEvaluations` 0.043 → 0.021 s — but evaluation is only 5.3% of `OverallAlgorithm` there *because Q4 already removed the rest of it*. Q6 removes the residue of a residue.
* *"…callback/GAMS front ends that assert the hint"* — false premise. Neither GAMS link asserted any of the four. They do now.
* `hessian_constant`, the hint §4's Lever 2 is written around, has the **least** reach on the NLP path (corpus `eval_h` −0.4%), because a model that qualifies for it is a QP and POUNCE routes QPs to `pounce-qp`/`pounce-convex`. The hints that pay are the two Lever 2 never mentions.

Decorators: `ScalingTnlp`, `CountingTnlp` and `SeededTnlp` forward the proofs (every transform they apply is a multiply by a constant, and neither direction of the proof is affected). Presolve wrappers deliberately do **not** — they change what the rows are — and the declining default makes that the free, safe choice.

## Tests

* **`crates/pounce-cli/tests/constant_derivative_proofs.rs`** — the load-bearing artefact, in the shape of Q3's, Q4's and Q5's differentials. Every `.nl` file in the repository is asked what it can prove, and then the derivative it made a claim about is evaluated at five points — and, for `∇²L`, at five *multiplier vectors*, because `∇²L = σ∇²f + Σᵢλᵢ∇²gᵢ` depends on λ and a proof that ignored that would be a proof of the wrong thing. Over 60 models: **59,756 derivative values compared, every `Constant` claim bit-identical, and all 349 `Varying` proofs witnessed by a derivative that actually moved** (none unwitnessed). Both directions are checked because both bite: a spurious `Varying` is not a wrong answer, but it refuses a user's *true* hint while claiming the model disproves it. A fourth case pins that the proofs do not move with parse-time recognition on/off, so `POUNCE_DBG_NO_QUAD=1` stays a real A/B.

  This test found the phase's one real defect on its own: the first draft read the Jacobian triplet rows as 1-based and compared row *i*'s values against row *i+1*'s proof. `NlTnlp` reports `IndexStyle::C`. It failed on `linear_eq_aggregation.nl` immediately.

* **`crates/pounce-algorithm/tests/constant_derivative_hints.rs`** — the `Unknown` state end to end, on a QP written as callbacks and nothing else: 12 Hessian evaluations without the hint, **1** with it, bit-identical objective and solution. Plus: an explicit `hessian_constant=no` asserts nothing.

* **`crates/pounce-cli/tests/unimplemented_options.rs`** — the pin the note asked to flip. `a_caching_hint_warns_but_solves` (string match on "pounce does not exploit") becomes `a_disproved_caching_hint_is_refused_with_a_warning` and `a_proved_constant_hessian_is_evaluated_once`, the latter asserting `num_hess_evals == 1` and `num_constr_jac_evals == 1` out of `--json-output` on a six-iteration solve.

* **`python/tests/test_gams_link.py`** — five license-free cases over the fake `GmoView`, including that a view which *declines* the linearity question never asserts the hint, and that the link sets it before the user's option file.

**How they bite on the parent commit:** all three Rust files are new or newly-inverted and fail on `19340e6c` for the right reason — `constant_derivative_proofs.rs` and `constant_derivative_hints.rs` do not compile there (`TNLP::derivative_proofs` and `pounce_nlp::constant_derivatives` do not exist), and `a_proved_constant_hessian_is_evaluated_once` fails on the count (`num_hess_evals == 7`, not 1).

**Deliberately not pinned:** the C GAMS link's half of the hint. There is no GAMS licence and no GAMS shared library in any CI runner, so `gams/gams_pounce.c` is untested here for the same reason the rest of that file is; it is code-reviewed against the `row_has_nl` array it already keys its own fast path on. The two `solve_view` Python cases could not be *run* in the container that produced this branch (the native `pounce._pounce` extension is not built there) — they are pure-Python monkeypatch tests and run in CI.

---

- [ ] `CHANGELOG.md` `[Unreleased]` entry — deliberately omitted, matching Q0–Q5: the series lands one CHANGELOG entry at the end
- [x] Book page under `docs/src/` updated — `options.md` (the three states, with `POUNCE_DBG_CONSTDERIV`) and `gams.md` (the Jacobian hints the links now set)
- [x] `cargo fmt --all -- --check`, `cargo clippy --workspace --exclude pounce-hsl --all-targets` (exit 0, no new lint instance in the new code), `cargo test --workspace --exclude pounce-hsl` (3008 passed, 0 failed; 2995 at Q5)
- [x] Every claim in this body is true of the code as it stands

---

### DIFF
--- a/crates/pounce-algorithm/src/application.rs
+++ b/crates/pounce-algorithm/src/application.rs
@@ -1296,6 +1296,51 @@ impl IpoptApplication {
         crate::unimplemented_options::hint_warnings(&self.options, &self.reg_options)
     }
 
+    /// Resolve the four constant-derivative hints for this solve and
+    /// install the result on the NLP (gh #588, phase Q6).
+    ///
+    /// `grad_f_constant` / `hessian_constant` / `jac_c_constant` /
+    /// `jac_d_constant` are, upstream, unchecked user assertions: Ipopt
+    /// reuses the derivative and returns a wrong answer if the assertion
+    /// was false. pounce asks the model first. Where the model *proves*
+    /// the derivative constant, the reuse happens whether or not the
+    /// option was set — the hint is redundant. Where the model proves it
+    /// **varies** and the option was set anyway, the option is refused
+    /// with a warning, which is the deliberate divergence. Where the
+    /// model can prove nothing — every callback front end, both GAMS
+    /// links — the user's assertion is honoured on trust, exactly as
+    /// upstream, because "unproved" is not "disproved" and silently
+    /// overriding the caller there would be its own wrong answer.
+    fn install_constant_derivative_hints(&self, orig_nlp: &mut OrigIpoptNlp) {
+        use pounce_common::journalist::JournalCategory;
+        use pounce_nlp::constant_derivatives::{HINT_OPTIONS, reconcile};
+
+        let asserted = HINT_OPTIONS
+            .map(|name| matches!(self.options.get_bool_value(name, ""), Ok((true, true))));
+        let proofs = orig_nlp.derivative_proofs();
+        let (outcomes, enabled) = reconcile(proofs, asserted);
+
+        for outcome in &outcomes {
+            if let Some(warning) = outcome.warning() {
+                eprintln!("{warning}");
+                self.journalist.print(
+                    JournalLevel::J_STRONGWARNING,
+                    JournalCategory::J_MAIN,
+                    &format!("{warning}\n"),
+                );
+            }
+        }
+        if std::env::var("POUNCE_DBG_CONSTDERIV").is_ok() {
+            for outcome in &outcomes {
+                eprintln!(
+                    "[const deriv] {:<15} proof={:?} asserted={} reused={}",
+                    outcome.name, outcome.proof, outcome.asserted, outcome.honoured,
+                );
+            }
+        }
+        orig_nlp.set_constant_derivatives(enabled);
+    }
+
     /// Resolve the five registered `derivative_test*` knobs. Every one
     /// of them was registered and never read, so `derivative_test=
     /// first-order` ran no test and printed nothing — a checker that
@@ -1432,13 +1477,16 @@ impl IpoptApplication {
             .ok()
             .and_then(|(v, f)| f.then_some(v))
             .unwrap_or(1.0);
-        let orig_nlp = match OrigIpoptNlp::new(
+        let mut orig_nlp = match OrigIpoptNlp::new(
             Rc::clone(&adapter),
             Rc::new(ConstObjScaling(obj_scaling_factor)),
         ) {
             Ok(n) => n,
             Err(_) => return ApplicationReturnStatus::InternalError,
         };
+        // Same Q6 reconciliation as the IPM route: the SQP driver
+        // evaluates the same derivatives through the same NLP object.
+        self.install_constant_derivative_hints(&mut orig_nlp);
         let nlp_rc: Rc<RefCell<dyn IpoptNlp>> = Rc::new(RefCell::new(orig_nlp));
 
         let mut sqp_adapter = crate::sqp::IpoptNlpAdapter::new(Rc::clone(&nlp_rc));
@@ -2262,6 +2310,9 @@ impl IpoptApplication {
             }
         };
         orig_nlp.set_timing_stats(Rc::clone(&timing));
+        // Q6: decide which derivatives may be reused across iterates,
+        // before anything is evaluated (gh #588).
+        self.install_constant_derivative_hints(&mut orig_nlp);
 
         // Mirror upstream `OrigIpoptNLP::InitializeStructures` (IpOrigIpoptNLP.cpp:299):
         // bail out with NotEnoughDegreesOfFreedom when there are fewer free
--- a/crates/pounce-algorithm/src/unimplemented_options.rs
+++ b/crates/pounce-algorithm/src/unimplemented_options.rs
@@ -237,12 +237,22 @@ pub const UNIMPLEMENTED_VALUES: &[UnimplementedValue] = &[UnimplementedValue {
 /// unaffected — but whose performance hint pounce does not exploit.
 /// These warn rather than fail: refusing them would stop a solve that
 /// returns the right result today, only a little slower.
-pub const UNEXPLOITED_HINTS: &[&str] = &[
-    "grad_f_constant",
-    "hessian_constant",
-    "jac_c_constant",
-    "jac_d_constant",
-];
+///
+/// **Empty since gh #588 phase Q6.** The four constant-derivative hints
+/// (`grad_f_constant`, `hessian_constant`, `jac_c_constant`,
+/// `jac_d_constant`) lived here and are now exploited:
+/// [`pounce_nlp::constant_derivatives`] reconciles each one against what
+/// the model can prove about its own algebra, and
+/// `OrigIpoptNlp` reuses the derivative across iterates when the answer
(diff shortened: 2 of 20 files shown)


jkitchin:

   produced.

Follow the series there.

PR #680, opened 2026-08-18
L-BFGS fidelity: σ, recalc_y, slack-based scaling, restoration densification, and the missing reset

### LINKED ISSUE #524: Vanderbei: cresc4 (false local-infeasibility) and steenbrf (3000-iter stall) — the only 2 problems Ipopt solves and POUNCE does not; both fixed by mu_strategy=adaptive

### LINKED ISSUE #677: L-BFGS initial Hessian scalar is pinned to scalar2; `limited_memory_initialization` is registered but never read
Every limited-memory solve uses σ = yᵀy/sᵀy (`scalar2`). Ipopt defaults to σ = sᵀy/sᵀs (`scalar1`). The option that selects this is registered with the correct default and read nowhere, so it cannot be set and does not warn.

Found while diagnosing a downstream report on #670 ([comment](https://github.com/jkitchin/pounce/pull/670#issuecomment-5323379210)): a 59,939-variable CasADi collocation model where POUNCE + L-BFGS diverges from Ipopt + L-BFGS with the duals running to `inf_du` ≈ 6e8, while Ipopt converges normally.

## The defect

`limited_memory_initialization` is registered with upstream's default:

```rust
// crates/pounce-algorithm/src/upstream_options.rs:1065
r.add_string_option("limited_memory_initialization", …, "scalar1",
    &[("scalar1", "sigma = s^Ty/s^Ts"), ("scalar2", "sigma = y^Ty/s^Ty"), …])?;
```

There is no read site. `AlgorithmBuilder` has no corresponding field, and the updater is constructed with:

```rust
// crates/pounce-algorithm/src/alg_builder.rs:1228-1234
HessianApproxChoice::LimitedMemory => Box::new(LimMemQuasiNewtonUpdater {
    update_type: self.limited_memory_update_type,
    max_history:  self.limited_memory_max_history,
    init_val_max: self.limited_memory_init_val_max,
    init_val_min: self.limited_memory_init_val_min,
    nonlinear_vars: self.limited_memory_nonlinear_vars.clone(),
    ..LimMemQuasiNewtonUpdater::default()      // <- initial_approx: Scalar2
}),
```

so `initial_approx` always comes from `lim_mem_quasi_newton.rs:114`. It is also absent from `unimplemented_options.rs`, so setting it produces neither effect nor warning.

This is the same miss as gh#483 / #191 round 2, one argument over. `application.rs:3316` records wiring `limited_memory_init_val_max`/`_min` into `initial_hessian_scalar` — `limited_memory_initialization` is that same function's third parameter and was not picked up in that sweep.

## Why it matters numerically

σ_scalar2 / σ_scalar1 = (yᵀy · sᵀs) / (sᵀy)² ≥ 1 by Cauchy–Schwarz, and grows without bound as the curvature pair becomes ill-conditioned. An over-large σ makes `B₀ = σI` dominate the six rank-2 corrections; the model collapses toward a large multiple of the identity, the primal step shrinks, and the equality multipliers absorb the mismatch.

`compute_sigma_bfgs` returns a hard-coded `1.0` while the history is empty, so iteration 1 cannot differ between the two settings — divergence begins at iteration 2, when the first curvature pair lands. In the reported logs, POUNCE and Ipopt agree to the last printed digit at iteration 1 and separate at iteration 2.

## Measured impact

`scripts/sweep-fixtures.sh`, 57 fixtures, scalar2 → scalar1.

**Default (exact-Hessian) path: zero diff, all 57 identical.** `initial_approx` is only reachable under `HessianApproxChoice::LimitedMemory`.

Under `hessian_approximation=limited-memory`, 20 of 57 move. Improvements:

| fixture | scalar2 (today) | scalar1 |
|---|---|---|
| `jit1` | MaxIter @3000, obj 177485.8 | **Solved, 35 it**, obj 173345.4 |
| `jit1_boxed` | MaxIter @3000, obj 177011.2 | **Solved, 35 it**, obj 173345.4 |
| `jit1_node` | MaxIter @3000, obj 398830.5 | **Solved, 36 it**, obj 173345.4 |
| `pooling_rt2stp` | RestorationFailed, 14 it | **Solved, 413 it**, obj −28.19 → −4391.83 |
| `autocorr_bern55-06` | 2924 it, obj −2304.0 | **106 it**, obj −2368.0 |
| `csfi2` | 60 it | 32 it |
| `eigenb2` | SolvedToAcceptableLevel, 52 it | **SolveSucceeded**, 69 it |
| `linear_eq_collapsed_box` | 18 it | 14 it |
| `user_scaling_suffix` / `_var_suffix` | 11 it | 9 it |

The three `jit1` variants previously returned three different answers at the iteration cap and now agree on one objective in ~35 iterations.

Lines that move the other way, or that need explaining before this lands:

- `cresc4` — MaxIter@3000 → RestorationFailed@1692. Failure either way, but the mode changed.
- `deb7` — MaxIter@3000 → RestorationFailed@597. Same shape.
- `eigena2` — ErrorInStepComputation at 150 → 178. Still errors.
- `hs13_bigstart` — solved both ways, obj 0.9943957 → 0.9941229 (~2.8e-4 relative). HS13 is MFCQ-degenerate at its solution so more than one tolerance-legal answer is plausible, but that is a hypothesis, not a finding.

Neutral: `airport` 54→56, `hs71_obj1e8` 15→16 (same obj), `issue_508_infeasible_gap_1em4` 72→76, `convex_qp_sens` 1.2e-30→1.8e-30. The two `InfeasibleProblemDetected` objective moves are not meaningful — that field is not a solution on an infeasibility certificate.

Caveat on the baseline: this sweep forces L-BFGS onto a corpus otherwise exercised with the exact Hessian, so the "before" column is not shipped default behaviour for these models. It **is** shipped behaviour for anyone on `hessian_approximation=limited-memory`, which is what CasADi uses whenever no Hessian is supplied.

## Proposed fix, in two separable parts

1. **Wire the option.** Add the `AlgorithmBuilder` field and read site so `limited_memory_initialization` does something. Sweep-free if the default stays `scalar2`. Note `InitialApprox` only has `Identity`/`Scalar1`/`Scalar2` — upstream also registers `scalar3` (arithmetic mean), `scalar4` (geometric mean) and `constant`, so honouring the full registered set means implementing three more.
2. **Flip the effective default to `scalar1`** for Ipopt parity. This is the trajectory change; the sweep above is its evidence, and the four lines listed need explanations. If `cresc4`/`deb7`'s new failure mode is accepted as a cost, that needs its own issue and an owner rather than a line in a commit message.

## Adjacent: 26 core options registered, never read, not declared unimplemented

Auditing all 353 registered options against every accessor form (`get_*_value`, `read_num`, `read_int`) plus the `unimplemented_options.rs` declarations leaves 26 core options that are silently inert (linear-solver knobs for backends POUNCE does not implement excluded):

```
alpha_for_y_tol            alpha_red_factor         corrector_compl_avrg_red_fact
delta                      derivative_test_perturbation  derivative_test_tol
eta_penalty                expect_infeasible_problem_ctol  expect_infeasible_problem_ytol
filter_margin_fact         filter_max_margin        fixed_mu_oracle
limited_memory_init_val    limited_memory_initialization   n_sens_steps
neg_curv_test_reg          neg_curv_test_tol        nu_inc
nu_init                    resto_failure_feasibility_threshold  rho
s_max                      sens_bound_eps           sens_max_pdpert
tau_min                    theta_min
```

None of these names appear as a string literal anywhere outside `upstream_options.rs`. Ones that look worth their own triage:

- **`limited_memory_init_val`** — same family, same function. Upstream defines B₀ as this multiple of the identity on the first iteration, before any updates; `compute_sigma_bfgs` hard-returns `1.0` for the empty-history case. That happens to equal the default, so it is currently invisible, but it is not tunable and `constant` mode has nowhere to read its value from.
- **`tau_min`** — the fraction-to-boundary floor. `MonotoneMuUpdate` carries the field with upstream's 0.99 default but nothing populates it from options.
- **`s_max`** — the s_d/s_c scaling cap in the KKT error test, a core convergence parameter.
- **`fixed_mu_oracle`** — `mu_oracle` and `adaptive_mu_globalization` are both wired; this one is not.
- **`theta_min`**, `filter_margin_fact`, `filter_max_margin`, `eta_penalty`, `alpha_red_factor` — filter / line-search parameters.

Each is a silent no-op today. Setting any of them gets no effect and no diagnostic, which is the property that made this one expensive to find. A cheap guard independent of the fixes: fail the registry build (or emit the `unimplemented_options.rs` warning) for any registered option with no read site, so the next one of these is caught at compile time rather than by a user comparing trajectories against Ipopt.

### LINKED ISSUE #684: Restoration densifies the L-BFGS Hessian: O(n²) allocation, 14 GB at n=60k
Entering restoration with `hessian_approximation=limited-memory` allocates a dense lower triangle of the full Hessian. On a 59,939-variable model that is a single 14 GB allocation and the solve dies.

Reported by @srikanth-gm on [#680](https://github.com/jkitchin/pounce/pull/680#issuecomment-5331841976), running a direct-collocation NLP:

```
memory allocation of 14379225712 bytes failed
```

## It is an n×n, not a big problem

```
14,379,225,712 / 8 bytes  =  1,797,403,214 entries
n(n+1)/2 = 1,797,403,214  →  n = 59,956
```

which is the model's variable count in pounce's compressed `x_var` space. A dense lower triangle, allocated in one call.

## Where

`materialize_orig_block`, `crates/pounce-restoration/src/aug_resto_system_solver.rs:535-560`.

Restoration cannot consume the low-rank L-BFGS `W` directly, so this densifies it into a `SymTMatrix`, reserving `n(n+1)/2` entries for `irows` (i32), `jcols` (i32) and `vals` (f64) — about 6.7 GB + 6.7 GB + 13.4 GB at this size. The `vals` allocation is the one that fails first.

The function's own comment states the assumption that breaks:

> Cost is `O(rank·n_orig²)`; restoration is a rare heavyweight fallback and `n_orig` is small, so this is negligible.

`n_orig` is the full primal dimension. "Small" holds across the fixture corpus, where the largest model is a few thousand variables. It does not hold for the problem class L-BFGS exists to serve — no exact Hessian available, tens of thousands of variables — which is exactly the population that reaches this code.

## Why nothing caught it

- **The corpus is too small.** Every fixture is a few thousand variables at most, so `n(n+1)/2` is a few million entries and invisible.
- **The corpus never ran L-BFGS** until #677 added the sweep's `lbfgs` leg — and even that leg only reaches this path on a fixture that both uses limited-memory *and* enters restoration.
- **The scaling probe added on #677 measured the wrong path.** `scripts/scaling-probe.sh` reported the limited-memory path linear (slope 1.05, R² 0.993) from n=2,000 to n=128,000. That result is real but does not cover this: none of the probe problems enter restoration, so the fallback was never exercised. A linear main iteration says nothing about a quadratic fallback, and the probe was presented with more confidence than it had earned.

## Fix

Stop densifying. The structure is already factored — `B = σI + VVᵀ − UUᵀ` with `rank ≤ 2·limited_memory_max_history` — and restoration should consume it in that form, which is `O(n·rank)` storage instead of `O(n²)`.

The obstacle is the one the comment describes: the resto `W` stores `D` as a flat `DenseVector` but its `V`/`U` columns as 5-block resto `CompoundVector`s, so no single probe-vector type threads through `mult_vector` (pounce#102). That is a plumbing problem, not a mathematical one.

An interim guard is worth having regardless: refuse or fall back with a clear message when `n_orig` is large enough that the dense form cannot fit, rather than attempting a 14 GB allocation and aborting. A user should not learn about this from a raw allocator failure.

## Test that would have caught it

Extend `scripts/scaling-probe.sh` — or add a companion — over a problem family that *enters restoration*, not just one that iterates. The current probe's own limitation is now documented in this issue; whatever fixes this should close that gap too, or the next fallback path will hide the same way.

### LINKED ISSUE #686: L-BFGS never resets the limited-memory approximation — `limited_memory_max_skipping` is registered but unread
**Version:** pounce 0.10.0 @ `642c36f9bbcb40d96a73985f2cf755b5723e4eff`
**Component:** `crates/pounce-algorithm/src/hess/lim_mem_quasi_newton.rs`
**Reference:** Ipopt @ `2695946fa79d2e84f3034e065e788933a81466eb`, `IpLimMemQuasiNewtonUpdater.cpp`

> **AI disclosure.** This report — the investigation, the instrumentation, the
> log analysis, the attached patch and this text — was produced with the help of
> an AI coding assistant, then reviewed by a human before submission. The
> measurements and log excerpts below are from real runs on real hardware and
> can be reproduced from the attached configurations. Please still treat the
> source-level reasoning with the scepticism you would apply to any drive-by
> contribution.

## Files
[some proprietary info scrubbed from header]
[0001-fix-lbfgs-reset-the-approximation-after-limited_memo.patch](https://github.com/user-attachments/files/31197845/0001-fix-lbfgs-reset-the-approximation-after-limited_memo.patch)
[config_ipopt.json](https://github.com/user-attachments/files/31197847/config_ipopt.json)
[config_pounce.json](https://github.com/user-attachments/files/31197842/config_pounce.json)
[ipopt_reference_60it.log](https://github.com/user-attachments/files/31197841/ipopt_reference_60it.log)
[ipopt_reference_full_converged.log](https://github.com/user-attachments/files/31197846/ipopt_reference_full_converged.log)
[pounce_AFTER_fix_60it.log](https://github.com/user-attachments/files/31197843/pounce_AFTER_fix_60it.log)
[pounce_AFTER_fix_250it.log](https://github.com/user-attachments/files/31197844/pounce_AFTER_fix_250it.log)
[pounce_BEFORE_fix_60it.log](https://github.com/user-attachments/files/31197840/pounce_BEFORE_fix_60it.log)

## Summary

`limited_memory_max_skipping` is registered in `upstream_options.rs:1069` with
upstream's default of 2, but no read site exists — it is listed as known debt in
`tests/no_silent_options.rs` under "#551 section 1 — feature runs, read site
missing". The underlying feature does **not** run: `LimMemQuasiNewtonUpdater`
has no skip counter and never discards its history.

Upstream resets the entire limited-memory approximation once the curvature
update has been skipped `limited_memory_max_skipping` times in a row
(`IpLimMemQuasiNewtonUpdater.cpp:226-260`), re-anchoring `last_x_` /
`last_grad_f_` / `last_jac_c_` / `last_jac_d_` at the current iterate, zeroing
`curr_lm_memory_`, and dropping `S_ Y_ Ypart_ D_ L_ V_ U_ SdotS_ STDRS_ DRS_`.

On any problem whose Lagrangian curvature is persistently negative, pounce
therefore keeps whatever `(s, y)` pairs it happened to accept in the first few
iterations for the remainder of the solve. The Hessian model stops representing
the current curvature; `inf_pr` converges normally while `inf_du` plateaus.

This is not a tuning difference — it is a missing algorithmic step. It is the
sole cause of the *early* divergence from Ipopt on the problem below, though —
see "What this does not fix" — it is not the whole story for that problem.

## Reproducer

A large direct-collocation optimal-control NLP, solved through CasADi's
`nlpsol` with `hessian_approximation=limited-memory` and `linear_solver=ma57`:

```
Total number of variables............:    59939
Total number of equality constraints.:    58271
Total number of inequality constraints:      33
```

Solver settings are attached as `config_pounce.json` and `config_ipopt.json`.
The two are the same tolerances, scaling and barrier settings under a different
plugin key; both cap `max_iter` at 60 for the side-by-side, and Ipopt runs at
`print_level 6` so the update decisions are printed.

Attached logs:

| File | What it is |
| --- | --- |
| `ipopt_reference_60it.log` | Ipopt, `print_level 6`, 60 iterations — the reference |
| `ipopt_reference_full_converged.log` | Ipopt, uncapped — `EXIT: Optimal Solution Found`, 171 iterations, `inf_du` 9.22e-05 |
| `pounce_BEFORE_fix_60it.log` | pounce @ 642c36f, 60 iterations, `[LBFGS]` instrumentation |
| `pounce_AFTER_fix_60it.log` | pounce with the attached patch, same run |

## Evidence

Both solvers were instrumented to print every curvature pair. They compute
**bit-identical** `s^Ty`, `snrm`, `ynrm` and make **identical skip decisions**.
The only difference is what happens after two consecutive skips:

```
Ipopt                                                              pounce @ 642c36f
------------------------------------------------------------------ ----------------------------------
Limited-Memory approximation started; store data at current iterate.
s^Ty = 2.727392e+00 snrm = 2.727720e+00 ynrm = 4.165026e+01         sy=2.72739188647343722e0  accept=true
  Perform the update. skipping: 0                                    hist=1 sigma=3.665623575e-1
s^Ty = 1.595978e+01 snrm = 1.368773e+01 ynrm = 3.140744e+02         sy=1.59597770485976511e1  accept=true
  Perform the update. skipping: 0                                    hist=2 sigma=8.518520749e-2
s^Ty = -2.136187e-01 snrm = 1.950495e+00 ynrm = 3.960641e+02        sy=-2.13618734172891317e-1 accept=false
  Skip the update.    skipping: 1                                    hist=2  (unchanged)
s^Ty = -2.065615e-01 snrm = 1.741449e+01 ynrm = 8.638597e+02        sy=-2.06561512657842672e-1 accept=false
  Skip the update.    skipping: 2                                    hist=2  (unchanged)
Resetting Limited-Memory Update.            <-- never happens        hist=2  (stale, for the rest of the run)
```

Over the same 60 iterations:

| | updates performed | updates skipped | **resets** |
| --- | --- | --- | --- |
| Ipopt | 29 | 21 | **9** |
| pounce @ 642c36f | 48 | 12 | **0** |
| pounce + patch | 30 | 22 | **8** |

Ipopt's `Wr` info-string appears on iterations 5, 8, 11, 14 and 23. Note that
`Wr` / `Ws` are only emitted at `print_level >= 6`; at level 5 they are silently
suppressed, which makes this easy to miss.

### Iterate sequences

Before the patch the two solvers are bit-identical through iteration 5 and split
at iteration 6 — exactly one iteration after Ipopt's first reset:

```
 it |      objective    inf_pr    inf_du    mu     ||d||     a_pr   || pounce @ 642c36f
  0 |  2.2904024e+02  1.06e+01  1.01e+02  -1.0  0.00e+00 0.00e+00  || identical
  ...
  5 |  1.3681753e+02  6.44e+00  4.92e+02  -1.5  6.43e+00 6.04e-02  || identical
  6 |  1.3312028e+02  5.85e+00  3.11e+03  -1.5  7.06e+00 9.90e-02  || 1.2557280e+02 8.44e+00 3.78e+02 -1.5 8.39e+00 2.32e-01   <-- DIVERGE
```

`lg(rg)` is `-` and `ls` is `1` on both sides through the split, so neither the
regularization path nor the line search is involved.

With the patch applied, pounce reproduces Ipopt to 7-8 significant figures on
every printed quantity out to iteration 18, instead of splitting at iteration 6:

```
 it |      objective    inf_pr    inf_du    mu     ||d||     a_pr   || pounce + patch
  6 |  1.3312028e+02  5.85e+00  3.11e+03  -1.5  7.06e+00 9.90e-02  ||  1.3312027e+02  5.85e+00  3.11e+03  -1.5  7.06e+00 9.90e-02
 12 |  1.2925531e+02  3.40e+00  4.80e+02  -1.5  8.25e+00 1.80e-01  ||  1.2925531e+02  3.40e+00  4.80e+02  -1.5  8.25e+00 1.80e-01
 18 |  1.0356563e+02  4.98e+00  1.64e+02  -1.5  3.54e+00 1.91e-01  ||  1.0356562e+02  4.98e+00  1.64e+02  -1.5  3.54e+00 1.91e-01
```

The residual last-digit differences are floating-point summation-order noise.
The two sequences separate at iteration 19 (relative objective difference
1.5e-06, growing from there) for reasons we have not yet identified.

## What this does not fix

We want to be explicit that the patch is **not** sufficient to make our problem
converge, so nobody reads more into it than the evidence supports.

Running the patched build to 250 iterations:

```
status            = MaximumIterationsExceeded
iteration_count   = 250
final_objective   = 74.638356
final_dual_inf    = 1.5500e-01
final_constr_viol = 3.3577e-04
final_compl       = 1.4064e-05
```

Ipopt on the same problem exits `Optimal Solution Found` after 171 iterations
with `inf_du` 9.22e-05. The patched pounce run reaches essentially the same
objective — 74.6384 against Ipopt's 74.6482 at its iteration 170, a relative
difference of 1.3e-04 — but its `inf_du` oscillates in the 5e-02 to 3e-01 band
and never approaches the 1e-03 `dual_inf_tol`.

Notably, that 0.155 plateau is **unchanged** from the unpatched build. So the
missing reset is real and worth fixing on its own merits, but the endgame
stationarity stall on this problem has a different cause that we are still
chasing. We are reporting the reset because the evidence for it is clean and
self-contained, not because it resolves our symptom.

## What we changed locally

The patch below is **what we tried in order to confirm the diagnosis**, not a
proposal for how pounce should solve this. The final shape of the fix — where
the counter lives, how the reset interacts with the `nonlinear_vars` mask, what
the test surface should look like — is entirely the maintainer's call, and we
have no attachment to this particular implementation. We are happy to rework it,
or to have it thrown away in favour of something that fits the codebase better.
If it is more useful as evidence than as code, treat it as evidence.

Attached as `0001-fix-lbfgs-reset-the-approximation-after-limited_memo.patch`
(applies to `642c36f`). 4 files, +47/-8:

* `LimMemQuasiNewtonUpdater` gains `max_skipping` (default 2) and `skipped_iter`,
  mirroring upstream's `limited_memory_max_skipping_` and `lm_skipped_iter_`.
  `ingest_pair` already returns whether the pair was accepted; that return value
  was being discarded and now drives the counter.
* `update_hessian` clears the history and falls through to the existing
  unconditional re-anchor once the counter reaches the threshold.
* `AlgorithmBuilder::limited_memory_max_skipping` plus a read site in
  `IpoptApplication::algorithm_builder_from_options`, alongside the existing
  `limited_memory_max_history` handling.
* `limited_memory_max_skipping` removed from `no_silent_options.rs`'s
  `known_debt`, as that test requires once an option gains a read site.
* `application_limited_memory_options_propagate_to_builder` extended to assert
  the default (2) and a user-supplied value (4) reach the builder.

`cargo test --release -p pounce-algorithm --features ma57` on the patched tree:
436 unit tests pass, and `no_silent_options`'s
`every_registered_option_is_read_or_declared_unimplemented` passes with the
option removed from `known_debt`.

One deliberate difference from upstream, flagged because it is the kind of
shortcut a maintainer may not want: upstream's reset branch re-anchors inside
the reset itself, whereas this patch relies on the re-anchoring that
`update_hessian` already performs unconditionally at the end of every call. The
observable effect appeared to be the same and it kept the diff smaller, but
mirroring upstream literally is obviously also defensible.

## Two related observations (separate from the above, impact not measured)

These were found while diffing against `IpLimMemQuasiNewtonUpdater.cpp` and are
*not* addressed by the patch, but both cite upstream behaviour that does not
match the referenced source:

1. **Skip tolerance.** `bfgs_curvature_pair_ok` (`lim_mem_quasi_newton.rs:679-682`)
   hardcodes `eps = 1e-8` and attributes it to upstream "~line 750". The actual
   test is `IpLimMemQuasiNewtonUpdater.cpp:995-1005`, and it uses
   `tol = sqrt(machine_epsilon)` ~= 1.4901e-08, not 1e-8.

2. **Powell damping.** `powell_damping_theta` (`:663-674`, coefficient 0.2) is
   applied in `build_low_rank:478`, with a doc comment saying it "Mirrors
   upstream's `IpLimMemQuasiNewtonUpdater.cpp:PowellDamping`". No such function
   exists in Ipopt; `CheckSkippingBFGS` takes `const Vector&` for both `s_new`
   and `y_new` and only returns a bool, so upstream cannot and does not damp.
   The `.hpp:305` comment mentioning Powell damping appears to be stale.
   Empirically, damping first alters the iterate sequence at iteration 10 on
   this problem; disabling it did **not** affect the iteration-6 divergence, so
   it is unrelated to the bug above.

### LINKED ISSUE #688: `recalc_y` regularizes the least-squares multiplier system; upstream does not — dual convergence stalls at scale
**Version:** pounce 0.10.0 @ `9caf48416283fb6dd8cb1d1260bb0a9d16aaf997`
**Component:** `crates/pounce-algorithm/src/eq_mult/least_square.rs`
**Reference:** Ipopt @ `2695946fa79d2e84f3034e065e788933a81466eb`, `IpLeastSquareMults.cpp:80-81`
**Related:** review item **M3** in `dev-notes/code-review-progress.md` (which introduced the perturbation), #686 (L-BFGS reset — merged, necessary but not sufficient here)

> **AI disclosure.** This report — the investigation, the log analysis, the attached patch and this text — was produced with the help of an AI coding assistant, then reviewed by a human before filing.

## Summary

`LeastSquareMults::calculate_y_eq` builds the W=0 augmented system with `delta_c = delta_d = 1e-8`. Upstream passes `0`:

```cpp
// IpLeastSquareMults.cpp:80-81
retval = augsyssolver_->Solve(GetRawPtr(zeroW), 0.0, NULL, 1.0, NULL, 1.0,
                              GetRawPtr(J_c), NULL, 0.,
                              GetRawPtr(J_d), NULL, 0.,
                              *rhs_x, *rhs_s, *rhs_c, *rhs_d,
                              *sol_x, *sol_s, y_c, y_d,
                              check_NegEVals, numberOfEVals);
```

Every other argument matches pounce's line for line — `rhs_x = ∇f + Pₗz_L − Pᵤz_U`, `rhs_s = Pₗv_L − Pᵤv_U`, `rhs_c = rhs_d = 0`, `delta_x = delta_s = 1.0`, `w_factor = 0.0`, `check_NegEVals = ProvidesInertia()`, `numberOfEVals = dim(c) + dim(d)`. The perturbation is the only difference we could find.

On a large direct-collocation NLP this is enough to make `recalc_y = yes` *degrade* the multipliers instead of improving them, and dual convergence plateaus. Setting the two deltas to `0` was the last change required to get the model to solve — the first time pounce has reached `Optimal Solution Found` on it for us, after the #686 reset fix on its own left it stalled.

This call site *did* originally pass `0.0`. The `1e-8` was introduced later, by review item M3, as a precaution copied from a sibling call site; the attached patch is a revert of M3 at this one site rather than a new divergence from the port. That history is worth reading before acting on any of this, so it has its own section below.

**We are reporting the change that produced convergence, not proposing an implementation.** How — or whether — to act on this is entirely yours; see "The deliberate decision" below, which we think is the crux.

## Mechanism

The (3,3)/(4,4) blocks of this system are structurally zero; that is ordinary saddle-point structure and `ma57` factors it with inertia control. Adding `δ` there turns the equality-constrained least-squares problem into a Tikhonov-regularized one, so the returned `y` is damped by roughly `O(δ / σ_min(J)²)` along the weakest singular directions of the constraint Jacobian.

Where `σ_min(J)² ≫ 1e-8` the estimate is numerically unchanged — which is the case for every NLP in the test suite, and is presumably why the in-code comment's claim that the perturbation "keeps the LS solution numerically identical" held up in practice. On a ~59 000-variable / ~58 000-equality collocation problem the two are comparable, and the damping is no longer negligible.

The damped `y` is then written straight back into the iterate, so the error appears directly in `∇ₓL` — i.e. in `inf_du`, the quantity the run is being judged on. Because `recalc_y` recomputes `y` the same biased way every iteration, nothing corrects it; the bias is a fixed point of the estimator rather than a transient.

## Evidence

Identical option sets throughout (`hessian_approximation=limited-memory`, `limited_memory_max_history=6`, `recalc_y_feas_tol=0.05`, `max_iter=250`), `linear_solver=ma57` except where noted; the two `solverOptions` nodes are attached.

| build | `recalc_y` | `δ_c`,`δ_d` | iterations | final `inf_du` | result |
|---|---|---|---|---|---|
| Ipopt `2695946` | yes | 0 | **171** | 9.22e-05 | **Optimal** |
| Ipopt `2695946` | no | — | 250 | 1.15e-02 | Max iterations |
| pounce `642c36f` + #686 reset (local) | yes | 1e-8 | 250 | 1.55e-01 | Max iterations |
| pounce `642c36f` + #686 reset (local) | no | — | 250 | 1.73e-02 | Max iterations |
| pounce `642c36f` + #686 reset (local) | yes | **0** | **203** | 1.79e-04 | **Optimal** |
| pounce `9caf484` + patch, `ma57` | yes | **0** | **220** | 1.86e-04 | **Optimal** |
| pounce `9caf484` + patch, **`feral`** | yes | **0** | **188** | 5.21e-04 | **Optimal** |

Rows 3 and 4 are the diagnostic tell: turning `recalc_y` **off** improved the final dual infeasibility ninefold and landed pounce next to Ipopt-without-recalc_y. A feature intended to sharpen the multipliers was the thing holding them back, which is what pointed at the solve rather than at the L-BFGS model.

Row 6 is stock `9caf484` with only the attached two-line patch applied, built with `--features pounce-algorithm/ma57`. Converged run against the Ipopt reference:

|  | Ipopt `2695946` | pounce `9caf484` + patch |
|---|---|---|
| Status | Optimal Solution Found | Optimal Solution Found |
| Iterations | 171 | 220 |
| Objective | 7.4648600022558227e+01 | 7.4621273559181745e+01 |
| Dual infeasibility | 9.2232118434186768e-05 | 1.8597077014402975e-04 |
| Constraint violation | 2.3726604201148097e-04 | 1.1351607880069792e-04 |
| Overall NLP error | 2.3726604201148097e-04 | 1.8597077014402975e-04 |

We have not run stock `9caf484` with `δ = 1e-8` on this model to 250 iterations; rows 3-5 were measured on the `642c36f` base carrying a local equivalent of the #686 reset. The `least_square.rs` blob is byte-identical between `642c36f` and `9caf484` (`45b71ab88a1580298675582447e1002110e3196f`), so we do not expect the difference to matter, but we would rather flag the gap than paper over it.

## The deliberate decision, and where it came from

The perturbation is not an oversight. `dev-notes/code-review-progress.md` records it as review item **M3**, "algorithm: `LeastSquareMults` lacks the δ_c/δ_d inertia workaround its sibling has", and the note is candid about both the reasoning and its limits. Paraphrasing it:

- Before M3, `calculate_y_eq` passed `delta_c = delta_d = 0.0` — i.e. it matched `IpLeastSquareMults.cpp`. M3 classified that as "duplicate logic that diverged" from the dual initializer in `init/default.rs:154-194`, which solves the same W=0 system and perturbs by `1e-8` because pounce-feral's LDLᵀ mis-reports the inertia of a structurally-zero (3,3)/(4,4) block — 0 negative eigenvalues on `nuffield2_trap` against a true `n_c + n_d`, raising `WrongInertia`. A spurious failure makes `calculate_y_eq` return `false`, and the caller (`init/default.rs:387-390`) treats that as "solver failed → leave at zero", appending `y0` to the info string and starting from `y_c = y_d = 0`.
- The fix was to mirror the sibling's `1e-8`, with a cross-reference comment asking future edits to keep the two sites in sync.
- The note states plainly that **no fail-first test was constructible**: the trigger is feral's data-dependent mis-report on a CUTEst matrix not present in the repo, and the aug-solver unit harness drives `DenseMock`, an exact LU oracle that reports correct inertia regardless of `delta`.
- Safety was argued from the suite staying green (323 passed, 0 failed), "confirming the `1e-8` perturbation is numerically inert on every covered problem".

We think that last line is where our result attaches, and we want to be precise about the claim we are making. M3's inertness argument is correct **on every problem it was checked against** — it holds exactly while `σ_min(J)² ≫ δ`, which is true of the whole in-repo corpus. It is a statement about the covered problems, not a scale-free property, and the measurements above are the first case we know of that leaves that regime.

So we are not asking you to discard M3's reasoning — and note that the feral result below does **not** discharge it either, since we never reproduced M3's actual trigger. What we can say is narrow: **M3's justification is entirely about feral's inertia reporting, and neither backend needed the help on this model.** Whether that means reverting M3 outright, gating the perturbation on the linear solver, fixing feral's inertia detection, or making `δ` scale-relative is a judgement we are not in a position to make from outside. The attached patch reverts M3 at this one call site only because that is the minimal change that isolates the diagnosis; please treat it as evidence, not as a proposal.

We deliberately left the sibling perturbation at `init/default.rs:171,174` untouched — it is on the initializer path, where the fallback-to-zero failure mode above is the real risk, and our problem converged without touching it.

## Why the test suite would not catch it

Offered as an observation, not a request:

- Every NLP in the suite is small enough that `σ_min(J)² ≫ 1e-8`, so the regularization genuinely is invisible there. The failure needs scale.
- `tests/optimize_hs71_lbfgs.rs:237` (`recalc_y_changes_the_lbfgs_trajectory`) asserts `assert_ne!(it_off, it_on)` and that the objective stays near 17.014017 — it checks that `recalc_y` *changes* the trajectory, not that it improves it. The doc comment explains this is deliberate, since re-estimating `y` is not uniformly better. That is a defensible choice, and it is also why a `recalc_y` that actively degrades the multipliers passes CI.

## feral with `δ = 0`

Since M3's justification is specifically about feral, we re-ran the same model on the same patched build with `linear_solver=feral` and nothing else changed. **It converges: 188 iterations to `Optimal Solution Found`, `inf_du` 5.21e-04** — fewer iterations than the ma57 run, at roughly 2× the wall time per iteration (8.6 s/iter against 4.15 s/iter, plausibly the `OrderingMethod::Auto` → AMD choice against our `ma57_pivot_order=5` METIS, though we have not tried to confirm that). The run's solve report records `"solver_name": "feral"`, and the ma57 and feral iterate sequences agree to eight significant figures before diverging at iteration 22 on rounding, which is what we would expect from two exact factorizations of the same system.

No `y0` appears in the info strings at any iteration, and there is no `WrongInertia` or inertia-correction message — that is, `calculate_y_eq` never once failed under feral with the perturbation removed.

We want to be careful about what this does and does not establish. It says the LS solve is well-conditioned enough on **this** problem that feral factors it without the `1e-8`. It is **not** a test of M3's actual trigger, which is feral's data-dependent inertia mis-report on `nuffield2_trap` — a matrix we do not have. If that matrix is available to you, it is the case worth running before acting on any of this, and we would be glad to run it if it can be shared.

The `pounce-algorithm` suite is also green on `9caf484` + the patch: 438 unit tests and ~50 integration binaries, 0 failures. Because `constr_mult_init_max` defaults to `1e3 > 0`, every constrained solve in that suite traverses `calculate_y_eq`, so these tests do exercise the changed line. That is the mirror image of M3's own evidence — M3 argued `1e-8` was inert because the suite stayed green, and the suite stays equally green at `0`.
## Attachments

- [`0001-fix-recalc_y-solve-the-least-squares-multiplier-syst.patch`](https://github.com/user-attachments/files/31199459/0001-fix-recalc_y-solve-the-least-squares-multiplier-syst.patch) — the two-line change, against `9caf484`
- [`config_pounce.json`](https://github.com/user-attachments/files/31199457/config_pounce.json), [`config_ipopt.json`](https://github.com/user-attachments/files/31199461/config_ipopt.json) — the `solverOptions` nodes only
- [`ipopt_reference_full_converged.log`](https://github.com/user-attachments/files/31199460/ipopt_reference_full_converged.log) — Ipopt, 171 iterations, Optimal
- [`pounce_stall_recalc_on.log`](https://github.com/user-attachments/files/31199456/pounce_stall_recalc_on.log) — `δ = 1e-8`, 250 iterations, `inf_du` 1.55e-01
- [`pounce_stall_recalc_off.log`](https://github.com/user-attachments/files/31199455/pounce_stall_recalc_off.log) — `recalc_y = no`, 250 iterations, `inf_du` 1.73e-02
- [`pounce_9caf484_delta0_converged.log`](https://github.com/user-attachments/files/31199462/pounce_9caf484_delta0_converged.log) — `9caf484` + patch, `ma57`, 220 iterations, Optimal
- [`pounce_9caf484_delta0_feral_converged.log`](https://github.com/user-attachments/files/31199458/pounce_9caf484_delta0_feral_converged.log) — `9caf484` + patch, `feral`, 188 iterations, Optimal

Model source and solution vectors are omitted deliberately; the logs carry the iteration history and the summary blocks but no primal or dual values.

jkitchin:
## Summary

Started as "L-BFGS σ is wrong" (#677) and grew, through four user reports, into
the L-BFGS fidelity PR. Six defects, each measured against upstream source
rather than inferred from behaviour.

Fixes #677
Fixes #684
Fixes #686
Fixes #688

> **The before/after numbers in this body were measured before the merge with
> `main` (`d38167a`, 35 commits including the #588 quadratic-structure work).**
> The per-commit "exact leg identical 57/57" claims were each verified at the
> time and still hold, and the final commit's sweep is identical 114/114 against
> the *post*-merge baseline. But the specific iteration counts and objectives in
> the tables below are pre-merge and should be read as indicative — `main` moved
> some of these fixtures itself, `eigenb2` among them (`14f0468`). Re-deriving
> them means a two-leg sweep per trajectory-affecting commit.

## What was wrong

**1. σ used `scalar2` where Ipopt uses `scalar1`.** `limited_memory_initialization`
was registered and never read, so the code ran `scalar2` unconditionally while
the registry advertised `scalar1`. Reported by @srikanth-gm with side-by-side
KKT dumps.

**2. `recalc_y` was registered and refused.** Implemented, structurally matching
upstream's `AcceptTrialPoint` block, including the `"y "` iteration marker that
was missing. Left opt-in: auto-enabling under limited-memory, which upstream
does, regressed 7 fixtures here.

**3. `linear_system_scaling=slack-based` parsed and did nothing.** A catch-all
match arm routed it to `None`. Now a real port of `IpSlackBasedTSymScalingMethod`.

**4. Restoration densified the L-BFGS Hessian — a 14 GB allocation** (#684).
`n(n+1)/2` triplets at n = 59,956 aborts the process rather than failing the
solve. Restoration now hands the orig block to the inner solver in factored
form whenever that solver applies a low-rank `W` by Sherman-Morrison-Woodbury.

**5. L-BFGS never reset its approximation, and damped every pair it kept**
(#686). `limited_memory_max_skipping` was registered, unread, and the feature
behind it did not exist — over 60 iterations of the reporter's model Ipopt
reset 9 times and pounce 0. Separately, `build_low_rank` blended each accepted
`y` toward `B·s` by a Powell damping factor citing
`IpLimMemQuasiNewtonUpdater.cpp:PowellDamping`, a function that does not exist;
checked for a deliberate reason and found the opposite, in
`dev-notes/issue-131-monotone-lbfgs-stall.md`, which already named the damping
as a *cause* of a stall.

**6. `recalc_y` degraded the multipliers it exists to sharpen** (#688). The
least-squares solve perturbed its `(3,3)`/`(4,4)` blocks by `1e-8` where Ipopt
passes `0`. Eliminating `w` gives `y = −(J Jᵀ + δ I)⁻¹ J·rhs_x`, so the
perturbation returns a Tikhonov-regularized multiplier damped by
`O(δ / σ_min(J)²)` — invisible while `σ_min(J)² ≫ δ`, `O(1)` on a 59,000
variable collocation NLP. Because `recalc_y` recomputes `y` identically every
iteration the bias is a *fixed point*, and it lands in `inf_du`:
`recalc_y=yes` stalled at `1.55e-01`, **worse** than `recalc_y=no` at
`1.73e-02`. At `δ=0` the reporter's model converges on both MA57 and FERAL —
the first time pounce has solved it.

### Why #688's fix is scoped

An unscoped `δ=0` was measured and rejected: it **inverts gh#540**. With the
initializer unregularized, that issue's `feral_inertia_pivot_floor` trigger
makes `eigena2` worse on the fixture it was built for — `3.097e-08` with the
trigger on against `2.2e-08` with it off. Shipping that means shipping a build
where disabling a documented fix improves the case it exists to serve.

Only one of the four callers has the problem #688 describes:

| caller | frequency | bias | cost of a failed solve |
|---|---|---|---|
| dual init, warm-start init, restoration inner | once | transient — Newton corrects it | `y = 0`, iteration-0 `inf_du` blow-up ← what M3 guarded |
| **`recalc_y`** | every iteration | **fixed point**, lands in `inf_du` | benign — Newton multipliers stand |

The unscoped change is *measurably better* on most of the corpus and fixes #524
at its cause, so it is worth doing — but it and the gh#540 pivot floor are
coupled through `y₀` and must be re-derived together. Filed as **#693** with the
measurements; **#524 stays open** with its root cause now named.

## Trajectory changes

Per CLAUDE.md, every moved line is accounted for. The **exact-Hessian leg is
identical, 57/57**, on every commit up to the damping removal; #688's commit is
identical on **both** legs, since `recalc_y` is off by default.

| fixture | before | after | exact reference |
| --- | --- | --- | --- |
| `cresc4` | RestorationFailed @976 | **SolveSucceeded @106** | 0.8718975273 |
| `cresc4`, `recalc_y=yes` | **InfeasibleProblemDetected @32** (false) | **SolveSucceeded @109** | 0.8718975273 |
| `hs13_bigstart` | 0.9945139766 | **0.985014399** | 0.9849287152 |
| `pooling_rt2stp` | 413 it | **203 it** | — |
| `issue_508_gap_1em2` | certified from a blown-up iterate | **certified at obj 25** | obj 25 |

Two lines move the wrong way and were run down rather than absorbed:

* `autocorr_bern55-06` is nonconvex and unstable under any perturbation —
  sweeping `limited_memory_max_skipping` 1→50 gives −2336, −2256, −2252, −2368,
  −2368, and none is the exact reference of −2304.
* `deb7` does not solve under L-BFGS at any setting; only the exact path does.
* `eigenb2` downgrades to `Solved_To_Acceptable_Level`.
  `docs/src/initialization.md` already documents this fixture doing exactly this
  under an unrelated change (gh#616) — it sits on the tolerance band at
  objective 1.6 and any perturbation can tip it.

**A caution on `pooling_rt2stp`:** its L-BFGS-leg outcome is chaotic. A 1%
change to `limited_memory_init_val_max` — a clamp — flips it between
`RestorationFailed`@26 (obj −10.2), `SolveSucceeded`@171 (−4391.8) and
`ErrorInStepComputation`@215 (−3273.9). Improvements quoted for that fixture on
that leg should be read as draws, not effects.

## Infeasibility: a false verdict, found and fixed

`cresc4` reported `Infeasible_Problem_Detected` for a feasible model at a point
with constraint violation 13.47 and restoration KKT error 1.95e2. The existing
divergence guard is *entry-relative*, so it is blind when the blow-up predates
restoration entry. Added the absolute companion — a point cannot be the best
feasibility available when this same solve has already reached one 3.5 million
times better.

## Guard against the whole class

`no_silent_options.rs` fails the build for any registered option that is neither
read nor declared unimplemented. Four of the six defects above were this shape,
and the scan turned an estimate of ~24 such options into a measured 151. The
remaining debt is inventoried in #682; this PR clears
`limited_memory_initialization`, `limited_memory_init_val`, `recalc_y`,
`linear_system_scaling=slack-based` and `limited_memory_max_skipping`.

## Test plan

- [x] CI green on the tip (`70e91e2`), 11/11, against `main` at `d38167a`
- [x] `cargo test`: pounce-algorithm 710, pounce-cli 561, pounce-restoration
      144, pounce-linsol 24, pounce-feral 40 — 0 failed
- [x] `cargo fmt --all -- --check` clean
- [x] Fixture sweep, both legs, on every trajectory-affecting commit; final
      commit identical 114/114 against the post-merge baseline
- [x] Operator-equivalence test for the factored restoration block (#684):
      every basis direction, agreement to 1e-12
- [x] Behaviour tests, not propagation tests: `recalc_y_changes_the_lbfgs_trajectory`,
      `slack_based_scaling_changes_the_solve`, `mc19_still_falls_back_to_no_scaling`
- [x] `saddle_point_inertia_is_correct_without_regularization` — feral counts the
      unregularized saddle-point shape correctly, which was M3's premise for
      adding the perturbation #688 removes
- [x] Upstream cross-checks fetched from coin-or/Ipopt stable/3.14 source for
      σ, `recalc_y`, slack-based scaling, the reset, the skip tolerance, the
      absence of Powell damping, and the LS multiplier deltas

### One test threshold changed, with the measurement

`test_halving_costs_materially_less_across_the_suite` went red because
`pounce.minimize` with a `jac` and no `hess` runs L-BFGS. Measured per model
rather than tuned until green:

| model | solver_evals | iters | user_evals |
| --- | --- | --- | --- |
| `_hs71` | 0.819 | 0.731 | 0.828 |
| `_rastrigin_eq` | **1.021** | 0.845 | 1.012 |
| `_double_well` | 0.983 | 0.808 | 1.043 |
| **TOTAL** | **0.920** | **0.783** | **0.908** |
| margin needed | 0.90 | 0.85 | 0.95 |

The ladder is not less effective: `iters` — the measure it acts on — clears its
own stricter margin at 0.783 with every model improving. What changed is the
margin that had no basis: `solver_evals` and `user_evals` are the same kind of
measure, diluted by the same per-rung fixed cost, and the test bounded
`solver_evals` *more* strictly (0.10) than `user_evals` (0.05). `git log -S`
finds that asymmetry arriving with no rationale. Both now carry 0.05.


jkitchin:
@srikanth-gm — opening this so you can follow it. It's a **draft / WIP**: what's described in the body is in and measured, but two things are still going onto this branch.

**Still to come:**

1. **`linear_system_scaling=slack-based` implemented.** Right now it only *warns* that it is falling back to no scaling. Since it's in your golden configuration, warning is the stopgap, not the fix — a run of yours that asked for slack-based scaling has been getting none.
2. **The rest of #551.** Auditing all 353 registered options against every accessor form turned up 24 more that are registered, never read, and not declared unimplemented — so they accept a value, do nothing, and say nothing. `tau_min`, `s_max`, `fixed_mu_oracle`, `theta_min` and the filter/line-search constants are among them. Worth checking your configs against that list once it lands; `slack-based` and `limited_memory_initialization` were both in this category and both were in yours.

**Already in and measured:** σ defaulting to `scalar1`, `recalc_y` implemented (opt-in — see the body for why we diverge from Ipopt's conditional default), the L-BFGS sweep leg, and the scaling probe.

Nothing here needs anything from you. If you do get a chance to run your model against the branch, the number worth reporting is whether the first 25 iterations track your Ipopt baseline more closely than before — but the sanitised-fixture question from #677 is the more valuable one long-term, since it's what would let us catch the next one of these ourselves.

---

srikanth-gm:
## Report
- Some early test results with the understanding that this is a draft PR
- I tested commit `642c36f9bbcb40d96a73985f2cf755b5723e4eff` which has the slack-based scaling implemented
- I can confirm both `recalc_y` and `slack-based` options work now
- Quite close to IPOPT Golden Configuration as mentioned in #677
- However the problem is still stalling in duals
- This seems to have surfaced another issue - it fails on hitting restoration

## Configuration
```json
	"solverOptions": {
		"verbose": true,
		"pass_nonlinear_variables": true,
		"pounce": {
			"print_level": 5,
			"max_iter": 2000,
			"solver_selection": "nlp",
			"presolve": "yes",
			"hessian_approximation": "limited-memory",
			"linear_solver": "ma57",
			"linear_system_scaling": "slack-based",
			"ma57_pivot_order": 5,
			"ma57_pre_alloc": 2.0,
			"nlp_scaling_method": "gradient-based",
			"nlp_scaling_min_value": 0.001,
			"mu_strategy": "monotone",
			"constr_viol_tol": 0.001,
			"dual_inf_tol": 0.001,
			"compl_inf_tol": 0.0005,
			"tol": 0.001,
			"acceptable_constr_viol_tol": 0.005,
			"acceptable_dual_inf_tol": 0.01,
			"acceptable_compl_inf_tol": 0.001,
			"acceptable_obj_change_tol": 0.005,
			"acceptable_tol": 0.005,
			"limited_memory_max_history": 6,
			"recalc_y": "yes",
			"recalc_y_feas_tol": 0.05,
			"mu_target": 1e-06,
			"mu_superlinear_decrease_power": 1.00000001,
			"mu_linear_decrease_factor": 0.33,
			"barrier_tol_factor": 1000.0,
			"mu_init": 0.1
		}
	}
```

## Iteration Log
[iteration-log.log](https://github.com/user-attachments/files/31190431/iteration-log.log)

jkitchin:
Thanks — and that OOM is a real bug, not a resource limit. We can name the line.

## The allocation is a dense n×n

`memory allocation of 14379225712 bytes failed` is not a large problem needing more RAM. Divide it out:

```
14,379,225,712 / 8 bytes  =  1,797,403,214 entries
n(n+1)/2 = 1,797,403,214  →  n = 59,956
```

That is your variable count (59,939 plus pounce's compressed `x_var` space after presolve). It is a **dense lower triangle of the full Hessian**, allocated in one go.

It comes from `materialize_orig_block` in `crates/pounce-restoration/src/aug_resto_system_solver.rs:547`, on the path taken when restoration is entered while `hessian_approximation=limited-memory`. Restoration cannot consume the low-rank L-BFGS `W` directly, so that function densifies it into a `SymTMatrix` — building `n(n+1)/2` entries for `irows`, `jcols` and `vals`.

The comment above it states the assumption that fails:

> Cost is `O(rank·n_orig²)`; restoration is a rare heavyweight fallback and `n_orig` is small, so this is negligible.

`n_orig` is not small here. At 60k variables it is ~1.8 billion entries per array, so the solve dies the moment restoration is entered. Every L-BFGS user large enough to need restoration hits this; it is not specific to your model beyond its size.

Worth being straight about how we missed it: the scaling probe added on this PR measured the limited-memory path as linear from n=2,000 to n=128,000, and reported no hidden quadratic. That measurement was real but incomplete — none of the probe problems enter restoration, so the probe never exercised this path. A clean linear result on the main iteration said nothing about the fallback, and we presented it with more confidence than it earned.

## What this means for your run

The dual stall and the OOM are two different problems, and the OOM is downstream: the solve stalls, stalls trigger restoration, restoration allocates 14 GB and dies. Fixing the allocation will not fix the stall — it will change the failure from a crash into whatever restoration actually does on your model, which is the thing we then need to look at.

We'll take the allocation as its own issue. The fix is to stop densifying: the low-rank structure is already there and restoration should consume it in factored form, which is `O(n·rank)` instead of `O(n²)`.

## On the rest

Good to have it confirmed that `recalc_y` and `slack-based` both work and that you are close to the golden configuration — that is the part we could not test here, since nothing in our corpus is this shape.

Two notes on your configuration:

- `linear_system_scaling: "slack-based"` is now honoured, but `linear_scaling_on_demand` defaults to `yes`, which computes scaling only once a factorization already looks troubled. If you want it applied from the start, as your Ipopt baseline effectively does, set `linear_scaling_on_demand: "no"` too.
- You dropped `ma57_automatic_scaling` from this run relative to the earlier one. Probably deliberate, but it is a difference from the golden config worth being aware of when comparing.

The standing ask from #677 has not changed and this is a good illustration of why: a sanitised or synthetic model of this shape would become a fixture here, and both of these — the restoration blowup and the dual stall — would be things we find rather than things you hit.

---

srikanth-gm:
## Report
- We noticed stall on our problem with POUNCE with the identical config converging in IPOPT 
- I ran some experiments with AI and found a possible issue with how IPOPTs `limited_memory_max_skipping` is implemented in POUNCE
- Made #686 for full context and logs

srikanth-gm:
## The End Zone
- Prof. @jkitchin we are happy to report that we have achieved convergence on POUNCE with **both** MA57 and FERAL :)
- Its great to see FERAL be competitive to MA57 in runtime perf and achieve convergence in lower iterations on such a large problem (since that is our real prize here)
- The fix that finally made it exit was a 2-line change
- full context and logs for your evaluation in #688

jkitchin:
Wow, great news! and thanks for all the debugging help! I am converging on a solution in pounce for all this that will hopefully land this evening.

jkitchin:
@srikanth-gm this should be very close to a working solution for you I hope! There are still a few more corners to polish on my end, and some benchmark tests but hopefully I didn't miss anything major and this is the end for this specific issue! Thanks for the feedback and debugging along the way, it was super helpful at getting to this point. Feel free to file a new issue if it doesn't get there.

srikanth-gm:
Prof. @jkitchin - just to close the loop we tested `main` at `68d6298cab7a516e26c3bc11844522d559d0e8a2` and our NLP is converging

thanks a lot for your time and patience :)

[pounce_solver_config.json](https://github.com/user-attachments/files/31206373/pounce_solver_config.json)
[pounce_iteration_log.log](https://github.com/user-attachments/files/31206374/pounce_iteration_log.log)

image

jkitchin:
This is great! I would love to see some comparisons with the solvers you usually use. If there is a problem you can share for benchmarking / testing that would also be helpful. The best way to do that is an an nl file, but a python script would also work.

srikanth-gm:
Prof. @jkitchin good timing, now that we have convergence, we can do apples-to-apples benchmarks 
- I have posted #698 with benchmarks against IPOPT 
- There are findings (with evidence) that might possibly bring the speed of POUNCE on-par with IPOPT on large problems like ours

## Background
- The model we're optimizing here is a production-grade multibody vehicle model on a  [differentiable track surface](https://www.tandfonline.com/doi/full/10.1080/00423114.2025.2525240) 
  - this gets compiled as a [DAE-FMU](https://arxiv.org/pdf/2606.22544) in a residual formulation
  - only uses C-API calls to get directional derivatives (only analytical jacobians, no analytical hessians)
- We are using the `DaeBuilder` feature of CasADi to build a [Framework](https://elib.dlr.de/213372/1/Konferenz-Vortrag_PDF.pdf) on top of it for transcription and efficiently assembling `nlp_f`, `nlp_g`, `nlp_grad_f`, `nlp_jac_g` for the NLP solver along with the residuals as NLP constraints
- Since the model and framework are both proprietary (probably SOTA) we can't directly share it in public
  - you would also need the entire framework to be able to run the optimization so its not one file
  
## NLP
- Transcribed as a direct collocation problem with 1000 grid points
- Currently the NLP is around 60-80K variables but is expected to get to 100K+ as we add more states to the model (currently around 50)
- I believe our problem is fairly non-convex because there are a lot of constraints (FIA rules) to satisfy and has risk of local minima (not-an-expert)
- I saw your previous note and are trying to see how we can make a synthetic version of this problem size or share something in private (I have to get some approvals)

## Solvers
- IPOPT - this is our main solver with either MA57 or MUMPS and solves in reasonable time
- POUNCE - only other solver outside of IPOPT that has solved it in comparable iterations :)
- KNITRO - solves but takes 4x iterations to exit (circling-the-drain near optimal solution)
- SNOPT - wont solve unless we give it an optimal solution as guess

Issue #685, opened 2026-08-18
A cancelled quadratic coefficient is still read out for evaluation and for routing: Q4 evaluates the row as zero, the classifier calls the model an LP


jkitchin:
## Summary

gh #683 is fixed on `feat/588-quadratic-structure` (`792c98a3`): the *degree* answer, `NlBody::provably_affine`, no longer reads "affine" off a quadratic term map that floating-point arithmetic emptied. Two consumers of the same cancelled coefficients are **not** fixed by that, and each is its own silent wrong answer:

1. **Evaluation** — `NlBody::admitted_quad_form` (`crates/pounce-nl/src/nl_reader.rs:369`) still reads the cancelled form out, and Q4's constant-structure evaluator evaluates the row from it. **Introduced by Q4 (`947b3aa3`), so this is on PR #603, not on `main`.**
2. **Routing** — the ungated `analyze_quadratic` at `crates/pounce-cli/src/dispatch.rs:383` classifies such a row as linear, so a model made of them routes **LP**. Pre-existing on `main`; noted in #683 §"Suggested fix".

Both were found while fixing #683 and are reported rather than folded into it: closing either changes which bodies Q4 and the classifier *reach*, which is a trajectory-visible change and a larger decision than the demotion #683 needed.

## Why the exactness gate does not stop it

Q4's gate is `is_expanded_quadratic`, and its rule is stated as: a form is admitted only when the recognizer's read-out **does no algebra the tape would not have done**. A flat sum of monomials qualifies. But summing several monomials that share a key *is* algebra the tape did not do — the tape multiplies each term by `x` and then sums; the read-out sums the coefficients first. When those coefficients cancel, the two are not near each other, they are unrelated.

The gate reasons about *factored versus expanded*. It has no clause about a coefficient that was summed.

## Reproduction

One row, `2⁵³·x₀² + x₀² − 2⁵³·x₀² ≤ 5`, written as an `o54` sumlist over three monomials on the same key (the middle one spelled `x₀^2` so the tape does not hash-cons all three into one node). The true body is `x₀²`. `min −x₀`, `0 ≤ x₀ ≤ 10⁶`.

```
$ pounce max.nl                                   # default route
pounce: problem class LP — every nonlinear part expanded to a linear (or constant) polynomial
Objective...............:  -1.0000000000000000e+06     # the row vanished

$ pounce max.nl solver_selection=nlp               # forced NLP, Q4's fast path
[...]

Issue #683, opened 2026-08-18
provably_affine can report a quadratic row as affine when its coefficients cancel, freezing that row's Jacobian for the whole solve


jkitchin:
## Summary

`NlBody::provably_affine` (`crates/pounce-nl/src/nl_reader.rs:417`) is documented and used as a **proof of degree**, but it derives that answer from floating-point-summed coefficients with exact zeros dropped. A genuinely degree-2 body whose quadratic coefficients cancel to exactly `0.0` reports `Some(true)` — proved affine.

Q6 (`8c0c2e27`, PR #603) uses this to decide `jac_c_constant` / `jac_d_constant`, so an affected row's Jacobian is **evaluated once and reused for the entire solve**. That is a silent wrong answer on a default route, with no `POUNCE_DBG_*` and no option overrides.

Found by a cumulative correctness review of PR #603. Blocks that PR.

## Mechanism

```rust
pub fn provably_affine(&self) -> Option<bool> {
    match self {
        NlBody::Tree(e) => {
            if is_trivially_zero(e) { return Some(true); }
            recognize_expr(e).map(|q| q.quadratic().is_empty())
        }
        NlBody::Quad(q) => Some(q.form.quadratic().is_empty()),
    }
}
```

`quadratic()` is populated through `is_live` (`crates/pounce-nl/src/nl_quadratic.rs:253`):

```rust
/// Is this coefficient worth storing? Exact zeros are dropped so that
/// "has a quadratic term" is a structural question; ...
fn is_live(c: f64) -> bool { c.abs() > 0.0 }
```

The comment asserts the property that fails. `is_live` is applied to the **sum** of a row's quadratic coefficients, so "has a quadratic term" is an arithmetic question, not a structural one. Two routes reach a false `Some(true)`:

1. **Cancellation.** `o54 3` over the same monomial with coefficients `−2^53, 1, +2^53`. The recognizer folds back-to-front: `2^53 + 1` rounds to `2^53`, then `− 2^53` gives exactly `0.0`. True body is `x²`.
2. **Underflow.** `(1e-200·x)·(1e-200·x)` is a monomial by `is_monomial`; its coefficient `1e-400` flushes to `0.0`. True body is `1e-400·x²`.

## Why Q6's safety argument does not cover this

[...]

PR #692, opened 2026-08-18
nl: flag the inexact fold, not the drop (#687)

### LINKED ISSUE #687: Quad2::dropped_terms conflates an exact cancellation with a lossy one — flag the inexact fold instead
## Summary

`Quad2::dropped_terms` records that the recognizer dropped a coefficient. It
does not record whether the arithmetic that produced that coefficient was
**exact**, and those are different questions. Because gh#683 and gh#685 both
gate on the flag, a body that cancels exactly is refused alongside one that
cancels catastrophically — same flag, same refusal, and only one of them had
to be refused.

## The two cases the flag cannot tell apart

- `x − x` → `fl(1) + fl(−1) = 0`, and that add is **exact**. The body really
  is degree 0. Nothing was lost.
- `2⁵³·x + x − 2⁵³·x` → `fl(2⁵³ + 1) = 2⁵³` **loses the `1`**, and only then
  does `2⁵³ − 2⁵³ = 0` drop the term. The body is `x`; the form says nothing.

The loss happens at the *inexact add*, not at the drop. The drop is only where
it becomes visible.

## What this costs today

Refusing both is sound and was the right first move — it is what closed two
wrong-answer defects on the default route. But it costs reach:

- gh#683: a row with an exactly-cancelling term is no longer *provably affine*,
  so its Jacobian is re-evaluated every iteration instead of once.
- gh#685 part 1: the same row goes back to the AD tape instead of being
  evaluated from its (correct, empty) matrices.
- gh#685 part 2: a model whose only nonlinearity cancels exactly now classifies
  **NLP** instead of **LP** — it loses the convex path entirely. This is the
  expensive one. `dispatch::tests::lp_reasons_distinguish_absent_from_cancelled`
  pins the current behaviour and says why.

Nothing in the fixture corpus is affected — no fixture body drops a term at
all, which is why none of the three sweeps moved — so this is a latent cost on
models we do not currently have, not a measured regression on ones we do.

## The sharper gate

Flag the **inexact fold** rather than the drop. A single float add can be
checked exactly with a two-sum:

```text
s  = a + b
bb = s - a
err = (a - (s - bb)) + (b - bb)      // err == 0.0  ⟺  the add was exact
```

Set the flag when `err != 0.0` (and keep setting it on the underflowing/NaN
multiply paths, where the same argument applies: `10⁻²⁰⁰ · 10⁻²⁰⁰` is an
inexact multiply, not an exact one that happens to reach zero). Then an exact
cancellation leaves the flag clear, all three consumers keep their fast paths,
and the catastrophic case is still refused — for the right reason, at the
point where the information is actually lost.

Cost: two extra flops per coefficient merge in the recognizer, on a path
`qssp180` runs 65 341 times. Worth measuring against the sweep before
adopting; the recognizer's own timings are in
`dev-notes/quadratic-structure-exploitation.md`.

## Acceptance

- The exact-cancellation cases in gh#683's and gh#685's test files keep their
  fast paths (degree proved, form admitted, model routes LP/QP).
- The `2⁵³` and `10⁻²⁰⁰` reproductions in both files still refuse, unchanged.
- `scripts/sweep-fixtures.sh` and the `Problem class:` sweep over the loadable
  `.nl` files are diffed before merge — this is a trajectory-visible change by
  construction, since it hands fast paths back.

jkitchin:
## Problem

`Quad2::dropped_terms` records that the recognizer dropped a coefficient. It does not record whether the arithmetic that produced that coefficient was **exact**, and those are different questions — so a body that cancels exactly is refused alongside one that cancels catastrophically:

| body | the fold | maps | truth | pre-#687 verdict |
|---|---|---|---|---|
| `x − x` | `fl(1) + fl(−1) = 0`, **exact** | empty | degree 0, nothing lost | refused |
| `2⁵³·x² + x² − 2⁵³·x²` | `fl(2⁵³ + 1) = 2⁵³`, **inexact**, then an exact cancel | empty | degree 2, `x²` lost | refused |

Refusing both is sound and closed two wrong-answer defects (#683, #685 part 1). What it costs is reach on the first row, on every consumer at once: not provably affine (its Jacobian is re-evaluated every iteration instead of once), not evaluated from its own — correct, empty — matrices, and once the classifier gates on the same bit (#685 part 2, merged in #691) it loses the LP/QP route as well.

Nothing in the fixture corpus is affected in either direction, so this is a latent cost on models we do not have, not a measured regression on ones we do.

## Cause

The loss happens at the **inexact add**, not at the drop. `fl(2⁵³ + 1) = 2⁵³` is where the `x²` goes; the `− 2⁵³` that follows is exact and only makes the loss visible. A float add can reach zero *only* by exact cancellation, so the drop site itself carries no information about whether anything was lost — the information is one or more folds upstream.

## Fix

`Quad2` carries two bits instead of one:

* **`inexact`** — some arithmetic behind these coefficients rounded. Clear ⇒ every stored coefficient is exactly the real-arithmetic value of the terms folded into it.
* **`lost_terms`** — a term may be missing: something was dropped *while* `inexact` was set, or a live coefficient was flushed to zero (or `NaN`) outright. This is the bit the consumers read, under the contract `dropped_terms` had.

Three exactness tests feed the first, none of them a tolerance:

| operation | test |
|---|---|
| `a + b` | Knuth two-sum — exact for every finite pair, 4 flops per merge |
| `a · b` | `fma(a, b, −p) == 0`, with `±1`/`0` answered directly (every variable enters with coefficient `1.0`) and a non-normal product — underflow, subnormal, overflow — answered *inexact* |
| `a / d` | the recognizer scales by `fl(1/d)`, the reciprocal only for a power of two: `fma(r, d, −1) == 0` |

Two smaller holes closed with them, both from `as_constant` reading a number out of a form and leaving its flags behind: a `Div`'s divisor and a `Pow`'s exponent now hand their `lost_terms`/`inexact` to the result. Dividing by a constant that swallowed a variable term is not dividing by a constant.

`dropped_terms` → `lost_terms` is a rename on purpose: the meaning changed, and a call site that does not compile is better than one that silently keeps reading the old question.

**Rejected: flagging every inexact add**, which is what the issue proposes literally. It is sound and gives back less than it costs — `0.1·x² + 0.2·x²` rounds, drops nothing, and has complete maps, so under that rule it would lose the matrix evaluation and the LP/QP route for a form that is missing nothing. The drop is still necessary; what changed is that it is no longer sufficient.

**Kept: one pair of bits per form, not per monomial.** Q3's `Quad2` allocates nothing per term and the recognizer is affordable at 65 341 rows because of it. The price is that this answer is conservative: an inexact fold on one coefficient makes an exact cancellation on an unrelated one count as lost.

## Blast radius

Baseline `d38167a` (the parent commit), release builds both sides:

* `scripts/sweep-fixtures.sh` over the 57 fixtures — **diff empty**.
* `Problem class:` over the 61 loadable `.nl` files in the repository — **diff empty** (4 LP, 24 convex QP, 2 convex QCQP, 1 nonconvex QP, 30 NLP; `idaes_helmholtz.nl` wants `AMPLFUNC` and does not load here, as before).

Bit-identical except where the flag fires, and no fixture body drops a term at all — the same reason #683's and #685's own reach costs measured zero. What moves is synthetic material that does drop terms:

| measurement | parent | this branch |
|---|---|---|
| ill-scaled degree battery (9 000 bodies): rows demoted for a lost term | 499 | **302** (197 regain the proof) |
| …rows proved affine / proved degree 2 | 5 500 / 1 812 | 5 697 / 1 812 |
| evaluator differential battery: seeds skipped | 204 of 1 500 | **7 of 1 500** |
| …problems compared against their tapes | 1 296 | 1 493 (4 479 quadratic rows, 54 132 Hessian entries, **0 not bit-identical**) |
| …worst `g`/Jacobian relative deviation | `4.385e-15` | `4.774e-15` (pinned tolerance `1e-14`) |

**Cost of the extra flops.** `qssp180` could not be re-measured — the bench corpus is not on this machine — so the worst case was measured synthetically instead (release, 200 recognitions, three runs a side): a body that is nothing but merges (20 000 monomials onto 64 keys) goes 7.00/7.04/7.15 ms → 7.72/7.79/7.80 ms, **+10%**; a product-heavy `(Σ⁶⁴ xᵢ)²` is unchanged within noise, because the `±1` shortcut answers almost every product without an `fma`. The sweep's wall clock shows no change.

**What admitting these rows turned up.** A class of body reached #683's battery that never had before, and it found two things — neither a defect in the claim, both the *tape* being ill-conditioned on a body the recognizer reads exactly:

* **seed 565** — `x₀²·((x₁ − x₁)·x₁)·10³⁰⁰` is identically zero and is proved so; its tape multiplies before it sums, so at `x₀ = 10⁸` the adjoint reaches `∞` and then `∞ · 0 = NaN` where every finite probe reads `0`.
* **seed 1293** — a row whose adjoint accumulation sums `±10³⁰⁸` contributions into a running `−10³⁰⁰`, so the constant it lands on drifts `4.5e-12` with `x`. The only drift in 3 000 seeds.

The battery now skips probe entries where the tape's own arithmetic left the finite range (415 of them) and bounds the drift at `TAPE_WOBBLE_REL = 1e-9` — three orders above what was observed, and far below what a *degree* defect moves (those take a derivative off zero, or scale it by `x`). This is the opposite failure from #683's: there the recognizer was wrong about a body the tape had right.

## Tests

New, and failing on the parent commit for the stated reason:

* `issue_683…::an_exactly_cancelled_quadratic_row_is_still_proved_affine` — `provably_affine` reads `None` pre-fix, on both parse arms; also asserts `DerivativeProof::Constant` on the row, so the claim is checked and not just the flag.
* `issue_685…::an_exactly_cancelled_row_is_still_admitted_for_evaluation` — `admitted_quad_form` is `None` pre-fix; the admitted form's `g` is checked against the tape's.
* `nl_quadratic` unit tests on the fold itself: the exact and the absorbed cancellation, the absorption flagged *before* anything drops, rounding carried across subexpressions into a later drop, cancellation inside a product (`(x₀+x₁)(x₀−x₁)` has no cross term and is not demoted for noticing), the underflowing product, the exactly-cancelled versus absorbed constant, a rounded scale that loses nothing, and the two-sum / two-product tests themselves.

Unchanged and still passing, i.e. the refusals this sharpens are still refusals: every `2⁵³` and `10⁻²⁰⁰` case in both issue files, on both parse arms, including the partial-cancellation model and the CLI A/B against `POUNCE_DBG_NO_QUAD=1`.

Not pinned: `qssp180`'s recognizer timing (no bench corpus on this machine — the synthetic worst case above stands in for it).

## Merged with `main` after #691 landed

This branch opened from `d38167a`, nine minutes before #691 merged gh#685 part 2, so the green run above never saw the classifier gating on this flag. `main` is now merged in, and the merge is the interesting part: `git merge` reports it clean and the result **does not compile** — #691 added four `dropped_terms()` call sites in `nl_reader.rs`, which is exactly what renaming the accessor was meant to catch.

Resolved by carrying #691's gate onto `lost_terms()` and by taking the two `dispatch.rs` assertions this PR's sharper gate is *meant* to flip. Both now assert **both** sides rather than merely relaxing: `x0² − x0²` (exact) routes `Lp`/`NonlinearPartsCancelled`, and a `2⁵³·x0² + x0² − 2⁵³·x0²` companion routes `Nlp`/`ObjectiveTermsDropped`. `ClassReason::explain()`'s wording moved from "dropped a term to floating-point cancellation" — no longer what the gate tests — to "lost a term to an inexact floating-point fold or a flush to zero".

Re-run on the merge result, against the post-#691 `main` baseline:

* `scripts/sweep-fixtures.sh`, 57 models — **byte-identical**.
* `Problem class:` over all 62 loadable `.nl` files — **identical**. (My earlier baseline held 61: `crates/pounce-wasm/tests/simple.nl` was dropped building that list. Checked against the `main` binary directly — same class.)
* `cargo test --workspace`: **3 198 passed, 0 failed**.
* `cargo fmt --all --check` clean; CI's clippy invocation (`-D clippy::correctness -D clippy::suspicious`) clean.

Fixes #687

---

- [ ] Book page under `docs/src/` updated — none applies; this is internal to the recognizer, and the trajectory-visible record is `dev-notes/quadratic-structure-exploitation.md`, which gains the section

https://claude.ai/code/session_01CkLW9jjXX98sQjFTQBkhNV

---
_Generated by [Claude Code](https://claude.ai/code/session_01CkLW9jjXX98sQjFTQBkhNV)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -163,9 +163,9 @@ changes.
   records that a term was dropped and not whether the arithmetic that
   dropped it was exact. Nothing in the corpus is affected — no fixture
   body drops a term — but a model that spells a linear row that way loses
-  the LP path. #687 tracks the sharper gate (flag the inexact *fold*,
-  which is where `2⁵³ + 1` loses the `1`, rather than the drop it leads
-  to), which would give those models their fast path back.
+  the LP path. #687 is the sharper gate (flag the inexact *fold*, which
+  is where `2⁵³ + 1` loses the `1`, rather than the drop it leads to);
+  the next entry lands it and hands those models their fast path back.
 
 - **…and a row whose coefficients cancel *exactly* keeps both fast paths**
   (#687, sharpening the two above).
@@ -177,8 +177,8 @@ changes.
   coefficient map is a lower bound. `2⁵³·x² + x² − 2⁵³·x²` loses its `x²`
   at `fl(2⁵³ + 1) = 2⁵³`, an inexact add, and only then cancels. Both set
   the same flag, so both were refused, and the first one paid a proved
-  degree, a matrix evaluation, and (once the classifier gates on it) the
-  convex route for arithmetic that never rounded.
+  degree, a matrix evaluation, and — once the classifier gated on it too
+  — the LP/QP route, all for arithmetic that never rounded.
 
   The gate is now the **fold** rather than the drop: a form records
   whether its arithmetic rounded — a two-sum on every coefficient add, an
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,34 @@ changes.
   Part 2 of #685 — a model of this shape being classified LP — predates
   this work and is unchanged here.
 
+- **…and a row whose coefficients cancel *exactly* keeps both fast paths**
+  (#687, sharpening the two above).
+
+  The flag those two fixes read said a coefficient had been **dropped**,
+  which is a different question from whether anything was lost. `x − x`
+  folds `fl(1) + fl(−1)` to zero and that add is exact — the term really
+  is absent, the body really is degree 0, and nothing about the
+  coefficient map is a lower bound. `2⁵³·x² + x² − 2⁵³·x²` loses its `x²`
+  at `fl(2⁵³ + 1) = 2⁵³`, an inexact add, and only then cancels. Both set
+  the same flag, so both were refused, and the first one paid a proved
+  degree, a matrix evaluation, and (once the classifier gates on it) the
+  convex route for arithmetic that never rounded.
+
+  The gate is now the **fold** rather than the drop: a form records
+  whether its arithmetic rounded — a two-sum on every coefficient add, an
+  FMA check on every product, and one on the reciprocal a `Div` scales by
+  — and a dropped term counts as missing only when something behind it
+  did. An underflowed or `NaN` coefficient still counts on its own; there
+  is no exact case to spare there.
+
+  No model in the corpus changes: the 57-fixture sweep and the
+  `Problem class:` line over every loadable `.nl` are both byte-identical
+  to the parent build, because no fixture body drops a term at all. What
+  moves is synthetic material that does: 197 rows of the ill-scaled
+  degree battery get their proof back, and the quadratic evaluator's
+  differential compares 197 more generated problems against their tapes
+  (bit-identical Hessians, worst `g`/Jacobian deviation `4.8e-15`).
+
 - **`least_square_init_primal=yes` costs one accepted downgrade on the
   fixture corpus, not two** (#681, revising the gh#616 measurement).
 
--- a/crates/pounce-cli/src/dispatch.rs
+++ b/crates/pounce-cli/src/dispatch.rs
@@ -156,11 +156,14 @@ pub enum ClassReason {
     ObjectiveNotQuadratic,
     /// Row `row`'s nonlinear part is not a degree-2 polynomial.
     ConstraintNotQuadratic { row: usize },
-    /// The objective is degree ≤ 2, but the recognizer dropped a term
+    /// The objective is degree ≤ 2, but the recognizer *lost* a term
     /// reaching its coefficients, so they are not the whole objective.
     ObjectiveTermsDropped,
-    /// Row `row` is degree ≤ 2, but the recognizer dropped a term reaching
+    /// Row `row` is degree ≤ 2, but the recognizer *lost* a term reaching
     /// its coefficients, so they are not the whole row (gh #685).
+    ///
+    /// "Lost", not merely "dropped": a term that cancels exactly is not
+    /// missing from the form, so those rows keep the fast path (gh #687).
     ConstraintTermsDropped { row: usize },
     /// The sense-adjusted objective Hessian has a negative eigenvalue.
     ObjectiveHessianIndefinite,
@@ -198,14 +201,15 @@ impl ClassReason {
                 format!("row {row}'s nonlinear part is not a degree-2 polynomial")
             }
             ClassReason::ObjectiveTermsDropped => {
-                "the objective's quadratic form dropped a term to floating-point \
-                 cancellation or underflow, so its coefficients are not the whole \
-                 objective"
+                "the objective's quadratic form lost a term to an inexact \
+                 floating-point fold or a flush to zero, so its coefficients are \
+                 not the whole objective"
                     .to_string()
             }
             ClassReason::ConstraintTermsDropped { row } => format!(
-                "row {row}'s quadratic form dropped a term to floating-point \
-                 cancellation or underflow, so its coefficients are not the whole row"
+                "row {row}'s quadratic form lost a term to an inexact \
+                 floating-point fold or a flush to zero, so its coefficients are \
+                 not the whole row"
             ),
             ClassReason::ObjectiveHessianIndefinite => {
                 "the objective Hessian (sense-adjusted for minimization) is not PSD".to_string()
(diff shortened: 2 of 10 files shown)


jkitchin:
CI is red on `d6a0e94` and expected to be: that commit is the base merge of #691 **without** the follow-up that reconciles the two. #691 put the flag in front of `analyze_quadratic_full` and calls it `dropped_terms`, which this branch renames to `lost_terms`, so the workspace does not compile — every job fails at the build, not at a test.

The reconciliation is written and green locally (see below); it cannot be pushed from this session because git push authentication stopped working partway through the work — `fatal: could not read Username for 'https://github.com'` on every attempt for the last ~15 minutes, while fetches over the same remote still succeed. The base merge on this branch was made through the API for that reason. **Do not merge this PR until the follow-up commit lands.**

What the follow-up does, measured against `2118871`:

- `quad_terms_dropped` → `quad_terms_lost`, `ClassReason::{Objective,Constraint}TermsDropped` → `…TermsLost`, and `explain()` names the loss (a coefficient that rounded and then cancelled, or one that underflowed) rather than the drop.
- The reach #691 recorded as its accepted cost comes back: `x₀·x₀ − x₀^2 ≤ 5` routes **LP** again and the CLI solves it as one — the row really is `0 ≤ 5`, so the bound really is the optimum — while `2⁵³·x₀² + x₀² − 2⁵³·x₀² ≤ 5` still routes NLP and still says why. Two new tests in `issue_685_cancelled_quadratic_routing.rs`, both failing on the parent with `Nlp`/`ConstraintTermsLost`; two dispatch unit tests go back to `Lp` with the reason asserted, not just the class.
- `scripts/sweep-fixtures.sh` (57 fixtures) and `Problem class:` over the 61 loadable `.nl` files: both **diff-empty** against `2118871`. `cargo test --workspace --exclude pounce-hsl`: 3183 passed / 0 failed. `cargo fmt --all --check` clean.

I will keep retrying the push and update here when it lands. If it stays broken, the commit can be taken from this session's checkout of `claude/issue-687-pu26fa` and pushed from a machine with credentials.

---

Issue #689, opened 2026-08-18
Direct convex driver diverges on `scaled_feasible_a`, which HSDE solves in 16 iterations


jkitchin:
Found while measuring the blast radius of #588 Q9a (Gondzio correctors, PR on `feat/588-q9-correctors`). **Pre-existing — not caused by that PR**, which is why it is filed rather than folded in, but it was invisible until a sweep at `qp_hsde=no` put a number on it.

## What happens

`crates/pounce-cli/tests/fixtures/scaled_feasible_a.nl`:

| route | result |
|---|---|
| default (`qp_hsde=yes`, HSDE) | `SolveSucceeded`, **16 iterations**, obj 236.85 |
| `qp_hsde=no` (direct driver, `ipm.rs::run_ipm`) | `MaximumIterationsExceeded`, **199 iterations**, obj 1.14e11 |

At the cap the direct driver is not near anything: `final_kkt_error` 8.37e45, `final_dual_inf` 3.69e41, `final_constr_viol` 1.14e4. It is diverging, not converging slowly.

Reproduce, at `d38167a9` or later:

```
pounce crates/pounce-cli/tests/fixtures/scaled_feasible_a.nl /dev/null \
  --json-output /tmp/a.json qp_hsde=no
```

`scaled_feasible_b` is worth checking at the same time; it converges on both routes today but is the other member of the pair and is objective-unstable under any trajectory perturbation (see below).

## Why it matters beyond one fixture

The direct driver is not a debug-only path. It is what `QpWarmStart`, the build-once `QpFactorization` handle, and the dual-infeasibility reverify guard all use, and those routes cannot fall back to HSDE the way the one-shot path can. A model of this shape reaching a warm-start caller gets the 199-iteration divergence with no second chance.

## A related observation, same fixture pair

`scaled_feasible_a` and `scaled_feasible_b` also move their *reported objective* by a large factor under trajectory perturbations that leave the status at `SolveSucceeded`:

* `qp_gondzio_corr=0` at default options: `a` 236.85 → 315.67, `b` 456.33 → 378.03.
* An experimental adaptive-τ tail in HSDE (#690): `a` 236.85 → 272.94, `b` 456.33 → 953.36.

Either these two have genuinely flat or degenerate optima — in which case the fixtures are poor trajectory sentinels and the sweep should know it — or the success criterion is passing on iterates that are not optimal. Worth deciding which, since `scripts/sweep-fixtures.sh` treats an objective move as signal.

## Not blocking

[...]

Issue #690, opened 2026-08-18
Adaptive τ tail in the HSDE driver: −4.1% from the τ/κ ray, blocked by a 20× trajectory regression on `scaled_feasible_a`


jkitchin:
> **Superseded by the re-measurement in [this comment](https://github.com/jkitchin/pounce/issues/690#issuecomment-5348502811) (2026-08-19).** Everything below the fold was measured *through* the stopping-test defect that #689/#696 removed, and none of its numbers survive it. It is kept verbatim as the record of how the question was originally posed. Read the comment, or `dev-notes/issue-690-hsde-tau-tail.md`, for the current state.

## Current state (post-#696)

The study asked for by `QpOptions::tau_max`'s doc comment, run as part of #588 Q9a and **still not shipped**. Re-run on `ac18ba6` across all 58 CLI fixtures, both sweep legs, against a control that sweeps bit-identical to `main`.

**The fork this issue was filed on is resolved.** The body below asked whether the three ~2× objective movers had flat optima or whether the success test was accepting iterates it shouldn't, and said the second would outrank the τ question entirely. It was the second: #696 found HSDE normalizing the duality gap by an objective displaced by the model's degree-0 term (`Σaᵢ² ≈ 5e11` on this pair). All three fixtures now report their true optimum of 0, and the corpus baseline moved 2082 → 2280 on the same 57 fixtures.

| variant | exact leg | lbfgs leg | excl. `scaled_feasible_a` |
|---|---|---|---|
| off | 2292 | 5438 | — |
| orthant | 2369 (+3.4%) | 5515 (+1.4%) | +0.0% / +0.0% |
| both | 2278 (−0.6%) | 5424 (−0.3%) | **−4.1% / −1.7%** |

* The −3.0% does not reproduce. The τ/κ ray is no longer merely where the gain is — it is where *all* of it is; `orthant` alone buys nothing.
* The original objection is gone: every objective that moves under `both` moves *toward* the exact optimum (`5.000000002 → 5`, `−5.999999998 → −6`, `1.000000001 → 1`). No fixture returns a materially different answer.
* **The new blocker:** `scaled_feasible_a` goes 123 → **2474** iterations, with status and objective both intact at the end — the gh#544 signature. It is a stall, not a divergence: by iteration 199 it has reached `final_kkt_error` 3.2e-07, four orders *better* than the 4.6e-03 point the baseline certifies at 123, then wanders non-monotonically for two thousand more before the stopping test accepts it.

[...]

PR #694, opened 2026-08-19
588 (Q9): Gondzio correctors on the direct convex driver, a shared `correctors.rs`, and cone-aware presolve for convex QCQPs

jkitchin:
Q9 of the #588 series has two unrelated changes under one label. **This PR is both of them**, at the user's request that Q9 land as a single unit:

* **Q9a — Gondzio correctors** on the direct convex driver, and a shared `correctors.rs`. Commits `8e48ca6b`, `66edd58f`, `8c6a72bb`.
* **Q9b — presolve validity for quadratic rows** (§7), and the `presolve_conic` wiring Q1 deferred here. Commits `52ada9f4`, `61021da6`, `f04b6152`, `c3cceea6`.

They share a branch and nothing else; each half is measured on its own and the two are kept separable by commit so a bisect can tell them apart.

### What could not be measured — stated up front, and it applies to both halves

**The benchmark corpus is gone.** `$POUNCE_BENCH_DATA` is unset, its default path does not exist, `plato.asu.edu` is blocked by egress policy, and nothing was downloaded. **`qcqp1000-2nc` — the single instance whose Q0(b) trace justified this entire topic — could not be run in this container**, and no number in this PR stands in for it. Every figure below is labelled **real** (in-repo fixture) or **generated** (`scripts/gen-qcqp-nl.py`).

---

# Q9a — Gondzio correctors on the direct convex driver

## Problem

Two things were wrong before this PR:

1. The Gondzio multiple-centrality-corrector scheme lived only inside `hsde.rs`'s embedding loop, tangled with that loop's `(τ, κ)` Schur solve and its single symmetric step length. The **direct** driver — `ipm.rs::run_ipm`, the route taken by warm starts, the build-once `QpFactorization` handle, the dual-infeasibility reverify guard and `qp_hsde=no` — had no correctors at all, and could not grow them without duplicating the band arithmetic or depending on `hsde`.
2. The scheme was hard-coded at 3 correctors with **no way to switch it off**, so a corrector regression in the field had no bisection handle.

NETLIB `afiro`, the small degenerate LP the scheme is literally designed for, on the direct driver:

| | status | objective | KKT error | iters |
|---|---|---|---|---|
| before | `SolveSucceeded` | −464.7531419 | 2.10e-7 | **135** |
| after | `SolveSucceeded` | −464.7531426 | **1.35e-8** | **118** |

## Cause

`hsde.rs:866–945` held the whole scheme as an inline loop: Gondzio's centered box `[β_lo·μ, β_hi·μ]`, the projection of the trial step's complementarity products into it, the enlarged trial length `min(α + δ, 1)`, and the acceptance test `α̃ ≥ α + γδ`, all interleaved with the τ-row Schur solve. Those four pieces are driver-independent; everything they were tangled with is not.

The direct driver differs in exactly one relevant way: its step is **split** primal/dual where HSDE's is symmetric. That is a parameter of the projection, not a reason for a second implementation.

## Fix

Two commits, deliberately split so a bisect can tell them apart.

**`8e48ca6b` — pure refactor.** `pounce_convex::correctors` gets `Band` (with `project`/`deviation` as separate primitives), `project_products` taking a `(primal, dual)` trial pair, `trial_step`, `accepts`, and the tuning constants. `hsde.rs` calls them. The `(τ, κ)` row, the Schur complement and the step-length rule stay with the driver that owns them.

One bit the move could have flipped and does not: the τκ target is still `v.clamp(lo, hi)` computed directly, never recovered as `v − deviation(v)`. Those differ in floating point once `v` sits many orders above `hi`, which is precisely the regime a corrector fires in — hence the two primitives, with the doc comment saying why.

**`66edd58f` — behaviour.** `run_ipm` runs the scheme, on the same `cone.is_orthant()` gate. A corrector is accepted only when **both** step lengths gain at least `γδ`. Gondzio's rule is stated per-length and a split step has two of them; a corrector that lengthens one while shortening the other trades a known gain for an unknown loss, and this driver's residuals are not symmetric in the two. The conservative conjunction is what ships.

New option **`qp_gondzio_corr`** (integer 0–10, **default 3**), read by *both* drivers. New `POUNCE_DBG_GONDZIO=1`, one line per convex solve: driver, iterations, correctors attempted, correctors accepted, mean step gain. Every count in this PR came out of it.

**`8c6a72bb` — follow-up analysis**, correcting two things the second commit left as open questions. See *Findings that did not become code* below.

### The default is ON, and the argument for it

* The default route **cannot be affected**: it runs HSDE, and the option's default *is* HSDE's old hard-coded constant. Three sweeps of 57 fixtures at three option sets, all byte-identical.
* The route that *is* affected has its own sweep (below): net −16 iterations, no status change, no objective outside tolerance, KKT error improving on both lines whose count moved.
* Off is one option away and is **proven exact**: `qp_hsde=no qp_gondzio_corr=0` reproduces `d38167a9` byte-for-byte.

### Measured

**Real, in-repo.** `lp_afiro` (NETLIB `afiro`) — the table above, plus HSDE 15 → **13**.

**Generated — labelled generated, every time.** `gen-qcqp-nl.py --quad-rows 0`, n = 400, 1 200 linear rows, 100 equalities, seed 7:

| driver | off | on | correctors accepted | mean α gain |
|---|---|---|---|---|
| direct | 12 | **10** | 17 / 22 (77.3%) | 4.24e-2 |
| HSDE | 17 | **15** | 22 / 31 (71.0%) | 4.16e-2 |

Wall clock 0.54 → 0.50 s and 0.75 → 0.75 s: inside this machine's ~1.3× noise floor, and **not** reported as a speedup. A corrector change is judged by iteration count.

### The justification instance would not have been served by this code anyway

That finding did not need the corpus, and it is the most important thing in the Q9a half. Q0(b)'s corrector-shaped stall on `qcqp1000-2nc` was traced on the **NLP** path — `line-search trials / iter` and `median α_primal` are filter-line-search quantities and the convex IPM has no filter. Q9's correctors are in `pounce-convex`. Different solvers.

Both routes were checked on generated instances:

* A generated `qcqp500-3c` shape classifies **`NLP`** — *"convex QCQP downgraded: cone reformulation costs 1 250 000 000 flops (budget 20 000 000)"*. It never reaches `run_ipm` or `solve_conic_hsde`; `POUNCE_DBG_GONDZIO` prints nothing because no convex driver ran. That is Q1's own `SOCP_REFORM_FLOP_BUDGET` working as designed — Q1 measured the conic path 2.7–3.8× *slower* on `nql180`/`qssp180`.
* A generated QCQP small enough to pass both guards *does* route conic, reaches HSDE, and reports `corrected_iters=0 attempted=0`. The SOCP reformulation is what makes it conic, so the cone carries SOC blocks, and the orthant gate never opens.

**So the corrector half of Q9 cannot serve the `qcqp*` family by either route on the current routing**, corpus or no corpus. §9's Q9 paragraph is corrected in place. What it does deliver is the "bonus" clause of that same sentence — orthant LP/QP through both drivers — plus a module a SOC-capable corrector can be built in. That is the change QCQP would need, and it is spectral machinery (the Jordan product is not elementwise and cannot be box-projected componentwise), not tuning.

### Adaptive τ — already shipped, and the HSDE study is measured but not shipped

Adaptive τ on the direct driver's corrector step is **gh #417**, shipped long before this series, exposed as `qp_tau`/`qp_tau_max`. There was nothing to build. What `tau_max`'s doc comment records as open is HSDE. That study was run here as a throwaway patch behind an env var, swept, and **reverted**:

| variant | total iterations (57 fixtures) | status changes |
|---|---|---|
| off (shipped) | 2082 | — |
| adaptive τ, orthant blocks | 2074 (−0.4%) | none |
| adaptive τ, orthant blocks **and** τ/κ ray | **2019 (−3.0%)** | none |

Not shipped, for two reasons stated in `§0h`: the entire gain lives in the τ/κ ray, which is what `on_infeasibility_ray` compares — so τ → 1 there touches the certificate logic, not just the step length; and the driver it would change is the default cold path for every LP/QP/SOCP, whose real coverage (NETLIB LP, Maros–Mészáros QP) is exactly what the missing corpus held. Shipping that on 57 small fixtures is the gh #544 pattern.

## Blast radius — Q9a

`scripts/sweep-fixtures.sh` against a binary built from **`d38167a9`**, 57 fixtures, 0 `NO_JSON`, comparing **status, objective AND iteration count**:

| option set | moving lines |
|---|---|
| **default** | **0** |
| **`nlp_scaling_method=none`** | **0** |
| **`mu_strategy=adaptive`** | **0** |
| `qp_hsde=no` | 5 |
| `qp_hsde=no qp_gondzio_corr=0` | **0** (attribution control) |

The last row is why the five are readable: switch the correctors off and the direct-driver sweep returns to byte-identical, so every moving line belongs to the correctors and to nothing else.

The **refactor commit alone** (`8e48ca6b`) is byte-identical at **all four** option sets, which is what makes it provably behaviour-preserving. Note that `tests/conic_hsde_vs_direct.rs` — the guard §9 names — is **not sufficient on its own here**: it drives SOC and PSD cones and the corrector loop is gated on `cone.is_orthant()`, so it never enters the moved code. It is green; the sweep is the evidence. The sweep does reach it: 30 of the 57 fixtures route to the convex driver (4 LP, 26 convex QP).

The five, each explained individually — no summarising:

| fixture | iters | final KKT error | reading |
|---|---|---|---|
| `lp_afiro` | 135 → **118** | 2.10e-7 → **1.35e-8** | the win. NETLIB `afiro`, the degenerate-LP family Gondzio's paper is about. Count and accuracy improve together, so the shorter trajectory is not bought by stopping earlier at a worse point. |
| `wyndor_min` | 7 → **8** | 1.69e-8 → **8.98e-10** | the only count that goes **up**, by one, while landing 19× more accurate: objective −35.99999997 → −36. The trajectory took one more step and reached a strictly better point. Recorded, not accepted as a cost — there is nothing to own. |
| `bound_active_qp` | 5 → 5 | 5.76e-10 → 7.34e-9 | count unchanged. A different interior point at termination; both an order of magnitude under `tol` = 1e-8. Objective moves in the 9th digit. |
| `lp_row_constant` | 5 → 5 | 7.15e-10 → 3.50e-9 | same. |
| `lp_row_constant_expr` | 5 → 5 | 7.15e-10 → 3.50e-9 | same — the `_expr` twin of the row above, which is why the two move together. |

One more line differs in *objective* while its verdict does not: `scaled_feasible_a` hits the 199-iteration cap under `qp_hsde=no` on **both** binaries and reports `MaximumIterationsExceeded` either way (1.14e11 vs 1.25e11, KKT 8.4e45 vs 3.4e30 — both meaningless). See below.

## Findings that did not become code — Q9a

Two things surfaced during the sweep. Neither is caused by this PR — `qp_hsde=no qp_gondzio_corr=0` reproduces `d38167a9` byte-for-byte including both — and one of them dissolved on inspection.

**1. Three fixtures move their objective while still reporting success — not a defect.** `scaled_feasible_a` 236.85 ↔ 315.67, `scaled_feasible_b` 456.33 ↔ 378.03, `feasible_x0_extreme_row` 0 ↔ 0.598. All three pass on the **noise-floor** test, not on a small residual: `final_kkt_error` 246.7 / 934.2 / 7.6e-4 against `final_kkt_error_above_noise` of exactly **0** in every case. `scaled_feasible_a` minimizes `(x₀+3.81)² + (x₁−500000)² + …` and stops at `x₁ ≈ 499987`; at data of magnitude 5e5 a residual of a few hundred *is* the arithmetic noise floor, so the objective is genuinely underdetermined and a wide set of points qualifies. Independent confirmation: two unrelated perturbations (`qp_gondzio_corr=0`, and the τ experiment) move exactly the same three. Written up in `§0h`; no code change needed.

**2. The direct driver diverges on `scaled_feasible_a` — real, pre-existing, and deliberately not fixed here.** Under `qp_hsde=no` it runs to the 199-iteration cap at KKT 8.4e45; at default options HSDE solves the same fixture in 16 iterations. The judgement is explicit rather than an oversight: the solver returns an honest `MaximumIterationsExceeded` rather than claiming success; it is reached only when a caller has explicitly disabled the default driver; and making the infeasible-start method robust across five orders of magnitude of scale — with `qp_equilibrate` already on for that route — is a research task on the driver's start point and regularization, with nothing to do with correctors. **Q9b measured a second instance of the same thing** (see below), so this is now a two-instance finding rather than a one-off. It needs a decision on whether it is worth a phase of its own. Both measurements are recorded in `§0h`/`§0i` so they are not lost.

**Disclosure:** before the instruction to stop filing issues reached me, I had already opened **#689** (finding 2) and **#690** (the HSDE τ study). No others were filed. Both are superseded by `§0h` and by this PR body, which are the durable record; #690's stated blocker is retracted by finding 1 above.

## Tests — Q9a

**`crates/pounce-cli/tests/issue_588_gondzio_correctors.rs`** (3, new) — drives the real CLI, per gh#551.

* `the_direct_driver_runs_the_correctors` — `lp_afiro` at `qp_hsde=no`, correctors off vs on: the trajectories must differ *and* the count must fall.
* `the_hsde_driver_honours_the_same_knob` — the same option at default options must move HSDE (15 → 13), so the knob demonstrably reaches both drivers.
* `the_registered_default_reproduces_an_unset_run` — `qp_gondzio_corr=3` is bit-identical to unset on both routes. This is the safety argument for the trajectory change, demonstrated rather than asserted.

**How I know they bite.** At `d38167a9` the CLI rejects the option outright — `Read Option: "qp_gondzio_corr". It is not a valid option` — and writes no report, so all three panic at `no report for lp_afiro`. That is failure for the stated reason: the option does not exist. `gondzio_correctors.rs` does not compile at the parent, since `QpOptions` has no `gondzio_max_corr` field.

**`crates/pounce-convex/tests/gondzio_correctors.rs`** (5, new) — `0` is a real off switch on both drivers; the default is the historical HSDE setting; correctors never move the optimum across two families × several sizes.

**Deliberately not pinned, and why.** The two aggregate tests in that file assert a **floor**, not a win, and their doc comment says so with the numbers: over the degenerate-corner and spread-LP families the correctors *do* fire (roughly one acceptance per solve, mean gain ~2e-2) and the iteration totals come out **exactly level** — 146 either way, 84 either way, individual instances moving in both directions and cancelling. They pin that the acceptance rule is not passing correctors that hurt. The evidence the scheme *pays* is `afiro` and the generated QP, not these.

---

# Q9b — presolve validity for quadratic rows (§7), and the `presolve_conic` wiring

## The scope boundary, established before anything was built

**Most of §7–§8 depends on work that does not exist, and the note now says so.**

§7 reasons about a native `QcqpProblem` carrying a per-row `Qᵢ` alongside `P, c, A, G`. That is **Q7, which was skipped**. §8's recommendation is explicitly "ship equilibration on by default for **the QCQP driver**" — that is **Q8, blocked on Q7**. Neither is on `main`.

What exists is the conic path, and it changes the problem more than it first appears. `extract_socp_with_map` does **not** keep `Qᵢ` on the side. It writes each quadratic row `½xᵀQᵢx + aᵢᵀx + bᵢ ≤ 0` as a second-order cone block of `r + 2` rows:

```
row 0    :  aᵢ,        h =  1 − bᵢ
row 1    :  aᵢ,        h = −(1 + bᵢ)
row 2..  : −√2·Fᵢ,     h =  0          (FᵢᵀFᵢ = Qᵢ)
```

So **the quadratic structure *is* rows of `G`.** Presolve sees all of it. §7's column-level hazards arise from a `Qᵢ` the presolve cannot see, and here there is no such thing.

### §7's table, row by row, with what was implemented / disabled / blocked

| §7 row | verdict on the conic path | status |
|---|---|---|
| Empty linear row | unchanged | already shipped |
| Empty quadratic row (`Q=0, a=0`) | a row with `Q=0` is not classified quadratic (`is_quadratic` needs a non-empty Hessian map), so it is emitted as an ordinary linear row and the existing empty-row screen handles it. A *fully empty cone block* needs `Q=0` **and** `a=0` at once, which the extractor cannot produce | **unreachable — not implemented, would be dead code** |
| **Parallel/duplicate row removal** | **real, and sharper than §7 says** — the shared `aᵢ` rows are byte-identical, right-hand sides included, not merely parallel | **already guarded; demonstrated and pinned** |
| Activity-redundancy drop, linear row | unchanged | already shipped |
| Activity-redundancy drop, quadratic row | means proving a whole SOC *block* redundant and removing the block: new block-level reduction, a `reduced_cones` case for a vanished non-orthant block, a postsolve entry. **No instance in the corpus, real or generable, exercises it** | **deferred, with the reason** |
| Infeasibility from activity, quadratic row | valid and structurally free, but a false `Infeasible` is the worst answer this solver can give (`presolve`'s own gh #523 note), and nothing available exercises it | **deferred, with the reason** |
| Forcing constraint, linear row | §7's caveat ("no pinned column in any `Qᵢ`/`aᵢ`, postsolve knows only `P, c, A, G`") **does not apply**: here the `Qᵢ` *are* `G` | **no guard needed, none added** |
| **Forcing constraint on a quadratic row** | skipped on `is_soc_row` (`presolve.rs:1351`) | **already disabled** |
| Bound tightening from a linear row | unchanged | already shipped |
| **Bound tightening from a quadratic row** | skipped on `is_soc_row` (`:1686`) — exactly §7's v1 verdict | **already disabled** |
| Fixed-variable elimination | fixing `x_k = v` folds `v·G[i,k]` into `h_i` of every row including cone rows, which is exact. §7's `Q̃`/`ã`/`b̃` update is a `QcqpProblem` concern | **valid here, no `Q` update to write** |
| Empty/free-column elimination | the "one line, must not be forgotten" is already there: the test is `p_col_present == false && a_col_count == 0 && g_col_count == 0`, and the SOC rows *are* `G` rows | **already implied** |
| **Free column-singleton substitution** | a free column singleton is in exactly one equality row and **nowhere else**, `G` included — so never in a cone row. §7's rank-one fill trap is unreachable | **cannot arise** |
| Dominated columns | blocked by `soc_col[col]` (`:1451`), and `soc_col` is used for **nothing else** — so this is §7's guard exactly, and not the over-correction §7 warns about | **already disabled** |
| Doubleton aggregation | `aggregate_once` is reachable only from `presolve`, never from `presolve_conic`, and its doc says why | **already excluded** |

**Nothing in §7 needed a new reduction, and nothing needed disabling that was not already disabled.** That is the finding, and it is a correction to §7 and to Q1's deferral note, both edited in place.

## The `dedup_rows` hazard: real, already guarded, and demonstrated anyway

§7 calls parallel/duplicate row removal "the silent-wrong-answer risk" because `parallel_signature` (`:2146`) builds a row's signature from its *linear* triplets alone. On the conic path this is **worse** than §7 describes rather than better: rows 0 and 1 of a cone block **are** `aᵢ`, so two quadratic rows sharing a linear part produce rows that are byte-identical. No quantization slack is needed to reach the collision.

`presolve_conic` already protects them: it marks every **non-orthant** row `protected` — generalized from `SecondOrder` alone by the H9 fix — and `dedup_rows` excludes protected rows from grouping entirely, "never dropped and never drop others". **So the fix §7 names was already in the tree, arriving for a different reason (exp/power/PSD blocks) than the one §7 gives.**

That is a claim worth demonstrating rather than asserting. `crates/pounce-convex/tests/presolve_conic_quadratic_rows.rs` uses a four-variable QCQP whose two quadratic rows are `2x0² + x0 + x1 ≤ 20` and `2x1² + x0 + x1 ≤ 20` — identical in `a` and in `b`, different only in `Q` — plus three orthant rows that are genuine scalar multiples of one another:

| pass | rows | cone partition | status | objective |
|---|---|---|---|---|
| told nothing about the cones | 9 → **4** | `Nonneg(1), SecondOrder(2), SecondOrder(1)` | `Optimal` | **−17.35083** |
| `presolve_conic` | 9 → **7** | `Nonneg(1), SecondOrder(3), SecondOrder(3)` | `Optimal` | **−10.40312** |

The true optimum is `−(√41 − 1) − 5 = −10.403124…`. The first row is a wrong answer **67% off, reported as `Optimal`**, with nothing anywhere raising an error — and the *only* difference between the two runs is the cone list the pass is handed. **The protection is removed as data, not as a source mutation**, so the test is a permanent statement that it is load-bearing.

All four `a` rows collapse into one parallel group, not two pairs — they all normalize to `(1,1)`, so the merge keeps the single most restrictive and drops three. **Both** blocks are damaged, not one.

Alongside it, `presolve_conic_agrees_with_no_presolve_over_a_random_qcqp_battery` is the differential sweep §7's warning actually calls for: 64 generated QCQP-shaped conic problems, each with a linear part deliberately shared across its quadratic rows plus duplicate and parallel orthant rows, solved bare and through presolve + postsolve, objectives required to agree and every cone block required to keep its dimension. The fixture corpus is what missed gh #683 and gh #685; a synthetic battery is the evidence that generalizes.

## The wiring: shipped as a contract fix, with no speedup claimed

Q1 deferred `presolve_conic` into `run_convex_socp` on two grounds — that it was unmeasurable, and that the safety work belonged to Q9. **The second is stale**: the safety work was already present, as above. **The first is still true**, and is stated rather than worked around.

What the wiring fixes is a contract. `qp_presolve` (and its `presolve` alias) is registered, documented, and defaults on for the convex path — and was **silently ignored on every convex QCQP**, because this driver did not presolve at all. Setting it did nothing in either direction.

The presolve arm mirrors `run_convex_qp` line for line: the `max_iter=0` stop stays above it, the debugger arm still steps the *extracted* problem, and the `Presolve:` lines are held back until the gh #535 fallback check has passed so a declined conic attempt leaves no stdout. `postsolve` restores `z` at the original row indices, so `con_map`'s `z_row0`/`z_row1` and `recover_socp_duals` need no remapping. One line differs deliberately: no `cap-truncated` suffix, because `presolve_conic` is a single pass and the round cap is a fixpoint notion.

### Measured — all of it

| model | source | presolve | iterations | final NLP error |
|---|---|---|---|---|
| `qcqp_ball` | **real**, in-repo | nothing to reduce | 17 → 17 | unchanged |
| `airport` | **real**, in-repo | nothing to reduce | 30 → 30 (declines to NLP either way) | unchanged |
| `qcqp_shared_linear_rows` | **added by this phase** | 9 → 7 rows | 11 → **12** | 9.92e-10 → **8.14e-11** |

`airport` and `qcqp_ball` are pure box-plus-cone models: 42 and 1 cone blocks, a variable box, and **not one orthant row between them**. There is nothing for a row reduction to remove. That is the honest content of "unmeasurable on the real corpus" — the reduction is not declining to fire, **it has nothing to fire on**. Pinned in `a_cone_only_model_is_untouched_by_the_wiring` so a later reader does not have to rediscover it.

Generated instances cannot close the gap either: `scripts/gen-qcqp-nl.py` emits every variable **free**, so no box-driven reduction (activity, bound tightening, forcing, dominated) can fire at all, and its linear rows are random with no duplicates. A generated `qcqp*` shape that routes conic reports no reduction.

**No speedup is claimed anywhere in this half**, and the one iteration count that moves moves *up*.

## Blast radius — Q9b

`scripts/sweep-fixtures.sh` against a binary built from **`db361779`** (this branch's head at Q9a plus the merge of `main`), **58** fixtures — the 57 of Q9a plus the one added here — two legs each, 0 `NO_JSON`, comparing **status, objective AND iteration count**:

| option set | moving lines |
|---|---|
| **default** | 2 |
| **`nlp_scaling_method=none`** | 2 |
| **`mu_strategy=adaptive`** | 2 |
| **`qp_hsde=no`** | **0** |

The same two lines at the first three option sets, and they are the `exact` and `lbfgs` legs of **one fixture**:

```
exact  qcqp_shared_linear_rows  SolveSucceeded  it=11  obj=-10.40312423
exact  qcqp_shared_linear_rows  SolveSucceeded  it=12  obj=-10.40312424
```

That fixture is the one this half adds, and it is the only conic model in the corpus presolve can act on. It moves because the wiring works: two parallel orthant rows go, the reduced problem is a different (smaller) problem, and its trajectory is one step longer and an order of magnitude more accurate at the end — `8.14e-11` against `9.92e-10`, both far inside `tol`. The objective agrees with the analytic optimum to the digit either way. **Nothing else in the corpus moves at any option set**: 55 of the 58 fixtures never reach a convex driver at all, and the two that reach the conic one have no orthant rows.

`qp_hsde=no` moves **nothing at all**, that fixture included, and the reason is worth stating rather than filing as a happy zero: on that option set the *direct* conic driver does not certify this model and gh #535's fallback hands it to the NLP path, **on both binaries** — so the swept line is the NLP path's and presolve is invisible to it. It also confirms the log hold-back works: a declined conic attempt prints no `Presolve:` line.

## Findings that did not become code — Q9b

**The direct conic driver gets worse on the reduced problem, and it is the second instance of Q9a's finding 2.** Under `qp_hsde=no`, `qcqp_shared_linear_rows` stops after **12** iterations at KKT error **4.81e2** with presolve on, where on the unreduced problem it manages **30** at **5.55e-2**. Neither certifies; the NLP fallback owns the verdict either way, so nothing user-visible moves and the sweep line does not budge.

This is **not** the reduction being wrong — HSDE solves the same reduced problem to the analytic optimum, and the differential battery agrees on 64 more instances. It is the same family as Q9a's finding 2 (`scaled_feasible_a` diverging on the direct driver under `qp_hsde=no` while HSDE solves it in 16): the infeasible-start method's start point and regularization are not robust on this class, and a smaller problem is simply a different problem for it. Two independent instances now, on unrelated models. **This is the one thing I could not fit into this PR** — it is a research task on the direct driver's start point, with nothing to do with either half of Q9 — and it needs a decision on whether it is worth a phase of its own. Recorded in `§0i` alongside `§0h`'s so it is not lost.

## Tests — Q9b

**`crates/pounce-convex/tests/presolve_conic_quadratic_rows.rs`** (3, new) — the hazard demonstration, the guard, and the 64-instance differential battery, all described above.

**`crates/pounce-cli/tests/issue_588_conic_presolve.rs`** (3, new) — drives the real CLI per gh#551. `the_conic_driver_honours_the_presolve_switch` asserts the `Presolve:` line reports `9 → 7 rows` and that `qp_presolve=no` suppresses it; `the_shared_linear_part_does_not_lose_a_quadratic_row` pins the objective against the closed-form optimum on both settings; `a_cone_only_model_is_untouched_by_the_wiring` pins that `qcqp_ball` is unchanged.

**How I know they bite, stated precisely rather than blurred.** `the_conic_driver_honours_the_presolve_switch` **fails at the parent commit for the stated reason** — reverting only `main.rs` and rerunning gives `the conic driver ignored qp_presolve — no Presolve: line in: …`, because on the parent no presolve runs on this route whatever the option says.

The other five **pass at the parent, and their doc comments say so.** They are guards on a protection that was already correct, not regressions, and the distinction is not blurred: **the hazard this half is about was latent, not live**, because until this PR no production caller pointed any presolve at a conic problem. What the tests pin is that the protection is load-bearing — remove it, as the first one does with data rather than a source edit, and the solver returns `Optimal` at an objective 67% wrong.

## What Q9b did not do

* **Nothing from §8.** Its recommendation names the QCQP driver (Q8) and its argument is about a native representation POUNCE does not have; on the conic path the obstacle `equilibrate.rs:44-49` documents is the live one — per-row scaling of `G` breaks a second-order cone. §8 is marked blocked in place. No code changed.
* **No Q7, no Q8.** Not started.
* **No new reduction, and none disabled** — the row-by-row table above is the record.
* **`presolve_conic` still runs one pass**, still opts out of doubleton aggregation, still does not iterate to a fixpoint. Correct as-is: the cone partition is recoverable from the kept rows only for a single layer.

---

## Gates

`cargo test --workspace --exclude pounce-hsl` (**257 suites, 3221 tests, 0 failures**), `cargo fmt --all -- --check`, `cargo clippy --workspace --exclude pounce-hsl --all-targets` (exit 0), `tests/conic_hsde_vs_direct.rs` (5/5), plus the five Q9a sweeps against `d38167a9` and the four Q9b sweeps against `db361779`.

- [x] Tests fail on the parent commit for the stated reason — and where they do **not**, the PR body and the tests' own doc comments say which ones and why
- [x] `CHANGELOG.md` `[Unreleased]` entries, in the user's terms — both halves
- [x] Book pages under `docs/src/` updated: `docs/src/lp-qp-routing.md` (the corrector option table and where the scheme applies; a new "Presolve on a convex QCQP" section) and `docs/src/options.md` (`POUNCE_DBG_GONDZIO`, and `qp_presolve`'s reach). Both pages already exist in `SUMMARY.md`

Refs #588


jkitchin:
Reviewed locally, and pushed a fix for the one blocking problem I found (76f54437).

## The claims in the body reproduce

Built release binaries for `c3cceea6` and `main` in separate target dirs and checked the numbers independently.

- **Fixture sweep**: exactly two moving lines, both legs of `qcqp_shared_linear_rows` (`it=11 -> 12`, objective agreeing to 8 digits). 116 legs, 0 `NO_JSON`.
- **`lp_afiro`**: default route 13 iterations with a bit-identical objective on both binaries; `qp_gondzio_corr=0` gives 15, matching the claimed "HSDE 15 -> 13".
- The `hsde.rs` extraction into `correctors.rs` is faithful; tau is applied at `ipm.rs:2571`, acceptance requires *both* step lengths, a failed corrector solve keeps the Mehrotra direction, and the accumulation happens only on accept.
- `cargo fmt --check` and clippy both exit 0. The `qp-convex` CLI suite (138 models) shows 0 status changes and 0 iteration changes.
- Two test failures on my macOS box (`slack_based_scaling_changes_the_solve`, `an_interior_solution_is_unmoved`) are pre-existing — they fail identically on `main`.

## The blocking problem: the safety argument has a hole

The body argues the default route cannot be affected because it runs HSDE. That is true for the `.nl` CLI, but it does not cover the Python surface. `pounce-py`'s `solve_qp` calls `solve_qp_ipm`, which **dispatches internally**: cold solves go to HSDE, but a **warm-started** `solve_qp` goes to the direct driver. I confirmed this empirically with `POUNCE_DBG_GONDZIO=1` rather than by reading the call graph — my first reading of it was wrong in the other direction.

So warm-started Python solves are a default path through the newly-corrected code, and `sweep-fixtures.sh` structurally cannot see that route.

Measured over 230 warm-started convex QP/LPs through the Python host:

| | main | PR head |
|---|---|---|
| total iterations | 967 | **1004** (+3.8%) |
| QP family | 601 | 601 |
| LP family | 366 | **403** (+10.1%) |

All still `Optimal` with agreeing objectives — 37 of the 80 LPs lost exactly one iteration and **none** gained one. This also explains the failing CI check, which the body does not mention: the Python job was 198 passed + 1 failed against 199 on main, on the same machine and command.

### Mechanism

The band is symmetric, `[0.1mu, 10mu]`, so a product the affine step drove *below* `BETA_LO*mu` gets corrected back **up** to the floor. On a warm start that is backwards — products far below mu are the superlinear tail. Traced on one instance, the corrected run stepped mu down by exactly 10x three iterations running where the uncorrected run managed 17x, then 276x, then 8000x. The acceptance rule is step length alone, so it cannot see the bill: the corrector delivers the alpha it advertises and spends the tail to buy it.

This is the gh#544 shape again — the right answer, more slowly, invisible to a suite that asserts status and objective but not iteration count.

## The fix

Gate the loop on the Mehrotra step still being short, `min(alpha_p, alpha_d) < 0.9` — Gondzio's own premise, since the scheme is written for a step *blocked short*. Direct driver only, one call site. HSDE deliberately untouched.

Two things worth recording about how the threshold was chosen, because both cut against the obvious approach:

**`lp_afiro` is not usable for tuning.** Its count is chaotic in the threshold:

| alpha | 2.0 | .99 | .95 | .92 | .90 | .88 | .86 | .85 | .82 | .80 | .70 |
|---|---|---|---|---|---|---|---|---|---|---|---|
| iters | 118 | 118 | 121 | 128 | 122 | 115 | 130 | 130 | 117 | 136 | 136 |

Clean asymptotes with ~15 iterations of noise between them. An early draft of this fix picked 0.88 because it read 115 — better than the ungated PR's own 118 — which was fitting noise.

**Two aggregates were used instead**, and they separate sharply:

| gate | 42 NETLIB LPs (cold) | 230 warm QP/LPs |
|---|---|---|
| no correctors | 1809 | 967 |
| ungated | **1757** (-52) | 1004 (+37) |
| alpha < 0.95 | 1757 (-52) | 1003 (+36) |
| **alpha < 0.9** (shipped) | **1759** (-50) | **967** (+0) |
| alpha < 0.85 | 1781 (-28) | 967 (+0) |

The cold column is the direct driver over every NETLIB LP that converges on it — **only 42 of 365 do**, the rest stall and fall back to the NLP path per gh#133. Worth stating plainly, because it is the first evidence in this phase that the scheme's cold benefit is real and general (27 models improve, 7 worsen) rather than a single-fixture artifact.

0.9 is where the curves cross. 0.95 keeps the cold win and fixes nothing warm; 0.85 pays a quarter of the cold win for nothing further, and below 0.9 the cold column goes chaotic too.

A **one-sided band** (correct only products above `BETA_HI*mu`) was tried first and rejected: it takes `lp_afiro` to 135, exactly the uncorrected baseline. The win and the harm are the same action, so band shape cannot separate them and the step length must.

### After the fix

- Warm battery **967**, per-instance identical to main (0 iteration diffs, 0 objective diffs). Cold battery 1864, unchanged.
- The failing Python CI check passes — as a consequence of fixing the defect, not by touching the test.
- Fixture sweep still moves exactly the same two lines. Blast radius unchanged.
- `lp_afiro` direct is **135 -> 122** (was 118 ungated), KKT still improving 2.10e-7 -> **2.49e-8**. Default route unchanged at 15 -> 13.
- `cargo test --workspace --exclude pounce-hsl`: 3220 passed, 1 failed (the pre-existing macOS one above). fmt and clippy exit 0.

The stale `135 -> 118` / `1.35e-8` figures are corrected in CHANGELOG, dev-notes and both test doc comments.

### One note on the existing tests

`correctors_do_not_cost_iterations_on_the_degenerate_family` and `..._on_the_spread_family` are **inert**: 146 = 146 and 84 = 84 at every threshold including gate-open. They only assert "not worse" and the correctors never usefully fire on them, so they never constrained this in either direction. The new `correctors_do_not_cost_iterations_on_a_warm_start` covers the route they miss (105 -> 117 ungated, 105 -> 106 gated).

## Smaller points, not blocking

1. **No off-switch on the affected surface.** `qp_gondzio_corr` is registered CLI-only (`main.rs:2319-2321`); `grep -i gondzio` over `python/` and `crates/pounce-py/` returns nothing, and `POUNCE_DBG_GONDZIO` is a print, not a switch. A Python user hitting a corrector regression has no bisection handle. Worth exposing `gondzio_max_corr` on the host.
2. `presolve_conic`'s `debug_assert_eq!` compiles out in release, so the invariant it documents is unchecked in shipped builds.
3. #689 and #690 are still open although the body describes them as superseded. (#689 has since been closed by #696.)

jkitchin:
## Correction: my calibration above was measured against a driver that has since changed

#696 merged between when I posted the comment above and now. It fixed #689 by
rescaling the direct driver's cold start to the data and stopping HSDE from
certifying non-optimal iterates. That is the same driver this gate sits in
front of, so it invalidated the numbers I quoted. Concretely, #696 took
direct-driver convergence on the NETLIB LP corpus from 60 of 365 models to 96,
and `lp_afiro` from 135 iterations to 10.

I have merged `origin/main` into this branch, re-measured everything on the
merged tree, and pushed `c82ae569`. **The threshold changed, 0.9 -> 0.85.**

### What the re-measurement says

52 NETLIB LPs converge on this driver under every arm; correctors off is 688
iterations.

| arm | total | vs off | better | worse |
|---|---|---|---|---|
| off | 688 | — | — | — |
| ungated | 649 | −39 | 28 | 3 |
| 0.95 | 647 | −41 | 28 | 2 |
| **0.90** | 644 | −44 | 29 | 1 |
| **0.85** | 644 | −44 | 29 | 1 |
| 0.80 | 651 | −37 | 27 | 1 |

The interesting change is not the threshold, it's that **the gate is no longer
a trade.** In my comment above I justified it as buying warm-start neutrality
at the cost of two cold iterations. Post-#696 it improves the cold aggregate
too (−44 against −39 ungated), and cuts the models correctors make *worse*
from 3 to 1. Correcting an already-long step was never paying for itself on
either surface; warm starts were just where it was loud enough to measure —
which is the surface `pounce-py`'s `solve_qp` actually routes to this driver,
and the one `scripts/sweep-fixtures.sh` structurally cannot see.

0.85 strictly dominates 0.9: same cold aggregate, same 230-instance warm
battery (967, per-instance equal to correctors off), and unlike 0.9 it leaves
no leak on the `boxed_lp` warm family (105 gated / 105 at 0.8 / 106 at 0.9 /
117 ungated). That is why the warm-start regression test now asserts
`on <= off` strictly instead of within a slack.

### `lp_afiro` is no longer evidence for anything

My earlier comment leaned on `135 -> 122` with the KKT error improving
`2.10e-7 -> 2.49e-8`. Post-#689 that fixture solves in 10 iterations off and 9
on, and its KKT error gets **worse** (2.25e-9 -> 6.72e-8 — both converged,
objectives agreeing to the asserted tolerance). The CLI test still asserts the
*direction*, because proving `qp_gondzio_corr` is wired to `run_ipm` is all it
was ever for; its doc comment now says exactly that, warns against re-tuning
the gate against a count that is chaotic in the threshold, and points at
`correctors::ALPHA_MAX` for the 52-LP aggregate that actually carries the
claim. CHANGELOG and `dev-notes/quadratic-structure-exploitation.md` are
updated the same way.

### Verification on the merged tree at 0.85

- **Fixture sweep**, both legs, against a freshly built post-#696 `main` binary
  run over this branch's corpus: 116 legs, 0 `NO_JSON`, **exactly 2 moving
  lines**, both `qcqp_shared_linear_rows` (`it=11 -> 12`, objective agreeing to
  8 digits). Byte-identical to the pre-recalibration diff, so 0.85 moves no
  fixture that 0.9 did not. (Note for anyone re-running this: `main` cannot
  sweep that fixture, because this branch adds it — you have to run main's
  binary against this branch's corpus.)
- **Warm battery**, 230 warm-started convex QP/LPs through the Python host:
  967 iterations, 0 non-optimal, per-instance identical to main.
- **Cold battery**, 180 instances: 1864 iterations, 0 non-optimal.
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets`
  clean.
- `cargo test --workspace --exclude pounce-hsl --no-fail-fast`: 3225 passed,
  1 failed — `slack_based_scaling_changes_the_solve`, which fails identically
  on `main` on this platform (asserts two scalings disagree; both give 103).
- Both #588 corrector test files and both #689 tests pass together.

The temporary `POUNCE_X_ALPHA_MAX` / `POUNCE_X_ONESIDED` hooks I used to sweep
the threshold are not in the pushed tree.

PR #701, opened 2026-08-19
Sum the presolve chain's objective offsets, so the accessor is honest (#697)

### LINKED ISSUE #697: Presolve::obj_offset is 0.0 for any multi-layer presolve chain
Found during the local review of #696 (merged). Not introduced by that PR —
it is pre-existing — but #696 is the first caller to depend on the value, so
it starts mattering now.

## The defect

`Presolve::obj_offset` (`crates/pounce-convex/src/presolve.rs:900-912`) is
`0.0` whenever the presolve fixpoint took **two or more layers**.

The fixpoint wrapper is constructed with `obj_offset: 0.0`. The real
per-layer offsets live in the `chain`, and only `postsolve` walks that chain
(`presolve.rs:2457-2486`). So the aggregate accessor silently reports no
offset for any multi-layer reduction, which this repo's own CHANGELOG
describes as the common case.

## Why it matters now

The CLI reads `ps.obj_offset` at `main.rs:2583` / `:2588` to build the new
`obj_constant`. When presolve took more than one layer the presolve
contribution is dropped and only the `.nl` term survives, so #696's
"plus presolve's own objective offset" holds only for single-layer
reductions.

## Impact — bounded

`obj_constant` is a **convergence-test normalizer only**. It is consumed by
`scale_g` (`hsde.rs:541-543`) and passed through scaling at `ipm.rs:569,
969, 2130`. It touches no residual, no direction, no dual, and not
`QpSolution::obj`. So this cannot produce a wrong answer — it can only make
the relative gap test slightly too loose on models where presolve moved a
large constant into the objective.

Flagging the bounded-impact reasoning explicitly because "it cannot produce a
wrong answer" is the argument that shipped gh#544: the failure mode here is a
too-loose stopping test, i.e. a trajectory/accuracy effect, which is exactly
the class that goes unnoticed.

## Suggested fix

Either sum the chain's offsets when building the fixpoint wrapper, or remove
the aggregate accessor and make callers walk the chain, so the value cannot
be read as complete when it is not.

jkitchin:
Fixes #697.

## Problem

`Presolve::obj_offset` was a literal `0.0` on the fixpoint wrapper. The real
per-layer offsets live in `chain`, and only `postsolve` walked it — so the
public accessor reported "presolve moved no constant into the objective" for
every reduction that took two or more layers, which this repo's own #530 sweep
records as the common case (the layer cap binds on 46% of LP and 25% of QP
models). Nothing distinguished that from a genuine zero.

Reached in practice, not merely reachable — three corpus fixtures presolve into
two layers with a nonzero offset and reported `0.0`:

| fixture | layers | true offset | `obj_constant` it feeds |
|---|---|---|---|
| `dual_order` | 2 | `-59` | 909 |
| `dual_scaled` | 2 | `-5999` | 9.0e6 |
| `tame` | 2 | `1` | — |

## Cause

The wrapper standing for the composition of the layers owns none of their
offsets, and was constructed with the placeholder the field was never updated
from. `postsolve` folds the layers in reverse and each applies its own, so the
*recovery* path was always correct; only the aggregate accessor was not, which
is why nothing caught it until a caller read the value directly.

## Fix

Telescope the chain: layer `Lₖ`'s objective is `Lₖ₊₁`'s plus `Lₖ₊₁.obj_offset`,
so the sum is exactly the constant between the final reduced problem — the one
this wrapper hands the solver — and the user's. Exact, not an approximation.

No double-count in `postsolve`: with a non-empty chain it folds the layers and
never reads this field. Verified by reading every use, and pinned by the test's
identity below, which fails just as loudly on a double-count as it did on the
drop.

The reader that made the stale value matter is #689's `QpOptions::obj_constant`,
which the CLI builds from the `.nl` degree-0 term plus this offset to normalize
the scale-relative stopping test. Where presolve took more than one layer only
the `.nl` term survived, leaving the relative gap test looser than intended.

**Bounded, and the bound is stated rather than leaned on.** `obj_constant` is a
convergence-test normalizer only — no residual, no direction, no dual, not
`QpSolution::obj` — so this could not produce a wrong answer, only a too-loose
stop. That is the trajectory/accuracy class, and "it cannot produce a wrong
answer" is precisely the argument that shipped gh #544, so it is not offered
here as a reason the defect did not matter.

## Blast radius

`scripts/sweep-fixtures.sh`, both legs (`exact` and `lbfgs`), 58 fixtures,
against a binary built from this PR's base — **`3e8383f`, current `main`,
including #694**:

| leg | moving lines |
|---|---|
| default | **0** |
| `qp_hsde=no` | **0** |

Bit-identical. The three fixtures whose `obj_constant` actually changes do not
move because on each the constant is small next to the objective itself
(`dual_scaled` solves at `8994002`), so the normalizer barely shifts and the
relative arm is not what governs their stopping test anyway.

Worth being precise about what that does *not* establish: the corrected value is
right rather than merely different, and no fixture in the corpus demonstrates
the size of error the defect can hide. A model whose presolve moves a constant
*comparable to* its objective would, and there isn't one.

This branch was rebased onto `3e8383f` specifically so the sweep is measured on
the tree the PR merges into — #694 recalibrated the corrector gate on the
post-#689 direct driver after the first measurement was taken, and a sweep
against the older base would not have covered that interaction.

## Tests

`crates/pounce-convex/tests/issue697_multilayer_obj_offset.rs` (2, new).

The multi-layer test asserts the documented contract — reduced objective plus
offset equals the original — against a solve rather than a literal constant.
That choice is deliberate: which of a doubleton's columns the aggregation folds
decides how much of the linear term lands in the constant, and the contract is
invariant to that choice while the number is not. A literal would have pinned an
implementation detail and broken on a benign change.

**How I know it bites:** restoring the `0.0` and rerunning gives
`original objective -20 != reduced 0 + reported offset 0` — the identity failing
by the whole constant, which is the defect itself and not an incidental
assertion.

---

- [ ] Book page under `docs/src/` updated — n/a, no user-facing option, routing
      or output change; `obj_offset` is an internal accessor
- [x] `cargo fmt --all -- --check`, `check-release-consistency.sh`,
      `check-docs-consistency.sh`, `cargo test --workspace --exclude pounce-hsl`
      (260 binaries) all clean
- [x] Every claim here is true of the code as it stands

---
_Generated by [Claude Code](https://claude.ai/code/session_01DFgsipcwFCsuDEzc7iPVyE)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,40 @@ changes.
   solver by the reformulation-cost guards, and the members that do route
   conic carry SOC blocks). `POUNCE_DBG_GONDZIO=1` prints the per-solve
   corrector tally so this is checkable rather than assumed.
+- **`Presolve::obj_offset` now reports the whole reduction chain, not
+  `0.0`** (#697).
+
+  The fixpoint wrapper `presolve` returns holds the composed reduction:
+  its per-layer objective offsets live in `chain`, and only `postsolve`
+  walked that chain. The public aggregate accessor was built with a
+  literal `obj_offset: 0.0`, so any reduction that took **two or more
+  layers** — the common case — reported that presolve had moved no
+  constant into the objective however large a constant it had actually
+  moved. It is now the sum over the chain, which telescopes to the
+  constant between the final `reduced` problem and the user's. No
+  double-count: with a non-empty chain `postsolve` folds the layers and
+  never reads this field.
+
+  The reader that made a stale `0.0` matter is #689's `obj_constant`,
+  which the CLI builds from the `.nl` degree-0 term *plus this offset* to
+  normalize the scale-relative stopping test. Where presolve took more
+  than one layer only the `.nl` term survived. Bounded, but not nothing:
+  `obj_constant` is a convergence-test normalizer only — it enters no
+  residual, no direction, no dual, and not `QpSolution::obj` — so it
+  cannot produce a wrong answer, it can only leave the relative gap test
+  too loose. That is a trajectory/accuracy effect, which is the class
+  gh#544 established goes unnoticed, so it is fixed rather than filed as
+  cosmetic.
+
+  Measured over the CLI fixture corpus, three models reach the convex
+  route with a multi-layer presolve and a nonzero offset that was being
+  dropped: `dual_order` (2 layers, `−59` against an `obj_constant` of
+  `909`), `dual_scaled` (2 layers, `−5999` against `9.0e6`) and `tame`
+  (2 layers, `1`). None of them moves: `scripts/sweep-fixtures.sh` is
+  bit-for-bit identical across all four legs — `exact` and `lbfgs`, at
+  the default route and at `qp_hsde=no` — so this corrects the value
+  every future reader gets without rerouting any solve today.
+
 - **HSDE no longer certifies `Optimal` a few hundred short of the optimum
   when the objective carries a large constant** (#689).
 
--- a/crates/pounce-convex/src/presolve.rs
+++ b/crates/pounce-convex/src/presolve.rs
@@ -499,7 +499,10 @@ pub struct Presolve {
     /// The reduced problem to hand to the solver.
     pub reduced: QpProblem,
     /// Constant added to the objective by variable substitutions; the
-    /// reduced objective plus this equals the original objective.
+    /// reduced objective plus this equals the original objective. For an
+    /// iterated presolve this is the sum over `chain` — the constant between
+    /// the *final* `reduced` problem and the user's, not one layer's share of
+    /// it (gh #697).
     pub obj_offset: f64,
     /// Original problem dimensions.
     orig_n: usize,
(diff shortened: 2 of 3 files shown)


jkitchin:
Closing as a duplicate — #700 fixes the same bug and has been merged.

Two sessions reached gh #697 independently and landed on the same diagnosis:
`presolve_fixpoint` left the offset at `0.0` while each layer in `chain`
carried its own, so the accessor reported "presolve moved no constant" for
every multi-layer reduction. Both PRs were green, and both added a test file
at the same path (`crates/pounce-convex/tests/issue697_multilayer_obj_offset.rs`),
so only one could land.

The two differ in where the sum lives, and that is what decided it:

* **This PR** keeps `pub obj_offset: f64` and fills it eagerly in
  `presolve_fixpoint`. Smaller, no API change, no caller churn.
* **#700** makes the field private as `layer_obj_offset` — layer-local, which
  is what it always actually was — and exposes a computed
  `pub fn obj_offset(&self)` that sums the chain on demand.

#700 was taken because it makes the bug structurally unable to recur. The
defect here was a *stored* value that was never updated to match a derived
one; summing it eagerly fixes today's construction site but leaves the same
shape in place for the next one. Deriving it removes the class. The cost is a
public API break on a published crate, and that turned out to be cheap: the
only consumer in the workspace is `crates/pounce-cli/src/main.rs` (3 call
sites), which #700 updates.

Nothing here was wrong. The analysis in this PR's comment on `postsolve` —
that a non-empty chain folds the layers and never reads the field, so there
is no double-count — is correct and was the thing worth confirming in either
design.

PR #708, opened 2026-08-19
degeneracy="directional": decide through the active-set QP engine, budgeted by degeneracy_iter

devin-griff:
The shipped directional decision enumerates candidate working sets over the weakly active bounds, one released re-factorization per trial. On a dynamic model with 792 weakly active bounds it spends 32 s exhausting its budget on every call before falling back, no budget reaches a meaningful candidate, and a candidate holding a bound whose variable an initial-condition equality already pins is singular, one such trial ending the whole search. Both failures are structural to enumeration.

This PR solves the decision as the QP it is, split across the two crates by what each can reach. pounce-sensitivity does the reduction against the held factor: one released factorization per call (the released Sigma is built once and every solve reuses it, where rebuilding it per call was most of the old per-trial cost), the released direction's violation scan naming the engaged rows, one basis column per engaged row, and a verification sweep that grows the engaged set until no weak row violates. The combinatorics go to pounce-qp: the engaged rows' pin forces form a small bound-constrained QP (one nonnegative variable per engaged row, dense Hessian from the basis columns) whose complementarity conditions are eq. 14's, solved cold by `ParametricActiveSetSolver` with the feral backend and the inertia hint left at Unknown, so working-set updates, anti-cycling, and tie-breaking are the engine's, not hand-rolled. pounce-sensitivity gains workspace dependencies on pounce-qp and pounce-feral. Equality-pinned rows never violate, never enter, and nothing is singular. Engagement uses a sqrt(mu) band relative to the direction's norm: a weak bound's slack and multiplier carry uncertainty equal to their own size, so a movement below that band cannot be resolved against the bound and keeps the released treatment. The gradient is scaled to order one before the engine (its tolerances are absolute, the decision's semantics are relative, exact by homogeneity in g).

The enumeration and the bounded and path directional wrappers are removed rather than left alongside: nothing consumes them after the rewiring, and `parametric_step_directional` is now the QP implementation under the original name.

`degeneracy_iter` (default 16) budgets the decision's back-solves on all three surfaces, replacing the borrowed `max_iter`, which keeps its meaning as the mode's own work. Exhaustion falls back to the one-sided step as before, the warning now naming the true engagement count.

Measured: kink fixtures decide identically to the enumeration (direction differences 0 and 2e-19, fewer solves). The demo notebook's held-breakpoint example is bit-identical, table and record. The 62k-variable column: default budget falls back diagnosed in 5 s against 32 s, and a raised budget produces the exact resolvable decision in 20 s (230 back-solves, 76 of 792 held), which enumeration cannot produce at any budget. The engine's `use_schur_updates` hits MaxIter on dense reduced problems of hundreds of rows and stays off here, noted as engine-side follow-up.


jkitchin:
Reviewed this locally (branch built and tested in a worktree). Verdict up front: **the math is right and I think this is mergeable as-is.** Two things I want your opinion on before it goes in, @devin-ai-integration.

## 1. Nothing in the repo tests `ke >= 2`

`crates/pounce-sensitivity/tests/released_direction.rs:355` asserts `weak.len() == 1`, and that is the only degenerate fixture we have. So every existing test covers `ke = 0` or `ke = 1`. The expansion loop, the dense off-diagonal `S` assembly, and the multi-lambda `held` filter — the parts this PR actually rewrites — are validated only by the manual run on the 792-weak-bound dynamic column model, which CI cannot reproduce.

I built the missing fixtures to close that. Zeroing *both* linear terms in the `released_direction.rs` QP puts the unconstrained minimizer at the origin, so `x1` and `x2` are each on a lower bound with a vanishing multiplier (`ke = 2`); a third variable with `H = I + G*(offdiag)` gives `ke = 3`. Both are hand-computable: on the hold side the dual QP must return `lambda = [A1, B1]` (both strictly positive) and the direction must come back exactly zero.

Ran them against this branch **and** against `main`:

| case | direction | `main` trials | PR trials |
|---|---|---|---|
| `ke = 2`, hold side | `[0, 0]` (matches hand algebra) | 6 | **3** |
| `ke = 3`, hold side | `[0, 0, 0]` | 11 | **4** |
| `ke = 2` / `ke = 3`, release side | free system `H dx = c` | 1 | 1 |

Two conclusions, both of which I think strengthen the case for merging:

- **The rewrite is answer-preserving.** On every case the old enumeration can still solve, this branch agrees with it to the last bit. That evidence did not exist before, and it is the strongest available given the real motivating model is out of CI's reach.
- **This is a complexity fix, not a correctness fix.** The PR's cost is `ke + 1`; enumeration grows combinatorially (6, 11, ...). So the 792-bound model isn't a *slow* case for the shipped code, it's an unreachable one — which is exactly what the CHANGELOG says, now measured rather than asserted.

The `ke = 3` run also shows the expansion loop engaging one row at a time and re-solving (4 trials = 1 initial + 3 expansions), with three simultaneous positive multipliers going through the `held` filter. So those paths are genuinely exercised.

The two test files run in about 10 ms total. **Do you want them folded into this PR, or filed as a follow-up?** My preference is in this PR — they are regression evidence for exactly what it changes.

## 2. `xi` / `xj` are named inversely to the math, in `solver.rs:989-990`

```rust
let xi = cols[engaged[j]].as_ref().expect("column built");   // actually holds X_j
let xj = cols[engaged[i]].as_ref().expect("column built");   // actually holds X_i
let s_ij = 0.5
    * (sign(engaged[i]) * xi[weak[engaged[i]].var_row]        //  a_i^T X_j
        + sign(engaged[j]) * xj[weak[engaged[j]].var_row]);   //  a_j^T X_i
```

The arithmetic is correct. `S_ij = a_i^T X_j`, and extracting `a_i^T` from a vector is picking component `var_row(i)` with its sign, so the two terms really are `a_i^T X_j` and `a_j^T X_i` — the symmetrized off-diagonal, which is what eq. 14 wants.

The issue is purely how it reads. Pattern-matching `s_ij = 0.5 * (a_i . xi + a_j . xj)` parses as `0.5 * (a_i^T X_i + a_j^T X_j)` — the average of two *diagonal* entries, which would be flatly wrong. The names invite a misreading the code does not commit, and verifying the block costs a detour because of it. Suggestion:

```rust
let col_j = cols[engaged[j]].as_ref().expect("column built"); // X_j
let col_i = cols[engaged[i]].as_ref().expect("column built"); // X_i
// S_ij = a_i^T X_j; symmetrize, since S is symmetric in exact arithmetic
let s_ij = 0.5
    * (sign(engaged[i]) * col_j[weak[engaged[i]].var_row]
        + sign(engaged[j]) * col_i[weak[engaged[j]].var_row]);
```

Cosmetic, no behavior change, and the existing "symmetrize" comment stays accurate. `col_i` / `col_j` rather than `x_i` / `x_j` also avoids two near-identical identifiers differing by one character.

## What else I checked

- **`d = d0 + sum lambda_k X_k`** — plus, not minus, is correct against the KKT condition `S lambda + m >= 0`.
- **`HessianInertia::Unknown`** is safe: `S` is a congruence of `K_rel^-1` and therefore PSD, so the declaration is conservative rather than wrong.
- Termination and memory are bounded by the back-solve budget; `engaged` ordering is deterministic.
- `scripts/check-release-consistency.sh` passes on the branch — the two new deps keep the publish topology valid (feral:41 -> qp:47 -> sensitivity:51).
- Not a trajectory change (`parametric_step_*` runs post-convergence), so no fixture sweep is required.
- pyomo fallback routing is right: `fail()` prefixes recoverable errors with `"directional derivative: "`, so a QP `MaxIter` warns and falls back while `BacksolveFailed` propagates.
- The CHANGELOG is unusually honest — it documents the breaking removals rather than burying them, and the docs correctly drop the strict-convexity uniqueness claim the new method no longer supports.

Two smaller notes, neither blocking:

- `g` is scaled to order one but `H` (= `S`) is not, so the engine's absolute tolerances aren't fully normalized. The code acknowledges this ("at fixed H") — real but bounded looseness. Was leaving `H` unscaled deliberate?
- `degeneracy_iter = 16` cannot solve a realistically degenerate problem (the 792-bound model needs ~793). I read that as deliberate fail-fast into the documented fallback, and the CHANGELOG says as much — just confirming that's the intent and not an oversight.

devin-griff:
Both folded in. The branch now has two more commits.

## 1. Fixtures

I recreated them from your description rather than having you push, since the recipe was complete. They are in `crates/pounce-sensitivity/tests/degenerate_expansion.rs`, five tests, tens of milliseconds.

`Ke2Qp` drops both linear terms so the unconstrained minimizer sits at the origin and `x1`, `x2` each land on a lower bound with a vanishing multiplier. `Ke3Qp` adds the third variable with `H = I + G*(offdiag)`. Both sides of the kink are asserted against hand algebra. The release side is the free system, `[(A1 - G B1)/det, (B1 - G A1)/det]` for `ke = 2` and `alpha (c - beta * sum c)` for `ke = 3` with `alpha = 1/(1-G)` and `beta = G/(1+2G)`. The hold side returns a direction of exactly zero with every row through the held filter.

The back-solve counts are asserted exactly rather than bounded, since they are the regression you identified: `assert_eq!(trials, 1)` on each release side and `assert_eq!(trials, 3)` and `assert_eq!(trials, 4)` on the hold sides. A fifth test gives the `ke = 3` hold side a budget of 3 and checks it errors with the recoverable prefix before spending the fourth solve.

One gap against your files. In this fixture family every engaged row enters on the QP's first pass, because with uniform negative coupling, holding the other rows only raises a weak row's movement. So the expansion loop's re-entry, where a row engages only after a first decision returns, is exercised on the column model but not in CI. If your `ke = 3` run showed re-entry (your "4 trials = 1 initial + 3 expansions" reads as one row at a time), your file covers something mine does not and is worth landing on top.

## 2. `xi` / `xj`

Applied as you wrote it, `col_i` / `col_j` with the `S_ij = a_i^T X_j` comment.

## 3. `H` unscaled

Deliberate, but only as the minimal fix for the failure I had measured, which was decisions changing with the perturbation's size. Your question held up when I went back to the engine.

`opt_tol` gates `p_inf` on the stationarity step and the working-set multiplier sign test at `solver.rs:564`, and both live at the scale of the QP's variables, which are the pin forces. Scaling `g` alone puts those forces at the reciprocal of `S`'s scale, and `S` entries are compliances in the model's units, so a model whose compliance runs to `1e9` would carry true pin forces near the tolerance and could be declared stationary at the origin with every row reported released. Real, and not bounded by anything the code controls.

Fixed in the second commit. `S` is scaled by its largest entry, `g` by the direction's scale as before, and the solution maps back by `g_scale / s_scale`. Exact by the same homogeneity argument, since scaling `H` by a positive scalar scales the minimizer by its reciprocal and the constraint `lambda >= 0` is unaffected.

Re-verified on the double column model, `zF1 + 1%`, the same 792 weakly active bounds:

| | before | after |
|---|---|---|
| direction, max abs difference | | `3.5e-16` at direction scale `0.16` |
| held rows | 76 of 792 | 76 of 792, same rows |
| back-solves | 230 | 230 |
| wall | 21.1 s | 21.7 s |

The crate suite and the 56 `pyomo-pounce` sens and degeneracy tests pass on the rebuilt wheel.

I stopped at a scalar rather than equilibrating `S`'s diagonal. Diagonal equilibration would also be exact here, for the same reason the scalar is, so it stays available. The column's `S` mixes trace compositions with holdups over 230 engaged rows and still decided identically to the enumeration on the fixtures it can reach, so internal spread is not currently biting and the scalar closes the failure mode I can actually name.

## 4. `degeneracy_iter = 16`

Intent, not oversight. The default is a statement about what `estimate()` costs by default, and the budget errs before the work rather than after, so the fallback costs one extra check and not sixteen factorizations. A model that is degenerate at this scale opts into the exact decision by raising the number, and the error names both the shortfall and the weak count so the caller knows what to raise it to. Worth saying plainly: 16 is not a guess at a sufficient budget, it is a ceiling on unrequested work.

## 5. A detail on the PSD argument

`Unknown` is safe and your conclusion holds. The route is worth pinning down, since `K_rel^-1` by itself is indefinite as the inverse of a saddle-point matrix, so the PSD property has to come from which block the congruence lands in. The basis columns use unit vectors in the primal rows, which puts `S` at a congruence of the primal block, `Z (Z^T H Z)^-1 Z^T` with `Z` spanning the equalities' null space and `H` the released barrier Hessian. That block is PSD whenever the released reduced Hessian is, which second-order sufficiency at the base point gives. So the declaration is conservative by way of the reduced Hessian.

crates/pounce-sensitivity/src/solver.rs
```diff
@@ -853,10 +832,282 @@ impl Solver {
...
+            // a clean base point takes the plain step and no decision
+            // happens, so the reported decision work is zero
+            let mut d = vec![0.0; dim];
+            if !bs.solve(&rhs_plain, &mut d) {
+                return Err(SolverError::BacksolveFailed);
+            }
+            return Ok((d[..n_x].to_vec(), Vec::new(), 0));
+        }
+
+        let budget = |need: usize, spent: usize| {
```
jkitchin: The budget error is hard-coded as `budget(1, work)` at both call sites (the `d0` guard below and the basis-column guard in the expansion loop), and the message interpolates `nw`, the *weakly active bound* count, rather than the engagement count.

On this PR's own 62k-column example the user reads:

```
directional derivative: 1 more back-solve(s) needed with 16 of 16 spent over 792 weakly active bound(s)
```

so they set `degeneracy_iter=17`, pay another ~5 s, and get the identical message -- 60+ retries to reach the 77 that example actually needs (76 held + 1). `engaged.len()` is in scope at the basis-column guard and is the real retry price.

This contradicts the contract the PR ships: CHANGELOG says "the warning now naming the true engagement count", and `docs/src/sensitivity.md:394` says "a warning naming the engagement count, which is also the retry price".

Separately, the message never names `degeneracy_iter`, so a user who hits it is not told which knob to raise -- and `max_iter` no longer works, which is exactly the knob they would reach for from the previous release.

crates/pounce-sensitivity/src/solver.rs
```diff
@@ -806,38 +784,39 @@ impl Solver {
...
         }
         let rhs_plain = self.parametric_rhs_full(pin_constraint_indices, deltas)?;
         let ctx = self.bound_context()?;
         let state = self.state.borrow();
         let state = state.as_ref().ok_or(SolverError::NotConverged)?;
-        let (d, _, trials) =
-            crate::boundcheck::directional_step(&state.backsolver, &rhs_plain, &weak, max_iter)
-                .map_err(SolverError::SensComputationFailed)?;
+        let released: Vec<usize> = weak.iter().map(|w| w.row).collect();
+        let pinned_rows: Vec<usize> = held_var_rows.iter().map(|&r| r as usize).collect();
```
jkitchin, line 812: `held_var_rows` is converted with `r as usize` and handed straight to `path_direction` as `pinned`, with no range check. `path_direction` (`crates/pounce-sensitivity/src/boundcheck.rs:945`) then does:

```rust
let rhs: Vec<Number> = pinned.iter().map(|&i| d[i]).collect();
```

so an out-of-range or negative row is an index-out-of-bounds panic rather than a returned error. `IndexSchurData::from_parts` only validates lengths and signs, so it does not catch it first.

This is reachable from Python: `crates/pounce-py/src/solver.rs:424` maps `Vec<i64>` to `Index` with no validation, while the sibling argument on the very same call goes through `validate_pins` (`crates/pounce-py/src/solver.rs:863`).

Concrete scenario: at a degenerate base point, `solver.parametric_step_bounded_decided(pins, deltas, [-1], 16)` -- or a stale row index carried over from a re-solved model -- panics out of the extension instead of raising `ValueError`. Since `Index` is `i32`, a negative value becomes a huge `usize` here.

`parametric_step_path_decided` happens to be safe only because it funnels the rows through a `HashSet::contains`, so this is worth fixing on the `bounded` side specifically (or validating in `validate_pins` style at the pyo3 boundary for both).

crates/pounce-py/src/solver.rs
```diff
@@ -455,38 +458,38 @@ impl PySolver {
...
-    /// `(dx, segments, trials)`. Rows the accepted working set leaves
-    /// are forced into the walk's base-activity table, so the record
-    /// carries each departure at the fraction where its multiplier
-    /// reaches zero, essentially zero at an exact kink.
+    /// The eq. 14 directional derivative, decided by the pounce-qp
+    /// active-set engine on the reduced weak-row problem. `max_iter`
+    /// is the total
+    /// back-solve budget: the all-released solve plus one basis column
+    /// per engaged weak row count against it, and a budget the weak
+    /// set cannot fit fails before any factorization happens. Returns
```
jkitchin: "a budget the weak set cannot fit fails before any factorization happens" is not true for any `degeneracy_iter >= 1`. The `work + 1 > max_iter` guard ahead of the `d0` solve passes, `solve_released_prebuilt` runs and pays the released factorization, and the shortfall is only detected later at the per-basis-column guard.

The PR's own measurement confirms it: the "default budget falls back diagnosed in 5 s" *is* that factorization. And the new test `a_tight_budget_fails_before_the_work` spends 3 of its 3 back-solves before erroring, which is the opposite of what its name asserts.

Same claim appears in `crates/pounce-sensitivity/src/solver.rs:862` ("errs before the work happens") and `docs/src/sensitivity.md:392` ("fails before any of that work happens"). Readers will size `degeneracy_iter` on the assumption that an over-budget call is free, which is precisely the case where it is not.

crates/pounce-sensitivity/src/solver.rs
```diff
@@ -853,10 +832,282 @@ impl Solver {
...
+                .max(1e-300)
+        };
+
+        let tol0 = band * scale_of(&d0);
+        let mut engaged: Vec<usize> = (0..nw).filter(|&k| movement(k, &d0) < -tol0).collect();
+        if engaged.is_empty() {
+            return Ok((d0[..n_x].to_vec(), Vec::new(), work));
+        }
+
+        let mut cols: Vec<Option<Vec<Number>>> = vec![None; nw];
```
jkitchin: `cols` retains one full `dim`-length vector per engaged row for the life of the call, but `S` only ever reads the `nw` entries `col[weak[j].var_row]`, and the final `d` update only needs the columns whose `lambda` is nonzero (the held set).

At this PR's raised-budget 62k configuration (230 back-solves, dim ~62k) that is ~114 MB held simultaneously, and it grows linearly with `degeneracy_iter`. A user raising the budget toward the 792 weak rows to "get the exact decision" -- which the PR presents as the supported way to get it -- would need ~390 MB of transient columns.

Storing the reduced `nw`-entry projection for `S`, plus only the held columns for the `d` update, would bound this by the weak-set size instead of by `dim * degeneracy_iter`.

jkitchin (commented):
Review of the enumeration-to-QP rewrite. I checked the dual derivation (`d = d0 + sum lambda_k X_k`, `S = A^T K^-1 A`, `m = A^T d0`; the KKT of `min 1/2 lambda^T S lambda + m^T lambda, lambda >= 0` is eq. 14's complementarity and `(S lambda + m)_k` is row k's movement under `d`) and it is right, including the sign convention, the symmetrized `S_ij`, the diagonal, and the `g_scale / s_scale` unscaling by homogeneity. Also verified: upper-triangle triplets are fine (`pounce-qp/src/kkt.rs` normalizes either triangle and `h_times_x` is triangle-agnostic); `use_schur_updates` really does default to `false`; no dependency cycle from the new `pounce-qp` / `pounce-feral` deps and `scripts/publish-crates.sh` already orders them ahead of `pounce-sensitivity`; no stale references to the removed `directional_step` / `*_directional` entries / `next_combination`; `weakly_active_bounds()` emits at most one row per `var_row` so no duplicate basis column can make `S` singular; the expansion loop terminates; and `solve_released_prebuilt` still zeroes released rows' rhs, so the weak-row zeroing is correctly not repeated in the new decision path.

Four findings inline, none of them blocking the math. Plain comment review, no approval or change request.

Nothing had to be re-anchored or folded into this summary -- all four findings sit on added lines inside diff hunks.

devin-griff:
All four fixed in one commit, `c6178a65`. Three were defects in things the PR itself claims, and the fourth is the memory. Verified on the column at `zF1 + 1%`: the direction is unchanged to 4.2e-13 at a scale of 0.158, the same 76 of 792 rows held, and 230 back-solves become 232. Full crate suite and the pyomo side green.

One thing that changes your measured table. Recovering the direction from a combined right-hand side costs one back-solve per expansion round, so the hold side is now `ke + 2` rather than `ke + 1`. Your comparison against enumeration becomes 4 against 6 at `ke = 2` and 5 against 11 at `ke = 3`. The fixtures assert the new counts and the CHANGELOG records them. Still `O(ke)` against combinatorial, and I would rather pay one solve than hold the columns.

On your expansion-loop gap from the earlier round, this makes it worth landing your `ke = 3` file if it exercises re-entry: mine still engages every row on the first pass, and the combined solve now sits inside that loop, so a second pass is a code path CI does not reach.

crates/pounce-sensitivity/src/solver.rs
devin-griff: Fixed, and you are right that it contradicted the shipped text. The message now reports the engaged count against the weak count, names `degeneracy_iter`, and gives the number to raise it to, with the caveat that a later pass can engage more rows so the figure is a floor. On the `ke = 3` fixture at a budget of three it reads:

`directional derivative: 3 of 3 back-solve(s) spent, and 3 of 3 weakly active bound(s) are engaged so far. Raise degeneracy_iter to at least 5; the engaged set can still grow, so that is a floor.`

The guard before the all-released solve cannot know the engagement, so it is now its own message saying it fires only at a budget of zero. A new fixture asserts both name the knob and are non-empty, which caught the first version of this producing a message with the source indentation baked into it.

crates/pounce-sensitivity/src/solver.rs
devin-griff: Fixed at the pyo3 boundary for both decided entries, with a `validate_var_rows` alongside `validate_pins`, checking against the primal block length from `block_dims()`. Confirmed on the column: `held_var_rows=[-1]` and `[1000000000]` now raise `ValueError: held_var_rows[..] = -1 out of range [0, n_x=62167)` where before they reached the raw index. I validated both entries rather than only `bounded`, since `path_decided` is safe by accident of funnelling through a `HashSet` and that is not a property worth relying on.

crates/pounce-py/src/solver.rs
devin-griff: You are right and the test name was the worst of it. Corrected in both docstrings and in `docs/src/sensitivity.md`. Only a budget of zero fails before any work, because which rows engage is not known until the all-released solve has run, so a short budget still pays that factorization before the shortfall is reported. The test is now `a_tight_budget_reports_what_to_raise_it_to`, which is what it actually checks, and its comment states the five back-solves the `ke = 3` hold side needs and why three cannot finish.

crates/pounce-sensitivity/src/solver.rs
devin-griff: Fixed, and this was raised during design and lost. Each column is now projected onto the one entry per weak row that `S` reads, and the full-length vector dropped. What the columns were otherwise held for, `d0 + Σ λ_k X_k`, is recovered in a single solve on `Σ λ_k a_k`, which is the same vector since each `X_k` is `K_rel⁻¹ a_k`. So the storage is bounded by the weak-set size rather than by `dim` times the budget, roughly 1.5 MB against the 114 MB you measured, and it no longer grows with `degeneracy_iter`.

The cost is one back-solve per expansion round, which is the `ke + 1` to `ke + 2` change noted at top level. On the column that is 230 to 232 for an identical answer.

jkitchin:
Reviewed `bd0e93f` locally: built, full `pounce-sensitivity` suite green (16 targets), `cargo fmt --check` clean, `cargo clippy -p pounce-sensitivity --all-targets -- -D clippy::correctness -D clippy::suspicious` clean, all 11 GitHub checks green.

I picked up where the last round left off rather than re-deriving what @jkitchin already verified. Both of you flagged the same gap twice — **the expansion loop's re-entry is not reached by any test in CI** — and neither round closed it. So I built a fixture that does reach it, ran it, and found one defect in the process.

## The expansion loop's second pass works — verified

A fixture that engages a row only on the *second* pass needs `(Mc)_1, (Mc)_2 > 0` (rows 1–2 engage at `d0`), `(Mc)_3 ≤ 0` (row 3 does not), and `M_31, M_32` negative enough that holding rows 1–2 drives row 3 into violation. This one does it:

```text
min ½ xᵀHx − p·cᵀx   s.t. x4 = p,  0 ≤ x1,x2,x3 ≤ 10

H = [[ 1.0, -1.2, -0.5],      c = [1.6, 1.4, 0.8]
     [-1.2,  5.0,  3.5],
     [-0.5,  3.5,  3.0]]      eig(H) = [0.186, 0.943, 7.87]
```

At `p = 0` the unconstrained minimizer is the origin, so all three variables sit on a lower bound with a vanishing multiplier — three weakly active bounds, same recipe as `Ke3Qp`, but with the coupling signs chosen so the first decision does *not* engage everything.

Measured on this branch:

| perturbation | `d[..4]` | `held` | back-solves |
|---|---|---|---|
| `δp = −1` | `[2.7e-15, 2.2e-15, −1.8e-15, −1.0]` | `[0, 1, 2]` | **6** |
| `δp = +1` | `[2.719101, 0.932584, −1.3e-15, 1.0]` | `[2]` | **3** |

Both match hand algebra:

- `δp = −1` is the re-entry case. Pass 1 engages `{0,1}` (`d0 = [−4.0797, −3.2536, +2.8493]`), returns `λ = [26/15, 7/15]` and `d = [0, 0, −4/15]`; row 2 now violates, so the set grows and pass 2 returns `λ = c = [1.6, 1.4, 0.8]`, all strictly positive, direction exactly zero. The back-solve count is the proof it re-entered: `6 = 1 + 2 cols + 1 combined` (pass 1) `+ 1 col + 1 combined` (pass 2). A single-pass decision could not exceed `ke + 2 = 5`.
- `δp = +1` is the **mixed** case, also untested until now — one row held while two release. With `x3` pinned the reduced system `[[1,−1.2],[−1.2,5]] d = [1.6,1.4]` gives `d = [9.68/3.56, 3.32/3.56] = [2.7191011…, 0.9325843…]`, matching to 1e-15.

So the two code paths nobody could exercise — expansion re-entry, and a partial hold with `0 < |held| < ke` — are both correct. Worth landing the fixture; happy to hand you the file.

## One defect: the budget message can tell a caller to raise `degeneracy_iter` to a number they are already at

`solver.rs:927` computes the floor as `engaged_now + 2`. That is the cost of a **one-pass** decision. Each expansion round adds a combined solve (which this PR introduced in `c6178a65`), so on a multi-pass decision the formula under-counts the true need, and once the engaged set has stopped growing it stops moving at all. Sweeping the budget on the fixture above:

```text
budget 1: ERR … 1 of 1 spent, 2 of 3 engaged. Raise degeneracy_iter to at least 4
budget 2: ERR … 2 of 2 spent, 2 of 3 engaged. Raise degeneracy_iter to at least 4
budget 3: ERR … 3 of 3 spent, 2 of 3 engaged. Raise degeneracy_iter to at least 4
budget 4: ERR … 4 of 4 spent, 3 of 3 engaged. Raise degeneracy_iter to at least 5
budget 5: ERR … 5 of 5 spent, 3 of 3 engaged. Raise degeneracy_iter to at least 5   ← already 5
budget 6: OK  held=[0,1,2] work=6
```

At budget 5 the message reads "5 of 5 back-solve(s) spent … raise `degeneracy_iter` to at least 5". A caller following it mechanically never terminates, and a human reads it as self-contradictory. It is exactly the retry-cost problem the surrounding comment says the engaged count exists to avoid, just relocated: the message stops being a strictly increasing sequence.

"That is a floor" covers *under*-estimating; it does not cover a floor at or below what was already spent. The fix is one line — the floor can never be less than one more than what is already gone:

```rust
let need = (engaged_now + 2).max(spent + 1);
```

Applied on this branch it makes the sequence strictly monotone, 4 → 5 → 6 → success, so every retry buys progress:

```text
budget 4: ERR … Raise degeneracy_iter to at least 5
budget 5: ERR … Raise degeneracy_iter to at least 6
budget 6: OK
```

Full crate suite still green with it; no existing assertion moves, since `degenerate_expansion.rs` checks the message names the knob rather than the number.

## Smaller notes, none blocking

- **`active_set_changes()` gained `degeneracy_iter` but its docstring does not mention it** (`sens.py:1683`). `estimate()` and `estimate_report()` both document theirs; this one only documents `degeneracy`, so the new knob is invisible to `help()` on one of the three surfaces the CHANGELOG names.
- **`estimate_report(max_iter=…)` is now accepted and completely ignored.** Documented as such, and signature parity is a reasonable call, but it silently no-ops what used to be the directional budget: `estimate_report(m, p, max_iter=0)` used to force the fallback and now does nothing. Worth a `DeprecationWarning` if anyone is likely to be passing it, or at least a CHANGELOG line, since the CHANGELOG currently describes `max_iter` as keeping "its meaning as the mode's own work" — true for `estimate()`, but `estimate_report()` has no mode work to budget.
- **CHANGELOG has no blank line between this entry and the `nlp_scaling_method` entry below it** (`CHANGELOG.md:71`); every other pair in the file is separated.
- `docs/src/sensitivity.md:398` has a long unwrapped line where `max_iter keeps its meaning…` was appended to the paragraph.

## What else I confirmed

- The restricted-master scheme is sound, not just plausible: solving the dual QP over a subset with `λ = 0` forced elsewhere, then admitting every omitted row whose gradient is negative, reproduces the full KKT system of the `nw`-row problem at termination. `engaged` grows monotonically, so it terminates in at most `nw` rounds, and since `S ⪰ 0` the restricted optimum is global.
- `weakly_active_bounds()` emits **one entry per `var_row`** (`solver.rs:1202` loops over variables and picks a single side), so `S` cannot pick up the duplicate-column singularity a doubly-weak variable would cause. The PSD argument @devin-griff gave holds without that caveat.
- The `band` reads `curr_mu` off the same `IpoptData` handle the activity classifier uses (`activity.rs:334`), so engagement and classification cannot disagree about which `mu` frame they are in — including under a declared/crossover frame.
- Removals are complete: no reference to `directional_step`, `next_combination`, `parametric_step_bounded_directional` or `parametric_step_path_directional` survives anywhere in the tree, notebooks included.
- Publish topology still valid with the two new deps — `publish-crates.sh` has feral at 41, qp at 47, sensitivity at 51.
- The `g`/`S` joint scaling maps back exactly: `λ = λ̃ · (g_scale / s_scale)` is the right inverse for scaling `H` by `1/s` and `g` by `1/g`.
- Not a trajectory change — `parametric_step_*` runs post-convergence — so no fixture sweep needed.

**Verdict: approve with the one-line `need` fix.** The rewrite is correct, including the two paths that had never been executed under test.

---

jkitchin:
@devin-griff I wrote the fix and the regression fixture, but **I could not push to `devin-griff/pounce`** — this session's git proxy is scoped to `jkitchin/pounce` and will not issue a credential for the fork, so the work is on a branch in this repo instead:

```
claude/pr-708-review-dcgjnk    (= bd0e93f + one commit)
```

One commit, `0f61872`, cherry-pickable straight onto `sens-degeneracy-qp`:

```bash
git fetch https://github.com/jkitchin/pounce.git claude/pr-708-review-dcgjnk
git cherry-pick 0f61872
```

## What it changes

**`solver.rs:936` — one line.** `(engaged_now + 2).max(spent + 1)`, with a comment on why the engaged count alone is not enough. On the fixture below the retry sequence a caller walks was `4, 5, 5, 5, …` and is now `4, 5, 6, done`.

**`crates/pounce-sensitivity/tests/degenerate_reentry.rs` — new, four tests, ~10 ms.** This is the file you both asked for two rounds running. It reaches the expansion loop's second pass, which `degenerate_expansion.rs` cannot: with uniform negative coupling, holding a weak row only raises the other rows' movement, so every row there engages on the first pass. Here the signs are chosen so one row does not.

```text
H = [[ 1.0, -1.2, -0.5],      c = [1.6, 1.4, 0.8]
     [-1.2,  5.0,  3.5],
     [-0.5,  3.5,  3.0]]      eig(H) = [0.186, 0.943, 7.87]
```

- `a_row_that_engages_only_after_the_others_are_held_still_enters` — `dp = -1`. Pass 1 engages `{0,1}`, holds them with `λ = [26/15, 7/15]`, and moves row 3 to `-4/15`; row 3 enters and pass 2 returns `λ = c`, direction exactly zero. `assert_eq!(work, 6)` is the proof it re-entered — one pass over a final `ke = 3` costs `ke + 2 = 5`, and no arrangement of a single pass reaches 6.
- `the_other_side_holds_one_row_and_releases_two` — `dp = +1`, the partial-hold case, also previously unreached. Checked against the reduced system `H[1..2,1..2] dx = c[1..2]`, i.e. `[9.68/3.56, 3.32/3.56]`, to 1e-15. `work == 3`.
- `every_budget_retry_the_message_asks_for_buys_progress` — walks the retry loop the way a caller would, parsing the number out of each message and calling again, asserting each answer is strictly greater than the budget it was given. It asserts the sequence is exactly `[4, 5, 6]`. **This is the one that fails without the one-line fix** (it trips the `need > budget` assertion at budget 5); I confirmed that by reverting the line and re-running. The other three pass either way — they are re-entry coverage, independent of the message.
- `all_three_bounds_are_weakly_active` — the base-point precondition, matching `degenerate_expansion.rs`'s pattern.

Note this shifts your `ke + 2` framing slightly: it is the cost of a decision that finishes on **one** pass, and each expansion round adds a combined solve on top. The fixture's doc comment says so, and the CHANGELOG's `d0 + Σ λ_k X_k` paragraph already describes the per-round solve correctly, so I left it alone.

Verified on `bd0e93f + 0f61872`: `cargo fmt --all --check` clean, `cargo clippy -p pounce-sensitivity --all-targets -- -D clippy::correctness -D clippy::suspicious` clean, full `pounce-sensitivity` suite green (17 targets). I did not touch the Python side, the CHANGELOG, or `docs/src/sensitivity.md` — the docs' "that is a floor" wording stays accurate under the fix.

The four smaller notes from my review are untouched and still yours to take or leave: `active_set_changes()`'s docstring not mentioning `degeneracy_iter`, `estimate_report(max_iter=…)` silently no-oping what used to force the fallback, the missing CHANGELOG blank line at `CHANGELOG.md:71`, and the long line at `docs/src/sensitivity.md:398`.

---

jkitchin:
@devin-griff **Superseding the cherry-pick instruction in my last comment — please don't pick `0f61872`.**

@jkitchin has opted to land it as a follow-up PR after this one merges, so nothing here needs to change and this PR does not need another round on my account. Picking it now would just mean resolving it against itself later.

`claude/pr-708-review-dcgjnk` stays as it is (`bd0e93f` + the one commit). Once this merges, that branch diffs against `main` as exactly the fix and its fixture, and I will open it as its own PR then.

---

Issue #702, opened 2026-08-19
qcqp1500-1c: Q4's quadratic evaluator costs 28 iterations (103 → 131) on the one family member it does not speed up


jkitchin:
`qcqp1500-1c` takes 131 iterations instead of 103 since Q4's quadratic
evaluator, and is the only member of the family that Q4 makes slower.

## What moves

On `main` (3e8383f8), Mittelmann `qcqp1500-1c.nl`, default options:

| | iterations | wall | final KKT | objective |
|---|---:|---:|---|---|
| `main`, default | **131** | 212.0 s | 2.99e-09 | 3882979.491 |
| `main`, `POUNCE_DBG_NO_QUAD=1` | **103** | 162.8 s | 7.31e-09 | 3882979.491 |
| `fd738d46` (Q3, pre-Q4) | **103** | 163.1 s | 7.31e-09 | 3882979.491 |

The answer is unchanged to ten digits and the *default* path's KKT error is
actually the better of the two. What moves is the trajectory: 28 extra
iterations, and 49 s of extra wall time.

## Attribution

`git bisect` over the first-parent chain blamed the PR #603 merge
(`d38167a9`); a second bisect inside the branch narrowed it to

    947b3aa3  Q4: evaluate recognized quadratics from their constant matrices (#588)

`fd738d46` (Q3) is 103, `947b3aa3` (Q4) is 131. `POUNCE_DBG_NO_QUAD=1` on
today's `main` reproduces the pre-Q4 run digit for digit, so the Q4 evaluator
is the whole of it and nothing after Q4 contributes.

There is no bug in the evaluator here. Q4 computes the same quantities by a
different summation, the rounding differs at the ULP level, and on this one
instance that kicks the barrier onto a longer central path.

## Why this is worth an issue and not a revert

Q4 is a large win on every other member of the family. Same binary, same day,
`POUNCE_DBG_NO_QUAD=1` as the control:

| instance | default (Q4) | tape (`NO_QUAD`) | Q4 effect |
|---|---:|---:|---|
| `qcqp500-3c` | 330 it / 13.7 s | 332 it / 110.6 s | 8.1x faster |
| `qcqp750-2c` | 63 it / 8.5 s | 67 it / 45.7 s | 5.4x faster |
| `qcqp500-3nc` | 121 it / 4.6 s | 122 it / 25.4 s | 5.5x faster |
[...]

PR #711, opened 2026-08-19
Evaluate factored quadratics from their own structure (#673)

### LINKED ISSUE #673: Constant-structure evaluation skips factored quadratics, because expanding them cancels

jkitchin:
Fixes #673.

> **Two rounds since this was opened, both recorded at the bottom.** #702 landed on `main` and forced a full re-measurement; then review found a merge-blocking defect in the #685 gate, fixed in 10fb7c8. Every number below is against `ac18ba6` and true of the current head.

## Problem

Q4 of #588 stopped building an AD tape for a degree-≤2 objective or row and evaluates it from constant structure instead. It is gated on `is_expanded_quadratic`, and for good reason: reading `(x − 500000)²` back as `x² − 10⁶x + 2.5·10¹¹` cancels five digits (2.4e-5 relative), and ungated it took `airport.nl` at `mu_strategy=adaptive tol=1e-12` from stopping at a tiny step in 16 iterations to grinding out the 300-iteration cap — #544's failure mode.

The gate is right. What was wrong is what it was read to mean. It is a verdict on a **representation**, and it shipped being read as a verdict on a **body**, so every model written as a sum of squared residuals kept its tape and gained nothing from the phase:

| | before | after |
|---|---|---|
| `airport.nl` rows on the fast path | 1 / 42 | **42 / 42** |
| `airport.nl` objective | taped | **quadratic** |
| `airport.nl` tape summands | 1804 | **0** |
| `feasible_x0_extreme_row.nl` forms | 0 | **1** |
| fixtures whose `[quad stats]` line moves | — | **24** |

The design note's §3.5 claim that constant-structure evaluation "pays on all 358 structured problems" was weakened to "the ones whose quadratic parts arrive expanded" when Q4 shipped. This restores it for the sums of squares.

## Cause

`½xᵀHx + aᵀx + c` was the only representation on offer. Reaching it from a factored body is an algebraic expansion about the origin, and expanding is not accuracy-neutral: the tape squares a small residual where the expansion adds and cancels two large numbers. So the gate had to refuse, and the refusal swept up every body whose *shape* was factored, regardless of whether constant-structure evaluation suited it.

## Fix

Add the missing representation. `nl_quadratic::recognize_factored_quadratic` reads a body as `Σ wₖ(bₖᵀx + dₖ)² + aᵀx + c` — the writer's own grouping, kept — and `QuadraticStructure::push_factored_form` stores and evaluates it in that shape.

This is the issue's option (2). Options (1) and (3) were rejected for the reasons the issue gives: storing the vertex `x* = −H⁻¹a` needs a factorization at load and a fallback for singular `H`, and routing per-form on a measurement at `x₀` is point-dependent while the solver moves. (2) is the only one that is **exact** rather than better-conditioned — and that exactness is measured end to end rather than argued: with the outer sum left naive, the build evaluates `airport.nl`'s 42 rows bit-identically to the 42 tapes it replaces.

The **Hessian is not factored and does not need to be.** `Σ 2wₖbₖbₖᵀ` is constant, assembled once from exactly the products the tape accumulates on every call, and `accumulate_hessian`, `add_hessian_vector` and `lower_triangle` are shared with the expanded arm unchanged. The expanded arm is still tried first — both can answer for the same body (a bare `x²` is a monomial and a square) and a merged-row matvec beats one squaring per term — so the `qcqp*` family routes exactly as it did.

**The #702 merge decision.** Both arms' outer sums are compensated. `Σ wₖlₖ²` is the same outer accumulator wearing a different shape, and a least-squares row is #702's case in its purest form: hundreds of same-signed terms and a running total that outgrows each one. Leaving it naive would make the compensation a property of which read-out a body happened to take. The inner accumulator on that arm is the per-residual dot product, `O(nnz)`, and stays naive for #702's stated reason.

Two things found while implementing, both of which had a wrong first version:

- **`w·(l·l)`, not `(w·l)·l`.** The tape multiplies the writer's constant into the square. The first version associated the other way — a one-ulp difference in general, and the entire answer on a body whose terms then cancel.
- **The gradient support must come from the `bₖ`, not the assembled Hessian.** `(x₀+x₁)² − (x₀−x₁)²` has no Hessian diagonal at all while the gradient depends on both variables; reading the pattern off `H` hands `eval_jac_g` a column list short of a column the row actually depends on.

**Four refusals**, deliberate, each keeping the body's tape:

- **A cross monomial `c·xᵢxⱼ` beside a square.** It would need an expanded *and* a factored quadratic part stored side by side. A diagonal `c·xᵢ²` is stored as `c·(xᵢ)²` and costs nothing; anything else refuses. The reach added is a strict superset of Q4's, never a swap.
- **A `Cse` on the sum spine.** `is_expanded_quadratic` may *skip* a re-referenced body because it answers yes/no; this walk accumulates terms, so skipping would drop them and visiting every reference is `Θ(2^depth)` — the shape that measured 0.00 s → 172 s in Q4.
- **A body refused for *loss* rather than shape** (#685) — the flat-sum-of-monomials case, caught by an explicit `is_expanded_quadratic` test in `admitted_factored_form`.
- **A factored body that loses a Hessian entry while `Σ 2wₖbₖbₖᵀ` is assembled** — added in 10fb7c8 after review; see below.

## The #685 gate, and the defect review found in it

`admitted_factored_form`'s gate is a test on the **shape** the coefficients came from, so it catches a lossy flat sum of monomials. It has nothing to say about three honest squares whose Hessian fold cancels — and `Σ 2wₖbₖbₖᵀ` was an unguarded naive `+=` with no `lost_terms` analogue. Review built the witness:

```
(2²⁷x₀)² + (x₀ + x₁)² − (x₀·2²⁷)²
```

`quad_terms_dropped()` calls it lossy and `admitted_quad_form` correctly refuses it — and it was admitted through the factored door. Measured at `x = (3, 1)` before the fix:

| | fast path | tape |
|---|---|---|
| `eval_g[0]` | **16.0** | 0.0 |
| `nnz_h` pattern | **`[(1,0), (1,1)]`** | `[(0,0), (1,0), (1,1)]` |

Two routes over the same bytes disagreeing about a constraint body against its own `≤ 5` bound. The fast path is the *more accurate* side here — 16 is the true value — and it is still the defect, for the reason #685's own test file gives: **the tape is the reference because it is what the row means to this solver, not because it is exact.**

**Fixed at the accumulation, not in the recognizer**, so the check cannot drift out of step with the arithmetic it gates. `factored_hessian` tracks exactness the way `Quad2::merge` does, and `push_factored_form` returns `Option`, refusing before it touches any state so a rejected body leaves nothing half-built; `push_body_form` falls through to the tape like every other refusal.

The inexactness has to be **carried per entry**. In the witness the fold that rounds (`2·2⁵⁴ + 2`) and the fold that drops are different terms, and the dropping add is *exact* — a per-fold check does not fire. The first version of the check made that mistake and a test caught it.

#687's rule is unchanged and pinned in both directions: `(x₀+x₁)² − (x₀−x₁)²` cancels both diagonal entries exactly and still takes the fast path.

The mechanism needs ~2⁵³ of dynamic range across the squared terms. On the diagonal that takes a negative weight; **off** the diagonal it does not — `(10⁹x₀ + 10⁹x₁)² + (x₀ + x₁)² + (10⁹x₀ − 10⁹x₁)²` loses `H(0,1)` with every weight positive, and `10⁹` is ordinary in a badly scaled model. Both are tested.

The same check settles a suspected item: a weight past `8.99e307` whose `2·w` overflows, and a coefficient product that underflows out of two nonzero factors (`10⁻²⁰⁰ · 10⁻²⁰⁰`), are both refused. Each is pinned by a test verified to fail when its half of the guard is disabled.

`add_is_exact` / `mul_is_exact` / `is_live` moved from `pounce-nl`'s recognizer to `pounce-common`, because `pounce-nlp` cannot name `pounce-nl` and two copies of the #683/#685/#687 argument free to drift was the worse outcome.

## Blast radius

**`scripts/sweep-fixtures.sh`, both legs, against `ac18ba6`: 116 fixture-legs, one line differs.**

```
- lbfgs  airport   SolveSucceeded  it=59  obj=47952.70165
+ lbfgs  airport   SolveSucceeded  it=57  obj=47952.70165
```

Which of the two changes moved it is answerable rather than arguable. Building with the factored arm's outer sum left **naive** reproduces the baseline exactly:

| `airport.nl`, `hessian_approximation=limited-memory` | iterations | objective |
|---|---|---|
| `ac18ba6` (all 41 rows on the tape) | 59 | 47952.70164698971 |
| branch, factored arm **uncompensated** | 59 | 47952.70164698971 |
| branch, as merged (compensated) | **57** | 47952.701646989735 |

So the factored read-out on its own is **bit-identical to the tape on this model**. The line moves because #702's compensation now reaches 42 rows it could not reach before, and it moves the way #702 predicts: fewer iterations, same status, same objective to 12 digits. A trajectory change, recorded as one — not a cost, so nothing to file. The other 115 legs are byte-identical, including every `qcqp`-shaped fixture and both `eigen*` models #702 itself moved.

**The sweep is unchanged by the 10fb7c8 fix** — still this one line. That is expected rather than reassuring: no repo fixture carries the ~2⁵³ of dynamic range the refusal needs, which is the same reason the defect survived review-by-corpus.

The sweep runs default options, which is how this was missed in Q4, so #512's case is run directly — `airport.nl` at `mu_strategy=adaptive tol=1e-12`:

| | `ac18ba6` | with the fix |
|---|---|---|
| iterations | 16 | **16** |
| status | `SearchDirectionBecomesTooSmall` | same |
| final dual inf | 9.42357762137488e-13 | **identical** |
| final constraint violation | 5.551115123125783e-16 | **identical** |
| final KKT error | 7.780583180375908e-12 | **identical** |
| objective | 47952.70164695195 | 47952.70164695197 |

Every number the stopping test reads is identical; the objective moves two ulps, which is the compensated sum.

**Load cost.** A body the expanded gate refuses now takes a second structural walk before falling back to a tape. Worst case, a generated 2000-row model whose rows are all transcendental so every row asks and every row is refused: load (`max_iter=0`, 25 runs) 33.8 ms → 35.8 ms, on a figure that is mostly process start. Both walks abort at the first offending leaf.

**Speed, in the regime this opens.** No fixture is big enough to time (`airport.nl` is 0.17 s either way), so `scripts/gen-qcqp-nl.py` gains `--form factored`. 1000 vars, 5107 rows, 7 nonlinear, 400 residuals of 20 variables each:

| | `ac18ba6` | with the fix | |
|---|---|---|---|
| iterations | 53 | 53 | unchanged |
| objective | 331.34374901077535 | 331.3437490107755 | agree to 15 digits |
| wall (best of 2) | 23.05 s | **15.40 s** | 1.50× |
| `OverallAlgorithm` | 21.269 s | **14.310 s** | 1.49× |
| `LagrangianHessianEvaluations` | 6.702 s | **0.256 s** | **26×** |
| peak RSS | 255 MB | 251 MB | no win either way |

**This is a generated instance and no number in that table may be quoted as a Mittelmann one.** The RSS line is measured at `residual-nnz=20` and does **not** generalize: `Σ 2wbbᵀ` is dense `O(nnz²)` per residual, so one *wide* residual is a different regime — at `n = 3000`, single residual, peak RSS 2188 MB → 2557 MB (+17%) with wall 36.3 s → 23.0 s. Still net favourable, and recorded in §0k rather than left as "memory is a wash".

Untouched: classification and routing read `analyze_quadratic_full`, not `admitted_quad_form`, so no model changes solver. The Q6 constant-derivative proofs read `provably_affine`. `POUNCE_DBG_NO_QUAD` still disables both arms.

## Tests

`quad_evaluator_differential.rs` compares `f`, `∇f`, `g`, `∇g` and `∇²L` against the `POUNCE_DBG_NO_QUAD=1` tape at four points on every `.nl` file in the repo. **It passes with `REL_TOL`, `WORST_OBSERVED_REL = 1e-14` and `MAX_HESS_ULPS = 1` exactly as they were.** No pin relaxed.

Its randomized battery **never reached the factored arm** — it emitted only flat sums of monomials, so every factored body the suite compared came from the 24 repo fixtures: real-writer models, unit weights, well-scaled residuals. That is the same hole #683 sat in, and review's witness went straight through it. `battery_square` now emits `w·(bᵀx + d)²` over a randomized affine base; **562 bodies per run take the factored arm**, and the count is asserted (`factored_bodies >= 100`) so a generator change cannot quietly close it again.

Stated plainly because it would otherwise be assumed: the battery keeps its coefficients within a few orders of magnitude on purpose, so it does **not** reproduce the loss case — reverting the `factored_hessian` refusal leaves the battery green. It covers the arm's *arithmetic*; the loss case is pinned by dedicated tests.

39 unit tests on the recognizer (`nl_quadratic`), 25 on the storage and evaluation (`pounce-nlp::quadratic`), 10 in `issue_685_cancelled_quadratic_evaluation.rs` — covering the admitted shapes, all four refusals, the compensation decision, and a 250 000-term chain for the stack-safety property the module is built around.

**On "tests fail on the parent commit":** most are new-feature tests referencing APIs the parent does not have, so they do not compile there — that box does not apply in its usual sense and I am not ticking it. These *are* regression pins, each verified to bite by reverting the code and watching it fail:

- `the_weight_multiplies_the_square_not_the_residual` — revert `w·(l·l)` to `(w·l)·l`: `8.470000000000002` against `8.47`.
- `a_row_that_dropped_a_term_is_not_admitted_as_a_factored_form_either` — remove the `is_expanded_quadratic` test: fails with two others.
- `the_factored_outer_sum_does_not_lose_small_terms_either` — naive accumulator: `0.0` against `200.0`.
- `a_factored_row_that_dropped_a_hessian_entry_keeps_its_tape` and `a_factored_form_that_loses_a_hessian_entry_is_refused` — neuter the cancellation refusal: both fail, and nothing else does.
- `a_weight_whose_double_overflows_is_refused` and `a_coefficient_product_that_underflows_is_refused` — disable either half of the product guard: exactly the matching test fails.

Deliberately not pinned: the generated-model speedup and the load-cost figure, both machine-dependent. Recorded in §0k with how to reproduce.

`cargo fmt --all -- --check` clean, `cargo clippy --all-targets -- -D clippy::correctness -D clippy::suspicious` clean, `cargo test --workspace --exclude pounce-hsl` green, `scripts/check-release-consistency.sh` OK. `pounce-hsl` is excluded because MA57 does not link in this container: pre-existing, unrelated to a diff that touches neither that crate nor anything it links.

No book page: this adds no option and no user-facing API. `POUNCE_DBG_NO_QUAD` in `docs/src/options.md` is updated to say it covers both read-outs.

## Two rounds of corrections, kept visible

**Round 1 — rebased onto #702.** Three claims were true against `5715579` and not against `ac18ba6`, corrected in place rather than dropped: the sweep is no longer byte-identical (one line moves, with an attribution experiment saying which change moved it); #512's objective moves two ulps; and the #685 gate mechanism, written up as a Hessian-pattern divergence measured on *uncompensated* code, diverges in the **value** too under compensation.

**Round 2 — the review finding.** The gate did not cover the new arm, and this body asserted in three places that it did. Fixed in 10fb7c8, with the CHANGELOG line that said lossy bodies "still keep their tapes" rewritten to name both routes to that loss, and "No answer changes:" narrowed to "No model changes its answer:". Also from that round: two doc claims corrected (`peel_square` folds constants in *reverse* source order; the reversed spine push does not buy the same association `recognize_expr` uses), a `debug_assert` making the `SquareTerm::coefs` ordering contract enforceable, and `gen-qcqp-nl.py` rejecting `--residuals`/`--residual-nnz` under `--form expanded` instead of ignoring them.

Earlier still, I claimed the #685 CLI test caught the `w·(l·l)` association bug. True before the gate was added, false after; the Tests section says what actually bites.

Full write-up in `dev-notes/quadratic-structure-exploitation.md` §0k, with the §3.5 and §0e corrections made in place.

---

- [ ] Tests fail on the parent commit for the stated reason — **partially; see Tests.** The regression pins do; the new-feature tests do not compile on the parent.
- [x] Book page under `docs/src/` updated — `options.md`; no new page, and why is stated above


jkitchin:
## Review — factored quadratics from their own structure (#673)

Reviewed at `5f0cd0b5` against `ac18ba6d`, in a throwaway worktree. The engineering here is careful and every headline measurement in the PR body reproduces exactly on my machine. **One merge-blocking finding**, in §3; everything else is verified-good or a nit.

### 1. Math, checked by hand

- **Hessian convention.** `s = 2·w`, `hess[(i,j)] += s·bi·bj`. For `w(bᵀx+d)²`, `∂²/∂xᵢ∂xⱼ = 2w·bᵢbⱼ` on and off the diagonal, so the upper-triangle convention `push_form` documents is honoured with no double-counting. `diagonal_square` returning the *polynomial* coefficient `c` rather than `2c` is right given `push_factored_form` applies the 2.
- The **`½` is correctly absent** on the factored arm and correctly retained on the expanded one.
- **Gradient support taken from `bₖ`, not from `H`** is the right call, and the justification in the doc holds: `(x₀+x₁)² − (x₀−x₁)²` has an exactly-zero Hessian diagonal but a gradient that depends on both variables.
- The `f = self.constant.len()` offset bookkeeping is correct in both `push_form` and `push_factored_form` — the new form's ranges are `form_*[f] .. form_*[f+1]` and `constant` is pushed last in `finish_form`.
- I re-derived every assertion in the four new value/gradient/Hessian tests. All four are arithmetically correct as written.
- `the_weight_multiplies_the_square_not_the_residual` is a genuine pin, and the witness `7·(1.1)²` → `8.47` vs `8.470000000000002` is right. The earlier `2⁵³·3²` version would have pinned nothing; good catch.

Two doc-only inaccuracies, neither a defect:

- `peel_square` pops its product tree off a stack, so multiple constant factors fold in **reverse** source order: `a·(b·(c·s))` becomes `((1·c)·b)·a·s`. The doc admits it "reassociates the writer's own constants" but doesn't say the order is reversed.
- `nl_quadratic.rs:1120-1123` claims the reversed spine push makes leftovers fold "front to back … the same terms `recognize_expr` sums." That is only true for a left-leaning spine — this is a *left* fold where `recognize_expr` does a *tree* fold, and `Quad2::add` additionally swaps operands by map width. Harmless (`lost_terms` gates the lossy cases), but the comment overstates what the pop order buys.

### 2. Blocking — the #685 gate does not cover the new arm

`admitted_factored_form` (`nl_reader.rs:546`) gates on `is_expanded_quadratic` only. Its own doc states the purpose:

> A body refused because a term went **missing** in the fold (`Quad2::lost_terms`, gh #685) keeps its tape, and the explicit `is_expanded_quadratic` test here is what keeps it there.

That holds only for lossy bodies that are *also* flat sums of monomials. A genuinely factored body can lose a term in `Σ 2wₖbₖbₖᵀ` — `quadratic.rs:299-307` is an unguarded naive `+=` with no `lost_terms` analogue for this representation — and nothing refuses it.

Witness, 2 vars, row 0 is `body ≤ 5`, row 1 is `sin(x₁)=0` (the gh#685 test model):

    body = (2**27 * x0)**2 + (x0 + x1)**2 - (x0 * 2**27)**2

True value at `x = (3,1)` is 16; true `H = [[2,2],[2,2]]`. Measured through `parse_nl_text_with_quadratic` + `NlTnlp`:

| predicate | result |
|---|---|
| `quad_terms_dropped()` | true — this **is** a gh#685 body by the repo's own predicate |
| `admitted_quad_form()` | false — correctly refused |
| `analyze_quadratic_full()` | false — correctly refused |
| `admitted_factored_form()` | **true — admitted** |

End-to-end at `x = (3,1)`, `lambda = (1,0)`:

| | `ac18ba6d` | this branch |
|---|---|---|
| `eval_g[0]`, fast path | 0.0 | **16.0** |
| `eval_g[0]`, `POUNCE_DBG_NO_QUAD=1` | 0.0 | 0.0 |
| `nnz_h_lag` pairs, fast path | `[(0,0),(1,0),(1,1)]` | **`[(1,0),(1,1)]`** |
| `nnz_h_lag` pairs, tape | `[(0,0),(1,0),(1,1)]` | `[(0,0),(1,0),(1,1)]` |

On `main` the two routes agree. Here they disagree on the value of a constraint body by 16 against a `≤ 5` bound, and on the Hessian pattern (2 entries vs 3). That is the `POUNCE_DBG_NO_QUAD`-shaped divergence this PR itself argues is still the defect. `a_row_that_dropped_a_term_is_not_admitted_as_a_factored_form_either` does not catch it, because it only feeds the three *expanded* bodies from the gh#685 file.

Scope, stated honestly:

- The fast path is the **more accurate** side here — 16 is the true value. By the repo's own stated principle ("the tape is the reference because it is what the row means *to this solver*, not because it is exact") that is still a defect, and it is the same argument used to justify the gate that *was* added.
- The mechanism needs ~2⁵³ of dynamic range across squared terms with cancellation. On the diagonal that requires a negative weight (difference-of-squares / DC formulations). **Off-diagonal it does not**: `(1e9*x0 + 1e9*x1)**2 + (x0+x1)**2 + (1e9*x0 - 1e9*x1)**2` loses `H(0,1)` with all weights positive. Coefficients of 1e9 are not exotic in badly scaled models.
- I could **not** construct a case where the fast-path Hessian is numerically *worse* than the tape's — in my witness both compute `H(0,0) = 0`. The mechanism exists but I did not realize it, and I would rather say so than guess.

Suggested direction — explicitly **not** "call `recognize_expr` and check `lost_terms`", which would expand the very bodies this PR exists to avoid expanding and would refuse real least-squares models. Track loss inside the `Σ 2wₖbₖbₖᵀ` accumulation at `quadratic.rs:304` the way `Quad2::merge` does: an `add_is_exact`-style check that refuses the form when an entry is dropped inexactly. Failing that, correct the three claims and file the gap with an owner — this repo's own convention for an accepted cost.

Whichever way it goes, the CHANGELOG line *"A body mixing a square with a cross term `c·xᵢxⱼ`, one whose sum spine goes through a declared common expression, and one that lost a coefficient to cancellation (#685) all still keep their tapes"* is false as written and has to change. The "No answer changes:" heading is likewise stronger than the evidence supports.

### 3. Trajectory — sweep independently reproduced, exactly

Re-ran it rather than taking the PR's word:

    scripts/sweep-fixtures.sh <ac18ba6d binary> base.txt   -> 116 lines
    scripts/sweep-fixtures.sh <5f0cd0b5 binary> new.txt    -> 116 lines
    diff base.txt new.txt
    59c59
    < lbfgs  airport  SolveSucceeded  it=59  obj=47952.70165
    ---
    > lbfgs  airport  SolveSucceeded  it=57  obj=47952.70165

Both legs, one line, identical to the claim. I also reproduced the **attribution experiment**: patching the factored arm's outer accumulator back to a naive `f64` sum and rebuilding gives `lbfgs airport` = 59 iterations, objective `47952.70164698971` — bit-identical to baseline in the last digit. So the moved line really is #702's compensation reaching 42 newly-eligible rows, not the new read-out. That is the right standard for this repo and it was met.

Also reproduced: the gh#512 case (`airport.nl mu_strategy=adaptive tol=1e-12`), both at 16 iterations / `SearchDirectionBecomesTooSmall`, objective `47952.70164695195` → `...197` (2 ulps); `[quad stats]` for `airport.nl` going `rows=1/42 obj=taped` → `rows=42/42 obj=quadratic`; exactly 24 fixtures whose `[quad stats]` line moves; and `scripts/gen-qcqp-nl.py` expanded output byte-identical to the pre-change script (`cmp` clean).

### 4. Test coverage

The refusals are well covered — cross monomial, product of two different affine forms, quartic, transcendental, non-expanded base, lost leftover, `Cse` on the spine, zero weight, `x²/0`, 250 000-term chain. Two gaps:

- **The randomized differential battery never reaches the new arm.** `quad_evaluator_differential.rs`'s `battery_monomial`/`battery_body` emit only flat sums of monomials; no `Pow(affine, 2)` with a non-atomic base is ever generated. That battery exists specifically to cover shapes the writers do not produce — which is exactly the hole gh#683 sat in — and the new representation is precisely such a shape. All differential coverage of the factored arm is therefore the 24 repo fixtures, real-writer models with unit weights and well-scaled residuals. My §2 witness fits straight through this gap. A `Pow(battery_body(degree ≤ 1), 2)` leaf would be a small change with high value.
- **Stale comment made false by this PR.** `quad_evaluator_differential.rs:621-622`: "A flat sum of monomials — the shape `is_expanded_quadratic` admits, *and therefore the only shape that reaches the fast path at all*." No longer true, and it is the sentence that would mislead the next reader into thinking the new arm is randomly covered.

One suspected-not-verified item: `peel_square` admits any finite nonzero weight, so `s = 2.0 * t.weight` can overflow to `inf` for `|w| > 8.99e307` and poison a Hessian entry for a body whose value is finite. Contrived, and I did not build an end-to-end witness.

### 5. Local verification

    cargo test -p pounce-nlp -p pounce-nl --release        -> 204/101/12/1 passed, 0 failed
    cargo test -p pounce-cli --release  (6 quad/#683/#685 test binaries)
                                                          -> 6/8/4/3/7/3 passed, 0 failed
    cargo fmt --all -- --check                            -> clean
    cargo clippy ... -D clippy::correctness -D clippy::suspicious  -> exit 0
    scripts/check-release-consistency.sh                  -> OK
    gh pr checks 711                                      -> 11/11 pass, MERGEABLE / CLEAN

`Fixes #673` is in the PR **body**, not just the title. No crates added or removed and no new inter-crate dependency, so the release guard is unaffected. Dev-note §0k is thorough, and the §3.5 / §0e corrections being made in place rather than quietly dropped is the right habit.

### 6. Non-blocking observations

- **Memory on a wide residual.** `push_factored_form` materializes `Σ 2wbbᵀ`, dense `O(nnz²)` per residual. A single wide residual — `(Σᵢ xᵢ − 1)²`, an extremely common idiom — used to be an `O(n)` tape. Measured on a generated `n=3000` single-residual model at `max_iter=0`: peak RSS 2188 MB → 2557 MB (+17%), wall 36.3 s → 23.0 s, `stored_h_entries` 0 → 18 000 000. Net favourable, because the baseline already colours the dense pattern for the tape — so this is an observation, not a regression. But the dev-note's "memory is a wash" was measured only at `residual-nnz=20` and does not generalize.
- `add_gradient`'s factored branch accumulates uncompensated into `out`. Consistent with the expanded arm, so not an inconsistency — just worth knowing the "slightly better than the tape" claim applies to `value` only.
- `SquareTerm`'s `coefs` invariant ("ascending, free of zeros") is a `pub` contract enforced only by a `debug_assert` on the Hessian triangle, not on the ordering — while `push_factored_form`'s `t.coefs[a..]` loop *relies* on ascending order for `i ≤ j`. An out-of-order caller would silently write a lower-triangle entry. A `debug_assert` on sortedness would make the contract enforceable; the only in-tree caller comes from a `BTreeMap`, so this is a nit.
- `gen-qcqp-nl.py` accepts `--residuals` / `--residual-nnz` silently under `--form expanded`, where they do nothing.

### Bottom line

Merge-blocking: **§2 only** — because it re-opens the exact defect class (gh#683 → gh#685) this codebase has already shipped twice, and because the PR asserts in three places that it is closed when it is not. Either extend the gate to the factored Hessian accumulation, or correct the three claims and file the gap with an owner.

Everything else is verified-good — the sweep discipline, the attribution experiment, the arithmetic, CI, the release guard, the closing keyword — or a nit. The two items in §4 are cheap and worth doing in the same pass.

jkitchin:
Thank you for the depth here — the §2 finding is real, I reproduced it before touching anything, and it's fixed in 10fb7c8.

## §2 — reproduced, then fixed at the accumulation

Your witness, measured on `5f0cd0b` before any change:

```
admitted_quad_form=false  admitted_factored_form=true
quad=true   g=[16.0, 0.841…]  nnz_h=2  pat=[(1,0), (1,1)]
quad=false  g=[ 0.0, 0.841…]  nnz_h=3  pat=[(0,0), (1,0), (1,1)]
```

Exactly your table. A body `quad_terms_dropped()` calls lossy, refused by `admitted_quad_form`, walking in through the factored door.

I took your suggested direction rather than the `recognize_expr`/`lost_terms` one, for the reason you gave. `factored_hessian` now tracks exactness the way `Quad2::merge` does, `push_factored_form` returns `Option`, and it refuses **before touching any state** so a rejected body leaves nothing half-built — `push_body_form` falls through to the tape like every other refusal. Keeping the check on the arithmetic it gates, rather than re-deriving the accumulation in the recognizer, was the part of your suggestion that mattered most: the two can't drift.

**One thing I got wrong first, which is worth recording.** My initial check tested inexactness at the dropping fold and *did not fire on your witness*. The rounding is on the second term (`2·2⁵⁴ + 2` → `2·2⁵⁴`) and the drop is on the third — whose add is **exact**. The inexactness has to be carried per entry, which is what `Quad2` does with `carried_inexact` and what I'd skimmed. The test caught it, and the sticky flag is now in the code with that reason written next to it.

gh#687's rule is unchanged and pinned in both directions: `(x₀+x₁)² − (x₀−x₁)²` cancels both diagonal entries *exactly* and still takes the fast path.

Your off-diagonal point drove the test selection — `(10⁹x₀ + 10⁹x₁)² + (x₀ + x₁)² + (10⁹x₀ − 10⁹x₁)²` with every weight positive is a separate test, because "needs a negative weight" would have been the wrong lesson for the next reader.

Your **suspected-not-verified** overflow item is settled by the same fix: `mul_is_exact(2.0, w, s)` refuses a weight past `8.99e307`, since an infinity has swallowed every term it touched. Pinned.

The predicates `add_is_exact` / `mul_is_exact` / `is_live` moved from `pounce-nl` to `pounce-common` — `pounce-nlp` can't name `pounce-nl`, and two copies of the gh#683/#685/#687 argument free to drift was the worse outcome.

## §4 — battery, and an honest limit on what it buys

`battery_square` now emits `w·(bᵀx + d)²` over a randomized affine base, one body in four. **562 bodies per run take the factored arm**, up from zero, with `REL_TOL`, `WORST_OBSERVED_REL` and `MAX_HESS_ULPS` unchanged. The count is asserted (`factored_bodies >= 100`) so a generator change can't quietly close the coverage again — that seemed the actual lesson of the hole.

But I checked whether it would have caught your witness, and **it would not**: reverting the `factored_hessian` refusal leaves the battery green. The file keeps coefficients within a few orders of magnitude on purpose, so the battery covers the arm's *arithmetic*, not the loss case. That's now stated in the dev-note rather than left for someone to assume. The loss case is pinned by the dedicated tests.

The stale `quad_evaluator_differential.rs:621` comment is gone with the generator it described.

## §1 — both doc inaccuracies corrected

- `peel_square` pops a stack, so `a·(b·(c·s))` folds as `((1·c)·b)·a`. Now says so, with why a reader reconstructing a last-ulp difference needs the real order.
- The reversed spine push does **not** buy the same association `recognize_expr` uses — that folds the tree, and `Quad2::add` swaps operands by map width, so the two can differ on a leaning spine. The comment now says that, and says what actually makes it safe (`lost_terms`).

## §6 and the CHANGELOG

- The CHANGELOG line was false as written. It now says there are two routes to that loss — shape, and the Hessian fold — and that both fall back to the tape. "No answer changes:" is now "No model changes its answer:", which is the claim the evidence supports.
- `SquareTerm::coefs` ascending: `debug_assert` on sortedness in `factored_hessian`, and the doc now says the upper-triangle walk depends on it. You're right that an out-of-order caller would silently write a lower-triangle entry.
- `gen-qcqp-nl.py` now errors on `--residuals` / `--residual-nnz` under `--form expanded` rather than ignoring them — a shell history reading "measured with 400 residuals" over an expanded model is exactly the trap.
- The wide-residual memory figures are recorded beside the `residual-nnz=20` table, with your framing: net favourable, an observation not a regression, and "memory is a wash" only ever true at the width it was measured at.

`add_gradient` staying uncompensated is deliberate and consistent with the expanded arm; I've left the "slightly better than the tape" claim scoped to `value`, as you noted.

## Re-verification after the fix

| | |
|---|---|
| sweep vs `ac18ba6`, both legs | **unchanged** — still the one `lbfgs airport` line, 59 → 57 |
| `airport.nl` routing | still `rows=42/42 obj=quadratic` |
| gh#512 case | 16 iterations, `SearchDirectionBecomesTooSmall`, KKT identical to baseline |
| `cargo test --workspace --exclude pounce-hsl` | 262 binaries, cargo exit 0 |
| fmt / clippy / `check-release-consistency.sh` | clean |

The sweep not moving is expected rather than reassuring: no repo fixture carries the ~2⁵³ of dynamic range the mechanism needs, which is the same reason it survived review-by-corpus in the first place.

Two things I did **not** do, so they're visible rather than silent: I did not file a follow-up issue, because the gap is closed in this PR rather than accepted; and I did not widen the battery's coefficient range to chase the loss case, because that would rediscover the ill-scaled storage defect once per seed, which the file explicitly puts out of scope.


---

jkitchin:
Verified at `10fb7c89`, not taken on trust. The §2 fix holds up.

**The guard is real, not just present.** Neutering the cancellation refusal —
`if false && was != 0.0 && !is_live(v) && inexact` — turns
`a_factored_row_that_dropped_a_hessian_entry_keeps_its_tape` red and leaves the
other nine green. So the new test is pinning the branch it claims to pin, and
it is the only thing standing between that branch and a silent regression.

**The sticky flag is the right call and I had missed it too.** Your note that
the rounding lands on one term and the drop on another, so a per-fold check
does not fire, is correct and is the non-obvious part of this. Carrying it per
entry the way `Quad2` does is what makes it work.

**Your overflow claim checks out**, and I withdraw the "suspected, not
verified" framing — it is pinned by
`quadratic::tests::a_weight_whose_double_overflows_is_refused`, which I found
only after disabling the branch and watching it fail. I had been looking in the
integration file.

**One small gap, not blocking.** Disabling *both* early refusals together —
`mul_is_exact(2.0, t.weight, s)` and `product_lost` — fails only the overflow
unit test. The underflow half of `product_lost`, the `10⁻²⁰⁰ · 10⁻²⁰⁰` case the
comment names, appears to have no test that fails when it is removed: the whole
workspace stays green except for that one overflow assertion. It is the same
shape as the hole the review found, one predicate over. Worth one more unit
test alongside the overflow one.

Everything else in your reply matches what I asked for, and the two places you
went further than I suggested are both improvements: keeping the check on the
arithmetic rather than in the recognizer, and asserting `factored_bodies >= 100`
so the battery cannot quietly lose the arm again.

The honesty about the battery **not** catching the witness — you reverted the
refusal and confirmed it stays green — is the most useful sentence in the
reply. That is the fact that would otherwise have been assumed by the next
person to touch the generator, and it is now written down instead.

No remaining blocking findings from me.

PR #716, opened 2026-08-19
Empty the silent-option debt list: wire, refuse, or warn for all 149

### LINKED ISSUE #551: Registered-but-unread options: the rest of #191 round 2 (split from #486)

### LINKED ISSUE #682: Empty the registered-but-unread option list: 151 options, grouped (follow-up to #677)
Follow-up to #677, continuing #551. Splitting this out so #680 can ship on the L-BFGS work it is really about.

#677 shipped `no_silent_options.rs`, which fails the build for any registered option that is neither read nor declared unimplemented. That closed the *class* — nothing new can go silent — and in doing so produced an exact inventory of what is already silent. This issue is the work of emptying that list.

## The number is 151, not the ~24 #551 estimated

The earlier count came from grepping option names. That undercounts twice over: it missed read sites reached through the `read_yes`/`read_num`/`read_int` helpers, and it excluded the per-backend families by pattern. The mechanical scan in `no_silent_options.rs` — which discovers the accessor helpers from the source rather than listing them — is the honest count.

Worth noting how easily this goes wrong: the first version of that scan hard-coded the accessor list and immediately reported three *wired* options as silent, because `read_yes` existed and was not in the list. That is #551's caution 1 happening inside the tool built to enforce it.

## Groups

These are the constant names in `crates/pounce-algorithm/tests/no_silent_options.rs`. Each group is independent; deleting an entry from the list is the visible act of fixing one, and the test fails if you delete without fixing, or fix without deleting.

| group | n | notes |
|---|---|---|
| `BACKEND_KNOBS` | 111 | **Policy decision, not plumbing** — see below |
| `LINE_SEARCH` | 11 | `alpha_red_factor`, `eta_penalty`, `nu_inc`, `nu_init`, `rho`, `theta_min`, `delta`, `filter_margin_fact`, `filter_max_margin`, `alpha_for_y_tol`, `accept_after_max_steps` |
| `SENSITIVITY` | 7 | sIPOPT — `run_sens`, `n_sens_steps`, `sens_bound_eps`, `sens_boundcheck`, `sens_max_pdpert`, `compute_red_hessian`, `rh_eigendecomp` |
| `BARRIER_KKT` | 5 | `tau_min`, `s_max`, `fixed_mu_oracle`, `neg_curv_test_tol`, `neg_curv_test_reg` |
| `RESTORATION` | 5 | `max_resto_iter`, `resto_failure_feasibility_threshold`, `limited_memory_special_for_resto`, `expect_infeasible_problem_ctol`/`_ytol` |
| `CORRECTOR` | 4 | `corrector_type`, `corrector_compl_avrg_red_fact`, `skip_corr_if_neg_curv`, `skip_corr_in_monotone_mode` |
| `NLP_HINTS` | 4 | `grad_f_constant`, `hessian_constant`, `jac_c_constant`, `jac_d_constant` |
| `MISC` | 4 | `derivative_test_perturbation`, `derivative_test_tol`, `limited_memory_max_skipping`, `timing_statistics` |

## The hard part is not the wiring

**Every option needs its struct default compared against its registered default before the read site goes in.** If they differ, adding the read site does not merely enable an option — it silently changes behaviour for every user who never set it. That is exactly what #677 was: `limited_memory_initialization` registered `scalar1` while the code ran `scalar2`, and the gap was invisible because nothing compares the registry against behaviour.

So the rule per option is: wire it, **keep the current effective default**, and report any mismatch separately. Moving a default is a trajectory change and needs `scripts/sweep-fixtures.sh` (both legs) with every moved line explained. On #677 that part — four fixtures run down individually — was most of the work, not the wiring.

And per #551, the deliverable is a test that the option **changes behaviour**, not that it propagates: "A read site that parses a value and discards it is the same silent no-op this whole line of work exists to kill, and it is indistinguishable from a real fix by inspection." `recalc_y_changes_the_lbfgs_trajectory` and `slack_based_scaling_changes_the_solve`, both added on #680, are the pattern.

## `BACKEND_KNOBS` needs a decision first

111 tuning knobs for linear-solver backends pounce does not implement (`ma27_*`, `ma77_*`, `ma86_*`, `ma97_*`, `mumps_*`, `pardiso_*`, `pardisomkl_*`, `spral_*`, `wsmp_*`, `pardisolib`). pounce ships FERAL and MA57.

Refusing them is the obvious move and is probably wrong: it breaks the stated goal that an `ipopt.opt` written for Ipopt parses unchanged, since such a file may configure several backends, and it would fail solves for users not even using those backends. The precedent that fits better is `a_caching_hint_warns_but_solves` — warn and proceed. A single grouped warning per backend family ("pounce does not implement the MA97 backend; 4 `ma97_*` options ignored") beats 15 lines of noise, and it should fire only on an explicitly-set option so a default run stays silent (`a_default_run_is_silent` must keep passing).

This is one mechanism once the call is made, and it removes 111 of the 151.

## Partial work exists

Six agents were started on these groups and stopped mid-run when parallel builds exhausted the disk. Their work is preserved, **unverified and unreviewed**, on `wip/551-*` branches:

| branch | group |
|---|---|
| `wip/551-a47e5ef539224f086` | LINE_SEARCH |
| `wip/551-ac2a3099c9e2f2a85` | BARRIER_KKT |
| `wip/551-a5f59fa041b61f884` | SENSITIVITY |
| `wip/551-a3f29dbeeec11e8d4` | RESTORATION + CORRECTOR |
| `wip/551-a34a65f7499075ee2` | BACKEND_KNOBS |
| `wip/551-a01831152cc1b85f4` | NLP_HINTS + MISC |

Treat all of it as a starting point, not as reviewed work — in particular none of it has had the default-mismatch check verified independently. Two of the six had reported `pounce-algorithm` green before being stopped, but neither had finished the CLI suite.

Practical note for whoever picks this up: these builds are large. Running more than two of these concurrently in separate worktrees exhausted a 252 GB volume (25 GB of build artifacts across six trees), which is what stopped them.

jkitchin:
Fixes #682
Fixes #551

## Problem

149 registered options were accepted and then ignored. Setting one did
nothing and said nothing — the run looked like it had honoured the file.
`no_silent_options.rs` knew about them, but only as a `known_debt`
allowlist that let CI stay green while the list stood.

| | silent options | `known_debt` entries |
|---|---|---|
| before | 149 | 149 |
| after | 0 | 0 |

The allowlist is now empty. Every registered option has a read site the
scan can see, a refusal that names the missing feature, or a warning.

## Cause

Three different causes, which is why the count stayed put so long:

- **31 genuinely unwired** — registered with upstream's default, no
  consumer. `tau_min`, `s_max`, the penalty-acceptor ν knobs, the sIPOPT
  names, and the rest. One of the 31, `max_resto_iter`, was hidden even
  from a grep of the option name: it has been enforced all along under the
  field name `maximum_resto_iters`, from a constant rather than the
  option.
- **111 knobs for backends POUNCE does not ship** (`ma27_*`, `mumps_*`,
  `pardiso_*`, …) — nothing to wire them to.
- **7 false positives.** `timing_statistics`, the two derivative-checker
  knobs and the four constant-derivative hints were wired *and consumed*.
  Each reached its accessor through a loop variable
  (`["timing_statistics", "print_timing_statistics"].iter().any(…)`,
  `HINT_OPTIONS.map(|name| …)`) or a local helper named `num` instead of
  `read_num`. The scan keys on the option name **as it appears at the
  accessor**, so all seven read as "no key here". This is #551's caution 2
  turned inside out: the scan is right to key that way, and the read sites
  now spell the name out there.

## Fix

Per option, one of three dispositions:

| group | n | wired | refused | already fine |
|---|---|---|---|---|
| LINE_SEARCH | 10 | 8 | 2 | — |
| BARRIER_KKT | 5 | 4 | 1 | — |
| RESTORATION + CORRECTOR | 9 | 1 | 8 | — |
| SENSITIVITY | 7 | 6 | 1 | — |
| NLP_HINTS + MISC | 7 | — | — | 7 |
| BACKEND_KNOBS | 111 | — | — | warn, or refuse — see below |

**Refusal over silence, but not over portability — until the knobs are
all there is.** The 111 backend knobs *warn* rather than refuse: a
portable `ipopt.opt` routinely configures several backends at once so one
file runs everywhere, and refusing would fail that file over knobs the run
never touches. One line per backend family, not per option, and only when
the value differs from the registered default — a default run stays
completely silent.

That argument has a premise: the file has other business here. A run whose
options are **nothing but** backend knobs is refused, because nothing in
it survives — warning and solving would answer "tune the linear solver" by
tuning nothing and reporting success, which is the shape of the defect
rather than a fix for it.

| what the run sets | result |
|---|---|
| backend knobs only | **error**, exit 2 |
| backend knobs + any option POUNCE reads | warning, solve continues |
| backend knobs at their registered defaults | silent, solve continues |
| no backend knobs | silent, solve continues |

Two boundary calls in row 2. It counts an option's **presence**, not
whether it was changed: a caller who writes `tol` at its default has
stated an intention about this solve, and presence is the direction that
makes the refusal rarer rather than more eager. And `option_file_name` is
exempt from counting as content — it says where the options came from, not
what to solve, so pointing at a backend-only `ipopt.opt` is refused
exactly as passing the same knobs on the command line is. Without that
exemption, the most common way to supply a backend-only file would have
been the one case that could never be caught.

A file that *selects* the backend it tunes never reaches any of this:
`linear_solver=ma97` is already refused on its own.

`hsllib` keeps its refusal, because POUNCE *has* an HSL backend, so that
one is a caller reaching for a solver it can actually run by a mechanism
it does not have.

**Rejected: adopting the registered default wherever it differed.** Three
options have a registered default POUNCE does not honour — `max_resto_iter`
(registry 3000000, POUNCE enforces 3000), `run_sens`, and `sens_max_pdpert`.
Taking upstream's number at the same time as adding the read site would
change answers for people who never set the option, hiding a trajectory
change inside a wiring PR. Each keeps POUNCE's behaviour for the unset
case, and a test pins both halves so neither drifts:
`max_resto_iter_default_is_pounces_cap_not_the_registered_one`. Moving
them is separate work with its own measurement.

**One option needed the feature built first.** `neg_curv_test_tol` /
`_reg` configure the inertia-free curvature test of Zavala & Chiang (2014);
the old field only turned the inertia check *off* with nothing in its
place, so honouring the option before this would have been worse than
ignoring it.

## Blast radius

**Bit-identical except where an option is explicitly set.** The fixture
sweep is byte-identical against `ac18ba6` across all 118 fixture-legs, on
**both** the `exact` and `lbfgs` legs.

That holds because no default moved. Every struct default was checked
against `upstream_options.rs` one at a time — for most of these options the
two already agreed, which is exactly why the omissions were invisible.

What *does* move is a run that sets one of the newly-live options, and two
are worth naming because they can make things worse:

- `tau_min=0.5` moves 33 of 118 fixture-legs (`csfi2` 35 → 28, `deb7`
  154 → 139, `cresc4` 81 → **1081**).
- `neg_curv_test_tol=1e-11` moves 11 of 59 fixtures — some sharply better
  (`csfi2` 35 → 27 and to `Solve_Succeeded`), some sharply worse
  (`eigenb2` 67 → 960), and on `autocorr_bern55-06` and `pooling_rt2stp`
  to a **different, worse objective while still reporting success**.

Both are off by default. `docs/src/options.md` now carries that measured
table instead of the assurance it had before — a tolerance-legal wrong
answer is the gh#544 failure mode, and a reader deciding whether to enable
the curvature test is owed the real numbers.

## Tests

Six new test files plus additions to five existing ones; the
`known_debt` allowlist in `no_silent_options.rs` is now empty, so the scan
itself is the backstop.

Confirmed failing on the parent commit (`ac18ba6`) by checking out a
worktree at `origin/main`, copying the new tests in, and running them:

- `no_silent_options` — fails at the probe, with
  "`timing_statistics` has a read site in the source but the scan missed
  it". That probe is what pins the literal-key rewrite of the seven false
  positives; on the parent the read site is still in loop-variable form.
- `sipopt_options_wiring` — 2 assertion failures
  (`every_registered_sipopt_option_is_consumed_or_refused`,
  `n_sens_steps_is_refused_above_its_single_implemented_tier`).
- `line_search_options_wiring`, `barrier_kkt_options`,
  `algorithm_options_wiring` — do not compile on the parent: `no field
  accept_after_max_steps on LineSearchOptions`, `no field tau_min on
  MuOptions`, `no field max_resto_iter on RestoOptions`, and so on. The
  fields they assert on did not exist, which is the defect.

Two behaviours are pinned beyond "the option is read":

- `each_constant_derivative_hint_lights_up_its_own_slot` — the hint
  rewrite touches code where `reconcile` pairs each flag with the model's
  proof for the same index, and a transposed pair would silently reuse the
  wrong derivative. This asserts one hint lights its own slot and no other.
- `the_derivative_checker_knobs_change_the_verdict` — drives an exactly
  correct gradient through the checker, shows a coarse
  `derivative_test_perturbation` making it look wrong, and
  `derivative_test_tol` clearing it again.

The backend-only refusal is pinned at both ends — `backend_knobs_alone_are_refused`,
`one_real_option_turns_the_refusal_back_into_a_warning`,
`backend_knobs_at_their_defaults_alone_are_silent`, and
`a_backend_only_options_file_is_refused_by_either_route` (implicit
`ipopt.opt` in the working directory, and explicit `option_file_name=`).
The two pre-existing warning tests now name a real option of their own:
they passed before only on the `print_level=0` the test helper injects,
which is a harness detail rather than a statement about what is under
test.

**Deliberately not pinned:** the backend-knob warning is asserted on its
text and its once-per-family grouping, not on any solver behaviour — there
is no MUMPS or Pardiso here to behave differently. And the two
default-discrepancy options `run_sens` / `sens_max_pdpert` are pinned to
POUNCE's current behaviour, not to a correct one; if the intent is to
adopt upstream's defaults, those tests are what will need to change.

---

- [x] Book page under `docs/src/` updated (`options.md`, `sensitivity.md`;
      both already linked from `SUMMARY.md`)
- [x] Every claim in this PR body and in the code comments is true of the
      code as it stands now

Notes on those last two boxes:

`cargo clippy --workspace --exclude pounce-hsl --all-targets -- -D
clippy::correctness -D clippy::suspicious` (CI's exact gate) initially
**failed** on `line_search_options_wiring.rs` — `approx_constant`, a
correctness lint, on the literal `-1.41421356`. Fixed in 89ef732; the gate
and `cargo fmt` are clean now. `cargo test` is clean for the five crates
this touches (`pounce-algorithm`, `pounce-cli`, `pounce-restoration`,
`pounce-sensitivity`, `pounce-common`); `cargo check --workspace
--all-targets` is clean workspace-wide. Two pre-existing warnings in
`pounce-restoration` and a `presolve_phase6` test are untouched by this PR
and were already on `main`.

No re-sweep was run for the backend-only refusal commit, and it does not
need one: the change is an `or_else` on an existing refusal accessor, no
solver code is touched, and the new refusal cannot fire on any sweep run
because the sweep sets no backend knobs. The byte-identical sweep result
above stands.

On the last box specifically: the `neg_curv_test_tol` table in
`docs/src/options.md` was re-measured against the final diff rather than
carried over. The earlier numbers were wrong — 10 models rather than 11,
`eigena2` as 26 → 282 rather than 26 → 421, `eigenb2` missing entirely, and
no mention of the two models that reach a different objective. The table in
the tree is the re-measured one.


jkitchin:
Reviewed at `ca1facab`. Approving. The claims I could check independently all hold, and the two I most expected to find wrong — the upstream fidelity of the two new *behavioural* branches — are exact.

**The sweep, re-run rather than taken on trust.** Built the branch and swept all 118 fixture-legs against a binary built at the merge base `ac18ba6d`, both the `exact` and `lbfgs` legs: byte-identical, `diff` clean. That is the claim this PR most needed to be true — 397 lines land in `application.rs` and a new branch lands in the middle of `PdFullSpaceSolver::solve_once` — and it is.

It holds for the reason the body gives, which I also checked directly: no default moved. The one place a default could have drifted without anyone noticing is `resto_inner_solver`, which now pushes `inner_alg_builder.mu.tau_min` into both `MonotoneMuUpdate` and `AdaptiveMuUpdate` where it previously left each at its own constructor value. All three are 0.99 (`monotone.rs:68`, `adaptive.rs:207`, `alg_builder.rs:635`), so the assignment is a no-op today — but it is a no-op that would start moving trajectories the moment any one of the three is edited, and nothing pins the three together. Worth a line in `MuOptions::tau_min` saying so; not worth holding this PR for.

**`accept_after_max_steps` matches upstream exactly, including the boundary.** I went to `IpBacktrackingLineSearch.cpp:759` rather than reasoning from the doc comment, because a 0-based `trial` compared with `>=` is precisely where an off-by-one hides:

```cpp
if( accept_every_trial_step_ || (accept_after_max_steps_ != -1 && n_steps >= accept_after_max_steps_) )
```

`n_steps` is 0 on the first trial, and the comparison is `>=`. `trial >= self.accept_after_max_steps` on a `for trial in 0..max_trials` is the same predicate, so `accept_after_max_steps=0` accepting the first trial — which `accept_after_max_steps_changes_the_trajectory` pins — is upstream's semantics and not a convenient reading of them. `>= 0` vs upstream's `!= -1` is equivalent given the registry's lower bound of -1. `Reset()` is `in_soft_resto_phase_ = false; soft_resto_counter_ = 0; acceptor_->Reset();` (cpp:910-917) — the three lines the branch performs, in that order.

**The curvature test is a faithful port, and its one divergence is an improvement.** `curvature_measure` reproduces `IpPDFullSpaceSolver.cpp:600-621` operation for operation — `MakeNew`/`MultVector`/`Dot`, then `Copy`/`ElementWiseMultiply`/`Dot`, then `MakeNewCopy` for the slack block, with the `neg_curv_test_reg_` term in the same place — so the comment's claim of bit-comparability rather than mere algebraic equality is earned. The `neg_values != numberOfEVals` gate, the `xWx < tol * xs_nrmsq` test, and the `PerturbForWrongInertia` → `SYMSOLVER_WRONG_INERTIA` fall-through all match.

The divergence: upstream asserts `DBG_ASSERT(augSysSolver_->ProvidesInertia())` — which is nothing at all in a release build. This branch instead keeps the inertia check when the backend provides no inertia (`neg_curv_test_tol <= 0.0 || !self.aug_solver.provides_inertia()`) and guards the heuristic the same way, so a backend without an inertia ends up with the check rather than with neither. That is the right call and the comment says why.

**The backend-only refusal.** The gate design is the part I'd have argued with, and the code argues with itself first: presence-not-value for "mentions something real", value-not-presence for "asked for something", and `DELIVERY_MECHANISM` exempting `option_file_name` so the most common way to supply a backend-only file isn't the one case that can never fire. `run_bare` existing at all — a second helper whose entire purpose is to *not* inject `print_level=0`, because injecting it would make every case look mixed and the refusal would never fire — is the kind of thing that usually gets discovered a release later. Both pre-existing warning tests now carry a real option of their own instead of leaning on the helper.

This is a new way for a previously-working run to exit 2, and it is the only user-visible behaviour change here that isn't gated behind setting a newly-live option. It's justified, it's in the CHANGELOG in the user's terms, and the four-row table in the body is the right way to state it.

**Also checked:** `known_debt` is genuinely `BTreeSet::new()`, and the test refuses to let it be re-extended. `cargo fmt --all -- --check` clean. CI's exact clippy gate (`--workspace --exclude pounce-hsl --all-targets -D clippy::correctness -D clippy::suspicious`) clean; the one `unused import` warning in `pounce-restoration/src/aug_resto_system_solver.rs:859` is present on `main` at the merge base too, so it is not this PR's. All 11 checks green.

Merging.

Issue #693, opened 2026-08-19
Remove the least-squares multiplier perturbation from the initializer path, and re-derive the gh#540 pivot floor with it


jkitchin:
Follow-up to #688, which fixed the `recalc_y` half. This is the other half, and the two changes it needs are coupled.

## What #688 left on the table

`LeastSquareMults::calculate_y_eq` perturbs the `(3,3)`/`(4,4)` blocks of its W=0 augmented system by `1e-8` where `IpLeastSquareMults.cpp:80-81` passes `0`. Eliminating `w` gives

```
y = −(J Jᵀ + δ I)⁻¹ J · rhs_x
```

so `δ>0` returns a Tikhonov-regularized multiplier, damped by `O(δ / σ_min(J)²)`, rather than the least-squares one.

#688 removed it from the `recalc_y` caller only. Three callers keep it — the dual initializer (`init/default.rs`), the warm-start initializer (`init/warm_start.rs`) and the restoration inner solve (`min_c_1nrm.rs`) — on the grounds that they compute `y` **once**, where a bias is transient and corrected by the following Newton steps, while a *failed* solve costs a fallback to `y = 0` and an iteration-0 `inf_du` blow-up. On the `recalc_y` path both considerations invert.

That reasoning is sound but it is not the whole story, because removing the perturbation everywhere is measurably *better* on most of the corpus.

## What the complete change buys

Measured on #680's branch with all four call sites unregularized, against the same branch with only `recalc_y` unregularized. Two-leg fixture sweep:

| fixture | leg | scoped (today) | unregularized everywhere |
|---|---|---|---|
| `eigenb2` | exact | 67 it | **21 it** |
| `pooling_rt2stp` | exact | 298 it | **181 it**, same objective |
| `cresc4` | exact | 81 it | **68 it** |
| `deb7` | exact | 154 it | **143 it**, same objective |
| `airport` | lbfgs | 59 it | 57 it |
| `cresc4` | lbfgs | 106 it | 93 it |
| `eigenb2` | lbfgs | 68 it | 55 it |

It also **fixes #524 at its cause.** `cresc4`'s false `Infeasible_Problem_Detected` is this same multiplier bias reached through the initializer: with the perturbation gone, `infeasibility_mu_strategy_retry=no` goes from `Infeasible_Problem_Detected` to `Solve_Succeeded` in 68 iterations, and it solves at plain defaults in the same 68 — no rung carrying it. Ipopt solves `cresc4` in 71.

[...]

Issue #641, opened 2026-08-17
[adversary] qp-active-set: success=False / optimal_inaccurate at the exact optimum — kkt_error is unnormalized (active-set analogue of #336)


jkitchin:
adversary-fingerprint: qp-active-set/absolute-kkt-false-nonconvergence

> Filed by the nightly adversary job. Correctness is intact — this is a
> status-labeling (false-failure) defect on the parametric active-set QP engine.

## Summary

`pounce.solve_qp(..., method="active-set")` returns `success=False` (status
`optimal_inaccurate`, or `numerical_failure` at a tight `tol`) at points that are
**exactly optimal**. The engine's `kkt_error` is the raw, unnormalized
`max(primal_infeasibility, dual_infeasibility, complementarity)`
(`crates/pounce-convex/src/qp.rs:499`) and is compared directly against `tol`.
The complementarity term scales with the magnitude of the problem data, so on a
QP whose data is merely *large* (not ill-conditioned — the minimal case below has
`cond(P) = 1`) the absolute floor can never reach the absolute tolerance, and a
machine-precision-exact solution is reported as not converged.

This is the **active-set-engine analogue of #336**, which reported the identical
mechanism ("raw (unnormalized) complementarity gap as `kkt_error`") for the
non-symmetric HSDE conic driver. #336 was fixed by #337, which made the
*post-loop adjudication of the HSDE driver* scale-relative. The `method="active-set"`
path (`pounce-qp`) was not covered by that fix and still uses the absolute test.
(Siblings in the same family: #324 (QP), #327 (NLP), #329, #532.)

The POUNCE convex IPM on the identical problem returns a clean `optimal` — the
two engines disagree on the same instance, and the one reporting failure is the
one with the *more* accurate answer.

## Reproduction

Self-contained, ~12 lines, deterministic (reproduced 3/3 runs):

```python
import numpy as np, pounce
# min 0.5*K*||x||^2 - K*(x1+x2)   s.t.   x1 + x2 <= 1
# Closed form: constraint active, symmetric  ->  x* = [0.5, 0.5], obj* = -0.75*K
K = 1e9
P = K*np.eye(2); c = -K*np.ones(2)
G = np.array([[1.0, 1.0]]); h = np.array([1.0])

r = pounce.solve_qp(P=P, c=c, G=G, h=h, method="active-set")
x = np.asarray(r.x, float)
print(r.status, r.success, r.kkt_error)          # optimal_inaccurate False 1.11e-07
[...]

PR #719, opened 2026-08-20
Refuse the gh#293 retry's uncertified Optimal, and correct the second gap normalizer (#712)

### LINKED ISSUE #712: `scaled_feasible_a` returns `SolveSucceeded` at an absolute KKT error of 2.3e3: the gh#293 retry's `Optimal` is certified by a gap normalizer #696 never corrected
Found while re-measuring #690 on `ac18ba6`. **Pre-existing and post-#696** — #696 narrowed this fixture's certified point from KKT 2.5e2 to 4.6e-3 but did not close the hole, because it corrected one of the two places that normalize by the objective magnitude.

## What happens

`crates/pounce-cli/tests/fixtures/scaled_feasible_a.nl`, default options (`tol = 1e-8`):

```
pounce crates/pounce-cli/tests/fixtures/scaled_feasible_a.nl /dev/null --json-output /tmp/a.json
```

Reports `SolveSucceeded`, `iters=123`, `obj=0`, `final_kkt_error=4.57e-03`. The point actually returned by the retry that produces this verdict has an **absolute KKT error of 2.283e+03** in the user's coordinates.

The true optimum is 0 and is reachable by this same solver: given a large enough budget it converges at 3596 iterations with `final_kkt_error` 1.9e-10.

## The symptom that exposed it: `max_iter` changes the answer

| `max_iter` | status | iters | `final_kkt_error` |
|---|---|---|---|
| 200 (default) … **2610** | `SolveSucceeded` | 123 | **4.6e-03** |
| **2611** … 3595 | `SolveSucceeded` | cap−1 | 5.2e-11 |
| ≥ 3596 | `SolveSucceeded` | 3596 | 1.9e-10 |

Deterministic across repeats, sharp at 2610/2611. A budget cap must not change the accuracy of a solve that reports success. Six other fixtures (`scaled_feasible_b`, `feasible_x0_wide_scale`, `feasible_x0_extreme_row`, `lp_afiro`, `bound_active_qp`, `tame`) are budget-invariant, so this is specific to this model.

## Root cause

1. `scaled_feasible_a` minimizes `Σ(xᵢ − aᵢ)²` with `a` feasible, so its objective carries a degree-0 constant `Σaᵢ² ≈ 5e11` and its true optimum is 0.
2. The cold HSDE solve exhausts `max_iter` and returns `IterationLimit`.
3. `solve_qp_ipm_core` (`ipm.rs:462`) fires the gh#293 Ruiz-equilibrated HSDE retry. It returns `Optimal` at iteration 123.
4. `accept` is `retry.status == Optimal`, so the retry is returned at `ipm.rs:474`. This early return also **bypasses `verify_or_repair_optimum`** (the gh#414 guard, `ipm.rs:482`) — though see 5: the guard would not have caught it either.
5. `optimum_is_genuine` passes the point on its second arm:

   ```
   abs_kkt = 2.283e+03          (fails the first arm, 2.283e3 > tol)
   equilibrated_kkt_rel = 4.566e-09   ≤ FALSE_OPTIMUM_REL_TOL = 1e-3   ← passes here
       components:  pr = 0.000e+00   du = 1.000e-15   gap = 4.566e-09
   ```

   The damage is entirely in the **gap** component. `2.283e3 / 4.566e-9 ≈ 5.0e11` — the gap is being normalized by `Σaᵢ²`, the degree-0 constant `QpProblem` does not model.

6. At `max_iter ≥ 2611` the first solve reaches a true KKT error below `tol` on its own, so step 2 never produces `IterationLimit`, no retry fires, and the good point survives. That is the whole of the budget sensitivity — a side effect, not a separate defect.

## Why #696 did not cover it

#696 introduced `QpOptions::obj_constant` and threaded it into HSDE's stopping test, where `scale_g` normalizes the duality gap. `equilibrated_kkt_rel_parts` normalizes *its* gap the same way and never received the correction. The #689 dev-note already identified this second site:

> The `#414` guard (`verify_or_repair_optimum`) does not catch it for the same reason: its `cscale` is the objective magnitude, so it reads `4.9e-10` on a point whose absolute KKT error is `2.5e2`.

At the time that read as "the guard fails to catch a bad point". Post-#696 it is stronger than that: the guard is what **certifies** the gh#293 retry's answer, so the same normalizer is now load-bearing for a success verdict rather than merely silent.

## Suggested direction

Feed `obj_constant` into `equilibrated_kkt_rel_parts`' gap normalizer, the same correction #696 made to `scale_g`. `FALSE_OPTIMUM_REL_TOL = 1e-3` has documented margin on both sides (false optima measured at `2e-2‥1.2e2`, genuine at `1e-12‥1e-9`, the #286 huge-magnitude optima at `4e-10` and `1.5e-8`), so the cut itself likely does not need to move — but the #286 family is exactly what a change here can break, and it is the reason the constant rather than a tighter cut is the right lever.

Two things to settle with it:

* Whether the gh#293 accept at `ipm.rs:474` should run the genuineness check before returning, independently of the normalizer fix. It is the one `Optimal` in `solve_qp_ipm_core` that reaches a caller without one.
* This is a verdict- and trajectory-affecting change, so `scripts/sweep-fixtures.sh` both legs before merge.

## Relation to #690

The adaptive-τ study in #690 is blocked on this same fixture, which was the only status change in either variant. The baseline it is measured against is this early-stopped point, so #690 should be re-run once this lands.

Refs #690
Refs #689
Refs #414
Refs #293
Refs #324

jkitchin:
## Problem

`crates/pounce-cli/tests/fixtures/scaled_feasible_a.nl` at default options
reported `SolveSucceeded` on a point it never reached. The model is
`min Σ(xᵢ − aᵢ)²` with `a` feasible and three constraints exactly active there,
so `x* = a` and the optimum is `0` — checkable by hand from the `.nl`, and what
the NLP route returns.

| | status | objective | `final_kkt_error` | true abs KKT | iters |
|---|---|---|---|---|---|
| before | `SolveSucceeded` | 0 | 4.566e-03 | **2.283e+03** | 123 |
| after | `MaximumIterationsExceeded` | 0 | 2.967e-07 | 2.967e-07 | 199 |
| after, `max_iter=4000` | `SolveSucceeded` | 0 | 1.883e-10 | 1.883e-10 | 3596 |
| oracle | — | 0 | — | 0 | — |

Two columns need reading together. `final_kkt_error` is the NLP-level report and
carries stationarity and feasibility; the returned point's *full* KKT error,
complementarity included, is `2.283e+03`, and complementarity is where all of
the damage was. That is why nothing downstream saw it. The objective column
discriminates nothing here — the user's objective is a near-total cancellation
against the `5e11` constant, so it reads `0` on a point that is `2.3e-3` per
variable away from the optimum.

The symptom that exposed it is the sharper statement: **`max_iter` changed the
answer while the status stayed the same.** Below the budget at which the cold
solve happens to converge, `SolveSucceeded` at 123 iterations; above it,
`SolveSucceeded` at a genuine `5e-11`. Twelve orders of accuracy, same verdict,
same options — which is precisely what a user cannot catch.

## Cause

The convex QP driver models `½xᵀPx + cᵀx` only, so on a least-squares objective
the degree-0 term `Σaᵢ² ≈ 5e11` is a displacement the solver is not carrying.
#696 told the in-loop stopping test about it (`QpOptions::obj_constant` →
HSDE's `scale_g`). It did not reach the **second** place that magnitude
normalizes a duality gap: `equilibrated_kkt_rel_parts`' `cscale`, which is what
`optimum_is_genuine` — the gh #414 guard — divides the complementarity residual
by. The #689 dev-note had already named this site; at the time it read as "the
guard fails to catch a bad point". After #696 it is stronger than that, because
the guard became what *certifies* an answer rather than merely failing to refuse
one.

Concretely, on the retry point: `2.283e3 / 5.0e11 = 4.566e-9`, comfortably
inside `FALSE_OPTIMUM_REL_TOL = 1e-3`.

And the verdict came from a path that never asked. The cold HSDE solve exhausts
`max_iter` and returns `IterationLimit`; the gh #293 Ruiz-equilibrated retry
fires and returns `Optimal`; `solve_qp_ipm_core` returned it **on status
alone** — the one `Optimal` in that function reaching a caller without a
genuineness check, on an early return that also bypasses the #414 guard
entirely.

## Fix

Three changes. None of them fixes it alone.

**1. `cscale` is told the objective constant.** The same correction #696 made to
`scale_g`, applied to the site it left, mapped into the equilibrated metric by
the equilibration's own `σ`.

**2. The gh #293 retry has to earn its `Optimal`.** It now passes through
`optimum_is_genuine`; a retry that cannot certify leaves the original status
standing, which is the honest answer — the loop really did run out of
iterations. A false `Optimal` is refused on the `NumericalFailure` arm too: it
is worse than an honest breakdown. Without this, change 1 is inert on this
model, because the early return never consults the corrected test.

**3. A complementarity product only counts where its slack is resolvable.**

This is not a companion correctness improvement — **it is what keeps change 1
from regressing a fixture.** Measured as a control (the shared `obj_constant`
term alone, numerator filter neutralized):

| | `scaled_feasible_a` | `feasible_x0_wide_scale` |
|---|---|---|
| main | `SolveSucceeded` 123 | `SolveSucceeded` 80 |
| control (constant only) | `MaximumIterationsExceeded` 199 ✅ | `SolveSucceeded` **199** ❌ |
| + numerator filter | `MaximumIterationsExceeded` 199 ✅ | `SolveSucceeded` **80** ✅ |

Restoring the constant on its own fixes the target model and pays for it with a
2.5× trajectory regression on `feasible_x0_wide_scale` — still `SolveSucceeded`,
still the right objective, just 199 iterations against a 200 cap. That is the
gh #544 signature exactly: the right answer, slowly, invisible to a suite that
asserts status and objective.

Why the constant alone does that: it does not merely re-centre the objective on
a least-squares model *at its optimum*, it cancels it — `½xᵀPx + cᵀx` and the
constant are equal and opposite, their sum is `~0`, the `max(1.0)` floor
governs, and the relative arm silently becomes an absolute one.

The two points are structurally identical: a variable pinned inside a box far
tighter than its own magnitude, with large multipliers on both sides. Measured
in the equilibrated metric, what separates them is only whether the slack is a
number double precision can hold:

| model | box width | \|x\| | slack | quanta | multiplier | product | wanted |
|---|---|---|---|---|---|---|---|
| `feasible_x0_wide_scale` | 1.4e-8 | 7.1e5 | 7e-9 | **46** | 1.8e7 | 0.13 | accept |
| `scaled_feasible_a` | 1e-5 | 3.8e4 | 5e-6 | **5.9e5** | 4.6e8 | 2.3e3 | reject |

("quanta" = the slack in units of `ε·max(|x|,|bound|)`, the quantum of the
subtraction that produced it.)

So `equilibrated_kkt_rel_parts` counts a product only where
`|s| > 64·ε·max(|a|,|b|)` over the two quantities `s` is the difference of. That
is the same reading of "numerically zero", and the same constant, the NLP-side
primal residual has used since #446 / #528 (`ROW_NOISE_KAPPA`,
`primal_noise_floor_kappa`). Only the *relative* measure abstains — the absolute
residual `optimum_is_genuine` consults first is untouched.

**`FALSE_OPTIMUM_REL_TOL` did not move**, and could not have: once the
normalizer is honest the two points read `2.3e3` and `0.13`, and no cut between
them survives the #414 and #286 families the constant is calibrated against.
That is why the fix is a noise floor on the numerator rather than a retuned cut.

Also rejected, earlier in the search: a *global* normalizer keyed on the
complementarity terms' own magnitude (`max_i termscaleᵢ·|zᵢ|`). It reads
`1.3e-10` on `scaled_feasible_a` — the max pairs a large row slack with a
different row's multiplier, so it accepts the very point this exists to reject.
The pairing has to be per-index, and per-index the only thing left to say about
a pair is whether its slack is real.

## Blast radius

`scripts/sweep-fixtures.sh`, both legs — **one line moves, and it is the
targeted one:**

```
exact  scaled_feasible_a   SolveSucceeded it=123 obj=0  ->  MaximumIterationsExceeded it=199 obj=0
lbfgs  scaled_feasible_a   SolveSucceeded it=123 obj=0  ->  MaximumIterationsExceeded it=199 obj=0
```

Everything else is bit-identical in status, objective and iteration count.
`feasible_x0_wide_scale` in particular stays at 80 iterations, as do
`scaled_feasible_b` (47) and `feasible_x0_extreme_row` (33). Re-verified on
current `main` (`9b713dd`) after the 11 intervening commits, including #693's
least-squares multiplier-initializer change and the FERAL 0.17.0 bump: same one
line, and the pre-state is unchanged — main still certifies `scaled_feasible_a`
at 123 iterations on a point whose absolute KKT error is `2.283e+03`.

That one line is the fix, not a regression: the model provably needs ~3596
iterations to reach a genuine optimum (`final_kkt_error 1.9e-10` there), so no
honest verdict inside a 200-iteration budget is a success. It still reports
objective `0`, and `max_iter=4000` reaches the optimum with a success code.

**A user-visible verdict change, called out deliberately.** A model that shipped
as `SolveSucceeded` at defaults now reports `MaximumIterationsExceeded`. Whether
`scaled_feasible_a` should carry a success verdict at all — i.e. whether the
default budget is the right one for this class — is the separate question #712
raises, and this PR does not decide it.

**Reach beyond the fixtures.** `equilibrated_kkt_rel_parts` is `pub(crate)`,
with two callers — `optimum_is_genuine` (the #414 guard) and the active-set
driver's `adjudicated_kkt_error` (#641). The two halves of the change reach them
differently, and the difference is what decides whether a future change here
needs a sweep:

* The `+ obj_constant·σ` term **is** inert for a caller that does not set it.
  `obj_constant` defaults to `0.0` and only the CLI's QP and SOCP routes assign
  it, from the `.nl` degree-0 term they already track, so every library caller
  reads the same `cscale` as before, bit for bit.
* The numerator filter is **not**. `resolvable_complementarity` replaces the
  complementarity residual *unconditionally* — it is not gated on a non-zero
  `obj_constant` — so it applies to every point that reaches this measure, on
  every route. That is deliberate: "this slack is under the quantum of the
  subtraction that produced it" is a statement about the arithmetic, not about
  where the caller's zero sits, and gating it on the constant would switch it
  off exactly where the arithmetic is worst. But it is a behaviour change for
  every caller, and the corpus sweep is what stands behind it.

Downstream: **#690 should be re-run.** Its adaptive-τ study is measured against
this fixture's early-stopped point, which this removes.

## Tests

New, in `crates/pounce-cli/tests/issue_689_direct_driver_scaled_feasible.rs`:

* `a_success_verdict_carries_the_same_accuracy_at_every_budget` — the property
  #712 opened on, stated directly: over budgets spanning the historical flip
  (200, 400, 2610, 2611, 3000, 3595, 3596, 4000), *every* budget that reports
  success must hand back a point accurate to `1e-6`. It deliberately does not
  care which budgets converge — the HSDE trajectory oscillates on this model
  (converging at 2610, 2999, 3596 and not at 2699, 3594) and that is its own
  business; what must not vary is what a verdict means.

`the_default_route_reaches_the_same_optimum` in the same file asserted a success
code for `scaled_feasible_a` at defaults, and is updated rather than deleted.
The #689 property it exists for — the reported objective is `0`, not the `236.85`
the old stopping test admitted — is still asserted for **both** fixtures, and the
verdict assertions are now split by what each model's budget can actually reach,
with `max_iter=4000` pinning that `a` does reach the optimum when allowed to.

New, in `crates/pounce-convex/src/ipm.rs` (`objective_constant_metric_tests`):

* `the_objective_constant_reaches_the_equilibrated_gap_normalizer` — the defect
  in one variable: the same point reads genuine at `obj_constant = 0.0` and is
  refused once told the constant.
* `the_correction_still_certifies_a_genuine_optimum` — the same model with the
  spurious multiplier gone.
* `a_slack_under_its_own_rounding_quantum_is_not_a_violation` and
  `a_resolvable_slack_in_an_equally_tight_box_is_counted` — the two geometries
  from the table above, as the measurement rather than the argument. If a future
  change to the floor collapses them onto the same verdict, these fail loudly.
* `the_quantum_scales_with_the_operands_not_the_slack` — the rule itself.

**Do they bite?** Verified directly, by reverting `crates/pounce-convex/src` to
the parent commit and running the suite against the new tests:

```
test direct_driver_solves_the_scaled_feasible_pair ... ok
test a_success_verdict_carries_the_same_accuracy_at_every_budget ... FAILED
test the_default_route_reaches_the_same_optimum ... FAILED

scaled_feasible_a.nl at max_iter=200: reports success (status SolveSucceeded) at
final_kkt_error 4.566208925098181e-3 after 123 iterations.

scaled_feasible_a.nl at defaults: solve_result_num=0 (status SolveSucceeded)
after 123 iterations at final_kkt_error 4.566208925098181e-3.
```

Both fail for the stated reason, and the pre-existing #689 test
(`direct_driver_solves_the_scaled_feasible_pair`) still passes on the parent —
so nothing in this change weakened it. The `pounce-convex` unit tests pin
functions that do not exist on the parent (`resolvable_complementarity`, and
`optimum_is_genuine`'s `obj_constant` parameter), so "fails on parent" is not
literally checkable for them; each states its before/after pair internally
instead — `the_objective_constant_reaches_the_equilibrated_gap_normalizer`
asserts the pre-fix reading (`obj_constant = 0.0` certifies the bad point) as an
explicit premise before asserting the fixed one.

Not pinned, and knowingly: which budgets converge on `scaled_feasible_a`. It
depends on where an oscillating trajectory happens to be when the cap lands, and
pinning it would pin noise.

Post-mortem in `dev-notes/issue-689-direct-driver-cold-start.md` §6, where the
#689 note had already identified this second site. #720's
`issue-712-genuineness-normalizer.md` calibration table and its unrelated
`issue-690-hsde-tau-tail.md` are carried on this branch rather than lost with
that one.

---

- [x] Tests fail on the parent commit for the stated reason — verified by
      reverting the solver source and running them, output above
- [ ] Book page under `docs/src/` updated — n/a, no user-facing option, API or
      documented behaviour changes; the verdict change is in the CHANGELOG
- [x] `cargo fmt --all -- --check` clean; `cargo clippy` warning count unchanged
      from the parent; `cargo test --release --workspace` green — 268 suites,
      3342 tests, 0 failures, and re-run green on the merge with current `main`
      (`pounce-convex` 222 unit + all integration). `pounce-hsl` is excluded:
      MA57 cannot link without `COINHSL_DIR`, absent in this environment and
      unrelated to this change.
- [x] Every claim in this body and in the code comments is true of the code as
      it stands now

Fixes #712

Refs #690
Refs #696
Refs #689
Refs #544
Refs #414
Refs #293


jkitchin:
CI green — all 11 checks pass on `8c09324`, `mergeable_state: clean`.

One note on the sweep baseline, since `main` moved between measuring it and opening this: the branch point is `b1bb3c3` and `main` is now `dd699e2`. The only intervening commit is `ac94848` (#711 follow-up), which adds two `#[test]` functions to `crates/pounce-nlp/src/quadratic.rs` and no production code. So the corpus sweep quoted in **Blast radius** is still measured against a numerically identical baseline — nothing in the solver moved under it, and it does not need re-running.

---

jkitchin:
## Review: this is the one to merge, with one correction to the body

Reviewed against #720, which fixes the same issue (#712) with a different
remedy. The two are mutually exclusive.

### What I verified

Both PRs' fixture sweeps were stale — #719's was taken at `b1bb3c3b` (11
commits behind), #720's at `dd699e24` (9 behind), both before #693's
least-squares multiplier-initializer change landed. I rebuilt a baseline on
current `main` (`9b713ddd`) and re-swept both PRs merged onto it, 120
fixture-legs each, both legs.

The claim holds on current main. Exactly one line moves, and it is the
targeted one:

```
exact  scaled_feasible_a   SolveSucceeded            it=123  ->  MaximumIterationsExceeded  it=199
lbfgs  scaled_feasible_a   SolveSucceeded            it=123  ->  MaximumIterationsExceeded  it=199
```

The pre-state is also unchanged by the 11 intervening commits — current main
still certifies `scaled_feasible_a` at 123 iterations on a point whose
absolute KKT error is 2.283e+03, so the defect is still live and still
exactly as described.

Also worth recording: **#719 and #720 produce byte-identical 120-leg
corpora.** The fixture corpus cannot distinguish these two fixes. The choice
between them has to be made on what happens *off* the corpus.

### The control experiment — the second half is load-bearing

Both PRs share the `+ obj_constant * sigma()` term. I built a control that
keeps only that shared term and neutralizes #719's numerator filter, to see
what each PR's second half actually buys:

```
main -> control (obj_constant term only):
  exact  scaled_feasible_a        SolveSucceeded 123  ->  MaximumIterationsExceeded 199   <- the fix
  exact  feasible_x0_wide_scale   SolveSucceeded  80  ->  SolveSucceeded            199   <- the cost
  lbfgs  (same two lines)

control -> #719 (what the numerator filter adds):
  exact  feasible_x0_wide_scale   SolveSucceeded 199  ->  SolveSucceeded             80
  lbfgs  feasible_x0_wide_scale   SolveSucceeded 199  ->  SolveSucceeded             80
```

So restoring the constant, on its own, fixes `scaled_feasible_a` and pays for
it with a 2.5x trajectory regression on `feasible_x0_wide_scale` — still
`SolveSucceeded`, still the right objective, just 199 iterations instead of
80. That is the gh#544 signature exactly: right answer, slowly, invisible to
a suite that asserts status and objective. The second half of each PR exists
to prevent that false *rejection*, and both remedies buy it back to exactly
80.

This is worth stating in the PR body. As written, the body frames the
numerator filter as a companion correctness improvement; it is actually the
thing that keeps the primary fix from regressing a fixture.

### Why this remedy over #720's

Both are unconditional relaxations of the sole gh#414 false-optimum guard.
They differ in how bounded the relaxation is.

- **#719** drops a slack from the numerator when it is below
  `64 * eps * max(|h_i|, |g_i x|)` — a local, per-index question ("is this
  slack distinguishable from zero at the precision of the operands that
  produced it?"), with a bound tied to the operands. `64 * eps` is not a new
  calibration: it is the same quantum already used by `ROW_NOISE_KAPPA` /
  `primal_noise_floor_kappa` from gh#446 / gh#528.
- **#720** floors the denominator at `sqrt(eps) * max(|quad|, |lin|, |konst|)`.
  That is a global fudge, and in the cancelling regime (`quad ~ -lin`, no
  degree-0 term at all) it raises `cscale` by up to `1/sqrt(eps) ~ 6.7e7`,
  loosening the guard by that factor. #720's own body declares this
  unmeasured, and there is no fixture in that regime.

#720's `dev-notes/issue-712-genuineness-normalizer.md` is nonetheless the
better write-up of the two — its 10-row calibration table over every point
that reaches the normalizer is the artifact this change deserves. If #720 is
closed, that note and its unrelated `issue-690-hsde-tau-tail.md` should be
salvaged rather than lost with the branch.

### One correction needed before merge

The body says the change is inert for existing callers:

> every existing caller passes the `0.0` that leaves them unchanged

That is not accurate, and the control experiment above disproves it.
`resolvable_complementarity` replaces the numerator **unconditionally** — it
is not gated on a non-zero `obj_constant`. `feasible_x0_wide_scale` reaches
this path with no degree-0 constant and moves 199 -> 80 because of it. The
`+ obj_constant * sigma()` half is genuinely inert for library callers
(`obj_constant` is set only on the CLI QP/SOCP routes); the numerator filter
is not. Please reword — this is the one sentence a future reader would rely
on when deciding whether this change needs a sweep.

### Checks

- CI 11/11 green.
- `pounce-convex` suite passes on the merge with current main (8 passed).
- The new `a_success_verdict_carries_the_same_accuracy_at_every_budget`
  passes in 16.5s — the budget loop is cheap, no concern there.
- Merges clean onto `9b713ddd`; GitHub's `CONFLICTING` status is stale.

**Recommendation: merge this one, close #720**, after the body correction.

jkitchin:
Corrected, and the control experiment is now in both the body and the dev-note. Thanks for running it — it changes what the third change *is*, not just how it reads.

**The inertness sentence.** You're right and the claim was wrong. `resolvable_complementarity` replaces the numerator unconditionally; it is not gated on a non-zero `obj_constant`. The body now splits the two halves explicitly — the `+ obj_constant·σ` term is inert for a caller that does not set it, the numerator filter is not, and the paragraph says why the filter is deliberately ungated (the quantum is a fact about the arithmetic, not about where the caller's zero sits, so gating it on the constant would switch it off exactly where the arithmetic is worst).

One correction back, on the supporting example only — the conclusion is unaffected. `feasible_x0_wide_scale` does carry a degree-0 constant: its objective is `(x₀+0.5)² + (x₁−499998.3031417761)²`, so the constant is `0.25 + 499998.3…²`, and instrumenting the call site measured `obj_constant = 2.499983e+11` reaching `equilibrated_kkt_rel_parts` on that model. So that fixture isn't the disproof. The disproof is the code-level one, which needs no fixture: the call is unconditional, so a library caller passing `0.0` gets a different numerator than before wherever a slack falls under its quantum. I've put it that way rather than reproducing the fixture claim, since this is the sentence you flagged as the one a future reader would lean on.

**Also pushed** (`b975f33`): the same two facts into `dev-notes/issue-689-direct-driver-cold-start.md` §6 — the scope split, and your control table. The body is the argument for this PR; the note is where someone lands two years from now asking whether a change here needs a sweep, and "the constant alone costs `feasible_x0_wide_scale` 80 → 199" is exactly the gh #544 shape that deserves to be written down rather than re-derived.

**On the stale sweeps.** Confirmed on my side after taking your merge of current `main`: rebuilt and re-ran, `scaled_feasible_a` → `MaximumIterationsExceeded` 199, `feasible_x0_wide_scale` → 80, `scaled_feasible_b` → 47, `feasible_x0_extreme_row` → 33; `pounce-convex` 222 unit tests + all integration green, and the three tests in `issue_689_direct_driver_scaled_feasible` green. I'd started my own merge of `main` in parallel and dropped it when I saw yours had landed — nothing of yours was overwritten, and `ce71df7` with the salvaged notes is intact on the branch.

`git merge-tree --write-tree` against `9b713dd` returns a tree cleanly, so the `CONFLICTING` status was stale as you said; the merge you pushed should have cleared it and re-run CI against current main.

---

Issue #286, opened 2026-07-22
[Adversary] SOLVER_LIMITATION: convex QP IPM exhausts default iteration budget at cond(P)=1e10 with O(1e19) objective


jkitchin:
> Discovered by the pounce adversary agent (automated solver correctness testing).

## Family / solver

Convex QP IPM (`solve_qp`), exposed via `solve_qp_batch`.

## Problem

At `cond(P) = 1e10` with an objective of magnitude `O(1e19)`, pounce exhausts its default iteration budget and returns a point violating the box constraint by `0.884`, on a problem that CLARABEL, OSQP, ECOS and SCS all solve at default settings.

Six box-constrained QPs in one `solve_qp_batch` call, item `k` with `cond(P) = 10^(2k)` (1e0 .. 1e10) and objective scale `10^(2.4k)` (spanning ~1e12).

## Expected vs actual

| | objective | box violation |
|---|---|---|
| exact (3^6 active-set enumeration) | `-6.8510532741e+19` | 0 |
| cvxpy CLARABEL / OSQP / ECOS / SCS | `-6.8510532741e+19` | 0 |
| **pounce, item k=5** | `-6.860539e+19` | **`8.84e-01`** |

Statuses: `['optimal' x5, 'iteration_limit']`. Per-item max relative-`x` error vs exact enumeration is `8.84e-01`, **entirely at k=5** — items 0–4 are `2.2e-09`, `2.9e-11`, `1.0e-08`, `9.8e-09`, `3.0e-08`.

pounce's k=5 objective is *lower* than the true optimum only because the returned iterate is outside the box — an unconverged interior-point iterate, not a better solution.

Note: the mean (~0.15) and median (~1e-8) of the per-item errors both **hide** this. Only the per-item max exposes it — worth remembering when writing batch regression tests.

## Why LIMITATION rather than BUG

The status is honestly `iteration_limit`, never `optimal`, so a status-checking caller is not silently misled. And `max_iter=5000` recovers the exact optimum to `5.04e-09` — the iteration is convergent, merely ~1000x slower than the conic oracles at this conditioning.

## Secondary observation worth chasing

At `max_iter=5000` the status is **still** `iteration_limit` even though the iterate is accurate to `5e-9`. That suggests the termination test is measured in unscaled units and becomes unsatisfiable when the objective is `O(1e19)`.

## The batch layer itself is clean

[...]

PR #720, opened 2026-08-20
Stop certifying a non-optimal point on a cancelling objective (#712)

### LINKED ISSUE #712: `scaled_feasible_a` returns `SolveSucceeded` at an absolute KKT error of 2.3e3: the gh#293 retry's `Optimal` is certified by a gap normalizer #696 never corrected

jkitchin:
## Problem

`scaled_feasible_a` returned `Solve_Succeeded` on a point whose absolute KKT error is `2.283e3`, at a `tol` of `1e-8`. It is `min Σ(xᵢ − aᵢ)²` with `a` feasible, so the true optimum is `0` and the objective carries a degree-0 constant `Σaᵢ² ≈ 5e11`.

| | status | objective | abs KKT | iters |
|---|---|---|---|---|
| before | `Solve_Succeeded` | 0 | **2.283e+03** | 123 |
| after | `Maximum_Iterations_Exceeded` | 0 | 1.5e-01 | 199 |
| after, `max_iter=4000` | `Solve_Succeeded` | 0 | 1.9e-10 | 3596 |
| NLP oracle | `Solve_Succeeded` | ~1e-09 | — | — |

Downstream this is a success everywhere — `success=True` from `solve_qp`, `solve_result_num=0` and exit 0 from the AMPL/Pyomo/GAMS drivers — so the wrong point is consumed as an answer.

The symptom that exposed it: **`max_iter` changed the answer.** At ≤2610 the solve returned 123 iterations at KKT 4.6e-03; at ≥2611 it ran on and reached 5.2e-11. Deterministic, sharp at the boundary, and unique to this model — six other fixtures are budget-invariant.

## Cause

Two defects in series, neither sufficient alone.

**1.** `solve_qp_ipm_core`'s gh#293 Ruiz-equilibrated retry — which fires when the first solve exhausts its budget — accepted `Optimal` on the *status alone* and returned it. That early return also bypasses `verify_or_repair_optimum`, so it was the one success verdict in the convex entry point reaching a caller without the gh#414 genuineness check.

**2.** That check would have passed the point anyway. `optimum_is_genuine`'s relative arm normalizes complementarity by `½xᵀPx + cᵀx` — the objective `QpProblem` models, which on a model carrying a degree-0 term is the caller's objective *displaced* by it. A gap of `2.283e3` over `5.0e11` read `4.566e-9` against a `1e-3` cut.

This is the second of the two sites #689 named. #696 corrected the first (HSDE's `scale_g`, via `QpOptions::obj_constant`) and left this one, where post-#696 it is no longer merely failing to catch a bad point — it is what *certifies* the retry's.

The `max_iter` sensitivity is a consequence, not a separate defect: at ≥2611 the first solve gets under `tol` on its own, so the retry never fires.

## Fix

Both sites. The gh#293 accept now earns its `Optimal` the same way the other one does, and a retry that cannot certify leaves the original status standing.

For the normalizer, **adding the constant is not enough, and that is the whole shape of the fix.** On these models `½xᵀPx + cᵀx` and the constant are equal and opposite (`−5.000e11` and `+5.000e11`), so their sum carries no scale, the `max(1.0)` floor takes over, and the relative test silently becomes an absolute one. Measured, that variant rejects `feasible_x0_wide_scale` — whose point matches the NLP oracle to 13 digits (`499998.30314177` vs `…78`) and whose absolute KKT error of `1.3e-1` is `5.3e-13` of its own objective terms — taking it from 80 iterations to 199 against a 200 cap, handing back the margin `cb43358` fixed. It is the same split `cb43358` already had to make inside HSDE: correct the gap's normalizer with the constant, keep the gate on the magnitudes actually being computed.

So the floor is *relative to the terms*:

```rust
cscale = max(|quad + lin + const·σ|,
             OBJ_CANCELLATION_FLOOR * max(|quad|, |lin|, |const·σ|),
             1.0)
```

A sum that retains any of its terms' magnitude is untouched by construction — every gh#414 and gh#286 instance — so **`FALSE_OPTIMUM_REL_TOL` does not move.**

### Two alternatives ruled out by measurement, not argument

Measured over all 13 points that reach the normalizer: the gh#414 false optima and their repairs, the gh#286 huge-magnitude optima, and the four #712 fixtures.

* **No cut on the current normalizer can work.** The genuine `#286` illcond point reads `1.47e-08` and the false `scaled_feasible_a` point reads `4.566e-09` — the good one is *worse* than the bad one. The constant is missing information, not a mis-tuned threshold.
* **A data-scale normalizer rejects the `#286` family outright.** Their objectives are huge because the magnitude lives in `‖x*‖`, not the data: `datascale` is `2e9` where the objective is `3.87e18`.

| point | truth | abs KKT | current | +const | **+const, floored** |
|---|---|---|---|---|---|
| `414` false ×3 | reject | 1.6e-01‥8.3e+03 | 2.0e-02‥1.2e+02 ✓ | same ✓ | same ✓ |
| `414` repaired ×4 | accept | ≤2.1e-04 | 2.9e-10‥1.0e-09 ✓ | same ✓ | same ✓ |
| `286` huge / illcond / mid | accept | 1.4e+03‥1.4e+13 | 3.7e-10‥1.5e-08 ✓ | same ✓ | same ✓ |
| `712` `scaled_feasible_a` | **reject** | 2.28e+03 | 4.57e-09 ✗ | 2.28e+03 ✓ | 1.53e-01 ✓ |
| `712` `feasible_x0_wide_scale` | **accept** | 1.32e-01 | 5.33e-13 ✓ | 1.33e-01 ✗ | 1.79e-05 ✓ |
| `712` `feasible_x0_extreme_row` | accept | 3.81e-05 | 3.81e-20 ✓ | 3.41e-13 ✓ | 2.29e-17 ✓ |

`φ = √ε`. The cancelling models admit any `φ ∈ (2.7e-10, 2.3e-6)`; `√ε` sits 56× above that floor and 153× below its ceiling, and is the classical cancellation threshold. The window is measured, not derived — plain `ε` is below the lower bound and rejects `feasible_x0_wide_scale`. Margins on the cut: tightest reject `2.01e-02` (20× above), loosest accept `1.79e-05` (56× below).

Full calibration and reasoning: `dev-notes/issue-712-genuineness-normalizer.md`.

## Blast radius

Fixture sweep, both legs, against **`dd699e2` (current `main`)** — re-measured after merging `main` in, since the branch was cut at `ac18ba6`:

```
exact  scaled_feasible_a   SolveSucceeded 123it -> MaximumIterationsExceeded 199it
lbfgs  scaled_feasible_a   SolveSucceeded 123it -> MaximumIterationsExceeded 199it
```

**Bit-identical except the targeted case**, on both legs — `feasible_x0_wide_scale` stays at 80, `scaled_feasible_b` at 47, `feasible_x0_extreme_row` at 33.

`scaled_feasible_a` at the default budget is now an honest exhaustion carrying the correct objective. Reaching a certifiable point takes it ~3596 iterations against a cap of 200; it certifies at `max_iter=4000`. That cost was always being paid — this makes it visible rather than skipped.

`obj_constant` defaults to `0.0`, so every library caller that does not set it is unchanged.

**One surface is reasoned rather than measured, and is flagged as such in the code and the note:** the floor is not conditioned on a constant being supplied, since cancellation is a property of the arithmetic — so an objective that cancels on its own (`½xᵀPx ≈ −cᵀx`, both huge, no degree-0 term) is also affected, and there it *raises* `cscale`, relaxing rather than tightening. No fixture is in that regime and neither family exercises it.

## Tests

`issue_689_direct_driver_scaled_feasible::the_default_route_reaches_the_same_optimum` is updated rather than deleted. Its objective assertion — the sharp half, and the one #689 was about — is unchanged on both models. The verdict is now asserted per model, with the budget `_a` actually needs recorded, so a future "certifies at 200 again" has to declare whether it is a speed-up or this regressing.

It bites: on the parent commit it fails with

```
scaled_feasible_a at defaults: solve_result_num=0 (status SolveSucceeded).
Expected an iteration-limit verdict: ... a success code here means the gh #712
false certificate is back
```

which is exactly the defect, verified by building the parent's `ipm.rs`/`active_set.rs` against the new test.

Deliberately **not** pinned: why the model needs ~3596 iterations. That is a convergence-rate question on a least-squares QP whose optimum is a near-total cancellation, it is what #690's adaptive-τ study kept tripping over, and it wants its own measurement.

---

- [ ] Book page under `docs/src/` updated — none applies; this changes no option, no routing and no documented interface, only when an existing verdict is honest
- [x] `cargo fmt --all -- --check` clean; `cargo clippy` clean on the changed crates (no finding on any added line); `pounce-convex` 38 suites, `pounce-cli` 85 suites, and `pounce-algorithm`/`pounce-py`/`pounce-rs` 756 tests all green. `pounce-hsl` does not link MA57 in my container — confirmed identical on an unmodified tree, so not caused by this branch

### Two notes for the reviewer

**This branch also carries #690.** Commit `8d2ccee` is the re-measurement of the adaptive-τ HSDE study, a dev-note only, no solver change. It is what turned up this bug. #690 stays open.

**Commit `0aff181` is a superseded WIP** — it is the normalizer fix before the cancellation floor, and its message says it is not ready to merge. Its content is fully captured by the final commit and the dev-note. Since this repo merges rather than squashes, **squash-merging this PR, or asking me to squash `0aff181` into `9369201`, would keep that message out of `main`'s history.**

Fixes #712

https://claude.ai/code/session_01V34QZ9xPuwyh8vptRtoNp3

---
_Generated by [Claude Code](https://claude.ai/code/session_01V34QZ9xPuwyh8vptRtoNp3)_


jkitchin:
## Review: correct fix, but I recommend #719's remedy instead

This and #719 fix the same issue (#712) and are mutually exclusive. Their
`active_set.rs` changes are byte-identical, and their gh#293 retry fix
(refusing an uncertified `Optimal` from the Ruiz-equilibrated retry) is
byte-identical too. The only substantive difference is the normalizer remedy.

### What I verified

Both PRs' sweeps were stale — this one's baseline is `dd699e24`, nine commits
behind, taken before #693's least-squares multiplier-initializer change
landed. I rebuilt a baseline on current `main` (`9b713ddd`) and re-swept both
PRs merged onto it, 120 fixture-legs each.

The claim holds. Exactly one line moves, and it is the targeted one:

```
exact  scaled_feasible_a   SolveSucceeded 123  ->  MaximumIterationsExceeded 199
lbfgs  scaled_feasible_a   SolveSucceeded 123  ->  MaximumIterationsExceeded 199
```

**This PR and #719 produce byte-identical 120-leg corpora.** The fixture
corpus cannot tell the two fixes apart, so the choice has to rest on what
happens off the corpus.

### Why #719's remedy, on the point this PR already raises about itself

Credit where due: this PR is honest about the tradeoff in a way #719 is not.
It states plainly that the floor is not conditioned on a constant being
supplied, so a self-cancelling objective (`0.5 x'Px ~ -c'x`, no degree-0
term) gets a raised `cscale` and a correspondingly more permissive guard,
with no fixture in that regime. That disclosure is exactly right, and it is
also the reason I'd take the other remedy.

`OBJ_CANCELLATION_FLOOR = sqrt(eps)` loosens the sole gh#414 false-optimum
guard by up to `1/sqrt(eps) ~ 6.7e7` in the cancelling regime, and the size
of that loosening is set by the cancellation magnitude, not by anything tied
to the actual KKT error. #719 instead filters the numerator per index, at
`64 * eps * max(|h_i|, |g_i x|)` — a local question with an operand-tied
bound, using the quantum already established in this codebase by
`ROW_NOISE_KAPPA` / `primal_noise_floor_kappa` (gh#446 / gh#528) rather than
a new constant calibrated to fit a 13-point corpus.

Both remedies are unconditional relaxations of the same guard; #719's is the
bounded one.

### For the record: what the second half is actually for

I built a control keeping only the shared `+ obj_constant * sigma()` term:

```
main -> control:  scaled_feasible_a       SolveSucceeded 123 -> MaxIterExceeded 199   (the fix)
                  feasible_x0_wide_scale  SolveSucceeded  80 -> SolveSucceeded    199   (the cost)
```

So restoring the constant alone fixes #712 and pays for it with a 2.5x
trajectory regression on `feasible_x0_wide_scale` — right answer, slowly, the
gh#544 signature. The floor (here) and the numerator filter (in #719) both
exist to prevent that false rejection, and both buy it back to exactly 80.
Your `dev-notes/issue-712-genuineness-normalizer.md` calibration table
already shows this from the reading side (`feasible_x0_wide_scale` accepted
at 1.79e-05); the sweep confirms it from the trajectory side.

### Scope

Two things sit outside the #712 fix and would come along with a merge:

- `dev-notes/issue-690-hsde-tau-tail.md` (+248) — unrelated to #712, no
  solver change attached.
- `0aff181a`, a superseded WIP commit that the body asks be preserved via
  merge rather than squash.

### Recommendation

Close in favour of #719, but **salvage the dev-notes first** — this PR's
`issue-712-genuineness-normalizer.md` is the better write-up of the two, and
the 10-row calibration table over every point reaching the normalizer is
worth keeping on whichever branch lands. `issue-690-hsde-tau-tail.md` should
not be lost with the branch either.

### Checks

- CI 11/11 green.
- `pounce-convex` suite passes on the merge with current main.
- `issue_689_direct_driver_scaled_feasible` passes (2 tests, 2.2s).
- Merges clean onto `9b713ddd`; GitHub's `CONFLICTING` status is stale.

jkitchin:
Closing in favour of #719, which merged as `8b5d5217` and closed #712.

Both dev-notes were carried over first, so nothing here is lost:

* `dev-notes/issue-690-hsde-tau-tail.md` — unrelated to #712, carried
  unchanged.
* `dev-notes/issue-712-genuineness-normalizer.md` — carried with its
  diagnosis and its calibration table intact. Its `## The rule` section now
  describes **both** candidates: the numerator filter that shipped, and this
  PR's `√ε` denominator floor, recorded under *the rule that did not ship*
  along with the reason. The floor's own disclosure — that it is not
  conditioned on a constant being supplied, so a self-cancelling objective
  gets a *raised* `cscale` and a guard relaxed by up to `1/√ε` — is preserved
  verbatim as that reason.

Two things from this PR that the merged branch would not otherwise have had,
and which are the durable part of the work here:

* **No cut on the current normalizer can work.** The genuine `286` illcond
  point reads `1.47e-8` while the false `scaled_feasible_a` point reads
  `4.57e-9` — the good one is worse than the bad one. The constant is missing
  information, not a mis-tuned threshold. That closes off an entire class of
  "just retune `FALSE_OPTIMUM_REL_TOL`" follow-ups.
* **A data-scale normalizer is not the answer either** — `datascale` is `2e9`
  where the `286` objective is `3.87e18`, so it rejects that whole family.

Also recorded there: the two remedies are **byte-identical on the corpus**
(120 fixture-legs, both legs, common `9b713ddd` baseline). That equivalence is
written down deliberately, so this floor is not re-proposed later as an
untried idea — it was tried, it works, and it was passed over on the bound of
the relaxation it grants rather than on anything measurable in the corpus.

Thanks — the calibration table in particular is the artifact that made the
choice decidable at all.

Issue #713, opened 2026-08-19
pooling_rt2stp on the L-BFGS leg regresses from SolveSucceeded to RestorationFailed under gh#693's undamped y0


jkitchin:
Split out of the gh#693 fix (removing the Tikhonov `δ=1e-8` from the
least-squares equality-multiplier initializer) so the cost is tracked rather
than buried in a commit message. Filed per the repo's own rule: a measured
regression recorded as "an accepted cost of the fix" needs an issue and an
owner, or it is indistinguishable from noise to the next reader.

## The regression

`pooling_rt2stp` on the **L-BFGS** leg of `scripts/sweep-fixtures.sh` is the
one success → failure in the whole 116-leg sweep:

| | status | iters | objective |
|---|---|---|---|
| 0.10.0 (`ac18ba6d`) | `SolveSucceeded` | 233 | `-4391.826001` |
| with gh#693 | `RestorationFailed` | 27 | `9.323117531` |

Reproduce:

```
pounce crates/pounce-cli/tests/fixtures/pooling_rt2stp.nl --no-sol \
       hessian_approximation=limited-memory
```

The exact-Hessian leg of the same model *improves* — 298 → 128 iterations at
the same objective — so this is specific to the L-BFGS trajectory.

## Mechanism, isolated

The entire divergence flows through `y0`. Forcing the initial equality
multipliers to zero makes the two builds bit-identical:

```
                                          status                  iters   objective
main,    constr_mult_init_max=0    ErrorInStepComputation   373    -3273.954615
patched, constr_mult_init_max=0    ErrorInStepComputation   373    -3273.954615
```

Identical to every printed digit. So nothing else in the change reaches this
model — only the value of `y0`.

Sweeping the cap localizes the magnitude, and rules out the obvious
explanation:

```
constr_mult_init_max     main                          patched
1                        ErrorInStepComputation 373    ErrorInStepComputation 373
10                       ErrorInStepComputation 373    ErrorInStepComputation 373
100                      SolveSucceeded 233            RestorationFailed 27
1000 (default)           SolveSucceeded 233            RestorationFailed 27
```

Both builds flip at the same rung, so **both** `‖y0‖∞` land in `(10, 100]`.
This is *not* the cap admitting an estimate that main rejected — it is a
genuinely different `y0` of the same magnitude, and the undamped one starts
the L-BFGS trajectory in a worse basin.

## What is *not* claimed

Neither `y0` is the principled one. At `y0 = 0` both builds fail
[...]

PR #728, opened 2026-08-20
Scale a QCQP by its coefficients, not by one derivative sample (#703)

### LINKED ISSUE #703: Equilibration for badly-scaled QCQPs: salvage dev-note §8 from the cut Q7/Q8
Section 8 of `dev-notes/quadratic-structure-exploitation.md` works out an
equilibration scheme for quadratically-constrained problems and closes with
"ship equilibration on by default for the QCQP driver". The section is then
marked **blocked on Q7/Q8** in its own header, because the driver it names is
Q8's native QCQP barrier.

Q7 and Q8 are being cut (see #588). This issue keeps the part of section 8
that does not actually depend on them.

## What the section establishes

`equilibrate.rs:44-49` is explicit that per-row scaling of `G` preserves
`z >= 0` for the orthant but breaks a second-order cone, whose rows must scale
uniformly. That is a real obstacle **on the conic path**, and it is why
`equilibrate` is wired only into `solve_qp_ipm`.

A quadratic row written natively has no such obstacle. Under `x = D x_hat`
with `D > 0` diagonal and a per-row scale `e_i > 0`, the row becomes a
congruence `D Q_i D`, so `Q_i >= 0` implies `D Q_i D >= 0` -- convexity is
preserved exactly, the barrier is invariant up to an additive constant, and
the central path in `x_hat` is unchanged.

The section also notes that Ruiz per `Q_i` individually is wrong, since each
would demand its own column scaling and there is only one `x`; `D` has to
equilibrate the whole family jointly via magnitude surrogates over
`H(lambda) = Q_0 + sum_i lambda_i Q_i`.

## Why it is worth pursuing without the native driver

The section's motivating measurement is about the *model*, not the solver
path it takes:

> The `qcqp1500-1c` right-hand sides are 1.58e5-1.80e5 while
> `lambda_max(Q_i) ~ 1.6e3` -- a 100x mismatch that directly biases
> `s_i = -g_i(x)` and hence the `-s_i/lambda_i` KKT diagonal.

That mismatch is a property of `qcqp1500-1c.nl`. It is there whichever driver
picks the problem up, and today `qcqp1500-1c` is routed to the **NLP** path
(it is a convex QCQP downgraded by `SOCP_REFORM_FLOP_BUDGET`, n*m = 2.7e10,
over the cap). So the badly-scaled instance the section diagnoses is one the
existing NLP path is solving right now.

#702 is the concrete symptom: `qcqp1500-1c` is unusually sensitive to
ULP-level rounding in the constraint evaluator -- 103 iterations on the AD
tape, 131 on Q4's coefficient matrices, same answer to ten digits. A
trajectory that fragile is exactly what a 100x row/curvature mismatch
predicts, and it is the only member of its family that behaves this way.

## Suggested work

1. Measure the actual row-scale spread on the six downgraded convex QCQPs
   (`nql180`, `qssp180`, `qcqp500-3c`, `qcqp750-2c`, `qcqp1000-2c`,
   `qcqp1500-1c`) and confirm `qcqp1500-1c` is the outlier the section says
   it is.
2. Check what the NLP path's existing scaling already does to these rows
   before designing anything new -- the section was written about a driver
   with no scaling at all, and that is not the situation on the NLP path.
3. Only if 1 and 2 leave a real gap, implement the joint-`D` scheme from
   section 8. The two-block magnitude surrogate and the `e_i` second stage
   are written out there in full and do not need redesigning.

A result of "the NLP path's existing scaling already handles this" is a fine
outcome and should be recorded in the note, since section 8 currently implies
otherwise.

## Not blocked

Nothing above needs `QcqpProblem` extraction (Q7) or a native QCQP barrier
(Q8). Section 8's header should stop saying it is blocked on them; that
should be corrected in the note either way.

jkitchin:
Closes #703.

> **Read this first.** #703 asked for four things. Three are done here.
> The fourth — *"measure the row-scale spread on the six downgraded convex
> QCQPs and confirm `qcqp1500-1c` is the outlier"* — **has not been done**:
> those instances live in `$POUNCE_BENCH_DATA`, and this environment has
> neither that directory nor egress to `plato.asu.edu` (the proxy returns a
> policy 403). This PR ships the tool that takes that measurement and
> deliberately leaves the new scaling method **off by default** because of
> it. If you'd rather #703 stayed open for that, drop the closing keyword —
> §8.7 of the dev-note is written as the standing description of what is
> left. A second, unrelated defect found along the way (below) needs its own
> issue either way.

## Problem

#703's question 2 was "check what the NLP path's existing scaling already
does to these rows before designing anything new". The answer is **nothing**,
and not because a cutoff is mistuned.

`nlp_scaling_method=gradient-based` (the default) is a point sample: it reads
`∇f` and the Jacobian once at x₀ and assigns row `i` the factor
`min(1, nlp_scaling_max_gradient / ‖∇gᵢ(x₀)‖_∞)`. A row `½xᵀQx ≤ b` written
about the origin has `∇g(0) = 0`. Started from `x₀ = 0` — which is how AMPL
emits `qcqp1000-2c`: every variable free, no initial guess — the row maximum
arrives as the `DBL_MIN` seed, the ratio overflows the clamp, and the row
comes out at **exactly 1.0** however far `Q` and `b` disagree. No cutoff
reaches it: `100/0` and `1e-6/0` both clamp to 1. The cutoff is also a
*per-block* gate, so a block with no row above it gets no scale vector at
all.

Measured across the CLI fixture corpus: **196 of 196 quadratic rows are left
unscaled**, `csfi2`'s at 250× right-hand-side-to-curvature mismatch and one of
`airport`'s at 52× among them.

There is also no column stage anywhere on this path. Given one convex QCQP
and the same QCQP after an exact change of variables `x_j → x_j / c_j`:

| column span | `gradient-based` | `curvature-based` (this PR) |
|---|---|---|
| 1 | 75 it, `2.4779690299303e4` | 16 it, `2.4779690299303e4` |
| 1e3 | 92 it, `2.4779690302194e4` | 16 it, `2.4779690299303e4` |
| 1e6 | 154 it, `2.4779690388034e4` (KKT 3.3e-7) | 16 it, `2.4779690299303e4` |
| 1e9 | **`Maximum_Iterations_Exceeded`** at 3000, objective `4.23e2` | 16 it, `2.4779690299303e4` |

Same problem in every row of that table. The oracle is its own first line.

## Cause

§8 of `dev-notes/quadratic-structure-exploitation.md` diagnosed this and was
marked *blocked on Q7/Q8*, which were cut. Two things in it were stale rather
than blocked:

* it says to extend `equilibrate()`, which belongs to `solve_qp_ipm` and is
  never called on the NLP path;
* it assumed §8's first stage (a joint variable scaling) was unreachable
  because `OrigIpoptNlp` refuses `x_scaling`. **Since #486 that refusal is
  gone** — `install_variable_scaling` wraps the model in `ScalingTnlp` and
  substitutes `x̃ = d ⊙ x` one level below the algorithm. Both of §8's stages
  were expressible through machinery already present.

## Fix

`nlp_scaling_method=curvature-based` (`crates/pounce-nl/src/nl_scaling.rs`),
derived from the model's coefficients rather than from a derivative sample:

1. **Joint `D`.** Build §8's two λ-independent magnitude surrogates over the
   pencil `Q₀ + Σλᵢ Qᵢ` — the ∞-norm envelope of every `H(λ)` that can arise —
   and Ruiz-sweep them as the augmented `K̂ = [[P̂, Ĵᵀ], [Ĵ, 0]]`, with the same
   `RUIZ_SWEEPS = 10` and `[1e-8, 1e8]` clamp `pounce_convex::equilibrate`
   uses. Ruiz *per `Qᵢ`* would be wrong: each would demand its own column
   scaling and there is only one `x`.
2. **Row scale.** `eᵢ = 1/max(‖D Qᵢ D‖_∞, ‖D aᵢ‖_∞, |bᵢ|)`.

Both are returned from `TNLP::get_scaling_parameters`, the callback user
factors already arrive on, so no new solver machinery exists. A
`scaling_factor` suffix the model carries overrides the computed factor
component by component.

Three decisions §8 does not cover, and why:

* **Every row enters `Ĵ`, not only the quadratic ones.** §8 was written for a
  driver whose rows are all quadratic. A model here has linear rows
  constraining the same `x`; omitting them balances the quadratic rows by
  unbalancing the linear ones. A linear row contributes `|aᵢ[j]|`, which is
  the formula at `Qᵢ = 0`.
* **The objective is left unscaled.** §8 says `σ` "carries over unchanged"
  from `equilibrate()`; that function's own comment is the authority on what
  unchanged means — it applies `σ` *only for a pure LP*, because the Ruiz pass
  already anchors the `P` block against the constraint blocks, and `σ < 1` on
  a problem that has a Hessian shrinks it below the constraint scale,
  degrades strong convexity and diverges the dual iterates. **Tried the
  symmetric-looking thing first and rejected it on that comment**, not on a
  measurement of my own.
* **It refuses rather than falls back.** The surrogates are λ-independent only
  because each `Qᵢ` is constant; a genuine nonlinearity has none, and
  equilibrating against its linear section alone equilibrates against a
  fiction. A model that is not degree ≤ 2 everywhere is a usage error naming
  the alternatives. An accepted scaling option that is quietly not applied is
  #483 exactly.

Also here: `pounce check-x0` gained a **scaling** section reporting the
factors gradient-based will pick *and* the coefficient magnitudes it cannot
see (this is the artefact #703's question 1 needs), plus
`--scaling-max-gradient`. `gradient_obj_scale` / `gradient_row_scale` /
`gradient_scaling_fires` were lifted out of `OrigIpoptNlp` so the preflight
calls the arithmetic that runs rather than a copy free to drift from it.
`gen-qcqp-nl.py --column-scale` writes the exact substitution that makes a
column-equilibration claim checkable against a known answer — without it the
generator draws every `Q` entry from one `U(−10,10)`, its columns are
homogeneous by construction, and a number taken there measures the generator.

## Blast radius

**Off by default. The corpus sweep against the pre-#703 baseline
(`8b5d521`, the branch point) adds exactly the four lines of the two new
fixtures — two legs each — and moves no existing line**, on both the
exact-Hessian and the limited-memory legs, 124 fixture-legs.

Under the option, over the 41 corpus models it accepts, **37 are unchanged**
in status, iteration count and objective. The rest:

| fixture | `gradient-based` | `curvature-based` |
|---|---|---|
| `airport` | 15 it | 14 it |
| `parametric`, `parametric_red_hessian` | 9 it | 8 it |
| `issue_508_infeasible_gap_1em2` | 114 it | 162 it |
| `issue_508_infeasible_gap_1em4` | 441 it | 276 it |
| `pooling_rt2stp` | 181 it, **−3273.955** | 1083 it, **−4391.826** |

`pooling_rt2stp` is nonconvex, so a change of scaling changes the trajectory
and therefore which local minimum is reached — here the published global
optimum of the RT2 pooling instance, for six times the work. Recorded as a
trajectory change, not offered as a win to generalise. Deterministic across
repeated runs.

## Tests

`crates/pounce-cli/tests/curvature_scaling.rs` and five unit tests in
`nl_scaling.rs`. The new tests cannot run on the parent commit — the option,
the module and the fixtures do not exist there — so they were **mutation-
tested** against the shipped code instead, which is the check that actually
matters here:

| mutation | invariance test | unit tests |
|---|---|---|
| emit `D` where `D⁻¹` belongs | **FAILS** | **FAILS** |
| stage 1 disabled (`D = I`) | **FAILS** | pass |
| stage 2 disabled (`eᵢ = 1`) | passes | **FAILS** |

The reciprocal is the one error no test of `D`'s *magnitudes* would catch —
applying `D` where `D⁻¹` belongs doubles the imbalance instead of removing
it — so it is pinned twice. Stage 2 is covered by `the_row_scale_normalizes_
the_row` and not by the fixture pair; that gap is stated rather than papered
over.

The fixture pair is one convex QCQP at n = 24 and the same QCQP under
`--column-scale 1e8`. The assertion is a **measured ulp distance**, not a
tolerance: `curvature-based` reproduces the well-conditioned answer to one
ulp (not bit-exactly — `D` recovers `c` only up to how the Ruiz factors
round), in the same 14 iterations, where the default needs 27 and loses seven
digits.

Two accuracy-guard interactions, both moved deliberately and attributed in
`quad_evaluator_differential.rs`:

* The fixtures are the corpus's first models with a quadratic objective over
  the same variables as a quadratic row, so every shared diagonal entry of
  `∇²L` has two writers and reassociates by one ulp. That module's claim that
  only a tape-plus-matrix entry can do this was a fact about the corpus, not
  the evaluator; corrected in place.
* They move its **frequency** assertion (≈96 one-ulp entries where the rest of
  the corpus produces 2), 0.1% → 1%. **`MAX_HESS_ULPS` is unchanged at 1** —
  and that is why the fixtures carry *one* quadratic row and not two: at two
  rows there are three writers and the distance becomes 2, which would have
  meant loosening an accuracy bound for the whole Q4 fast path to accommodate
  a fixture added for an unrelated reason.

**Deliberately not pinned:** anything about the six downgraded Mittelmann
instances. See the note at the top.

## A separate defect this turned up — needs its own issue

`qcqp_columns_illcond.nl` at n = 24 clears both conic guards, so its *default*
route is the SOCP driver. There it returns

```
SolveSucceeded   28 iterations   objective -4.0065155951e+02
Constraint violation ...  2.66e-15     (the solver's own report)
```

against a true optimum of `-3.6421025382e+02`. Re-evaluating the returned
point with POUNCE's own evaluator violates the quadratic row by `4.948e+01` —
**38% of that row's right-hand side**. The well-conditioned twin is fine on
the same route; the NLP route is fine on both; **a binary built before this
work reproduces it digit for digit**, so it is pre-existing and independent of
everything above.

The mechanism is the one `equilibrate.rs` states in its own module docs:
Ruiz equilibration is wired only into `solve_qp_ipm` because per-row scaling
of `G` breaks a second-order cone, so a badly-column-scaled conic problem
gets no scaling at all. That module predicts a `NumericalFailure`. What
happens here is a silent wrong answer. The fixture pair is a minimal
reproducer and `tests/fixtures/README-qcqp-columns.md` says so beside the
files; §8.5 of the dev-note carries the analysis. **Happy to file this
separately on request.**

---

- [x] Tests fail on the parent commit for the stated reason — *see the
      mutation table; the parent has neither the option nor the fixtures, so
      the tests were mutation-tested against the shipped code instead*
- [x] Book page under `docs/src/` updated (`scaling.md`, `initialization.md`;
      no new page, so `SUMMARY.md` is unchanged)
- [x] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean — 264
      test binaries green (`cargo test` over the default members;
      `--workspace` needs the HSL/Fortran toolchain)
- [x] Every claim in this body and in the code comments is true of the code
      as it stands now

https://claude.ai/code/session_01SnKyT8cw7FWFXXGVtbeer3

---
_Generated by [Claude Code](https://claude.ai/code/session_01SnKyT8cw7FWFXXGVtbeer3)_


jkitchin:
## Review — five findings, all fixed on this branch (88ba9bc)

### 1. Wrong answer on the conic route

`qcqp_columns_illcond.nl` classifies as a convex QCQP, so its **default**
route is the SOCP driver. On that route it returned:

```
SolveSucceeded   28 iterations   objective -4.0065155951e+02
Constraint violation ...  2.66e-15   (the solver's own report)
```

against a true optimum of `-3.6421025382e+02`. Re-evaluating the returned
point with POUNCE's own evaluator violates the quadratic row by `4.948e+01`
— 38% of that row's RHS.

Cause: the QCQP→SOCP reduction factors `Q = sum_k f_k f_k^T` by pivoted
Cholesky, and **the number of factor rows is the cone's dimension**.
`psd_outer_factor` decided rank against one global tolerance,
`tol = 1e-12 * max_diag`. On the ill-conditioned twin `max_diag ~ 4.3e9`, so
`tol ~ 4.3e-3` — larger than 7 of the 24 genuine pivots. The cone was built
in 17 dimensions and then satisfied exactly; `2.66e-15` is an honest report
about the wrong problem. A relative residual check would not have caught it
either: `5.4e-13` against `||Q|| = 4.3e9`.

Diagonal congruence `Q -> CQC` is a change of units and preserves rank
exactly (Sylvester), so the test is now per-pivot, relative to that pivot's
own starting diagonal. One knock-on: once the threshold is relative, pivot
order and rank order decouple, so `break` on first failure had to become a
`settled` mask plus `continue` — a spent pivot's roundoff residual
(`2.0 - (2/sqrt 2)^2 = +4.4e-16`) can outrank a live `1e-20` pivot.

| fixture | rank found | pivot range | residual |
|---|---|---|---|
| `qcqp_columns_wellcond` | 24 / 24 | 8.84e0 … 5.47e1 | 1.07e-14 |
| `qcqp_columns_illcond` | 24 / 24 (was 17) | 1.54e-7 … 4.26e9 | 2.30e-3 |

**Fixture sweep: exactly 2 of 132 lines move**, both this fixture on both
legs, `it=28 obj=-400.6515595` → `it=26 obj=-364.2102508`.

### 2. gh#483: the option was accepted and then not applied

`curvature-based` reaches the engine through `get_scaling_parameters`, which
the convex drivers never call. **38 of the 47 corpus fixtures the option
accepts classify convex**, including both fixtures added for it — so the
headline feature was inert by default on most of the models it exists for.
Now wired into the `decline_convex` machinery #483 established for
`user-scaling`: reroute under `auto`, warn under an explicit
`solver_selection`.

### 3. …but the reroute must not be unconditional

An LP is degree ≤ 2 with every `Q` empty, so the scheme degenerates to Ruiz
equilibration of `[A b]` — which `pounce-convex` already does internally.
Rerouting those buys a different engine and nothing else:

| `lp_israel` (NETLIB, 142×174) | iterations |
|---|---|
| convex, gradient-based (default) | 29 |
| NLP, gradient-based | 135 |
| NLP, curvature-based | 296 |

10× for asking. The gate now requires a nonzero second-order coefficient
(`CurvatureScaling::quadratic`); those 7 fixtures — including all four
gh#690 convex trajectory sentinels — keep the fast path bit-for-bit and get
a note rather than a reroute.

### 4. The blast-radius table measured inertness, not neutrality

"41 accepted, **37 unchanged**" was taken before any routing gate existed,
so most of the 37 were unchanged because *nothing had been applied to them*.
Re-measured with the gate in place: **47 accepted, 33 move, 14 do not**,
decomposed in the dev-note into the reroute (31 fixtures), the
no-curvature keepers (7) and the scaling itself (5).

Every status change on the reroute (`Invalid_Problem_Definition`,
`Not_Enough_Degrees_Of_Freedom`, `Invalid_Number_Detected`,
`Restoration_Failed`) is reproduced on the **pre-#703 binary** by passing
`solver_selection=nlp`; two are also reachable there by
`obj_scaling_factor=-1`, which has triggered the same reroute since #483.
Nothing new.

One number needed chasing: `inconsistent_eq_qp` goes to 2993 iterations
against a cap of 3000. The factors are ordinary (`x = (√2,√2)`,
`g = (1/2,1/3)`), and feeding exactly those to the pre-#703 binary through
`scaling_factor` suffixes reproduces `it=2993` digit for digit — so it is
the NLP path's restoration behaviour under a mild reweighting of a
contradictory row pair, not this scheme. Recorded in the dev-note.

### 5. Two claims that were single-draw artifacts

**"The ulp distance tracks the number of writers"** — it does not; it is a
cancellation statistic. Reassociating `k` terms costs
`~(k-1)·eps·sum|t_i|/|sum t_i|`, and the runaway factor is the second one.
Measured: 2 dense quadratic rows reach **8** ulps where 4 rows reach 2.
`quad_evaluator_differential` now asserts a relative bound as the actual
guarantee, keeps the ulp count as a corpus measurement, and carries the
counter-example as a test.

**"`curvature-based` reaches the published global optimum on
`pooling_rt2stp`"** — one draw of a bistable model. This is the same
mistake #713 was opened and closed on. Across 11 `mu_init` perturbations at
`0.1·(1 ± k·1e-12)`:

| | `-3273.955` | `-4391.826` | other | iters min/median/max |
|---|---|---|---|---|
| `gradient-based` | 7 | 3 | 1 | 81 / 204 / 383 |
| `curvature-based` | 6 | 4 | 1 | 71 / 108 / **1083** |

Same distribution, and the median is *lower* under `curvature-based`. The
sweep's draw is the worst of eleven for the new option.

### Also

Doc comment moved back onto `quadratic_row`; length `debug_assert`s in
`get_scaling_parameters`; a pre-existing temp-file race in the test helper
(`tmp()` keyed only on pid + fixture name) fixed with a sequence counter.

### Verification

- `cargo test --workspace --exclude pounce-hsl` — all green
- CI clippy gate (`-D clippy::correctness -D clippy::suspicious`) — clean
- `cargo fmt --all` — clean
- `scripts/sweep-fixtures.sh` against a `main` baseline — 2 of 132 lines
  move, both explained above

Issue #732, opened 2026-08-21
fix_relax: the predictor_iter budget is the loop's only termination condition, and picks the answer


devin-griff:
`mode="fix_relax"` returns a step 8.6x worse than the unrefined one, because its `predictor_iter` budget is the loop's only termination condition.

On the CSTR of `python/notebooks/36_active_set_parametric_sensitivity.ipynb`, 1406 variables and 1206 equality rows, so 200 degrees of freedom. Perturbing the two initial conditions all the way to the steady state, and varying only the budget:

| `predictor_iter` | largest relative error | variables pinned |
|---|---|---|
| 16 | 0.9500 | 16 |
| 30 | 0.9500 | 30 |
| 60 | 0.9500 | 60 |
| 100 | **7.1554** | 100 |
| 200 | 0.7293 | 3 |
| 400 | 0.6258 | ? |

`mode="linear"` on the same perturbation gives 0.833, so at a budget of 100 the refinement returns an answer **8.6 times worse than not refining at all**. `mode="path"` gives 0.6258 at every budget from 16 to 400.

The pin count equals the budget at 16, 30, 60 and 100. `refine_step_onto_bounds` never terminates on its own here: each pin it adds pushes other coordinates out, so `worst_violation` keeps returning something and the loop runs until `for _ in 0..max_iter` stops it. At a budget of 100 it pins half the problem's degrees of freedom.

The 7.1554 is reproducible run to run, and the warning at that budget shows it pinning 31 `zc` state variables it never touches at lower budgets, not just the `v1` control it is supposed to be repairing.

The same non-monotonicity appears at a smaller perturbation, 25% farther from the steady state: 0.2916, 0.8550, 0.4362, 0.2916, 0.2916, 0.3177 across the same budgets, against `path` steady at 0.2916.

## Where this comes from

The loop diverges from `SensStdStepCalc.cpp` in four ways. I wrote it, in 0787a0d4 and e995cb59, and did not check the reference loop's structure, only the two repairs it implements.

| | upstream | `refine_step_onto_bounds` |
|---|---|---|
| violations per pass | all of them, collected into one list | the single worst |
| termination | `while (bounds_violated)` | `for _ in 0..max_iter` |
| releasing a bound | same list, same pass | only once no pin violation is left, and it clears every pin |
| cost of a release | one more Schur row on the multiplier's row | a refactorization |

[...]

PR #733, opened 2026-08-21
corrector_iter: Newton iterations on the barrier system, against the factorization the solve left behind

devin-griff:
Every prediction mode returns a step built from the held factorization, and that step leaves a residual in the barrier KKT system at the perturbed parameter values. This adds `corrector_iter`, which drives that residual down with Newton iterations against the same factorization, so each one costs a back-solve rather than a factorization. It is available on `estimate()` and `estimate_report()` under every mode and defaults to zero, so nothing changes for existing callers.

## How it works

The residual comes from the algorithm's own calculated quantities by way of the trial iterate, so scaling, fixed variables and the bound expansions are handled exactly as the solve handles them. Setting a trial point leaves `curr` alone, and `curr` is what the held factorization was built from, so nothing here disturbs the factor or any other consumer of the session.

Iterating alone cannot change an active set, because the held operator carries every bound's barrier diagonal from the base point, `z / s`, which at a tightly held bound is `z**2 / mu`. So the predictor's decision is applied once before iterating. A released bound comes out of the operator with its multiplier held at zero and its complementarity row gone, a bound the step brings in has its diagonal raised to what the barrier assigns at that slack, and every other row keeps the base point's term. Both directions are the same change to one diagonal, so one factorization serves the whole correction and every iteration after it is a back-solve.

The changes come from the step's own endpoint, so every mode hands over something. `fix_relax` and `path` decide an active set and pass it on. `mode="linear"` holds the active set fixed as it builds the step, so what reaches the correction is whatever the clamp left sitting on a bound, one change where the other two pass seven.

## What it reaches

How far the correction gets is set by how many bound crossings the predictor hands over rather than by the size of the perturbation. Measured on the CSTR of notebook 36, whose first crossing is at 1.3% of the change to its steady state, with `fix_relax` and eight iterations:

| change | crossings | largest relative error | |
|---|---|---|---|
| 2% | 1 | 2.4e-3 to 1e-6 | converges |
| 5% | 2 | 1.4e-2 to 2e-6 | converges |
| 10% | 4 | 6.0e-2 to 6.0e-5 | converges |
| 25% | 7 | 0.199 to 0.189 | 5% better |

At that last case `mode="linear"` improves by the same five percent, from 0.564 to 0.538. It stays further from the re-solve because its step started further away, not because the correction did less for it.

## What it does not do

It aims at the barrier solution at the `mu` the solve finished on, not at a re-solve, so its accuracy floors out at that offset. It does not claim to converge to the exact answer.

It does not set the multipliers. They arrive extrapolated over the whole perturbation, and once that is large they are the dominant error rather than the operator, which is why the gain at seven crossings is five percent. `estimate()` warns when a correction ends without at least halving the residual, so an uncorrected step is never returned as a corrected one.

## Reporting

`estimate_report(corrector_iter=...)` carries what the iterations did on a `corrector` attribute: back-solves spent, the residual before and after, `released` and `pinned` for the bounds that left the active set and the ones that joined, `active_set_changes` for their total, and the residual split into stationarity, feasibility and complementarity. The three parts carry different units and different consequences. A correction can leave the model's equations nearly satisfied and the multipliers complementary while the Lagrangian's gradient is far from zero, and only the first two say whether the values can be acted on.

`estimate_report` also gains `mode` and `predictor_iter`, because `violation` is evaluated at the predicted point and the corrector block starts from that point's residual, so both are properties of the step. Reporting the linear step's numbers for a `fix_relax` estimate described a step the caller did not take. On a nonlinear model whose linear step leaves a variable outside its bound, the violation is 3.64 under `"linear"` and 0.427 under the other two.

Under `"fix_relax"` and `"path"` the step stops at the bound, so `alpha` is 1.0 and `crossed` is empty for every model, which is the correct answer for such a step rather than a missing one. `activity`, `row_activity`, `mu`, `perturbations` and `bounds_relaxed` come from the base point and the solve and do not move with the mode. The two arguments come after `corrector_iter`, since `estimate_report` shipped in 0.10.0 and its positional order is fixed.

A `refine_stop` attribute carries why the `"fix_relax"` refinement stopped, from the `stop` #734 added to `parametric_step_bounded`. It is None under the modes that run no refinement. A pass pins every crossing it sees, so the pin count says nothing about which limit was reached and this is the only thing that does.

## Testing

Eight integration tests in `crates/pounce-sensitivity/tests/corrector.rs`, nine unit tests in the module, one added to `crates/pounce-sensitivity/tests/cd_split_pin_mapping.rs`, ten in `pyomo-pounce/tests/test_corrector.py` and ten added to `pyomo-pounce/tests/test_estimate_report.py`.

Three of those came out of review and each fails on the commit before its fix. The pin-mapping one uses the leading-inequality fixture already in `cd_split_pin_mapping.rs`, where indexing the pin by its user `g` index instead of its KKT row made the corrector converge to 2.6e-16 against the wrong perturbed problem. The two scaling ones run the release case at variable factors 1, 2 and 10 and at row factors 1, 1e3 and 1e4, where mixing the algorithm's scaled iterate with a step and bounds in the model's own units returned an answer eight orders worse than the step it was given, with the residual halved so nothing warned.

The active-set application is covered in both directions on both layers, because it is the mechanism the correction rests on. The rust QP holds `x2` on its lower bound at the base point. Between p = 0.573, where the true `x2` turns positive, and p = 0.728, where `x1` reaches its upper bound, a perturbation releases exactly one bound and touches nothing else. At p = 0.65 the correction goes from 1.5e-2 to 4.0e-11 and lands on the closed form, which the test computes from the fixture's constants. A second test takes the same p through the plain step, whose endpoint keeps `x2` on the bound, and asserts the residual does not fall, which is the limit the module documents.

The python model holds `x` with a multiplier of 2 and wants it off the bound. `fix_relax` and `path` release and converge, `linear` cannot, so the modes give different answers and `mode=` is load-bearing in a test.

The clamp only engages on badly scaled problems, so no fixture this size reaches it. It is covered on its contract in the module, the way the crate already covers `boundcheck` and `activity`.

I checked the fixtures by mutation rather than by reading them. Each of eight mechanisms was disabled in turn and both suites re-run. Three survived everything at first: releasing a bound, pinning one, and the clamp. The test meant to show the correction works passed with pinning off, because `improved()` is a strict inequality and the iterations still moved the residual by 1e-10 of itself. All eight are caught now, and that assertion carries a magnitude. Making `estimate_report` ignore its `mode` fails seven tests.

That method only asks whether the tests notice a mechanism being disabled. It says nothing about a mechanism written wrongly along a dimension no fixture varies, which is what the three review findings were: no fixture had an inequality ahead of the pin, none had non-unit scaling, and the frozen multipliers were documented as intended.

Full rust suite and 435 pyomo-pounce tests pass. Two failures are pre-existing and reproduce with the branch stashed: `pounce-hsl`'s MA57 test does not link on Windows, and three cases in `python/tests` unrelated to sensitivity.

## `max_iter` becomes `predictor_iter`, and this touches a recent decision

On `estimate()` and `active_set_changes()` the argument that bounds the `fix_relax` refinement passes and the active-set changes `path` applies is renamed to `predictor_iter`, so it says which work it limits alongside `corrector_iter` and `degeneracy_iter`. Neither function is in a release, so nothing downstream carries the old name.

This overlaps with 1187bf3c, which deprecated `max_iter` on `estimate_report` and kept it on the other two. That deprecation is preserved exactly. `estimate_report` still takes `max_iter=None` and still warns. The rename applies only where the budget is live, and `estimate_report`'s new `predictor_iter` is a separate, live argument for the mode's own work. `test_the_other_two_surfaces_keep_max_iter_silent` is updated to the new name and still checks what it checked. Say the word if you would rather the rename came out and went in separately.

## Notebook 36

Restructured to seven sections. Section 4 compares the two prediction modes where they agree and where they part, section 5 sweeps the predictor budget for both on one set of axes, section 6 sweeps the corrector, and section 7 covers `degeneracy` at a held kink. The introduction says why case E is not in the table of measured states.

Rebased onto #734 and re-executed, which moved three passages. Section 4 said the two prediction modes are 7e-7 apart at case A and much further apart at the steady state, where the measured gaps are 4.4e-6 and 4.1e-7, so the largest gap is now at the smallest change. Section 4's full-change passage said the order reverses there and that `fix_relax` lands below the linear step, which was #732 showing: both active-set modes now reach 0.626 against the linear step's 0.833. Section 5 said the modes differ by about 2e-6 at case C and that `path` costs about twice `fix_relax`, where they now agree to 5e-15 and `path` is the cheaper of the two at every cap.


jkitchin (changes requested):
## Review

The design is sound and the write-up is unusually good — the active-set-once-then-back-solve argument, the residual split, and the mutation testing are all the right instincts. The measured tables reproduce: I ran the six integration tests and they pass, and the `max_iter` → `predictor_iter` rename checks out (`git show pyomo-pounce-v0.10.0:pyomo-pounce/pyomo_pounce/sens.py` has `estimate(model, perturb, clamp=True)` — no `max_iter`, and no `active_set_changes` at all, so the claim that neither name is in a release holds).

But the corrector is validated only on fixtures where it happens to be exercised in one frame and one constraint shape. Three defects fall outside that, and each one turns an **essentially exact** step into a wrong one while reporting `improved() == true`. Details and reproducers below; all three are on the branch head (a135342), unmodified except for the probe fixtures.

---

### 1. `residual_at` indexes the pin by user-`g` index, not by `c`-block row

`corrector.rs:150`

```rust
let r = yc_a + row as usize;
```

`algorithm_backsolver.rs:995` states the contract explicitly:

> Pin-row construction must route through this: the flat KKT row of a pinned equality is `n_x + n_s + full_g_to_c_block(g)`, **NOT** `n_x + n_s + g` — those differ whenever any inequality precedes the pinned equality in `g(x)`.

`parametric_step_full` routes through `map_pin_g_to_kkt_rows`; `correct_step` passes `pin_constraint_indices` straight down to `residual_at`, which does the forbidden thing. When the offset runs past the block it errors; when it lands *inside* the block it silently shifts the wrong equality's right-hand side.

Reproducer — the PR's own QP with one **inactive** inequality in front of the pin and a second dummy equality (`g[0]: x1+x2 <= 100`, `g[1]: p = 0` ← pinned, `g[2]: q = 0`). No scaling anywhere:

```
exact                 [0.904752604, 0.034830729]
uncorrected endpoint  [0.904752604, 0.034830729]  err (2.54e-10, 7.12e-11)
CORRECTED endpoint    [0.810526316, 0.000000000]  err (9.42e-2, 3.48e-2)
report: iterations: 6, residual: 0.5732, initial_residual: 0.6500, converged: true,
        released: 1, pinned: 0, feasibility: 0.5732
```

The delta landed on `q = 0` instead of `p = 0`, so the iterations converged toward the wrong perturbed problem. `feasibility: 0.573` is the never-shifted `p` row. `improved()` is true. This needs `full_g_to_c_block` on each pin, and a fixture with an inequality ahead of the pin — the pattern `scaling_invariance.rs` already uses for exactly this reason.

### 2. The corrector mixes the scaled and the natural-units frame

`corrector.rs:364`, `solver.rs:848` / `solver.rs:861-862`

`curr_flat` reads `curr.x`/`curr.z_l`/… — the algorithm's **internally scaled** iterate. `parametric_step_full` returns a step in **natural units** (`solve_released_prebuilt` post-multiplies by `F`). `ctx.lo`/`ctx.hi` are natural units too — `bound_context` divides out `d` on purpose (`solver.rs:1261-1279`). Three places add them together:

- `iterate = base(scaled) + start(natural)` — `corrector.rs:364`
- `iterate += alpha * dir` where `dir` comes back from `solve()` with `F` applied, i.e. natural — inside the loop
- `lo`/`hi` (natural) against `iterate[..n_x]` (scaled) in `released_rows`, `pinned_rows`, `clamp_multipliers`, the fraction rule, and the initial put-back-inside-bounds clamp

Also `out[r] -= d` in `residual_at` subtracts an unscaled delta from a `dc`-scaled equality row.

Reproducer A — `nlp_scaling_method=user-scaling` with per-variable factors `d` on the PR's QP fixture, `RELEASE_DP = 0.65`:

```
d = 1    uncorrected err (2.54e-10, 7.12e-11)   CORRECTED err (3.22e-11, 8.01e-12)   ok
d = 2    uncorrected err (2.54e-10, 7.13e-11)   CORRECTED err (8.48e-02, 3.48e-02)   improved()=true, NO warning
d = 10   uncorrected err (2.54e-10, 7.12e-11)   CORRECTED err (1.52e+00, 6.03e-11)   improved()=true, x1 = -0.62
```

At `d = 10` the returned point is `x1 = -0.62` against a declared lower bound of `0` — contradicting the documented guarantee that "the returned point always satisfies the variable bounds". At `d = 2` the residual halves (4.33e-1 → 1.98e-1), so `residual > 0.5 * initial_residual` is **false** and `estimate()` issues no warning at all while handing back an answer eight orders of magnitude worse than the step it was given.

Reproducer B — the **default** `gradient-based` path, same fixture with the pin row scaled so `dc != 1`:

```
row scale 1e3   uncorrected err (2.54e-10, 7.12e-11)   CORRECTED err (9.52e-2, 3.52e-2)   improved()=true
row scale 1e4   uncorrected err (2.54e-10, 7.12e-11)   CORRECTED err (9.52e-2, 3.52e-2)   improved()=true
```

This is reachable from `estimate()` today: `pyomo-pounce/pyomo_pounce/scaling.py` translates the `scaling_factor` Suffix into variable factors, `repair.py:141` turns on `user-scaling` on its own, and gh#486 stage 3's stated invariant is that "a variable-scaled solve answers every query in the model's own units. Stages 1 and 2 refused those queries rather than answering them in scaled coordinates; nothing is refused now." The corrector breaks that invariant silently.

The CHANGELOG entry — "so scaling, fixed variables and the bound expansions are handled exactly as the solve handles them" — is true of the residual assembly and not true of the corrector as a whole; worth restating once the frames agree.

Whichever frame you settle on, a scaling-invariance test belongs next to `corrector.rs`, the way `scaling_invariance.rs` and `variable_scaling_sensitivity.rs` sit next to the accessors.

### 3. A released multiplier held at zero pins `alpha_d` to zero

`corrector.rs:439`

```rust
let alpha_d = fraction_to_boundary(&iterate[off[4]..off[8]], &dir[off[4]..off[8]]);
```

Released rows are held at `iterate[r] = 0.0`, and they are still in the slice the dual fraction rule reads. For `v = 0` and `d < 0`, `-TAU * 0 / d == 0`, so a single released row with a negative direction drives `alpha_d` to exactly zero — freezing **every** multiplier for that iteration, and for the rest of the correction, since the row stays at zero.

This fires in this PR's flagship test. Instrumenting `the_corrector_releases_a_bound_the_solve_held` on the branch as-is:

```
PROBE iter 1: released row 5: z=0.000e0 dir=-4.550e-1 -> alpha_d=0.000e0
PROBE iter 2: released row 5: z=0.000e0 dir=-3.292e-8 -> alpha_d=0.000e0
```

The dual block never moves. Excluding released rows from the dual fraction rule gives `alpha_d = 1.0` then `0.253`, and all six tests still pass. So "it does not set the multipliers" is currently enforced by an accident of the fraction rule rather than by design — and the `complementarity` figure the report hands callers as a decision quantity cannot be corrected at all whenever a bound is released.

---

### Smaller notes

- `_correct` (`sens.py`) spends an extra full Schur solve on `parametric_step_full` even under `mode="linear"`, where `step` already **is** that step's primal block. Also, under `degeneracy="directional"` the multipliers it grafts on come from the one-sided plain step rather than the directional one that produced `step`.
- `active_set_sigma_x(released: &[usize], pinned: &[(usize, Number)])` takes compound rows in one argument and var-x rows in the other. Both are `usize` and nothing at the call site says which is which — worth a newtype or at least a note on the signature.
- The residual reported to callers excludes released rows (`clear`), which is right, but `initial_residual` is measured after the active-set decision and the put-back-inside-bounds clamp. So `initial_residual` is not the residual of the step the caller handed in, which is how `estimate()`'s warning text reads it. Worth stating.

Happy to push the three probe fixtures as regression tests if that's useful — they're small and each one fails on the current head.

---

devin-griff:
All three are real and all three are fixed, in `205bb676` and `f64417a3`. Each has a test that fails on the parent for the stated reason.

## 1. Pin indexing

`correct_step` now routes through `pin_rows_and_c_scales` and passes flat KKT rows, so it goes the same way every other pin consumer does. The test sits in `cd_split_pin_mapping.rs` beside the existing one, on the same leading-inequality fixture, since that module already exists to pin this contract.

On the parent it returns `[0.1, 0, 0.1]` where `[0.1, 0.1, 0]` is the answer, and reports `residual: 2.6e-16, converged: true`. It converged to sixteen digits against the wrong perturbed problem, which is worse than the wrong answer on its own.

## 2. The frame

Settled on the model's own units, which is what the step and `bound_context`'s bounds already arrive in. The converged iterate is converted once on the way in, and `residual_at` converts back before reading the calculated quantities. That covers all four places you listed, since `released_rows`, `pinned_rows`, `clamp_multipliers`, the fraction rule and the initial clamp now compare quantities in one frame. `out[r] -= d` takes the row scales from the same `pin_rows_and_c_scales` call.

Two tests next to `corrector.rs`, one per reproducer.

Variable factors 1, 2 and 10. On the parent, `d = 2` corrects to `(0.820, 0.0109)` against an exact `(0.9047, 0.0348)`, with the residual falling 0.325 to 0.196, so `residual > 0.5 * initial_residual` is false and nothing warns. The test also asserts the returned point is inside the declared bounds, which is your `x1 = -0.62` at `d = 10`.

Row factors 1, 1e3 and 1e4. This is the one I had fixed and not tested. `use_g_scaling` was false in the first test, so every row factor was 1 and the scales I had just added went unexercised, which is the same gap that produced the three defects. On the parent, a row scale of 1e3 corrects to `(0.8105, 1.1e-18)`, matching your `(9.52e-2, 3.52e-2)`.

The CHANGELOG sentence is restated. It claimed scaling was handled as the solve handles it, which was true of the residual assembly and not of the corrector, and now says what the conversion does.

## 3. The dual step length

Released rows are out of the slice the fraction rule reads. Your numbers reproduce exactly: `alpha_d` was 0 in both iterations and is now 1.0 then 0.253. Two unit tests pin the arithmetic, that a zero entry sets the length to zero and a skipped one does not.

You are right that this was an accident rather than a design limit, and I defended it in the PR body as one. The `complementarity` figure is correctable when a bound is released now, which it was not before.

## The smaller notes

`active_set_sigma_x` takes the note rather than the newtype, naming the two index spaces on the signature.

`initial_residual` now says it is measured at the point the iterations start from, after the active-set decision and the clamp, in the Rust field, in the `EstimateReport` docstring, and in `estimate()`'s warning text, which read it as the step handed in.

`_correct` is documented rather than changed, and I would rather do it separately. The second solve under `mode="linear"` is real, and removing it means restructuring what `estimate` computes and hands over. The directional half cannot be fixed today, since there is no directional full step to take multipliers from, so the docstring says where they come from instead of leaving it to be found.

## On the fixtures

Yes please, if they cover ground mine do not. Mine reproduce all three on the parent and sit where the repo already keeps this kind of test, but yours are the ones you verified and I would rather compare than assume they are equivalent.

jkitchin (changes requested):
Re-verified all three on `f64417a3`, using my own reproducers rather than the shipped tests. **1 and 3 are fixed. 2 is half fixed** — the frame mix-up that produced wrong answers is gone, but the residual is still handed to `solve()` in the wrong frame, and it costs seven orders of magnitude on the correction.

### 1. Pin indexing — fixed ✅

My leading-inequality fixture (`g[0]: x1+x2 <= 100`, `g[1]: p = 0` pinned, `g[2]: q = 0`), which is not the fixture you used:

```
uncorrected err  (2.54e-10, 7.12e-11)
CORRECTED  err   (3.20e-11, 7.86e-12)      was (9.42e-2, 3.48e-2)
feasibility: 1.11e-16                      was 0.5732
```

Routing through `pin_rows_and_c_scales` is the right call. `cd_split_pin_mapping.rs` is the right home for the test.

### 3. Dual step length — fixed ✅

Instrumented `the_corrector_releases_a_bound_the_solve_held` on your head:

```
PROBE iter 1: alpha_d=1.000e0
PROBE iter 2: alpha_d=2.531e-1
```

Matches what you reported. The two unit tests pin the arithmetic in both directions, which is the right level for it.

### 2. The frame — half fixed ⚠️

The half you fixed is real and I confirmed it independently. My reproducer B used a *badly scaled constraint function* under the default `gradient-based` method, so `dc` is derived rather than supplied — a different route to `dc` than your `use_g_scaling` test. It now passes, bit-identical across row scales:

```
row scale 1      CORRECTED err (3.22e-11, 8.01e-12)
row scale 1e3    CORRECTED err (3.22e-11, 8.01e-12)
row scale 1e4    CORRECTED err (3.22e-11, 8.01e-12)
```

Variable scaling gives correct answers and stays inside the declared bounds at every factor. `x1 = -0.62` is gone.

**But `E` is still applied twice.** `residual_at` assembles `out` from the algorithm's calculated quantities, so it is in the **scaled** frame — the comment you added says so ("the residual above is in the algorithm's scaled equality block"), and the deltas are multiplied by the row scales to match. `run` then does `rhs = -resid` and hands it to `solve`, which pre-multiplies by `E` internally because it expects its right-hand side in **natural** units.

Since `K̃ = E K F` and `v_scaled = F⁻¹ v_nat`, the residual transforms as `r_scaled = E r_nat`. So the corrector needs `r_nat = E⁻¹ r_scaled` before the solve, and instead passes `r_scaled` straight in.

`E` is diagonal and nonsingular, so the fixed point is unchanged — which is why the answers are right and your tests pass. What it costs is the Newton direction. The release case, `d` the per-variable factor:

| `d` | final residual, head | with `E⁻¹` applied | stationarity, head |
|---|---|---|---|
| 1 | 3.662e-11 | 3.662e-11 | 4.7e-17 |
| 2 | 1.777e-6 | 3.662e-11 | 1.777e-6 |
| 10 | 4.112e-4 | 1.592e-11 | 4.112e-4 |
| 100 | 1.291e-4 | 9.977e-12 | 1.291e-4 |

The leftover is entirely **stationarity** — feasibility is 0.0 and complementarity sits at the `mu` floor at every factor. That is the `x` block, which is exactly where `E` differs from 1 under a change of variables. Answer error improves too (8.4e-10 → 2.0e-10 at `d = 10`; 1.3e-9 → 2.0e-10 at `d = 100`).

I confirmed the diagnosis by patching only the RHS conversion — all 12 tests in `corrector.rs` and `cd_split_pin_mapping.rs` still pass, and the row-scaling numbers are unchanged, so it is a strict improvement rather than a trade. There is no `E` accessor today; `natural_units_factor` exposes `F` only.

`the_correction_is_the_same_under_variable_scaling` misses this because it asserts only that the corrected answer is within `1e-7`, and on that fixture the step handed in is *already exact* — nothing needs correcting, so an under-corrected direction still lands. Worth asserting the residual reduction too, at a `dp` where the step is genuinely wrong. That is the gap pattern you named yourself: a mechanism written wrongly along a dimension no fixture varies.

### Nit

`corrector.rs:34` — `Index` is now unused and warns on build. CI passes `-D clippy::correctness -D clippy::suspicious` rather than `-D warnings`, so it will not fail the build, but it is new noise.

### On the smaller notes

`initial_residual`, `active_set_sigma_x` and `_correct` all read fine as answered. Deferring `_correct` is the right call — agreed the directional half has nowhere to take multipliers from today, and documenting where they come from is better than a silent graft.

### On the fixtures

Yours cover the same ground as mine for 1 and 3, and I would not add a second copy. The one thing mine reach that yours do not is the **derived** `dc` path — my reproducer B scales the constraint function and lets `gradient-based` compute the factor, where your test supplies `g_scaling` under `user-scaling`. Both routes now work, so it is a coverage question rather than a bug: happy to push that one as a third case in `the_correction_is_the_same_under_constraint_scaling` if you want the default method exercised.

CI is green on all 11 checks at this head.

---

jkitchin:
Pushed the `E` fix as a single commit for you to cherry-pick: **`3e519fb`** on `jkitchin/pounce` branch `claude/pounce-pr-733-review-qc78o4`. It is based directly on your `f64417a3`, so it applies clean:

```
git fetch origin claude/pounce-pr-733-review-qc78o4
git cherry-pick 3e519fb
```

I could not push to `sens-corrector` directly — cross-owner repos cannot be attached to this session, so the fork is out of reach from here. Take it, rewrite it, or drop it; it is yours to land.

### What it changes

Three files, +36/−3.

- `algorithm_backsolver.rs` — adds `scaled_rhs_factor`, the `E` counterpart of the existing `natural_units_factor`.
- `corrector.rs` — divides the residual by `E` before `solve`, and drops the `Index` import your pin-row change left unused (it was warning on build).
- `tests/corrector.rs` — `the_correction_is_the_same_under_variable_scaling` now asserts the residual reduction, not only the answer.

### On the test

Worth saying plainly, because it changed what I shipped. I first wrote a *second* test at `p = 1.0`, where the step is genuinely wrong, expecting that to be the sharp one. It passes on your parent commit unmodified — that case reduces the residual by the same 0.909 ratio at every factor even with `E` applied twice, so it discriminates nothing. I deleted it rather than ship a test that looks like a guard and is not.

The assertion that does bite is the one added to your existing test, and it fails on `f64417a3`:

```
d = 2: the iterations should drive the residual down as far as they do
unscaled, got 7.280e-3 -> 1.777e-6 (stationarity 1.777e-6)
```

Your fixture's step is already exact, so a merely damped direction still lands within `1e-7` and the answer assertion stays green. Only the residual sees it.

### Verification

`cargo test -p pounce-sensitivity` — all 18 test binaries pass, 8/8 in `corrector.rs` and 4/4 in `cd_split_pin_mapping.rs`. `cargo clippy -p pounce-sensitivity --all-targets -- -D clippy::correctness -D clippy::suspicious` is clean, and the warnings that remain in the crate are all pre-existing and outside these lines. `cargo fmt --check` clean.

Residual on the release case after the change:

| `d` | before | after |
|---|---|---|
| 1 | 3.662e-11 | 3.662e-11 |
| 2 | 1.777e-6 | 3.662e-11 |
| 10 | 4.112e-4 | 1.592e-11 |
| 100 | 1.291e-4 | 9.977e-12 |

Constraint scaling is untouched by it — both the `use_g_scaling` route your test uses and the derived-`dc` route through default `gradient-based` give numbers identical to unscaled, before and after.

### Still open, and not in this commit

The derived-`dc` coverage gap I mentioned. Your `the_correction_is_the_same_under_constraint_scaling` supplies `g_scaling` by hand under `user-scaling`; my reproducer B scales the constraint function itself and lets `gradient-based` — the default method — compute the factor. Both work today, so it is coverage rather than a defect, and it seemed wrong to fold an unrelated fixture into a commit about `E`. Happy to send it separately if you want the default method exercised.

---

devin-griff:
Took it. `3e519fb` cherry-picked as `39788d94`, and `391d6694` on top.

I had written the same fix before your comment landed and dropped it for yours. Two things in yours are better than what I had: `scaled_rhs_factor` on `PdSensBacksolver` rather than widening the `SensBacksolver` trait for a single caller, and branching once on `Some`/`None` instead of per element. My numbers were the same as yours, 3.662e-11, 3.662e-11, 1.592e-11 and 9.977e-12 at factors 1, 2, 10 and 100, so the diagnosis is confirmed independently either way.

`391d6694` adds the two things your commit does not carry. The constraint-scaling test gets the same residual assertion the variable-scaling one now has, since the answer landing is no evidence the direction was right in either. And the CHANGELOG says what the conversion does, next to the frame paragraph it belongs with.

Yes to the derived-`dc` fixture, please send it. Supplying `g_scaling` under `user-scaling` and letting `gradient-based` derive it are different enough routes that exercising only one is the same shape of gap as the rest of this.

The test you deleted is worth having said out loud. A case that reduces the residual by 0.909 at every factor with `E` applied twice discriminates nothing, and shipping it would have read as a guard. That is exactly what the answer-only assertion was doing in the test I did ship.

18 test binaries pass, no build warnings, `cargo fmt` clean.

jkitchin:
Derived-`dc` fixture sent: **`d3dda90`**, on the same branch, rebased onto your `391d6694` so it cherry-picks clean.

```
git fetch origin claude/pounce-pr-733-review-qc78o4
git cherry-pick d3dda90
```

Test-only, one file. `ParamQp` gains a `row_scale` written into the constraint itself — `row_scale * x3 = 0` — and `solved_gradient_based` runs it under the default method so `dc` is derived rather than supplied. `solved_with` is refactored through a shared `solved_inner`; its behaviour is unchanged, including leaving `nlp_scaling_method` unset when no factor is given, which is how the fixture always ran.

### It guards the frame, not `E`

Worth being exact about, since the commit message is where this gets read later.

Against the source at **`a135342`** it fails at row scale 1e3 — corrects to `(1, 0.0700)` where `(0.9048, 0.0348)` is the answer, residual 585 → 577. Those are the numbers I reported for this route in the original review, reproduced by the fixture rather than by a probe.

Against **`f64417a3`** — frames agreed, right-hand side still handed over scaled — it **passes**. `E` is applied twice there too, but this fixture's feasibility residual is zero, so the equality block `E` would distort contributes nothing. I kept the residual assertion anyway, for the reason your two carry it, not because it catches that. Saying so rather than letting the commit imply it covers both.

### Not vacuous

The one way this test could quietly rot is `gradient-based` declining to fire and the fixture becoming the unscaled one under another name — the same shape as `use_g_scaling` sitting false. So it asserts the derived factor first:

```
dc == 100 / row_scale     (nlp_scaling_max_gradient is 100)
```

0.1 and 0.01 at the two scales, 1.0 at row scale 1. If the default method's behaviour ever changes, this fails loudly instead of going green on a fixture that stopped testing anything.

### Verification

`cargo test -p pounce-sensitivity` — 18 test binaries, no failures; `corrector.rs` is 9/9. Clippy under CI's flags clean, `cargo fmt --check` clean. I ran the other 8 corrector tests against the refactored helper before adding the new one, to be sure the `solved_inner` change moved nothing.

That closes everything I raised. Nice work on the turnaround — and agreed on `scaled_rhs_factor` belonging on `PdSensBacksolver` rather than widening the trait for one caller.

---

jkitchin (approved):
All four findings addressed and verified against my own reproducers, not only the shipped tests. Clearing the earlier REQUEST_CHANGES.

- **Pin indexing** — routes through `pin_rows_and_c_scales`; my leading-inequality fixture goes 9.42e-2 → 3.20e-11, feasibility 0.573 → 1.1e-16.
- **Frame** — converted once on the way in; both my reproducers pass, including the derived-`dc` route through default `gradient-based`.
- **Dual step length** — released rows out of the fraction rule; instrumented `alpha_d` is 1.0 then 0.253, was 0 throughout.
- **`E` applied twice** — cherry-picked as `39788d94`, with `391d6694` on top adding the matching residual assertion to the constraint-scaling test.

11/11 checks green on `391d6694`.

One thing not in this PR, recorded so it is not mistaken for covered: the derived-`dc` fixture (`d3dda90` on `claude/pounce-pr-733-review-qc78o4`) is **not** merged. `the_correction_is_the_same_under_constraint_scaling` supplies `g_scaling` under `user-scaling`; nothing here exercises `gradient-based` deriving the factor from the Jacobian, which is the default method. Both routes work at this commit, so this is missing regression coverage rather than a live defect — the fixture fails on `a135342` and would have caught the frame bug on that route. The branch keeps it if it is ever wanted.

---

PR #734, opened 2026-08-21
sens: one violation list per fix-relax pass, so max_iter stops picking the answer

### LINKED ISSUE #732: fix_relax: the predictor_iter budget is the loop's only termination condition, and picks the answer

jkitchin:
Fixes #732

*Body kept current with the head; the numbers below are `7a5cfb8`'s. Reviewed by @devin-griff across two rounds — the [first](https://github.com/jkitchin/pounce/pull/734#issuecomment-5371441872) found a regression the first commit introduced, the [second](https://github.com/jkitchin/pounce/pull/734#issuecomment-5373330052) verified the fix on notebook 36's real CSTR, the double column and seven synthetic models, and recommends merging.*

## Problem

`refine_step_onto_bounds` pinned the single worst violation per pass, so
repairing `k` crossings took `k` passes and `max_iter` — documented as a cost
budget — was the loop's only termination condition on any model with more
crossings than passes. The budget then selected the answer.

Notebook 36 needs `pyomo_cvp`, which is not in this repo, so my own
measurements are on a CSTR built the same way with the piecewise-constant
controls written as explicit hold equalities instead of `cvp.parameterize`:
1806 variables, 202 degrees of freedom. Perturbing both initial conditions to
the steady state, `mode="linear"` is 0.8199; "err" is the largest relative
error against a warm re-solve, "rows" what the refinement constrained:

| `max_iter` | before: err / rows | after: err / rows |
|---|---|---|
| 16 | 0.9500 / 16 | 0.7293 / 244 |
| 30 | 0.9500 / 30 | 0.7293 / 244 |
| 60 | 0.9500 / 60 | 0.7293 / 244 |
| 100 | **4.8593** / 100 | 0.7293 / 244 |
| 200 | 0.9500 / 32 | 0.7293 / 244 |
| 400 | 0.9500 / 64 | 0.7293 / 244 |

The row count is exactly the budget at 16/30/60/100; at 100 rows the refined
step is six times worse than not refining at all; the drop to 32 rows at 200 is
the release wiping the pins.

@devin-griff's numbers on the real models are stronger. On the **double column**
(25,356 variables, 1963 required active-set changes, all releases), `main`
returns `rows == cap` at every budget and spends 77.72 s applying 200 of the
1963; this branch applies all 1963 in 0.79 s at every cap including 4. On
**notebook 36's CSTR**, all four perturbations are now flat in the budget:

| case | linear | main | this branch |
|---|---|---|---|
| A 25% out | 1.3007 | 0.2916 | 0.2916, 78 rows, `settled` |
| B quarter | 0.5637 | 0.1995 | 0.1995, 43 rows, `settled` |
| C halfway | 0.9197 | 0.4175 | 0.4175, 51 rows, `settled` |
| D steady state | 0.8329 | 0.9500 / 7.1554 / 0.7293 / 0.6258, = cap | **0.6258, 95 rows** |

Case D is the one #732 opens with. `main` reaches 0.6258 only at a cap of 400;
this reaches it at 16. Of seven synthetic models built to vary the structure,
`main` shows #732 on five; all seven settle here.

And `fix_relax` now agrees with `path` — a mode that decides the active set by
an entirely different route, walking the breakpoints instead of deciding at the
base point — at every case and every cap:

| case | gap before | gap on this branch |
|---|---|---|
| A | 7e-7 | 4.4e-6 |
| B | 1.1e-6 | 5.0e-7 |
| C | 1.8e-6 | 4.8e-15 |
| D | **7.6e-1** | **4.1e-7** |

## Cause

Four differences from `SensStdStepCalc.cpp`, as the issue lays out. Two of them
are what the tables above are:

* one violation per pass instead of the whole list, which makes the number of
  passes the number of crossings, so the budget binds and decides;
* a release clearing every pin and restarting, which is the discontinuity at
  200.

The guard that was meant to catch the 100-pin case checks only that each pinned
row landed where it was asked to. It does, to 1e-3, while the correction does
whatever it likes to the other 1300 coordinates.

## Fix

A pass now takes the whole violation list — every coordinate outside a bound
and every multiplier driven negative, collected together, ordered worst-first —
constrains all of it and re-solves. That is upstream's structure: one
`x_bound_violations_idx` per pass, `while (bounds_violated)` as the termination
condition. The loop ends when the list empties.

**Both batches back off, and the asymmetry between them is the subtle part.**
A pin ADDS a condition, so an over-large batch reports itself: the augmented
system cannot be solved, and the pass retries with the single worst of the new
crossings. A release REMOVES one — each bound taken out of the active set is
stiffness that is no longer holding its variable there — so a batch that takes
too many carries variables off bounds they were sitting on, with no degrees of
freedom left to pin them back and *nothing that fails* to say so. The first
commit here had a backoff only on the half that announces its own failure. On
the CSTR that was 56 releases where 41 were right, carrying five `v1` intervals
below the valve minimum, and cases B and C came back as the linear step.

A release batch of more than one is now kept only when the step it produces is
no further outside the bounds than the step in hand; otherwise the most
negative multiplier goes alone, unconditionally — refusing that too would leave
the loop with a negative multiplier and nothing to do about it — and the next
pass re-measures the rest under it. The batch is judged on the step *after* the
pins already placed are re-solved into the released system rather than on the
raw released base: it costs one Schur solve on a pass that already pays for a
factorization, and it is what lets case A settle in 16 passes instead of 60.
Backing off spends passes, which is the cost: on the double column it is about
9% (0.65 s → 0.71 s), and two of the seven synthetic models report
`iteration_limit` at a cap of 4 with identical answers.

**A release also survives a pin batch that cannot be solved**, and that is two
separable things — the released set, and the step it produced. Rolling back only
the first brings the released bounds back (case D: 0 rows → 95) while the answer
stays the unrefined step's, because `dx` is still restored from a snapshot taken
before any release happened; the snapshot has to sit below the release phase
too, and that is what turns 95 rows into 0.6258. Both are named at the snapshot
in the code (separated by @devin-griff's forward bisection from `a0358c5`, since
reverting from the fixed state shows nothing — the two fixes mask each other).

A bound whose release the factorization refuses is barred rather than asked for
again on every later pass, and its stop is `degrees_of_freedom`, which no budget
reaches, rather than the pass limit.

**A release keeps its re-factorization** rather than becoming a Schur row as
upstream has it. Computing a release from the held factor is what the
`eps · sigma` cancellation documented on `solve_released` measures: 2e-4 off at
`tol = 1e-10` against 7e-9 at `1e-6`, i.e. worse the better the solve converged.
The issue proposed the Schur row and withdrew it in review ("algebraically the
same, not the same in floating point"). What #732 correctly identifies about
releases is fixed — the pins now **survive** one, their right-hand sides
re-measured against the re-solved base instead of the pin set being cleared.

`max_iter` is a safety limit now and the refinement reports which stopping
condition fired. `parametric_step_bounded`, `parametric_step_bounded_decided`
and their Python bindings return `(dx, pinned, stop)` with `stop` one of
`settled` / `iteration_limit` / `degrees_of_freedom` / `worse_than_plain`.
`estimate()` names it in the warning rather than inferring the reason from
`len(pinned) >= max_iter`, which a batched pass makes meaningless.

Both guards from the issue, independent of the loop shape:

* a pass is refused when its correction is out of scale with the step it
  corrects (`CORRECTION_SCALE_LIMIT`, 1e4 — above the leverage a real pin can
  have, below the ~1e15 a dense LU returns for a singular system it does not
  report), not only when a pinned row misses its target;
* the unrefined step is returned when the refinement ends more than 10x further
  outside the bounds than it started (`WORSE_THAN_PLAIN_FACTOR`). This is the
  guard that caught the over-release above — working as intended, on a defect
  that should not have reached it.

Also fixed on the way: the refinement's augmented solves routed an *empty*
released set through `solve_released`, which a backsolver without release
support answers `false` to — such a backsolver could not pin at all. The real
backsolver already short-circuits an empty set to `solve`, so nothing that
worked before changes; it is what makes the guards testable against
`DenseLuBacksolver`.

## Blast radius

**Not a trajectory change: none of this runs during a solve.** The refinement
is post-convergence parametric-step machinery, so the fixture sweep does not
apply and no corpus number moves.

Signature change on `Solver::parametric_step_bounded` and
`parametric_step_bounded_decided` and on their `pounce` Python bindings, from
`(dx, pinned)` to `(dx, pinned, stop)`. In-tree callers are `pyomo-pounce`'s
`estimate()` and the CLI's `--sens-boundcheck`, both updated. #733 adds two more
call sites in its `estimate_report` dispatch; @devin-griff takes those on the
rebase, along with surfacing `stop` on `EstimateReport`.

Not touched: **notebook 36's stored outputs** for the `fix_relax` cells, which
are now stale — section 6 is the `max_iter` sweep this flattens, and the
`fix_relax`/`path` convergence above replaces prose in section 5 and in the
mode-comparison section. Re-running it needs `pyomo_cvp`, which isn't available
here; @devin-griff regenerates it on the #733 rebase.

I have no trustworthy wall-time comparison of my own: my before and after runs
straddle a change in machine load, and the re-solve I measure against moved by
the same factor. The column numbers above are @devin-griff's, measured in one
session on one machine. What is stable across every run is that the cost no
longer varies with `max_iter`.

Separately noticed while measuring, not from this change:
`estimate(mode="path", max_iter=400)` on my CSTR raises
`step_along_path: augmented solve failed (holds [...], released [...])`.
It reproduces identically on `main`. Say the word and I'll file it.

## Tests

`the_pass_limit_no_longer_picks_the_answer` (`parametric_cpp.rs`) is the
headline regression test: the three-free model crosses three bounds at once, and
`max_iter` of 1 and of 8 must agree. **Verified failing on `main`** in a
worktree, with the test back-ported to the two-tuple API so it compiles there:

```
assertion `left == right` failed: all three, worst first
  left: [2]
 right: [2, 1, 0]
```

which is the stated reason — one pin per pass at a budget of one.

Two unit tests on a scripted-release mock cover the release half, both
**verified failing on the first commit** (`a0358c5`) for the stated reason:

* `a_release_batch_that_makes_the_step_worse_backs_off_to_one` — `[2, 3, 0]`
  against `[2]`, which is the CSTR regression in miniature: it takes both
  releases and then has to pin the coordinate the batch carried out of bounds;
* `a_release_the_factorization_refuses_is_not_asked_for_twice` — 3 calls
  against 1, and `IterationLimit` where `DegreesOfFreedom` is the truth.

Writing the second of those turned up a real bug in the barring it tests: with
the refused row filtered out of the candidate list, the loop's own termination
check saw an empty list and called itself `settled` while a multiplier it had
given up on was still negative. Fixed in the same commit.

Two more unit tests pin the guards on a synthetic lever system:
`a_refinement_that_ends_further_out_returns_the_unrefined_step` and
`a_pass_whose_correction_is_out_of_scale_is_refused`. Being honest about these:
they cannot run on `main` at all, not merely fail — `RefineStop` does not exist
there, and the empty-released-set bug above means the refinement could not pin
with a `DenseLuBacksolver` in the first place. The behaviour they pin is new, so
there is no pre-fix version of them that bites.

`fix_relax_pins_three_crossings_at_once` keeps asserting `[2, 1, 0]` — batching
preserves worst-first ordering within a pass — and
`a_pin_beyond_the_degrees_of_freedom_is_refused` gains the
`RefineStop::DegreesOfFreedom` assertion, so the "which limit" answer is pinned
and not just its wording in a warning.

Green here: `cargo test -p pounce-sensitivity` (17 suites), the CLI's sens
tests, `python/tests/test_sensitivity.py` (10), `pyomo-pounce/tests` (320
passed; the 9 failures are `test_block_init_row_scale.py` needing `networkx`,
present before this branch). `cargo fmt --all -- --check` and CI's clippy gate
clean. @devin-griff independently ran the sensitivity suites and `pyomo-pounce`
(413 passed) on `abb43c7`.

---

- [x] Tests fail on the parent commit for the stated reason — the headline test
      and both release tests do, verbatim above; the two guard unit tests cannot
      exist on `main`, as noted
- [x] Book page under `docs/src/` updated (`sensitivity.md`; no new page, so no
      `SUMMARY.md` change)
- [x] Every claim in this body and in the code comments is true of the code as
      it stands now — rewritten at `abb43c7` and again at `7a5cfb8` rather than
      left describing an earlier commit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,26 @@ changes.
   discontinuity came from (a budget of 200 reached a release, wiped
   100+ pins and ended with 3).
 
+  The release batch backs off the way the pin batch does, and the
+  asymmetry is the point: a pin ADDS a condition, so an over-large batch
+  shows up as an augmented system that cannot be solved, while a release
+  REMOVES one — each bound taken out is stiffness no longer holding its
+  variable, and taking too many at once carries variables off bounds
+  they were sitting on with nothing left to pin them back, and no failed
+  solve anywhere to say so. On notebook 36's CSTR that was 56 releases
+  where 41 were right, and two of its four perturbations came back worse
+  than the unrefined step. A batch of more than one is now kept only
+  when the step it produces is no further outside the bounds than the
+  step in hand; otherwise the most negative multiplier goes alone and
+  the next pass re-measures the rest under it. A release also survives a
+  pin batch that cannot be solved, rather than being rolled back with
+  it: a release repairs the active set on its own terms.
+
+  A bound whose release the factorization refuses is barred rather than
+  retried — the same factorization was being asked for again on every
+  later pass — and the stop for it is `degrees_of_freedom`, which no
+  budget reaches, rather than the pass limit.
+
   `max_iter` is a safety limit now, and the refinement reports which of
   its stopping conditions fired. `Solver::parametric_step_bounded`,
   `parametric_step_bounded_decided` and their Python bindings return
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,55 @@ changes.
 
 ## [Unreleased]
 
+- **`mode="fix_relax"` pins every crossing in a pass, so `max_iter` no
+  longer picks the answer (gh#732).**
+
+  The refinement took the single worst violation per pass, so repairing
+  `k` crossings needed `k` passes and `max_iter` — documented as a cost
+  budget — was the loop's only termination condition on any model with
+  more crossings than passes. On the CSTR of notebook 36 the pin count
+  equalled the budget at 16, 30, 60 and 100; at 100 pins, half that
+  problem's degrees of freedom, the refined step came back with a
+  largest relative error of 7.16 against `mode="linear"`'s 0.83 — 8.6
+  times worse than not refining at all — and the error was not even
+  monotonic in the budget.
+
+  A pass now collects the whole violation list, both halves of
+  fix-relax together, constrains all of it and re-solves, which is
+  upstream's own structure: one `x_bound_violations_idx` per pass and
+  `while (bounds_violated)` as the termination condition. The loop ends
+  when the list empties.
+
+  A release keeps its re-factorization rather than becoming a Schur row
+  as upstream has it — that is the computation the `eps · sigma`
+  cancellation measured at 2e-4 off at `tol = 1e-10` is about — but it
+  no longer clears the pins: their right-hand sides are re-measured
+  against the re-solved base. Clearing them is where the budget table's
+  discontinuity came from (a budget of 200 reached a release, wiped
+  100+ pins and ended with 3).
+
+  `max_iter` is a safety limit now, and the refinement reports which of
+  its stopping conditions fired. `Solver::parametric_step_bounded`,
+  `parametric_step_bounded_decided` and their Python bindings return
+  `(dx, pinned, stop)`, where `stop` is `"settled"`,
+  `"iteration_limit"`, `"degrees_of_freedom"` or `"worse_than_plain"`.
+  `estimate()` names it in the warning instead of inferring the reason
+  from the pin count, which a batched pass makes meaningless.
+
+  Two guards independent of the loop shape, both suggested on the
+  issue. A pass is refused when its correction is out of scale with the
+  step it corrects, not only when a pinned row misses its target: the
+  hundred pins above each landed within `1e-3` of where they were asked
+  to go, because achieving the pinned coordinates says nothing about
+  what the correction did to the other 1300. And the unrefined step is
+  returned when the refinement ends further outside the bounds than it
+  started, which is free — `dx_plain` is already in scope.
+
+  Also fixed on the way: the refinement's augmented solves routed an
+  empty released set through `solve_released`, which a backsolver
+  without release support answers `false` to, so such a backsolver
+  could not pin at all.
+
 - **`degeneracy="directional"` decides through the active-set QP engine,
   budgeted by its own `degeneracy_iter`.**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,10 @@ changes.
   step in hand; otherwise the most negative multiplier goes alone and
   the next pass re-measures the rest under it. A release also survives a
   pin batch that cannot be solved, rather than being rolled back with
-  it: a release repairs the active set on its own terms.
+  it: a release repairs the active set on its own terms. That is two
+  separable things — the released set itself, and the step it produced
+  — and a pass that rolls back only the first recovers the released
+  bounds while still answering with the unrefined step.
 
   A bound whose release the factorization refuses is barred rather than
   retried — the same factorization was being asked for again on every
--- a/crates/pounce-cli/src/sens.rs
+++ b/crates/pounce-cli/src/sens.rs
@@ -476,13 +476,31 @@ fn try_compute_sens_step(
             eps,
             16,
         ) {
-            Ok((refined, pinned)) => {
-                if !pinned.is_empty() {
+            Ok((refined, rows, stop)) => {
+                if !rows.is_empty() {
                     eprintln!(
                         "pounce: --sens-boundcheck pinned or released {} bound(s) \
                          and re-solved",
-                        pinned.len()
+                        rows.len()
                     );
+                }
+                // A safety limit that fired, or a refinement that gave
+                // up, is the caller's business: the step returned is
+                // not the one the refinement was asked for.
+                match stop {
+                    pounce_sensitivity::boundcheck::RefineStop::Settled => {}
+                    pounce_sensitivity::boundcheck::RefineStop::IterationLimit => eprintln!(
+                        "pounce: --sens-boundcheck stopped at its pass limit with \
+                         bounds still violated"
+                    ),
+                    pounce_sensitivity::boundcheck::RefineStop::DegreesOfFreedom => eprintln!(
+                        "pounce: --sens-boundcheck could not hold every bound at \
+                         once; the problem's degrees of freedom are spent"
+                    ),
+                    pounce_sensitivity::boundcheck::RefineStop::WorseThanPlain => eprintln!(
+                        "pounce: --sens-boundcheck ended further outside the bounds \
+                         than the unrefined step, which was returned instead"
+                    ),
                 }
                 dx_full = refined;
             }
--- a/crates/pounce-py/src/solver.rs
+++ b/crates/pounce-py/src/solver.rs
@@ -253,7 +253,7 @@ impl PySolver {
     }
 
     /// Parametric step with the bounds respected by pinning rather than
-    /// clamping, as `(dx, pinned)`.
+    /// clamping, as `(dx, pinned, stop)`.
     ///
     /// `parametric_step` points where the linear predictor points,
     /// which can be outside the box. Clamping a coordinate back to its
@@ -280,18 +280,34 @@ impl PySolver {
     /// already respects every bound, and the step is then the plain
     /// one.
     ///
-    /// `max_iter` is a budget rather than a safeguard: the
-    /// refinement is only worth running while it stays cheaper than a
-    /// re-solve. The factorization is never rebuilt, but the Schur
-    /// complement is rebuilt each pass, so pass `k` costs one dense
-    /// `k × k` solve and `k + 1` back-solves and the total grows
-    /// quadratically. The default of 16 is 136 back-solves. What counts
-    /// as outside a bound comes from the solve's own
-    /// `bound_relax_factor` rather than from an argument. Passes stop
-    /// when nothing is outside by that much, at `max_iter`, or when
-    /// the conditions exhaust the problem's degrees of freedom, which
-    /// makes the augmented system singular. None is an error, and
-    /// `pinned` says how far the refinement got.
+    /// A pass takes every crossing it can see, pins them together and
+    /// re-solves, so the number of passes tracks how far the active set
+    /// has to move rather than how many bounds it moves. The
+    /// factorization is never rebuilt for a pin, but the Schur
+    /// complement is, so a pass carrying `k` pins costs one dense
+    /// `k × k` solve and `k + 1` back-solves. What counts as outside a
+    /// bound comes from the solve's own `bound_relax_factor` rather
+    /// than from an argument.
+    ///
+    /// `max_iter` is a safety limit rather than a budget. It was a
+    /// budget while a pass took one pin, which needed as many passes as
+    /// there were crossings, and on a model with more crossings than
+    /// passes the limit picked the answer (gh#732).
+    ///
+    /// `stop` says why the refinement stopped, and is one of:
+    ///
+    /// * `"settled"` — nothing is outside a bound and no multiplier is
+    ///   negative. Only this one says it finished.
+    /// * `"iteration_limit"` — `max_iter` passes were spent with
+    ///   violations left. Raising it may finish the job.
+    /// * `"degrees_of_freedom"` — the conditions exhausted the
+    ///   problem's degrees of freedom, so no step holds every bound at
+    ///   once. No budget helps.
+    /// * `"worse_than_plain"` — the refinement ended further outside
+    ///   the bounds than the unrefined step, which was returned instead
+    ///   with an empty `pinned`.
+    ///
+    /// None is an error, and `pinned` says how far the refinement got.
     #[pyo3(signature = (pin_constraint_indices, deltas, max_iter=16))]
     fn parametric_step_bounded<'py>(
         &self,
@@ -299,7 +315,7 @@ impl PySolver {
         pin_constraint_indices: Vec<i64>,
         deltas: Vec<Number>,
         max_iter: usize,
-    ) -> PyResult<(Bound<'py, PyArray1<Number>>, Vec<i64>)> {
+    ) -> PyResult<(Bound<'py, PyArray1<Number>>, Vec<i64>, &'static str)> {
         let s = self.state.as_ref().ok_or_else(|| {
             PyRuntimeError::new_err(
                 "parametric_step_bounded: no converged factor (call solve() first)",
@@ -313,13 +329,14 @@ impl PySolver {
                 pins.len(),
             )));
         }
-        let (dx, pinned) = s
+        let (dx, pinned, stop) = s
             .inner
             .parametric_step_bounded(&pins, &deltas, max_iter)
             .map_err(solver_error_to_py)?;
         Ok((
             dx.into_pyarray_bound(py),
             pinned.into_iter().map(|p| p as i64).collect(),
+            stop.as_str(),
         ))
     }
 
@@ -397,8 +414,8 @@ impl PySolver {
 
     /// `parametric_step_bounded` with the weak-row decision
     /// supplied by the caller (var-x rows the direction holds) instead
-    /// of searched for. Study surface for an externally solved eq. 14
-    /// QP.
+    /// of searched for, as `(dx, pinned, stop)`. Study surface for an
+    /// externally solved eq. 14 QP.
     #[pyo3(signature = (pin_constraint_indices, deltas, held_var_rows, max_iter=16))]
     fn parametric_step_bounded_decided<'py>(
         &self,
@@ -407,7 +424,7 @@ impl PySolver {
         deltas: Vec<Number>,
         held_var_rows: Vec<i64>,
         max_iter: usize,
-    ) -> PyResult<(Bound<'py, PyArray1<Number>>, Vec<i64>)> {
+    ) -> PyResult<(Bound<'py, PyArray1<Number>>, Vec<i64>, &'static str)> {
         let s = self.state.as_ref().ok_or_else(|| {
             PyRuntimeError::new_err(
                 "parametric_step_bounded_decided: no converged factor (call solve() first)",
@@ -432,13 +449,14 @@ impl PySolver {
             .block_dims()
             .ok_or_else(|| PyRuntimeError::new_err("no converged factor (call solve() first)"))?[0];
         let held: Vec<Index> = validate_var_rows(&held_var_rows, n_x)?;
-        let (dx, pinned) = s
+        let (dx, pinned, stop) = s
             .inner
             .parametric_step_bounded_decided(&pins, &deltas, max_iter, &held)
             .map_err(solver_error_to_py)?;
         Ok((
             dx.into_pyarray_bound(py),
             pinned.into_iter().map(|p| p as i64).collect(),
+            stop.as_str(),
         ))
     }
 
--- a/crates/pounce-sensitivity/src/boundcheck.rs
+++ b/crates/pounce-sensitivity/src/boundcheck.rs
@@ -345,6 +345,161 @@ mod tests {
         );
         assert_eq!(dx, dx_plain.to_vec());
     }
+
+    /// A backsolver whose release half is scripted: the plain solves go
+    /// through a `DenseLuBacksolver`, `solve_released_step` answers
+    /// from `steps` keyed by how many rows are released — a missing
+    /// entry is a factorization that failed — and every call to it is
+    /// counted.
+    #[derive(Clone)]
+    struct ScriptedRelease {
+        base: crate::backsolver::DenseLuBacksolver,
+        rows: Vec<crate::backsolver::BoundRow>,
+        steps: std::collections::BTreeMap<usize, Vec<Number>>,
+        calls: Rc<std::cell::Cell<usize>>,
+    }
+
+    impl crate::backsolver::SensBacksolver for ScriptedRelease {
+        fn dim(&self) -> usize {
+            self.base.dim()
+        }
+        fn solve(&self, rhs: &[Number], lhs: &mut [Number]) -> bool {
+            self.base.solve(rhs, lhs)
+        }
+        fn bound_rows(&self) -> Option<&[crate::backsolver::BoundRow]> {
+            Some(&self.rows)
+        }
+        fn supports_release(&self) -> bool {
+            true
+        }
+        fn solve_released(&self, _released: &[usize], rhs: &[Number], lhs: &mut [Number]) -> bool {
+            self.base.solve(rhs, lhs)
+        }
+        fn solve_released_step(
+            &self,
+            released: &[usize],
+            _rhs: &[Number],
+            lhs: &mut [Number],
+        ) -> bool {
+            self.calls.set(self.calls.get() + 1);
+            match self.steps.get(&released.len()) {
+                Some(s) => {
+                    lhs.copy_from_slice(s);
+                    true
+                }
+                None => false,
+            }
+        }
+    }
+
+    /// `n × n` identity with `lever` at `(1, 0)`, so solving `K y = e0`
+    /// gives `y = (1, -lever, 0, …)`: pinning row 0 drags row 1.
+    fn lever_matrix(n: usize, lever: Number) -> Vec<Number> {
+        let mut a = vec![0.0; n * n];
+        for i in 0..n {
+            a[i * n + i] = 1.0;
+        }
+        a[n] = lever;
+        a
+    }
+
+    #[test]
+    fn a_release_batch_that_makes_the_step_worse_backs_off_to_one() {
+        // Two multipliers are negative. Releasing both takes x0 five
+        // below its lower bound; releasing the most negative one alone
+        // settles. The batch has to earn its place the way the pin
+        // batch does — without that, the CSTR of notebook 36 released
+        // 56 bounds where 41 were right and the step came back worse
+        // than not refining at all (gh#734 review).
+        let calls = Rc::new(std::cell::Cell::new(0));
+        let bs = ScriptedRelease {
+            base: crate::backsolver::DenseLuBacksolver::from_dense(4, &lever_matrix(4, 0.0))
+                .expect("nonsingular"),
+            rows: vec![
+                crate::backsolver::BoundRow {
+                    row: 2,
+                    var_row: 0,
+                    lower: true,
+                },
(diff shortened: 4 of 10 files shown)


jkitchin:
@devin-griff — this is your #732; could you take a look? (GitHub won't let me put you on the reviewer list, so this mention is the request.)

Three things worth your eyes specifically:

1. **The release.** I did not make it a Schur row. Your issue reads upstream's `i + z_L_ItVec_idx` as "the same thing without touching the factorization", but that is precisely the computation the `eps · sigma` cancellation on `solve_released` measures — 2e-4 off at `tol = 1e-10` against 7e-9 at `1e-6`, worse the better the solve converged. So the re-factorization stays and the pins now survive it instead: their right-hand sides are re-measured against the re-solved base rather than the set being cleared, which is the discontinuity you found at a budget of 200. If you think that measurement is answering a different question than the Schur row would face, say so — it's a small change from here.

2. **The batch-overshoot fallback.** When a batch exceeds the remaining degrees of freedom, the pass retries with the single worst of the *new* crossings and lets the next pass re-measure the rest under it. Without that, `a_pin_beyond_the_degrees_of_freedom_is_refused` loses the one pin that does fit. It is not what upstream does, and I'd like a second opinion on whether it should instead bisect the batch.

3. **The two guard thresholds** are judgement calls, not measurements: 1e4 for "correction out of scale with the step it corrects", 10x for "ended further outside than it started". The first sits above the leverage a real pin can have and below the ~1e15 a dense LU returns for a singular system it does not report; the second is loose enough not to fire on the legitimate degrees-of-freedom stop. Both are the loop's safety net rather than what fixes #732, so I kept them conservative — push back if you'd rather they bite harder.

One thing I could not do: notebook 36 needs `pyomo_cvp`, which isn't in the repo, so the table in the PR body is a CSTR built the same way with the piecewise-constant controls as explicit hold equalities (1806 vars, 202 DOF). It reproduces your pathology — row count exactly equal to the budget, a blow-up at 100, the drop to 32 rows at 200 — but the numbers are not yours. Section 6 of the notebook is the `max_iter` sweep this change flattens, and its stored outputs are now stale; re-running it needs your environment.

---

devin-griff:
I have `pyomo_cvp`, so I ran this against notebook 36's CSTR, the model #732 cites under "Reproducing", plus a set of synthetic models and the double column. One finding, with a patch that I have tested against all three.

## The batching works, and the double column is the case that shows it

Double column, N=10 after collocation and `drto.parameterize`, 25,356 variables, initial condenser holdups shifted 2% on five trays. The model needs 1963 active-set changes, all of them releases.

| cap | main rows | main | this branch rows | this branch | stop |
|---|---|---|---|---|---|
| 4 | 4 | 1.69s | 1963 | 0.63s | `settled` |
| 16 | 16 | 6.29s | 1963 | 0.64s | `settled` |
| 60 | 60 | 23.29s | 1963 | 0.65s | `settled` |
| 200 | 200 | 77.72s | 1963 | 0.65s | `settled` |

On main `rows` equals the cap exactly at every budget. At cap 200 it spends 77.72 seconds applying 200 of the 1963 and is nowhere near done, with each pass a dense solve one row larger. This branch applies all of them and settles in 0.65 seconds at every budget including 4. At a 15% shift the same thing happens with 3582 changes.

I also built seven small models to vary the structure: a chain with 1 and with 2 degrees of freedom where pinning one crossing determines the rest, a wide model with 7 free directions where many crossings fit at once, a model that only releases and never pins, and two nonlinear ones. Main shows #732 on five of the seven, `wide` most clearly at 0.1673, then 1.5714, then 0.1673 as the budget goes 4, 16, 60. On this branch all seven are `settled`, flat in the budget, with violation at machine zero.

## The one regression: a release batch has no backoff

On the CSTR, two of four perturbations come back as the linear step.

| case | linear | main | this branch | stop |
|---|---|---|---|---|
| A, 25% farther out | 1.3007 | 0.2916 | 0.2916, 78 rows | `settled` |
| B, quarter of the way | 0.5637 | **0.1995** | 0.5637, 0 rows | `worse_than_plain` |
| C, halfway | 0.9197 | **0.4175** | 0.9197, 0 rows | `worse_than_plain` |
| D, at the steady state | 0.8329 | 0.9500 / 7.1554 / 0.7293 | 0.8329, 0 rows | `degrees_of_freedom` |

D is not a loss. Those main numbers are the erratic ones the issue is about, and `degrees_of_freedom` is more honest than any of them. B and C are different. Main's answer there is 41 releases and zero pins, bound-feasible at exactly 0.0 violation, the same value at caps 16, 60 and 200 and at 25, 50 and 100 intervals. Nothing about it looks budget-decided.

The refinement at those two cases is entirely the release half. Disabling only the `worse_than_plain` early return shows what is being discarded:

| build | releases | left outside a bound | error | stop |
|---|---|---|---|---|
| main | 41 | none | 0.1995 | settled |
| this branch | 56 | 5 coords, worst 0.157 | 0.1404 | `degrees_of_freedom` |
| this branch, one release per pass | 41 | none | 0.1995 | `settled` |

The five are `v1[0]` through `v1[4]`, at 0.157, 0.121, 0.084, 0.046 and 0.007 below the valve minimum, and none of them was pinned. A release takes a bound's `sigma` out of the operator, so the extra 15 releases remove the stiffness that was holding those intervals on their own bound and the step carries them past it. There are no degrees of freedom left to pin them back, the loop stops, and the guard correctly refuses what it produced. The guard is doing its job here.

Limiting releases to one per pass reproduces main exactly, which is the proof of cause. It is also not the fix: the same patch puts the double column back to `rows == cap` and `iteration_limit` at 55.33 seconds for cap 200. Batching releases is what wins the column and what loses the CSTR.

The asymmetry is in the loop. The pin batch backs off when its augmented system cannot deliver:

```rust
let mut next = solve_pins(&pins, &released, &dx_base)?;
if next.is_none() && fresh_pins.len() > 1 {
    pins.truncate(keep.1.len());
    pins.push((fresh_pins[0].0, fresh_pins[0].1));
    next = solve_pins(&pins, &released, &dx_base)?;
}
```

The release batch takes the whole negative-multiplier list, commits whenever `solve_released_step` returns true, and never checks what the step it produced looks like.

## A patch

Keep the batch when the step it produces is no further outside the bounds than the one in hand, otherwise fall back to the most negative multiplier alone and let the next pass re-measure. Replacing the release block with:

```rust
let step_worst = |d: &[Number]| {
    bound_violations(x_curr, d, lo, hi, eps, &[])
        .first()
        .map_or(0.0, |&(_, _, over)| over)
};
let before = step_worst(&dx);
let apply = |set: &[usize]| -> Option<(Vec<usize>, Vec<Number>)> {
    let mut trial = released.clone();
    trial.extend_from_slice(set);
    let mut base = vec![0.0; n_full];
    if !backsolver.solve_released_step(&trial, rhs_plain, &mut base) {
        return None;
    }
    for &r in &trial {
        if let Some(m) = multipliers.iter().find(|m| m.row == r) {
            base[r] = -m.base;
        }
    }
    Some((trial, base))
};
let taken = apply(&fresh_releases)
    .filter(|(_, base)| {
        fresh_releases.len() == 1 || step_worst(base) <= before.max(eps)
    })
    .or_else(|| {
        if fresh_releases.len() > 1 {
            apply(&fresh_releases[..1])
        } else {
            None
        }
    });
match taken {
    Some((trial, base)) => {
        released = trial;
        dx_base = base;
    }
    None if fresh_pins.is_empty() => {
        stop = RefineStop::DegreesOfFreedom;
        break;
    }
    None => {}
}
```

Measured with it applied:

| | main | this branch | with the patch |
|---|---|---|---|
| CSTR B | 0.1995, 41 rel | 0.5637, discarded | **0.1995, 43 rel, `settled`** |
| CSTR C | 0.4175 | 0.9197, discarded | **0.4175, 51 rel, `settled`** |
| CSTR A | 0.2916 | 0.2916 | 0.2916 |
| CSTR D | erratic | 0.8329, dof | 0.8329, dof |
| column, cap 200 | 200 rows, 77.72s | 1963 rows, 0.65s | **1963 rows, 0.71s** |
| seven small models | #732 on five | all settled | all settled |

`cargo test -p pounce-sensitivity` is green across all 17 suites and `pyomo-pounce/tests` passes 413 with it.

What it costs: the column goes from 0.65s to 0.71s, about 9%. Backing off spends passes, so at a budget of 4 two of the small models now report `iteration_limit` where they said `settled`, with identical answers, and CSTR case A needs 60 passes to settle where it settled at 16, again with 0.2916 at every budget. Take the shape or rewrite it, the finding is the asymmetry rather than these particular lines.

## Case D stays open

Zero rows and `degrees_of_freedom` at every budget, with or without the patch, so the first pass cannot place even a single pin after the retry. That is a different path from B and C and I do not have an explanation for it. Main's numbers there were the bug, so nothing is lost, but it is worth knowing why.

## Two smaller things

When `solve_released_step` fails and `fresh_pins` is non-empty, the pass continues without recording the release, so the next pass finds the same negative multipliers and tries the same factorization again until `max_iter`. The stop reported is `IterationLimit`, which does not say a release was the obstacle.

On the re-factorization you are right and the issue was wrong. I wrote that pinning the multiplier row to zero is "the same thing without touching the factorization". It is the same algebraically and not in floating point, and the comment on `shift_released_rhs` already said as much before I filed. Keep it.

## Order with #733

This changes `parametric_step_bounded` and `parametric_step_bounded_decided` to three-tuples, and `sens.py` unpacks them at four call sites. Two of those are in the `estimate_report` dispatch #733 adds and do not exist on main yet. And notebook 36's stale `fix_relax` outputs are the ones #733 restructures. If this lands first I will rebase, fix all four call sites, and regenerate the notebook, which I can run. Worth settling the release backoff first either way, since section 5's numbers move with it.

jkitchin:
Thank you — that is a much better review than the PR deserved, and the finding is right. Pushed `abb43c7` with your patch's shape plus two things it turned up.

## The release backoff: taken

Your framing of the asymmetry is the part I'd missed, so I wrote it into the code and the book rather than just the diff: a pin **adds** a condition, so an over-large batch reports itself as an augmented system that cannot be solved; a release **removes** one, and a batch that takes too many carries variables off bounds they were sitting on with *nothing that fails* to say so. That is why the pin batch got a backoff and the release batch did not — I had generalised from the half that announces its own failure.

I rebuilt my CSTR at your four perturbations and it reproduces B and C exactly in kind before the patch, and settles after:

| case | linear | before | after |
|---|---|---|---|
| A 25% out | 1.2923 | 0.2575, 186 rows | 0.2575, 186 rows |
| B quarter | 0.5611 | 0.5611, 0 rows, `worse_than_plain` | **0.2082, 126 rel, `settled`** |
| C halfway | 0.9083 | 0.9083, 0 rows, `worse_than_plain` | **0.4029, 153 rel, `settled`** |
| D steady state | 0.8199 | 0.7293, 244 rows | 0.7293, 244 rows |

Flat in the budget from 60 passes up, and your predicted cost shows: C reports `iteration_limit` at a cap of 16, at 0.5661 — still under linear, and honestly labelled.

One deviation from your lines. I test the batch on the step *after* the pins already placed are re-solved into the released system, rather than on the raw released base. With no pins the two are identical, which is every case either of us measured; with pins it compares what the step actually becomes. It costs one Schur solve on a release-bearing pass, which already pays for a factorization.

## Your case D: candidate explanation, and it is fixed

I think this was the pass rolling your releases back. The pass snapshotted `released` alongside `pins` and restored **both** when the pin batch could not be solved, so a sound release batch was discarded because the pins that came with it did not fit — leaving zero rows, the plain step, and `degrees_of_freedom`. Exactly your symptom, and it never reaches the "after the retry" reasoning you were looking for, because the retry succeeding or failing is not what discards it.

A release repairs the active set on its own terms, so it now survives a pin batch that fails. My case D settles either way so I cannot confirm it is yours — worth a re-run on the real notebook model.

## The refused-release path

Barred rather than retried, and the stop is `degrees_of_freedom` rather than the pass limit, per your second smaller point.

Writing the test for it turned up a bug in my own barring: with the refused row filtered out of the candidate list, the loop's own termination check saw an empty list and called itself `settled` while a multiplier it had given up on was still negative. Fixed in the same commit — the in-loop check now asks the unfiltered list.

## Tests

Two unit tests on a scripted-release mock, both verified failing on `a0358c5` for the stated reason:

* `a_release_batch_that_makes_the_step_worse_backs_off_to_one` — `[2, 3, 0]` vs `[2]` before, which is your CSTR in miniature: it takes both releases and then has to pin the coordinate the batch carried out of bounds.
* `a_release_the_factorization_refuses_is_not_asked_for_twice` — 3 calls vs 1, and `IterationLimit` where `DegreesOfFreedom` is the truth.

Everything else green: `cargo test -p pounce-sensitivity` all 17 suites, CLI sens tests, `pyomo-pounce` 320 passed (the 9 failures are the `networkx`-missing ones that predate this branch), `python/tests/test_sensitivity.py` 10 passed, fmt and the clippy gate clean. I did not re-measure wall time against your 9% — my two runs straddle a change in machine load and the absolute numbers moved for both the refinement and the re-solve, so your column measurement is the one to trust.

## Ordering

Yours to call, but landing this first sounds right given you can regenerate notebook 36 and I can't — and section 5's numbers move with the backoff, so they want to be regenerated after it rather than before. The four `sens.py` call sites are all yours to fix on the rebase; shout if you'd rather I did anything to make that easier.

---

devin-griff:
Re-ran everything on `abb43c7`. The backoff is right, case D is recovered, and your explanation for it is half of the mechanism.

## Case D, on the real model

| build | error | rows | stop |
|---|---|---|---|
| linear | 0.8329 | | |
| main | 0.9500 / 7.1554 / 0.7293 / 0.6258 | = cap | budget-decided |
| `a0358c5` | 0.8329 | 0 | `degrees_of_freedom` |
| `abb43c7` | **0.6258** | **95** | `degrees_of_freedom` |

Flat at 16, 60 and 200. 0.6258 is what main reached only at a cap of 400, now had at 16.

A, B and C come with it: 0.2916 with 78 rows, 0.1995 with 43, 0.4175 with 51, all `settled` and flat. Case A settles at 16 rather than 60, and reverting your deviation puts it back to 60, so testing the batch against the re-solved step rather than the raw released base is what buys that.

## The mechanism, forward-bisected

Reverting your changes one at a time from `abb43c7` says nothing, because two of them independently keep case D fixed. Applying them one at a time to `a0358c5` separates it:

| on top of `a0358c5` | case D | rows |
|---|---|---|
| nothing | 0.8329 | 0 |
| do not restore `released` on a failed pin batch | 0.8329 | **95** |
| that, plus the release phase writing `dx` | 0.8329 | 95 |
| release writes `dx`, and `keep` moves below the release phase | **0.6258** | 95 |

`a0358c5` captures `keep = (dx, pins, released)` above the release phase and restores all three when the pin batch cannot be solved, so a failed pin batch undid the release twice. Dropping `released` from the snapshot recovers the set, which is the 0 to 95 rows and is exactly what you said. It leaves the estimate at the linear value, because `dx` is still rolled back to a point captured before any release happened. Moving the snapshot below the release phase is what lets the step keep the release, and that is the half that turns 95 rows into 0.6258.

Your code comment covers both, since it says the rollback "discards a repair the step needed and returns the plain one instead". It is the PR message that names only the `released` field, and that is the half that moves the row count rather than the answer.

## Recommend merging

Behaviour is right on every model I have: the four CSTR perturbations, the seven small models, and the double column at two perturbations, all flat in the budget. `cargo test -p pounce-sensitivity` all 17 suites, the CLI sens tests, `pyomo-pounce` 413 passed, `cargo fmt --all -- --check` clean, CI 11 of 11.

The column keeps its result at 1963 rows and `settled`, at 0.79s against 0.65s on `a0358c5`. Measured in one session on one machine so those two are comparable to each other, and against main's 77.72s for 200 of the 1963 it does not signify.

## Ordering, and what moves in the notebook

Agreed, yours first. The gap between the two prediction modes, flat at every budget:

| case | before | on `abb43c7` |
|---|---|---|
| A | 7e-7 | 4.4e-6 |
| B | 1.1e-6 | 5.0e-7 |
| C | 1.8e-6 | 4.8e-15 |
| D | **7.6e-1** | **4.1e-7** |

`fix_relax` and `path` now agree at every case and every cap. That is more than a re-run for notebook 36. Section 5's claim that both modes reach the same accuracy at every cap stops needing the #732 caveat #733 currently carries, and the section comparing the two modes quotes the old four gaps in its prose, so those get replaced. I will do both on the rebase.

Nothing needed to make the four `sens.py` call sites easier, they are a two-tuple to three-tuple change. While I am there I will surface `stop` on `EstimateReport`, since your PR puts it on `parametric_step_bounded` and #733 puts `mode` on `estimate_report`, so without it `estimate_report(mode="fix_relax")` builds a refined step and drops the stop reason on the floor.

jkitchin:
The forward bisection is the right call and it caught a real hole in what I wrote. Pushed `7a5cfb8` — comment and changelog only, no behaviour change.

You are right that the two are separable and that I named one of them. Dropping `released` from the snapshot recovers the released bounds (0 rows → 95) and leaves the answer at the plain step, because `dx` was still being restored from a snapshot taken before any release happened; taking the snapshot below the release phase is what keeps the step, and that is the half that turns 95 rows into 0.6258. Both are now named at the snapshot itself and in the changelog, with your bisection cited, so the next reader gets the mechanism rather than the field name.

Reverting one at a time from the fixed state and seeing nothing is a failure mode I should have expected and did not — two independent fixes masking each other. Applying forward from `a0358c5` is what separates them, and I would not have got there from my own model, where case D settles under either.

Good to know case A settling at 16 rather than 60 traces to testing the batch against the re-solved step. That was a judgement call I made without evidence — I said it costs one Schur solve on a pass that already pays for a factorization, and I did not have a case where it bought anything. Now it has one.

The `fix_relax`/`path` convergence is a better result than anything in my PR body: case D from 7.6e-1 to 4.1e-7, and C at 4.8e-15. Two modes that decide the active set by completely different routes — one at the base point, one walking the breakpoints — agreeing to machine zero is a much stronger statement that the base-point decision is now right than "flat in the budget" is.

Surfacing `stop` on `EstimateReport` is clearly correct and yours to place, since `mode` on `estimate_report` is #733's. Agreed on the rest of the ordering. Nothing further from me unless something new turns up — over to @jkitchin for the merge call.

---

PR #735, opened 2026-08-21
perf(kkt): cut the FERAL back-solve on large limited-memory problems (gh#698 obs 5, #729, #730)

### LINKED ISSUE #729: [performance] Batch the low-rank SMW columns into one `multi_solve`
Filing the measurement from #698 as its own issue so it is findable.

**Base:** pounce `cf461b5595774a3ca017acb437fd74534940bb35`
**Site:** `crates/pounce-algorithm/src/kkt/low_rank_aug_system_solver.rs`, `multi_solve_block`
**Full write-up and logs:** https://github.com/jkitchin/pounce/issues/698#issuecomment-5358901429

> **AI disclosure.** This report — the patch, the runs, the log analysis and this text — was produced with the help of an AI coding assistant, then reviewed by a human before filing.

## Why this is here

This was suggested in #698 as item 1 of a suggested order, we tried it, and it did what it was expected to do. The detail, the method and the attachments are all in the comment linked above; this issue exists so the observation is not buried in a closed thread, and so there is something to close if you decide to act on it.

Not a defect report and not a pull request — an observation that there is measurable performance available here.

## The observation

Under `hessian_approximation = limited-memory`, `multi_solve_block` issues one single-RHS solve per low-rank column. Every column shares one factorization, so all but the first are back-substitutions against the same matrix, and `nrhs` never reaches the backend — so the batched path that exists in both backends is never entered.

Batching them into a single `try_resolve_many_flat`, on a 118 276-row augmented system at `limited_memory_max_history = 6`:

| | stock | batched | |
|---|---|---|---|
| FERAL, `LinearSystemBackSolve` | 72.043 s | 41.948 s | −41.8% |
| FERAL, solver internal | 147.780 s | 115.701 s | −21.7% |
| MA57, `LinearSystemBackSolve` per iteration | 0.1634 s | 0.1280 s | −21.7% |

On FERAL the trajectory is unchanged — all 195 printed iteration rows byte-identical, final quantities equal to all 17 digits — so this is work removed rather than work approximated away.

## Two things worth knowing before acting on it

**It is not bit-neutral on MA57.** HSL's blocked multi-RHS solve reassociates against repeated single-RHS solves, so the SMW correction differs in the last bits and the interior-point trajectory amplifies it. Our run converged in 173 iterations instead of 206, to a marginally better objective. Both optimal, but results move, not just timings.

**The existing tests do not cover the batched arm.** `CountingInner` does not override `try_resolve_many_flat` and its `system_dim()` defaults to 0, so both guards decline and the per-column fallback is what runs. A suite pass is not evidence the batched path is correct.

## What we are offering

The patch we measured with is attached to the linked comment as `smw_batch.patch` — 93 insertions, 7 deletions, one site, with a `None` fallback to the existing loop when the backend declines. It exists to make the measurement possible rather than as a proposed implementation; how you would want this shaped is yours to decide.

Happy to re-run anything on this problem if it would help, or to close this if you would rather track it elsewhere.

### LINKED ISSUE #730: [performance] Under limited-memory, the Hessian-free solves re-analyze the KKT structure for the whole run
*A follow-up to #698. The fix there removed a real callback hazard; this is a structural side effect further down the same path. It is **not** in `uninitialized_h` — we checked that first, and that part matches Ipopt exactly.*

> **AI disclosure.** This issue — the patch, the runs, the log analysis and this text — was produced with the help of an AI coding assistant, then reviewed by a human before filing.

## Why we are reporting this

Our goal is performance parity with Ipopt on a large-scale direct-collocation problem, so that POUNCE can be adopted as a drop-in replacement without a regression.

To make that a like-for-like comparison we integrated POUNCE against MA57 — the same linear solver our Ipopt reference uses, from the same coinhsl build, linked against the same BLAS. That removes the linear solver as a variable and leaves only the two solvers' own behaviour.

The headline of that comparison is that POUNCE already wins it on wall clock. The interesting part is the one phase where it does not, and that is what this issue is about. It is offered as an observation that may be useful rather than as a defect report.

## Summary

Under `hessian_approximation = limited-memory`, `LowRankAugSystemSolver::solve` forwards the Hessian-free solves — the least-square multiplier estimates and the default iterate initializer — to the inner `StdAugSystemSolver` with *the caller's own* W block, which is structurally empty. The main primal-dual solves reach that same inner solver with an `n`-diagonal `Wdiag`.

`StdAugSystemSolver` keys its structure signature on W's nonzero count, so the two shapes alternate and the symbolic factorization re-runs — for MA57 with `ma57_pivot_order = 5` that is a full MeTiS nested dissection of a 118 276-row system, roughly 0.4 s each time.

Handing the inner solver the same zeroed `n`-diagonal in that bypass takes the symbolic row from **4.378 s to 0.837 s with a bit-identical trajectory**.

## Evidence

Minimum-time optimal control problem, direct collocation, Radau order 3: 59 939 variables, 58 271 equality constraints, 33 inequalities, 912 872 Jacobian nonzeros, 118 276-row augmented system. `nnz_h_lag = 0` — there is no Hessian callback.

A/B run back to back on one machine, the only thing changing being `pounce_cinterface.dll`:

| | unpatched | with the patch | Δ |
|---|---|---|---|
| Iterations | 173 | 173 | — |
| Objective | 7.4662814104076219e+01 | 7.4662814104076219e+01 | bit-identical |
| Dual infeasibility | 3.2412858424781281e-04 | 3.2412858424781281e-04 | bit-identical |
| Constraint violation | 2.7556284620633420e-04 | 2.7556284620633420e-04 | bit-identical |
| Complementarity | 5.0278416139812678e-06 | 5.0278416139812678e-06 | bit-identical |
| **`LinearSystemSymbolicFactorization`** | **4.378 s** | **0.837 s** | **−3.541 s, −81%** |
| `LinearSystemFactorization` | 26.319 s | 25.902 s | −0.417 |
| `LinearSystemBackSolve` | 23.948 s | 23.190 s | −0.758 |
| solver internal | 82.287 s | 77.014 s | −5.273 |

The bit-identical trajectory is the load-bearing part: `w_factor` is already `0.0` at both call sites and the substituted diagonal is exactly zero, so nothing numerical moves. The work was removed, not approximated away.

We are deliberately **not** claiming the wall-clock delta. The two runs perform identical callback work, and their callback totals still differ by 15.3 s (217.765 → 202.477) — that is our in-session noise floor, and it is three times the size of the effect. The symbolic row is the only number here that is safely outside it.

For scale, an earlier internally-consistent set on the same problem, Ipopt against POUNCE:

| | Ipopt MA57 | POUNCE MA57 | POUNCE MA57, SMW-batched |
|---|---|---|---|
| Iterations | 201 | 206 | 173 |
| `LinearSystemSymbolicFactorization` | **0.809 s** | **12.631 s** | **4.269 s** |
| Wall | 279.393 s | 335.684 s | 268.953 s |

Two things to read from that. POUNCE is already ahead of Ipopt on wall clock. And the symbolic row swings by 3× between two runs of nearly the same code, because `recalc_y` is gated by `recalc_y_feas_tol` and only engages in the tail — so how many analyses you pay for depends on when feasibility crosses that threshold. Patched, the row lands on 0.837 s against Ipopt's 0.809 s.

## Mechanism

The empty `zeroW` is **not** the problem, and this was our first hypothesis. Ipopt does exactly the same thing:

```cpp
SmartPtr<const SymMatrix> OrigIpoptNLP::uninitialized_h()
{ return h_space_->MakeNewSymMatrix(); }
```

With `nnz_h_lag == 0` that block is empty, and both our Ipopt and POUNCE runs report `Number of nonzeros in Lagrangian Hessian: 0`. POUNCE's `OrigIpoptNlp::uninitialized_h` (`orig_ipopt_nlp.rs:2223`) returns an empty `SymTMatrixSpace` in the same situation. That is faithful, and we do not think it should change.

The difference is what the two projects do with it downstream.

Ipopt pins the augmented-system structure once and treats it as invariant thereafter:

- `IpStdAugSystemSolver.cpp:127` — `if( !IsValid(augmented_system_) )` guards `CreateAugmentedSpace`; afterwards only values are rebuilt.
- `IpTSymLinearSolver.cpp:173` — `if( !initialized_ )` guards `InitializeStructure`, latched at `:419`.
- `IpTSymLinearSolver.cpp:182` — `DBG_ASSERT(nonzeros_triplet_ == TripletHelper::GetNumberEntries(sym_A));`

That assert states the contract: the triplet count must not change for the life of the solver. Ipopt's inner solver never sees two W shapes, because under limited-memory the low-rank layer owns those solves.

POUNCE re-derives the decision on every assemble instead, and the bypass at `low_rank_aug_system_solver.rs:759` is what makes the input vary:

```rust
// bypass the SMW machinery and solve directly through the inner
// augmented-system solver with the original coefficients.
let Some(lr_w) = lr_w_opt else {
    ...
    let status = self.inner.solve(coeffs, rhs, sol, check_neg_evals, num_neg_evals);
```

Those original coefficients carry the empty W. The main path, by contrast, substitutes an `n`-diagonal (`low_rank_aug_system_solver.rs:286-288`). The inner solver then keys on the difference (`std_aug_system_solver.rs:361`):

```rust
if !self.initialized || self.struct_sig != Some(sig) {
```

where `sig` contains `w_nnz`, and `w_nonzeros` (`std_aug_system_solver.rs:826-834`) maps a `SymTMatrix` to `nonzeros()` — `0` here — and a `DiagMatrix` to its dimension — `59 939`. The signature therefore flips on every alternation, and each flip costs an `ANALYZE`.

The comment already sitting at `std_aug_system_solver.rs:473-479` describes this arrangement as intentional, so the alternation is known; what appears not to have been costed is that it re-triggers the symbolic factorization rather than merely rebuilding a triplet.

## On the attached patch

`wdiag_bypass.patch` makes the bypass hand the inner solver the same zeroed `n`-diagonal shape the main path already passes, so the signature holds.

It is attached so the measurement can be reproduced, **not** as a proposed design — how this should be shaped is yours to decide. Other shapes that would work, and may well be preferable:

- Relax the signature so a structurally-subset W reuses the pinned structure instead of forcing a rebuild.
- Assemble the union pattern once at `initialize_structure` time and keep it fixed, which is closest to what the Ipopt assert above encodes.
- Have the low-rank solver own the Hessian-free solves outright, so the inner solver's W shape is invariant by construction.

One note on where the patch is *not*. We first tried this in `OrigIpoptNlp::uninitialized_h`, returning an `n`-diagonal from the `nnz_h_lag == 0` arm. It worked and passed the suite, but it is wrong: that arm is also reached by an exact-Hessian NLP whose Hessian is genuinely structurally empty, and there the main path's W *is* the empty block — so it would have introduced the mirror-image mismatch for that case. Nothing in the test suite covers it. Fixing it inside the low-rank solver confines the change to the limited-memory path, where the `n`-diagonal is already the established shape.

`cargo test -p pounce-algorithm -p pounce-nlp --features ma57 --release` is unchanged at 60 binaries, 875 passed, 0 failed. As with our earlier report, we would not treat that as evidence the new path is correct — the rejected variant above passed it too. The evidence we would point at is the bit-identical trajectory over 173 iterations of a 118 276-row system.

## Correction to an earlier report

This investigation also turned up an error in our own SMW-batching numbers (the report on #698, now tracked as #729), which we would like to correct here since this issue supplies the explanation.

We reported the MA57 arm as **`solver internal, per iteration: 0.5288 → 0.4501, −14.9%`**. Roughly half of that delta is the `LinearSystemSymbolicFactorization` row (0.0613 → 0.0247 s/iter), which batching cannot influence — it moved because the batched run's different trajectory engaged `recalc_y` fewer times. We had switched to per-iteration figures specifically to correct for the shorter trajectory, but that normalization does not remove this term, because the number of analyses is not proportional to the iteration count.

The defensible MA57 attribution for batching is therefore about **−7% on solver internal**, not −14.9%. The −21.7% back-solve row is unaffected and stands, as do all the FERAL figures — that arm is bit-identical and its symbolic row is flat (0.213 → 0.251 s).

## Configuration

- `hessian_approximation: limited-memory`, `limited_memory_max_history: 6`
- `linear_solver: ma57`, `ma57_pivot_order: 5` (MA57 automatic — probes for and uses MeTiS), `ma57_pre_alloc: 2.0`
- `linear_system_scaling: slack-based`, `nlp_scaling_method: gradient-based`
- `recalc_y: yes`, `recalc_y_feas_tol: 0.05`
- `mu_strategy: monotone`, `presolve: yes`, `print_timing_statistics: yes`
- Base commit `cf461b5595774a3ca017acb437fd74534940bb35`, target `x86_64-pc-windows-gnu`, `-C target-cpu=haswell`
- OpenBLAS built `USE_THREAD=0 USE_OPENMP=0`, and `OPENBLAS_NUM_THREADS=1` in the environment — both solvers link the same prefix, so BLAS threading is not a variable here
- Both arms of the A/B include our SMW batching change, so it cancels

## Attachments

- [wdiag_bypass.patch](https://github.com/user-attachments/files/31280688/wdiag_bypass.patch) — the change measured above
- [ma57_base_resample.log](https://github.com/user-attachments/files/31280684/ma57_base_resample.log) — unpatched, this session
- [ma57_wdiag_bypass.log](https://github.com/user-attachments/files/31280689/ma57_wdiag_bypass.log) — patched, this session
- [ipopt_ma57.log](https://github.com/user-attachments/files/31280685/ipopt_ma57.log), [pounce_stock_ma57.log](https://github.com/user-attachments/files/31280682/pounce_stock_ma57.log), [pounce_batched_ma57.log](https://github.com/user-attachments/files/31280687/pounce_batched_ma57.log) — the earlier internally-consistent Ipopt/POUNCE set
- [config_pounce_ma57.json](https://github.com/user-attachments/files/31280686/config_pounce_ma57.json), [config_ipopt_ma57.json](https://github.com/user-attachments/files/31280683/config_ipopt_ma57.json)

jkitchin:
Four independent pieces of work that all land on the same hot path — the KKT
back-solve on large limited-memory problems — plus the two performance issues
Srikanth filed against it.

Fixes #729
Fixes #730

## The headline

`laptime` (58 014 variables, L-BFGS, 126 028-dimension KKT, `max_iter=60`,
FERAL). Round-robin interleaved timing, three rounds per arm, so both arms of
each pair see identical ambient load:

| | before | after |
|---|---:|---:|
| Overall | 62.52 s | **14.74 s** |
| Back-solve | 49.47 s | **5.35 s** |
| Factorization | 10.05 s | 6.51 s |
| Iterations | 60 | 60 |

`dirichlet120` (53 881 variables, exact Hessian, to convergence):

| | before | after |
|---|---:|---:|
| Overall | 23.08 s | **15.25 s** |
| Back-solve | 9.90 s | **2.56 s** |
| Status | Optimal | Optimal |
| Iterations | 170 | 176 |

Objective agrees to 8 significant figures. MA57 on `laptime` is unchanged
(10.17 s -> 10.20 s), which is the control: none of this should reach a backend
that has no in-backend refinement.

## What is in it

**`2c4f25f1` — wire `increase_quality`, turn the backend refinement off for the
IPM (gh#698 obs 5, gh#710).** FERAL's inner refinement ran nested inside
`PdFullSpaceSolver`'s own refinement loop, at roughly 4x the back-solve cost,
working to a residual ~300x tighter than anything the caller reads. Ipopt
disables backend refinement on every direct solver it ships, for the reason
Wachter gives: refinement belongs on the *unreduced* system, which the IPM
already refines. The rung Ipopt uses in place of it is `IncreaseQuality`, which
pounce's FERAL interface returned `false` from — so the flip was unsafe until
this commit wires the escalation ladder (scaling Identity -> InfNorm, then
`pivot_threshold^0.75`), mirroring `IpMa57TSolverInterface.cpp:832`.

The flip is scoped to `feral_config_from_options`, **not** `FeralConfig::default()`.
Turning the backend loop off is only safe for a caller that refines the
unreduced system *and* escalates. pounce's IPM does both; pounce-convex,
pounce-rs and pounce-py do the first only (`hsde.rs:67 IR_MAX_PASSES = 5`), so
they keep `refine: true`.

**`a1dcc226` — port Ipopt's square-problem path (gh#508).** `IsSquareProblem()`
(`IpIpoptCalculatedQuantities.cpp:3732`) gates `ComputeFeasibilityMultipliers()`
(`IpIpoptAlg.cpp:409`, `:857`, postprocess at `:949`) and the corresponding
branches in `IpRestoMinC_1Nrm.cpp:152`/`:269`. pounce had none of it. Adds
`probe_convergence` — a snapshot/restore wrapper needed because pounce's
`OptErrorConvCheck` carries state Ipopt's does not (gh#505 streak, gh#200 veto
budget, gh#533 progress window).

**`dbe32150` / `571baa99` — batch the low-rank SMW columns (#729).** One
multi-RHS back-solve instead of a loop. Gated on `FERAL_BITWISE_MULTI_SOLVE_MAX_NRHS`
(16), deliberately below feral's private `BLAS3_NRHS_THRESHOLD` (32), so the
blocked back-substitution is **bit-identical** to looping rather than merely
tolerance-equal. A unit test exercises a real factor at exactly that width and
fails if feral ever lowers its threshold.

**`7623dca7` — separate inner solver for the L-BFGS Hessian-free bypass (#730).**

**`37665b12` / `d240dc4a` — `feral_refine_target` and the notes behind it.**

**`2b65e1d5` — make the square-problem restoration guard relative (gh#508).**
The absolute violation test was not scale-invariant: the same model with rows
scaled by 1e-4 took the other branch. Adds `ROW_NOISE_KAPPA` and a fixture
(`infeasible_square_scaled_1em4.nl`) that pins it.

**`11baabcb` — restore the refinement the `feral_refine = false` default
removed.** Turning the backend's inner refinement off removed the *only*
refinement the sens held-factor back-solves were silently relying on. Fixed at
the place the refinement belonged; see "The hole the default opened" below,
which also covers the one fixture the default moves across a knife edge
(`deb7` under `least_square_init_primal`) and why reference Ipopt says the
assertion it broke was never true of that fixture.

## Evidence for #729, isolated

Built main plus gh#729 alone and compared against clean main on `dirichlet120`:

| | main | main + #729 |
|---|---:|---:|
| Back-solve | 9.41 s | **7.61 s (-19%)** |
| Overall | 22.48 s | 20.39 s |
| Objective | *bit-identical* | *bit-identical* |
| Iterations | 170 | 170 |

Reproduces the reported win on FERAL.

**One thing the reporter should know:** their measurement was on MA57 (-21.7%),
and this PR gates batching **off** for MA57. MA57 reassociates the blocked
back-solve, so it is not bit-identical to looping — and on `pooling_rt2stp` MA57
batching reaches a 25% worse objective while still reporting `Optimal`. The
FERAL win ships; the MA57 variant does not, deliberately. Reopening that is a
separate decision about accepting a tolerance-legal trajectory change on MA57.

## Evidence for #730

This one took a while to reproduce, because the precondition in the report is
not the operative one. The report describes the bypass's `W` as "structurally
empty", which reads as a property of the model (`nnz_h_lag = 0`). It is not —
no model in the benchmark corpus has a zero Hessian, and every measurement
against `laptime` (`nnz_h_lag = 175 000`) and `dirichlet120` (`260 191`) showed
exactly zero difference.

The gate is the **`recalc_y` option**. The Hessian-free solves — the
least-square multiplier estimates — are only issued when `recalc_y` engages.
Under default options on these models it never does, the bypass is entered once
at startup, and there is nothing to alternate with.

Force it on and the pathology is unambiguous. `laptime`, MA57, limited-memory,
`max_iter=30`:

| `recalc_y` | arm | Symbolic factorization | Overall |
|---|---|---:|---:|
| no (control) | before | 0.905 s | 5.607 s |
| no (control) | after | 0.889 s | 5.481 s |
| **yes** | **before** | **27.987 s** | **38.538 s** |
| **yes** | **after** | **0.921 s** | **8.232 s** |

Objective bit-identical within each pair, 30 iterations both. Turning `recalc_y`
on used to cost 33 extra seconds; it now costs 2.7, which is what the extra
least-square solves should cost.

Attribution without needing a #730-only build: the delta exists *only* when
`recalc_y` is on, and with it off the two binaries are indistinguishable.
Nothing else on this branch is keyed to `recalc_y` — #729's batching is gated
off for MA57, the refinement flip only touches `FeralConfig`, and the
square-problem path cannot fire on a non-square model with an unchanged
iteration count.

The FERAL arm shows the same fix but a small absolute win on this axis
(symbolic 0.119 s -> 0.006 s), because FERAL's analysis is cheap. The symbolic
thrash is an MA57-scale problem, consistent with the reporter running MA57.

## The hole the default opened

`feral_refine = false` is the change this branch is *for*, and it is right on
the IPM path — Ipopt disables backend refinement on every direct solver it
ships. But two paths were relying on that backend loop without saying so, and
both surfaced as CI failures rather than as anything the sweep could see.

**Sens held-factor back-solves.** pounce#77 turned `PdFullSpaceSolver`'s
refinement loop off for sens back-solves (`allow_inexact = true`) on the
argument that a back-solve already lands below `tol`. That argument had a hidden
dependency: the loop was off, but the backend was refining underneath, so the
answers were accurate anyway. With nothing refining, two exactly-dependent rows
of `K^-1` come back differing by ~1e-12 instead of ~1e-16.

"Below `tol`" is not the property these callers need. A sens back-solve is read
as a *derivative* — one shot, no outer loop to absorb the error — and
`_rank_deficient` thresholds `matrix_rank` at `sigma_max · n · eps` (~9e-16 for
a 2x2). At 1e-12 two duplicated design points read as rank 2 and the "linearly
dependent coordinates" refusal never fires
(`test_wrt_within_count_dependent_block`). MA57's `icntl[9] = 0` means this hole
was already open there.

So the ordinary back-solves refine again — **except** under a `SigmaOverride`.
With an override in play (crossover, gh#654; or the bound-release path) the held
factor does not factor the system being solved, so the refinement loop measures
a residual against a different matrix, cannot converge, and escalates. Refining
all three sites is what broke `crossover_sigma_downstream`; `may_refine()` now
names the condition once.

Cost, over 120k/150k/300k KKTs: **0.136/0.120/0.217 s** refined against
0.138/0.123/0.215 s unrefined — inside the noise.

**deb7 under `least_square_init_primal` is not fixed, and is not an accuracy
problem.** `issue_616_ls_init_downgrades` asserts that solved-or-acceptable does
not depend on `least_square_init_primal`. On x86-64 this branch breaks that for
`deb7`: ls_init=yes reaches `ErrorInStepComputation` at it=358 where ls_init=no
succeeds at it=131. `feral_refine=yes` cures it (145 / 143, both succeeded), so
it is squarely caused by this branch's default.

The obvious reading — the backend loop was covering a genuine accuracy deficit,
so escalate to it when the step computation is in trouble — was tried and is
wrong on the facts:

- `increase_quality` is **never called** on the deb7 trajectory (instrumented: 0
  calls), and `PdFullSpaceSolver`'s refinement loop never stagnates there (0
  escalation events). The escalation chain latches on `augsys_improved`, so
  `increase_quality` fires at most once per solve, and
  `TSymLinearSolver::increase_quality` can satisfy that one call at the wrapper
  level by flipping `use_scaling` on without reaching the backend at all. A rung
  added to the feral side is unreachable code here.
- `feral_refine_steps` is **non-monotone**: `steps=1` fixes
  `obj_scaling_factor=10` but not `max_soc=0`; `steps=2` fixes `max_soc=0` and
  breaks `obj_scaling_factor=10`. `feral_refine_target` scatters the same way
  (`1e-8` fixes two of three local repros, `1e-10` and `1e-12` fix none). More
  refinement is not monotonically better, which is not how an accuracy fix
  behaves.

What it is instead: deb7 + `least_square_init_primal` is a chaotic case, and the
assertion is sampling that chaos. Perturbing one innocuous option at a time (20
options, macOS), the `solved(on) == solved(off)` assertion is violated by **main
itself 2/20 times**, by **MA57 2 of the 8 sampled**, and by this branch 7/20.
The perturbations that break main (`obj_scaling_factor=10`, `max_soc=0`) are not
the ones that break the branch. Both binaries are walking a knife edge; this
branch stands on a different side of it at the one sampled point the test pins.

The decisive check is what **reference Ipopt** does. Ipopt 3.14.19 on `deb7.nl`,
one innocuous option perturbed at a time, solved-or-acceptable per route:

| perturbation | `least_square_init_primal=yes` | `=no` |
|---|---|---|
| `max_soc=0` | Solved To Acceptable Level | **Error in step computation** |
| `tol=1e-6` | **Error in step computation** | Optimal Solution Found |

2 of 8 sampled perturbations flip Ipopt's own verdict with the option. The
implementation POUNCE is a port of does not satisfy the invariant on this
fixture, so the invariant is not a property of `deb7` to begin with.

`deb7` therefore leaves the invariant list and gets its own test,
`deb7_is_route_sensitive_in_ipopt_too`, pinning what it does guarantee — it
solves on the default route — with the Ipopt table recorded in the doc comment.
The other six fixtures keep the full assertion, unweakened.

**The headline win survives the sens fix.** `laptime`, `max_iter=60`, same machine back
to back:

| | overall | back-solve |
|---|---:|---:|
| main | 16.260 s | 1.683 s |
| this branch, before the sens fix | 14.214 s | 0.420 s |
| **with the sens fix** | **14.660 s** | **0.462 s** |

Re-swept both legs against the pre-fix branch baseline: **zero moved legs.**

## Fixture sweep

This is a trajectory change, so `scripts/sweep-fixtures.sh` was run against a
0.10.0 baseline binary, **both legs**. The sweep was first confirmed
load-insensitive: two runs of the same binary launched concurrently were
byte-identical.

**17 of 128 legs move.** Full accounting below. The MA57 column is the
reference — MA57 has no in-backend iterative refinement, so it is the backend
whose behaviour matches Ipopt, and "FERAL moved toward MA57" is the expected
direction here rather than a regression.

| leg | baseline (FERAL) | MA57 (reference) | new (FERAL) | reading |
|---|---|---|---|---|
| lbfgs autocorr_bern55-06 | ErrorInStep 782, -2282.454 | Acceptable 694, -2304.000028 | Acceptable 945, **-2304.000028** | lands on MA57's status and objective exactly |
| lbfgs pooling_rt2stp | RestoFailed 27, 9.32 | Succeeded 100, -4391.826001 | ErrorInStep 105, -4391.825986 | baseline quit at it=27; now reaches MA57's objective at MA57's iteration count |
| lbfgs deb7 | RestoFailed 2941, 119.806 | MaxIter 3000, 97.56362 | RestoFailed 2993, 97.56380 | objective moves to MA57's |
| lbfgs cresc4 | Succeeded 143 | Succeeded 110 | Succeeded 105 | toward MA57, 38 iters faster |
| lbfgs eigena2 | Acceptable 242, 82.5 | **ErrorInStep 249, 82.5** | ErrorInStep 262, 82.5 | MA57 has *always* failed this tail; FERAL's refinement was concealing it. Same objective, feasible to 1e-14 |
| lbfgs eigenb2 | Succeeded 59, 1.6 | ErrorInStep 55, 1.6 | Acceptable 41, 1.6 | better than MA57, 18 iters faster than baseline |
| exact deb7 | Succeeded 171 | Acceptable 124 | Succeeded 146 | toward MA57 |
| exact pooling_rt2stp | Succeeded 128 | Succeeded 160 | Succeeded 107 | 21 iters faster, same objective |
| exact issue_508_gap_1em2 | Infeasible 114 | — | Infeasible 112 | square path |
| exact issue_508_gap_1em4 | Infeasible 441 | Infeasible 48 | Infeasible 245 | square path, 196 iters faster |
| lbfgs issue_508_gap_1em2 | Infeasible 97 | — | Infeasible 68 | square path |
| lbfgs issue_508_gap_1em4 | Infeasible 81 | Infeasible 72 | Infeasible 130 | **the one leg moving away from the reference** — see below |

Five further legs are last-digit only, same status and iteration count, ULP
drift from removing the backend refinement pass:

- exact infeasible_equalities  it=28  1.201505736 -> 1.201505752
- exact unbounded_exp          it=23  -8.688703979e+20 -> -8.688655722e+20 (diverging by construction)
- lbfgs hs13_bigstart          it=42  0.9850143991 -> 0.985014399
- lbfgs infeasible_equalities  it=33  2.663255238e-18 -> 2.663255234e-18
- lbfgs jit1_boxed             it=31 -> 32, objective identical

No leg turns a correct answer into a wrong one, and no leg loses a verdict on
an infeasible problem.

### The one leg that moves the wrong way

`lbfgs issue_508_infeasible_gap_1em4` goes 81 -> 130 iterations, +49. Verdict
(`InfeasibleProblemDetected`) and objective (25) are unchanged, and the `exact`
leg of the same fixture goes 441 -> 245, -196. Net across the fixture is
strongly positive, but this is a measured cost of the change and is called out
here rather than left in a commit message — per
`dev-notes/trajectory-regressions-and-the-fixture-sweep.md`, which exists
because a 206 -> 812 regression once rode to release inside a commit message.
Happy to file it with an owner if reviewers want it tracked.

### #729 and #730 are trajectory-neutral

Re-swept after merging both: **zero moved legs** across all 128, on top of the
refinement work. That is the claim #729's bitwise-stability gate was written to
make, confirmed on the corpus rather than argued from the predicate. So the
17-leg accounting above is the accounting for the whole branch.

## Gates

- `cargo clippy --workspace --all-targets --features ma57` — clean, 0 errors.
- Full suite — 267 ok, 0 failed. (The `ma57` feature run exits 101 on an
  environmental `libcoinhsl.dylib` rpath doctest; the control without `ma57`
  passes 5/5.)
- Fixture sweep — both legs, accounted for above.

## Not in this PR

feral#191 (`fix/componentwise-refine-default`) changes FERAL's default
refinement stop criterion. I measured pounce against that branch by path. It
does not affect this PR — with `refine = false` none of it executes on the IPM
path — but two things are worth recording for whoever does the 0.18 bump:

1. Its new default costs pounce **+18% overall and +21% back-solve** on
   `laptime` versus feral 0.17.0 — measured, not extrapolated from its own tail
   table.
2. `solve_refined_into` / `solve_many_refined_into` change return type from
   `Result<(), _>` to `Result<RefineOutcome, _>`. pounce will not compile
   against 0.18 without adapting the call sites.

**Neither is something pounce has to accept.** `FeralConfig` is ours, and 0.18
adds `StopCriterion` to `RefineOptions`, so at the bump every pounce caller gets
an explicit criterion rather than inheriting feral's default:

- The IPM keeps `refine = false`. Nothing changes.
- pounce-convex / pounce-rs / pounce-py get
  `StopCriterion::RelativeResidual(<their own tolerance>)` instead of the new
  default. Measured at `1e-10` that is **22.8 s against the new default's
  69.4 s** on `laptime`, and 58.6 s for feral 0.17.0's current rule — so those
  callers come out ahead of where they are today, not behind, while gaining the
  componentwise accuracy fix feral#191 is actually about.

That also retires `FeralConfig::refine_target`, which documents itself as a
placeholder for exactly this knob ("it goes away when feral#190 lands"), along
with the duplicated initial solve its pre-check pays on a miss.

The one number that stays true regardless: arm C is still 1.6x slower than not
entering the backend loop at all, which is why the IPM's `refine = false` is
right independent of any of this.

Full measurement is posted at
https://github.com/jkitchin/feral/pull/191#issuecomment-5372497342


jkitchin:
@ srikanth-gm I am pretty sure this fixes the issues you reported. I am running some final benchmarks to check for overall performance, and hope to merge this tomorrow.

Issue #655, opened 2026-08-17
Sigma = z/s overflows to inf on a SolveSucceeded when the slack goes subnormal


jkitchin:
Found while closing out #653's second open question. Independent of
crossover — it reproduces with crossover off.

## Reproduction

    mu_strategy    = adaptive
    tol            = 1e-306
    compl_inf_tol  = 1e-306
    mu_min         = 5e-324

on any problem with an active bound carrying a nonzero multiplier. The solve
returns **`SolveSucceeded`**, and at the `on_converged` hook:

    mu         = 9.0909e-308
    min slack  = 2.0202e-308      (subnormal)
    max Sigma  = inf

Neighbouring points, same fixture, for scale:

| `tol` | status | `mu` | min slack | `max Sigma` |
|---|---|---|---|---|
| `1e-8` (default) | SolveSucceeded | `2.51e-09`  | `5.57e-10`  | `8.08e+09`  |
| `1e-300`         | SolveSucceeded | `9.09e-302` | `2.02e-302` | `2.23e+302` |
| `1e-304`         | SolveSucceeded | `9.09e-306` | `2.02e-306` | `2.23e+306` |
| `1e-306`         | SolveSucceeded | `9.09e-308` | `2.02e-308` | **`inf`**   |
| `1e-308`         | SearchDirectionBecomesTooSmall | — | — | — |

## Why the existing floor does not catch it

`CalculateSafeSlack` is doing its job and still overflows. Its trigger
threshold is `s_min = eps * min(1, mu)`; at `mu = 9.09e-308` that product
underflows to `0` and the code substitutes `f64::MIN_POSITIVE` (`2.2e-308`).
The raw slack `2.02e-308` is below that, so the floor **does** fire — and
pins the slack to `2.2e-308`. Then `Sigma = z/s = 4.5 / 2.2e-308 = 2.02e308`,
which is past `f64::MAX = 1.8e308`.

So `f64::MIN_POSITIVE` is the wrong floor for the purpose. The quantity that
has to stay finite is `z/s`, and the floor that guarantees it is
`s >= z / f64::MAX`, not the smallest representable positive double.

## What this corrects in #653

#653 listed "`mu == 0` exactly" as the one corner its measurement did not
bound, reasoning that `CalculateSafeSlack` would then pin to
`f64::MIN_POSITIVE` and `Sigma` would overflow. Both halves of that turned
out slightly wrong:

[...]

Issue #654, opened 2026-08-17
Crossover under a nonzero bound_relax_factor loosens the barrier pin, degrading a downstream reduced Hessian


jkitchin:
Split out of #653, which measured it. Same root cause as #646; #647 fixed
only the reporting half of that, and this is the other half.

## What happens

`bound_relax_factor` (default `1e-8`) widens every bound by `delta` before
the solve. Crossover then parks the iterate exactly on the **declared**
bound, which is a full `delta` **inside** the live relaxed one. The barrier
therefore sees a slack of exactly `delta` at every active bound, where an
interior iterate would have carried `mu/z`.

The barrier diagonal is `Sigma = z/s`, so:

    no crossover     Sigma = z / (mu/z) = z^2 / mu
    crossover        Sigma = z / delta

and crossover **loosens** the pin whenever `z*delta/mu > 1` — which is the
normal case, since `delta` is capped at `constr_viol_tol` and `mu` ends near
`tol/(barrier_tol_factor+1)`.

That matters because `Sigma` is the stiffness with which the barrier holds a
bounded variable, and any reduced Hessian read off the held KKT factor
carries a residual error of exactly `O(1/Sigma)` — the leftover of that pin
being finite. A looser pin is a less accurate covariance.

## Measured

`min 1/2 x'Qx - q'x` over `(a, b, w)`, `a` and `b` held by pin rows, `w`
capped by a bound binding with multiplier `z`. `Q` non-diagonal, so the
reduced Hessian over the pins differs by `O(1)` between the bound-pinned
answer (`Q_ab`) and the free one. Error below is `|H_R - Q_ab|_inf`.

At `z = 4.5`, `bound_relax_factor = 1e-8`:

| | `Sigma` | `H_R` error |
|---|---|---|
| `crossover=no`  | `8.08e+09` | `4.95e-10` |
| `crossover=yes` | `4.50e+08` | `8.89e-09` (**18x worse**) |

The degradation tracks the bound multiplier, exactly as `z*delta/mu`
predicts, and the downstream `H_R` error ratio tracks the `Sigma` ratio 1:1:

| `z` | `Sigma` no-xo | `Sigma` xo | ratio | `H_R` error no-xo -> xo | ratio |
|---|---|---|---|---|---|
| `4.5`    | `8.08e9`  | `4.50e8` | `18.0`  | `4.95e-10` -> `8.89e-9`  | `18.0x`  |
[...]

Issue #653, opened 2026-08-17
Crossover + bound_relax_factor=0: the barrier diagonal Sigma = V/S grows ~690x on the sensitivity path (measured; not the blow-up it was assumed to be)


jkitchin:
Follow-up from #612 / #646 (PR #647), which raised this and deliberately did
not resolve it.

## Summary, corrected by measurement

The open question was phrased as "what does the barrier diagonal
`Sigma = V/S` do at exact-zero slack" — the worry being that crossover puts
the iterate *on* a bound, and `on_converged`'s sensitivity hook then builds
`Sigma` by dividing the multiplier by that slack.

**It does not blow up.** I probed all eight combinations and `Sigma` is
finite in every one. The failure mode that motivated the question does not
occur, and PR #647 chose the reporting-side fix partly to avoid it — that
choice is still right, but not for the reason stated.

What the probe *did* find is a conditioning change worth its own attention.

## Measured

`min (x0 - a)^2 + x1^2  s.t.  x0 <= 1`, `a = 2` gives a strictly
complementary optimum at `(1, 0)` with multiplier `2`; `a = 1` gives the
weakly active one (multiplier `0`) that crossover exists for. Values read in
`on_converged` off `curr_sigma_x()` / `curr_slack_x_u()`.

**Strictly complementary (multiplier 2):**

| crossover | `bound_relax_factor` | min slack | `max Sigma_x` |
|---|---|---|---|
| no  | `1e-8` | `1.25e-09` | `1.60e+09` |
| yes | `1e-8` | `1.00e-08` | `2.00e+08` |
| no  | `0`    | `1.25e-09` | `1.60e+09` |
| yes | `0`    | `1.82e-12` | **`1.10e+12`** |

**Weakly active (multiplier 0):**

| crossover | `bound_relax_factor` | min slack | `max Sigma_x` |
|---|---|---|---|
| no  | `1e-8` | `5.87e-05` | `2.00e+00` |
| yes | `1e-8` | `1.00e-08` | `0.00e+00` |
| no  | `0`    | `5.87e-05` | `2.00e+00` |
| yes | `0`    | `1.82e-12` | `0.00e+00` |

`mu` at termination was `2.5e-09` / `9.1e-10` — never `0`.

## The three findings

1. **No exact-zero slack.** The crossed-over slack bottoms out at `1.82e-12`,
   the residual of the QP step plus line search, not `0`. So the division
[...]

PR #738, opened 2026-08-22
Cap the barrier diagonal at what its constraint rows can be seen against

### LINKED ISSUE #737: A declared parameter pinned to exactly a bound of its variable makes the whole sensitivity column read zero
A declared parameter pinned to a value that equals a bound of the variable it pins makes the whole derivative column through that variable read zero, including a derivative that is 1 by construction. `estimate()` returns the baseline value for that variable. sIPOPT solves the same model and perturbation correctly.

## Symptom

The CSTR of notebook 36 has `zc_init: zc[0] == zc0`, with `zc0` declared and `zc` declared with bounds `(0, 1)`. Sweeping the baseline `zc0` with `zt0` held at 0.70, every solve optimal, `d zt[0]/d zt0 = 1.00000` in every row:

| `zc0` | `d zc[0]/d zc0` | `d v1[0]/d zc0` | `gradient()` warning |
|---|---|---|---|
| 0.00 | **0.00000** | -0.00000 | degenerate base point |
| 0.01 | 1.00000 | 1.05472 | none |
| 0.62 | 1.00000 | -0.03578 | none |
| 0.99 | 1.00000 | -0.08145 | none |
| 1.00 | **0.00000** | -0.00000 | **none** |

`d zc[0]/d zc0` is 1 by construction. It reads zero exactly at the two values where `zc0` equals a bound of `zc`, and at `zc0 = 1.00` nothing warns.

`estimate()` follows the column. From `(zc0, zt0) = (1.00, 0.70)` to `(0.79, 0.57)`:

| mode | `zc[0]` | `zt[0]` |
|---|---|---|
| linear | 1.000000 | 0.570000 |
| fix_relax | 1.000000 | 0.569903 |
| path | 1.000000 | 0.570022 |
| re-solve | 0.790000 | 0.570000 |

`zt[0]` is right in every mode. From `(0.99, 0.70)` to the same target, `zc[0]` is `0.790000` exactly.

## sIPOPT gets this right

Same model, same perturbation, through `pyomo.contrib.sensitivity_toolbox` and `ipopt_sens`, reading the `sens_sol_state_1` suffix:

| baseline | pounce | sIPOPT | re-solve |
|---|---|---|---|
| `zc0 = 0.62`, inside the bounds | 0.790000 | 0.790000 | 0.790000 |
| `zc0 = 1.00`, on `zc`'s upper bound | **1.000000** | **0.790000** | 0.790000 |

Both agree at the control, so this is not a harness artefact.

It is not solving the case by knowing about it. `SensStdStepCalc.cpp` has no special handling for `sens_state` variables or for variables fixed by the parameter constraints, and its `BoundCheck` collects two categories uniformly, `x` outside a bound and `z_L` or `z_U` below `-bound_eps`.

What differs is the converged point:

| | `zc[0]` | slack to the upper bound | multiplier |
|---|---|---|---|
| pounce | 1.000000000002 | -1.819e-12 | `sigma` 1.888e27 |
| ipopt | 1.000000000000 | +0.000e+00 | `z_U` = **-0.2506** |

IPOPT lands exactly on the bound with a negative upper-bound multiplier, which says the bound is not what holds the variable. sIPOPT's second category catches that and relaxes it, which is equation 18 of Pirnay, Lopez-Negrete and Biegler. pounce's multiplier is large and positive, so nothing in the release logic fires.

## The plain step violates a linear equality

`zc_init` is linear, so the parametric step has to satisfy it exactly. It does not:

| | `d zc[0]` | `d zc0_var` | requested | residual of `zc[0] == zc0` |
|---|---|---|---|---|
| bounds on `zc[0]` | +3.7e-19 | **-0.105** | -0.21 | **+0.105** |
| bounds removed | -0.210000 | -0.210000 | -0.21 | 0 |

The parameter's own Var moves half the requested delta and the missing half is the residual. An even split is what a solve returns when two rows have become indistinguishable to it.

## `sigma` is what separates working from broken

`sigma` follows `mu / s²` as `zc0` approaches the bound, and every value short of it is correct. The failure is not gradual. It appears only when the pinned value equals the bound exactly:

| case | where the solve lands | `sigma` | `d zc[0]/d zc0` |
|---|---|---|---|
| `zc0 = 0.9999999` | slack +1.0e-07 | 2.506e5 | 1.000000 |
| `zc0 = 0.00` | slack **+2.3e-15** | 3.356e23 | **0.00000** |
| `zc0 = 1.00` | slack -1.819e-12 | 1.888e27 | **0.00000** |
| two-variable model, pinned to its bound, four scales | slack **exactly 0.0** | 2.0e9 to 7.1e14 | 1.000000 |

The slack's sign is not the discriminator, since `zc0 = 0.00` fails with positive slack. `sigma` is. Everything at or below 7.1e14 is correct, everything at or above 3.4e23 is zero, and I sampled nothing between. That gap straddles double precision. Against an equality row whose coefficient is 1.0, a diagonal of 1e23 leaves the equality unrepresentable, which is the condition under which the delta gets split rather than delivered.

A two-variable model with the same structure lands exactly on the bound and keeps `sigma` under 1e15, which is why it does not reproduce.

## Removing the bound fixes it, with an identical solve

Clearing the bounds on the single Var data `zc[0]` before the solve, changing nothing else:

| | `zc[0]` solved | `sigma` | `d zc[0]/d zc0` | `estimate` `zc[0]` |
|---|---|---|---|---|
| bounds on `zc[0]` | 1.000000000002 | 1.888e27 | 0.00000 | 1.000000 |
| bounds removed | 1.000000000000 | 0.0 | 1.00000 | 0.790000 |

Identical to twelve digits. Only the sensitivity system changes.

## No mode rescues it, and `fix_relax` reports success

`estimate_report(mode="fix_relax", predictor_iter=32)` returns `refine_stop="settled"` having crossed and repaired over 200 coordinates, `zc[43]` through `zc[99]`, most of `v1` and most of `v2`. `zc[0]` is not among them.

That is circular. The refinement reads releases off the step's endpoint, the step in `zc[0]` is zero because the barrier holds it, so the endpoint stays on the bound and no release is proposed. The one coordinate whose bound has to leave the active set is the one it cannot see moving.

## Where a fix has to go

I tried releasing bounds with no slack inside `Solver::parametric_step` and it never runs. `estimate()` defaults to `degeneracy="directional"` and takes `parametric_step_directional`, and `gradient()` goes through the solve-time callback in `convenience.rs`. A guard covering all of them belongs at the barrier diagonal every sensitivity path shares, `barrier_sigma_x`.

It also has to key on `sigma` rather than on slack. A slack-based rule would not fire at `zc0 = 0.00`, where the slack is positive and the derivative is still zero.

## Reproducing

Notebook 36's CSTR at `zc0 = 1.00`, `zt0 = 0.70`, perturbed to `(0.79, 0.57)`. Needs `pyomo_cvp`. Every solve terminates optimal and `kkt_perturbations` is at most 7.1e-11, so no inertia correction is involved.

Versions: pyomo-pounce 0.10.0 editable from this repo at a135342f, POUNCE 0.9.0 with FERAL, Pyomo 6.10.1, Windows.

jkitchin:
Fixes #737

> **Updated twice.** Once after merging `main` (#733, the sensitivity corrector) —
> that merge is clean textually and was wrong semantically, and the third commit
> fixes it. Once after @devin-griff's review, which corrected what actually
> bounds the corrector's addend; the fourth commit carries that. All
> measurements below have been re-taken on the merged head.

## Problem

A declared parameter pinned to a value that *equals* a bound of the variable it
pins made the whole derivative column through that variable read zero —
including a derivative that is `1` by construction.

On notebook 36's CSTR, `zc_init: zc[0] == zc0` with `zc` bounded `(0, 1)`, `zt0`
held at `0.70`. Measured on this branch against its parent (release build):

| `zc0` | `d zc[0]/d zc0` before | after | `d v1[0]/d zc0` before | after |
|---|---|---|---|---|
| 0.00 | **0.00000** | 1.00000 | **-0.00000** | 1.04145 |
| 0.01 | 1.00000 | 1.00000 | 1.05472 | 1.05472 |
| 0.62 | 1.00000 | 1.00000 | -0.03578 | -0.03578 |
| 0.99 | 1.00000 | 1.00000 | -0.08145 | -0.08145 |
| 1.00 | **0.00000** | 1.00000 | **-0.00000** | -0.06934 |

`estimate()` from `(zc0, zt0) = (1.00, 0.70)` to `(0.79, 0.57)`:

| | `zc[0]` before | after |
|---|---|---|
| `mode="linear"` | 1.000000 | 0.790000 |
| `mode="fix_relax"` | 1.000000 | 0.790000 |
| `mode="path"` | 1.000000 | 0.790000 |
| re-solve (oracle) | 0.790000 | 0.790000 |

sIPOPT is the other oracle and solves the same model and perturbation
correctly. At `zc0 = 1.00` nothing warned; `fix_relax` reported success having
repaired 1255 other coordinates and not this one.

## Cause

Such a point is held by a bound and an equality at once, so the force holding it
has no unique split between them. The solve lands with a bound multiplier far
larger than the geometry needs over a slack near roundoff, and `Σ = z/s` is the
product of both: `1.9e27` on the CSTR, against Jacobian entries of `1`.

Eliminating the variable through that diagonal leaves each of its constraint
rows holding `a²/Σ ≈ 1e-27` — the whole of what the row still knows about it,
and far under the roundoff of the row it lands in. The factorization then has a
row it cannot pivot on, and the step comes back as the even split of two rows
that are no longer distinguishable: the parameter's own Var moving half the
requested delta, and the linear equality carrying the other half as its
residual. The fixture shows exactly that pre-fix — `[0, -0.105, 0]` for a
requested `-0.21`.

## Fix

`Σ` is capped, at construction and on every sensitivity path, at the stiffness
the variable's own constraint rows can still be seen against:
`Σ_i ≤ a_i²/eps`, with `a_i` the largest of that variable's Jacobian
coefficients and `SIGMA_PIN_HEADROOM = 64` backing off from the edge. The
quadratic form is the scale-invariant one — a change of variables sends
`Σ → Σ/c²` and `a → a/c` — and both quantities are read in the solver's own
scaled space, which is the space the factor lives in.

It is a **ceiling, not a release**. A capped bound still holds its variable to
within roundoff of the scale its constraint coefficients set, so a bound that
genuinely holds a variable keeps holding it. Zeroing the entry instead — the
first thing tried — answers a different question: it lets a genuinely held
variable off its bound entirely. A variable in *no* constraint row is not capped
at all: there is no row for its diagonal to swamp, and that is the #653 / #654
case, where the stiffness *is* the accuracy being measured. The rule is the one
#655's slack floor already makes one step further down — a pin the matrix cannot
represent is not a stiffer pin.

**Where the ceiling sits.** The theory puts the edge at `a²/Σ = eps`, i.e.
`4.5e15` for a unit coefficient. Measured, the failure is much further out: on
the fixture here `Σ = 3.6e22` still returns the exact step and `6.9e27` returns
none of it, and the issue's own bracket runs `7.1e14` correct to `3.4e23` zero.
So the theoretical edge is conservative against both brackets, and the headroom
backs off a further 64 to `7.0e13`. That is deliberate: the cost of backing off
is a residual leak in the pin of `1.4e-14` rather than `2e-16` — an order under
the roundoff a back-solve on a real model carries — and it is only ever paid by
an entry that was going to be capped. The cost of the other direction is the
defect surviving between the ceiling and the failure.

Applied at the frame choice itself, so `barrier_sigma_x` / `barrier_sigma_s`,
`sigma_override` and the released rebuild all read one capped pair.

## The merge with #733

`main` gained the sensitivity corrector while this branch was open. The merge
has no textual conflict and is still incomplete, which is worth stating plainly.

#733 turned `released_sigma_x` into `active_set_sigma_x(released, pinned)`,
which both rebuilds released entries **and adds** to entries a step brings onto
a bound. Git matched the ceiling to the released rebuild, where its context
lived. The pinned additions happen after it, and came out uncapped. The ceiling
is a property of the entry rather than of where the entry came from, so it now
applies to the sum: two contributions that are each representable can still
swamp the row together.

**How far that addend can actually reach** — corrected by @devin-griff's
review, and the reason the fourth commit exists. I wrote that the addend is
unbounded above because `pinned_rows` floors its slack by nothing beyond `> 0`.
That is true of `pinned_rows` and false of the code: `correct_step` clamps the
iterate to `margin = 1e-10 * (1 + |base_i|)` inside each bound *before*
measuring it. So the addend is bounded at `mu / margin²` — `2.5e10` at a
converged `mu` of `1e-9` on a variable of order one, three orders under the
`7.0e13` a unit coefficient allows. The ceiling binds on this path only where
the coefficient is small enough to bring it down to meet the addend, under
`sqrt(mu · eps · SIGMA_PIN_HEADROOM) / margin`, about `2e-2` there.

The clamp stays, restated as what it is: **prophylaxis, not a repair for a
reachable failure.** What makes the corrector's door narrow is a `1e-10` margin
in `corrector.rs` that no rule keeps in step with this ceiling. Holding the
ceiling here costs nothing and does not depend on that margin staying put.

## Blast radius

Everything that reads the held factor: `gradient()`'s solve-time callback,
`estimate()`'s directional and fix-relax modes, the plain and released
parametric steps, `compute_reduced_hessian`, `classify_activity()`,
`covariance()`, and the diagonal `corrector_iter` assembles. All of them get the
same capped diagonal, so they cannot disagree about it.

**No trajectory moves.** Nothing in the interior-point iteration reads
`solve_with_sigma` — the cap lives entirely in the sensitivity system — so
`scripts/sweep-fixtures.sh` does not apply to this change.

An entry under its ceiling is left as the object the calculated quantities
already cache (`cap_sigma` returns `None` rather than an equal copy), so an
ordinary solve does not pay a re-factorization for a bit-identical diagonal.
On the reported model, the baselines away from a bound (`zc0` = 0.01, 0.62,
0.99) are unchanged to every digit printed.

## Tests

`crates/pounce-sensitivity/tests/issue_737_pinned_to_bound_sigma.rs` — a
three-variable fixture reproducing the reported signature exactly (interior
`Σ = 6.88e27`, crossed over `1.13e25`, `d x/d p = 0`, and the parameter's Var at
half the delta), asserted through both the plain and the directional step.

**These fail on the parent commit for the stated reason.** Run there with the
source file reverted:

```
d x = 8.550615924378022e-20, want -2.1e-1
(whole step was [8.550615924378022e-20, -0.105, 8.5506159239085e-20])
```

— the zero derivative and the exact half-step, which is the mechanism above. The
fourth test in the file is the control: a point that never needed the ceiling,
which passes on the parent too and must keep passing here.

Nine unit tests on the ceiling itself: the representability identity,
invariance to rescaling the variable, the refusal to report a ceiling that would
release rather than loosen, that a diagonal under its ceiling is left as the
cached object, three on `pinned_entry`, and one carrying the two numbers that
make the corrector's door narrow — `corrector.rs`'s margin and `pinned_rows`'
form of the addend, both of which live elsewhere, so a retune of either
surfaces there.

**Deliberately not pinned**, two things:

1. **The corrector's path has no end-to-end fixture that bites here**, and after
   @devin-griff's four attempts on top of my one, that now looks structural
   rather than lucky. The clamp above floors `s_end`, so exceeding the ceiling
   needs a coefficient small enough that the constraint barely sees the
   variable. Driving it at `1e-3` and `1e-4` does reach it: two of twelve cases
   move — an iteration count of 1 against 12 at identical residuals, and a
   residual differing in the fifth digit. That is real end-to-end evidence the
   wiring is live, and far too small to assert on. The `pinned_entry` unit tests
   pin the arithmetic; the wiring stays unpinned and the code comment says so.
   Also worth recording: the existing `corrector.rs` fixture could not have
   bitten either way — its `x1` is in no constraint row, which is the uncapped
   case by construction.
2. **The `s`-block half of the cap.** The argument is identical to the `x` block
   — in the space `curr_sigma_s` lives in, an inequality's coupling to its own
   slack is exactly `1` — but reaching it needs a model where an inequality's
   slack is itself pinned that hard, and the fixture here does not produce one.

## Re-measured after the merge

#733 restructured notebook 36 (646 lines) and reworked `estimate()`, so every
number above was re-taken on the merged head rather than carried over. `build()`
is untouched, and:

- the gradient sweep is **identical in all five rows**, to every digit printed;
- `estimate()` to `(0.79, 0.57)` returns `zc[0] = 0.790000` in all three modes,
  at `corrector_iter = 0` and `corrector_iter = 8` alike.

@devin-griff reproduced both independently on the same model.

`estimate(max_iter=)` is now `predictor_iter=` (that rename is #733's, from
gh#732).

Verified: `cargo test --workspace --exclude pounce-hsl`; `pytest
pyomo-pounce/tests` (435 passed, 4 skipped) and `pytest python/tests` (953
passed, 36 skipped — the one failure is `pounce.jax` with no jax installed)
against a `maturin develop --release` build of the merge head and a release CLI
built from it; `cargo fmt --all --check`; `cargo clippy -p pounce-sensitivity
--all-targets -D clippy::correctness -D clippy::suspicious`;
`scripts/check-release-consistency.sh`; `scripts/check-docs-consistency.sh`.
The fourth commit is comments and one unit test, re-verified against the Rust
workspace, fmt, clippy and both guards.

Docs: `docs/src/sensitivity.md` gains "A Param pinned to exactly a bound", which
also names the corrector's diagonal; `docs/src/crossover.md`'s floor discussion
gains the ceiling at the other end and the reason it is not symmetric with them.

---

      since moved


devin-griff:
Verified on notebook 36's CSTR, the model #737 was filed against, rather than on the fixture.

## It fixes it

Your gradient sweep reproduces to every digit. `d zc[0]/d zc0` goes 0.00000 to 1.00000 at both `zc0 = 0.00` and `1.00`, `d v1[0]/d zc0` lands on 1.04145 and -0.06934, and the three interior rows are unchanged at 1.05472, -0.03578 and -0.08145.

Against both oracles, `estimate()` from `(1.00, 0.70)` to `(0.79, 0.57)` returns `zc[0] = 0.790000` in all three modes and at `corrector_iter = 8`, matching the re-solve and sIPOPT. At the interior control all three still agree, so nothing moved where it should not have. `fix_relax` now reports `alpha = 4.762` with `crossed` empty, where before it claimed `settled` having repaired over 200 coordinates and not that one.

The ceiling is a better answer than the release I proposed in the issue, and for the reason you give: releasing lets a genuinely held variable off its bound entirely, and this only removes the part the matrix cannot represent.

## The corrector's unpinned wiring

I tried to close it and could not, and I think the door is shut for a structural reason rather than by luck.

Four attempts. The `corrector.rs` fixture across its whole perturbation range, byte-identical between this branch and main. The perturbation tuned to where the crossing coordinate reaches its bound exactly, identical. Looser solve tolerances, which move it the wrong way: `s_end` grows in proportion to `mu`, so `mu / s_end²` behaves like `1 / mu` and shrinks, and by `tol = 1e-4` the pin stops firing at all. And a new fixture putting the pinned variable in a constraint with coefficient `C`, so the ceiling `C²/eps` comes down to meet the addend instead.

That last one gets closest. At `C = 1e-3` and `1e-4` the perturbation drives the variable onto its bound with `s_end` at 1e-6 and 1e-8, `pinned` fires, and on main the correction stops on iteration 1 with the ratio at exactly 1.000000. Capped, those rows are unchanged. Across twelve cases two move: at `C = 1e-3, dp = 0.6` the iteration count goes 1 to 12 with identical residuals, and at `C = 1e-4, dp = 0.500001` the residual differs in the fifth significant digit.

That pair is worth something. It is end-to-end evidence the wiring is live and the ceiling reaches the corrector's addend, which the `pinned_entry` unit tests assert at the arithmetic level and nothing showed above it. It is far too small to assert on, so I would not ship it either.

The reason it cannot get bigger is the corrector's own initial clamp, which puts any coordinate outside a bound back at `hi - 1e-10 * (1 + |base|)`. That floors `s_end` near 1e-10 and so bounds the addend near `mu * 1e20`. Exceeding the ceiling from there needs a coefficient small enough that the constraint barely sees the variable, and by then the correction has nothing to gain on that coordinate anyway.

Also worth saying: the existing `corrector.rs` fixture could never have bitten. Its `x1` appears in no constraint row, which is the uncapped case by construction.

## With #739

No textual conflict, and I merged the two and ran both suites rather than trusting that, since a clean merge being wrong is the thing you flagged about #733. 19 rust suites and 438 pyomo-pounce tests pass on the merge, including the three new option tests.

jkitchin:
You are right about the clamp, and it means my comment was wrong. Pushed `a8836ba`.

`correct_step` clamps the iterate to `1e-10 * (1 + |base_i|)` inside each bound *before* `pinned_rows` measures it, so "nothing bounds that slack from below beyond `> 0`" is true of `pinned_rows` and false of the code. The addend is bounded at `mu / margin²`, which at `mu = 1e-9` on a variable of order one is `2.5e10` — three orders under the `7.0e13` a unit coefficient allows. I checked the arithmetic against your `C` sweep and it lands where you found it: the ceiling comes down to meet the addend at `sqrt(mu · eps · SIGMA_PIN_HEADROOM) / margin`, about `1.9e-2` at that `mu`, so `1e-3` and `1e-4` are inside and `1` is not. That is why two of twelve moved and none of the unit-coefficient cases did.

Your twelve-case sweep is the part I could not get. My one attempt died before it started — the corrector stopped on iteration 1 with the residual it began with, so the fixture never discriminated and I could not tell a shut door from a badly chosen model. Bringing the ceiling down to the addend instead of pushing the addend up to the ceiling is the move I missed. It settles the question for me: the wiring is live and the ceiling does reach the corrector's addend. Recorded on the PR with your numbers, and agreed it is far too small to assert on.

So the clamp stays, restated as prophylaxis rather than a repair for a reachable failure. What keeps that door narrow is a `1e-10` margin in `corrector.rs` that nothing keeps in step with the ceiling in `algorithm_backsolver.rs`; holding the ceiling on the sum costs nothing and does not depend on that margin staying where it is. There is one new unit test carrying both numbers so a retune of either surfaces there rather than silently widening the door.

Also took your point about the existing `corrector.rs` fixture — its `x1` is in no constraint row, so it is the uncapped case by construction and could not have bitten either way. I had noticed the same thing about `ParamQp` while looking for somewhere to put a test and did not think to write it down. It is on the PR now.

Nothing needed from #739 on my side — it does not touch the diagonal, and I am happy to take your merge check for the pair rather than duplicate it. Whichever lands first, I will merge `main` and re-run rather than trusting the clean merge, for the reason this PR now has a section about.

---

devin-griff:
Checked `a8836ba`. 19 suites green, and I worked the threshold independently rather than taking it: `sqrt(mu · eps · 64) / margin` is 1.88e-2 at `mu = 1e-9` on a variable of order one, with the addend at 2.5e10. That predicts the boundary my sweep found, since `1e-2` sat on it and did not move while `1e-3` and `1e-4` did.

Agreed on keeping the clamp as prophylaxis. The reason I would keep it is the one you give, that the two numbers live in different files with nothing holding them in step, and the unit test carrying both is the right place for that to surface.

Nothing further from me on this one.

PR #739, opened 2026-08-22
Expose bound_eps and max_pdpert on the pyomo interface

### LINKED ISSUE #736: Expose bound_eps and max_pdpert on the pyomo interface
Two of the sIPOPT options wired in 6346135d are settable through the CLI and the `SensSolve` builder but not from pyomo. Expose both on the pyomo interface for completeness.

## `bound_eps`

The margin that decides whether a step has left a bound is `--sens-bound-eps` on the CLI and `sens_bound_eps` on the builder. From pyomo it is not settable. `Solver::bound_context` derives it from the solve's own bound relaxation, floored at `1e-9`, and no caller of `estimate()` can change it.

Add `bound_eps` to `estimate()` and `estimate_report()`, defaulting to `None` for the current behaviour.

It governs `mode="fix_relax"` only. `path` detects its changes from the multiplier's zero crossing and never reads a bound-violation margin, and `active_set_changes()` is path-only, so neither takes the argument. That is the same mode-dependent shape `predictor_iter` already has, which caps passes under `fix_relax` and caps applied changes under `path`.

Threading it is contained. `bound_context` takes an override, the two bounded-step functions pass it, and the Python binding carries it. Nothing else reads the margin.

## `max_pdpert`

`sens_max_pdpert` caps the inertia-correction perturbation the converged factor may carry before the sensitivity outputs are withheld. Setting it from pyomo does nothing: on a rank-deficient model whose factor carries `7.08e-11`, a cap of `1e-12` returns the step and the gradient without complaint.

Add `max_pdpert` to `estimate()`, `estimate_report()` and `gradient()`, defaulting to `None` for the current behaviour of no cap.

`EstimateReport.perturbations` already reports the values, which lets a caller judge but not refuse.

## Not needed

`bound_relax_factor` is already settable through `solve(options=...)` and reaches the solve.

devin-griff:
Fixes #736. Both margins are settable through the CLI and the `SensSolve` builder and were unreachable from pyomo.

## `bound_eps`

How far outside a bound a step has to end to count as having left it. `Solver::bound_context` derived it from the solve's own bound relaxation and no caller could change it.

`bound_context` takes an override, `parametric_step_bounded` and `parametric_step_bounded_decided` take it as a trailing `Option<Number>` and pass it, their Python bindings take `bound_eps=None`, and `estimate()` and `estimate_report()` carry it to all four call sites.

One margin decides the whole call. The `fix_relax` refinement pins against it, `estimate()` clamps against it, and `EstimateReport.crossed` and the variable ratio test report against it. Deciding it twice is what let a margin wider than the crossing come back reporting `refine_stop == "settled"` beside a coordinate 2.0 outside its bound. The margin is absolute, as the refinement's own `over > eps` is. The python side used to scale it by the coordinate, which gave a variable of order 1e4 a tolerance the refinement never gave it, so a step the refinement pinned could come back with `crossed` empty.

It is a margin on variable bounds. The refinement pins variable bounds only, so a constraint row keeps its own floor whatever the margin, and a wide margin cannot hide a row the step carries past its limit from `crossed_rows` or from `alpha`.

Unset, the report keeps its `1e-9` floor rather than taking the solve's relaxation. On a relaxed solve the relaxation is exactly how far out the solve itself left a coordinate, and naming that coordinate is what `crossed` is for.

A margin wide enough to cover the crossing pins nothing, so the step is the linear one and reaches a bound at a fraction below one. `alpha` reports that, where under `fix_relax` it is otherwise 1.0.

Only the `fix_relax` refinement reads the margin. `mode="linear"` and `mode="path"` warn and are unaffected, and `active_set_changes()` does not take it, since the path decides its changes from where a multiplier reaches zero.

**The refinement's release test has its own threshold.** `refine_step_onto_bounds` read one `eps` for three jobs: the primal violation margin, the threshold a multiplier has to go negative past before its bound is released, and the two accept guards. Exposing that number made a caller's `bound_eps` of `1e-2` silently stop every release on a model with multipliers of order `1e-3`. It now takes `release_eps` separately, which is the solve's own floored `bound_relax_factor` on every caller, and the accept guards stay on the primal margin, since they compare primal overshoot and a caller who widened the margin has said that about overshoot too. The CLI's two callers pass the same floor, so `sens_bound_eps` no longer gates release there either, which is a behaviour change: a multiplier the step drives to between `-sens_bound_eps` and the solve's margin now releases its bound.

## `max_pdpert`

Refuses rather than answering when the converged KKT factor carries an inertia correction above the value given. Every sensitivity output inverts that factor, so a perturbed one answers for a nearby problem rather than this one.

It is on every surface that inverts the factor: `gradient()`, `estimate()`, `estimate_report()`, `active_set_changes()`, `covariance()` and `information()`. The last two warned about the same perturbations already and go on warning when no cap is set.

The comparison is `pounce_sensitivity::pdpert_verdict`, returning `(refuse, worst)`. `SensOptionOverrides::pdpert_refusal` calls it for `sens_max_pdpert`, the new `Solver.pdpert_verdict` binding exposes it, and `_refuse_on_pdpert` calls that, so the two surfaces cannot drift apart on what counts as too perturbed.

Each words its own message from `worst`. The option's says the sensitivity was skipped and tells the caller to raise `sens_max_pdpert`, and neither is true of a call that raises from `estimate()`.

## Both

Rejected at or below zero, and NaN with them, which is the strict lower bound `sens_app.rs` registers each option with. A `bound_eps` of zero reinstates the roundoff pinning the floor prevents, a NaN makes every comparison against it false so the refinement pins nothing and still reports settled, and a negative `max_pdpert` refuses on every model with a message that reads as though the factor were bad.

The checks run before the factor is read, so a typo'd mode plus a tight cap names the typo.

## Breaking

`Solver::parametric_step_bounded` and `Solver::parametric_step_bounded_decided` are `pub` on a published crate and take the margin as a trailing `Option<Number>`, so an out-of-tree caller passing the old argument list stops compiling. Passing `None` reproduces the previous behaviour exactly. `boundcheck::refine_step_onto_bounds` takes `release_eps` after `eps`, and passing the same value for both reproduces the previous behaviour. The CHANGELOG says so under *Breaking (Rust API)*.

## Testing

Eleven new python tests and two rust ones, three rewritten and two extended, across `test_estimate_report.py`, `test_information.py`, `options.rs` and `boundcheck.rs`.

The stop reason and `crossed` are asserted together across four margins. The unset test names both numbers rather than comparing a value against itself. The two validation tests match the wording of the check that fired rather than the argument name the refusal message also carries. The ordering test uses a typo'd mode with a tight cap on a regularized factor. The singular-free-block model is a fixture now, since it is what forces inertia corrections into the held factor, and the cap is covered on `covariance()` and `information()` there rather than only wired. The cap's boundary is crossed from python through the binding, which is what shows the comparison is shared rather than duplicated: widening `>` to `>=` in `options.rs` alone fails it.

Every new and rewritten test was checked against the code it guards by putting the old behaviour back and watching it fail, including a rebuild with the release threshold set back to the primal margin for the split.

455 pyomo-pounce tests pass and 4 skip, every `pounce-sensitivity` suite, the CLI sens tests, `cargo fmt` clean, and clippy reports nothing in the files this branch changes. Main is merged in and both suites were re-run on the merge with the extension rebuilt.

## Also here

`a6a97e00` is your `d3dda90`, the derived-`dc` fixture, which I said I would land alongside the next change rather than spend a PR on. Authorship kept. Two adaptations: its `parametric_step_bounded` call takes the `bound_eps` argument the parent commit adds, amended into that commit rather than added as a fixup so the tree is not broken at it, and it sets `nlp_scaling_max_gradient` rather than relying on the default it reasons about.


jkitchin (commented):
## What holds up

I checked the claim that `bound_eps` only matters for `fix_relax`, since five of the seven `bound_context` call sites pass `None`. It's correct, not lazy: `step_along_path` never takes an eps (`boundcheck.rs:1101`), `correct_step` reads only `ctx.lo`/`ctx.hi` (`solver.rs:818`), and `weakly_active_bounds` reads only `x_curr`/`lo`/`hi` (`solver.rs:1299`). `refine_step_onto_bounds` is the sole consumer of `ctx.eps`. Every in-tree caller is updated, and the `regularized()` fixture reads `worst` at runtime rather than hardcoding 7.08e-11, so it won't rot across platforms.

## Should fix

**1. `bound_eps` makes `EstimateReport` contradict itself, and the file already has a test saying it can't.**

`refine_step_onto_bounds` decides "outside a bound" with `over > eps` (`boundcheck.rs:115`); when no pins and no releases are fresh it sets `stop = RefineStop::Settled` (`boundcheck.rs:840-847`). `EstimateReport.crossed` is decided separately, in Python, with a fixed `tol = 1e-9 * max(1, |reached|)` (`sens.py:1644`) that `bound_eps` does not reach.

So the new `test_bound_eps_sets_what_counts_as_leaving_a_bound` — `coupled()`, `p=-1.2`, `bound_eps=10.0`, asserting `slack.crossed != []` — should also be reporting `refine_stop == "settled"`. Twenty lines above it, `test_settled_means_nothing_is_left_outside_a_bound` runs the *same model and same target* and asserts exactly the opposite pairing:

```python
assert r.refine_stop == "settled"
assert list(r.crossed) == [], (
    f"settled but {[v.name for v in r.crossed]} is outside its bound")
```

And `_ratio_test`'s own docstring names this failure mode: *"Deciding it twice is what let the two disagree."* The new argument reintroduces it. I reasoned this from the code path rather than executing it — adding `assert slack.refine_stop == ...` to the new test settles it in one line. The fix I'd suggest is threading `bound_eps` into the tolerance `crossed` uses, so both sides keep answering with one margin.

**2. Two doc comments now say the opposite of what the code does.**

- `crates/pounce-sensitivity/src/solver.rs:641` — "What counts as outside a bound is taken from the solve **rather than from the caller**."
- `crates/pounce-py/src/solver.rs:289` — "What counts as outside a bound comes from the solve's own `bound_relax_factor` **rather than from an argument**."

Both are the paragraph a reader consults before reaching for the new argument, and the pyo3 one is the user-visible Python docstring. Neither mentions `bound_eps`.

**3. `max_pdpert` parity with the CLI is partial.** The Rust cap guards the step *and the reduced Hessian* — `cli/sens.rs:241` refuses under "reduced Hessian", and `convenience.rs:445` gates on `want_dx || want_rh`. On the pyomo side, `covariance()` and `information()` invert the same factor and today only `warnings.warn` (`sens.py:2919`, `sens.py:3295`); neither gained a cap. `active_set_changes()` runs the same predictor and inverts the same factor with no cap either — the PR body explains why it skips `bound_eps` (mode-dependent) but that reasoning doesn't apply to `max_pdpert`. Worth extending, or saying why not.

**4. Neither argument is validated, where the option surface bounds both.** `sens_bound_eps` and `sens_max_pdpert` are registered `add_lower_bounded_number_option(..., 0.0, strict, ...)` (`sens_app.rs:295-310`), so the CLI rejects what pyomo now accepts. `bound_eps=0.0` reinstates exactly the roundoff pinning the `.max(1e-9)` floor exists to prevent; `bound_eps=float("nan")` makes `over > eps` false everywhere, so `fix_relax` pins nothing and still reports `settled`; a negative `max_pdpert` always refuses, with a message that reads as though the factor were bad.

## Worth doing

**5.** `_refuse_on_pdpert` is a line-for-line Python re-implementation of `SensOptionOverrides::pdpert_refusal` (`options.rs:164`) — same `worst = max|delta|`, same `>`, same message text, in two languages with nothing tying them together. `kkt_perturbations` is already exposed through pyo3; exposing the refusal itself, or at minimum a comment pointing at `options.rs`, keeps the next editor from updating one.

**6.** `estimate()` already has `session` in hand three lines earlier but calls `_refuse_on_pdpert(_session_for(model), ...)` (`sens.py:1294`). Two lookups of the same thing in one function.

**7.** `_refuse_on_pdpert` runs *before* the `mode` and `degeneracy` validation in both `estimate` and `estimate_report`, so a typo'd mode plus a tight cap reports the cap. Argument validation should come first.

**8.** `bound_eps` is silently ignored under `mode="linear"` and `mode="path"` with no warning. The PR body calls this "the same mode-dependent shape `predictor_iter` already has," but `predictor_iter` is read by two of three modes and `bound_eps` by one. Given gh#677 — a registered option that looked wired and wasn't — a `warnings.warn` when `bound_eps is not None and mode != "fix_relax"` fits the house style.

**9.** `test_bound_eps_unset_matches_the_previous_behaviour` doesn't test its name. Unset is `bound_relax_factor.abs().max(1e-9)`, which for a default solve is `1e-8`, not the `1e-9` it compares against; and the unset path can't regress here by construction, since `None` reaches the identical `unwrap_or_else`. It asserts two nearby explicit margins agree on one model. Pinning the actual values, or asserting a `1e-3` margin *moves* the answer while unset does not, would carry weight.

**10.** `Solver::parametric_step_bounded` and `parametric_step_bounded_decided` are `pub` on a crates.io-published crate and gained a positional parameter — out-of-tree callers stop compiling. The CHANGELOG describes the new shape but files it under an additive bullet. Either mark it breaking or keep the old arity delegating to an `_eps` variant.

## Nits

- Stray double blank line in `estimate()`'s docstring (`sens.py:1253-1254`).
- The cherry-picked fixture hardcodes `100.0` for `nlp_scaling_max_gradient` and relies on it being the default, while the doc comment states it as though the fixture set it. One `set_numeric_value` call makes it self-describing — and the test's own argument is that a fixture shouldn't be able to go vacuous in silence.

Item 1 is the one I'd hold the merge on; 2-4 are cheap.

devin-griff:
All ten and both nits. Item 1 was real and I ran it rather than reasoning it, since you said one line would settle it.

## 1. Deciding it twice

Adding `assert r.refine_stop` reproduces it exactly as you described. On `coupled()` at `p=-1.2`:

```
bound_eps  refine_stop   alpha    crossed
None       settled       1        {}
1e-09      settled       1        {}
0.01       settled       1        {}
1.0        settled       1        {}
10.0       settled       0.485    {'x': 1.9986}
```

`_ratio_test`, `crossed`, `crossed_rows` and the clamp now read the margin the refinement pinned against, so one call answers with one number and the last row comes back `crossed={}`.

One thing changed my first attempt. Threading it through unconditionally broke `test_a_relaxed_solve_reports_a_bound_its_own_solve_left`: unset resolves to `bound_relax_factor`, and on a relaxed solve that is exactly how far out the solve left the coordinate, so `crossed` stopped naming the case it exists for. The report takes the caller's margin when there is one and keeps the `1e-9` floor when there is not, which confines the change to callers who pass the argument.

`alpha` stays at 0.485 in the last row and that is the honest number. A margin that wide pins nothing, so the step is the linear one and does reach a bound at 0.485. The docstring claimed 1.0 for every model under `fix_relax`, so it carries the exception now.

## 2, 4, 6, 7, 10

Both doc comments say the margin is the argument when one is given, and I checked the pyo3 one through the rebuilt extension rather than only in source. `_check_margins` rejects zero, negatives and NaN on both arguments, which is the strict lower bound `sens_app.rs` registers, and it runs before the factor is read, so a typo'd mode reports the typo. `estimate()` uses the `session` it already has. The CHANGELOG carries *Breaking (Rust API)* and says `None` reproduces the old behaviour.

## 3. The cap

`active_set_changes()`, `covariance()` and `information()` take it now, so it covers every surface that inverts the factor. The two that already warned go on warning when no cap is set.

Wiring it was the easy half. I had added it to `covariance()` and `information()` and tested neither, which is the same shape as gh#677, so the singular-free-block model is a fixture now and four tests cover it. A cap above the perturbation gets out of the way and each call then does what it does on that model for its own reasons, which is not the same thing for both: `covariance` returns the regularized matrix, `information` refuses because the free block is singular.

## 5. The duplication

`pdpert_verdict(perturbations, limit) -> (refuse, worst)` in `options.rs` is the comparison now. `pdpert_refusal` calls it, the `Solver.pdpert_verdict` binding exposes it, and `_refuse_on_pdpert` calls that rather than recomputing `max|δ|` and the `>`.

The messages stay apart. Yours says the sensitivity was *skipped* and tells the caller to raise `sens_max_pdpert`, and from `estimate()` the call raised and that is not the argument they passed, so surfacing the text would have traded a duplicated comparison for a wrong error. Returning `worst` lets each surface word its own from the same number.

The python test crosses the boundary through the binding rather than asserting a value, since that is what shows the sharing is real: changing `>` to `>=` in `options.rs` alone, with nothing touched on the python side, fails it. The rust test covers the comparison directly, including that a correction is reported signed while a cap is a magnitude.

## 8, 9

`bound_eps` under `mode="linear"` or `mode="path"` warns and changes nothing.

You were right that the unset test proves nothing, and the number is the other way round: pyomo-pounce solves with `bound_relax_factor` at 0.0, so unset resolves to the `1e-9` floor rather than `1e-8`. The test asserts that explicitly now, and asserts a `10.0` margin moves the answer while unset does not.

## Nits

Both. The fixture sets `nlp_scaling_max_gradient` rather than relying on the default it reasons about, and the derived-`dc` test still passes with it set.

## Testing

Eight new python tests and one rust one, two rewritten and one extended. 452 pyomo-pounce tests pass and 4 skip, every `pounce-sensitivity` suite, the CLI sens tests, `cargo fmt` clean.

Every one of them was checked by putting the old behaviour back and watching it fail: the report deciding the margin on its own, the refusal ahead of the argument checks, the warning removed, `_check_margins` removed, the two `_refuse_on_pdpert` calls taken out of covariance and information, `bound_eps` stopped short of the refinement, and `>` widened to `>=` in `options.rs`. That was worth doing. My first attempt at the ordering mutation left the mode check above the line I had moved, so the test passed against what I thought was the old code and proved nothing.

Main is merged in, and I re-ran both suites on the merge with the extension rebuilt rather than trusting the clean merge, since that is what you flagged about gh#733.

jkitchin:
## Code review

Scope: `git diff main...pyomo-sens-options` — 12 files, +654/-60. The wiring is thorough: every Rust and Python caller of the changed `pub fn` signatures was updated, and all six `sens.py` entry points that invert the factor got `max_pdpert`. Five issues below.

### 1. `crates/pounce-sensitivity/src/options.rs:279` — medium — duplicated `#[test]` attribute breaks `make clippy`

The new test was inserted *between* the existing doc comment and its `#[test]`:

```rust
    /// The cap compares against the largest perturbation of the four,
    /// and only fires when it is actually exceeded.
    #[test]
    /// The comparison two surfaces share, on its own. ...
    #[test]
    fn the_verdict_reads_the_largest_correction_by_magnitude() {
```

rustc emits `warning: duplicated attribute [duplicate_macro_attributes]`. `Makefile:116` runs `cargo clippy --workspace --all-targets -- -D warnings`, so `make clippy` now fails. ci.yml only denies `clippy::correctness`/`suspicious`, so CI stays green and this bites locally / at release time.

Separately, `the_perturbation_cap_fires_only_above_the_limit` has silently lost its doc comment to the new test, which now carries a description that does not match it.

### 2. `pyomo-pounce/pyomo_pounce/sens.py:1942` (and `:1954`) — medium — `bound_eps` widens the constraint-row tolerances, but the refinement it comes from only governs variable bounds

`eps` (line 1909) is threaded into the *row* ratio test and into `gtol` for `crossed_rows`. But the margin reaches Rust via `Solver::bound_context` -> `refine_step_onto_bounds`, whose `lo`/`hi` are `expand_bounds(n_x, ...)` over the **x block only** (`solver.rs:1338-1344`; `BoundContext.x_curr = state.x[..n_x]`), and whose multiplier set is `z_l`/`z_u` only — slack/row bounds are never pinned.

Scenario: `estimate_report(m, p, mode="fix_relax", bound_eps=10.0)` on a model where the step predicts an inequality row 5.0 outside `g_u`. `gtol = 10.0 * max(1, |g_pred|) >= 10`, so the row is dropped from `crossed_rows`; and in `_ratio_test` a row sitting on its bound is scored only via `crosses` (`interior` is false there), so a wide `eps` also removes it from `alpha`. A genuine constraint violation the margin has no authority over disappears from the report.

The row tolerances should stay at the 1e-9 floor (or the solve's own margin), not take `bound_eps`.

### 3. `crates/pounce-sensitivity/src/solver.rs:1381` — medium — `BoundContext.eps` is overloaded, so a user-supplied primal margin also retunes the dual bound-release test

`refine_step_onto_bounds` uses the same `eps` for three different jobs:

- the primal violation margin (`bound_violations`, `over > eps`),
- the **multiplier-negativity threshold** in `releasable` (`boundcheck.rs:724`, `.filter(|&(_, v)| v < -eps)`),
- the accept / "worse than plain" guards (`boundcheck.rs:889`, `:967`).

Previously `eps` was always the solve's `bound_relax_factor` floored at 1e-9, so the dual test was effectively "multiplier went negative". Now a caller who passes a plausible primal margin — `bound_eps=1e-2` on a model with `x ~ 1` and bound multipliers of order 1e-3 — silently disables bound *release* entirely: a bound whose multiplier the step drives to -5e-3 is no longer released, and the refinement pins it at its bound, returning a different (wrong) active set.

Nothing in the docstring, CHANGELOG, or the docs page hints at this; all describe `bound_eps` purely as "how far outside a bound a step has to end". Either split the release threshold off `ctx.eps`, or document and clamp the override.

### 4. `pyomo-pounce/pyomo_pounce/sens.py:1947` — low — the Python margin is relative, the Rust one absolute

Rust compares `over > eps` (absolute, `boundcheck.rs:115`); Python computes `tol = eps * np.maximum(1.0, np.abs(x_new))`. On a coordinate with `|x| ~ 1e4` and a `fix_relax` run that stopped at `iteration_limit` or `worse_than_plain`, an overshoot of 1e-7 exceeds the Rust margin (1e-9) but sits under `tol = 1e-5` — so `crossed` is empty while `refine_stop` says the refinement did not settle. That is the same class of disagreement this PR set out to eliminate, relocated to large-magnitude coordinates.

Pre-existing (the old code also used `1e-9 * max(1, |x|)`), but the PR's stated invariant and `test_bound_eps_decides_crossed_and_the_stop_reason_together` only hold because `coupled()` has O(1) variables.

### 5. `pyomo-pounce/pyomo_pounce/sens.py:2043` — low — `active_set_changes` reads the factor before validating `degeneracy`

`estimate` (1348-1356) and `estimate_report` (1842-1850) both validate `mode`/`degeneracy` first and only then call `_check_margins` / `_refuse_on_pdpert` — which is what `test_arguments_are_checked_before_the_factor_is` asserts. `active_set_changes` does the opposite: lines 2043-2044 run before the `degeneracy not in (...)` check at 2046.

So `active_set_changes(m, p, degeneracy="one-sided", max_pdpert=<tight>)` raises the perturbation refusal instead of naming the typo, contradicting "the checks run before the factor is read, so a typo'd mode plus a tight cap names the typo."

### Checked and clean

- All Rust and Python callers of `parametric_step_bounded` / `_decided` updated; positional argument order in `sens.py` matches the new pyo3 signatures. No `.pyi` stubs to update.
- `correct_step` and `parametric_step_directional` pass `bound_context(None)` but never read `ctx.eps` (only `lo`/`hi`), so not threading `bound_eps` there is correct.
- `_check_margins` rejects `0.0`, negatives and NaN — matches the strict `add_lower_bounded_number_option(..., 0.0, true, ...)` registration for both `sens_bound_eps` and `sens_max_pdpert` (`sens_app.rs:294-310`).
- `max_pdpert` is on all six public entry points that invert the factor; no `sens.py` public function was missed.
- `pdpert_verdict` is genuinely shared (Rust `SensOptionOverrides::pdpert_refusal` -> `pdpert_verdict` <- `PySolver::pdpert_verdict` -> `_refuse_on_pdpert`), so the strict-`>` boundary cannot drift.

devin-griff:
All five.

## 1. The duplicated attribute

Moved the new test below the one whose doc comment it had taken, so each carries its own. `cargo clippy -p pounce-sensitivity -p pounce-cli -p pounce-py --all-targets -- -D warnings` is clean. `make clippy` itself fails on main for nine `collapsible_if` in `pounce-studio-core`, which this branch does not touch, so I ran it per crate rather than report a green I did not get.

## 2. The row tolerances

Rows keep the `1e-9` floor whatever the caller's margin. `_ratio_test` takes the tolerance it is to agree with rather than a scale factor, so the variable call gets the caller's margin and the row call gets the rows' own, and `crossed` and `crossed_rows` are filled from the same two numbers. `test_bound_eps_leaves_constraint_rows_alone` runs `with_row` at a margin of ten and asserts the row is still in `crossed_rows`, still reached at `alpha = 0.5`, and still `first_kind == "constraint"`. Widening the row tolerance back to the margin fails it.

## 3. The overloaded `eps`

Split, as you offered. `refine_step_onto_bounds` takes `release_eps` after `eps`, the release test reads it, and the two accept guards stay on the primal margin. `BoundContext` carries it as the solve's own floored `bound_relax_factor` whatever `bound_eps` is, and the two CLI callers read the same number from the options, so `sens_bound_eps` no longer gates release either. That is a behaviour change on the CLI: a multiplier the step drives to between `-sens_bound_eps` and the solve's margin now releases its bound, where before it was pinned. The CHANGELOG says so, and both docstrings and the docs page now say that a bound is released when the step drives its multiplier negative past the solve's own margin, whatever `bound_eps` is.

The guards staying on the primal margin has a consequence worth stating. On the two-multiplier fixture, a primal margin of ten with the release threshold at its floor releases both bounds and keeps the batch, where the sibling test at `1e-9` backs off to one. Five below a bound is inside a margin of ten, so the batch stands. That is the guard reading the one number the caller changed, and the unit test asserts `[2, 3]` and says why. The other direction is in the same test: a release threshold of ten with the primal margin at its floor releases nothing.

The python twin is `test_a_wide_primal_margin_does_not_stop_a_release`. On `coupled()` at `-1.2`, `parametric_step_bounded` at `bound_eps=10.0` returns two released rows and no pin, where at `1e-9` it pins the crossing and releases one. Before the split a margin of ten returned nothing at all.

One existing test moved because of this. `test_bound_eps_sets_what_counts_as_leaving_a_bound` asserted the wide-margin step equals the plain one to `1e-9`, and that held only because the overloaded `eps` suppressed every release at ten. The bounds the refinement now releases there are inactive ones, so the step moves by `1.6e-9`, the size of their multipliers. The tolerance is `1e-6` and the docstring says why it is not bit for bit.

## 4. Relative versus absolute

The python side compares absolutely now, as `over > eps` does. `test_the_margin_is_absolute_as_the_refinement_is` puts one variable of order `1e4` at its bound and steps `1e-6` past it. Under `linear` that is in `crossed` and the clamp warns, where a relative tolerance of `1e-5` hid it. Under `fix_relax` the refinement pins it and the report says `settled` with `crossed` empty, which is the pairing this PR exists to keep. Scaling the tolerance back fails it.

## 5. The ordering

`active_set_changes` validates `degeneracy` before it reads the factor, and `test_arguments_are_checked_before_the_factor_is` covers it with a typo'd `degeneracy` and a tight cap. Swapping the order back fails it.

## Testing

Three new python tests and one rust one, one python test re-asserted and the ordering test extended. 455 pyomo-pounce tests pass and 4 skip, every `pounce-sensitivity` suite, the CLI sens tests, `cargo fmt` clean. Clippy reports nothing new at the lines this round changes. The two `expect()` calls I first wrote for the `bound_relax_factor` read are `map_or(1e-9, ...)`, since the floor is `1e-9` by definition when the option is unreadable, and `refine_step_onto_bounds` at ten arguments carries the `too_many_arguments` it carried at nine.

Items 2, 4 and 5 were each checked by putting the old behaviour back and watching the new test fail. Item 3 is checked both ways inside its rust unit test, and the python twin was run against a build with the release threshold set back to the primal margin, where it fails. Item 1 is checked by clippy.

I read the diff before pushing this time.

PR #743, opened 2026-08-22
Implement reusable PSD certificate

GermanHeim:
## Problem

The sparse PSD classifier currently lives in pounce-cli and it is not reusable.

## Fix

- Extract the CLI’s sparse PSD classifier into pounce-convex 
- Check inertia support before initializing or factoring the backend.
- Re-export EMatrixFormat and ESymSolverStatus through pounce-rs::linsol.

## Blast radius

End users of pounce that do not go through the CLI. Easier to add to oximo.

## Tests

      since moved

> The changelog and CLI were updated using codex

### DIFF
--- a/crates/pounce-cli/src/dispatch.rs
+++ b/crates/pounce-cli/src/dispatch.rs
@@ -37,6 +37,7 @@
 
 use crate::nl_reader::NlProblem;
 use pounce_common::types::{lower_bound_present, upper_bound_present};
+use pounce_convex::{Triplet, certify_psd_lower_triangle};
 
 /// Tolerance for the smallest-eigenvalue sign test in the convexity
 /// check. A Hessian eigenvalue below `-PSD_TOL` is treated as a genuine
@@ -727,100 +728,16 @@ fn socp_reform_flops(h: &QuadHessian) -> u128 {
 
 /// Is the (symmetric, sparse) Hessian positive semidefinite?
 ///
-/// A purely diagonal Hessian is settled in `O(nnz)` by sign — its
-/// eigenvalues *are* its diagonal entries — with no factorization at all;
-/// this keeps large separable / least-squares QPs cheap. A *coupled*
-/// Hessian is certified by a sparse symmetric factorization (see
-/// [`coupled_hessian_is_psd`]): feral's LDLᵀ reports the matrix inertia in
-/// roughly `O(nnz · fill)`, so even the large but sparse coupled Hessians of
-/// the CVXQP family (n ≈ 1000) are classified in well under the solve cost —
-/// no dense `k×k` allocation and no `O(k³)` eigensolve. Returns `true` only
-/// when the smallest eigenvalue is `≥ -PSD_TOL`; an indefinite or
-/// inconclusive result returns `false`, routing to the safe (more general)
-/// class.
-fn hessian_is_psd(h: &QuadHessian, _n: usize) -> bool {
-    if h.is_empty() {
-        return true; // zero matrix is PSD (the linear case)
-    }
-    // Fast path: a diagonal Hessian is PSD iff every diagonal entry is
-    // `≥ -PSD_TOL`. No factorization — essential for large but separable
-    // objectives, where the answer is trivial.
-    if h.keys().all(|(i, j)| i == j) {
-        return h.values().all(|v| *v >= -PSD_TOL);
-    }
-    coupled_hessian_is_psd(h)
-}
-
-/// PSD certificate for a *coupled* Hessian via a sparse symmetric
-/// factorization.
-///
-/// The test is positive-definiteness of the `ε`-shifted matrix `H + ε·I`
-/// with `ε = PSD_TOL`. A genuinely-PSD `H` (smallest eigenvalue `λ_min ≥ 0`,
-/// even a singular one) becomes strictly positive definite after the shift,
-/// so feral factors it with no negative pivots (`inertia.negative == 0`); a
-/// truly indefinite `H` with `λ_min < -PSD_TOL` keeps a strictly-negative
-/// shifted eigenvalue and yields `negative > 0`. The `negative == 0` test on
-/// the shifted matrix is therefore exactly `λ_min ≥ -PSD_TOL` — the same
-/// tolerance the dense path used — and it scales to large sparse Hessians
-/// because the factorization cost tracks the nonzero/fill count, not a dense
-/// `k³`.
-///
-/// The Hessian is compressed to its active variable set so the factored
-/// dimension is `k` (the number of distinct variables in the form). The
-/// [`QuadHessian`] is upper-triangular (`i ≤ j`); feral wants the lower
-/// triangle (`row ≥ col`), so each entry `(i, j)` is emitted at
-/// `(row = j, col = i)`. Every active diagonal is seeded with `ε` (the shift;
-/// `from_triplets` sums it with any diagonal entry already in `H`), which
-/// also guarantees no structurally empty column. A non-`Success`
-/// factorization (singular/fatal — should not occur given the strictly-PD
-/// shift, but possible on a pathological form) is treated conservatively as
-/// not-provably-PSD.
-fn coupled_hessian_is_psd(h: &QuadHessian) -> bool {
-    use feral::{CscMatrix, FactorStatus, Solver};
-
-    // Compress to the active variable set so the factored dimension is `k`.
-    let mut active: Vec<usize> = Vec::with_capacity(2 * h.len());
-    for (i, j) in h.keys() {
-        active.push(*i);
-        active.push(*j);
-    }
-    active.sort_unstable();
-    active.dedup();
-    let k = active.len();
-    let idx = |v: usize| active.binary_search(&v).unwrap();
-
-    // Lower-triangle triplets: H's entry (i ≤ j) maps to (row = j, col = i).
-    // Capacity covers H's nonzeros plus one ε-shift per active diagonal.
-    let mut rows: Vec<usize> = Vec::with_capacity(h.len() + k);
-    let mut cols: Vec<usize> = Vec::with_capacity(h.len() + k);
-    let mut vals: Vec<f64> = Vec::with_capacity(h.len() + k);
-    for ((i, j), v) in h {
-        let (ri, rj) = (idx(*i), idx(*j));
-        // i ≤ j by the upper-tri convention, so rj ≥ ri ⇒ lower triangle.
-        rows.push(rj);
-        cols.push(ri);
-        vals.push(*v);
-    }
-    // εI shift: seed every active diagonal (summed with H's own diagonal).
-    for d in 0..k {
-        rows.push(d);
-        cols.push(d);
-        vals.push(PSD_TOL);
-    }
-
-    let mat = match CscMatrix::from_triplets(k, &rows, &cols, &vals) {
-        Ok(m) => m,
-        Err(_) => return false, // malformed ⇒ be conservative
-    };
-    let mut solver = Solver::new();
-    match solver.factor(&mat, None) {
-        FactorStatus::Success => {
-            // PD ⟺ no negative pivots in the LDLᵀ of the ε-shifted matrix.
-            solver.inertia().map(|i| i.negative == 0).unwrap_or(false)
-        }
-        // Singular / wrong-inertia / fatal: cannot certify ⇒ safe fallback.
-        _ => false,
-    }
+/// A diagonal Hessian is settled in `O(nnz)` by sign. A coupled Hessian is
+/// certified from the inertia of `H + PSD_TOL·I`.
+fn hessian_is_psd(h: &QuadHessian, n: usize) -> bool {
+    let lower: Vec<_> = h
+        .iter()
+        .map(|(&(row, col), &val)| Triplet::new(col, row, val))
+        .collect();
+    certify_psd_lower_triangle(n, &lower, PSD_TOL, || {
+        Box::new(pounce_feral::FeralSolverInterface::new())
+    })
 }
 
 #[cfg(test)]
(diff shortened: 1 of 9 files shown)


crates/pounce-cli/src/dispatch.rs
```diff
@@ -727,100 +728,16 @@ fn socp_reform_flops(h: &QuadHessian) -> u128 {
...
-    }
+/// A diagonal Hessian is settled in `O(nnz)` by sign. A coupled Hessian is
+/// certified from the inertia of `H + PSD_TOL·I`.
+fn hessian_is_psd(h: &QuadHessian, n: usize) -> bool {
+    let lower: Vec<_> = h
+        .iter()
+        .map(|(&(row, col), &val)| Triplet::new(col, row, val))
+        .collect();
+    certify_psd_lower_triangle(n, &lower, PSD_TOL, || {
+        Box::new(pounce_feral::FeralSolverInterface::new())
```
jkitchin: Routing the certificate through `FeralSolverInterface` (rather than the raw `feral::Solver` this replaces) changes zero-pivot handling, and it is measurable.

`FeralSolverInterface::factor` returns `Singular` whenever the inertia reports *any* zero pivot (`crates/pounce-feral/src/lib.rs:1093-1100`), and `Factorization::new` turns that into `Err`. The old code read `solver.inertia().map(|i| i.negative == 0)` and ignored the zero count entirely.

Differential-tested on `s * [[1, 1], [1, 1]]` (rank-deficient PSD, eigenvalues `{0, 2s}`) with `PSD_TOL = 1e-9`:

| scale `s` | old | new |
|---|---|---|
| 1e-3 … 1e6 | true | true |
| 1e9, 1e12 | true | **false** |

Past `s ~ 1e9` the absolute `1e-9` shift drops under the relative precision floor, so the certificate stops meaning anything there and the old `true` was luck as much as anything. The direction is conservative — `ConvexQp` becomes `Nlp`, still a correct answer — but it is a dispatch change for badly-scaled rank-deficient convex QPs, which is the "right answer, slowly" class `CLAUDE.md` asks to have named up front rather than found later.

Either call it out in the PR body, or scale the shift by the matrix norm instead of using an absolute `tolerance`.

Two things I checked that are *not* problems: FERAL's `ScalingStrategy::Auto` is a symmetric diagonal scaling, so by Sylvester it cannot change the inertia; and `check_neg_evals=false` means the `inertia_pivot_floor` branch never fires here. Separately and much lower stakes: `FeralSolverInterface::new()` reads the `POUNCE_FERAL_*` env knobs, so those now influence problem *classification*, which bare `feral::Solver::new()` did not.

---

crates/pounce-convex/src/psd_certificate.rs
```diff
@@ -0,0 +1,411 @@
...
+where
+    F: FnOnce() -> Box<dyn SparseSymLinearSolverInterface>,
+{
+    if !tolerance.is_finite() || tolerance < 0.0 {
+        return false;
+    }
+
+    let mut entries = HashMap::<(usize, usize), f64>::with_capacity(triplets.len());
+    for &Triplet { row, col, val } in triplets {
+        if row >= n || col >= n || row < col || !val.is_finite() {
```
jkitchin: This collapses "not certified PSD" and "you handed me garbage" into the same `false`: upper-triangle triplets (`row < col`), an out-of-range index, a non-finite value, a negative or non-finite tolerance all return the same answer as a genuinely indefinite matrix.

That was defensible while this was a private CLI helper with one call site. It is now public on `pounce_rs::convex`, and handing it the *upper* triangle is an easy mistake to make — with no panic, no `Result`, and no log to catch it. Given how much of this repo's commentary is about silent misroutes, a `Result<bool, _>` (or at minimum a `tracing::warn!` / `debug_assert!` on the reject paths) would fit better.

---

crates/pounce-cli/src/dispatch.rs
```diff
@@ -727,100 +728,16 @@ fn socp_reform_flops(h: &QuadHessian) -> u128 {
...
-    match solver.factor(&mat, None) {
-        FactorStatus::Success => {
-            // PD ⟺ no negative pivots in the LDLᵀ of the ε-shifted matrix.
-            solver.inertia().map(|i| i.negative == 0).unwrap_or(false)
-        }
-        // Singular / wrong-inertia / fatal: cannot certify ⇒ safe fallback.
-        _ => false,
-    }
+/// A diagonal Hessian is settled in `O(nnz)` by sign. A coupled Hessian is
+/// certified from the inertia of `H + PSD_TOL·I`.
```
jkitchin: The ~55 lines of rationale this replaces were deleted, not relocated — I checked `psd_certificate.rs` and they are not there either.

What is gone: the argument that `negative == 0` on `H + eps*I` is *exactly* `lambda_min >= -PSD_TOL` (and why the shift makes that true for a singular PSD `H`), why the diagonal case is `O(nnz)` and matters for the large separable/least-squares shapes, and why the sparse factorization beats the dense `k^3` eigensolve on the CVXQP family. The new module doc states the mechanism but not the argument, so after this PR nothing in the tree explains why the test is valid.

Worth carrying at least the Sylvester/shift paragraph into `psd_certificate.rs`, where it now belongs.

---

crates/pounce-cli/src/dispatch.rs
```diff
@@ -727,100 +728,16 @@ fn socp_reform_flops(h: &QuadHessian) -> u128 {
...
+fn hessian_is_psd(h: &QuadHessian, n: usize) -> bool {
+    let lower: Vec<_> = h
+        .iter()
+        .map(|(&(row, col), &val)| Triplet::new(col, row, val))
```
jkitchin: Two things here.

**Naming.** The `QuadHessian` key is `(i, j)` with `i <= j` (upper-triangular, per `nl_quadratic.rs:46-49`). Binding it as `row, col` and then swapping into `Triplet::new(col, row, ...)` reads like a transpose bug on first pass. The old code carried an explicit comment for exactly this step; suggest:

```rust
// QuadHessian is upper-triangular (i <= j); the certificate wants the
// lower triangle, so (i, j) is emitted at (row = j, col = i).
.map(|(&(i, j), &val)| Triplet::new(j, i, val))
```

**Cost.** The diagonal fast path got about 8x slower, because this materializes a `Vec<Triplet>` and `certify_psd_lower_triangle` then builds a `HashMap` before either can notice the matrix is diagonal — where the old path was a zero-allocation `keys().all()` scan. Measured on the n=50,000 diagonal case from `large_diagonal_hessian_is_cheap_and_psd`, test profile (optimized), 100 iterations: 33.6ms -> 265.1ms, i.e. 0.34ms -> 2.65ms per call.

Negligible next to a solve, so not a blocker — flagging it only because the CHANGELOG entry advertises the fast path, and `hessian_is_psd` runs once per quadratic constraint row.

---

crates/pounce-convex/src/psd_certificate.rs
```diff
@@ -0,0 +1,411 @@
...
+
+/// Certify that a sparse symmetric matrix is positive semidefinite within an
+/// absolute tolerance.
+///
+/// `triplets` supplies the matrix's lower triangle in zero-based
+/// coordinates (`row >= col`). Duplicate coordinates are summed. The
+/// diagonal fast path accepts an eigenvalue down to `-tolerance`, while coupled
+/// matrices are tested by factoring `A + tolerance * I` and inspecting its
+/// inertia. A coupled matrix exactly on that shifted singular boundary may be
+/// rejected conservatively when the backend cannot produce an inertia.
```
jkitchin: "when the backend cannot produce an inertia" is not the mechanism. On the shifted singular boundary the backend produces an inertia perfectly well — the rejection happens because the factorization reports `Singular`, so `Factorization::new` returns `Err` at line 110 and the inertia is never consulted. The `provides_inertia()` guard at line 108 is a separate path.

Worth restating, since this doc line is the only place the boundary behaviour is described. While here, it is also worth noting that the diagonal path accepts `lambda_min >= -tolerance` whereas the coupled path effectively tests `lambda_min > -tolerance` (the shifted matrix must be strictly PD); the two branches differ exactly on the boundary.

---

crates/pounce-convex/src/psd_certificate.rs
```diff
@@ -0,0 +1,411 @@
...
+        return false;
+    };
+    let mut shifted = BTreeMap::<(usize, usize), f64>::new();
+    for ((row, col), value) in entries {
+        let Ok(compressed_row) = active.binary_search(&row) else {
+            return false;
+        };
+        let Ok(compressed_col) = active.binary_search(&col) else {
+            return false;
+        };
```
jkitchin: Both `else { return false }` arms are unreachable: `active` is built at lines 57-63 from exactly the `row` and `col` of every key in `entries`, then sorted and deduped, so `binary_search` on either coordinate of a key always finds it. `.expect()` under the crate's test-only unwrap allowance, or a plain index lookup, would say that more honestly than a `false` that reads as a real failure mode.

---

crates/pounce-rs/tests/convex_surface.rs
```diff
@@ -28,6 +28,19 @@ fn boxed_qp(t: &[f64]) -> QpProblem {
...
+#[test]
+fn sparse_psd_certificate_and_linsol_enums_are_available_through_the_facade() {
+    let lower = [
+        Triplet::new(0, 0, 2.0),
+        Triplet::new(1, 0, 1.0),
+        Triplet::new(1, 1, 2.0),
+    ];
+    assert!(certify_psd_lower_triangle(2, &lower, 1e-9, backend));
+
+    assert_eq!(ESymSolverStatus::Success, ESymSolverStatus::Success);
```
jkitchin: This asserts nothing — `Success == Success` holds for any type deriving `PartialEq`, and the `use` on line 12 already proves the name is reachable through the facade. The `EMatrixFormat` assertion on the next line does carry content (it pins the default backend's format), so this one reads as an oversight rather than a deliberate pair.

Either drop it, or make it test something, e.g. that a real status round-trips out of a backend call.

---

jkitchin (commented):
Built the branch and differential-tested the extracted certificate against the code it replaces. **The extraction is faithful — no correctness defect found.**

I ran 1600 randomized symmetric matrices (k in {2, 3, 6, 12}; random-indefinite, `A^T A`, and deliberately rank-deficient `A^T A`; scales 1e-2 … 1e3) through both the old raw-`feral::Solver` classifier and the new `certify_psd_lower_triangle`: **zero disagreements**, in either direction. `cargo fmt --all -- --check` is clean, `cargo clippy -p pounce-convex --all-targets` adds no warnings on the new module, and the `pounce-convex` (7 new tests), `pounce-cli dispatch` (48), and `pounce-rs convex_surface` (5) suites all pass locally. CI is green on all 11 checks.

The one behaviour change I did find is at large matrix scale, and it is conservative in direction — details inline on `dispatch.rs:739`. Short version: `FeralSolverInterface` maps any zero pivot to `Singular` where the old raw-solver path ignored the zero count, so `s * [[1, 1], [1, 1]]` (rank-deficient PSD) certifies `true` under the old code and `false` under the new for `s >= 1e9`. `ConvexQp` becomes `Nlp` — still a correct answer, just via the general path — but it is a dispatch change, and per `CLAUDE.md` that is the class worth naming in the PR body rather than leaving to be rediscovered. It is also a fair prompt to ask whether the shift should be relative to the matrix norm rather than an absolute `PSD_TOL`; the old code had the same absolute-shift weakness, this just changes where it bites.

Two other substantive points, both inline:

- **`certify_psd_lower_triangle` returns `false` for malformed input** (`psd_certificate.rs:39`), conflating "not certified PSD" with "you handed me garbage". Fine for a private helper with one call site; less so now that it is public on `pounce_rs::convex` and passing the upper triangle is a silent no-op.
- **The rationale comments were deleted rather than relocated** (`dispatch.rs:731`). The argument for *why* `negative == 0` on the shifted matrix is exactly `lambda_min >= -PSD_TOL` no longer exists anywhere in the tree.

Smaller items inline: an ~8x slowdown on the diagonal fast path (0.34ms -> 2.65ms at n=50,000 — negligible next to a solve, but the CHANGELOG advertises that path), a misleading doc line about the singular boundary, confusing `row`/`col` naming at the `QuadHessian` -> `Triplet` conversion, two unreachable `return false` branches, and a tautological assertion in the facade test.

Two process notes, not inline:

- The certificate depends only on `pounce-linsol` and `Triplet`, but lives in `pounce-convex` — so the oximo consumer the PR body names pulls in the whole IPM, simplex, presolve, and rayon to get it. `pounce-linsol` (which already owns `Factorization`) may be the better home if `Triplet` can follow.
- The checklist leaves "Tests fail on the parent commit" and the book-page box unchecked while adding a new public API to `pounce_rs::convex`; `docs/src/rust.md` is not updated.

None of this blocks on correctness. Item 1 is the one I would want acknowledged in the PR body before merge.

---

crates/pounce-convex/src/psd_certificate.rs
jkitchin: Addressed in 0727af5 — confirming against the current head (97106eb).

`PsdCertificateError` gives each reject path its own variant, and the coordinates come back with it, which is more than I asked for. Verified the four input classes discriminate correctly:

```
upper  -> UpperTriangle { row: 0, col: 1 }
oob    -> OutOfBounds { row: 2, col: 0, dimension: 2 }
nan    -> NonFiniteValue { row: 0, col: 0 }
negtol -> InvalidTolerance
```

The upper-triangle case — the mistake I was actually worried about — is now impossible to make silently.

`hessian_is_psd`'s `.unwrap_or(false)` is the right call at that site: the CLI constructs the triplets itself from a `QuadHessian` it already knows is well-formed, so the error can only be a bug in the conversion, and conservative-to-NLP is the correct posture for a classifier either way.

Also re-checked the rest of the round on this head: differential sweep against the old classifier is still 0/1600 disagreements in either direction; the diagonal fast path is back to 0.41ms/call at n=50,000 versus a 33.6ms/100 old-code baseline (it was 2.65ms after the regression); `fmt` clean and no new clippy warnings on the changed code. The new `row as Index + 1` cast is safe — `Index::try_from(active.len())` has already succeeded above, so `row <= Index::MAX - 1` and the increment cannot overflow.

Resolving.

---

PR #756, opened 2026-08-23
sensitivity: read the degeneracy weak set off the barrier diagonal

devin-griff:
## The defect

On a 62k-variable collocation model with 246 perturbed parameters, `degeneracy="directional"` returned a step whose first-order error was fifty times the plain one-sided step's. At a 1% perturbation the rms error against a warm re-solve was 5.0e-5 for the directional step and 1.0e-6 for the one-sided step, and the gap scaled linearly with perturbation size, so the error was in the direction itself.

## The cause

The weak set can hold coordinates far from their bounds. The classifier's ambiguous class is a band on `sigma_i / |H_ii|`, and low curvature widens it until coordinates a fifth of their range from a bound land inside. That membership is honest, the status of those bounds is genuinely unresolved at the base point. The defect was downstream: the QP treats an engaged row as sitting at its bound, so on this model it pinned 34 interior coordinates whose bounds no step of this size could reach, and the direction came back wrong at first order.

The QP could also come back unbounded. A coordinate that an equality pins enters the QP when that pin's parameter is perturbed, because the pin moves the coordinate. But a bound force on that coordinate moves nothing, since the equality determines it, so the row's entire column of the QP's reduced matrix is zero. A zero column beside a nonzero linear term is an unbounded QP.

## The fix

Engagement is untouched from main, a row engages when the direction moves its coordinate toward the bound past the relative noise band, so nothing reads the perturbation's size and the decision is linear in the step. The change is one drop rule at the point where the QP builds an engaged row's column: the row is decided only when its bound is at a kink, read off `kappa = sigma * S_kk`, the barrier weight times the row's own diagonal of the reduced matrix, which the column provides for free. `sigma` equals the curvature reduced along the coordinate at an exact kink and `S_kk` is that reduced curvature's inverse, so `kappa` is 1 there at any curvature, coupling, or scaling, and falls as the squared ratio of kink width to slack away from one. A row below `KAPPA_MIN = 1e-3` is dropped and its plain movement stands. A coordinate an equality pins is the limiting case, `S_kk` exactly zero, dropped by the same test.

The threshold is a constant, but on a dimensionless, step-free, frame-invariant statistic with an exact value at the object it tests for. Measured populations, with `sigma` in natural units so scaling cancels:

| population | kappa |
|---|---|
| exact fixture kinks, curvature 1 to 1e4, coupling ratio to 1e-2 | 1.000 |
| held solves near a real release, two collocation models | 4.4e-2 and 6.4e-3 |
| genuinely interior weak rows, 792 measured across two models | 3e-8 and below |
| pin-owned rows | at or below zero |

A row at the threshold sits about 30 kink widths from its bound, and the error of deciding it either way is bounded by that slack, order `sqrt(mu)`. The first-order failure modes on both sides require `kappa` far on the wrong side of the middle, which the rule precludes.

## Results

| setup | before | after |
|---|---|---|
| N=25, 1% perturbation, directional rms | 5.0e-5 | 1.0e-6 (matches one-sided) |
| N=25, 10% perturbation, directional rms | 4.8e-4 | 7.0e-5 |
| N=5 with terminal segment, 10% | QP unbounded | completes, 2.2 s |

Directional-call cost on the 62k model returns to the pre-PR shape, 4.7 s against 2.3 s for the withdrawn reach-test design, since interior rows near bounds get their column built before being dropped.

## Tests

`degenerate_weak_curvature.rs` holds two identical kinks with curvature 1 and 1e4 and asserts both are weak, both move exactly, and both hold at a holding-side step of 1e-10. `degenerate_weak_set.rs` adds the two rows the QP must not decide: an interior coordinate in the weak set through the ambiguous class, moved toward its bound through an equality and an order-one partner, must not be held, and a coordinate an equality pins is dropped while the rest is decided. All fail on main. The twelve invariance legs of #762, now merged into this branch from main, pass, including the ambiguous-kink leg that fails the previous head.

## History

The branch went through three withdrawn designs before this one: a membership band on the barrier diagonal (a curvature band, drops stiff kinks), a slack term in the engagement test (breaks the linearity a directional derivative owes), and a classifier-certified split (the class divides by diagonal curvature where the kink question is about reduced curvature, so coupled kinks fell through). Each was killed by a measured counterexample, and the final design was chosen after measuring `kappa` on every one of them. The earlier commits carry the withdrawn versions.


crates/pounce-sensitivity/src/solver.rs
```diff
@@ -1268,51 +1288,59 @@ impl Solver {
...
+    /// and the multiplier vanish together, `s ~ z ~ sqrt(mu)`, and
+    /// their ratio is order one; an inactive bound's entry is `O(mu)`
+    /// and a strongly active one's is `O(1/mu)`, both orders of
+    /// magnitude outside the band. No curvature appears: whether a
+    /// bound sits at a kink is a fact about the bound's own geometry.
+    ///
+    /// The activity classifier is deliberately not consulted. Its
+    /// classes are bands on `sigma / |H_ii|`, the right object for the
+    /// covariance question of whether the barrier pins the variance
+    /// against the objective, and the wrong one here. Low curvature
```
jkitchin: **This band is itself a curvature band — the defect it replaces, mirrored.**

> No curvature appears: whether a bound sits at a kink is a fact about the bound's own geometry.

This doesn't hold. At a kink the multiplier is *generated by* the curvature along the coordinate: dual feasibility gives `z ≈ H·s`, and with `s·z = mu` that is

```
s = sqrt(mu/H),   z = sqrt(mu·H),   sigma = z/s ≈ H
```

So `sigma ≈ 1` at a kink exactly when the curvature there is ≈ 1. `[1e-2, 1e2]` is a two-decade band on curvature, where the classifier's was a one-sided threshold on `sigma/H`.

Worth noting `sigma/|H_ii|` equals 1 at *every* kink regardless of `H` — that ratio is the curvature-invariant statistic. The real defect the PR found is that its *bands* were too loose on the interior side (there `sigma/H = mu/(s²H)`), not that the normalization was wrong.

**Reproducer.** Two coordinates, both on their lower bound, both with multipliers → 0 as `mu` → 0, both with genuinely different one-sided derivatives — identical in kind, differing only in curvature (1 vs 1e4):

```
min 0.5·x0² − A1·p·x0 + 0.5·C·x1² − A1·C·p·x1
s.t. p = 0,  0 ≤ x0 ≤ 10,  0 ≤ x1 ≤ 10,  C = 1e4, A1 = 1.10
```

Exact answer for `dp = +1e-3`: both leave the bound, `dx0 = dx1 = 1.1e-3`.

| | weak set | `parametric_step_directional` |
|---|---|---|
| base `5dc835f` | `[0, 1]` | `[1.100000e-3, 1.100000e-3]` ✅ exact |
| this PR `16f6df1` | `[0]` | `[1.100000e-3, 5.499863e-4]` ❌ 2× off |

A first-order error on `x1` — the same failure mode this PR set out to fix, reached from the other side. Both coordinates are confirmed kinks and not strongly-active bounds: tightening `tol` from `1e-8` to `1e-11` moves `x` from `[2.80e-4, 3.32e-6]` to `[1.82e-5, 1.82e-7]`, both slacks and both multipliers shrinking together. And since `sigma ≈ H` is `mu`-independent, no tolerance ever brings `x1` into the band.

This isn't exotic for the target models: the motivating 62k collocation model had curvature small enough to widen the old bands, which means its curvature spans decades across coordinates. A model like that will have kinks on *both* sides of a two-decade window centred on 1.

Two directions that keep the fix's benefit without this cost:

- Keep the scale-free ratio and require a **conjunction**: `sigma/|H_ii|` near 1 *and* the slack small. The interior coordinate that motivated this PR fails the second test (it was a fifth of its range from the bound); a high- or low-curvature kink passes both.
- Or classify the slack and the multiplier directly against their own scales, rather than their ratio against an absolute constant.

Related: `sigma` here is `z_L/s_L + z_U/s_U` in the algorithm's **scaled** frame (`ipopt_cq.rs:778`, built from `iv.x`), so `nlp_scaling_method` and `obj_scaling_factor` also shift membership. The old ratio was invariant to both.

Finally, `degenerate_weak_set.rs` encodes the coupling in its own comment: `S0 = 1e-5` is chosen so `mu/S0²` lands in the band "for the final `mu` a tol of 1e-8 reaches". Raise `tol` to `1e-6` and the fixture's own precondition assertion fails. Good that it asserts rather than silently passing, but it is a signal about the criterion.

---

crates/pounce-sensitivity/src/solver.rs
```diff
@@ -1097,6 +1097,16 @@ impl Solver {
...
+        // A weak row whose own diagonal entry of `S` is zero cannot be
+        // decided by a pin force: an equality owns its coordinate, and
+        // under a parametric step the one equality that moves a
+        // coordinate is the pin of a perturbed parameter. Its movement
+        // is the perturbation itself, so the engagement test cannot
+        // exclude it, and admitting it puts a zero on the diagonal of
+        // `S` beside a nonzero gradient, which is an unbounded QP.
+        // The bound's status there is the parameter's business, inside
+        // the bound or infeasible, so the row stays out for good.
+        let mut inert: Vec<usize> = Vec::new();
```
jkitchin: The `inert` half of this PR I'd take as-is. `S_kk = 0` on a PSD `S` forces the whole column to zero by Cauchy–Schwarz, so dropping the row is exactly `lambda_k = 0`, and the alternative genuinely is an unbounded QP (`g_k < 0` is the engagement condition). Termination still holds: `inert` is monotone, expansion skips it, and each pass either grows `engaged` or breaks.

One thing to fix alongside it — **`parametric_step_directional`'s own doc contract now contradicts the code.** At `solver.rs:961-963` (just above, unchanged by this PR) it still says:

> rows the equalities already pin (movement exactly zero under every candidate) never enter and never make anything singular.

That sentence is the claim this block exists to refute: they *do* enter, and they *do* make the QP unbounded. Worth rewriting to describe the drop.

---

crates/pounce-sensitivity/src/solver.rs
```diff
@@ -1113,7 +1123,17 @@ impl Solver {
                     return Err(SolverError::BacksolveFailed);
                 }
                 work += 1;
-                proj[k] = Some(weak.iter().map(|w| xk[w.var_row]).collect());
+                let col: Vec<Number> = weak.iter().map(|w| xk[w.var_row]).collect();
+                let own = sign(k) * col[k];
+                let col_scale = col.iter().fold(0.0_f64, |a, &b| a.max(b.abs()));
+                if own.abs() <= 1e-12 * col_scale.max(1e-300) {
+                    inert.push(k);
+                }
```
jkitchin, line 1173: Nit: `sign(k)` is ±1 and `own` is only ever read through `.abs()`, so the multiply cannot change the test — `own.abs() == col[k].abs()`.

Either drop the `sign(k) *`, or make the test signed (`own <= 1e-12 * col_scale.max(1e-300)`), which would additionally catch a **negative** diagonal. `S_kk < 0` is likewise a QP the engine cannot bound, and with `HessianInertia::Unknown` it would reach `feral` as a nonconvex reduced problem rather than being classified here.

---

crates/pounce-sensitivity/src/algorithm_backsolver.rs
```diff
@@ -493,6 +493,15 @@ impl PdSensBacksolver {
         }
     }
 
+    /// [`Self::barrier_sigma_x`] as a dense var-x vector, `None` when
+    /// the diagonal is not a `DenseVector`.
+    pub(crate) fn barrier_sigma_x_dense(&self) -> Option<Vec<Number>> {
+        self.barrier_sigma_x()
+            .as_any()
+            .downcast_ref::<DenseVector>()
```
jkitchin: Minor, and partly pre-existing: at the call site a `None` from here yields an empty weak set, i.e. degeneracy silently undetected and the caller takes the plain step at a degenerate point.

The guard immediately below it raises `BadShape` on a length mismatch, on the stated argument that "skipping rows over it would silently shrink the weak set" — but a failed downcast shrinks it all the way to nothing, silently. The two guards read inconsistently.

It matches how `bound_rows()`'s `None` is already handled, and `curr_sigma_x` builds from `iv.x.make_new()` so in practice this is dense, so low priority — but if the length check is worth raising on, this probably is too.

---

CHANGELOG.md
```diff
@@ -9,6 +9,31 @@ changes.
...
+  whether the barrier pins the variance against the objective. For
+  degeneracy that ratio is the wrong object. Low curvature widens the
+  bands until they admit coordinates far from any bound, and the QP
+  then held interior coordinates and returned a direction wrong at
+  first order. A bound now counts as weak when the barrier diagonal
+  entry the KKT matrix carries for it lies within a factor of `1e2`
+  of one. At a kink the slack and the multiplier vanish together, so
+  the entry is order one, while an inactive bound's is `O(mu)` and a
+  strongly active one's is `O(1/mu)`, both orders of magnitude
+  outside the band.
```
jkitchin: Two user-facing docs describe the route this entry replaces, and neither is touched by the PR:

- **`crates/pounce-py/src/solver.rs:401-405`** — the Python `weakly_active_bounds` docstring still says the result is "on the bound with a multiplier of the same order as the slack, the classifier's WEAKLY_ACTIVE and AMBIGUOUS labels." After this change the classifier is deliberately not consulted, so the docstring names a mechanism that is gone.

- **`docs/src/sensitivity.md`**, §"A held solve at a kink: `degeneracy`" — still routes through "The activity classifier reports such a bound as ambiguous, which is its honest answer." Its neighbouring sentence, "at a strongly active bound that is around 1e8 … and at a kink it is of order one", was a descriptive aside before; this PR promotes it to the load-bearing membership rule. Per the comment on `solver.rs`, it holds only when the curvature at the kink is order one, so if that framing survives review it needs the qualifier stated.

---

jkitchin (dismissed):
Checked out both head (`16f6df1`) and base (`5dc835f`), read the changed code in context, ran `cargo test -p pounce-sensitivity` (passes on both), and built a reproducer. CI is green on all 11 checks.

The diagnosis in the body is right and well argued: `sigma/|H_ii|` is the covariance question rather than the degeneracy question, and low curvature widening the classifier's bands until they admit interior coordinates is a real defect worth fixing. The `inert`-row drop is a clean fix with correct reasoning — on a PSD `S`, `S_kk = 0` forces the whole column to zero, so dropping the row is exactly `lambda_k = 0`, and the alternative really is an unbounded QP.

The concern is the replacement membership test, which has a defect of the same shape as the one it removes.

**`sigma ∈ [1e-2, 1e2]` is itself a curvature band.** At a kink the multiplier is generated by the curvature along the coordinate — `z ≈ H·s` with `s·z = mu` gives `sigma = z/s ≈ H` — so the band is a two-decade window on curvature, not the curvature-free geometric test the doc comment claims. On a fixture with two coordinates that are identical kinks differing only in curvature (1 vs 1e4), the base returns the exact direction `[1.100000e-3, 1.100000e-3]` and this PR returns `[1.100000e-3, 5.499863e-4]` — a first-order error, the same failure mode the PR set out to fix. Details, reproducer and two suggested alternatives are inline on `solver.rs`.

Net: I'd take the `inert` half as-is (modulo the stale doc contract noted inline). The membership half needs rework before merge — as it stands it converts a class of false positives into a class of silent false negatives.

Nothing here calls for the fixture sweep: this is sensitivity post-processing, not a solver trajectory change.

Other findings inline — a stale doc contract on `parametric_step_directional`, two user-facing docs still describing the classifier route, and two nits.

---

crates/pounce-sensitivity/src/solver.rs
devin-griff: *Superseded: the design described here was withdrawn after the ambiguous-kink finding, the current design is the kink-measure drop rule in the PR body and the 2026-08-24 comment.*

Confirmed, and the band is withdrawn. Your fixture reproduces as written and is now `degenerate_weak_curvature.rs`, failing only on the banded version. My first transcription dropped the cross Hessian entry and gave a confusing hour, the numbers were yours all along.

The rework takes a third direction rather than the conjunction. Membership reverts to the classifier untouched, and the slack condition moves into the QP's engagement test, which now engages a row only when the step carries its coordinate past the remaining slack plus the existing `sqrt(mu)` band. That answers "small compared to what" with the step itself, so no constant is chosen. A bound the classifier certifies weakly active has its slack treated as zero, barrier width is the solve's own inflation, so a certified kink decides by direction alone at any scale, and the measured slack applies to the ambiguous class, where a kink and a genuinely interior coordinate share one verdict and the slack separates them. The scaled-frame point and the tol-coupled fixture precondition go away with the revert, and the sigma accessor is deleted.

On the motivating model the numbers match the banded version: directional rms 5.0e-5 to 1.0e-6 at 1%, matching one-sided, and 4.8e-4 to 7.0e-5 at 10%.

crates/pounce-sensitivity/src/solver.rs
devin-griff: Kept as-is, with the contract rewritten: the doc above `parametric_step_directional` now describes the drop instead of denying pinned rows can enter. Every engaged pin-owned row ends at the drop either way.

On the final head the drop is the limiting case of the kink measure: a pin-owned row's diagonal is exactly zero, so kappa is at or below zero and the same rule drops it.

crates/pounce-sensitivity/src/solver.rs
devin-griff: Went signed, so a nonpositive diagonal is dropped the same way instead of reaching the QP engine as a nonconvex reduced problem.

Subsumed on the final head: a nonpositive diagonal has kappa at or below zero and the kink-measure threshold drops it.

crates/pounce-sensitivity/src/algorithm_backsolver.rs
devin-griff: Deleted along with the band. The call site reverts to the classifier route.

CHANGELOG.md
devin-griff: With membership reverted the pounce-py docstring is accurate again, so it stands. sensitivity.md now states the kink measure and the pinned-row case, and the kink aside carries your qualifier: sigma at a kink is the order of the curvature along the coordinate, order one on a well-scaled model.

devin-griff:
*Superseded: the design described here was withdrawn after the ambiguous-kink finding, the current design is the kink-measure drop rule in the PR body and the 2026-08-24 comment.*

Since the review, the branch reworks the membership half and keeps the inert half. The band on the barrier diagonal is withdrawn. As the review showed, sigma at a kink is the curvature along the coordinate, so any band on it is a curvature band. `weakly_active_bounds` and the sigma accessor revert to main byte for byte.

The fix moves to the QP's engagement test, which now engages a row only when the step carries its coordinate past the remaining slack plus the existing `sqrt(mu)` band. An interior coordinate whose bound the step cannot reach keeps its plain movement, and no constant is chosen anywhere, both lengths are measured at decision time. One qualifier came out of CI: the pyomo scale-invariance test caught the unqualified slack term breaking linearity at steps below the barrier width. A bound the classifier certifies weakly active now has its slack treated as zero, so a certified kink decides by direction alone at any scale, and the measured slack applies to the ambiguous class, where a kink and a genuinely interior coordinate share one verdict and the slack separates them. The review's reproducer is committed as `degenerate_weak_curvature.rs` and passes here and on main.

The motivating model's numbers match the banded version, directional rms 5.0e-5 to 1.0e-6 at 1%, and the CSTR notebook's stored outputs reproduce exactly. Details are in the thread replies, the updated body, and commits e3021a91, 02fca671, f962d746, 9b4a919e.

jkitchin:
**The certified/ambiguous split is not a safe discriminator for a kink — the holding side goes first-order wrong on `9b4a919`.**

The rework rests on "a genuine kink is certified `WEAKLY_ACTIVE`, so its slack is zero and its decision is linear at any scale." My own review comment is what made that look safe, and it was wrong: I told you `sigma/|H_ii|` equals 1 at *every* kink regardless of curvature. It doesn't, and the reason is the same index-space-shaped confusion the rest of this PR has been about — but between two *curvatures* rather than two index spaces.

`classify` divides `sigma` by the Hessian's **diagonal** `H_ii`. The multiplier at a kink is generated by the curvature **reduced along that coordinate**. Eliminating a free partner `y` from `[[h, c], [c, m]]`:

```
reduced = h - c²/m,    sigma = z/s = reduced,    ratio = reduced / h
```

so the ratio is 1 only when the coordinate is **decoupled**. My reproducer for you was two *independent* kinks differing in curvature — both decoupled, both ratio 1, both certified. That is exactly the case this design handles. Drive `c²/(h·m)` toward 1 instead and the ratio falls below the `1e-1` band edge: a genuine kink, classified `AMBIGUOUS`, which then takes the measured-slack path.

## Reproducer

```
min  0.5·h·k² + c·k·y + 0.5·m·y² − A·p·k
s.t. p = 0,  0 ≤ k ≤ 10,  y free
```

`h = m = 1`, `A = 1.10`, `c = sqrt(1 − rho)` so the reduced curvature along `k` is `rho`. At `p = 0` the reduced gradient at `k = 0` is zero and the multiplier vanishes with `mu` — a kink by construction, same as `TwoCurvatureKinks`. Exact answer: `dk/dp = A/rho` moving the pin up, and **`0`** moving it down, since the pin would drive `k` through its own bound.

Holding-side `dk/dp`, `tol = 1e-8`, slope taken as a difference quotient on one side (the step is affine in `delta`, so `dx/delta` would not do):

| reduced | ratio | class | `main` | **`9b4a919`** | exact |
|---|---|---|---|---|---|
| 1.0 | 1.0 | WEAKLY_ACTIVE | 0 | 0 | 0 |
| 1e-1 | 0.1 | WEAKLY_ACTIVE | 0 | 0 | 0 |
| 1e-2 | 0.01 | **AMBIGUOUS** | 0 | **110.0** | 0 |
| 1e-3 | 0.001 | **AMBIGUOUS** | 0 | **1100.0** | 0 |

`main` is correct in every row, including both ambiguous ones. The leaving side is correct on this branch everywhere — the defect is one-sided.

The transition sits exactly at the measured slack, which is what identifies the mechanism rather than guessing at it. At `reduced = 1e-2` the slack is `5.9e-4`, so `delta = -1e-3` still clears it and the row engages and holds; `delta = -1e-6` moves only `1.1e-4`, fails the engagement test, and the kink is treated as free — so `k` walks through its lower bound and the holding derivative reads the *leaving* side's answer.

This is gh#672 finding 4's shape. You named the class yourself over on #762 an hour ago:

> an absolute length compared against a homogeneous quantity breaks linearity even when the length is measured rather than chosen

That is precisely it. Measuring the length rather than choosing it removes the arbitrary constant, but the length is still fixed while the step shrinks, so linearity still breaks — just at a data-dependent scale instead of a chosen one. `9b4a919` fixed that for the certified class by zeroing the slack; the ambiguous class still carries it.

Relevant to the motivating model: strong coupling between neighbouring coordinates is what collocation *is*, so a kink at a ratio well below `1e-1` isn't a corner there. Your statement that the column's holdups "classify ambiguous and stay excluded" is right for genuinely interior coordinates — it's *kinks* in that class that are now mishandled, and the 62k validation would not surface it, since it measures rms against a warm re-solve at 1% and 10% perturbations, both far above the slack where this is still correct.

## Not a blocker on the approach

The engagement test is still the right home for this — the interior-coordinate defect the PR set out to fix is real and the fix is sound where the slack means what it should. What needs rethinking is only the discriminator: the certified/ambiguous split is a proxy for "is this a kink", and it's the wrong proxy, because the class is set by `reduced/diagonal` while the question is about `reduced` alone. Comparing the slack against something homogeneous with the step, or reading the kink test off the reduced curvature rather than the class, both look like they'd hold — but you know this surface far better than I do and I'd rather not prescribe.

The case is committed as a leg on #762 (`leg_magnitude_an_ambiguous_kink_decides_the_same_way_at_every_step_size`, with `the_coupled_fixture_carries_an_ambiguous_kink` asserting the precondition so it can't go vacuous). On `9b4a919` it is the **only** one of the twelve that fails, so it names the dimension rather than the model; on `main` all twelve pass. Feel free to pull it over here if it's easier to iterate against.

---

devin-griff:
Confirmed, and the certified/ambiguous split is withdrawn too. Your coupled reproducer walked through its bound on `9b4a919e` exactly as your table shows, and the discriminator is now the object you pointed at, the reduced curvature itself, read where the QP already has it.

The rule on the new head: engagement is main's movement test again, so nothing anywhere reads the perturbation's size and every decision is linear in the step. Each engaged row is decided only when `kappa = sigma * S_kk` is at least `1e-3`, computed from the column the QP builds anyway. `sigma` equals the reduced curvature at an exact kink and `S_kk` is its inverse, so `kappa` is 1 there at any curvature, coupling, or scaling, and equals the squared ratio of kink width to slack away from one. A pin-owned coordinate is the limiting case at exactly zero, dropped by the same test. Your coupled fixture measures `kappa = 1.000` through this code path.

The threshold is a constant, but on a dimensionless, step-free, frame-invariant statistic with an exact value at the object it tests for, and the homogeneity class you named cannot reach it since no step appears. Measured populations, `sigma` in natural units so scaling cancels:

| population | kappa |
|---|---|
| exact fixture kinks, curvature 1 to 1e4, coupling ratio to 1e-2 | 1.000 |
| held solves near a real release, two collocation models | 4.4e-2 and 6.4e-3 |
| genuinely interior weak rows, 792 measured across two models | 3e-8 and below |
| pin-owned rows | at or below zero |

A row at the threshold sits about 30 kink widths from its bound, and the error of deciding it either way is bounded by that slack, order `sqrt(mu)`, so a first-order error on either side requires `kappa` far on the wrong side of the middle, which the rule precludes.

Verification on the new head, `3abaaac5`:

| check | result |
|---|---|
| rust suites | 22, all pass |
| the twelve #762 invariance legs, including the new ambiguous-kink leg | all pass, run from the merged tree |
| pyomo suite | 455 pass |
| the active-set notebook | reproduces exactly |
| regression tests on main | fail as intended |
| 62k model, 1% and 10% and terminal-segment setups | rms unchanged, no fallback |
| a second collocation model, near-release and interior probes | decided and dropped respectively |

Thanks for committing the leg, it is in the branch via the merge and it is the test that killed the previous head.

Housekeeping: the earlier comments describing the withdrawn designs were edited in place rather than left stale. The first summary and the membership-thread reply now open with a superseded note pointing here, and the inert, sign, and docs replies state how the final rule subsumes them.

jkitchin:
Pushed two commits to this branch: a merge of current `main` (the head was 71 commits behind, and CI's green run was against that stale head), and one test file.

## The gap

`kappa`'s column back-solve already carries the sign in its right-hand side —

```rust
unit[weak[k].var_row] = sign(k);
```

— so by linearity `col[k] = sign(k) * S_kk`, and the second multiply in `own = sign(k) * col[k]` is what recovers the bare diagonal. Applied once instead of twice, an upper-bound row lands at `kappa <= 0` for any geometry whatsoever, is dropped as though an equality owned it, and its bound goes undecided.

Nothing in the crate could see that. Every fixture reaching `parametric_step_directional` has its weak rows at lower bounds: `degenerate_expansion.rs` and `degenerate_reentry.rs` assert it outright ("weak bounds here are all lower"), `sens_invariance_legs.rs` pins its kink as `(0, true)`, and both fixtures this branch adds are lower-bound too. So `sign(k)` is `+1` everywhere the drop rule has ever run.

Worth flagging because my own inline nit on the earlier head asked to **drop** the `sign(k) *`, on the grounds that `own` was read through `.abs()` and the multiply could not change the test. That was right then. On this head `own` is read signed — which is how nonpositive diagonals get dropped — so applying that nit now would silently strand every upper-bound weak row, with the whole crate still green. My mistake to have left it unqualified.

## The fixture

`degenerate_weak_upper_bound.rs` is `degenerate_weak_curvature.rs` reflected through `x = 0`: the same two kinks at curvature 1 and 1e4, bounds `-10 <= x <= 0`, optimum at the upper bound. Three tests — the weak rows are upper-bound rows, they leave the bound by the exact step, they hold on the holding side.

Mutation-checked to the standard the invariance legs are held to:

| run | result |
|---|---|
| this head, full crate | 204 passed, 0 failed |
| `own = col[k]` (one `sign(k)`) | 1 failed: `an_upper_bound_kink_holds_on_the_holding_side`, alone |
| control, `KAPPA_MIN = -1e300` (never drop) | the two `degenerate_weak_set.rs` tests fail — the fixtures do reach the rule |

The holding side is the guard rather than the leaving side because a wrongly dropped row keeps its *plain* movement, and here that movement is through the bound: `-A1 dp` with `dp < 0` is positive and the bound is above. The mutation does not cost precision, it returns an infeasible step.

`solver.rs` is untouched — the mutation was applied and reverted, and the diff is empty.

## Where this leaves the PR

The rework answers the review. The curvature band is withdrawn, my two-curvature reproducer is `degenerate_weak_curvature.rs`, `kappa = sigma * S_kk` is curvature-free with an exact value at the object it tests for, and there is no branch on activity class — so the gh#756 lesson about certified-row early returns, and open #763, do not apply to this head. All nine invariance legs pass against current `main`, including the ambiguous-kink leg.

Once CI is green on the merged head I'll re-review and merge.

---

jkitchin:
@devin-griff I think this will be ready to merge when CI is green. Take a look at the last push and let me know if you agree.

devin-griff:
Agreed, and verified locally on the merged head: the full crate passes 204, your one-sign mutation fails exactly `an_upper_bound_kink_holds_on_the_holding_side` and nothing else, and the pyomo suite passes 472 on a fresh build of this head. CI is green on all eleven checks.

The gap is real and the fixture is the right shape for it. Every drop-rule decision to date ran at sign plus one, and the reflected fixture is the cheapest way to keep the minus branch load-bearing. Point taken on the earlier nit as well: the double multiply now has a test that says why it is not redundant.

Ready to merge from my side.

PR #759, opened 2026-08-23
benchmarks: read the POUNCE time limit from its stamp, and rerun mittelmann at 1800s

### LINKED ISSUE #581: benchmarks: mittelmann compares a 300s POUNCE arm against an 1800s Ipopt reference
## Summary

The mittelmann suite compares a **300s** POUNCE arm against an **1800s** Ipopt
reference. One instance in the 1,326-model corpus is decided by that gap, and
it happens to be the only Ipopt-exclusive solve in the whole report.

This is now disclosed in `BENCHMARK_REPORT.md` (#580) rather than hidden, but
the underlying asymmetry is still there.

## Why the limits differ

`benchmarks/ipopt_ma57.provenance.json` carries a per-suite override:

    "mittelmann": { "timelimit": 1800, "generated": "2026-08-07",
                    "threads": "pinned to 1 (OMP/OPENBLAS/VECLIB/RAYON)" }

The reference was regenerated at 1800s because `max_cpu_time` counts CPU
summed across threads, so unpinned multithreaded runs burned a 300s budget in
~90s wall and truncated six instances. The fix was correct. But the POUNCE arm
has no time-limit flag at all — `run_nl_bench.sh` wraps it in
`timeout $BENCH_TIMELIMIT` (default 300) — so a plain `make benchmark-rerun`
leaves the two columns on different clocks for this suite.

Notably, the raised ceiling was only *needed* by one of the four instances it
flipped. `henon120` (108.5s -> 77.5s), `lane_emden120` (89.3s -> 69.7s) and
`qcqp1000-2c` (104.2s -> 129.7s) all now finish inside 300s; thread pinning
fixed those on its own.

## The affected instance

`WM_CFy` was a **mutual** timeout before the reference was regenerated
(`e157b036~1`: Ipopt `Maximum_CpuTime_Exceeded`, 340.5s, 157 iters). It became
an Ipopt-only solve by raising one column's ceiling, not by any solver change.

Measured out of band at 1800s (same host, threads pinned, binary from
`ce41b5bc`):

| | POUNCE | Ipopt ma57 |
|---|---|---|
| status | `Optimal Solution Found` | `Solve_Succeeded` |
| wall | 672.8s | 1146.5s |
| iterations | 239 | 556 |
| objective | 1.2234907111354598 | 1.2218235931043862 |

So it is not a capability gap — POUNCE converges in 59% of Ipopt's time and
43% of its iterations.

## Proposed work

Rerun the mittelmann POUNCE arm at `BENCH_TIMELIMIT=1800` so the suite is
like-for-like, then regenerate the report. Expected effect: `WM_CFy` moves
from "Ipopt only" to both-Optimal, and the report's Ipopt-exclusive count
goes to **0**.

Only that one instance can change. The other five POUNCE timeouts in the
corpus (`vanderbei/cresc132`, `drcav3lq`, `drcavty3`, `lpopt/ex10`,
`irish-electricity`) are instances Ipopt fails as well.

Cost is a few hours of machine time, which is why it was deferred out of the
0.10.0 release rather than done inline.

## Separate observation worth its own look

The two objectives on `WM_CFy` differ by **0.136%**, far past the report's
0.01% match threshold. The problem minimizes (`O0 0`), so Ipopt's point is the
better one — POUNCE converges cleanly (KKT error 6.3e-09, constraint
violation 9.4e-10) to a *worse* local optimum. That is a solution-quality
question rather than a benchmarking one, and it may deserve splitting out.

## References

- `dev-notes/research/wm-cfy-timelimit.md` — full write-up
- #580 — the disclosure that surfaced this
- `dev-notes/research/mittelmann-post-546-sweep.md` — the original threading fix

jkitchin:
Reruns the mittelmann POUNCE arm at 1800s so the suite is like-for-like with its
1800s Ipopt reference, and fixes the report machinery that described the limit.

Fixes #581

## The measurement

| | |
|---|---|
| POUNCE | `5dc835f5`, clean tree, release, sha256 `702f586f…` |
| limit | 1800s, **stamped** in `pounce.env.json` |
| threads | OMP / OPENBLAS / VECLIB / RAYON all pinned to 1 |
| corpus | the local mirror, **not** Dropbox |

**47/47 `Solve_Succeeded`.** For mittelmann the report goes to **0
Ipopt-exclusive solves, 6 POUNCE-exclusive**, 41 both-Optimal, 40/41 objectives
matching within 0.01%. That is the headline #581 predicted.

Two setup details that changed the numbers and are worth knowing:

- The corpus was moved off Dropbox first. `scripts/bench_data_root.sh` warns
  that timings taken in a sync folder include the daemon's I/O, and the
  previously published mittelmann numbers were taken with Dropbox in the path.
- `make -C benchmarks mittelmann-rerun` does not work on a fresh checkout: it
  depends on `mittelmann-translate`, which delegates to
  `benchmarks/mittelmann/Makefile` — a file `.gitignore:44` excludes and that
  has never been committed. It dies with "No rule to make target fetch".
  The `.nl` already existed, so the run invoked `run_nl_bench.sh` directly with
  the make rule's exact arguments. Filed separately; not specific to this run.

## What the run does *not* show

#581 predicted only `WM_CFy` could change. Six instances changed against the
previously published column — but that column came from
`perf/feral-refine-target @ 96846891-dirty`, a different dirty branch, so the
comparison confounds binary, limit, corpus location and pinning.

A **same-binary 300s control** isolates the limit exactly:

| instance | 300s | 1800s | attributable to |
|---|---|---|---|
| `WM_CFy` | killed, 98 it | Optimal, 1533.6s, 556 it | **the limit** |
| `nql180` | killed, 48 it | Optimal, 329.8s, 52 it | **the limit** |
| `robot_a/b/c` | Optimal, 23–27s | Optimal, 23–27s | the binary |
| `NARX_CFy` | Optimal, 166.0s | Optimal, 162.1s | the binary |
| `qcqp1000-1nc` | Optimal, 16.4s | Optimal, 16.1s | the binary |

Exactly two instances are decided by the limit. #581's reasoning was sound; it
was the published column that had moved. **The `robot_a/b/c` improvement is a
solver change and should not be cited as an outcome of #581.**

`nql180` is limit-attributable only *because* the binary made it ~6x slower
(54.5s published → 329.8s here), pushing it past a 300s cap it used to clear
comfortably.

Byproducts of the control, both recorded in the note: this solver is
**deterministic** under pinned threads (all 45 instances finishing under 300s in
both arms took identical iteration counts), and the **wall-clock noise floor on
this host is 0.9% median / 1.3% mean**. Any future claim on this suite worth
less than ~2% on a multi-second instance is measuring the machine.

## The code change

`time_limit_note()` claimed the POUNCE limit "is not stamped in the results and
is inferred here from the longest run that was killed". It has been stamped
since `run_nl_bench.sh` started writing `<suite>/pounce.env.json`, which
`read_env_stamp()` already loads for the threading note.

The inference also fails precisely in the case this PR creates: it goes blind
when a suite stops timing out, because with no kills there is nothing left to
infer from. A suite at genuine parity would have printed no POUNCE limit at all.
It also cannot tell a harness kill from a solver-side timeout.

Now it prefers the stamp, keeps the inference only for older results, and when
the limits agree says so and stops hunting for instances "decided by the gap" —
at equal limits there is no gap, and a POUNCE timeout there is a real capability
difference that the suite's ordinary timeout accounting already reports:

> POUNCE ran this suite at the same 1800s (recorded in `pounce.env.json`), so
> both columns are held to the same clock here.

## What is deliberately not here

**`BENCHMARK_REPORT.md` is not regenerated.** Only `benchmarks/mittelmann/pounce.json`
exists in this checkout; every other suite's results are gitignored and absent,
so regenerating rewrites the committed 1,326-model report as a 47-model
mittelmann-only file (−440 lines, plus the three `figures/profile_*.png`). That
trade is not worth one suite's update. The committed report picks up the parity
at the next full sweep, and this fix means that sweep will state the POUNCE
limit from the stamp instead of guessing.

Raw arm, env stamp, control run and the mittelmann-only render are kept outside
the repo (`pounce.json` is gitignored by design); paths are in the note.

Full record: `dev-notes/research/mittelmann-1800s-parity.md`.

https://claude.ai/code/session_01Kz7rXUtFXAtRX25CTgK55a

### DIFF
--- a/benchmarks/benchmark_report.py
+++ b/benchmarks/benchmark_report.py
@@ -373,7 +373,24 @@ def _followup_for(prov, name):
     return (prov.get('ipopt_followups') or {}).get(name)
 
 
-def time_limit_note(prov, comps):
+def _pounce_limits(env_stamps):
+    """Per suite, the time limit the POUNCE arm was actually given.
+
+    `run_nl_bench.sh` stamps the `timeout` it wrapped the solver in into
+    <suite>/pounce.env.json, so this is a record rather than an inference.
+    Keyed by suite display name, matching `read_env_stamp`'s caller.
+    """
+    limits = {}
+    for name, stamp in (env_stamps or {}).items():
+        if not stamp:
+            continue
+        lim = stamp.get('timelimit')
+        if isinstance(lim, (int, float)):
+            limits[name] = lim
+    return limits
+
+
+def time_limit_note(prov, comps, env_stamps=None):
     """Build the Time limits note, disclosing any per-suite Ipopt override.
 
     The base provenance stamp carries one `timelimit`, but
@@ -383,21 +400,40 @@ def time_limit_note(prov, comps):
     did not actually run under, and hides that a suite may compare a
     300s POUNCE arm against a 1800s reference. So: print the base, print
     every override, and name any instance the asymmetry decides.
+
+    The POUNCE limit is *recorded*, not guessed: `run_nl_bench.sh` writes it
+    into <suite>/pounce.env.json. This used to be inferred from the longest
+    run that was killed, which had two failure modes — it read a genuine
+    solver-side timeout as evidence of the harness limit, and it went blind
+    exactly when the suite stopped timing out, since a run with no kills
+    leaves nothing to infer from. Prefer the stamp; fall back to the
+    inference only for results predating it (gh#581).
     """
     base = prov.get('ipopt_timelimit')
     overrides = prov.get('ipopt_suite_overrides') or {}
     if base is None and not overrides:
         return []
 
-    lines = [f"> **Time limits.** The saved Ipopt reference ran at "
-             f"`max_cpu_time` = {base}s unless overridden below. The POUNCE arm "
-             "carries no time-limit flag — it is wrapped in "
-             "`timeout $BENCH_TIMELIMIT` (default 300s) and a kill is recorded "
-             "as `Maximum_CpuTime_Exceeded` — so its limit is not stamped in the "
-             "results and is inferred here from the longest run that was killed."]
-
     cutoffs = _timeout_cutoffs(comps)
-    by_lower = {s.lower(): s for s in cutoffs}
+    limits = _pounce_limits(env_stamps)
+
+    if limits:
+        how = ("The POUNCE arm carries no time-limit flag — it is wrapped in "
+               "`timeout $BENCH_TIMELIMIT` and a kill is recorded as "
+               "`Maximum_CpuTime_Exceeded` — but the limit in force is stamped "
+               "per suite into `<suite>/pounce.env.json` and is read from there "
+               "below.")
+    else:
+        how = ("The POUNCE arm carries no time-limit flag — it is wrapped in "
+               "`timeout $BENCH_TIMELIMIT` (default 300s) and a kill is recorded "
+               "as `Maximum_CpuTime_Exceeded` — and these results carry no "
+               "`pounce.env.json` stamp, so its limit is inferred here from the "
+               "longest run that was killed.")
+
+    lines = [f"> **Time limits.** The saved Ipopt reference ran at "
+             f"`max_cpu_time` = {base}s unless overridden below. {how}"]
+
+    by_lower = {s.lower(): s for s in set(cutoffs) | set(limits)}
 
     affected = []
     for suite, ov in sorted(overrides.items()):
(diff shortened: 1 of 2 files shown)


jkitchin:
Follow-up on the issue's "Separate observation worth its own look" — the
`WM_CFy` solution-quality gap. **It has closed on this binary.**

#581 recorded the two solvers landing 0.136 % apart, with POUNCE converging
cleanly (KKT 6.3e-09, constraint violation 9.4e-10) to the *worse* local
optimum, and suggested splitting that out as its own question.

Measured on the 1800 s arm in this PR (`5dc835f5`, threads pinned):

| | objective | vs Ipopt |
|---|---|---|
| Ipopt ma57 (saved reference) | 1.2218235931043862 | — |
| POUNCE, #581 era | 1.2234907111354598 | 0.1364 % |
| **POUNCE, this run** | **1.2218307805290987** | **0.0006 %** |

The report's match threshold is 0.01 %, so the two now agree on that instance —
it is counted among the 40/41 matching objectives rather than as the single
mismatch. Constraint violation on the new point is 1.2e-11.

This is a solver change rather than anything to do with the time limit: the
same-binary 300 s control in this PR reaches the same objective, just truncated.
So the split-out question #581 proposed appears to need no follow-up issue,
unless you want one tracking *why* it moved.

Found while running the Mittelmann option-tuning study on top of this arm.

PR #762, opened 2026-08-24
test(sens): invariance legs over two kink fixtures

jkitchin:
## Problem

Sensitivity defects keep shipping through a corpus that is uniform in exactly
the dimension the defect lives in. Three examples, all merged:

| | defect | what the corpus was uniform in |
|---|---|---|
| `205bb67` | the corrector added the algorithm's **scaled** iterate to a step and bounds in the model's units; at a variable factor of 2 it returned an answer eight orders worse than the step it was given, with the residual halved so nothing warned | unit scaling — "which coincide only at unit scaling, **which is every fixture it had**" |
| gh#672 f4 | an acceptance tolerance was absolute on a quantity that scales with the perturbation, so a `1e-10` step cleared feasibility everywhere and the holding side's derivative read `-1` instead of `0` | one perturbation magnitude |
| gh#672 f1 | `weakly_active_bounds` read a full-x index as a var-x factor row, so with a fixed variable ahead of a kink the detection pinned or released the kink variable's **neighbour** (the gh#450 hazard) | no `make_parameter`-removed variables |

Each was found by a reviewer, not by a test. There is no leg that sweeps any of
these dimensions over a degenerate base point.

The gap is live, not historical — see **Blast radius**, where the fourth leg
catches a first-order defect on the current head of #756.

## Cause

`variable_scaling_sensitivity.rs` already runs the scaling leg, and already
states the rule:

> Sigma and the ratio are BARRIER quantities, proportional to the mu the run
> stopped at … **So the invariant is Sigma/mu, not Sigma.**

But it asserts that only about `classify_activity`, and its fixture has **no
weakly active bound**. I added the obvious row — `weakly_active_bounds` must be
unmoved by the change of variables — and it passes vacuously, comparing `[]`
with `[]`. The degeneracy surface has never been under a scaling leg at all.
Same for perturbation magnitude, for a fixed column ahead of the kink, and for
the activity class the engagement rule branches on.

## Fix

A new `crates/pounce-sensitivity/tests/sens_invariance_legs.rs`: two kink
fixtures, swept down three dimensions, the way `sweep-fixtures.sh` sweeps the
exact and L-BFGS Hessians.

**Fixture 1** — a decoupled kink, certified `WEAKLY_ACTIVE`:

```text
min  0.5 k² − A·p·k + 0.5 (w − 2)²   [+ 0.5 (xf − 0.7)²]
s.t. p = 0,  0 ≤ k ≤ 10,  0 ≤ w ≤ 10  [, xf ≡ 1.5]
```

`k` is a genuine kink — on its lower bound with a multiplier that vanishes with
`mu`, derivative `A` where the pin moves up and `0` where it moves down. `w` is
interior and decoupled: it must never be called weak and must never move. `xf`
is the optional `make_parameter`-removed column that puts full-x and var-x out
of step in front of everything interesting.

**Fixture 2** — a *coupled* kink, which lands in `AMBIGUOUS`:

```text
min  0.5·h·k² + c·k·y + 0.5·m·y² − A·p·k
s.t. p = 0,  0 ≤ k ≤ 10,  y free
```

`classify` divides `sigma` by the Hessian's **diagonal** `H_ii`, but a kink's
multiplier is generated by the curvature *reduced* along that coordinate:
eliminating `y` leaves `reduced = h − c²/m` and `sigma = reduced`, so the ratio
is `reduced/diagonal` and equals 1 only when the coordinate is **decoupled**.
Driving `c²/(h·m)` toward 1 puts a genuine kink below the `1e-1` band edge. On
a collocation model that coupling is the normal case, not a corner.

| leg | sweeps | asserts |
|---|---|---|
| 1 — scaling | `user-scaling` factors `[1e-2, 5, 1e3]` vs none | weak set and directional derivative unmoved |
| 2a — magnitude, certified | `delta` over 1e-2 → 1e-10, both sides | derivative independent of step size |
| 2b — magnitude, ambiguous | same sweep, fixture 2 | same, in the other activity class |
| 3 — fixed variable | leading fixed column present/absent | weak set and derivative unmoved |

Leg 2 runs over both fixtures because the rule that engages a weak row
*branches on the activity class*. A rule exact in one class and length-based in
the other is invisible to a corpus carrying only the certified one.

Both arms of every scaling leg run under `user-scaling`, so the only difference
is whether the TNLP hands factors back — the option, the objective factor and
the row factors are identical, which is the same discipline
`variable_scaling_sensitivity.rs` uses.

### What the legs compare, and what was rejected

**Not `dx / delta`.** That was the first attempt and it is wrong: the
parametric step is affine in `delta`, not linear. It carries a base-point term
of order `mu` — here `9.1e-11`, the same vector at every `delta` — because the
converged iterate sits that far off the exact solution and the step corrects
it. Dividing by `delta` inflates that term until, at `1e-10`, it *is* the whole
answer, and leg 2 fails on clean code for a reason that has nothing to do with
the leg's name.

So the legs compare **slopes**: a difference quotient between two perturbations
on the same side of the kink, which cancels the constant and leaves the
one-sided derivative. `the_step_is_affine_in_delta` pins that constant's size
and its constancy across magnitudes, so if it ever stops being negligible that
test says so instead of a leg failing obliquely.

Each leg also checks both arms against the **exact** derivative, not only
against each other, so a shared error cannot pass the pair off as agreement.

## Blast radius

Test-only. No `src/` change, so nothing in the solver or the sensitivity layer
moves; the CLAUDE.md addition documents the legs the way the fixture sweep is
documented.

**Leg 2b catches a live first-order defect on #756.** Reported in full at
[#756 (comment)](https://github.com/jkitchin/pounce/pull/756#issuecomment-5389742881);
in short, that PR gives a row the classifier certifies `WEAKLY_ACTIVE` a slack
of zero and applies a *measured* slack to the `AMBIGUOUS` class. That is safe
only if a genuine kink is always certified, and per fixture 2 it is not. Below
the measured slack the kink stops engaging and the holding side returns the
*leaving* side's answer:

| reduced | ratio | class | `main` | #756 `9b4a919` | exact |
|---|---|---|---|---|---|
| 1.0 | 1.0 | WEAKLY_ACTIVE | 0 | 0 | 0 |
| 1e-1 | 0.1 | WEAKLY_ACTIVE | 0 | 0 | 0 |
| 1e-2 | 0.01 | **AMBIGUOUS** | 0 | **110.0** | 0 |
| 1e-3 | 0.001 | **AMBIGUOUS** | 0 | **1100.0** | 0 |

On `9b4a919` leg 2b is the **only** one of the twelve tests that fails, which
is the diagnostic separation these legs exist to buy: the red leg names the
dimension that broke, not the model. The other eleven pass there, and all
twelve pass on `main`.

The earlier revision of #756 (`16f6df1`, the withdrawn `Sigma` membership band)
failed the scaling legs instead; that version has since been withdrawn and
membership reverts to `main` byte for byte.

## Tests

12 tests. They pass against `main` at `32fea00` — that is the point of an
invariance leg, and it is **not** the template's "fails on the parent commit"
case: these are guards, not regression tests for a fix, so there is no parent
commit they fail on. The equivalent evidence is a mutation check per leg —
reintroduce the historical defect and confirm the matching leg, **and only that
leg**, goes red:

| mutation | reintroduces | result |
|---|---|---|
| `report.var_status.get(full_of[var_row])` → `get(var_row)` | gh#672 f1 | both fixed-variable legs fail; scaling and magnitude legs stay green |
| `let tol0 = band * scale_of(&d0)` → `let tol0 = band`  (and the same for `tol`) | gh#672 f4 | magnitude legs fail — holding side reads `-1.1e-3` where `0` belongs; weak-set legs stay green |
| (no mutation) the legs against #756 `16f6df1` | the withdrawn membership band | both scaling legs fail; magnitude, fixed-variable and affine legs stay green |
| (no mutation) the legs against #756 `9b4a919` | the live measured-slack defect | leg 2b alone fails; the other eleven stay green |

Preconditions are asserted rather than assumed, so no leg can quietly go
vacuous the way the existing scaling fixture did:

- `the_fixture_carries_a_kink_and_an_untouchable_interior_variable` — `k` is on
  its bound and in the weak set, `w` is not, and the two sides genuinely
  disagree.
- `the_coupled_fixture_carries_an_ambiguous_kink` — the bound is weak, the
  class really is `AMBIGUOUS`, and the ratio really is `reduced/diagonal`. A
  classifier change that moves it fails loudly rather than letting leg 2b pass
  against the wrong branch.
- `leg_fixed_the_index_spaces_actually_diverge` — `n_full_x() == 4` while
  `x_primal_rows(&[0,1,2,3]) == [None, Some(0), Some(1), Some(2)]`, i.e. the
  kink's full-x index 1 really is var-x row 0.

**Not pinned, deliberately:** the corner where a leg's own fixture is large
enough to exercise the 62k-scale resource paths (gh#672 f2's `2^n` masks,
gh#708 f4's full-length basis columns). Those need a fixture too large to
vendor into a unit test, and both were found by profiling on the collocation
model rather than by a corpus.

---

## The rest of the plan — @devin-griff

Tiers 3 and 4 as you framed them, with your additions folded in; Tier 2
downgraded on your evidence.

**Tier 3 — a `sens-review` skill.** Building it next, with your line: *an
absolute length compared against a homogeneous quantity breaks linearity even
when the length is measured rather than chosen*. Leg 2b is that exact class
caught in the wild, so the checklist entry gets a worked example rather than a
maxim.

**Tier 2 — you were right, and I verified it rather than taking the
correction.** I reproduced your mutation (`w.row` → `w.var_row` at the
directional caller): seven legs fail, every one with
`SensComputationFailed("directional derivative: released sigma unavailable")`.
Tracing it: `active_set_sigma_x` resolves each released row through
`rows.iter().find(|b| b.row == r)?`, the reverse direction hits
`sigma.get_mut(var_row)?`, and all four call sites (`corrector.rs:428`,
`algorithm_backsolver.rs:523` and `:593`, `solver.rs:1050`) turn the `None`
into an error or a failed back-solve. So the swap cannot pass silently, and my
body's argument from the `will not be caught here` comment overstated it —
that comment is true of the line it sits on, not of the function.

The useful rule is narrower than either of our framings: **table lookup fails
loudly, direct array indexing fails silently.** `released` rows are resolved
through a lookup table, so a swap errors. gh#672 f1 was
`report.var_status.get(var_row)` — a direct index into an array — which
silently returns a *neighbouring variable's* answer. That, not "two spaces in
one signature", is where a newtype earns its cost. Agreed on deferring it
behind Tier 4.

**Tier 4 — a re-solve verification leg.** Agreed it is the highest value per
hour, and both practicalities are taken: steps above the barrier width, warm
start from the base iterate. Leg 2b is an argument for it — the 62k validation
measures rms at 1% and 10%, both far above the slack where #756 is still
correct, so an oracle hung only on large perturbations would have missed this
too. Worth hanging it on a *span* of step sizes rather than a representative
one. I'll take your suggestion of this fixture plus one where the step crosses
a strongly active bound, so it covers `fix_relax`.

**Fixture — kept fresh, as you argued.** Fixture 2 is a second standalone one
for the same reason: one varied thing per leg is what makes the mutation table
possible.

---

- [x] Tests fail on the parent commit for the stated reason — **N/A** for
      invariance legs (nothing is being fixed, so there is no parent commit
      they fail on). Discharged instead by the per-leg mutation checks in
      **Tests** above: reintroduce the historical defect, and the matching leg
      — and only that leg — goes red
- [ ] `CHANGELOG.md` `[Unreleased]` entry — deliberately none: test-only, no
      user-visible behaviour change. Say the word if you'd still like one
- [ ] Book page under `docs/src/` — N/A; the legs are documented in CLAUDE.md
      alongside the fixture sweep, which is where the sweep's own contract lives
- [x] `cargo fmt --all -- --check` clean; `cargo clippy -p pounce-sensitivity
      --all-targets -- -D clippy::correctness -D clippy::suspicious` exits 0;
      `cargo test -p pounce-sensitivity` green (21 binaries, 0 failures, 12
      tests in this file). Full-workspace `cargo test` was **not** run locally
      — this sandbox ran out of disk building it — and is left to CI. The
      change adds one integration test file to `pounce-sensitivity` and edits
      CLAUDE.md, so it cannot reach another crate's tests, but flagging the gap
      rather than ticking the box for a run that did not happen
      as it stands now

The fixtures trip `clippy::unwrap_used` (`used unwrap() on a Result`) the same
way the sibling sensitivity test files do; that lint sits outside what CI
denies, and the gate above exits 0.

https://claude.ai/code/session_014ZLf5Ph2UWtr8tDpUqKocD


devin-griff:
I read the test file and ran the legs against the current #756 head, `9b4a919e`: all ten pass. The review there withdrew the band and reverted membership, which is exactly the frame-dependence your scaling legs probe, so the two PRs are compatible. Merging this as-is looks right, and I agree no CHANGELOG entry. A bonus worth noting: your scaled arm gives the kink an effective curvature of 1e4, so the legs also cover the stiff-kink case from the #756 review.

**Tier 2: lower priority than your framing suggests, and I checked rather than guessed.** I mutated the directional caller to pass var-x rows as `released`: six legs fail immediately. Every consumer of a released row resolves it against the bound-row table and errors on a miss, and the two spaces never overlap in range, since bound rows sit above the primal and constraint blocks. So the swap cannot pass silently today and your leg 3 catches it loudly. The scoped newtype buys compile-time detection of an already-loud runtime failure, plus insurance against a refactor that loses the lookup guard. I would take it as opportunistic cleanup after #756, not schedule it ahead of Tier 4.

**Tier 3: build it.** One line to add from the #756 round: an absolute length compared against a homogeneous quantity breaks linearity even when the length is measured rather than chosen. My slack term was exactly that, and the scale-invariance frame is what caught it.

**Tier 4: highest value per hour of the four, agreed.** It is the methodology the 62k-model validation already runs out-of-repo. Two practicalities: the oracle's steps must sit above the barrier width, since below `sqrt(mu)` the warm re-solve and the directional-derivative contract genuinely diverge, which is the boundary the #756 scale episode mapped, and the re-solve should warm-start from the base iterate, for cost and to stay in the same basin. I would hang it on this PR's kink fixture plus one fixture where the step crosses a strongly active bound, so the oracle also covers the fix_relax release test.

**Fixture: keep the fresh one.** One varied thing per leg is what made your mutation table possible, and the existing degeneracy fixtures each encode one historical defect and would need restructuring to serve the legs. The #756 round made the same point in miniature: the interior-coordinate fixture there had to be rebuilt around an equality-mediated coupling, and being standalone is what made that safe.

jkitchin:
The plan discussed above now has a durable home, so it does not stay as prose in a PR body that stops being read once this merges:

- **#764** — sensitivity hardening roadmap: the re-solve oracle (with both of @devin-griff's design constraints, plus "hang it on a span of step sizes, not a representative one" from what leg 2b found), the `sens-review` checklist, and the scoped newtypes with the table-lookup-vs-array-indexing rule that came out of verifying his correction. Nothing assigned — that is the point of filing.
- **#763** — the root cause leg 2b surfaced: `classify_activity` normalizes a variable's `Σ` by the Hessian *diagonal* rather than the reduced curvature, so a coupled kink reads `AMBIGUOUS`. Filed separately because it outlives #756 and is visible through the public API. Severity stated narrowly there: `main` returns the *correct* derivative for those cases, so it is a misleading label rather than a live miscomputation — but it is the label #756 trusted.

Also pushed `964f596`, which records in CLAUDE.md the general lesson rather than only the instance: **a leg is only evidence about the branch its fixture reaches.** These legs passed on #756's head while its defect was live, because their kink certifies `WEAKLY_ACTIVE` and the rule under review gave certified rows an early return — the code actually being changed was never executed. Worth stating as a contract, since "the guard was green" is exactly the kind of thing that gets trusted later.

An internal asymmetry worth noting from tracing #763, which suggests the diagonal is an oversight rather than a choice: the classifier's **row** path already normalizes by a directional curvature, `∇dᵀH∇d / ‖∇d‖²` (`activity.rs:502`, `:553`), while the **variable** path passes the bare diagonal (`activity.rs:410`).

---

PR #771, opened 2026-08-24
feat(convex): ActiveSetSession — persistent active-set driver over QpProblem

### LINKED ISSUE #769: Add `ActiveSetSession` in `QpProblem`
I think it would be a good idea to add a persistent `ActiveSetSession` over `pounce_convex::QpProblem`.
The session should own the `pounce-qp` translation and preserve the behavior of `solve_qp_active_set` (possibly including):

- Presolve and postsolve,
- Screening,
- Standard defaults,
- Status/certificate verification, 
- Parametric reuse when valid,
- Cold fallback when reuse is invalid.

I am not sure if other behavior is applicable.
This would prevent every downstream frontend from reimplementing the private convex-to-native translation (including oximo).

I tried to do this a few days ago but I remembered there was still some parameters in the CLI that needed to be removed and imported to `pounce_convex`, but maybe it is fine now.

jkitchin:
Fixes #769.

## Problem

Two things were dropped on the floor at every call to `solve_qp_active_set`.

The **translation** from `pounce_convex::QpProblem` to `pounce_qp::QpProblem` —
the `+1` index shift, the `[A_eq ; G]` row stacking, the `±1e19` free-bound
convention, and the dual sign transform on the way back — lived in locals
inside `solve_translated`. Nothing was public, so a frontend wanting the
active-set engine over a convex `QpProblem` had to restate it. A restatement
that gets the sign transform wrong still returns something that looks like a
multiplier, which is the failure mode worth caring about.

The **previous solve** went with it. `pounce-qp`'s headline capability is
parametric reuse, and the convex path could not reach it at all: every
convex-path solve was cold, even when the caller was sweeping one parameter
over an otherwise fixed model.

Measured on the new `active_set_session` example — an 8-step path over
`n = 40`, one parameter moving, release build, 4-core VM:

|                    | total    | per step |
|--------------------|----------|----------|
| cold, per step     | 104.4 ms | 13.1 ms  |
| through a session  |  19.3 ms |  2.4 ms  |

All 7 reusing steps trace the homotopy — asserted, not assumed — with status
and objective identical to the cold solve on every step. A repeat run reads
111.4 vs 24.2 ms.

## Fix

`pounce_convex::ActiveSetSession` — a persistent handle that owns the
translation, the presolve/postsolve wrapper, and the previous solve, and
traces `QpSolver::solve_parametric` from it when the next problem is
eligible. Re-exported from `pounce-rs` as `convex::ActiveSetSession`.

The translation is an owned public type, `ActiveSetQp`, and so is the rest of
the recipe around it: `screen_variable_box` / `BoxScreen` ahead of it,
`engine_options` for the settings this path was measured under, and
`back_translate_verified` (read back → re-derive the verdict → reject
non-finite fields → relabel a deadline crossing) after, with `back_translate`
and `verify_status` exported for callers that need the pieces.
`solve_qp_active_set` is now a thin cold wrapper over the same parts the
session uses, which is what `session_cold_matches_free_function` pins.

Everything the issue listed is covered: presolve/postsolve, the variable-box
screen on both paths, standard engine defaults, status/certificate
verification, parametric reuse, cold fallback.

**The reuse rule is one attempt, then the cold ladder.** Eligibility beyond a
shape match is left to `solve_parametric`'s own guards — deliberately *not*
re-adding the `A`/box guard that gh#602 measured and declined; re-adding it
here would have quietly reversed a decision that was made with numbers. A
warm answer is verified against the problem as posed by the same
`verify_status` the cold path runs — the engine's `Optimal` is re-derived, not
propagated — and is reported only if it stands up. Otherwise the full cold
ladder (screen → unscaled → Ruiz retry → simplex-seeded retry) runs and owns
the answer. So reuse changes the cost, never the verdict.

**`last_reuse()` and `stats()` name the route the engine took, not the call
that was made.** `solve_parametric` has three exits: the traced path, a
decline that falls back to the previous working set, and — when the previous
solve is not a usable base — a cold solve inside the call. All three return an
ordinary conclusive `Ok`. `QpStats::parametric_source` now stamps which one
ran, and the session reports it (`Reuse::Homotopy` / `WorkingSet` /
`EngineCold`, with `SessionStats` broken out the same way). See the review
round below — this was a defect, not a nicety.

Rejected: caching the translated `QpProblem` and mutating it in place across
steps. It saves the rebuild, but it makes the session's state and the caller's
model two sources of truth for the same numbers, and the failure is silent
when they drift. Rebuilding per step is a small fraction of the win above.

## Review round (@GermanHeim — two findings raised, three fixes)

1. **The session counted every conclusive `solve_parametric` return as
   parametric reuse** (2e739cd). Worse than reported: the third exit is a
   *fully cold* solve, also labelled `Reuse::Parametric`. The claim was
   unfalsifiable — a family that never traced a single path reported perfect
   reuse. Fixed by propagating the engine's actual source rather than
   re-deriving the guards in `pounce-convex` (a second copy of them would
   drift silently). `parametric_accepted` is gone rather than renamed: as a
   stored counter it read as "the homotopy ran this often", which it never
   measured. Making it falsifiable is also what lets the measurement above say
   *homotopy* rather than *accepted*.
2. **`ActiveSetQp` shipped the forward translation public and the read-back
   crate-private** (2e739cd), so an external driver (oximo, the case in the
   issue) could build and solve the native problem and then had to restate the
   dual sign transform, the objective reconstruction and the verification —
   the three parts that fail silently.
3. **The screen ahead of the translation was still crate-private** (2632ca3),
   found by auditing the whole recipe rather than only the three functions
   named in 2. `screen_variable_box` is documented as something *every* solve
   entry point runs, and the public path was one that could not. Skipping it
   is not a lost optimization: a reversed box reaches `pounce_qp`'s `validate`
   as a hard `InvertedBounds` error where the driver certifies
   `PrimalInfeasible` (gh #491, which used to panic through PyO3), and a
   *present* `+∞` lower bound is dropped as if absent and comes back `Optimal`
   at a point violating it (gh #295) — the silent wrong answer, one step
   earlier than finding 2.

Screening deliberately stays outside `from_convex`: that translates the
problem *as posed*, and a version that quietly snapped a crossed box would
hide the decision. The four-step recipe (screen → translate →
`engine_options` + solve → `back_translate_verified`) is documented on
`ActiveSetQp::from_convex`, in the `pounce-rs` module docs and in
`docs/src/rust.md`, along with what it deliberately does *not* reproduce: the
cold driver's retry ladder (Ruiz, simplex seed, feasibility probe), which is
what `solve_qp_active_set` and `ActiveSetSession` are for.

## Blast radius

`scripts/sweep-fixtures.sh` is **byte-identical to the merge base `dcebdc5`**
— both legs (`exact`, `lbfgs`), 142 fixture-legs each, at the default
`solver_selection=auto` *and* pinned at `solver_selection=qp-active-set` so
the convex-QP fixtures actually reach this driver rather than the
interior-point arm. Status, objective, iterations and engine all unmoved.

Re-baselined after merging `main`, and not only re-run: #772 changed what the
convex path reports, and the sweep records exactly that, so a comparison
against this branch's original parent (`20be289`) would no longer have
isolated my diff from theirs. Four legs — base and head, `auto` and pinned —
with the binaries built from `dcebdc5` and from this head. (For the record the
corpus did not move between `20be289` and `dcebdc5` either, so the earlier
per-commit sweeps still say what they said.)

That is the expected result and not a strong one on its own: no existing caller
constructs a session, so the cold path is all the corpus can exercise. The
equivalence it does pin is that the refactor did not move the cold path, which
is the part every current frontend runs.

`QpStats` gains a public field. Additive, but it breaks exhaustive struct
literals — called out in the CHANGELOG. #772 added none, which is why the
merge is clean in substance and not just textually.

No frontend is rewired onto the session in this PR. The CLI's `run_convex_qp`
interleaves presolve with its own logging, the IPM branch and the debug hook;
`pounce-py`'s `solve_qp(method="active-set")` has no presolve at all today, so
adopting the session there is a *behaviour* change that needs its own
justification and its own sweep. Both are follow-ups; this PR is the thing they
would adopt.

On the issue's last note — CLI parameters still needing to move into
`pounce_convex` — that is already resolved upstream:
`QpOptions::try_from_options_list` and `ActiveSetOverrides::try_from_options_list`
live in `pounce-convex`/`pounce-qp`, so the session needs nothing from the CLI.

## Tests

`crates/pounce-convex/tests/issue769_active_set_session.rs`, 15 tests, plus
`parametric_source_names_the_route_actually_taken` in `pounce-qp`. The ones
that carry weight:

- `session_cold_matches_free_function` — every field of the solution against
  `solve_qp_active_set`, not just status and objective. This is what makes the
  refactor a refactor.
- `warm_solve_matches_cold_across_a_parameter_sweep` — the 8-step family, each
  step also solved cold and compared, and now asserting `homotopy_accepted`
  rather than a route-blind count.
- `a_declined_homotopy_is_reported_as_a_working_set_reuse` — the review case
  (`P`'s diagonal 2 → 4), which the first version of this PR reported as a
  traced path.
- `parametric_source_names_the_route_actually_taken` — all three exits of
  `solve_parametric` pinned at the source.
- `an_external_caller_can_translate_solve_and_read_back` — public API only, no
  session and no free function, starting from the box screen, so the exported
  surface is pinned from outside the crate.
- `the_box_screen_is_reachable_and_decides_what_the_engine_cannot` — a wide
  crossing certified infeasible with no solve behind it, a hairline crossing
  repaired and solved.
- `infeasible_member_is_reported_infeasible_from_a_warm_session` — a same-shape
  member that reaches the warm path and still comes back certified infeasible,
  rather than inheriting the previous step's optimality.
- `rejected_warm_verdict_falls_back_to_the_cold_ladder` — drives the fallback
  branch directly.

**These tests cannot fail on a parent commit — each calls an API that does not
exist there, so they do not compile.** That is the honest statement for a
feature PR, and it holds for the review-round tests too: on 525b605 the
variant `Reuse::WorkingSet` did not exist to assert. What can be checked
against that commit is the *behaviour*, and it is the defect: `solve_engine`
set `Reuse::Parametric` and incremented `parametric_accepted` for any
conclusive attempt, unconditionally, so the `2 → 4` scenario reported a traced
path there — which is what @GermanHeim reproduced independently.

Deliberately not pinned: the speedup itself. The example asserts *agreement*
across the sweep and that every attempt traced the path, and prints the
timings; it does not assert a ratio, because a wall-clock threshold in CI is a
flake generator on shared runners.

## Checks

On the merged head (`22ce1b6`): `cargo test -p pounce-convex -p pounce-qp`
green across 53 test binaries — 52 plus #772's new `convex_timing_rows.rs`,
which is the interaction worth naming, since `QpStats::parametric_source`
would have broken any exhaustive literal that PR added. `cargo fmt --all --
--check` clean; CI's clippy gate (`-D clippy::correctness -D
clippy::suspicious`) clean on `pounce-qp`, `pounce-convex` and `pounce-rs`;
`pounce-rs` doctests green. `pounce-cli`'s suite was last run green at
2632ca3's parent — CI covers it on the merge.

Not a full `--workspace` run: this container ran out of disk for one. The
crates above are the ones the diff can reach.

---

- [x] Book page updated — `docs/src/rust.md`, already linked from `SUMMARY.md`
- [x] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean (scope as stated above)
- [x] Every claim in this body is true of the code as it stands now — re-read
      against the final diff after each review commit and after the merge,
      including the timing table and the sweep baseline
- [ ] Tests fail on the parent commit for the stated reason — **N/A, see Tests**


jkitchin:
@GermanHeim can you check if this fixes the issue you posted?

GermanHeim:
I took a quick look. This is a great fit for my needs, but some issues:

First, `solve_parametric` can internally decline the homotopy and fall back to `solve_with_working_set`. However, this code counts every conclusive return as `parametric_accepted` and reports `Reuse::Parametric`.
I reproduced this with two same-shaped QPs whose Hessian changed from 2 to 4. This declined the homotopy, but the session reported `Reuse::Parametric` and `parametric_accepted = 1`.
We could propagate the engine's actual source.

Second, `ActiveSetQp` is public, however, only the forward translation is public, since `back_translate` and `verify_status` are crate-private.
An external caller (e.g., oximo) can construct and solve the native problem, but must reimplement the dual conversion, objective reconstruction, and status verification.

jkitchin:
Both findings were real, both fixed in 2e739cd. Thanks — the first one was a claim my own tests were asserting.

## 1. Every conclusive return counted as parametric reuse

Confirmed, and worse than the case you reproduced: `solve_parametric_scoped` has **three** exits. A guard decline or a tracer that returns `Ok(None)` falls through to `solve_with_working_set`; a previous solve that is not `Optimal` (or whose working set does not fit) falls through again to a plain `self.solve(qp_new, None, opts)` — a fully cold solve, which the session was also reporting as `Reuse::Parametric`.

Fixed the way you suggested, by propagating the source rather than re-deriving it: `QpStats::parametric_source` is `Some(Homotopy | WorkingSet | Cold)` on any `solve_parametric` return and `None` everywhere else, stamped at each exit (including the deadline stub, which reused nothing and says `Cold`). Re-deriving it in `pounce-convex` would have meant keeping a second copy of `solve_parametric`'s guards, which drift silently by construction.

The session reports what the engine said: `Reuse::Homotopy` / `WorkingSet` / `EngineCold`, `Reuse::is_warm()` for the coarse question, and `SessionStats` broken out the same way — `homotopy_accepted`, `working_set_accepted`, `engine_cold_accepted`, with `attempts_accepted()` and `warm_accepted()` derived. `parametric_accepted` is gone rather than renamed: as a stored counter it read as "the homotopy ran this often", which it never measured.

Two things fell out of making the claim falsifiable:

- Your `H` 2 → 4 case is now `a_declined_homotopy_is_reported_as_a_working_set_reuse`, and `parametric_source_names_the_route_actually_taken` pins all three exits at the source in `pounce-qp`.
- The PR's own measurements re-ran clean: the sweep family and the `n = 40` example both report `homotopy_accepted == parametric_attempts` — 5/5 and 7/7 — so the numbers in the PR body were genuine traces, not mislabelled working-set fallbacks. The example asserts that now rather than trusting it; a timing table that quietly measured a different route would be the same defect one level up.

Worth stating: `Reuse::EngineCold` is currently unreachable **through the session**, since `remember` only keeps a base the engine and the driver both called `Optimal` and the shape guard covers the dimensions. It exists anyway — that decision belongs to the engine's guards, which can change — and `the_session_never_reports_a_route_the_engine_did_not_take` pins the current state so a future engine change surfaces there instead of as an inflated reuse count.

## 2. `ActiveSetQp` shipped the forward translation only

Also confirmed, and it undercut the point of the issue: `back_translate`, `verify_status` and `engine_options` were all `pub(crate)`, so an external caller could build and solve the native problem and then had to restate the dual sign transform, the objective reconstruction and the verification — the three parts that go wrong *silently*.

All three are public now, plus `back_translate_verified`, the composition: read back → re-derive the verdict → reject non-finite fields → relabel a deadline crossing. That exists because the ordering is not optional — `back_translate` returns a *provisional* `Optimal`, and a caller who stops there has propagated the engine's claim unchecked, which is exactly what `verify_status` is for (`QSC205` returns `Optimal` at the wrong objective; `DUALC1` certifies a feasible QP infeasible). Two calls with a rule about their order is an API that reads as complete when it is half-applied.

`an_external_caller_can_translate_solve_and_read_back` is written the way oximo would write it — integration test, public API only, no session and no free function — so the surface is pinned from outside the crate rather than asserted about it. `pounce-rs` re-exports all four.

Still not public: the cold driver's second rung on a rejected infeasibility claim (`feasibility_probe`, which needs a backend and another solve). That is documented on `back_translate_verified` rather than left to be discovered.

## Verification

Both sweep legs byte-identical to `20be289`, at `solver_selection=auto` and pinned at `solver_selection=qp-active-set`. `pounce-qp` (105 unit + suites), `pounce-convex` (241 + 14 here), `pounce-cli`, `pounce-rs` doctests, `cargo fmt --check` and the clippy gate all clean.

The `QpStats` field is additive but breaks exhaustive struct literals, so the CHANGELOG calls it out as a breaking change for anyone constructing one by hand.

---

jkitchin:
Follow-up on finding 2 — it was not fully fixed by 2e739cd. Auditing the recipe an external caller actually runs, rather than only the three functions you named, turned up a fourth piece one step earlier. Fixed in 2632ca3.

`screen_variable_box` and `BoxScreen` were still crate-private, so the public path — `ActiveSetQp::from_convex` → solve → `back_translate_verified` — could not run the variable-box screen. The screen's own doc comment says *"Every solve entry point runs this"*, and that path is now one.

Same shape as your finding, and the same severity — skipping it does not cost performance, it costs correctness:

- A **reversed** box (`lb > ub`) reaches `pounce_qp`'s `validate` as `InvertedBounds` — a hard `Err` — where the driver reports a certified `PrimalInfeasible`. (On the active-set path this used to abort the process outright: the simplex-seeded last resort clamped a seed into the inverted interval and `f64::clamp` panicked, which crossed the PyO3 boundary as a `PanicException`. gh #491.)
- The **impossible** class — a *present* `+∞` lower bound — is dropped as if absent, and the solve returns `Optimal` at a point that violates it. gh #295. That is the silent-wrong-answer case, which is exactly what your point about the dual sign transform was about one function later.

Screening deliberately stays outside `from_convex`: that is a translation of the problem *as posed*, and a `from_convex` that quietly snapped a crossed box would hide the decision. So the recipe is four steps, documented on `ActiveSetQp::from_convex`, in the `pounce-rs` module docs and in `docs/src/rust.md`:

1. `screen_variable_box` — `Empty` is a certified `PrimalInfeasible` with no solve; `Snapped` hands back a repaired copy to solve instead.
2. `ActiveSetQp::from_convex` on whatever step 1 returned.
3. `engine_options`, then solve `ActiveSetQp::problem`.
4. `back_translate_verified`.

What the recipe deliberately does **not** reproduce is the cold driver's retry ladder — Ruiz equilibration, the simplex-seeded retry, the objective-free feasibility probe. Those are salvage rungs rather than correctness gates, and wanting them is a reason to call `solve_qp_active_set` or `ActiveSetSession`. Said explicitly in the docs rather than left to be discovered, since that is the same trap in miniature.

Tests: `an_external_caller_can_translate_solve_and_read_back` now starts with the screen, and `the_box_screen_is_reachable_and_decides_what_the_engine_cannot` pins both classes — a wide crossing certified infeasible with no solve behind it, a hairline crossing repaired and solved.

While re-checking finding 1 I also verified the one route that could still have been mislabelled: `solve_homotopy` returns `Ok(Some(..))` only after the path reaches `t = 1`, using `solve_with_working_set` as the corrector on the discovered set, and `Ok(None)` when it stalls short. So `Homotopy` is honest there, and a stalled path falls through to `WorkingSet` like a declined one.

Verification for this commit: both sweep legs byte-identical to `20be289` again (`auto` and pinned `qp-active-set`), `pounce-convex` + `pounce-qp` green across 52 test binaries, `pounce-rs` doctests, clippy gate and `cargo fmt --check` clean.

---

PR #777, opened 2026-08-25
Phase-envelope fold refinement, sensitivities, and inverse design

### LINKED ISSUE #773: Tutorial/research case: differentiable phase-envelope tracing and inverse mixture design
> Migrated upstream from [bernalde/pounce#1](https://github.com/bernalde/pounce/issues/1). Existing implementation: [fork PR #6](https://github.com/bernalde/pounce/pull/6).

## Objective

Turn the existing Peng–Robinson phase-envelope notebook into a reproducible chemical-engineering case study that does two things which are usually implemented as separate, specialized workflows:

1. trace a complete multicomponent phase envelope through folds and the critical region; and
2. differentiate that solution to solve an **inverse mixture-design problem** (for example, minimally change composition so that a specified pressure-reduction path remains single phase, or so that the cricondenbar/cricondentherm moves to a target).

The intended paper claim is not merely that POUNCE can solve one flash calculation. It is that a differentiable nonlinear solver plus arclength continuation can replace a brittle collection of pressure/temperature marching, critical-point location, finite-difference sensitivity, and outer design scripts with one coherent workflow.

## What already exists in POUNCE

[`python/notebooks/34_phase_envelope_peng_robinson.ipynb`](https://github.com/jkitchin/pounce/blob/main/python/notebooks/34_phase_envelope_peng_robinson.ipynb) is a strong starting point. It already contains:

- a natural-gas mixture with a Peng–Robinson equation of state;
- a differentiable closed-form cubic-root calculation;
- the Michelsen incipient-phase equations as a square feasibility NLP;
- a useful failure comparison: a pressure ladder fails at a fold, ordinary parameter continuation can collapse onto the trivial `K_i = 1` branch, while arclength continuation traces dew → critical → bubble branches;
- extraction of the cricondenbar and cricondentherm from turning points; and
- a demonstration that the continuation tangent contains phase-boundary sensitivity information.

The current notebook is therefore more than a toy, but it is not yet a defensible tutorial or inverse-design result. Turning points are only localized to the continuation step size, the physical branch is guarded manually, composition derivatives are not validated, and there is no outer design calculation.

## Why this is a worthwhile chemical-engineering case

Complete phase envelopes contain folds, critical points, retrograde regions, and nonphysical/trivial roots. These are exactly the features that make naive repeated flash calculations cumbersome. A good case study would expose four POUNCE capabilities in one recognizable thermodynamics problem: nonlinear feasibility, pseudo-arclength continuation, factorization-based implicit sensitivity, and constrained inverse design.

This case is attractive for a paper only if the derivative-enabled design step is central. Complete phase-envelope construction itself is mature technology; the contribution should be the transparent, end-to-end transition from tracing to verified sensitivities and design.

## Literature landscape

The tutorial and paper should place the result against the following primary literature:

- Michelsen, **“Calculation of phase envelopes and critical points for multicomponent mixtures”** (1980), introduced a foundational algorithm for complete envelopes and the location of critical points and pressure/temperature extrema. This is the baseline against which the formulation and terminology should be checked. [DOI: 10.1016/0378-3812(80)80001-X](https://doi.org/10.1016/0378-3812(80)80001-X)
- **“Phase Envelope Construction for Mixtures with Many Components”** develops a practical procedure for saturation curves, quality lines, extrema, and the difficult critical region in high-component-count mixtures. It is a useful reference for the validation set and for deciding which computations belong in a production workflow. [DOI: 10.1021/ef700251k](https://doi.org/10.1021/ef700251k)
- Nikolaidis et al., **“Phase-envelope construction for complex mixtures using a bead-spring model”** (2016), is a relevant continuation-based alternative and emphasizes robustness to difficult envelope geometry rather than simple one-parameter marching. [DOI: 10.1002/aic.15064](https://doi.org/10.1002/aic.15064)
- **“Calculation of phase envelopes of fluid mixtures through parametric marching”** gives a modern, generic EOS/mixing-rule implementation and a useful independent description of numerical marching and critical-region handling. [Open-access article](https://pmc.ncbi.nlm.nih.gov/articles/PMC7756196/)
- DiffThermo demonstrates differentiating thermodynamic models, including equilibrium calculations, with respect to model parameters. It supports the value of differentiability but also makes clear that automatic differentiation alone is not the scientific novelty. [arXiv:2102.10705](https://arxiv.org/abs/2102.10705)
- Ben Hicham et al., **“Differentiable Thermodynamic Phase Equilibria for Machine Learning”** (2026), shows that differentiable phase-equilibrium layers are now an active research area. The POUNCE case should distinguish itself through branch-safe continuation, critical extrema, and constrained inverse design rather than claiming differentiable equilibrium as new. [DOI: 10.1021/acs.jcim.6c00728](https://doi.org/10.1021/acs.jcim.6c00728)

### 1. Extract a reusable, tested thermodynamics problem

- Move the Peng–Robinson mixture calculations and incipient-phase residuals out of notebook-only cells into a small reusable example module.
- Make component constants, overall composition, and binary interaction parameters explicit inputs.
- Define the phase/root-selection conventions and units in one place.
- Add residual-level tests for material consistency, fugacity equality, and the pure-component/near-pure-component limit where appropriate.

This does not need to become a general thermodynamics library. The goal is a transparent benchmark model whose equations can be audited.

### 2. Replace sampled extrema with solved extrema

- Use the arclength tangent to bracket `dP/ds = 0` and `dT/ds = 0` events.
- Refine each event with an augmented nonlinear solve, rather than reporting the nearest continuation point.
- Add step-size refinement tests demonstrating that the reported cricondenbar and cricondentherm converge to tolerance.
- Treat the critical neighborhood explicitly; document how the formulation detects or approaches `K_i → 1` without confusing the critical point with an arbitrary trivial solution.

### 3. Make physical-branch protection an explicit part of the example

- Detect collapse to the `K_i = 1` branch away from the critical neighborhood.
- Verify the chosen liquid/vapor cubic roots and reject mechanically valid but thermodynamically inadmissible states.
- Evaluate whether a tangent-plane-distance/stability check is needed for the selected validation mixtures. If it is omitted, state the domain in which the root guard is sufficient.
- Record branch events and rejected trial points so the tutorial teaches *why* continuation succeeds or fails.

### 4. Validate implicit sensitivities

Compute sensitivities of at least the following outputs:

- cricondenbar pressure and its temperature;
- cricondentherm temperature and its pressure; and
- one phase-boundary temperature or pressure at a specified operating condition.

Differentiate with respect to:

- `N-1` independent composition coordinates on the composition simplex; and
- at least one Peng–Robinson binary interaction parameter `k_ij`.

Compare the implicit derivatives against carefully converged re-solves/central differences. The check must retrace the same physical branch, not merely solve an unconstrained nearby flash.

### 5. Add a genuine inverse-design problem

Implement at least one of these, with the first preferred for the main narrative:

- **single-phase operating-path design:** minimize a weighted composition change subject to a decompression/cooling path remaining outside the two-phase envelope with a specified margin;
- **envelope-extremum targeting:** minimize blend cost subject to lower/upper bounds on the cricondenbar or cricondentherm; or
- **robust blend design:** impose those constraints under a small set of `k_ij` or feed-composition scenarios.

After optimization, retrace the *entire* designed envelope and independently verify the operating-path margin. A local linearized prediction by itself is not enough.

### 6. Benchmark the numerical value of the combined workflow

For the same cases, compare:

- independent pressure or temperature ladder solves;
- ordinary parameter continuation;
- pseudo-arclength continuation; and
- continuation plus implicit sensitivities versus finite-difference retracing for the design derivatives.

Report nonlinear iterations, failed points/restarts, factorizations if exposed, derivative error, and wall time. Stamp every table with the POUNCE commit, model-data revision, tolerances, and hardware/runtime environment.

## Validation cases

Use at least two mixtures:

1. the existing natural-gas-like mixture, because it exhibits a fold and retrograde behavior; and
2. an independently published mixture with enough data to compare critical/extremum values or a reference envelope.

One case should be small enough for a fast regression test. Reference numbers must come from an auditable source or an independent thermodynamics implementation, with the EOS/mixing-rule conventions reconciled before comparison.

- [ ] Reusable Peng–Robinson/envelope example code with residual and regression tests.
- [ ] A notebook or user-guide chapter that traces dew, critical, bubble, and retrograde regions without manual point editing.
- [ ] Cricondenbar and cricondentherm located by refined event solves and shown to converge with step-size refinement.
- [ ] Physical/trivial-branch guards documented and exercised by an intentional failure example.
- [ ] Composition and `k_ij` sensitivities checked against full branch retraces to an agreed tolerance.
- [ ] At least one constrained inverse-design result, followed by independent full-envelope verification.
- [ ] At least two validation mixtures, one suitable for CI.
- [ ] A commit-stamped comparison of ladder, ordinary continuation, and arclength workflows.
- [ ] Clear statement of local/nonconvex limitations; no claim that every thermodynamic branch or global blend optimum is found.

## Scope guard

Do not turn this issue into a general-purpose property package. Use a small, inspectable Peng–Robinson implementation and focus the contribution on branch-safe continuation, verified sensitivities, and inverse design. General branch switching and arbitrary multiphase equilibrium can be follow-on work if the two-mixture study shows they are needed.

bernalde:
> Upstream submission of [bernalde/pounce#6](https://github.com/bernalde/pounce/pull/6); tracks #773.

## Problem

Issue #773 asks the phase-envelope tutorial to become a quantitative, reusable design workflow rather than a plot whose extrema are limited by the pseudo-arclength step size.  The existing notebook embedded its Peng–Robinson model, assumed every `k_ij = 0`, sampled the cricondenbar/cricondentherm from the trace, and had no path for composition sensitivities, an inverse design, an independent published mixture, or CI regression coverage.

The ordinary pressure ladder also illustrates why the additional machinery matters: it stops at 97.00 bar, while the refined natural-gas cricondenbar is 97.586395 bar.  Prescribed-parameter continuation is worse here: it reports success after landing on the exact but physically vacuous `K_i = 1` branch.

Independent oracle:

| quantity | published | POUNCE | absolute error |
|---|---:|---:|---:|
| methane/propane maxcondentherm | 282.530000 K | 282.525286 K | 0.004714 K |

The reference is Deiters and Bell, *AIChE Journal* 65 (2019), e16730, DOI 10.1002/aic.16730, using their methane/propane constants, `x_methane = 0.8`, and `k_12 = 0.042823`.

## Cause

`PathTrace.turning_points` detects a fold from a tangent-sign reversal, so its coordinate is necessarily one trace sample.  It supplies a bracket, not a converged extremum or a design derivative.  The notebook also had no reusable parameterization of the composition simplex or binary interactions, so a derivative or inverse design would have duplicated the thermodynamic model and had no independent verification surface.

## Fix

- Add the example-scoped `pounce.examples.phase_envelope` module with validated Peng–Robinson mixture data, differentiable fugacity/residual functions, low-pressure anchors, pressure/temperature pseudo-arclength traces, local residual/root/normalization/trivial-branch diagnostics, and explicit units/provenance.
- Refine a traced extremum by solving the augmented simple-fold equations `F = 0`, `F_x v = 0`, and `v^T v = 1` with POUNCE.
- Parameterize composition with all `N-1` unconstrained log ratios plus one selected `k_ij`; expose the implicit extrema derivatives in physical pressure/temperature units.
- Add a fixed-pressure implicit-sensitivity example checked against fresh central re-solves.
- Add a local inverse-composition NLP that targets an extremum while keeping every iterate inside the composition simplex.
- Extend notebook 34 with refined natural-gas extrema, the independent Deiters–Bell benchmark, central validation in which every perturbation retraces and refines a complete envelope, and an inverse design accepted only after a fresh low-pressure anchor, trace, and fold refinement.
- Document the scientific limits: largest/smallest-`Z` and branch guards are local checks, not a global TPD stability test; the augmented equations assume a simple fold; and the inverse NLP has no global-optimality certificate.

The source is commit `fefa6ec3`; notebook outputs were generated from that exact commit and recorded separately in `f14146ac`.

## Blast radius

This is an additive Python example/tutorial change.  It does not change the Rust solver, termination logic, public optimization APIs, or existing examples.  SciPy is already a required Python dependency.

The new test is the only CI-cost increase, and it is substantial: measured on GitHub Actions, the `Python tests (jax)` job's `Run Python tests` step goes from **6m06s** on base `dc103aa` to **18m44s** on `f14146ac` — about +12m38s, or 3.1x — because the module performs real continuation and fold solves rather than using fixtures derived from the expected answer.  That makes `Python tests (jax)` the critical path of `ci.yml`, taking the whole run from ~10m30s to ~20m47s.  Tracked for a follow-up trim in #788.

## Tests

- `python -m pytest python/tests/test_phase_envelope_example.py -q` — **4 passed in 416.41s** on the contributor's machine.
- Executed all 18 code cells of `34_phase_envelope_peng_robinson.ipynb` — **no error outputs; all assertions passed**.
  - natural-gas refined extrema residuals: `4.16e-14` and `3.55e-15`
  - independent reference error: `0.004714 K`
  - maximum relative error versus four complete central-perturbation retraces: `1.07e-08`
  - inverse design after a completely fresh retrace/refinement: `3.02e-11 K` from the 280 K target
  - pure n-butane saturation pressures: `1.034`, `2.573`, `9.469` bar at 273.15 / 300 / 350 K (literature ~1.03, ~2.58, ~9.42)
- `ruff format --check` and `ruff check` over the new module and test — clean.
- `git diff --check origin/main...HEAD` — clean.
- Full CI green on `f14146ac`: [run 32992430900](https://github.com/jkitchin/pounce/actions/runs/32992430900), 11/11 jobs.

This is a new feature, so there is no meaningful parent-commit regression test: the parent has neither the example module nor its test file.  No Rust files changed, so the Rust formatting/clippy/test checklist item is not applicable.

---

- [x] Existing book page under `docs/src/` updated; no new `SUMMARY.md` entry is needed
- [x] Every claim in this PR body and in the code comments is true of the code as it stands now — no test named that does not exist, no doc describing a design that was revised mid-review, no measurement whose baseline has since moved

Closes #773


jkitchin (changes requested):
Read the full `phase_envelope.py` module, `test_phase_envelope_example.py`, all 40 notebook cells with their committed outputs, the docs/CHANGELOG diffs, and the `pounce.jax` APIs this builds on.

The physics and the validation strategy are good — see the closing section. Two things need fixing before merge, one of which puts a visibly wrong number in the notebook.

## Blocking

### 1. `vapor_pressure()` returns its bracket endpoint — the notebook's only pure-component validation is wrong by ~30×

`python/pounce/examples/phase_envelope.py:1053`. Committed output of notebook cell 7:

```
pure n-butane vapor pressure (equal fugacity), vs literature:
  T =  273.15 K   Psat = 35.000 bar   (lit ~1.03 bar)
  T =  300.00 K   Psat = 35.000 bar   (lit ~2.58 bar)
  T =  350.00 K   Psat = 35.000 bar   (lit ~9.42 bar)
```

35.000 bar is exactly `bracket_pa[1] = 3.5e6`. It is identical at every temperature because the residual `ln φ_v − ln φ_l` is **identically zero** wherever the PR cubic has a single real root — `cubic_real_root(..., largest=True)` and `largest=False` return the same value there, bitwise, since `jnp.where(three, three_root, one)` selects `one` on both branches. At 273–350 K, 35 bar is deep in that region for n-butane, so `f(hi) == 0.0` and `brentq` returns `hi` on its first check without ever iterating.

Reproducing the residual in plain Python from the module's own cubic and pure-component fugacity formulas:

| T (K) | 1e5 Pa | 2.5e5 Pa | 1e6 Pa | 3.5e6 Pa |
|---|---:|---:|---:|---:|
| 273.15 | −3.20e−2 | +8.24e−1 | **0.0** | **0.0** |
| 300.00 | −8.95e−1 | −2.64e−2 | +1.09e0 | **0.0** |
| 350.00 | −2.06e0 | −1.17e0 | +4.17e−2 | **0.0** |

The true root is bracketed and is right — at 273.15 K the sign change sits just above 1e5 Pa, i.e. ≈1.03 bar, matching the literature value printed in the same cell. The thermodynamics is fine; the root-find is broken.

The search has to stay where the cubic has three distinct roots. The module already owns the right guard — `PhasePointDiagnostics.roots_are_admissible` / `Z − B` — it just isn't applied here. Minimally: scan down from `hi` for the first pressure with `Z_v − Z_l > 0` and use that as the upper bracket, and raise rather than silently return an endpoint when the residual there is zero.

On how this survived: notebooks aren't executed in CI here and I'm not asking for that. But `vapor_pressure` isn't notebook code — it's a public function in the example module, exported in `__all__`, shipped in the wheel, and this PR already adds `python/tests/test_phase_envelope_example.py` for exactly that module. Three temperatures asserted against literature is ~10 lines and milliseconds next to the 143 s that file already costs. Cell 7 only *prints*, which is why "no error outputs; all assertions passed" is true and the number is still wrong.

### 2. `CHANGELOG.md` cites the fork's issue number

The entry ends `(#1)`. In this repo, #1 is "Enable feral's `solve_refined` / `solve_many_refined` in `FeralSolverInterface`" (closed, not planned) — unrelated. Every other entry in the file uses the upstream number. Should be `(#773)`.

## Should fix

### 3. The tests assert that continuation *fails*, in a specific way

`python/tests/test_phase_envelope_example.py:227-228`:

```python
assert verification_trace.status == "corrector_failed"
assert verification_trace.n_steps > 145
```

plus the terminal-point assertions that follow, which depend on where the break landed. `trace_arclength` breaks the first time the Newton corrector misses `newton_tol` within `newton_max` (`python/pounce/jax/_path.py:665`); where that happens is a function of the LAPACK backend and jaxlib build, and CI installs `jax[cpu]` unpinned and runs the whole directory. A path-follower improvement, or a different backend shifting the break by a step, turns this red for a reason unrelated to what the test means. Assert the physically meaningful thing — the trace got past the target fold and every point on it is admissible — not the failure mode and the step count. The same two lines appear in notebook cell 38.

## Minor

**4.** `factor_reuse=False` is a no-op on the path taken (`phase_envelope.py:952`, `:1036`). `batched_solve_with_jacobian` pins the factor unconditionally — "Always pin for the duration of the Jacobian build so the back-solve resolves regardless of `factor_reuse`" (`_problem.py:2520`). `refine_fold` only calls `solve_with_jacobian`, so the flag doesn't select the dense path it appears to. Drop it, or comment why it's there.

**5.** `_fold_initial_guess` only ever finds a maximum (`:886`, `argmax(theta[reversals])`). Correct for cricondenbar/cricondentherm and all this example needs, but `refine_fold`'s docstring says "a traced extremum" with no way to pick — a trace with a minimum fold silently refines the wrong one. One sentence, or a `select=` argument. Also `:883` brackets with `<= 0.0` while `reparameterize_trace` at `:642` uses `< 0.0`; harmless today, reads as a bug later.

**6.** PR body provenance is stale: "The source is commit `3562e97a`; notebook outputs were generated from that exact commit." Cell 1's committed output stamps `70b3d64`, and `0f6cac0` changed the module after `3562e97a`. The outputs themselves are current — diffing cell sources across `3870f24` → `141975b`, only cell 39's markdown moved — so this is body accuracy rather than stale outputs, but checkbox 3 claims otherwise.

**7.** CI cost: ~143 s on every PR, since `python-test` runs `pytest python/tests` wholesale. Disclosed in the body, and it is a genuine end-to-end regression test. Just noting the third test does 5 full traces + 6 fold refinements, and the ds=0.10 / ds=0.08 convergence pair duplicates a check the notebook already publishes.

## What's good

The independent Deiters–Bell benchmark (0.0047 K), the four-output validation against full central retraces at ~1e−9 relative, the fresh-retrace verification of the inverse design (3.02e−11 K), and the explicit "local guards, not a global TPD stability test" caveat are all well above the usual bar for a tutorial. The β convention in `envelope_residual` is right, and cell 26's demonstration that dew and bubble are one curve through the critical point resolves the obvious worry about the hardcoded `largest=True` / `largest=False` root selection — K → 1 there, so the two roots coincide and the fixed selection degenerates gracefully rather than picking the wrong phase.

No Rust or solver change, so CLAUDE.md's fixture-sweep requirement doesn't apply.

---

bernalde:
Addressed the requested changes on `f14146ac` (with current upstream `main` merged at `dc103aa2`).

### Review findings

1. **Pure-component vapor pressure:** `vapor_pressure()` now scans within the distinct-root region, rejects inadmissible/zero-residual endpoints, and raises when no two-phase saturation root exists. The regression checks n-butane at 273.15, 300, and 350 K against 1.03, 2.58, and 9.42 bar; the computed values are 1.034, 2.573, and 9.469 bar. A supercritical case is also required to fail loudly.
2. **Changelog issue:** corrected the fork reference from `#1` to upstream `#773`.
3. **Continuation assertions:** replaced the backend-sensitive `corrector_failed`, step-count, and terminal-point assertions in both the test and notebook with physical invariants: the trace must cross the temperature fold, descend afterward, and keep every traced point residual-small, root-admissible, and phase-normalized.
4. **No-op option:** removed both `factor_reuse=False` arguments.
5. **Fold selection contract:** documented that `refine_fold()` selects the traced maximum and does not support minima, and aligned the reversal test from `<= 0` to `< 0`.
6. **Provenance correction:** this comment supersedes the PR body's historical `3562e97a` provenance. The freshly rebuilt native extension and notebook execution used source commit `fefa6ec3`; the output-only commit is `f14146ac`. All 18 code cells executed, no error outputs were present, and cell sources were unchanged during execution.
7. **CI cost:** no end-to-end validation was removed. The focused module took 416.41 s on this machine, so the review's cost observation remains valid and is now reported with the current measurement.

### Verification

- Full Python suite with exact-head release CLI on `PATH`: **1190 passed, 21 skipped** in 975.31 s.
- Focused phase-envelope module after the final upstream merge: **4 passed** in 416.41 s.
- Newly merged upstream second-opinion module: **7 passed**.
- Ruff formatting check and lint: passed.
- Documentation consistency: 47 pages reachable, no dead links.
- Executed notebook: all checks passed; max trace residual `9.91e-10`, augmented fold residuals `4.16e-14` and `3.55e-15`, Deiters–Bell error `0.0047 K`, sensitivity error `1.07e-08`, and inverse-design error `3.02e-11 K`.

GitHub Actions has **not run** for this head: [CI run 32992430900](https://github.com/jkitchin/pounce/actions/runs/32992430900) is `action_required`. I am therefore not claiming CI is green. The formal review decision also remains `CHANGES_REQUESTED` until the maintainer updates it.

There were no inline review threads, so this top-level summary covers findings 1–7. @jkitchin, the blocking and actionable findings are addressed; please re-review when convenient.

jkitchin (approved):
All seven findings addressed on `f14146ac`. Verified against the code rather than the summary; this supersedes my `CHANGES_REQUESTED`.

**1. `vapor_pressure` — confirmed fixed, independently.** The new implementation excludes the single-root region (`vapor_z > liquid_z` and both `> b_reduced`), scans geometrically down from the upper limit for a strict sign change, and raises instead of returning an endpoint. I re-derived the algorithm in a separate pure-Python implementation of the module's own cubic and fugacity formulas and got **1.0339, 2.5730, 9.4692 bar** at 273.15 / 300 / 350 K — matching the committed notebook cell 7 output to every printed digit, against literature 1.03 / 2.58 / 9.42. The supercritical guard raises at 450 K. `test_n_butane_vapor_pressure_matches_literature` puts it in the module test file, which is where it belongs.

**2.** `CHANGELOG.md:13` now reads `(#773)`.

**3.** `corrector_failed` and `n_steps > 145` are gone from both the test and notebook cell 38, replaced by physical invariants: the fold is interior to the trace, temperature descends after it, and every traced point is residual-small, root-admissible, and normalized. The new `diagnose_phase_trace` helper is `jax.vmap`-ed, so the all-points check costs one trace rather than one per point — which is the obvious way a fix of that shape could have backfired, and didn't.

**4.** Both `factor_reuse=False` arguments removed.

**5.** `refine_fold`'s docstring now states that it selects the traced maximum and does not support minima; both reversal tests are `< 0.0`.

**6.** I edited the PR body directly rather than leaving the correction in a thread comment — the body is what survives into the merge commit. Provenance now reads `fefa6ec3` / `f14146ac`, the test line reads 4 passed, and the two notebook figures that had moved (`4.16e-14` / `3.55e-15`, and `1.07e-08`) are updated to the current outputs.

**7. CI cost is real and is now tracked in #788.** Measured on Actions, same job and step: `Run Python tests` goes 6m06s on base `dc103aa` → 18m44s on `f14146ac`, about +12m38s (3.1x). That makes `Python tests (jax)` the critical path and takes the whole `ci.yml` run from ~10m30s to ~20m47s. I'm accepting it here rather than blocking, because the end-to-end shape is what found the `vapor_pressure` defect in the first place — but a doubled critical path on every PR in the repo is not free, and #788 names the candidate trims and the properties any trim has to preserve.

CI green on this head: 11/11 in [run 32992430900](https://github.com/jkitchin/pounce/actions/runs/32992430900).

---

PR #778, opened 2026-08-25
docs(pyomo): add closed-loop active-set NMPC tutorial

### LINKED ISSUE #774: Tutorial/research case: closed-loop active-set-aware advanced-step NMPC for a constrained CSTR
> Migrated upstream from [bernalde/pounce#2](https://github.com/bernalde/pounce/issues/2). Existing implementation: [fork PR #7](https://github.com/bernalde/pounce/pull/7).

## Objective

Turn the existing active-set-aware sensitivity notebook into a complete **closed-loop advanced-step nonlinear model-predictive-control (asNMPC) tutorial** for a constrained exothermic CSTR.

The intended result is a reproducible demonstration that POUNCE can solve the background optimal-control problem, reuse its KKT factorization for a fast measurement update, follow active-set changes without producing inconsistent controls, and safely fall back to a fresh NLP solve when the local correction is no longer trustworthy.

The paper-level question is not whether sensitivity-based NMPC exists—it has a mature literature—but whether POUNCE makes a technically difficult workflow inspectable and usable end to end, including active-set events, degeneracy, residual guards, and closed-loop consequences.

## What already exists in POUNCE

[`python/notebooks/36_active_set_parametric_sensitivity.ipynb`](https://github.com/jkitchin/pounce/blob/main/python/notebooks/36_active_set_parametric_sensitivity.ipynb) already provides the core numerical ingredients:

- a direct transcription of the Hicks–Ray exothermic CSTR with concentration and temperature states and two manipulated variables;
- a 100-interval finite-horizon control problem;
- a held nominal solve followed by several measurement perturbations;
- plain linear sensitivity/clamping, `fix_relax`, and active-set `path` modes;
- active-set-change records and corrector iterations that reuse the held factorization; and
- a directional degeneracy example.

That is the difficult inner kernel, but it remains an open-loop sensitivity experiment. A tutorial must add the controller around it: plant propagation, horizon shifting, asynchronous/background preparation, disturbances and measurement error, correction acceptance tests, fallback solves, and closed-loop performance/latency metrics.

## Why this is a worthwhile chemical-engineering case

Fast NMPC updates are often presented using sensitivities that assume a fixed active set. Chemical processes routinely invalidate that assumption because actuator bounds, temperature constraints, and state/path constraints switch as measurements and disturbances arrive. Simply clamping a sensitivity-corrected control can violate the linearized KKT system and may hide a bad state trajectory.

This case can show something unusually concrete: the same factorization can provide a rapid update *and* expose exactly when the set of binding process constraints changes. The tutorial should connect solver internals to closed-loop engineering outcomes rather than report only single-update timings.

## Literature landscape

- Zavala and Biegler, **“The advanced-step NMPC controller: Optimality, stability and robustness”** (2009), formalized the advanced-step architecture: solve in advance using a predicted state, then apply a fast sensitivity correction when the measurement arrives. Their CSTR study makes this the natural baseline architecture. [DOI: 10.1016/j.automatica.2008.06.011](https://doi.org/10.1016/j.automatica.2008.06.011)
- Pirnay, López-Negrete, and Biegler, **“Optimal sensitivity based on IPOPT”** (2012), describes sIPOPT and the reuse of the KKT factorization for parametric NLP sensitivities, including treatment of active inequalities through a fix/relax strategy. This is the closest methodological comparator for POUNCE's held-factorization path. [DOI: 10.1007/s12532-012-0043-2](https://doi.org/10.1007/s12532-012-0043-2)
- Jäschke, Yang, and Biegler, **“Fast economic model predictive control based on NLP-sensitivities”** (2014), studies an active-set path-following update and an exact-penalty formulation in a detailed CSTR setting. It is essential prior art for the `path` versus `fix_relax` comparison and for deciding what closed-loop evidence is needed. [DOI: 10.1016/j.jprocont.2014.04.009](https://doi.org/10.1016/j.jprocont.2014.04.009)
- **“Advances in sensitivity-based nonlinear model predictive control and dynamic real-time optimization”** surveys the broader sensitivity-update landscape, including practical limits of local corrections and the relation to real-time iterations. [DOI: 10.1016/j.jprocont.2015.02.001](https://doi.org/10.1016/j.jprocont.2015.02.001)
- **“An algorithm for tuning NMPC controllers with application to chemical processes”** is useful for making the final experiment credible as a control study rather than a solver microbenchmark; it discusses tuning and chemical-process implementation of advanced-step NMPC. [DOI: 10.1021/acs.iecr.6b01121](https://doi.org/10.1021/acs.iecr.6b01121)
- **“Advanced-step multistage NMPC: Robustness and stability”** extends the architecture to uncertainty-aware multistage control and reports CSTR and quadruple-tank studies. It motivates a bounded uncertainty/stress campaign without requiring that this first tutorial become a full stochastic NMPC implementation. [Article](https://www.sciencedirect.com/science/article/pii/S0959152419303270)

### 1. Build an explicit controller/plant loop

Separate the following objects in code:

- the controller model and finite-horizon NLP;
- a plant simulator, initially the same model but integrated independently;
- the measurement received at each sample;
- the state predicted when the background solve was launched; and
- the control actually applied after correction and safety checks.

Implement horizon shifting, state/control warm starts, and preparation of the next background problem. The notebook should make the time ordering visible: predict → solve in background → measure → correct → validate → apply → shift.

### 2. Define and implement correction policies

Run the same closed-loop scenarios with at least:

1. full NLP re-solve at every sample (accuracy reference);
2. nominal/background control without a measurement correction;
3. plain sensitivity update plus bound clamping (intentional weak baseline);
4. `fix_relax` correction;
5. active-set `path` correction; and
6. `path` correction with a guarded full-solve fallback.

The policy must validate the *whole corrected primal/dual point*, not only clamp the first control move.

### 3. Add an explicit acceptance/fallback guard

A fast update should be rejected when any agreed condition is exceeded, for example:

- primal feasibility or path-constraint violation;
- stationarity/KKT residual;
- complementarity/active-bound inconsistency;
- correction step or measurement displacement outside a trust threshold;
- failed corrector iteration or ambiguous degeneracy; or
- a predicted closed-loop safety constraint violation.

Document which checks can be evaluated from the corrected NLP state without another factorization. A rejected correction triggers a full solve and resets the background state/factorization.

### 4. Exercise the active set deliberately

Design at least three campaigns:

- **nominal tracking/regulation:** small measurement deviations, few active-set changes;
- **constraint-switching:** feed or temperature disturbance that activates/deactivates coolant, residence-time, and temperature constraints; and
- **stress/model-mismatch:** larger disturbance, parameter mismatch, or measurement bias that forces one or more safe fallbacks.

Record an event timeline containing the sample index, changing constraints, path length/number of breakpoints, corrector residuals, accepted/rejected update, and any full re-solve.

### 5. Include degeneracy as a controlled experiment

Retain the existing directional degeneracy example, but integrate it into a small closed-loop or multi-step test. State exactly which regularity assumption fails, how the directional rule chooses a derivative, and when the controller elects to re-solve instead. The result should not imply that a single sensitivity is mathematically unique at a degenerate point.

### 6. Measure control and computational outcomes

Report, for each policy and scenario:

- integral absolute/squared tracking error or economic stage cost;
- total control movement;
- maximum process/path-constraint violation;
- number of active-set changes;
- number/fraction of fallback full solves;
- update latency (median and a tail statistic such as p95);
- full-solve latency; and
- solver failures/recoveries.

Use an explicitly stated sample-time deadline and report whether each policy meets it. All timing and iteration tables must be stamped with the POUNCE commit, model revision, tolerances, platform, and whether compilation/warm-up time is excluded.

### 7. Keep the example reproducible

- Factor the CSTR transcription and controller loop into reusable example code; keep the notebook focused on explanation and plots.
- Seed noise/disturbance sequences.
- Add a short nominal smoke test and a deterministic active-set-switch regression test.
- Store enough numerical output to detect a change in feasibility or policy behavior without making wall-clock timing a brittle CI assertion.

## Proposed experiment design

Use the current Hicks–Ray formulation first so the new result is attributable to the closed-loop machinery rather than a model rewrite. A reasonable first campaign is 30–60 control samples with the existing horizon and sampling interval, followed by one shorter, higher-stress campaign. If the 100-interval NLP makes CI too slow, preserve it for the paper artifact and use a smaller deterministic transcription for regression tests.

The plant/model mismatch can be modest: a perturbation in reaction-rate or heat-transfer coefficient plus bounded measurement noise is enough to demonstrate repeated updates and fallbacks. Multistage/stochastic NMPC is a follow-on, not an acceptance requirement.

- [ ] Reusable closed-loop controller and independently propagated plant around the existing CSTR NLP.
- [ ] Correct implementation of horizon shifting, warm starts, predicted-state background solve, measurement update, and fallback reset.
- [ ] Side-by-side full-resolve, no-correction, clamped-linear, `fix_relax`, and active-set-path policies.
- [ ] At least nominal, constraint-switching, and stress/model-mismatch campaigns with deterministic inputs.
- [ ] An acceptance guard based on full corrected-point residuals; unsafe/inconsistent updates are never applied silently.
- [ ] Active-set event timeline and a retained directional-degeneracy demonstration.
- [ ] Closed-loop quality, constraint, fallback-rate, and latency results—not just a one-step speedup.
- [ ] Nominal smoke and active-set-switch regression tests.
- [ ] Commit-stamped benchmark artifact and a clear statement that local sensitivity corrections have a limited validity region.

## Scope guard

Keep the first version single-scenario/single-model NMPC. Do not add state estimation, multistage uncertainty trees, or hardware-in-the-loop until the active-set-aware closed-loop workflow is validated. The main comparison is correctness and deadline reliability of local updates versus fresh solves, not a claim that the CSTR model or asNMPC method itself is novel.

bernalde:
> Upstream submission of [bernalde/pounce#7](https://github.com/bernalde/pounce/pull/7); tracks #774.

## Problem

Notebook 36 demonstrated individual held-factor sensitivity updates for the
Hicks–Ray CSTR, but it did not close the control loop. There was no explicit
predicted-state background solve, independent plant propagation, horizon
shift, repeatable disturbance campaign, whole-point acceptance guard, or
closed-loop quality/latency evidence.

Closes #774.

## Cause

The repository had the numerical update kernel but no reusable orchestration
layer connecting prediction, background preparation, measurement, correction,
validation, control application, plant propagation, and the next warm-started
horizon. Consequently, it could not show how active-set changes or
out-of-validity corrections affect repeated controller decisions.

## Fix

- Add `pounce.examples.asnmpc_cstr`, a reusable controller/plant driver for
  the existing direct-transcription CSTR model.
- Compare six policies on identical seeded campaigns: full resolve, no
  correction, clamped linear, fix-relax, active-set path, and guarded path.
- Validate the full corrected primal/dual point using displacement,
  feasibility, stationarity, complementarity, corrector progress, path budget,
  predicted temperature, and manipulated-variable activity checks.
- Retry a failed warm-started NLP once from a cold model, count both the
  failure and recovery, and abort on an unrecovered cold solve rather than
  apply an unverified control.
- Extend notebook 36 with the 100-interval, 30-sample artifact, embedded plots,
  closed-loop tables, an active-set/fallback event timeline, and the
  directional degeneracy experiment.
- Add the notebook dependency/CI wiring, user-guide explanation, changelog
  entry, and deterministic small-model regression tests.

The saved artifact executes 18 campaign/policy combinations. It records a
coolant-bound transition in the constraint-switching campaign and multiple
control-bound transitions in the stress campaign. Guarded updates accept the
nominal campaign and visibly fall back under switching and model mismatch.
Timing tables are stamped with source commit
`d3d96ab9c7e25a09a073bf6d40bc4c071db5bbb2`, model revision, tolerance,
platform, Python version, and warm-up policy.

## Blast radius

This changes only the Python example/notebook, Pyomo adapter tests, notebook
dependency installation in CI, and documentation. It does not change the Rust
solver or existing sensitivity APIs. The paper-scale notebook remains the
artifact; CI uses the same model structure with a smaller deterministic
discretization.

The guard is an inspectable example policy, not a safety certificate. It does
not replace plant interlocks, robust tightening, state estimation, or a
real-time scheduler, and its fallback solves may miss the illustrative
deadline. Those misses are reported rather than hidden.

## Tests

- Parent proof: the new regression file run against `origin/main`
  (`20be2890`) fails during collection with
  `ModuleNotFoundError: No module named 'pounce.examples'`.
- `python/.venv/bin/python -m pytest -q pyomo-pounce/tests/test_asnmpc_example.py`
  — 12 passed with current `pyomo-cvp` 0.7.2.
- The same targeted file with `pyomo-cvp` 0.4.0 first on `PYTHONPATH` —
  12 passed.
- `python/.venv/bin/python -m pytest -q pyomo-pounce/tests` — 463 passed,
  8 skipped.
- Ruff check and format check — clean for the reusable module, package marker,
  and regression file.
- `scripts/check-docs-consistency.sh` — 47 pages reachable, no dead links.
- Frozen notebook lock install reports `pyomo-cvp 0.4.0`.
- Executed notebook validation — 17/17 code cells executed, zero error outputs,
  14 embedded PNG outputs, source cells unchanged by execution, and the exact
  source-commit stamp present.
- Mutation check: inverting the guard's primal-violation comparison makes the
  nominal guarded-policy test fail for unexpected fallbacks; restoring the
  comparison makes it pass.

## Issue 2 acceptance criteria

- [x] Reusable closed-loop controller and independently propagated plant
  around the existing CSTR NLP.
- [x] Horizon shifting, warm starts, predicted-state background solve,
  measurement update, fallback reset, and visible time ordering.
- [x] Side-by-side full-resolve, no-correction, clamped-linear, fix-relax,
  active-set-path, and guarded-path policies.
- [x] Deterministic nominal, constraint-switching, and stress/model-mismatch
  campaigns.
- [x] Full corrected-point acceptance guard; rejected points trigger a fresh
  solve and reset.
- [x] Active-set event timeline and directional-degeneracy experiment.
- [x] Closed-loop quality, constraint, fallback, failure/recovery, and latency
  results.
- [x] Nominal smoke, active-set-switch, and warm-failure recovery regressions.
- [x] Commit-stamped 100-interval benchmark artifact and explicit
  local-validity/safety limitations.

---

- [x] Book page under `docs/src/` updated; no new page requires a
      `SUMMARY.md` entry
      as it stands now


jkitchin (commented):
Reviewed against the code, the executed notebook artifact, and the `pyomo_pounce` sensitivity API this builds on. The engineering is careful and the structure is good — but two defects undercut the exact evidence the PR exists to produce, and the artifact itself is what proves them.

## Blocking

**1. The guard rejects on the *pre-corrector* primal violation** — `python/pounce/examples/asnmpc_cstr.py:739`

```python
if float(report.violation) > config.feasibility_tol:
    reasons.append("primal_feasibility")
```

`report.violation` is measured at `x_new`, the predictor step, *before* `_correct()` runs (`pyomo-pounce/pyomo_pounce/sens.py:1986`, then `corrector` at `:1990`). The docstring is explicit: `corrector_iter` reports what the Newton iterations did "without changing anything the rest of the report measures." But the trajectory actually applied is the **corrected** one, and its primal feasibility is already checked separately as `corrector["feasibility"]`. So the guard holds the uncorrected point to the corrected point's tolerance — it gates on the residual the corrector exists to remove.

The artifact demonstrates the cost. `guarded_path` rejects **19 of 30** switching samples, and the fallback summary lists `primal_feasibility` as the *only* observed reason. Meanwhile the unguarded `path` policy on the same seeded campaign returns numbers identical to `full_resolve`:

| policy | IAE | ISE | stage cost | movement |
|---|---|---|---|---|
| `path` | 1.4091 | 0.072361 | 3.1455 | 1.7722 |
| `full_resolve` | 1.4091 | 0.072361 | 3.1455 | 1.7722 |

Those 19 rejections bought nothing and cost 338.91 ms median online (vs 72.28 ms for `path`) and 19 deadline misses. Fix: drop `report.violation` from the guard when a corrector ran, or gate it only for the `corrector_iter == 0` policies.

**2. Reported online latency double- and triple-counts the update** — `asnmpc_cstr.py:657-691`

`_corrected_update` opens its timer, then calls `estimate_report(...)`, `estimate(...)` with identical arguments, and — for path policies — `active_set_changes(...)`. All three redo the full predictor and corrector back-solve sequence against the same held factor, and all three are inside the window that becomes `update_latency_s`.

So every "online median/p95" cell and every `deadline_misses` count for `clamped_linear`/`fix_relax`/`path`/`guarded_path` is ~2–3× the work a real controller does. The table's own shape corroborates it: `fix_relax` 70.43 ms vs `path` 72.28 ms — the path record adds ~2 ms, so ~35 ms of that 70 is the duplicated call. `path`'s 1 deadline miss on switching and 2 on stress are artifacts. This is the headline evidence of the PR, and it understates the advanced-step advantage.

The guard genuinely needs the report online, so the minimal honest fix is to move `active_set_changes` out of the timed region (it re-derives a path `estimate(mode="path")` already took) and say in the notebook that the remainder still counts the predictor+corrector twice.

**3. The stamped commit in the PR body is wrong.** The body says `d3d96ab9c7e25a09a073bf6d40bc4c071db5bbb2`; cell 1 and the provenance table both say `01363b38183543b9c6ceb4a5d421ebeb7f6f33a1`. `d3d96ab` is four commits back and predates `3c0a044 fix(asnmpc): preserve accepted warm starts`, which changed the driver. The notebook stamp is the correct one — it's the body that's stale, and the checklist item "Every claim in this PR body … is true of the code as it stands now" is ticked.

**4. The frozen lock cannot reproduce the artifact.** `python/pyproject.toml`/`uv.lock` pin `pyomo-cvp==0.4.0`; the notebook ran on `pyomo-cvp 0.7.2`; `ci.yml` installs it unpinned. Three surfaces, three versions, for a PR whose value proposition is a stamped reproducible artifact.

## Worth fixing

**5. `corrector_no_progress` has no absolute floor** (`asnmpc_cstr.py:754`). `final > 0.5 * initial` rejects a point the predictor already landed near-exactly on, because nothing is left to halve — a *better* predictor makes this guard fire more. Add `and final > max(feasibility_tol, stationarity_tol, complementarity_tol)`.

**6. Two tests run the paper-scale model**, contradicting "CI uses the same model structure with a smaller deterministic discretization". `test_switching_campaign_records_active_set_events` builds `CstrConfig(deadline_s=2.0)` — defaults to 100 intervals, ncp=3 — and drives 12 closed-loop samples; `test_directional_degeneracy_experiment_reaches_the_control_kink` calls `directional_degeneracy_experiment()` with no config. At ~270 ms/solve that's ~5–10 s added to CI.

**7. Those same two tests are trajectory-coupled and have one event of margin.** In the 30-sample artifact the entire switching campaign produces exactly **one** active-set event, at sample 11 — and the test runs 12 samples of a differently shaped campaign (`make_campaigns(steps=12)` moves the feed-shift indices and redraws the noise) before asserting `active_set_changes > 0`. A failure reads `assert 0 > 0`. The degeneracy test hardcodes `startswith("v1[")`, `bound == "lower"`, `0 < fraction < 0.1`. Per CLAUDE.md's fixture-sweep discipline these are exactly the assertions an unrelated solver change turns red with no diagnosis; at minimum give them messages naming what stopped being true. Note also that `assert all(step.guard_reasons ...)` currently passes largely *because* of finding 1 — fixing the guard may flip it.

**8. `_solve_with_recovery` catches only `RuntimeError`** (`asnmpc_cstr.py:556`). A solve raising Pyomo's `ApplicationError` skips the documented cold-start retry and aborts the campaign, with the advertised failure/recovery counters still at zero.

## Minor

- `pounce/examples/asnmpc_cstr.py` ships inside the `pounce-solver` wheel (`python-source = "."`) but imports `pyomo`, `pyomo_cvp`, and `pyomo_pounce` at module top level — none a dependency or extra of `pounce-solver`. `pip install pounce-solver; import pounce.examples.asnmpc_cstr` gives a bare `ModuleNotFoundError`, and `docs/src/sensitivity.md` names the module without saying what to install. An `[asnmpc]` extra or one sentence closes it.
- `_measure_state` clips measured temperature to `config.temperature_limit` (`asnmpc_cstr.py:963`), so the controller structurally cannot receive a state above its own bound. Inert here (`max_corrected_violation == 0.0`), but it means the safety story never exercises the case it is most interesting for, and the reason for the clip — `build_controller` would otherwise raise — isn't stated.
- `stress_model_mismatch` sets `measurement_bias=(0.180, 0.016)` against `concentration_scale=0.04`: a constant offset 4.5× the trust radius. The 30/30 fallback rate is that arithmetic, not local-validity detection — worth saying so rather than presenting it as the guard noticing model mismatch.

## What's good

The predict/measure/correct/validate/apply/propagate/shift ordering is genuinely visible in `run_closed_loop`, the plant is honestly independent (separate RK4, separate parameters), `bound_relax_factor: 0.0` is the right call given `activity` is empty under relaxed bounds, and rejecting on ambiguous *and* weakly-active control rows is the conservative direction CLAUDE.md's gh#763 corollary asks for. The dataclass diagnostics and event ledger make the discrete behaviour inspectable, which is the actual contribution.

Findings 1 and 2 both point the same way: the guard looks stricter than it is and the update looks slower than it is, so the notebook's central comparison is measuring the harness rather than the method. Both are small code changes; the notebook needs re-executing after either.


---

bernalde:
Addressed the findings from review `5025558823` in append-only commits through `1faab9a7`.

### What changed

- The guard now treats corrected-point feasibility as authoritative when a corrector ran; predictor violation is consulted only when there was no corrector. `corrector_no_progress` also has an absolute feasibility/stationarity/complementarity floor.
- Active-set diagnostics now run outside the online timing window. The tutorial and notebook explicitly disclose that `estimate_report(...)` plus `estimate(...)` still replay the predictor/corrector, so the remaining latency is a conservative harness measurement.
- `pyomo-cvp==0.7.2` is pinned consistently in project metadata, CI installation, and the regenerated lock. The sensitivity guide now states the optional Pyomo/NMPC packages needed for the example.
- The paper-scale tests were replaced with a deterministic 20-interval, 2-collocation-point fixture that produces several active-set transitions. The trajectory-sensitive assertions now carry diagnostic messages.
- Recovery now catches both `RuntimeError` and Pyomo `ApplicationError`, with parameterized coverage for both paths.
- The docs now explain why measured temperature is clipped at the model bound and identify that as a limitation of this demonstration. They also state that the stress campaign starts with a bias 4.5x the local trust scale and intentionally forces one out-of-validity fallback.
- The notebook was re-executed against a native extension rebuilt from the exact branch source. It has 19 sequential code cells, 0 execution errors, and 14 plots. The corrected switching campaign has zero guarded fallbacks and zero corrected temperature violation; the stress campaign fallback rate is 1/30.

### Provenance correction

The commit recorded in the existing PR body (`d3d96ab...`) is stale. The regenerated notebook's source stamp is `9e11922a96eee7a4ebc9850a23a277a66f1525d3`; the exact output artifact is committed in `1faab9a70eaace57f5743b8f29e7f55e1f41a291`.

### Verification

- `python/.venv/bin/python -m pytest -q pyomo-pounce/tests` — **468 passed, 8 skipped**
- focused guard/timing/recovery tests — **17 passed**, with four mutation checks proving each new regression test fails for its corresponding reverted defect
- Ruff check and format check — passed
- `uv lock --check` — resolved 150 packages
- `scripts/check-docs-consistency.sh` — 47 reachable pages, no dead links
- notebook validation — source cells match `9e11922a`, 0 execution errors, all benchmark invariants passed

GitHub created an exact-head CI suite for `1faab9a7`, but it currently concludes `action_required` with zero jobs because the fork workflow is awaiting maintainer approval. There are no inline review threads to reply to or resolve.

bernalde:
@jkitchin, the review findings have been addressed through `1faab9a7`; the change-to-finding map and verification results are in the summary above. Could you please re-review when convenient? GitHub would not accept a formal reviewer request from this fork (`404` from the narrow review-request endpoint).

Issue #765, opened 2026-08-24
[adversary] python Problem: missing optional jacobianstructure raises AttributeError instead of defaulting to a dense Jacobian


jkitchin:
adversary-fingerprint: python-problem/missing-jacobianstructure-attributeerror

> Filed by the nightly adversary job. A constrained `problem_obj` that omits the
> **optional** cyipopt callback `jacobianstructure` crashes `Problem.solve()` with
> a bare `AttributeError` instead of falling back to a dense `(m, n)` Jacobian.

## Summary

POUNCE documents a "cyipopt-style interface". In cyipopt, `jacobianstructure` is
optional: when it is absent the Jacobian is taken to be **dense `(m, n)`**.
POUNCE's own Python helper implements exactly that fallback — `_preflight.py`
carries the comment *"Jacobian: honor `jacobianstructure()` when present, else
dense (m, n)"* and guards the call with `hasattr`.

The native solve path does not. Any `problem_obj` with `m > 0` that omits
`jacobianstructure` raises:

```
AttributeError: 'O' object has no attribute 'jacobianstructure'
```

Two things make this an oversight rather than a contract:

1. **It is asymmetric with the Hessian.** `hessianstructure` *is* feature-detected
   (`has_hessian` is computed with `hasattr("hessianstructure")`, and the
   Hessian block is gated on it). Only the Jacobian block calls unconditionally.
2. **POUNCE's own entry points disagree about the same object.**
   `pounce.preflight(obj, x0, ...)` accepts it and reports `ok: True`; then
   `Problem.solve()` and `solve_nlp_batch()` both reject it.

Source, `crates/pounce-py/src/problem.rs` (`build_tnlp_init`, ~line 1054) —
no `hasattr` guard, unlike the Hessian block a few lines below:

```rust
let (jac_rows, jac_cols, nele_jac) = if m > 0 {
    let s = call0(&self.problem_obj, "jacobianstructure")?;   // <-- unconditional
    ...
} else { (Vec::new(), Vec::new(), 0) };

// Hessian sparsity. When the user provides one, use it verbatim ...
let (hess_rows, hess_cols, nele_hess) = if self.has_hessian {   // <-- feature-detected
    let s = call0(&self.problem_obj, "hessianstructure")?;
```

Affects `Problem.solve()` and `pounce.solve_nlp_batch()`. `m = 0` is unaffected
(the branch is skipped). Deterministic, 3/3 runs.

## Reproduction

Self-contained, 12 lines:

[...]

Issue #3, opened 2026-05-13
pyomo Solver Factory for pounce


jkitchin:

jkitchin:
Implemented: the pyomo-pounce solver plugin landed in cdba9f8, and the `notebooks` uv dependency group + 05_pyomo demo notebook in 718b150. Pyomo can now solve with pounce via the standard SolverFactory interface.

PR #779, opened 2026-08-25
feat(py): add uncertainty-aware catalyst pellet design

### LINKED ISSUE #775: Flagship tutorial/research case: uncertainty-aware nonisothermal catalyst-pellet inverse design
> Migrated upstream from [bernalde/pounce#3](https://github.com/bernalde/pounce/issues/3). Existing implementation: [fork PR #8](https://github.com/bernalde/pounce/pull/8).

## Objective

Create a flagship chemical-engineering tutorial on **uncertainty-aware inverse design of a nonisothermal catalyst pellet**.

The target application is a radially heterogeneous pellet for CO₂ methanation (or a comparably well-documented exothermic reaction), with spatially varying catalyst activity and optionally effective transport properties. POUNCE should optimize the internal profile subject to reaction–diffusion–heat equations, catalyst-inventory/manufacturability constraints, and temperature/concentration limits, then quantify how parameter uncertainty changes the recommended design.

This is intended to be the most ambitious of the proposed application cases. A classical problem that once required a purpose-built optimal-control/BVP code should become one transparent differentiable nonlinear program with solver diagnostics, exact or verified gradients, uncertainty information, and a reproducible physical validation suite.

## Scientific positioning

Optimizing activity distributions inside catalyst pellets is **not new**. Analytical and numerical results have existed for decades, and recent work already optimizes structured methanation particles. The paper-worthy contribution would be the accessible end-to-end workflow enabled by POUNCE:

- formulate the distributed model and design in one system;
- impose bounds, inventory, path, and thermal constraints directly;
- compute sensitivities/covariance information from the solved system;
- redesign under uncertainty without hand-derived adjoints or a separate BVP optimizer; and
- inspect convergence and physical residuals with the same tooling as the NLP.

The tutorial must reproduce known qualitative/limiting behavior before presenting any new robust design result.

## Relevant POUNCE building blocks

There is not yet a catalyst-pellet example, but several existing capabilities are directly relevant:

- [`docs/src/bvp.md`](https://github.com/jkitchin/pounce/blob/main/docs/src/bvp.md) describes constrained BVP solves with state/parameter bounds, nodewise inequality constraints, and an optional objective.
- The JAX/PyTorch BVP paths provide differentiable fixed-mesh solves, while the adaptive NumPy path is not differentiable through mesh changes.
- [`python/notebooks/26_parameter_covariance.ipynb`](https://github.com/jkitchin/pounce/blob/main/python/notebooks/26_parameter_covariance.ipynb) demonstrates reduced-Hessian covariance from a solved parameter-estimation NLP.
- [`python/notebooks/31_information_identifiability.ipynb`](https://github.com/jkitchin/pounce/blob/main/python/notebooks/31_information_identifiability.ipynb) demonstrates information/identifiability calculations from the same factorized system.
- A direct simultaneous JAX collocation transcription can expose all states and radial design variables to POUNCE with exact derivatives and explicit path constraints.

A key early decision is whether to use a nested differentiable BVP solve or a simultaneous collocation NLP. The simultaneous formulation is the recommended primary route because it makes thermal/path constraints and design regularization explicit and avoids finite-differenced objective gradients in the current constrained NumPy BVP interface. The nested BVP path remains valuable as a smaller differentiability demonstration and an independent check.

## Literature landscape

- Morbidelli and collaborators, **“Optimal catalyst activity profiles in pellets—VII. Arbitrary kinetics with finite external heat and mass transfer”** (1987), develops classical optimal-profile results including concentrated/Dirac-like activity in limiting cases. It is a required reminder that unconstrained profile optimization can produce singular, unmanufacturable designs. [DOI: 10.1016/0009-2509(87)87013-6](https://doi.org/10.1016/0009-2509(87)87013-6)
- Dougherty and Verykios, **“Optimization of catalytic activity distributions in series and parallel reaction schemes”** (1986), uses numerical collocation for nonisothermal pellets and reports transitions among shell/core structures. This is directly relevant to mesh choice, nonconvexity, and qualitative validation. [DOI: 10.1002/aic.690321111](https://doi.org/10.1002/aic.690321111)
- **“Optimal catalyst activity profiles in pellets—X. The role of catalyst loading”** (1993) shows that bounded catalyst loading leads to step-like optimal profiles and that nonlinear kinetics require numerical optimization. This motivates explicit activity bounds, inventory constraints, and regularization rather than allowing a mesh-dependent spike. [DOI: 10.1016/0009-2509(93)80357-V](https://doi.org/10.1016/0009-2509(93)80357-V)
- Zimmermann, Bremer, and Sundmacher, **“Optimal catalyst particle design for flexible fixed-bed CO₂ methanation reactors”** (2020), jointly optimizes radial activity, permeability, and thermal conductivity and obtains an egg-yolk-like design with an inert diffusion barrier and reduced thermal sensitivity. This is the strongest application reference and a natural target for qualitative reproduction. [DOI: 10.1016/j.cej.2019.123704](https://doi.org/10.1016/j.cej.2019.123704)
- **“Tailored catalyst pellet specification for fixed-bed reactor optimization”** (2018) motivates a shortcut method because rigorous optimization of heterogeneous pellets embedded in reactors is computationally costly. This directly supports the “previously cumbersome” motivation, while also providing a comparator for how much model fidelity is justified in a tutorial. [DOI: 10.1016/j.cherd.2018.06.043](https://doi.org/10.1016/j.cherd.2018.06.043)
- Kaczmarski, **“An efficient and robust method for numerical analysis of a dead zone in a catalyst particle and packed-bed reactor”** (2021), documents the difficulty of free-boundary/dead-core calculations and limitations of standard algorithms. It motivates a dead-core continuation extension but also supports keeping it out of the minimum viable tutorial. [DOI: 10.1002/eng2.12370](https://doi.org/10.1002/eng2.12370)
- A recent forced-periodic reactor study with Gaussian activity distributions shows that structured-pellet design remains relevant for flexible, nonisothermal operation rather than being only a historical mathematical exercise. [DOI: 10.1016/j.ces.2024.120945](https://doi.org/10.1016/j.ces.2024.120945)

## Proposed physical model

Start with a spherical pellet. For each reacting species and energy, use a conservative radial formulation equivalent to

```text
(1/r²) d/dr (r² D_eff(r) dc_i/dr) + ν_i R(c,T; θ) = 0
(1/r²) d/dr (r² k_eff(r) dT/dr) + (-ΔH) R(c,T; θ) = 0
```

with symmetry/zero-flux conditions at `r = 0` and finite external mass/heat-transfer conditions at the surface. Avoid a numerical `1/r` singularity by using flux variables, a transformed coordinate, or analytically regularized center equations.

Candidate design fields are:

- local activity/catalyst loading `a(r)` (required);
- effective diffusivity or porosity/permeability `D_eff(r)` (optional second stage); and
- effective thermal conductivity `k_eff(r)` (optional second stage).

Parameterize profiles with a small piecewise-constant, piecewise-linear, or B-spline basis. Include pointwise bounds, a total catalyst-inventory constraint, and a total-variation/smoothness or minimum-feature constraint so the optimum is interpretable and not set by mesh resolution.

The objective can maximize methane production/effectiveness or minimize pellet volume/catalyst amount for a required conversion, subject to a maximum internal temperature and nonnegative/conservative species profiles. Selectivity belongs in the objective only if the chosen kinetic network contains meaningful side reactions.

### Stage 0 — fix a reproducible data/model scope

- Select one published kinetic and transport parameter set with complete units and operating conditions; the Zimmermann et al. methanation study is the preferred anchor if enough data are recoverable.
- Document every assumption required to reduce the source model to a tutorial-size pellet.
- Establish a uniform-pellet reference solution and compare it with a published curve/result or an independent implementation.

Do not tune undocumented parameters merely to reproduce an attractive optimized profile.

### Stage 1 — solve and validate limiting BVPs

- Implement an isothermal, first-order single-reaction sphere.
- Verify the computed effectiveness factor against the analytical Thiele-modulus formula over diffusion- and reaction-limited regimes.
- Demonstrate radial mesh convergence, surface film limits, center regularity, and integrated reaction-rate/molar-flux balance.
- Add the nonisothermal/multicomponent model only after these checks pass.

### Stage 2 — choose the POUNCE transcription by evidence

Prototype both:

1. a nested differentiable fixed-mesh JAX BVP with design variables outside the solve; and
2. a simultaneous JAX collocation NLP with states, fluxes, and profile coefficients as variables.

Compare residual accuracy, gradient accuracy, constraint handling, solve robustness, and computational cost on the uniform-pellet and low-dimensional design cases. Choose one as the tutorial route and retain the other as an independent check if practical.

### Stage 3 — nominal inverse design

- Optimize `a(r)` first under activity bounds and fixed total catalyst inventory.
- Increase the number of profile degrees of freedom gradually through continuation/warm starts.
- Add `D_eff(r)` and/or `k_eff(r)` only if the activity-only model is validated and the extra field has a defensible manufacturing interpretation.
- Compare uniform, conventional egg-shell, and optimized profiles at identical catalyst inventory.
- Re-solve every reported design on a refined mesh and with the profile represented independently enough to detect discretization artifacts.

### Stage 4 — uncertainty and robust redesign

Choose a bounded, interpretable uncertainty set. A preferred workflow is:

- estimate a small set of kinetic/transport parameters from a synthetic or published data set;
- use POUNCE's reduced-Hessian information to diagnose identifiability and form a covariance model;
- propagate local uncertainty to conversion, hot-spot temperature, and active constraints; and
- compare the nominal optimum with either a small scenario-based robust design or a chance/safety-margin approximation.

At minimum, validate local covariance/delta-method predictions against sampled parameter re-solves near the nominal point. If experimental data are not sufficiently documented, label the study “uncertainty-aware” with synthetic calibration rather than claiming experimental robustness.

### Stage 5 — continuation and difficult regimes (stretch)

- Continue designs across Thiele modulus, external heat transfer, feed temperature, or uncertainty radius to reveal shell/core topology changes.
- Investigate a dead-core/free-boundary formulation only after the smooth positive-concentration model works reliably.
- If an optimized profile changes discontinuously with parameters, report the branch dependence instead of presenting one local optimum as globally unique.

## Numerical and physical validation

Every reported solution should include:

- collocation/BVP residuals and boundary-condition residuals;
- integrated species and energy balances;
- mesh/refinement error in objective, peak temperature, and key profile features;
- positivity and path-constraint checks between nodes when needed;
- derivative checks for profile coefficients and uncertain parameters;
- multiple initializations or continuation histories to expose local optima; and
- a refined forward solve of the final design independent of the optimization mesh.

Benchmark tables must be stamped with the POUNCE commit, model-data revision, tolerances, mesh/basis, and execution environment.

- [ ] Auditable pellet equations, parameter data, units, and source mapping.
- [ ] Analytical effectiveness-factor regression for the isothermal first-order sphere.
- [ ] Validated nonisothermal multicomponent uniform-pellet reference with mesh, balance, and center-regularity checks.
- [ ] Evidence-based selection of nested BVP or simultaneous collocation as the main POUNCE formulation.
- [ ] Verified design gradients against full perturb-and-resolve calculations.
- [ ] Nominal activity-profile optimization with bounds, fixed catalyst inventory, and manufacturability regularization.
- [ ] Uniform/egg-shell/optimized comparison on a refined independent solve.
- [ ] Qualitative comparison with a published shell/core or egg-yolk result, with model differences stated.
- [ ] An uncertainty model tied to parameter information/covariance, plus nominal-versus-uncertainty-aware designs.
- [ ] Sampled re-solve validation of uncertainty predictions and thermal/conversion constraints.
- [ ] A short CI case and a commit-stamped paper/tutorial artifact.
- [ ] Explicit local-optimum and model-form limitations; no unsupported claim of globally optimal pellet architecture.

## Scope guard and decision gates

The minimum publishable tutorial is **activity-profile design in one validated spherical pellet plus a small uncertainty study**. Pellet-in-reactor co-design, arbitrary pore-network morphology, transient cycling, three design fields, and a free dead-core boundary are stretch goals. If a complete, reproducible CO₂-methanation parameter set is unavailable, use a simpler exothermic reaction with analytical limits rather than inventing missing data.

bernalde:
> Upstream submission of [bernalde/pounce#8](https://github.com/bernalde/pounce/pull/8); tracks #775.

## Problem

POUNCE has differentiable BVP, covariance, and JAX NLP building blocks, but no end-to-end chemical-engineering case that joins them. Issue #775 asks for a reproducible nonisothermal spherical catalyst-pellet design with analytical and physical validation, explicit bounds/inventory/manufacturability constraints, verified gradients, reduced-Hessian uncertainty, a robust redesign, and a commit-stamped tutorial.

Before this branch, importing `pounce.examples.catalyst_pellet` on parent `20be28908c91157c0edea707ba1c7f2c4d242034` fails with `ModuleNotFoundError`; there is no model, notebook, book page, or focused CI case.

## Cause

The capabilities existed separately but there was no conservative pellet model or application layer connecting fixed-mesh JAX derivatives, simultaneous POUNCE constraints, curve-fit covariance, and physical validation. A full reproduction of the Zimmermann reactor/pellet co-design would also exceed the issue's scope guard and require transport/reactor correlations that are not completely specified for this tutorial.

## Fix

- Add `pounce.examples.catalyst_pellet`, a four-species/energy equal-volume finite-volume model with the published Koschany methanation kinetics, finite external films, exact JAX residual derivatives, balance diagnostics, and mesh refinement.
- Add the analytical first-order spherical effectiveness regression and a focused six-test CI case covering rate data, center regularity, integrated balances, films, refinement, equal-volume activity-basis validation, signed full perturb-and-resolve gradients for both design and uncertain parameters, two starts, nested/simultaneous agreement, covariance, worst-case scenarios, and sampled uncertainty re-solves.
- Make the simultaneous JAX/POUNCE NLP the primary design route. Keep a small independent nested SLSQP/implicit-derivative route as evidence: on the recorded 8-cell/4-zone case their activity profiles differ by `4.37e-7` at most and production agrees to better than `2e-6` relative.
- Add a covariance-derived principal-axis uncertainty set and a genuine worst-case epigraph. The recorded robust solve enforces a `7.0416656096e-7 mol/s` finite-scenario floor, closes constraints to `1.82e-13`, and keeps every scenario below `579.63 K` versus the `613 K` validity ceiling.
- Add an executed notebook stamped with reviewed source commit `7043b33e15ab8761c610ccedc32ae03d1f1db7d0`, a book page with equations/units/source mapping, notebook index entry, and changelog entry.
- Enforce `nodes % zones == 0` at every solve/design/refinement path so each activity coefficient owns exactly the same volume and inventory cannot drift between meshes.

Two larger/less defensible routes were rejected. The issue's full 30,750-state reactor-coupled reference was not copied because the minimum scope is one validated pellet and undocumented correlations would have to be invented. An initial `80 W m^-2 K^-1` heat-film assumption was also rejected before optimization: at the fixed activity inventory it produced no admissible low-temperature balance solution (`1.30e-2` maximum scaled residual and `0.989` relative energy-balance error). The documented `250 W m^-2 K^-1` tutorial assumption keeps the uniform reference on a steady branch; it is not fitted to obtain an attractive activity profile.

## Blast radius

This is additive Python/example/documentation work. It changes no Rust solver code, public solver routing, dependency manifests, or lockfiles. Existing users are unaffected unless they import the new module or run the tutorial.

The branch is based on fork `main` at `20be2890`. Open sibling PRs #6 and #7 also edit central catalog files and `python/pounce/examples/__init__.py`, so those small index/init hunks may need ordinary conflict resolution depending on merge order; the new model, notebook, test, and book page have distinct paths.

## Tests

- Parent red check: copied the current new test into an archive of `20be2890`; collection fails for the intended reason, `ModuleNotFoundError: No module named 'pounce.examples'`.
- `PYTHONPATH=python python/.venv/bin/python -m pytest -q python/tests/test_catalyst_pellet_example.py python/tests/test_curve_fit.py` — **66 passed in 31.65s**.
- `python -m ruff check python/pounce/examples python/tests/test_catalyst_pellet_example.py` — clean.
- `bash scripts/check-docs-consistency.sh` — **48 pages**, all reachable; no dead TOC links.
- `jupyter nbconvert --to notebook --execute --inplace python/notebooks/37_catalyst_pellet_inverse_design.ipynb --ExecutePreprocessor.timeout=300` — clean; saved outputs record source commit `7043b33e`, the environment, and signed scenario-gradient relative errors below `9e-10`.
- `make book` could not run locally because the `mdbook` executable is not installed. The repository's docs workflow installs its pinned mdBook version and remains the authoritative rendered-book check.
- Rust formatting, clippy, and tests were not run because this PR changes no Rust source, Cargo manifest, or lockfile.

---

- [ ] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean — not applicable; no Rust/Cargo changes

Closes #775


jkitchin (changes requested):
Thanks for this — the model is careful work, and the docs page is unusually honest about scope: stating the egg-yolk/egg-shell divergence from Zimmermann as a model-form difference rather than burying it is exactly right. The suite passes for me (6 passed in 76s) against `pounce-solver` 0.10.0, and I reproduced the rejected `80 W m^-2 K^-1` assumption exactly (`1.30e-2` max scaled residual, `0.989` energy-balance error), so that reasoning checks out independently.

One thing I'd like fixed before merge.

## `PelletConfig()` defaults to a mesh that doesn't converge

`nodes: int = 12` (`python/pounce/examples/catalyst_pellet.py:115`) is the default on every public entry point, and nothing in the PR exercises it — the tests use `nodes=6`/`nodes=8`, the notebook sets `PelletConfig(nodes=8, zones=4)`, and the only place 12 appears is `refine_solution(..., nodes=12)`, which warm-starts from an interpolated 8-cell state. Cold-starting at the default:

```python
>>> import numpy as np
>>> from pounce.examples.catalyst_pellet import (
...     PelletConfig, solve_forward, solve_design, solve_nested_design)
>>> solve_forward(np.full(4, 0.16), PelletConfig()).max_scaled_residual
0.38536828996307626          # success=False; species balance 0.88, energy balance 1.24
>>> solve_design()
RuntimeError: design seed failed for nominal: The maximum number of function
evaluations is exceeded.; max scaled residual=3.854e-01
>>> solve_nested_design()
RuntimeError: The maximum number of function evaluations is exceeded.;
max scaled residual=3.854e-01
```

The cause is the hard-coded `max_nfev=250` at `catalyst_pellet.py:701`. Cold-start evaluation counts against mesh size:

| nodes | 4 | 6 | 8 | 10 | 12 | 16 | 24 |
|---|---|---|---|---|---|---|---|
| nfev | 18 | 30 | 34 | 83 | **349** | 1287 | 2776 |

The budget clips off one step above the largest tested mesh, and it lands exactly on the default. Raising it converges `nodes=12` in 349 evaluations to a `4.6e-13` residual, so the physics is fine — it is purely the budget. Note that `max_nfev` is not exposed on `solve_forward`, so a caller cannot work around it either.

I get identical evaluation counts under your recorded environment (jax 0.9.2, scipy 1.17.1) and under jax 0.10.2, so this is not version drift on my side.

Simplest fix is `nodes: int = 8`, matching what the tests and notebook actually validate. If you'd rather keep 12 as the default, `solve_forward` needs a larger or caller-settable budget, or an internal mesh continuation. Either way, could a test call `solve_design()` with no arguments? The reason this survived nine commits and a green suite is that no test path ever reaches the default branch.

## Also cheap: `CHANGELOG.md` cites the wrong issue

The entry ends `(#3)`, which resolves to jkitchin/pounce#3 — "pyomo Solver Factory for pounce", closed in May, unrelated. It should be `(#775)`; the neighbouring `(#765)` entry shows the convention. The `Closes #775` in the PR body is correct, so this is just the fork's issue number leaking through.

## Non-blocking, for a follow-up issue

`_state_bounds` puts the temperature upper bound at exactly `temperature_limit_k` and `solve_forward` passes those bounds to `least_squares`, so `thermal_margin` (line 1160) returns `613 - max_T >= 0` identically — the SLSQP inequality can never be violated. A thermally infeasible design therefore isn't pushed back by the constraint; it makes the inner root-find fail and `evaluate()` raise:

```
replace(config, heat_transfer_coefficient_w_m2_k=80.0)
solve_nested_design -> RuntimeError: `ftol` termination condition is satisfied.;
                       max scaled residual=1.302e-02
```

This changes no reported number — every recorded design sits near 579 K — and the simultaneous route enforces the ceiling correctly as a variable bound. But it does mean the two routes only agree where the ceiling is slack, which is worth either a line in the docs' validation ladder or a widened forward-solve temperature bound so the 613 K limit is enforced by the optimizer in both routes.

Minor, same category: the PR body's stamped commit (`7043b33e`), production floor (`7.0416656096e-7`) and scenario-gradient bound (`9e-10`) are stale relative to the committed notebook, which records `2c14bbe`, `7.037435140168223e-07`, and a maximum of `1.23e-09`.

Nothing I found in the numerics is wrong: the Koschany parameters and equilibrium correlation match the published table, the equal-volume scheme genuinely avoids the `1/r` centre singularity, the residual scaling drops out of `ds/dp` at the root so the implicit derivatives are unaffected by it, the double-`where` in `_analytical_effectiveness_backend` is the correct NaN-safe form, and the robust epigraph is a real one.

---

bernalde:
## Review feedback addressed

- `fa9089f39011ea91b9dad730185aa5c4e79de1b2` merges current upstream `main` append-only; the PR now merges cleanly against `dc103aa2916a7ce0181d1f9c85d745510f0770a1`.
- `180685d21fb96242f0a34d6159ecfd30dd5ab3fb` changes the public `PelletConfig` default from 12 to the validated eight-cell mesh. `solve_design()` now succeeds with no arguments, and a new regression exercises that exact public path.
- The same commit changes the changelog reference from the fork's `#3` to upstream issue `#775`.

The default-path regression was proved against the pre-fix source: it failed with `RuntimeError: design seed failed ... max scaled residual=3.854e-01`, then passed after changing the default. Current focused verification is:

- `python/.venv/bin/python -m pytest -q python/tests/test_catalyst_pellet_example.py python/tests/test_curve_fit.py` — **67 passed in 34.45s**;
- `python -m ruff check python/pounce/examples/catalyst_pellet.py python/tests/test_catalyst_pellet_example.py` — clean.

The nested-route thermal observation was reproduced independently and deferred as requested to [#787](https://github.com/jkitchin/pounce/issues/787), with a coordination note on [#775](https://github.com/jkitchin/pounce/issues/775#issuecomment-5429653376). No finding was declined.

This comment also supersedes the head-specific fields in the PR body. The committed notebook records source commit `2c14bbe94507936904c4a86eecc86692cf3402f7`, a finite-scenario production floor of `7.037435140168223e-07 mol/s`, and a maximum signed scenario-gradient relative error of `1.23045531e-09`.

Current-head CI has not run: [workflow run 33001583665](https://github.com/jkitchin/pounce/actions/runs/33001583665) concluded `action_required` with zero jobs. The code-review decision therefore remains `CHANGES_REQUESTED` pending maintainer re-review, and the Actions run still needs maintainer approval before CI evidence exists for this head.

bernalde:
@jkitchin, the requested default-mesh fix, no-argument regression, changelog correction, and thermal follow-up are now at `180685d21fb96242f0a34d6159ecfd30dd5ab3fb`. Please re-review when convenient. The current Actions run still requires maintainer approval before its jobs can start.

PR #784, opened 2026-08-26
feat(start): diagnose degenerate starts, and condition them on request

jkitchin:
Running KRONOS (Ahmed & Hasan, 2026, *Comput. Chem. Eng.* **215**, 109839)
head-to-head against POUNCE on its own 244-problem benchmark set turned up
fifteen models POUNCE failed from their bundled starting point. Ten of them
are models KRONOS proves feasible to 2.4e-7 or better, so the
`Infeasible_Problem_Detected` verdict was simply wrong.

What recovers them is the finding:

| remedy | recovered |
|---|---|
| default | 0 |
| `start_with_resto=yes` | 0 |
| `expect_infeasible_problem=yes` | 0 |
| `mu_strategy=adaptive` | 4 |
| Adam warm-up (KRONOS stage 0) | 3 |
| **one displaced start, relative 1e-2** | **13** |

The two zeros at the top are the point. Restoration is the textbook answer
to an infeasibility verdict and it recovers *nothing*, because it inherits
the same degenerate point — it is a different objective solved from the
same place. The iterate does not need to be **better**, it needs to be
**non-degenerate**. At a start where the constraint Jacobian is
rank-deficient, LICQ fails and a filter line search has no descent
direction to find, whatever else you hand it. Two ways in dominate the
corpus: a squared slack at zero (KRONOS's own reformulation loses rank
exactly on the active set), and an origin start on a homogeneous quadratic.

## What this changes

- **A local-infeasibility verdict reached from a rank-deficient point says
  so.** The CLI audits the starting point on a failed run and names the
  identically-zero Jacobian rows and columns, and says that LICQ failing
  there makes the verdict as much a statement about the start as about the
  problem. It reports zero rows, not a rank estimate — an SVD is not
  affordable to run speculatively — so absence of the finding is not a
  clean bill of health, and the docs say that.
- **Rung 3 of the second-opinion ladder** (`infeasibility_perturbed_start_retry`,
  on by default): one re-solve from a start displaced by a relative `1e-2`,
  promoted only on `Solve_Succeeded` / `Solved_To_Acceptable_Level`. Like
  the other rungs it restores every earlier knob to baseline first, so
  exactly one thing differs from the original solve — non-cumulativeness is
  load-bearing (gh#524, `cresc4`). It is the only rung that also fires on
  `Invalid_Number_Detected`, where re-running the same evaluation under
  different linear algebra would reproduce the NaN exactly.
- **`Invalid_Number_Detected` names the number.** All four such cases in
  the corpus were correct stops reported unhelpfully. The audit covers `x`,
  `f`, `grad f`, `g` and the Jacobian, names the index and column, and
  reports an infinity's sign. One corpus model ships a start that is
  literally `[nan, nan, nan, nan, 0, ...]`.
- **`start_point_conditioner=adam`**, off by default, runs KRONOS's stage-0
  Adam warm-up on `f(x) + rho*||violation(x)||^2`, generalised from that
  paper's equality-only form to two-sided bounds. Guarded: if it does not
  reduce the merit it hands the original point back.

## Why Adam is an option and never a default

Over 40 problems POUNCE already solves, it broke none and cut 22
(`rk23` 82 -> 11, `bt5` 45 -> 9), median **0.83x** iterations — and the
**total rose 1.62x** (3030 -> 4900), on `palmer1c` 71 -> 1023 and
`biggs6` 1906 -> 2938. Excluding those two it is 0.89x. A fixed, unscaled
`rho` against a badly-scaled model. A conditioner that is a median win and
an aggregate loss is an option, not a policy.

## Measurements

Trajectory check per `CLAUDE.md`. `scripts/sweep-fixtures.sh` against a
baseline built at `f6231f40`, both legs, whole corpus: **142 fixture-legs
each, empty diff.** That is evidence rather than a skipped check — 22 of
those legs end `Infeasible_Problem_Detected`, so rung 3 ran on all 22 and
promoted none. The fixtures that assert infeasibility are genuinely
infeasible and the new rung does not manufacture false promotions.

End-to-end re-run of the corpus against the built branch, 244 problems,
K=1, identical starting points:

| leg | solved | global optimum | solve time |
|---|---|---|---|
| KRONOS | 225/244 | 175/244 | 42.9 s |
| POUNCE, as KRONOS states them (squared slacks) | 191/244 | 155/244 | 14.6 s |
| POUNCE, natural inequality form | 223/244 | 189/244 | 7.4 s |
| **POUNCE, natural form + the ladder** | **239/244** | **199/244** | **9.4 s** |

The first three rows reproduce the pre-branch run problem for problem, and
that is the reason to keep them in the table: the diagnosis changes nothing
an embedder can observe, so the library legs must not move, and they do
not.

Sixteen problems move — twelve on rung 3, three on `mu_strategy=adaptive`,
one on `feral_scaling=mc64` — and every promoted answer is genuinely
feasible (worst residual 1.0e-8, `hs020`). Four of the sixteen were
`Invalid_Number_Detected`, which no earlier rung would have been offered.
Cost across the whole corpus: **34 extra solves, +2.00 s**, spent only on
runs that had already failed. The fifteen losses are down to three
(`a10_perm`, `a29_rump`, `hong`).

Two caveats it would be misleading to omit:

1. **The ladder is CLI-only.** `grep -rn "second_opinion\|scaling_retry"
   crates/pounce-py/src/ python/` returns nothing — the conditioner options
   are reachable from Python, the retry logic that drives them is not. The
   benchmark harness reproduces the ladder option-for-option rather than
   measuring it through the library, and a Python embedder does not get
   rung 3 today. That is a real gap between the frontends, not an artifact
   of the measurement.
2. `lch` needed its 300 s harness timeout lifted for a reason that is not
   POUNCE's: n=600, and JAX spends 3m30s compiling the Lagrangian Hessian.
   Given the room it solves in 1.14 s and 27 iterations.

The speed ratio on the both-solved set falls from 12x to 6.7x, which is
the expected direction: the fourteen problems added to the intersection are
the hard ones, and POUNCE now pays two or three failed solves on each.

## Notes for review

- The audit runs on `inner_tnlp` — the user's own TNLP, before presolve,
  elimination, scaling and the counting wrapper — for the same reason
  `run_derivative_test` does. A wrapper renumbers variables, so naming
  `x[3]` of a presolved model points at a *neighbouring* variable's answer,
  which is the gh#450 failure mode. Going around the counting wrapper also
  keeps the reported eval counts the solver's rather than the diagnosis's.
- The new options are read with their literal tags rather than through a
  `|name, fallback|` helper, because `init_options_wiring.rs` proves
  consumption by scanning the source for `get_*_value("<tag>"` and a tag
  passed as a variable is invisible to it. A registered knob nothing reads
  validates, accepts a value, and lies.
- The displacement is SplitMix64-seeded and reads no clock, address or
  thread identity, so a promoted retry is reproducible and a failed one is
  reportable.
- `issue_508_infeasibility_gap_status.rs`'s sentinel moved from
  "keeping the original local-infeasibility verdict" to
  "keeping the original `Infeasible_Problem_Detected` verdict". The banner
  now names the status because rung 3 also fires on
  `Invalid_Number_Detected`, where the old prose would have been a lie.

Inspiration is credited in `docs/src/acknowledgments.md`; the full write-up
with the corpus, every measurement and the rationale for each default is
`dev-notes/degenerate-starts.md`.


jkitchin (commented):
I built the branch, ran the touched crates' tests, clippy at CI's lint levels, and independently reproduced the mandated trajectory check.

## What I verified

- `cargo test -p pounce-nlp -p pounce-algorithm` green; `cargo test -p pounce-cli --test issue_508_infeasibility_gap_status` green (7/7, including the new `.sol` pin); `cargo fmt --check` clean; `cargo clippy … -D clippy::correctness -D clippy::suspicious` clean.
- **The fixture sweep claim holds.** Release-built `f6231f40` vs. the head, `scripts/sweep-fixtures.sh` both legs: 142 lines each, **empty diff**. 22 legs end `InfeasibleProblemDetected` — matching the PR's count, so rung 3 really did run on all of them and promoted none.
- **The feature works end to end.** I hand-wrote an HS008 `.nl` (`x²+y²=25`, `xy=9`, start at the origin — the case the code comments name). Default ladder: rungs 1 and 2 fail, rung 3 promotes `SolveSucceeded`, and the `.sol` carries `(4.6016, 1.9558)`, feasible to machine precision. With the ladder off, the new degeneracy banner fires with `2 of 2 constraint rows … (rows 0, 1)`.

The measurement discipline here is the strongest part — the "restoration recovers 0" row is what makes the diagnosis convincing rather than a plausible story, and the `f6b005a` follow-up (a non-promoted rung leaving its rejected iterate in the `.sol`) is a real pre-existing bug found and pinned.

---

## 1. The Adam warm-up hands back a `NaN` start it claims to have sanitised — `crates/pounce-nlp/src/start_conditioner.rs:292` / `:416`

`let start = x.to_vec()` is taken **before** the sanitise loop, and the merit guard restores `start` verbatim. So on a start carrying a `NaN` where Adam fails to reduce the merit, the un-sanitised point goes to the solver while the report says the variable was repaired.

Confirmed with a probe test (`target = [0, 2, 3]`, `start = [NaN, 2, 3]`, `adam(50)`):

```
x = [NaN, 2.0, 3.0]
report = ConditionerReport { sanitised: [0], max_shift: 0.0, iters: 0, merit_initial: 0.0, ... }
```

This contradicts the module comment three lines above it ("so a NaN start would make the whole warm-up a no-op that reports success"), the option's long description, and `docs/src/initialization.md`. The fix is one line — take the snapshot after sanitising:

```rust
for i in 0..n { if !x[i].is_finite() { … x[i] = self.sanitised_value(lo, hi); } }
let start = x.to_vec();
```

Nothing in the 17 conditioner tests covers sanitise + revert together — `the_warm_up_sanitises_a_non_finite_start_too` picks a start where Adam does improve, and `a_warm_up_that_does_not_help_returns_the_original_point` uses a finite start. That's the branch-coverage trap `CLAUDE.md` describes for the invariance legs: a rule that branches, with a fixture that only ever takes one branch. Worth a test that lands in the other one.

(Also: `max_shift` is folded over `(a - b).abs()` against that same `start`, so a `NaN` there silently reads as `0.0` — which is exactly what the probe printed.)

## 2. The new diagnosis second-guesses a presolve *proof* — `crates/pounce-cli/src/main.rs:1857`

The ladder is deliberately exempt when `presolve_certified.is_some()` (`main.rs:1709`) — "re-solving to second-guess a proof is a waste." The diagnosis block has no such guard, so on a certified-infeasible model with any structurally-zero Jacobian row it prints:

```
pounce: FBBT detected infeasibility (witness constraint 1)
pounce: the constraint Jacobian is rank-deficient there: 1 of 2 constraint rows …
pounce: LICQ fails at a point like that, so a local-infeasibility verdict reached from it is
        as much a statement about the starting point as about the problem. Try a different
        starting point, or `start_point_perturbation 1e-2`.
Status: Infeasible_Problem_Detected
```

(Reproduced on a two-row fixture: one linear row presolve refutes by bound propagation, one `y²=4` row started at `y=0`.) The `.sol` on that same run writes `solve_result_num` **201** — "proved", per `presolve_verdict` — and then the console tells the user to doubt it. The non-finite half of the audit is still useful there; it's the rank-deficiency advice that should carry `&& presolve_certified.is_none()`.

## 3. The warm-start rationale argues against what the code does — `crates/pounce-nlp/src/start_conditioner.rs:465`

```rust
// Only the primal start is conditioned. Displacing `x` while keeping a
// warm-started `z` / `λ` would pair a moved primal point with
// multipliers certified at the old one.
```

That second sentence names a hazard as if it were excluded, but the duals *are* passed through untouched — so under `warm_start_init_point=yes` the code produces exactly the pairing it describes. `docs/src/initialization.md` reproduces the same circularity ("passed through untouched, **because** displacing the primal point while keeping multipliers … pairs a moved iterate with duals that no longer belong to it"). Either the comment is inverted, or the conditioner should decline to displace when `init_z || init_lambda` — worth deciding which.

Related, same combination: `fetch_warm_start_snapshot` (`crates/pounce-nlp/src/orig_ipopt_nlp.rs:1658`) is a second `get_starting_point` call, so warm start + `conditioner=adam` runs the 200-iteration warm-up twice. Harmless for Jitter (deterministic), pure cost for Adam.

## 4. `with_bound_inf` clamp defeats its own comment — `crates/pounce-algorithm/src/application.rs:869`

The comment says the sentinels must come from the options so "a caller who moved `nlp_lower_bound_inf` would otherwise have a bound the algorithm treats as absent clipped against as if it were real." But `lower.min(-DEFAULT_BOUND_INF)` only honours a *loosened* sentinel. Set `nlp_lower_bound_inf=-1e10` and the conditioner still uses `-1e19`, so a `-1e15` bound is absent to the algorithm and present to the clipper — the stated failure mode, intact. Low impact; either drop the clamp or say why it's there.

## 5. Smaller things

- `main.rs:1771` and `:1780` are two adjacent `if !scaling_retry_promoted(retry_status)` blocks with nothing between them — merge.
- `docs/src/troubleshooting.md`, the degenerate-Jacobian example shows `it survived 1 independent re-solve(s) (mu_strategy=adaptive)` under an `Infeasible_Problem_Detected` verdict. With defaults that trigger gets three rungs and lists all three; the snippet as written can't occur.
- No CLI-level test covers rung 3 recovering a model or the two new banners. The wiring test proves the tags are *read*; nothing proves the effect. The HS008 fixture used above is ~35 lines of `.nl` and pins both.

---

None of the above is a blocker for the ladder itself, which measures well and is default-transparent. Items 1 and 2 are the two I'd want fixed before merge.


---

jkitchin:
All five findings addressed in abbe856c, plus one the review led me to that was not on the list.

**The one that was not on the list.** Chasing finding 3 I went looking for what else the ladder restores by value rather than by set-ness, and found that the rung table itself did it. Each rung wrote the earlier rungs' knobs back to their *resolved* values so that exactly one thing would differ from the original solve (the gh #524 contract). That is a no-op by value and not by set-ness, and set-ness is load-bearing here: `is_mu_strategy_fallback_enabled` is default-on precisely while `mu_strategy` is unset, so rung 3 writing back a resolved `mu_strategy monotone` switched off `run_with_mu_strategy_fallback` — the max-iteration stall retry — on the solve most likely to need it.

Mutation-checked on KRONOS's `a18_ackley1`: `Solve_Succeeded` in 237 iterations with the fix, `Maximum_Iterations_Exceeded` at 3000 without. This is a **pre-existing defect**, inherited verbatim from the CLI's inlined ladder rather than introduced by the library move — it just could not be observed while every rung ran in a process that was about to exit. Fix: rungs no longer emit write-backs at all, and the driver re-applies the baseline `OptionSnapshot` before each one.

**The five.**

1. `start_conditioner` captured its revert point *before* the sanitise loop, so a warm-up that did not help handed the raw `NaN` back, and `max_shift` folded `(NaN - NaN).abs()` to `0.0` — a clean-looking report over a poisoned start. The capture moved after the loop. Pinned by `a_sanitised_start_survives_a_warm_up_that_does_not_help`, mutation-checked (only that leg goes red).
2. The rank-deficiency advice is now gated on the verdict it explains. A presolve-certified infeasibility is a statement about the model's structure that no starting point bears on, and a paragraph about LICQ there is misdirection. The non-finite audit above it stays ungated deliberately — a `NaN` in `x` is worth naming whatever the status.
3. The warm-start rationale comment asserted the opposite of what the code does. Rewritten: duals are forwarded untouched, which is a caveat on the displaced start rather than a reason to decline to displace under `init_z || init_lambda`.
4. `ConditionedStartTnlp` now reads the application's own `nlp_lower_bound_inf` / `nlp_upper_bound_inf` instead of clamping them outward to `DEFAULT_BOUND_INF`. A model that deliberately narrows what counts as infinite is no longer silently widened back.
5. New `crates/pounce-cli/tests/degenerate_start_ladder.rs`, five tests over two new HS008 fixtures started at the origin, where **both** Jacobian rows vanish. One fixture is recoverable — rung 3 promotes, and the assertion is on the shipped `.sol` satisfying both constraints, not merely on the status. The other flips `x²+y²=25` to `=-25`, so it is degenerate *and* genuinely infeasible: all three rungs spend and the verdict stands, which is the property that distinguishes a ladder from laundering failures. Also pinned: both banners naming rows and columns, `print_level=0` silencing the diagnosis (it goes through `eprintln!`, so nothing else enforces the level), and `infeasibility_perturbed_start_retry=no` restoring the upstream verdict — that last one is the mutation guard for the first.

**Verification.** `scripts/sweep-fixtures.sh` against a `f6231f40` baseline is an empty diff, both legs, 142 fixture-legs each. `cargo test --workspace --no-fail-fast` is green, 290 test groups.

And the review prompted me to actually measure a claim this PR's changelog was making without evidence. It asserted that a plain `Problem.solve` "reproduces the CLI's 239 solved / 199 global" on the KRONOS corpus — written, not measured. Re-run properly, with the rung fix in: **zero per-problem differences over 243 problems**, same status and same global-optimum verdict on every one, the same 34 rungs spent, and the same rung promoting each of the sixteen recoveries. The 244th (`lch`) has no library-leg result to compare — the harness's 300 s per-problem cap expires inside its 221 s *Python model construction*, before any solve — which is why the full-corpus figures stay 239 / 199 rather than 238 / 198. The changelog now says that instead.

jkitchin:
Two loose ends from your comment that my previous reply did not cover — both now fixed in 2c9a4472.

**The double warm-up (your "related, same combination" note under item 3) was real, and worse than a nit.** `fetch_warm_start_snapshot` is a second `get_starting_point` on the same TNLP, and `ConditionedStartTnlp` had no memoization, so `warm_start_init_point=yes` + `start_point_conditioner=adam` ran the full 200-iteration warm-up twice and used one. Deterministic, so never a wrong answer — but the wasted half is the entire cost of the feature.

Cached now, keyed on the **raw incoming point** rather than on "have I run yet". The weaker key would have been silently wrong: a `Problem` can be solved again from a start the caller has since changed, and that must miss. Comparison is on `to_bits` so a `NaN` component matches itself — a non-finite start is precisely the case that must hit the cache rather than re-sanitise. `the_warm_up_does_not_run_twice_for_the_same_start` pins no-recompute, same-point, and recompute-on-changed-start; mutation-checked, and disabling the lookup reddens only that leg.

**The docs half of item 3.** You flagged that `docs/src/initialization.md` reproduces the same circularity, and my first pass only fixed the code comment. The doc said duals are forwarded untouched *because* displacing the primal would pair a moved iterate with stale multipliers — the consequence stated as the reason. Rewritten to match the code: a caveat, not a safeguard, plus why declining to displace under a warm start is the wrong trade.

**Item 5's first bullet** resolved itself: `main.rs` no longer references `scaling_retry_promoted` at all once the ladder moved to the library, so there are no adjacent duplicate blocks left to merge.

`scripts/sweep-fixtures.sh` against the `f6231f40` baseline moves **no pre-existing line** on either leg — the only diff hunks are `a` (append) hunks for the two fixtures this branch adds.

jkitchin (commented):
Re-reviewed against `2c9a447`, including the library move and the Python surface.

## Verification

Built the branch, plus the Python extension via `maturin develop --release`.

- **All five findings fixed, confirmed.** I re-ran my own reproducers: the Adam sanitise/revert probe is now `a_sanitised_start_survives_a_warm_up_that_does_not_help` verbatim, and my presolve-certified + degenerate `.nl` no longer prints the LICQ paragraph (just `FBBT detected infeasibility` and the verdict).
- **Sweep independently reproduced.** Release `f6231f40` vs `2c9a447`, both legs: 142 → 146 lines, diff is exactly four `>` lines for the two new fixtures. No pre-existing line moved. Matches the claim.
- **Tests green**: `pounce-nlp` / `-algorithm` / `-restoration` / `-rs` / `-cli` no failures; clippy at CI's `-D clippy::correctness -D clippy::suspicious` clean across those plus `-cinterface`; `cargo fmt --check` clean. `pytest python/tests`: **979 passed, 38 skipped, 1 failed** — `test_continuation.py::test_step_controller_is_what_the_jax_follower_uses`, environmental (no `jax` in my venv; `importorskip` does not catch the chained `ModuleNotFoundError` raised inside `pounce/jax/__init__.py`). Not this PR's.
- **The Python promotion path works.** HS008 from the origin through `Problem.solve`: `promoted_by='start_point_perturbation=1e-2'`, `x=[4.60159492, 1.95584361]`, residuals `0.0` and `1.8e-15`, `iter_count=6`. The gate forwards the winner's `finalize_solution` through `PyTnlp` correctly.

Two design calls I checked and agree with. `SecondOpinionTnlp` is a better fix than the CLI snapshot it replaces — `on_converged` only fires on Success/Acceptable (`application.rs:3520`), so dropping the `nominal_capture` snapshot is safe, and the gate reaches `PyTnlp` / `LastSolve` where a per-caller snapshot could not. And the `pounce-rs` collector-scope comment is right: `IterCaptureGuard` is per-solve and `stats.iterations` is assigned per solve (`application.rs:3371`), so dropping the scope early really would empty a promoted rung's capture.

---

## 1. `SecondOpinionTnlp` does not forward `is_presolve_wrapper()` or `scaling_factors()` — `crates/pounce-restoration/src/second_opinion_driver.rs:327`

The `impl TNLP` block forwards eighteen methods but not these two, so both fall to the trait defaults (`false` / `None`). `tnlp.rs:303` states the requirement and the consequence: "A transparent decorator around another TNLP should override this and forward `inner.borrow().is_presolve_wrapper()`. Otherwise the public solve entry point cannot see a presolve wrapper below the decorator and may add a second one." `ConditionedStartTnlp`, written in this same PR, forwards both.

Honest reachability: I could not construct a live failure. `optimize_tnlp` calls `wrap_from_options` unless `presolve_already_applied`, the CLI sets that (`main.rs:1475`), and Python / C / Rust do not pre-wrap — so today every path gets one presolve either way. It bites a caller who calls `wrap_with_presolve` themselves *and* has `presolve=yes`: the original solve gets one wrapper, every rung gets two, and the rung is no longer "baseline plus one knob". `scaling_factors` is masked by `ScalingTnlp` sitting above it (it returns `Some(self.d)`), so that one is latent only. Two lines, and it removes the need to reason about reachability at all.

## 2. `Problem.solve_with_sens` does not run the ladder, and the docs say it does — `crates/pounce-py/src/problem.rs:797`

`docs/src/troubleshooting.md`: "Every single-solve entry point runs it … the CLI, the Python `Problem.solve`, the C `IpoptSolve`, and the `pounce-rs` builder." `solve_with_sens` is a public, tested, single-solve Python entry point and is in neither that list nor the code. Same model, same `x0`:

```
Problem.solve           -> Solve_Succeeded              x = [4.60159492 1.95584361]
Problem.solve_with_sens -> Infeasible_Problem_Detected  x = [0. 0.]   ('second_opinion' key absent)
```

Excluding it may well be right — a displaced start changes which local solution the sensitivity is taken about. But then it deserves what `solve_nlp_batch` got: a comment saying why, a line in the docs, and a test. As it stands two methods on the same object disagree about whether the model is feasible, silently.

## 3. The C interface's narration ignores `print_level` — `crates/pounce-cinterface/src/lib.rs:766`

`&mut |line| eprintln!("{line}")`, ungated. `pounce-cinterface` is the Ipopt drop-in, where `print_level=0 sb=yes` is a request for silence; a failing solve now writes seven `pounce:` lines to stderr. (The CLI does the same — I checked, it emits all seven at `print_level=0` — but that is pre-existing there and new here.) The CLI already gates its verdict re-emit on `print_level >= 1`; the same gate on the `report` closure would do it.

## 4. `pounce-rs` gives a Rust embedder no way to know the ladder ran — `crates/pounce-rs/src/builder.rs:399`

Narration is dropped and `Solution` has no ladder field, so `tried` / `promoted_by` are unreachable. Python gets `info["second_opinion"]`. A Rust caller sees a 4x slower failing solve with nothing to attribute it to. Worth mirroring the field.

## 5. Two doc slips

- `second_opinion_driver.rs:620`: the doc block for `restoring_an_unset_option_leaves_it_unset` is glued onto the front of `each_rung_starts_from_the_baseline` — two comments, one `#[test]` — and the real test at `:673` has none.
- `second_opinion.rs:522`: `start_rung_is_dropped_when_the_baseline_scaling_has_no_tag` still says "rung 3 has a baseline scaling to restore". Rungs no longer restore anything; the field doc above already says `baseline_scaling` is now purely a gate.

## 6. Release note: measure the embedder cost

`Problem.solve` on a failing model, 30-solve average: **61.5 ms with the ladder, 14.5 ms without — 4.2x.** Correct and by design, but it is a default-on change to every existing embedder's failure path. Anyone using POUNCE as a feasibility oracle inside a branch-and-bound or MPC loop pays it on exactly their hot calls. The changelog says "on by default"; worth also giving the number and naming the three `*_retry` opt-outs in the same breath.

---

Items 1 and 2 are the ones I would want settled before merge — 1 because a transparency decorator that is not transparent is a trap for the next caller, 2 because it is a user-visible contradiction between two documented methods.


---

jkitchin:
Thanks — all six addressed. Items 1 and 2 are fixed as asked; 3 turned into a
slightly bigger change than "add the gate", for a reason below.

## 1. `SecondOpinionTnlp` now forwards `is_presolve_wrapper` / `scaling_factors`

Fixed, with your reachability argument written down next to it rather than
re-derived — the point of forwarding is that nobody has to redo that argument
the next time something upstream moves.

`the_decorator_is_transparent_about_what_is_underneath_it` pins it against a
minimal `Underneath` that answers `true` / `Some([2.0, 4.0])`; on the previous
code it reads `false` / `None`.

## 2. `solve_with_sens` — the exclusion is deliberate, and now says so

Kept the exclusion, for the reason you named: a displaced start changes *which*
local solution the sensitivity is about, and the caller's
`pin_constraint_indices` and `deltas` are posed against the one they expected.
Quietly answering about a different local optimum is worse than reporting the
failure. (`SensSolve` also builds its Schur complement from the solve's final
factorization, so a rung would have to carry its factorization out with it.)

It now gets the three things `solve_nlp_batch` got:

- a comment at the call site saying why;
- a row in `docs/src/troubleshooting.md`, in a table of all three deliberate
  exclusions, plus the composition that gets you both — run `solve`, feed its
  `x` back in as `x0`, which makes the base point explicit and reproducible;
- `test_the_ladder_stays_out_of_the_sensitivity_path`, whose mutation guard is
  the pair: on the *same* failing problem the plain `solve` in the same test
  does open the ladder, so `None` is the sensitivity path's choice and not an
  artifact of the model.

One change from what you saw: the key is now **present and `None`**, not
absent, so `info` has the same shape from either method.

## 3. Narration gate — shared, not copied

Gated on both printing surfaces. But I didn't want the same four-line branch
living in two files: one of them silently losing it would look exactly like the
other keeping it. So the predicate is
`pounce_algorithm::second_opinion::narration_is_wanted(options)`, called from
the CLI and the C interface, unit-tested once (`narration_follows_print_level`).

Two notes on the semantics:

- It is `!found || level >= 1`, so only an **explicit** `print_level` silences.
  A bare `OptionsList` reads `0` for an unregistered tag, and I would rather a
  surface that forgets to register it be loud than silent.
- `pounce-rs` and Python don't call it — they never print, they hand the
  narration back.

Pinned end-to-end by `the_ladder_respects_print_level_zero` (CLI): silent at
`print_level=0` for the rank diagnosis *and* the ladder narration, and — the
positive control — all three lines present at `print_level=1`, so the test
can't pass by the ladder having stopped running. On
`degenerate_start_infeasible.nl` that is 11 `pounce:` lines at `print_level=1`
and 1 at `0` (the `.sol` written notice, pre-existing).

This one had a live consequence in the suite:
`issue_508_infeasibility_gap_status.rs::a_non_promoted_second_opinion_does_not_replace_the_solution_it_rejected`
used stderr narration at `print_level=0` as its own "this fixture still
exercises the path" guard, and went red. Its guard now takes a separate
`print_level=1` run; the two runs whose `.sol` files are compared stay at 0.

## 4. `pounce-rs` reports the ladder

`Solution::second_opinion: Option<SecondOpinion>` with `tried`, `promoted_by`
and `log` — the same three fields as Python's `info["second_opinion"]`, and the
narration is collected into `log` rather than dropped. `SecondOpinion` is
exported from the crate root and the prelude.

Three tests, using a model with a singularity at exactly `x == 0` (the value a
modelling layer hands over for a variable nobody initialised):

- `a_recovered_solve_reports_the_rung_that_recovered_it` — rung 3 promotes,
  `tried == ["start_point_perturbation=1e-2"]` (an invalid number opens only
  the rung that moves the point), `log` non-empty;
- `a_succeeding_solve_has_no_second_opinion` — `None` means *did not run*, not
  *found nothing*;
- `turning_the_start_rung_off_gives_the_upstream_verdict_back` — mutation guard
  for the first: same model, rung off, verdict back, field back to `None`.

## 5. Doc slips

Both fixed. The orphaned block moved onto `restoring_an_unset_option_leaves_it_unset`.

For the second I also fixed the sibling you didn't flag —
`barrier_rung_is_dropped_when_the_baseline_scaling_has_no_tag` carried the same
stale "no tag to write back" reasoning, and the module doc said "rung 2 restores
the baseline scaling first" when it is the driver that restores. All three now
say what the `baseline_scaling` field doc says: the gate is conservatism about a
trajectory change on externally-scaled models, not a restore requirement.

## 6. The cost, measured

In the changelog now, as a number and with the opt-outs named in the same
breath. My own measurement, `Problem.solve` on a small failing-and-not-recovered
model, 30 solves averaged over four repeats: **5.5 ms → 21.4 ms, 3.9×**. Yours
was 14.5 → 61.5 (4.2×) on a different model — the absolute numbers are the
model's, the ~4× is the ladder's (one extra solve per rung, no overhead beyond
them). A converged solve still pays nothing, and a model that *recovers* pays
less than 4× because the ladder stops at the first promotion.

---

**Verification.** `scripts/sweep-fixtures.sh` against release `target/release/pounce`,
both legs, 146 lines: **byte-identical** to the sweep on `2c9a447`. Tests green
on `pounce-algorithm`, `-restoration`, `-cinterface`, `-rs`, `-cli`, and
`pytest python/tests/test_second_opinion.py test_sensitivity.py
test_solver_session.py` (26 passed) against a fresh `maturin develop --release`.

**One thing I found and did not chase.** The same toy model
(`x² + 1 ≤ 0`, `x ∈ [-10, 10]`) exits `Infeasible_Problem_Detected` through
`Problem.solve` and `Restoration_Failed` through the `pounce-rs` builder, so
rungs 1–2 are not reachable from that surface on it. It is not caused by this
PR — the builder is L-BFGS-only and finite-differences derivatives by default,
which is enough to explain a different restoration outcome — and it is why the
`pounce-rs` tests use the invalid-number trigger rather than the infeasibility
one. Flagging it as an observation, not a claim; happy to open an issue if you
think it is one.

Issue #764, opened 2026-08-24
Sensitivity hardening roadmap: re-solve oracle, sens-review checklist, scoped index newtypes


jkitchin:
## Summary

Tracking issue for the three pieces of sensitivity-layer hardening agreed with
@devin-griff on #762, so they do not stay as prose in a PR body. #762 itself
lands the first piece (the invariance legs); this records what was deliberately
left out, in the priority order we settled on.

Ordered by value per hour, highest first. **None of these is assigned yet** —
that is the point of filing.

## 1. Re-solve verification oracle (highest value)

The defect class none of the current guards touch is the one with the worst
blast radius, and it is @devin-griff's own framing from `205bb67`:

> All three turn an essentially exact step into a wrong one while reporting
> `improved()` and converged … the residual halved so nothing warned.

A leg that computes the step and checks the reported `improved()` / residual
against an **independent warm re-solve** would have caught all three of those,
and #733's frame mixing besides. It is the methodology the 62k-model validation
already runs out of repo; this is about bringing it in.

Design constraints, from @devin-griff:

- The oracle's steps must sit **above the barrier width**. Below `sqrt(mu)` the
  warm re-solve and the directional-derivative contract genuinely diverge —
  that boundary is what the #756 scale episode mapped.
- The re-solve should **warm-start from the base iterate**, for cost and to
  stay in the same basin.
- Hang it on #762's kink fixture plus one where the step crosses a **strongly
  active** bound, so it also covers the `fix_relax` release path.

One addition from #756's outcome: hang it on a **span** of step sizes rather
than a representative one. The 62k validation measures rms at 1% and 10%, both
far above the slack where #756 was still correct, so an oracle placed only at
large perturbations would have missed that defect too.

## 2. A `sens-review` checklist skill

`.claude/commands/` already has `adversary.md`. A companion checklist encoding
the classes that actually recur here, so the same questions get asked in the
same order by whoever reviews next:

- index space (var-x vs full-x vs KKT rows vs user-g)
- frame and scaling (algorithm's scaled frame vs the model's units)
- absolute thresholds on scale-dependent quantities
- doc drift, where a contract above a function outlives the code below it
- silently-wrong-while-reporting-success

@devin-griff's line for the third item, which is the sharpest statement of it
we have:

> an absolute length compared against a homogeneous quantity breaks linearity
> even when the length is measured rather than chosen

#762's leg 2b is that exact class caught in the wild, so the entry can carry a
worked example instead of a maxim.

[...]

PR #800, opened 2026-08-27
test(sens): re-solve verification oracle over two branch-disjoint fixtures

jkitchin:
Lands **item 1 of #764**, the re-solve verification oracle, to the design
settled on that issue and in @devin-griff's comment there.

Deliberately **no closing keyword**: #764 is a tracking issue carrying three
items, and this is one of them. Item 2 (the `sens-review` checklist skill) and
item 3 (scoped index newtypes) are untouched, so auto-closing it on merge would
lose them. Worth flagging since CLAUDE.md's rule pushes the other way — the rule
is about a PR that fixes a filed issue, and this fixes a third of one.

## Problem

Every guard in `pounce-sensitivity` today compares a number the sensitivity
layer produced against another number the same layer produced, or against a
derivative someone worked out by hand. `sens_invariance_legs.rs` sweeps three
dimensions and every one of its assertions is internal in exactly that sense.
That catches a rule which is not invariant. It cannot catch a step that is
*self-consistently* wrong — the class with the worst blast radius, and the one
`205bb67` named in its own commit message:

> All three turn an essentially exact step into a wrong one while reporting
> `improved()` and converged … the residual halved so nothing warned.

Nothing in the crate reads a number the layer did not produce. Concretely, here
is what that costs, measured on this PR's own release fixture — the corrector
reporting success while the answer is the entire quantity at stake away from
truth:

| | residual | distance to warm re-solve |
|---|---|---|
| step handed in | 9.999999635e-2 | 1.3333333e-1 |
| after 8 corrector iterations | 9.999996705e-2 | 1.3333329e-1 |
| `improved()` | **true** | |
| what `x1` should have travelled off its bound | | 1.3333333e-1 |

## Cause

Not a defect — the held barrier diagonal has no way to represent a bound
leaving the active set, and `correct_step`'s own doc says so. The point is that
the residual is the machinery's opinion of itself, and no internal guard can
distinguish "converged" from "converged against the wrong thing".

## Fix

`crates/pounce-sensitivity/tests/sens_resolve_oracle.rs`. The oracle is a full
re-solve at the perturbed parameter, warm-started from the base iterate (`x`,
`z_l`, `z_u` and `lambda` seeded through `get_starting_point` under
`warm_start_init_point`), at `tol = 1e-10` against the base solve's `1e-8`.
Nothing in the sensitivity layer takes part in computing it.

**The oracle is checked before it is trusted.** Both fixtures have a closed
form at every parameter, and `the_resolve_is_an_independent_oracle` pins the
re-solve against it before any leg compares a step to it — an oracle sharing
the machinery's defect is worse than none. A second precondition,
`the_oracle_out_resolves_the_base_offset`, requires it to stay two orders finer
than the quantity the legs budget against.

**Two truth regimes, per the issue.** Above `sqrt(mu)` the re-solve is the
truth and this file owns it; below it the re-solve and the directional-derivative
contract genuinely diverge by the row's slack and *both are correct*, so the
invariance legs own that region by comparing slopes, which cancels the base
offset. A sub-width oracle arm would fail on correct code. Measured on the kink
fixture — `mu` 9.1e-10, width 3.0e-5, base-to-oracle floor 5.4e-5:

| delta | \|predicted − resolved\| | err/delta |
|---|---|---|
| 1e-2 | 6.36e-5 | 6.4e-3 |
| 1e-4 | 6.35e-5 | 6.3e-1 |
| 1e-6 | 5.47e-5 | 5.5e1 |
| 1e-8 | 5.42e-5 | 5.4e3 |
| 1e-10 | 5.42e-5 | 5.4e5 |

The absolute error is the same throughout — it *is* the base offset — and it
swamps the first-order content as the step shrinks. So the crossover is a band,
not a point (`err/delta ≈ floor/delta` passes 1 where `delta` passes the floor),
and the band between `width` and `100 · width` is deliberately left unasserted
rather than pinned to a sharp edge the data does not show.

**Rejected:** comparing `err/delta` against a budget that grows with `delta`.
It reads as the natural thing for a first-order predictor, and it forgives
exactly the first-order error the leg exists to catch. Both fixtures are QPs
whose exact solution is affine in the parameter on each side, so a correct step
carries no second-order term to allow for and a flat budget is the honest one.

**Also rejected:** the first version of the plain-step leg asserted that
`linear` *agrees* with the re-solve where the active set does not change. The
measurement disproved it — see below.

## Blast radius

Test-only. One new integration test file and a CLAUDE.md section; no `src/`
change, so nothing in the solver or the sensitivity layer moves. The fixture
sweep does not apply.

Three things the oracle measured that were not previously written down
anywhere, all now pinned:

- **The plain step is the two-sided average at a kink**, off by `A/2 · delta`
  against the exact `A · delta` leaving and `0` holding, because the barrier
  linearization at a weakly active bound sits strictly between the two
  one-sided values at every `mu`. That is the documented reason the directional
  mode exists; this turns it into a number. The separation runs 864× at
  `delta = 1e-1` down to 7.8× at `1e-3`, so the leg asserts the *structure*
  (the directional error is the constant floor, the plain error tracks `delta`)
  rather than a flat ratio that would encode wherever the list stops.
- **Across a strongly active release, `linear` and `directional` are both wrong
  by the whole released distance** — `directional` is a no-op with no weak rows
  — and only `fix_relax` reproduces the re-solve. Asserted, so the fixture
  cannot go vacuous.
- **The corrector table above.** Pinned deliberately, in the manner of #762's
  `the_coupled_fixture_carries_an_ambiguous_kink`: a corrector that learns to
  cross a release makes this fail, and that failure is the signal to update it
  on purpose rather than a regression.

## Tests

10 tests. They pass on the parent commit, which is what a guard does — this is
**not** the template's "fails on the parent commit" case, the same as #762. The
equivalent evidence is a mutation table: reintroduce a defect and only the
matching legs go red.

| mutation | reintroduces | red here | red elsewhere |
|---|---|---|---|
| `releasable` returns empty | fix-relax before its release half — "the difference between returning 0.0 and 1.667" in `boundcheck`'s own words | O3 and the corrector-release leg | `boundcheck` unit tests (3), `corrector.rs` (4) |
| `parametric_step_bounded` passes `&[]` for the multipliers | a wiring defect: the refinement called without what a release needs | O3 and the corrector-release leg | `corrector.rs` (4); the `boundcheck` unit tests stay **green**, since they supply their own multipliers |
| `KAPPA_MIN` `1e-3` → `1e2` | #756's class: a genuine kink dropped, so the two-sided step stands | O1 and the two-sided-average leg | 7 of 12 invariance legs, 5 `degenerate_*` files |

The separation is the point: the release mutations move only the release legs
and the kink mutation moves only the kink legs, so a red leg names the dimension
that broke.

**Stated plainly rather than buried: none of the three is caught by this file
alone.** The argument for it is not that it finds what nothing else can — it is
that it is the only guard in the crate reading a number the sensitivity layer
did not produce, and the class #764 filed it for is the one that survives every
internal check. Row 2 is the closest demonstration: a defect where the algorithm
is correct and the call site is not passes every unit test in the module it
breaks.

**Two fixtures, chosen to reach different branches** per CLAUDE.md's rule —
#762's kink (its pin's RHS lifted to a field so the parameter can move), and a
**strongly** active bound with multiplier 0.6 that a large enough perturbation
drives negative. That second one is upstream's equation 18, the release half of
fix-relax, which no fixture in `sens_invariance_legs.rs` reaches at all.
`the_fixtures_take_different_branches` asserts the split rather than assuming
it.

**Deliberately NOT pinned**, and the module docs say so in their own section:

- **Index spaces.** Both fixtures have `m = 1`, so the pin's user-g index and
  its KKT row coincide and a defect confusing them is invisible here.
  `cd_split_pin_mapping.rs` already owns that dimension end-to-end including
  `correct_step`, so a split-index fixture here would be redundant rather than
  additive.
- **Scaling.** Both run unit-scaled; leg 1 and `variable_scaling_sensitivity.rs`
  own it.
- **`improved() == false`.** No measured point on either fixture reaches it, so
  that arm is written but unexercised and this file is no evidence about it. A
  fixture whose corrector declines to improve is the missing one.
- **Scale.** Three variables each. The 62k resource paths (#672 f2's `2^n`
  masks, #708 f4's full-length basis columns) are out of reach by construction,
  as #764 puts them out of scope.

---

- [x] Tests fail on the parent commit for the stated reason — **N/A**: these are
      guards, not regression tests for a fix, so there is no parent commit they
      fail on. Discharged by the mutation table above
- [ ] `CHANGELOG.md` `[Unreleased]` entry — deliberately none: test-only, no
      user-visible behaviour change, same call as #762. Say the word if you want
      one
- [ ] Book page under `docs/src/` — N/A; documented in CLAUDE.md alongside the
      fixture sweep and the invariance legs, which is where those contracts live
- [x] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean. Locally:
      fmt clean, `cargo clippy -p pounce-sensitivity --all-targets -- -D
      clippy::correctness -D clippy::suspicious` exits 0, `cargo test -p
      pounce-sensitivity` green (214 passed, 0 failed, 10 of them in this file).
      I originally flagged `cargo test --workspace` as unrun here, because
      `pounce-hsl` will not link without `COINHSL_DIR`; **CI has since closed
      that gap** — `Test (ubuntu-latest)` runs `cargo test --workspace --exclude
      pounce-hsl`, excluding that crate as standing repo convention rather than
      as a local workaround, and all 11 checks are green on `656f607`
      as it stands now. Three corrections were made rather than left standing,
      since they are the exact class this box exists for: a leg asserting the
      plain step *agrees* with the re-solve (the measurement said it is the
      two-sided average, so the leg was rewritten to assert that instead); three
      doc comments describing an earlier design of the budget formula and the
      `SPAN` constant; and the workspace-test caveat above, which was true when
      written and stale once CI ran

The fixtures trip `clippy::unwrap_used` the same way the sibling sensitivity
test files do; that lint is in the `restriction` group, which `ci.yml` keeps as
an informational warn — the gate it denies is `correctness` + `suspicious`, and
that is what exits 0 above.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -195,6 +195,72 @@ coupled, because the ratio is `reduced/diagonal` (gh#763). Never use the
 activity class as a proxy for kink-ness — that inference is exactly what
 produced the gh#756 defect.
 
+## The re-solve oracle: the only guard that reads an outside number
+
+`crates/pounce-sensitivity/tests/sens_resolve_oracle.rs` (gh#764 item 1)
+checks the step against a **warm re-solve at the perturbed parameter**,
+tolerance two orders tighter than the base solve, started from the base
+iterate. Every other guard in the crate — the invariance legs included —
+compares a number the sensitivity layer produced against another number
+it produced, or against a derivative someone worked out by hand. Those
+catch a rule that is not invariant. They cannot catch a step that is
+*self-consistently* wrong, which is the class with the worst blast
+radius and the one `205bb67` named: "turn an essentially exact step into
+a wrong one while reporting `improved()` and converged … the residual
+halved so nothing warned."
+
+**The oracle owns steps above the barrier width, and nothing below it.**
+Below `sqrt(mu)` the warm re-solve and the directional-derivative
+contract genuinely diverge, by an amount of order the row's slack, and
+*both are correct*: the converged base point sits `O(sqrt(mu))` off the
+exact solution, because at a kink `s · z = mu` with both factors
+vanishing. So an oracle arm placed below the width **fails on correct
+code**. The invariance legs own that region instead, comparing slopes,
+which cancels the base offset exactly. Measured on the kink fixture:
+`mu` 9.1e-10, width 3.0e-5, base-to-oracle offset 5.4e-5, and `err/delta` runs
+6.4e-3 at `delta = 1e-2` to 5.4e5 at `1e-10` — the same absolute error
+throughout, swamping the first-order content as the step shrinks. The
+crossover is a band, not a point, and the band is deliberately left
+unasserted.
+
+Two fixtures, chosen to reach **different branches**, per the rule
+above:
+
+- a weakly active kink (gh#762's model, with the pin's RHS lifted to a
+  field), which reaches the directional decision;
+- a **strongly** active bound with multiplier 0.6 that a large enough
+  perturbation drives negative — upstream's equation 18, the *release*
+  half of fix-relax, which no fixture in `sens_invariance_legs.rs`
+  reaches at all.
+
+Measured on those two: the plain step is the **two-sided average** at a
+kink, off by `A/2 · delta` — the directional mode exists to repair
+exactly that, and the oracle is what turns the claim into a number.
+Across a strongly active release, `linear` *and* `directional` are both
+wrong by the full released distance (`directional` is a no-op with no
+weak rows); only `fix_relax` reproduces the re-solve.
+
+**What a falling residual does not tell you.** Across a release the
+corrector reports `improved()` and drops its residual by 3e-8 while the
+point stays 0.1333 from the truth — the whole distance the variable
+should have travelled off its bound. That is not a defect: the held
+barrier diagonal cannot represent a bound leaving the active set, and
+`correct_step` says so. It is the gh#764 thesis as a number —
+`improved()` plus a converged residual does **not** imply the answer is
+close, and no internal guard can tell you that. It is pinned, so a
+corrector that learns to cross a release fails the test deliberately
+rather than silently.
+
+The file carries its own mutation table and, in the same breath, the
+list of what it is **not** evidence about: index spaces (both fixtures
+have `m = 1`, so g index and KKT row coincide — `cd_split_pin_mapping.rs`
+owns that), scaling (both run unit-scaled — leg 1 and
+`variable_scaling_sensitivity.rs` own it), the `improved() == false`
+branch (unexercised by this corpus), and anything that only appears at
+62k. Add a row to the oracle for a change that reroutes which
+correction a mode reaches for; check first which branch the fixture you
+are leaning on actually takes.
+
 ## Working GitHub issues
 
 When opening a PR that fixes a filed issue, the PR **body** (not just the
--- a/crates/pounce-sensitivity/tests/sens_resolve_oracle.rs
+++ b/crates/pounce-sensitivity/tests/sens_resolve_oracle.rs
@@ -1,0 +1,1201 @@
+//! The re-solve verification oracle: check the sensitivity step
+//! against an INDEPENDENT warm re-solve, over a span of step sizes.
+//!
+//! # Why an oracle, when the invariance legs exist
+//!
+//! `sens_invariance_legs.rs` sweeps three dimensions a shipped defect
+//! hid in, and every one of its assertions is an *internal* one: two
+//! arms of the same machinery must agree, and both must match a
+//! closed-form derivative. That catches a rule that is not invariant.
+//! It cannot catch a step that is self-consistently wrong.
+//!
+//! The defect class with the worst blast radius is exactly that one,
+//! and `205bb67`'s own framing names it:
+//!
+//! > All three turn an essentially exact step into a wrong one while
+//! > reporting `improved()` and converged … the residual halved so
+//! > nothing warned.
+//!
+//! Every guard in the crate today reads a number the same machinery
+//! produced. This file reads one it did not: a full re-solve at the
+//! perturbed parameter, warm-started from the base iterate, at a
+//! tolerance two orders tighter than the base solve. Nothing in the
+//! sensitivity layer participates in computing it.
+//!
+//! # The two truth regimes
+//!
+//! An oracle cannot be hung on every step size, and the reason is not
+//! a tolerance to be tuned. Below the barrier width the warm re-solve
+//! and the directional-derivative contract genuinely diverge, by an
+//! amount of order the row's slack, and **both are correct**. The
+//! converged base point sits `O(sqrt(mu))` off the exact solution --
+//! at a kink `s · z = mu` with both factors vanishing, so the slack is
+//! `sqrt(mu)` -- and the step is measured from there.
+//!
+//! So the composition is:
+//!
+//! | regime | truth | owned by |
+//! |---|---|---|
+//! | `delta >> sqrt(mu)` | the warm re-solve | this file |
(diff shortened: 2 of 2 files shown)


jkitchin:
@devin-griff would you like to take a look at this?

devin-griff:
Verified end to end: all ten pass here, I read the file whole, checked both fixtures' closed forms against their stationarity algebra, and reproduced the `KAPPA_MIN` mutation row exactly, O1 and the two-sided leg red and nothing else.

The design matches what we settled on #764 and improves it in two places worth naming: the oracle is verified against closed forms before anything trusts it, with the tight and loose bounds split by where the fixture's own multiplier vanishes, and the regime crossover is left as an unasserted band instead of a sharp edge. Both are right, and the second is one I would have gotten wrong by pinning `sqrt(mu)` exactly.

Two small things. Doc drift, your own checklist class: `the_oracle_out_resolves_the_base_offset`'s comment says the oracle is three orders finer than the floor at every step size the leg uses, and the assert enforces two, `floor / 100`. And the fine end of O1's span asserts weakly: at `delta = 1e-4` the budget, about `2.2e-4`, exceeds the entire exact step, `1.1e-4`, so a zero step would pass those entries. The power lives in the large end plus the separation leg, which is fine, but a reader trusting O1 alone at the fine end would over-read it, and one line in the not-evidence-about section would close that.

On the corrector pin: `the_corrector_reports_improved_without_crossing_a_release` is the acceptance test the in-loop active-set corrector work will flip, and your "good kind of red" note is exactly how I intend to trip it. Good to have the tripwire in place before that work starts.

Ready to merge from my side.

jkitchin:
Both fixed in `cb5888f` — doc-only, no assert or fixture moved. Flagging that the head changed after your sign-off.

**The margin claim.** You're right, and the assert is the half that was correct. I measured `floor / err` across the span rather than just matching the comment to the code, because the interesting part turned out to be that the margin isn't uniform:

| delta | 1e-1 | 1e-2 | 1e-3 | 1e-4 |
|---|---|---|---|---|
| `floor / err` | 6.0e5 | 4.4e4 | 2.5e3 | **5.4e2** |

The re-solve has less parameter distance to work with as `delta` shrinks, so the fine end is the binding one at ~2.7 orders — which is why two is the honest bound and three was never true anywhere except the coarse end. The comment now carries the range and names `1e-4` as the entry that would close first.

**O1's fine end.** Confirmed by measurement, and narrower than I'd have guessed from the argument alone — it is *only* the `±1e-4` pair:

| delta | null-step error | budget | null passes? |
|---|---|---|---|
| ±1e-1 | 1.1e-1 | 2.17e-4 | no |
| ±1e-2 | 1.1e-2 | 2.17e-4 | no |
| ±1e-3 | 1.0e-3 | 2.17e-4 | no |
| ±1e-4 | 1.0e-4 | 2.17e-4 | **yes** |

So at `1e-3` a null step is already five budgets out and the margin grows from there. Recorded in the not-evidence-about section as you suggested, rather than by tightening the budget: making `1e-4` bite on its own means squeezing the threshold between the measured error `6.3e-5` and that null step's `1.0e-4`, at most `1.58×` of headroom. That trades a robust leg for a flaky one to duplicate power the coarse entries and the separation leg already carry — but if you'd rather have the assertion than the caveat, say so and I'll take the trade.

On the corrector pin — good, that's exactly the intent. When the in-loop active-set corrector crosses a release, `the_corrector_reports_improved_without_crossing_a_release` going red is the signal to update the recorded distance on purpose; the numbers in it are a snapshot of today's behaviour, not a contract to preserve.

---

PR #808, opened 2026-08-27
sens: the review checklist and scoped index newtypes (gh#764 items 2-3)

### LINKED ISSUE #764: Sensitivity hardening roadmap: re-solve oracle, sens-review checklist, scoped index newtypes

jkitchin:
Closes #764 — items 2 and 3, plus the two review fixes from item 1 that
missed the merge window on #800.

Item 1 landed as #800 at `656f607`. This carries the rest, so the closing
keyword is correct here where it was deliberately omitted there.

## Problem

Three separate things, smallest first.

**The item 1 doc fixes never reached `main`.** #800 merged at 15:57:36; the
fix commit for @devin-griff's two review findings landed ~40 seconds later and
the merge went in against the pre-fix head. `main` has been carrying a comment
claiming the oracle is three orders finer than the floor while the assert
enforces two.

**Item 2:** the recurring defect classes in `pounce-sensitivity` live in commit
messages, issue threads and CLAUDE.md. Whoever reviews next asks whichever
subset they happen to remember.

**Item 3:** two index spaces — `var_x` (the primal block) and `full_x` (the
model's variables) — coincide until the first `make_parameter`-removed variable
and diverge after it. A swap returns a *neighbouring variable's* answer: in
range, plausible, wrong. That is gh#450, and gh#672 finding 1 shipped it again
as `report.var_status.get(var_row)`.

## Cause

For item 3 specifically, the reason it keeps recurring is in the issue's own
words, and it is a property of *how* the index is consumed rather than of the
index:

> **Table lookup fails loudly; direct array indexing fails silently.**

A released row resolved through `rows.iter().find(|b| b.row == r)?` turns a
space-swap into `SensComputationFailed`, immediately and every time. A direct
`Vec` index does not.

### Item 2 — `.claude/commands/sens-review.md`

A companion to `/adversary` in the same directory and form: that one hunts
wrong answers from the outside, this one reads a diff. Eight entries, ordered
by **how silently the class fails** — entry 1's defects return a plausible
neighbour, entry 8's announce themselves.

Entries 1–5 and the ordering are the issue's list. 7 and 8 are @devin-griff's
additions from the issue comment (name the measured populations behind a new
threshold; confirm which binary the harness loaded). Each entry is *Ask / Why /
Discharge*: the question, a worked example with a file and symbol, and the test
that settles it.

**Entry 6 is beyond the filed list** and is the one to push back on if you want
it out. It is "which branch does the fixture actually take?" — CLAUDE.md's most
emphasized sensitivity rule, and the one whose absence let gh#756's defect pass
a green suite, since the invariance legs' kink certifies `WEAKLY_ACTIVE` and
that PR's rule gave certified rows an early return. It carries the gh#763
corollary that `AMBIGUOUS` is not "probably not a kink".

Every citation was checked against the source rather than against CLAUDE.md's
prose, which caught one error in the draft: it claimed `nat_sigma` is
`Sigma/mu`, but `ActivityReport::var_sigma` is the barrier diagonal `Σ = z/s`
in natural units. The conclusion survived — `nat_sigma[k] * own` is a curvature
times an inverse curvature, dimensionless, which is *why* a bare `KAPPA_MIN` is
legitimate at `solver.rs:1243` — but the stated reason was wrong and is fixed.

### Item 3 — `crates/pounce-sensitivity/src/index.rs`

`VarX`, `FullX`, `VarToFull`, `FullXSlice`. **Not** the 342-site sweep: that
changes `SensBacksolver` (re-exported from the crate root) and reaches 31
`pounce-py` sites where indices cross into Python as `i64`. Re-running the
measurement, the crate has 11 conversion sites and only **two** match the shape
that earns a newtype — a conversion feeding a direct array index in a scope
where the other space is also live. Both are in `solver.rs`:

- the kappa `var_sigma` read, whose `unwrap_or(0.0)` turns a miss into a zero
  that silently drops the row from the engaged set — @devin-griff's "exactly
  one new cross-space read" from the issue comment;
- `weakly_active_bounds`, whose loop body indexes `ctx` by var-x **and**
  `report` by full-x with a bound row live as a third space.

The other nine are left alone and the module says why for each: four
`activity.rs` scatter loops convert at the point of use with one space live;
`algorithm_backsolver.rs:871` fails loudly twice over (checked `get_mut`, then
an unfilled-NaN sweep); the rest *produce* indices rather than index with them.

**`FullX` has no public constructor.** It is obtainable only by putting a
`VarX` through `VarToFull`, which moves the assertion "this number is in
full-x" to once per map instead of once per read.

**Rejected, and the measurement is why.** The first draft made `FullX::new`
public. Its own first mutation showed `FullX::new(var_row.get())` compiling
clean — the swap the module exists to prevent, one short line away, with the
docs claiming otherwise. A newtype whose guarantee is not mutation-checked is a
comment with a `struct` around it.

## Blast radius

Additive: new `pub` module, no existing signature or field changed.
`pounce-py` and `pounce-rs` build unchanged, so `SensBacksolver` and the 31
Python `i64` sites are untouched. No behaviour change — the two converted sites
compute exactly what they did, and no `src/` logic moves — so the fixture sweep
does not apply.

The `.gitignore` change is the one piece with mechanical effect outside the
crate: `.claude/` was excluded wholesale under a comment reading "local state",
so a new command silently failed to stage. `adversary.md` is tracked only
because it was force-added past the rule. Now `.claude/*` with an exception for
`commands/`; verified `settings.local.json` and every other subdirectory stay
ignored.

## Tests

Item 3's evidence is a mutation table, in the module rather than only here,
because these are guards and pass on the parent commit:

| mutation | outcome |
|---|---|
| mint a `FullX` from the var-x row at the kappa read | **compile error**, `E0624: associated function 'new' is private` |
| the same at `weakly_active_bounds` | **compile error**, same |
| bypass the type: `report.var_status.get(row.get())` | **compiles.** Caught at runtime by `sens_invariance_legs.rs` leg 3 — three legs go red |

The third row is the honest limit and the reason **leg 3 is not retired**:
`ActivityReport`'s `Vec` fields are public and must stay, so the type closes
the typed path and the leg fences the untyped one. Neither alone covers the
site — which is @devin-griff's "part of it is already fenced" as a measurement.

Four `compile_fail` doctests pin the type-level half, each **paired with a
passing twin**, because a `compile_fail` that fails on a typo or an unresolved
import passes vacuously — the failure mode this crate cares most about. Two
were extracted and compiled by hand to confirm they fail with `E0308`
(mismatched types) rather than a resolution error.

249 passed, 0 failed on `pounce-sensitivity` (up from 238 pre-merge; main's
`reduced_row_activity.rs` accounts for the rest). `cargo fmt --all -- --check`
clean, `cargo clippy -p pounce-sensitivity --all-targets -D
clippy::correctness -D clippy::suspicious` exits 0.

**Deliberately not pinned.** Item 2 is a document: nothing executes it, nothing
in CI enforces it, and it fails on nobody. It is a review aid, not a guard, and
the difference from item 1 is real — `sens_resolve_oracle.rs` goes red on its
own and `/sens-review` never will. Three of its entries (1, 4, 8) are
mechanically checkable and could grow teeth later; entries 2, 3, 5, 6, 7 need
judgement about what a quantity means and would produce noise if automated.
That is a follow-up worth its own issue, not scope to widen here.

`main` was merged in at `18f289f` after #806 touched `solver.rs` by 87 lines;
revalidated rather than trusting the clean merge, and re-counted the conversion
sites to confirm #806 added none — the module's "two sites" claim still holds.

---

- [x] Tests fail on the parent commit for the stated reason — **N/A** for item
      2 (a document) and for the item 1 doc fixes. For item 3 the equivalent is
      the mutation table above, where two of three mutations do not compile
- [ ] `CHANGELOG.md` `[Unreleased]` entry — deliberately none. Item 2 is a
      review aid, item 3 is additive with no behaviour change, and the doc
      fixes are comments. Say the word if you want one for the new public
      module
- [ ] Book page under `docs/src/` — N/A; both items are documented in CLAUDE.md
      beside the guards they relate to, which is where those contracts live
- [x] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean, and
      `cargo doc -p pounce-sensitivity` produces **zero** warnings attributable
      to `index.rs`. The 15 remaining doc warnings are pre-existing
      `reduced_activity` links to a private item, from `main`
      as it stands now. Two corrections were made rather than left standing,
      both found by running the checklist's own entry 4 against this diff: a
      `VarToFull` doc linking to `Solver::var_to_full`, a method that was never
      written (rustdoc flags it; the call sites build the map inline), and a
      paragraph still describing the first draft's escape hatch after `FullX`
      became unforgeable

---
_Generated by [Claude Code](https://claude.ai/code/session_01P8W6S9roszB5YXFgZGxmyX)_


jkitchin:
@devin-griff this is close to finishing out this issue I think. LMKWYK.

devin-griff:
Verified: 249 pass on the branch, the doctest pass runs all four compile-fail cases with their passing twins, both converted sites compute exactly what they did with the reads now typed, the checklist's citations spot-checked against source, and the `.gitignore` carve-out confirmed in both directions, `settings.local.json` still ignored and `commands/` tracked.

On the one thing you flagged for pushback, entry 6: keep it. It is the gh#756 postmortem distilled to the question that would have caught the defect in review, and its gh#763 corollary now has an API remedy to point at rather than only a warning, which is what separates a checklist entry from a maxim.

The two item-1 follow-ups are addressed beyond what the review asked. The fine-end note in particular: working out that a null step passes exactly the two `±1e-4` entries and that tightening the budget would leave `1.58x` of headroom is the analysis my one-line observation should have contained.

Worth recording how item 3 ended: the scope converged below what either of us argued on the issue, 342 sites to 11 to 2, each cut by a measurement, and the module documents the rule that did the cutting. The private-constructor story, the first draft failing its own first mutation, is the discipline applied to the guard itself.

Ready to merge from my side, and with it #764 closes.

PR #809, opened 2026-08-27
perf(kkt): merge the limited-memory SMW factorize+batch into one backend call

jkitchin:
**Draft.** The change is structurally right and bit-identical, but the measured gain on FERAL is inside run-to-run noise. The number that would justify merging is an MA57 one, and I cannot produce it here — see **Why this is a draft** at the bottom.

Follow-up to [#730](https://github.com/jkitchin/pounce/issues/730), prompted by @srikanth-gm's re-measurement in [this comment](https://github.com/jkitchin/pounce/issues/730#issuecomment-5441740884). No closing keyword: #730 is already closed by #735, and this addresses part of the residual that comment identified.

## Problem

@srikanth-gm re-ran a 59,939-variable CasADi direct-collocation model under `hessian_approximation=limited-memory` against the current build and reported POUNCE/MA57 now within 1.6% of Ipopt/MA57 on wall time — but still spending **+39% solver-internal time per iteration**, of which the largest named component was **+41% on back-solve**, with numeric factorization at parity (+2.9%).

Their per-iteration figures:

| per iteration (s) | Ipopt/MA57 | POUNCE/MA57 | Δ |
|---|---:|---:|---:|
| numeric factorization | 0.1595 | 0.1642 | +2.9% |
| back-solve | 0.1301 | 0.1836 | **+41%** |
| solver internal | 0.3423 | 0.4763 | +39% |

Two observations on that decomposition, which are worth recording because they bound what this PR can be expected to fix:

- **"Single largest identifiable component" is doing real work in that sentence.** Subtracting both linear-algebra rows from solver-internal leaves 0.0758 s/iter — **57% of the gap sits in neither row**, versus back-solve's 40%. In relative terms it is the bigger discrepancy (POUNCE 0.1285 vs Ipopt 0.0527, +144%). This PR does not touch it, and it remains unmapped.
- **The two back-solve rows are not measured over the same region.** Ipopt's `LinearSystemBackSolve` wraps `ma57c` and nothing else (`IpMa57TSolverInterface.cpp:756-807`). POUNCE's guard wraps `TSymLinearSolver::multi_solve`, which also applies the RHS scale and solution unscale — two passes over 118,276 doubles per call, with `linear_system_scaling=slack-based` on — where Ipopt bills those to a separate `LinearSystemScaling` row. The magnitude is small (~0.07% of the row) but the rows are not strictly comparable.

## Cause

Under limited-memory, `LowRankAugSystemSolver` builds its Sherman-Morrison-Woodbury correction by solving the augmented system for the low-rank columns. Upstream issues **one** `MultiSolve` over all of them (`IpLowRankAugSystemSolver.cpp:487`), so the factorization and every column's back-substitution share a single traversal of the factor.

POUNCE split that in two: a single-RHS `solve` to pay the factorization, then a batched `try_resolve_many_flat` for the remaining columns. A sparse triangular solve streams the whole factor once per call and then applies it to each column, so its cost is `F + nrhs·W`. Instrumenting the backend entry point on `benchmarks/large_scale/laptime` and fitting the two best-sampled widths gave **F ≈ 19 ms, W ≈ 7 ms** — the fixed traversal is roughly 3× the marginal per-RHS cost. The split therefore discarded one full `F` on every SMW update.

Measured call distribution over 40 iterations (179 backend calls), before the change:

```
nrhs  refactor  calls  median
   1     true      42   271 ms   <- factorize, serving ONE rhs
   1     false     63    26 ms
   5     false     34    59 ms
   6     false     33    62 ms
```

## Fix

A factorizing multi-RHS path, `AugSystemSolver::try_solve_many_flat`, implemented on `StdAugSystemSolver` as a single `TSymLinearSolver::multi_solve(new_matrix = true, nrhs)`. The cold SMW block takes it, so all columns ride along with the factorization.

The gh#729 bit-identity gate still governs and is only widened: the merged call is taken solely when the backend affirms `multi_solve_matches_single_solve` at the full column count. The old split remains the fallback, and both KKT-dump surfaces force it so a dump run keeps its one-record-per-solve shape.

**What was tried and rejected.** The first cut gated the merged path on `dim == inner.system_dim()`, mirroring the warm batch below it. That is unanswerable before `assemble()` — cold, `system_dim()` is still 0 — so it rejected the first factorization of every run. It was not inert: later updates took the merged call once `dim` was set, so the timing looked right. It also meant the existing test never executed the new code, because the mock's `system_dim()` is 0 whenever it holds no factor. That is the gh#756 failure mode from CLAUDE.md verbatim. The precondition is gone; `try_solve_many_flat` now assembles and then declines with `None` on a packed-length mismatch, which asks the same question where the answer exists. Assembling and declining is safe — `assemble` is idempotent and no factorization ran.

Also caches the deprecated `POUNCE_DUMP_KKT` probe. It was an environment lock plus a linear scan of `environ` on every solve, for a surface that is off in every production run.

## Blast radius

Baseline `74c96ec^` (= current `main`).

**Fixture sweep: empty diff**, 152 fixture-legs, both `exact` and `lbfgs`, all three engines (36 `cvx-qp`, 35 `nlp`, 5 `cvx-qcqp` per leg). No status, objective, iteration count or engine moved.

**Trajectory bit-identical** over 100 iterations on `laptime` (58,014 vars / 62,014 cons, limited-memory), in both A/B pairs — every printed digit of objective, dual infeasibility, constraint violation and complementarity. The baseline is deterministic run-to-run, so the comparison is single-variable.

Timing, two interleaved pairs, only the binary changing:

| row | base 1 | base 2 | merged 1 | merged 2 | mean Δ | rel |
|---|---:|---:|---:|---:|---:|---:|
| LinearSystemBackSolve | 34.891 | 33.567 | 24.093 | 23.282 | −10.541 s | −30.8% |
| LinearSystemFactorization | 57.014 | 54.140 | 65.438 | 63.898 | +9.091 s | +16.4% |
| **linear algebra total** | 91.927 | 87.727 | 89.548 | 87.197 | **−1.454 s** | **−1.6%** |
| OverallAlgorithm | 114.794 | 110.167 | 112.964 | 110.321 | −0.838 s | −0.7% |
| callbacks | 9.780 | 9.577 | 9.888 | 9.710 | +0.120 s | +1.2% |

**Read the back-solve row with care: most of its drop is bookkeeping.** The merged call bills its substitutions to `LinearSystemFactorization` by construction, so −31% on one row against +16% on the other nets to −1.45 s. Baseline run-to-run spread is 4.2%, so **the net is inside noise on wall time and should not be quoted as a wall-clock win.**

It is, however, the size the mechanism predicts — ~20 ms per removed traversal over ~75–90 SMW updates ≈ 1.5–1.8 s — so it is real work removed, just not much of it on this model and this backend. Callback time is flat, which is the control: identical trajectory means identical callback work, so the delta is POUNCE's own.

Earlier in this branch I reported −7.8% overall. That does not reproduce and should be disregarded: the host changed mid-investigation, and the identical baseline binary now takes 110–115 s where it took 72–76 s. The figures above are the corrected ones.

## Tests

`batched_smw_columns_match_the_per_column_path` is extended to pin the merged call: **one** factorizing multi-RHS call carrying all 3 V columns, **zero** separate back-solve passes over the same factor, and still bit-identical to the per-column arm. The single remaining `resolve` is the actual RHS being solved against that factor, which upstream pays too.

**It bites.** With the merged path disabled (`&& false` on the gate) it fails at `factor_batched_calls >= 1` — "the packed mock must actually reach the merged arm" — which is the assertion that would have caught the first cut's dead gate. On the parent commit it fails for the same reason.

Deliberately not pinned: the timing itself. It is inside noise here and machine-dependent, so a performance assertion would be flaky. The sweep and the bit-identity assertion are what guard correctness.

- 960 tests pass, 0 failed, across 62 binaries (`pounce-algorithm`, `pounce-nlp`, `pounce-linsol`)
- `cargo fmt --all --check` clean
- CI's `clippy -D correctness -D suspicious` clean; warning count on the three touched files is 29 before and 29 after

## Why this is a draft

The reported +41% back-solve is an **MA57** number on @srikanth-gm's model. Everything above is **FERAL** on `laptime`, because there is no coinhsl in this container. The `F + nrhs·W` cost shape holds for any supernodal triangular solve, so the direction transfers — but `F/W` is exactly the ratio that decides whether this is worth 1.5 s or considerably more, and it is backend-specific. On FERAL the answer is "real but inside noise", which is not on its own a reason to merge.

What would settle it: this branch measured under MA57, ideally on the reporting model. @srikanth-gm, if you have the appetite for another A/B, that is the run that would tell us whether this moves your back-solve row — the trajectory is bit-identical, so it is a clean swap of `pounce_cinterface` with nothing else changing.

Independent of the timing, the change makes POUNCE's SMW call structure match upstream's and removes a demonstrably redundant pass over the factor, which may be worth taking on its own terms.

---
_Generated by [Claude Code](https://claude.ai/code/session_018DSYzDXdPyz4fEukH1bBZS)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1143,6 +1143,48 @@ changes.
   because a fitted vertex value is blind to a uniform rescaling of the
   abscissa — which is why the arclength is asserted beside it rather than
   the temperature being trusted alone.
+- **Limited-memory: the SMW update now costs one pass over the KKT factor,
+  not two (gh #730 follow-up).** Under `hessian_approximation=limited-memory`,
+  `LowRankAugSystemSolver` builds its Sherman-Morrison-Woodbury correction by
+  solving the augmented system for the `2q` low-rank columns. Upstream issues
+  a single `MultiSolve` over all of them
+  (`IpLowRankAugSystemSolver.cpp:487`), so the factorization and every
+  column's back-substitution share one traversal of the factor. POUNCE split
+  that in two: a single-RHS `solve` to pay the factorization, then a batched
+  `try_resolve_many_flat` for the rest. A sparse triangular solve streams the
+  whole factor once per call and then applies it to each column, so its cost
+  is `F + nrhs·W` — and on a large KKT `F` is several times `W`. The split
+  therefore threw away one full `F` on every SMW update.
+
+  There is now a factorizing multi-RHS path (`try_solve_many_flat`) and the
+  cold SMW block takes it, so all columns ride along with the factorization.
+
+  Be careful reading the timing rows, because most of the movement between
+  them is bookkeeping: the merged call bills its substitutions to
+  `LinearSystemFactorization`, so `LinearSystemBackSolve` falls by far more
+  than any work is removed. Measured on `benchmarks/large_scale/laptime`
+  (58 014 variables, 62 014 constraints, FERAL, 100 iterations, two
+  interleaved A/B pairs): back-solve 34.2 s → 23.7 s (−31%) against
+  factorization 55.6 s → 64.7 s (+16%), for a **net −1.45 s of 89 s of
+  linear algebra, −1.6%**, and −0.7% overall. The baseline's own run-to-run
+  spread is 4.2%, so the net is **inside noise on wall time** and should not
+  be quoted as a wall-clock win. It is, however, the size the mechanism
+  predicts — roughly 20 ms per removed factor traversal over ~75–90 SMW
+  updates — so it is real work removed, just not much of it on this model
+  and this backend.
+
+  The trajectory is **bit-identical** over all 100 iterations in both pairs
+  and the fixture sweep is empty on both legs, so nothing was re-associated.
+  Whether the effect is larger under MA57, where the fixed-versus-per-RHS
+  cost ratio differs, is unmeasured here — no coinhsl in the dev container.
+
+  The gh#729 bit-identity gate still governs: the merged call is taken only
+  when the backend affirms `multi_solve_matches_single_solve` at the full
+  column count, and the old split remains as the fallback. Reported by
+  @srikanth-gm, whose measurements on a 59 939-variable CasADi collocation
+  model put POUNCE/MA57 within 1.6% of Ipopt/MA57 on wall time while spending
+  41% more per iteration on back-solve; this addresses part of that row.
+
 - **The NLP arm no longer certifies a constrained maximum as
   `Solve_Succeeded` (gh #797).** On the CLI fixture `nonconvex_qp.nl` —
   `min x₀·x₁ s.t. x₀ + x₁ = 2, 0 ≤ x ≤ 4`, whose restriction to the feasible
--- a/crates/pounce-algorithm/src/kkt/aug_system_solver.rs
+++ b/crates/pounce-algorithm/src/kkt/aug_system_solver.rs
@@ -231,6 +231,36 @@ pub trait AugSystemSolver {
         None
     }
 
+    /// Factorize **and** solve `nrhs` right-hand sides in a single
+    /// backend call, with `packed_rhs` in the same column-major layout
+    /// [`Self::try_resolve_many_flat`] uses. Solutions overwrite
+    /// `packed_rhs` in place.
+    ///
+    /// This is the factorizing counterpart of `try_resolve_many_flat`,
+    /// and it exists to reproduce upstream's single
+    /// `AugSystemSolver::MultiSolve` (`IpLowRankAugSystemSolver.cpp:487`)
+    /// in one step rather than two. Splitting that call into a
+    /// single-RHS `solve` followed by a batched `try_resolve_many_flat`
+    /// costs one **extra full traversal of the factor**: a sparse
+    /// triangular solve streams the whole factor once per call and then
+    /// applies it to every column, so its cost is `F + nrhs·W` with `F`
+    /// several times `W` on a large KKT. Merging the two calls removes
+    /// one `F` per Sherman-Morrison-Woodbury update.
+    ///
+    /// Returns `None` when the backend does not support the path, in
+    /// which case the caller keeps the split. Inertia bookkeeping
+    /// (`check_neg_evals` / `num_neg_evals`) matches [`Self::solve`].
+    fn try_solve_many_flat(
+        &mut self,
+        _coeffs: &AugSysCoeffs<'_>,
+        _packed_rhs: &mut [Number],
+        _nrhs: usize,
+        _check_neg_evals: bool,
+        _num_neg_evals: Index,
+    ) -> Option<ESymSolverStatus> {
+        None
+    }
+
     /// Whether [`Self::try_resolve_many_flat`] with `nrhs` columns returns
     /// **bit-identical** results to `nrhs` separate [`Self::resolve`] calls.
     ///
--- a/crates/pounce-algorithm/src/kkt/low_rank_aug_system_solver.rs
+++ b/crates/pounce-algorithm/src/kkt/low_rank_aug_system_solver.rs
@@ -595,10 +595,19 @@ impl LowRankAugSystemSolver {
         // `W` on a KKT this size, so the split throws away one `F` per
         // SMW update. The same bit-identity gate as the warm batch
         // below applies, at the wider `n_cols_us`.
+        //
+        // Note there is deliberately no `dim == inner.system_dim()`
+        // precondition here, unlike the warm batch below. Cold, the
+        // inner solver has not assembled yet, so its `system_dim()` is
+        // still 0 and that check would reject every first factorization
+        // — silently, leaving the merged path unexercised by any mock
+        // whose `system_dim()` is 0 when cold. `try_solve_many_flat`
+        // assembles first and then declines if the packed length
+        // disagrees, which is the same guard applied where the answer
+        // is actually known.
         if !self.inner_has_factor
             && n_cols_us > 1
             && dim > 0
-            && dim == self.inner.system_dim() as usize
             && self.inner.multi_solve_matches_single_solve(n_cols_us)
         {
             let mut packed = vec![0.0; dim * n_cols_us];
@@ -1296,6 +1305,11 @@ mod tests {
         backsolves: Cell<usize>,
         batched_calls: Cell<usize>,
         batched_cols: Cell<usize>,
+        /// Calls to the *factorizing* multi-RHS path, and the columns
+        /// they carried. Separate from `batched_calls` so a test can
+        /// tell "one merged call" from "a factorization plus a batch".
+        factor_batched_calls: Cell<usize>,
+        factor_batched_cols: Cell<usize>,
         factored_diag: RefCell<Vec<Number>>,
         factored_delta_x: Cell<Number>,
     }
@@ -1417,6 +1431,53 @@ mod tests {
         /// shows up as a wrong answer rather than a wrong count.
         fn multi_solve_matches_single_solve(&self, _nrhs: usize) -> bool {
             self.packed
+        }
+
+        /// The factorizing counterpart: captures the factor like
+        /// `solve` does, then applies it to every packed column. One
+        /// call, one factorization, all columns — which is the whole
+        /// point of the path.
+        fn try_solve_many_flat(
+            &mut self,
+            coeffs: &AugSysCoeffs<'_>,
+            packed_rhs: &mut [Number],
+            nrhs: usize,
+            _check_neg_evals: bool,
+            _num_neg_evals: Index,
+        ) -> Option<ESymSolverStatus> {
+            if !self.packed {
+                return None;
+            }
+            let wdiag = coeffs
+                .w
+                .expect("CountingInner requires W")
+                .as_any()
+                .downcast_ref::<DiagMatrix>()
+                .expect("CountingInner requires W to be a DiagMatrix");
+            let diag_rc = wdiag.get_diag().expect("Wdiag has no diag set").clone();
+            let diag = downcast_dense(diag_rc.as_ref()).expanded_values();
+            let dim = diag.len();
+            if packed_rhs.len() != dim * nrhs {
+                return None;
+            }
+            *self.stats.factored_diag.borrow_mut() = diag.clone();
+            self.stats.factored_delta_x.set(coeffs.delta_x);
+            self.stats
+                .factorizations
+                .set(self.stats.factorizations.get() + 1);
+            self.stats
+                .factor_batched_calls
+                .set(self.stats.factor_batched_calls.get() + 1);
+            self.stats
+                .factor_batched_cols
+                .set(self.stats.factor_batched_cols.get() + nrhs);
+            let delta_x = coeffs.delta_x;
+            for col in packed_rhs.chunks_mut(dim) {
+                for (i, x) in col.iter_mut().enumerate() {
+                    *x /= diag[i] + delta_x;
+                }
+            }
+            Some(ESymSolverStatus::Success)
         }
 
         fn try_resolve_many_flat(
@@ -1969,13 +2030,10 @@ mod tests {
             "a mock without the packed path must take the per-column arm"
         );
         assert!(
-            stats_batch.batched_calls.get() >= 1,
-            "the packed mock must actually reach the batched arm"
-        );
-        assert_eq!(
-            stats_batch.batched_cols.get(),
-            2,
-            "3 V columns: column 0 pays the factorization, 2 are batched"
+            stats_batch.factor_batched_calls.get() >= 1,
+            "the packed mock must actually reach the merged arm — a mock \
+             whose `system_dim()` is 0 when cold silently would not, and \
+             then this test would be green about code it never ran"
         );
         assert_eq!(
             stats_batch.factorizations.get(),
@@ -1985,6 +2043,38 @@ mod tests {
         assert_eq!(
             sol_loop, sol_batch,
             "batched and per-column arms must agree bit-for-bit"
+        );
+
+        // All three V columns ride along with the factorization in a
+        // single backend call. Before the merge this read
+        // `factor_batched_calls == 0`, `batched_cols == 2` — column 0
+        // through the single-RHS `solve`, then a second call carrying
+        // the other two. That second call streams the whole factor
+        // again, which is the cost this removes.
+        assert_eq!(
+            stats_batch.factor_batched_calls.get(),
+            1,
+            "the cold SMW block must be ONE factorizing multi-RHS call"
+        );
+        assert_eq!(
+            stats_batch.factor_batched_cols.get(),
+            3,
+            "all 3 V columns must ride along with the factorization"
+        );
+        assert_eq!(
+            stats_batch.batched_calls.get(),
+            0,
+            "no separate back-solve pass may remain over the same factor"
+        );
+        // One `resolve` remains, and must: it is the actual RHS being
+        // solved against the same factor after the V columns have built
+        // the SMW correction. Upstream pays it too. What the merge
+        // removes is the second pass over the factor that used to carry
+        // the V columns.
+        assert_eq!(
+            stats_batch.backsolves.get(),
+            1,
+            "only the main RHS solve may remain over the cached factor"
         );
     }
 
--- a/crates/pounce-algorithm/src/kkt/low_rank_aug_system_solver.rs
+++ b/crates/pounce-algorithm/src/kkt/low_rank_aug_system_solver.rs
@@ -557,23 +557,100 @@ impl LowRankAugSystemSolver {
 
         // Every column here shares one matrix, so only the first one
         // needs a factorization; the rest are back-substitutions
-        // against it. Upstream gets that from a single `MultiSolve`
-        // over all `nrhs` columns (`IpLowRankAugSystemSolver.cpp:487`),
-        // which reaches `StdAugSystemSolver::MultiSolve` and factorizes
-        // once. We reproduce it in two steps: pay the factorization on
-        // column 0 through the single-RHS path when the inner solver is
-        // cold, then hand every remaining column to the inner solver's
-        // packed multi-RHS back-substitution in one call. That is the
-        // only way `nrhs > 1` reaches the backend — both FERAL
-        // (`solve_many_into`) and MA57 (`ma57cd_` with `nrhs`) block the
-        // triangular solves, and neither can do so one column at a time
-        // (gh#729).
+        // against it. Batching is the only way `nrhs > 1` reaches the
+        // backend at all — both FERAL (`solve_many_into`) and MA57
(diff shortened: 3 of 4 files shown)


srikanth-gm:
# The MA57 measurement

Thanks for the detailed writeup, and for being explicit about what the draft could and could
not establish — that made it easy to know exactly which run would be useful. We have coinhsl
here, so I was glad to do the A/B you asked for on the reporting model.

One thing surfaced that affects how the numbers should be read: as the branch stands, the
merged path cannot execute on MA57 at all. Rather than leave it at a null, I measured three
things — the shipping comparison, a variant that makes the path reachable so your actual
question can be answered, and an Ipopt/MA57 reference in the same block.

Model: 59,939 variables, 58,271 equality + 33 inequality constraints,
`hessian_approximation=limited-memory`, `limited_memory_max_history 6`, `linear_solver=ma57`,
KKT dimension 118,276.

**Summary.** As shipped the branch is a no-op on MA57: bit-identical trajectory, all timing
deltas inside noise, and an instrumented build takes the merged path zero times. With the
gate opened it does what it claims and by almost exactly the FERAL proportions
(back-solve −32.0% vs your −30.8%), so MA57 is not the larger win the draft hoped for. But
opening it moves POUNCE's back-solve row from **+51% to −19% against Ipopt/MA57**, and
linear-algebra total from +27.5% to +8.0%, which is a bigger structural change than the −4.2%
internal net suggests. Three defects are in the way; they are in §7.

---

## 1. The merged path is unreachable on MA57

The merged call is gated on the backend affirming bit-identity:

```rust
if !self.inner_has_factor
    && n_cols_us > 1
    && dim > 0
    && self.inner.multi_solve_matches_single_solve(n_cols_us)
```

The PR touches `aug_system_solver.rs`, `low_rank_aug_system_solver.rs`,
`std_aug_system_solver.rs` and the changelog. It does not touch `pounce-hsl`, and at
92700b67 `Ma57SolverInterface` has no `multi_solve_matches_single_solve` override, so it
inherits the trait default `false` from `sparse_sym_iface.rs:143`. FERAL overrides it
(`nrhs <= 16`, `pounce-feral/src/lib.rs:1308`); MA57 never has, and there is no option to
make it.

I added a `tracing::debug!` at the gate and at the point the merged call is issued, built
92700b67, and ran the model:

```
count  observation
    2  n_cols=2  cold=true   dim=118276  affirms=false     <- the case this PR adds
    2  n_cols=2  cold=false  dim=118276  affirms=false
    1  n_cols=1  cold=true   dim=118276  affirms=false
    1  n_cols=1  cold=false  dim=118276  affirms=false

merged calls taken: 0
```

The same instrumentation with MA57 affirming reports `merged calls taken: 2`, on precisely
the two `cold=true, n_cols=2` evaluations, so the probe is not blind.

This also means the pre-existing warm batch below it never ran on MA57 either. On this
backend the SMW columns have always been solved one column at a time — which is why the
back-solve row you were shown is as large as it is.

---

## 2. Method

Five binaries, two configs, only `pounce_cinterface.dll` changing between POUNCE arms. All
built on one host with one toolchain (rustc 1.92.0) and identical flags
(`cargo build --release -p pounce-cinterface --features pounce-algorithm/ma57`; workspace
profile `lto="thin"`, `codegen-units=1`). I built the 4c9abf76 baseline locally rather than
using the shipped wheel, because the wheel's DLL uses different codegen settings and
comparing against it would measure the build.

| arm | binary |
|---|---|
| `ipopt` | Ipopt/MA57, same model and tolerances |
| `base` | 4c9abf76, unmodified — the build behind the numbers you quoted |
| `pr809` | 92700b67, unmodified |
| `base_g` | 4c9abf76 + three-line MA57 affirmation |
| `pr809_g` | 92700b67 + the same three-line affirmation |

The affirmation is only:

```rust
fn multi_solve_matches_single_solve(&self, _nrhs: usize) -> bool { true }
```

`base_g`/`pr809_g` differ only by the PR, so that pair is what answers your question.
Two interleaved replicates per arm; arms run adjacent in time. Every arm was deterministic:
both replicates produced identical trajectories and identical objectives to the last digit.
The Ipopt reference ran in a second block with `base` repeated as a drift anchor — it
reproduced to **+0.7%** on solver-internal and +0.6% on wall, so the two blocks read
together.

---

## 3. As shipped: no effect

Trajectory **bit-identical** to 4c9abf76 over all 207 printed rows; objective identical to
every digit (`7.4688930868538648e+01`).

| per iteration (s) | `base` | `pr809` | Δ |
|---|---:|---:|---:|
| numeric factorization | 0.1462 | 0.1490 | +1.9% |
| back-solve | 0.1580 | 0.1629 | +3.1% |
| linear algebra total | 0.3121 | 0.3202 | +2.6% |
| solver internal | 0.4056 | 0.4159 | +2.5% |

Within-arm replicate spread on `pr809` is 4.1–8.1%, so all of that is noise. This is the null
the instrumentation predicts.

---

## 4. With the gate opened: the number you asked for

| per iteration (s) | `base_g` | `pr809_g` | Δ |
|---|---:|---:|---:|
| numeric factorization | 0.1443 | 0.1711 | **+18.5%** |
| back-solve | 0.1256 | 0.0854 | **−32.0%** |
| linear algebra total | 0.2796 | 0.2678 | −4.2% |
| solver internal | 0.3755 | 0.3544 | −5.6% |

Against your FERAL figures:

| | FERAL (`laptime`) | MA57 (this model) |
|---|---:|---:|
| back-solve | −30.8% | −32.0% |
| factorization | +16.4% | +18.5% |
| linear algebra net | −1.6% | −4.2% |

**MA57's `F/W` ratio is not materially different from FERAL's.** The back-solve row moves by
almost the same fraction, the factorization row absorbs almost the same amount, and the net
is small on both. Your reading of your own result — real work removed, not much of it, most
of the back-solve drop being bookkeeping — transfers to MA57 essentially unchanged. If the
case for merging rested on MA57 having a much larger fixed traversal cost relative to the
per-RHS cost, this does not support it.

The −4.2% net also sits against per-arm spreads of 3.4–5.3%, so I would not defend it as more
than "probably real, small".

---

## 5. Against Ipopt/MA57

Same block, adjacent runs. This is where the change looks more interesting than §4 does.

| per iteration (s) | Ipopt/MA57 | POUNCE `base` | POUNCE `pr809_g` |
|---|---:|---:|---:|
| numeric factorization | 0.1344 | 0.1467 (+9.1%) | 0.1702 (+26.6%) |
| back-solve | 0.1055 | 0.1594 (**+51.0%**) | 0.0856 (**−18.9%**) |
| symbolic factorization | 0.0042 | 0.0081 | 0.0104 |
| linear algebra total | 0.2465 | 0.3142 (+27.5%) | 0.2662 (**+8.0%**) |
| non-linear-algebra | 0.0368 | 0.0943 (+156%) | 0.0871 (+137%) |
| solver internal | 0.2833 | 0.4085 (+44.2%) | 0.3533 (+24.7%) |

Two things worth pulling out.

**The back-solve row inverts.** The +41% @srikanth-gm reported (+51% in my block, same
direction) becomes −18.9%: with the merged call, POUNCE does the SMW columns in fewer
traversals of the factor than Ipopt does. Linear-algebra total goes from +27.5% to +8.0%,
and solver-internal from +44.2% to +24.7%. On an equal-iteration basis that is
0.4085 → 0.3533 s/iter, i.e. **11.4 s of solver-internal removed over a 206-iteration run.**

**What remains is not linear algebra.** Of the residual 0.0700 s/iter gap against Ipopt in
`pr809_g`, the non-linear-algebra row is 0.0503 s/iter — **72% of it**, and still +137%
in relative terms. Your own observation on the draft ("57% of the gap sits in neither
linear-algebra row") is confirmed and, after this change, understated: with the back-solve
row now negative, non-LA is essentially the entire remaining gap. That is where the next
work is, and this PR correctly does not claim to touch it.

Symbolic factorization is ~2x Ipopt in absolute terms too (1.66 s vs 0.84 s per run), but
that is ~0.6% of wall and not worth chasing. Note the per-iteration symbolic figures above
are misleading: total symbolic time is constant at ~1.67 s regardless of arm, so the
apparent rise from 0.0081 to 0.0104 is only the shorter iteration count. It is not a
regression.

---

## 6. Why the gate cannot simply be opened

`base` and `base_g` are **not** bit-identical. They diverge at iteration 20, in the last
printed digit of the objective:

```
   base    20  9.7709874e+01 3.29e+00 5.87e+01  -1.5 3.71e+00  - 2.75e-01 2.21e-01f 1
   base_g  20  9.7709873e+01 3.29e+00 5.87e+01  -1.5 3.71e+00  - 2.75e-01 2.21e-01f 1
```

`ma57cd_` with `nrhs > 1` does not reproduce the per-column result bit-for-bit. That is about
one ulp, but it is enough: the runs end at different iteration counts (201 / 206 / 173 / 160)
and different objectives. The gh#729 default of `false` is doing real work for MA57.

**Two consequences, and the first is a trap.**

The wall-clock figures across gate-closed and gate-open arms — Ipopt 326.0 s, `base` 356.5 s,
`pr809_g` 269.2 s — **must not be quoted.** `pr809_g` converging in 160 iterations against
`base`'s 206 is a one-ulp perturbation landing favourably on a chaotic trajectory, not work
removed. It could as easily have gone the other way. Only the per-iteration figures in §4
and §5 are defensible, and the §5 equal-iteration projection is the honest way to state the
benefit.

Second, because the trajectories differ, `base_g` vs `pr809_g` is not single-variable at the
trajectory level the way your FERAL A/B was. Both arms carry the same affirmation and differ
only by the PR, but they walk different paths. I believe the §4 numbers — they reproduce your
FERAL ratios closely and the mechanism is unambiguous — but they are not the clean
bit-identical comparison you were able to make.

---

## 7. Bugs

**(a) MA57 has no way to reach any batched path. Blocks this PR.**
No `multi_solve_matches_single_solve` override and no option to supply one, so both the
merged path this PR adds and the pre-existing warm batch are dead on MA57. Worth noting that
the code comment in this very PR reads

> both FERAL (`solve_many_into`) and MA57 (`ma57cd_` with `nrhs`) block the triangular
> solves, and neither can do so one column at a time

which describes a capability MA57 has but never exercises. Given §6, the fix is an opt-in
option rather than a default `true`, and it needs (b) fixed first.

**(b) Every `ma57_*` option is silently discarded. Independent of this PR.**
All production constructions of the backend are `Ma57SolverInterface::new()`, which uses
`Options::defaults()`:

- `crates/pounce-algorithm/src/application.rs:4675`
- `crates/pounce-algorithm/src/application.rs:4710`
- `crates/pounce-cli/src/main.rs:3468`

`Ma57Options::from_options_list` exists but is called nowhere outside its own tests, and
`SparseSymLinearSolverInterface` has no options entry point — FERAL is configured through
`default_backend_factory(feral_cfg)`, and MA57 has no equivalent. So `ma57_pivtol`,
`ma57_pivot_order`, `ma57_pre_alloc`, `ma57_block_size`, `ma57_node_amalgamation`,
`ma57_small_pivot_flag` and `ma57_automatic_scaling` are registered, documented, accepted
without warning, and dropped.

Confirmed empirically: `ma57_pivtol` at its default `1e-8` versus `0.5` gives a bit-identical
20-iteration trajectory. A pivot tolerance spanning that range cannot leave the factorization
unchanged if it is reaching MA57.

This is not academic for the numbers above. The POUNCE config sets `ma57_pivot_order: 5` and
`ma57_pre_alloc: 2.0` and gets neither; the Ipopt config sets `ma57_pre_alloc: 2.0` and does
get it. **The two solvers in §5 are not running MA57 with the same settings**, and the
difference is not in POUNCE's favour. I will file this separately.

**(c) MA57 batched multi-solve is not bit-identical to the per-column path.**
~1 ulp, §6. Not a defect in this PR, but it is the thing that decides how (a) can be landed,
and it should be recorded somewhere rather than rediscovered.

---

## 8. Recommendation

I would not merge this on the strength of the MA57 timing, for the same reason you did not
merge it on the FERAL timing: on a like-for-like per-iteration basis the net is −4.2%,
against a 3–5% noise floor.

But I would not close it either, and I would not read §4 as the whole story. Against Ipopt
the change is the difference between a back-solve row that is 51% worse and one that is 19%
better, and it takes solver-internal from +44% to +25%. That is worth having. The blocker is
(a) and (b), not this PR — and with those fixed, this PR is what makes the batch worth
turning on rather than merely possible.

The structural argument — matching upstream's single `MultiSolve` and removing a
demonstrably redundant traversal of the factor — is unaffected by any of the above and
stands on its own.

jkitchin:
# Status update — one of the two blockers is gone, the other is not

Re-checked against `main` @ `d194d9ec`. Short version: **still correctly a draft.**
@srikanth-gm's blocker (b) is fixed; blocker (a), the one that actually holds this PR,
is untouched.

## What has landed since the MA57 measurement

**(b) — every `ma57_*` option silently discarded — is fixed.** #827 (gh#825, merged
2026-08-28) threads them through. All three production construction sites named in §7(b)
now go via `Ma57SolverInterface::from_options_list` rather than `::new()`; the only
remaining mentions of `::new()` at those line numbers are doc comments describing the old
behaviour. It carries two guard tests, `ma57_options_reach_the_backend.rs` and
`no_production_site_builds_ma57_with_defaults.rs`, the second of which is specifically
there to stop the defect returning. A follow-up in the same PR also refuses an explicit
`ma57_pivtolmax` below `ma57_pivtol`.

**Adjacent, and relevant to reproducing your run:** #834 (merged today) extends gh#811's
rpath re-emission from the `pounce` binary to the five further crates that link coinhsl —
`pounce-cinterface`, `pounce-py`, `pounce-restoration`, `pounce-rs`, `pounce-sensitivity`.
Before it, their test binaries aborted at process start under any MA57 build, and
`pounce._pounce` failed at `import pounce`. It also adds `make test-ma57`, which runs the
workspace with the backend actually linked. None of that changes the timing question, but
it does mean the MA57 side of this is now exercisable in-tree by anyone with an install
rather than only in your environment.

## What has not landed

**(a) — MA57 cannot reach any batched path — is still live.** Verified on `main` today:
`impl SparseSymLinearSolverInterface for Ma57SolverInterface` overrides eight methods
(`as_any`, `initialize_structure`, `values_array_mut`, `multi_solve`,
`number_of_neg_evals`, `increase_quality`, `provides_inertia`, `matrix_format`) and
`multi_solve_matches_single_solve` is not among them, so it still inherits the trait
default `false` from `sparse_sym_iface.rs:143`. There is no registered option to supply
one. Both the merged path this PR adds and the pre-existing warm batch below it remain
dead on MA57, exactly as your instrumented run reported.

**(c) — `ma57cd_` with `nrhs > 1` is not bit-identical to the per-column path — is still
unrecorded.** Nothing in `crates/pounce-hsl/` or `dev-notes/` mentions it. Per your §7(c)
this should be written down somewhere before it is rediscovered, and it is the fact that
decides (a) has to be an opt-in rather than a default `true`.

## This PR itself

Unchanged and unsuperseded. The branch head is still `92700b67` — the exact commit you
measured — and is now 104 commits behind `main`. Of the four files it touches, the three
under `crates/pounce-algorithm/src/kkt/` have not moved on `main` since the merge-base
(`7eb0b201`); the only one that has is `CHANGELOG.md`. So nothing merged in the interim
has done this work by another route, and the change still applies: `git merge-tree` against
`main` produces a clean tree, though the GitHub UI is currently showing the PR as
conflicting and will want a rebase for the changelog in practice.

## Where that leaves it

The order @srikanth-gm set out in §8 still holds, and only its first step is done:

1. ~~(b) `ma57_*` options reach the backend~~ — done, #827.
2. **(a) an opt-in for MA57 batched multi-solve** — not started. Opt-in rather than
   default `true`, because of (c) and §6: the one-ulp difference moves the trajectory, and
   the gate-open/gate-closed wall-clock figures (Ipopt 326.0 s, `base` 356.5 s, `pr809_g`
   269.2 s) must not be read as work removed.
3. **(c) recorded** — not started.
4. Then this PR, which is what makes the batch worth turning on rather than merely
   possible: on the §5 equal-iteration basis it takes the back-solve row from +51% to
   −18.9% against Ipopt/MA57 and linear-algebra total from +27.5% to +8.0%.

I am not merging it on the timing, for the reason both @srikanth-gm and the draft already
gave: −4.2% per-iteration net against a 3–5% noise floor is not a merge case on its own.
The structural argument — matching upstream's single `MultiSolve` and removing a
demonstrably redundant traversal of the factor — is what it would be merged on, and that
is unaffected by any of the above.

jkitchin:
Ready to leave draft. The two prerequisites this PR was blocked behind are both on `main` now.

## What unblocked it

| finding | disposition |
|---|---|
| **(b)** `ma57_*` options never reached the backend | fixed, #827 |
| **(a)** MA57 had no way to answer the bit-identity gate | fixed, #836 (`ma57_batched_backsolve`) |
| **(c)** the batch is a trajectory change, not a free speed-up | recorded, #836 (`dev-notes/ma57-batched-backsolve.md`) |

(a) is the one that mattered to this PR. `try_solve_many_flat` is taken only when the backend affirms `multi_solve_matches_single_solve` at the full column count — and MA57 had no override at all, so it inherited the trait default `false` and **declined this PR's merged path unconditionally**. The merged call was reachable on FERAL and nowhere else. It is now reachable on MA57 too, for a user who opts in.

That does not change what this PR measures. It changes whether the mechanism is available on the backend @srikanth-gm's report was about.

## Rebased

Rebased onto `main` (`79cc45ce`, 106 commits). One conflict, in `low_rank_aug_system_solver.rs`: #836 had appended a note to the same comment block this PR rewrites. Resolved by keeping this PR's three-path list and folding #836's point in underneath it — who answers the gate (FERAL a measured ceiling, MA57 an opt-in permission) rather than only where it is asked.

The CHANGELOG entry auto-merged into `[Unreleased]` at the position its context anchor put it; moved to the top so it sits next to #836's, which it now depends on.

Verified after the rebase: `cargo fmt --check` clean; `cargo test -p pounce-algorithm --lib kkt::` 25/25 pass, including `batched_smw_columns_match_the_per_column_path`, the test the second commit rewrote to actually execute the merged path.

## What still stands from the original measurement

Unchanged by the rebase, and worth restating because the first number posted here was wrong:

- Net **−1.45 s of 89 s** of linear algebra (−1.6%), −0.7% overall, against a **4.2% baseline spread** — inside noise on wall time, and it should not be quoted as a wall-clock win. Most of the movement between the back-solve (−31%) and factorization (+16%) rows is attribution: the merged call bills its substitutions to the factorization row.
- It is, however, the size the mechanism predicts (~20 ms per removed factor traversal over ~75–90 SMW updates), so it is real work removed.
- Trajectory **bit-identical** over 100 iterations in both A/B pairs; fixture sweep empty over 152 fixture-legs on both `exact` and `lbfgs`.
- The earlier −7.8% was measured on a host that has since changed and is withdrawn.

## Not measured

Whether the merge is worth more under MA57, where the fixed-versus-per-RHS cost ratio differs. No CoinHSL here. #836 makes the path *reachable* on MA57; nobody has measured it there, and that is stated in both PRs rather than assumed away.

Issue #810, opened 2026-08-27
MA57 declines the gh#729 bit-identity gate, so the batched low-rank SMW path is FERAL-only (measured: ~10% of OverallAlgorithm, at the cost of a trajectory change)


jkitchin:
**Components:** `crates/pounce-hsl`, `crates/pounce-linsol/src/sparse_sym_iface.rs`, `crates/pounce-algorithm/src/kkt/low_rank_aug_system_solver.rs`
**Related:** gh#729 (closed), PR #809
**Configuration:** `linear_solver=ma57`, `hessian_approximation=limited-memory`

> **AI disclosure.** The measurements, instrumented builds and this text were produced with the help of an AI coding assistant.

## Summary

`pounce-hsl` does not override `SparseSymLinearSolverInterface::multi_solve_matches_single_solve`, so MA57 inherits the `false` default at `crates/pounce-linsol/src/sparse_sym_iface.rs:143`. Every batching path gated on that predicate is therefore dead code under MA57 — including the merged low-rank SMW arm added in PR #809, which is gated at `low_rank_aug_system_solver.rs:601`. `pounce-feral` does override it (`nrhs <= FERAL_BITWISE_MULTI_SOLVE_MAX_NRHS`, currently 16), so the feature is FERAL-only in practice.

The prize is measured and large — roughly the entire remaining gap to Ipopt on our large limited-memory model — but it is **not free**: turning the gate on changes the trajectory, which is exactly the failure gh#729 introduced the gate to prevent. This issue exists so that trade-off has an owner rather than sitting implicit in a default.

## Evidence

Confirmed three independent ways.

1. **Source.** No `multi_solve_matches_single_solve` impl anywhere in `crates/pounce-hsl`; the trait default is `false`.
2. **Runtime probe.** A build printing the predicate at each call site reports `gate=false` at every width under `linear_solver=ma57`, and `gate=true` under `linear_solver=feral`.
3. **Measurement.** MA57 before and after PR #809 is identical to the printed digit on every timing row.

FFI counters on the MA57 entry points confirm what the models actually issue on `benchmarks/large_scale/nl/laptime.nl` (58 014 x 62 014, `hessian_approximation=limited-memory`, `mu_strategy=monotone`, `max_iter=100`):

    ma57ad (analyse) = 2    ma57bd (factor) = 110    ma57cd (solve) = 1351    total RHS columns = 1351

[...]

PR #813, opened 2026-08-27
fix(hsl): MA57 declines the gh#729 batching gate explicitly, with the measurement

jkitchin:
Addresses the diagnostic half of #810 and settles one of its three options with a measurement. Refs #810 — deliberately **not** a closing keyword; see "What this leaves open".

## What was wrong

`Ma57SolverInterface` inherited `SparseSymLinearSolverInterface::multi_solve_matches_single_solve`'s `false` default. The value is correct. Inheriting it is not, because a bare default is indistinguishable from an omission nobody got to yet — and the merged low-rank SMW arm (gh#729, PR #809) is gated on exactly this predicate, so under `linear_solver=ma57` that whole mechanism is dead code. That went unnoticed long enough to reach a changelog claiming the opposite.

## What this does

Writes the predicate out on the MA57 impl with the numbers that justify it, and adds `multi_solve_reassociates_from_two_columns_up` as the guard — the counterpart to feral's `multi_solve_bitwise_matches_single_solve_at_the_documented_ceiling`, pointing the other way.

Against **one shared factor** — so a difference cannot be blamed on a nondeterministic analyse/factor, and the test asserts a repeated single-RHS solve reproduces itself before it compares anything — `ma57cd` at `nrhs = 2` already disagrees with two `nrhs = 1` calls:

| nrhs | entries differing bitwise | max relative difference |
|---:|---:|---:|
| 2 | 648 / 1152 | 1.3e-14 |
| 3 | 1221 / 1728 | 1.3e-11 |
| 7 | 2964 / 4032 | 6.5e-11 |
| 14 | 6579 / 8064 | 1.1e-10 |

Tolerance-legal and trajectory-moving is precisely the dangerous combination gh#729 named. The test asserts both halves — that the answers differ, *and* that they differ only slightly, since a large difference would mean a broken solve rather than reassociation.

## The useful negative result

**There is no safe narrow window for MA57.** feral is bit-identical below its private BLAS-3 threshold, so it can honestly return `true` for `nrhs <= 16`; an `nrhs`-capped MA57 override was therefore the obvious repair for #810, and it would have recovered the ~10% with no trajectory change. It is empirically dead: MA57 reassociates from the second column on, well below its own `ICNTL(13) = 10` level-3 BLAS threshold, and nothing observable changes as that threshold is crossed. The test brackets widths 2/9/10/16 to record that.

Also recorded because it is the natural thing to assume and it is wrong here: feral's guard needs a wide separator or its blocked panel degenerates to the same scalar operations as the rank-1 cascade and agrees bit-for-bit at every width. MA57 has no such degenerate case — swapping the 2-D Laplacian for a plain tridiagonal band still shows reassociation at every width. The Laplacian is kept for comparability with feral's guard, not because this test depends on it.

## Verification

Behaviour is unchanged. The override returns the same constant the default returned, so no code path moves — **no fixture sweep applies**, and this is not a trajectory change.

- `COINHSL_DIR=... cargo test -p pounce-hsl` — 15 + 3 pass, including the new guard.
- `cargo test -p pounce-linsol` — 24 + 1 pass (the only change there is a doc comment).
- `cargo check -p pounce-hsl --all-targets` with `COINHSL_DIR` **unset** — clean, i.e. the CI step that actually covers this crate still passes.
- `cargo clippy -p pounce-hsl --all-targets -D clippy::correctness -D clippy::suspicious` — clean.
- `cargo doc` — one pre-existing warning in `pounce-linsol` (`Self::min_row_value`), untouched by this branch.

Both new assertions are mutation-checked, and the test carries the table:

| mutation | result |
|---|---|
| `multi_solve_matches_single_solve` returns `true` | fails at `nrhs=1`, "must decline the gh#729 batching gate" |
| compare the looped arm against itself instead of the batched one | fails at `nrhs=2`, "now bit-identical to looping single-RHS" |
| replace the Laplacian with a tridiagonal band | **does not fail** — recorded in the test, see above |

## Caveat the test states about itself

`ci.yml` excludes `pounce-hsl` from build/test/clippy and only `cargo check`s it, so this guard runs under a local `COINHSL_DIR=... cargo test -p pounce-hsl` and nowhere else. That is the pre-existing situation for every test in the crate, not something this PR introduces, but it means the guard is a local one and should not be mistaken for CI coverage.

## What this leaves open

The speed is **not** recovered. Batching is worth roughly 10% of `OverallAlgorithm` under MA57 on a large limited-memory model (back-solve −43.4%, measured with the gate forced on as a diagnostic), and with the gate forced on the trajectory does move — objective 78.21 split against 65.30 merged, same model, same iteration cap. Recovering it means accepting that deliberately behind an opt-in, with a full `scripts/sweep-fixtures.sh` diff over both legs and a benchmark pass over `benchmarks/qp` and the Maros-Meszaros families. #810 stays open for that decision; this PR just removes the option that looked free and makes the remaining choice legible.


jkitchin:
Closing unmerged.

This PR changed no behaviour — the explicit `multi_solve_matches_single_solve` returns the same constant the trait default returned. What it added was a doc comment, a CHANGELOG entry, and a test.

The test was the part meant to carry weight, and it cannot: `ci.yml` excludes `pounce-hsl` from build/test/clippy and only `cargo check`s it, because executing it needs a licensed `libcoinhsl` CI does not have. So it compiles but never runs, which makes it documentation with executable syntax rather than a guard — and the PR body oversold it as feral's counterpart, which does run.

The measurement it existed to record is preserved in https://github.com/jkitchin/pounce/issues/810#issuecomment-5445108573, together with a reproducer. #810 stays open for the option-1-or-3 decision.

PR #817, opened 2026-08-28
fix(815): a solved square problem is reported as solved — band, Python surfaces, and the restoration iteration count

### LINKED ISSUE #815: Local infeasibility reported on a feasible square IDAES flowsheet that Ipopt solves
> [!IMPORTANT]
> **Corrected in [this comment](https://github.com/jkitchin/pounce/issues/815#issuecomment-5451845993).**
> The comparison below is not apples-to-apples: Ipopt was run through
> `writer_config={"linear_presolve": True, "scale_model": True}` while pounce got the
> raw model. The headline finding survives on a corrected, like-for-like comparison
> (`ipopt_v2` converges in 4.5 s, `pounce_v2` fails, both presolved and scaled), but
> the **"Options tried" table and the scaling hypothesis below are superseded** —
> pounce fails even when handed the presolved, scaled problem. Read the comment first.

---

## Summary

On a 536×536 **square** IDAES flowsheet (zero degrees of freedom, trivial `f(x)=0`
objective), pounce exits `Converged to a point of local infeasibility` after
**2–3 iterations**. Ipopt 3.13.2 (the IDAES build) converges on the identical
model from the identical starting point in 7.9 s.

The problem is genuinely feasible, and pounce itself can confirm that: if the
model is first converged with Ipopt and then handed to pounce, pounce terminates
`optimal` in 5.6 s, and its sensitivity backsolves agree with an independent
reduced-Jacobian calculation (PyNumero + ASL Jacobian + sparse solve) to 0.06%.
So the `.nl` is fine and the solution is reachable — pounce cannot get there from
the initialization.

**Model:** an IDAES flowsheet of a naphtha hydrotreater — 5 unit models
(feed/effluent heat exchanger, trim heater, adiabatic CSTR with HDS kinetics,
trim cooler, flash) with a genuine recycle arc, on a Generic property package
with PR-style VLE. 536 equations, 536 unknowns after the specs are fixed. Nothing
exotic; this is the ordinary shape of an IDAES flowsheet.

## Reproducer

```python
m = build_model(); specify_model(m); initialize_model(m)
m._obj = pyo.Objective(expr=0.0)          # square problem needs an objective
pyo.SolverFactory("pounce").solve(m, tee=True)
```

```
### Ipopt ###
CONVERGED in 7.9 s

### pounce 0.10.0 ###
Total number of variables...............: 536
Total number of equality constraints....: 536
Total number of inequality constraints..:   0
   0  0.0000000e+00 5.02e+03 1.00e+00    0.0 0.00e+00      - 0.00e+00 0.00e+00   0
   1  0.0000000e+00 5.02e+03 1.02e+05    3.0 5.88e+06      - 7.26e-01 2.13e-05f  1
   2  0.0000000e+00 4.97e+03 5.09e+05    5.3 3.37e+07      - 1.90e-02 3.49e-03f  1
Number of Iterations....: 2
```

The second-opinion ladder fires and does not recover:

```
pounce: local infeasibility — re-solving along 2 different trajectories before believing it
pounce: feral_scaling=mc64 re-solve did not recover (InfeasibleProblemDetected).
pounce: mu_strategy=adaptive re-solve did not recover (InfeasibleProblemDetected).
pounce: keeping the original local-infeasibility verdict
```

Total wall clock 61.2 s, against Ipopt's 7.9 s.

## What the log suggests

Iteration 0 starts at a constraint violation of 5.02e+03. Accepted steps are tiny
(`2.13e-05f`, `3.49e-03f`, `6.14e-05h`) while the regularization / `du` column is
very large (`5.88e+06`, `3.37e+07`), and pounce concludes after 2–3 iterations.

This has the shape of a scaling problem. IDAES models routinely mix pressures of
~3e6 Pa, molar flows of ~30 mol/s, and trace mole fractions of ~1e-6 in the same
system, and Ipopt's default `gradient-based` NLP scaling is a large part of what
makes them tractable.

## Options tried — all still infeasible

| option | result | time |
|---|---|---|
| defaults | InfeasibleProblemDetected | 60 s |
| `nlp_scaling_method=gradient-based` | InfeasibleProblemDetected | 65 s |
| `bound_relax_factor=0` | InfeasibleProblemDetected | 71 s |
| `mu_init=1e-6` | InfeasibleProblemDetected | 35 s |

## A fairness caveat

Ipopt does not solve this in a single call either. The flowsheet's `solve_model()`
deliberately re-issues `solve()` up to four times, because each call resets the
barrier parameter and restoration state and effectively warm-restarts; a single
Ipopt call with a larger `max_iter` does *not* substitute for that (it wanders
rather than converging).

So part of the gap may simply be that pounce gives up where Ipopt-with-restarts
persists. But pounce stops after 2–3 iterations, far short of any iteration
limit, which is why the infeasibility verdict looks premature rather than
budget-limited.

## Why this matters

This is the standard shape of an IDAES flowsheet, so it likely affects any attempt
to use pounce as a drop-in Ipopt replacement in that ecosystem. It is also a
gating problem for the sensitivity feature specifically: `pyomo_pounce`'s declared
-parameter sensitivity is genuinely nicer than the sIPOPT/k_aug workflow it
replaces (declare, solve, `gradient(...)` as a cheap backsolve — 3.5 ms for four
sensitivities here), but it is only reachable if the solve converges.

**Environment:** pounce 0.10.0 built from `5cf01b16`, macOS arm64 (Apple Silicon),
`pyomo-pounce` from the same tree, Pyomo 6.10.1, IDAES with Ipopt 3.13.2.

### LINKED ISSUE #819: Number of Iterations omits restoration iterations, making a 900-iteration restoration grind read as an early exit
## Summary

`Number of Iterations....:` reports the **outer** iteration count only. Restoration iterations — the `Nr` lines in the same table, immediately above the summary — are not counted anywhere in the output.

On a model that spends its whole solve in restoration, the result is a summary line that is off by three orders of magnitude and reads as an early bail:

```
   0  0.0000000e+00 5.02e+03 1.00e+00   -1.0 0.00e+00      - 0.00e+00 0.00e+00   0
   1  0.0000000e+00 5.00e+03 1.50e+00   -1.0 3.57e+06      - 3.32e-03 5.02e-03h  1
   2  0.0000000e+00 4.97e+03 1.49e+00   -1.0 3.55e+06      - 5.18e-03 5.59e-03h  1
   3  0.0000000e+00 4.97e+03 6.00e+02   -1.0 3.53e+06      - 1.33e-02 6.14e-05h  1
   4r ...
   ... 894 restoration iterations ...
 897r 0.0000000e+00 1.60e+01 1.13e+04    0.5 2.72e-06    8.6 1.00e+00 7.81e-03f  8

Number of Iterations....: 3
Total seconds in POUNCE                              = 11.180
```

Three iterations, eleven seconds. The work is visible in the evaluation counts just below (`Number of Lagrangian Hessian evaluations = 894`), but nothing labelled "iterations" reflects it.

## Why this is worth fixing rather than documenting

This misled three separate readings inside gh #815 alone, two of them mine:

1. The original report characterised the failure as "pounce exits after **2–3 iterations**... far short of any iteration limit, which is why the infeasibility verdict looks premature rather than budget-limited." The verdict was not premature; the solver had ground through ~900 restoration iterations.
2. Having explained that, I then read a second-opinion re-solve reporting `Number of Iterations....: 0` as "a perturbed start declaring local infeasibility without completing a single iteration — possibly a separate bug." It had done 455 restoration iterations.
3. The reporter independently flagged the same `0` as suspicious, in the same comment in which they accepted the outer-count explanation for the first case.

A number that three attentive readings in one thread get wrong is not a documentation problem.

For comparison, Ipopt 3.13.2 on the same `.nl` reports `Number of Iterations....: 2418`, which is its total — the `r` iterations included. So this is also a **parity gap**: the same field means different things in the two solvers, and pounce advertises drop-in compatibility.

## Suggested shape

Not proposing a specific fix, but the constraint is that the existing field's meaning is load-bearing for anything parsing the output. Options roughly in order of disruption:

- Add a separate line (e.g. `Number of restoration iterations.......: 894`) and leave the existing field alone. Safest, and enough on its own to make the log self-explaining.
- Report the total in `Number of Iterations` to match Ipopt, and break out the outer count separately. Better parity, but changes an existing number.
- Surface the restoration count in the JSON report as well — it is currently absent there too, so a programmatic consumer has no way to tell a 3-iteration solve from a 900-iteration one.

Whichever way, the summary block should make it impossible to read an 11-second restoration grind as an early exit.

## Not a duplicate of #815

gh #815 is about whether a particular flowsheet converges. This is about the reporting, and it stands whatever happens to that model.

### LINKED ISSUE #820: Three Python success sets still call a square-problem feasible point a failure
Split out of gh #815, which fixed the `.sol` / CLI / `pyomo_pounce` surfaces but
deliberately left these three alone.

## The status

`Feasible_Point_Found` has exactly one producer in the engine:
`min_c_1nrm.rs` returns `RestorationOutcome::FeasiblePointFound`, reached only
through `square_feasible_point_found` in `resto_inner_solver.rs`, whose first
conjunct is `is_square_problem` — `c.x.dim() == c.y_c.dim()`, a port of
`IpoptCalculatedQuantities::IsSquareProblem`. So POUNCE emits it only for square
problems, exactly as Ipopt does, and on a square problem the objective is
constant: a feasible point *is* the solution. Ipopt's ASL driver reports it as
AMPL code `2`, in the `0..=99` solved band.

## What gh #815 changed

- `status_to_solve_result_num`: `100` → `2`, so both of Pyomo's `.sol` readers
  load the solve as a success. At `100` the contrib v2 reader
  (`pyomo/contrib/solver/solvers/asl_sol_reader.py`) returned
  `TerminationCondition.error` for a correct answer.
- `nlp_solve_succeeded` (`pounce-cli/src/main.rs`): exit 0, so the process exit
  code and the `.sol` it wrote cannot disagree.
  `the_exit_code_and_the_sol_band_never_disagree` pins that as an `iff` over
  the whole enum.
- `pyomo_pounce.v2._V2_STATUS` and `pyomo_pounce.sens._STATUS_RESULT`: both rows
  now match what the `.sol` route produces for the same solve.

## What is still split

Three Python success sets still count the status as a failure:

- `python/pounce/_minimize.py` — `_NLP_SUCCESS_STATUS = frozenset({0, 1})`.
  `Feasible_Point_Found` is code `6`. The comment above it reads "Codes 2..6
  (infeasible, tiny step, diverging, …) stay failures", which does not name
  this status and does not engage the square-problem argument.
- `python/pounce/_curve_fit.py` — imports the same frozenset, so it moves with
  the above.
- `python/pounce/jax/_path.py` and `python/pounce/torch/_path.py` —
  `_OK_STATUS = ("Solve_Succeeded", "Solved_To_Acceptable_Level")`.

These were left out of gh #815 on purpose, not by oversight. They are a
different contract: a scipy-style `success` flag on a library call that never
writes a `.sol`, so unlike the CLI they were not put in contradiction with a
file the same process had just written. gh #815's argument does carry over —
if a square solve's feasible point is the solution, `minimize` should say
`success=True` — but it is a user-visible change to a public return value and
deserves its own decision rather than being folded into a reporting fix.

## The work

1. Decide whether the square-problem argument should apply to the library
   surfaces. If yes, add code `6` to `_NLP_SUCCESS_STATUS` and
   `"Feasible_Point_Found"` to both `_OK_STATUS` tuples, and rewrite the
   "Codes 2..6" comment, which will otherwise be false.
2. Whichever way it goes, say so where a reader will find it. The present
   state is three surfaces silently disagreeing with the other four, and the
   reason gh #815 existed at all is that a divergence like this was recorded
   in a doc comment whose justification had stopped being true of the code.
3. If they change, `python/tests/` needs a case that reaches the status. Note
   the difficulty found in gh #815: **no fixture in the CLI corpus reaches this
   exit**, and 72 generated square models across three nonlinearity families,
   three start points and three row-scale spreads, run under five option sets,
   failed to reach it. The exit needs a model larger than the corpora carry —
   gh #815's own model is 536x536. A unit-level test against the status code is
   the practical option.

jkitchin:
Three related fixes, all of them found while working gh #815 and all of them
about the same solve. Keeping them in one PR because they are one story: a
square flowsheet that solved correctly, was reported as a failure by three
separate surfaces, and whose iteration count could not be used to tell what
had happened.

Fixes #815
Fixes #819
Fixes #820

---

## 1. A restoration failure opens the second-opinion ladder (`b604eb20`)

`SecondOpinionTrigger::for_status` recognised exactly two verdicts,
`Infeasible_Problem_Detected` and `Invalid_Number_Detected`. A solve ending
`Restoration_Failed` got no second opinion at all — even though that verdict
is a statement about the *path* in exactly the sense the other two are. The
rung that answers it, a displaced start (`start_point_perturbation=1e-2`), was
already implemented, on by default, and the highest-yield rung there is (13
recoveries of 15 over the KRONOS corpus, against `mu_strategy=adaptive`'s 4).
It was never reached.

**Only rung 3 opens on this trigger**, so the cost is exactly one extra solve,
not three. Rungs 1 and 2 vary the linear-solver scaling and the barrier
strategy from the *same* start — and the start is what put the iterate where
restoration failed.

It is deliberately not treated like a budget exit: the paragraph excluding
`Maximum_Iterations_Exceeded` argues the answer there is a bigger budget, and
a restoration failure stops far short of `max_iter`, so that argument does not
apply.

New fixture `square_flowsheet_resto.nl` — square (32 × 32, zero degrees of
freedom, `f(x) = 0`), flowsheet-shaped, mixing `1e-6` mole fractions with
`3e6` Pa, feasible by construction. It went `Restoration Failed!` where Ipopt
3.13.2 solves in 34; it now recovers to `Solve_Succeeded` in 54.

**This part does not fix gh #815's own model** — that one exits
`Infeasible_Problem_Detected` with the ladder already firing. Item 2 is the
part that does.

## 2. A square problem solved to feasibility is a solved-band answer (#815)

`Feasible_Point_Found` wrote AMPL `solve_result_num = 100`. Pyomo's contrib v2
`.sol` reader maps `100..=199` to `TerminationCondition.error`, so a correct
answer reached the caller as a *solver error*; the legacy reader maps it to
`optimal` with `status=warning` — the same complaint that moved
`Solved_To_Acceptable_Level` out of that band in gh #591. It is now `2`,
Ipopt's own code, in the `0..=99` solved band. The process exit code follows
(0, not 1), and both `pyomo_pounce` status tables report it as a success.

**The reason the band was `100` had stopped being true of the code.** The doc
comment defending it said POUNCE used the status loosely, for any usable
feasible point that missed convergence. It does not. The status has one
producer — `min_c_1nrm.rs` — behind one gate, now extracted as
`square_feasible_point_found`, whose first conjunct is `is_square_problem`
(`c.x.dim() == c.y_c.dim()`, a port of
`IpoptCalculatedQuantities::IsSquareProblem`). Same condition as Ipopt, same
meaning; and on a square problem there is no further criterion to miss,
because the objective is constant. Two surfaces in this repo had already
written the correct reading down (`gams/link.py` maps it to
`MODELSTAT_FEASIBLE`; `issue_390_nonlinear_equality_scale.rs` calls it "a
success-band answer — AMPL `objno` code 2"), so the repo had been
contradicting itself in comments for as long as the divergence existed.

**Measured.** gh #815 is a 536×536 IDAES naphtha-hydrotreater flowsheet, zero
degrees of freedom. Written the way IDAES writes it for Ipopt
(`writer_config={"scale_model": True}`), POUNCE reaches a constraint violation
of 2.208e-06 in the model's original units in 1.0 s, 18 iterations; with
`linear_presolve=True`, 2.899e-06 in 0.8 s, 7 iterations. The two agree on the
flowsheet's thiophene conversion to six digits. That solve used to load as
`TerminationCondition.error`; it now loads `optimal` / `ok`, and the `.sol`
line moved from `objno 0 100` to `objno 0 2`.

**It does not make the model as filed solvable, and the issue's premise did
not survive measurement.** On the raw unscaled `.nl` POUNCE stalls at
`inf_pr ≈ 4.97e+03`; so does Ipopt 3.13.2/ma27, which exits `Restoration
Failed` at 2418 iterations from a stall at 1.56e+01 against POUNCE's 1.60e+01,
the two trajectories bit-identical over iterations 0–3. An 18-configuration
option sweep moved none of it. What makes the model tractable is IDAES's own
per-variable scaling applied by the NL writer — a frontend configuration, not
reachable from any solver option.

### The Python surfaces move with it (#820)

Four success sets counted the status as a failure, so `minimize` returned
`success=False` on a solve whose `.sol` the same build calls optimal:

- `_minimize._NLP_SUCCESS_STATUS` (shared by `_curve_fit`, which imports the
  frozenset rather than copying it),
- `jax._path._OK_STATUS`,
- `torch._path._OK_STATUS`,
- and — **a fourth the issue did not name** — `_starts._DONE_STATUS`, where
  the cost was worse than a wrong flag: a multi-start candidate that *solved*
  was scored `solve failed`, given a `+10` health penalty, and eliminated at
  its rung.

Both path followers already require an all-equality model
(`_require_equality_constraints`), which is the shape that produces this
status, so refusing it aborted the path at a converged anchor.

`python/tests/test_issue_815_square_feasible_point.py` pins all four, and pins
the premise underneath them by parsing `status_to_solve_result_num` out of
`crates/pounce-solve-report/src/lib.rs` — so an edit that moves the Rust band
back out of `0..=99` fails the Python test that depends on it, instead of
leaving the two sides to drift silently.

## 3. A solve that ends inside restoration reports the iterations it spent there (#819)

`Number of Iterations....:` is, in Ipopt, the index of the last printed
iteration row — `r` rows included — on **every** exit path. Measured over four
upstream logs from this repo's issue corpus rather than assumed: `2418r`/2418
(`Restoration Failed`), `412r`/412 (local infeasibility), `1547r`/1547,
`1348r`/1348.

POUNCE printed the same `r` rows and then reported a number that ignored them.
On `square_flowsheet_resto` the log ended at row `131r` above a summary that
said **47**; on gh #815's raw `.nl` the three solves reported **3 / 0 / 3**
iterations for work that actually ran 3000 / 1471 / 1945.

Three defects stacked:

1. The outer counter's roll-forward lived only on the `Recovered` path
   (`min_c_1nrm.rs`, step 2g). All four *terminating* restoration outcomes
   return ahead of it, so the entire sub-solve vanished from the summary. It
   is now applied in `invoke_restoration` for every non-`Recovered` outcome,
   which is trajectory-safe by construction: each of those arms `Terminate`s,
   and `optimize_inner`'s loop breaks without consulting the counter again.
2. `restoration_inner_iters` summed the inner IPM's **absolute** terminating
   `iter_count`. That counter is seeded from the outer's
   (`inner.iter_count = outer_iter + 1`, mirroring `IpRestoMinC_1Nrm.cpp:181`),
   so the sum was a position in a shared numbering, not a length — the same
   misreading gh#664 documents for the stall gate. It is now the delta against
   the outer count at entry, i.e. the number of `r` rows actually printed.
3. The count was read off the `Some(result)` arm, so on every path that bailed
   — which is every path that *ends* in restoration, the ones the issue is
   about — it recorded `0`. The hook now returns
   `RestoInnerReturn { inner_iter_count, result }` with the count outside the
   `Option`, and the caller records it ahead of the `None` check.

The console summary gained a line for the split, printed only when restoration
ran:

```
Number of restoration iterations                     = 84 (in 1 call)
```

Two doc comments in `solve_statistics.rs` are corrected with it, and two
`pounce-studio-core` surfaces that printed `restoration_outer_iters` as
"outer iters spent in restoration" now print `restoration_inner_iters`:
`restoration_outer_iters` is bumped once per call, so it always equals
`restoration_calls`, and the studio markdown was printing the same number
twice in one sentence.

**A finding worth flagging, because it contradicts gh #815's own text.** With
the count fixed, the first solve on the raw `.nl` reports 3000 — the iteration
cap. So the verdict *was* budget-limited, and the issue's "it stopped far
short of any iteration limit, which is why the infeasibility verdict looks
premature" was an artefact of the miscount, not an observation about the
solve.

---

## Verification

- `cargo test --workspace` green.
- `python -m pytest python/tests` green; `pyomo-pounce` matches its
  main-worktree baseline (2 pre-existing failures and 1 pre-existing
  collection error on both sides).
- **`scripts/sweep-fixtures.sh`, both legs, 156 fixture-legs, against a
  binary built from `main` (`f677f244`) — exactly 3 lines move, and all three
  are explained:**

```
exact  issue_372_infeasible_bounds  nlp  InfeasibleProblemDetected  it=5  -> it=10
lbfgs  issue_372_infeasible_bounds  nlp  InfeasibleProblemDetected  it=4  -> it=9
exact  square_flowsheet_resto       nlp  RestorationFailed  it=47 -> SolveSucceeded it=54
```

  `issue_372_infeasible_bounds` is item 3 doing its job and nothing else: the
  two logs are **row-for-row identical** — same iterates, same alphas, same
  line-search codes — and the solve ends at row `10r`, which is the number now
  reported. The restoration line reads `= 5 (in 1 call)`, matching rows
  `6r`–`10r`. Objective unchanged to all printed digits; status unchanged;
  engine unchanged.

  `square_flowsheet_resto` is item 1's new fixture, moving because the
  baseline binary predates the ladder. Its **lbfgs** leg is the control: same
  model, same start, exits `Maximum_Iterations_Exceeded` at the 3000 cap and
  is unmoved — the budget-exit exclusion executing on a real leg rather than
  only in a unit test.

  No status, objective or **engine** moves anywhere else in the corpus.

## New tests

- `crates/pounce-cli/tests/issue_819_restoration_iteration_count.rs` — three
  tests against a real failing solve, all parsing the numbers out of the log
  rather than hardcoding them, so they survive a trajectory change that moves
  the values without breaking the identity: the reported total equals the last
  printed row's index; the restoration line equals the `r`-row count; and the
  restoration span sits strictly inside the total (which a fix that just set
  one field from the other would fail).
- `min_c_1nrm::tests::the_inner_iteration_count_survives_a_failed_sub_solve` —
  the `None` path, which is the path the issue is about and the one the old
  code could not report on.
- `python/tests/test_issue_815_square_feasible_point.py` — 7 tests over the
  four Python surfaces plus the Rust band they rest on.
- `the_exit_code_and_the_sol_band_never_disagree` — the guard stated as an
  `iff` over all twenty statuses, so a status added to the exit-code success
  set but not the solved band (or the reverse) fails it. gh #815 is that
  predicate failing on `FeasiblePointFound`; gh #591 was the same shape one
  status over.

## What these tests are *not* evidence about

Item 2 has **no end-to-end fixture**: no model in the CLI corpus reaches the
`Feasible_Point_Found` exit, and 72 generated square models — three
nonlinearity families, three start points, three row-scale spreads, five
option sets — failed to reach it too. It needs a model larger than the corpus
carries, and the end-to-end evidence is gh #815's own 536×536 flowsheet, cited
above. The new unit tests hold the conjunct the band argument rests on
(`only_a_square_problem_yields_a_feasible_point_verdict` sets every other
input to a passing value, so squareness alone decides it), not the exit.

Item 3's CLI test covers one fixture, one exit path (`Restoration_Failed`),
one linear solver. The `Recovered` path had a working roll-forward before this
change and still does; `restoration_deadline.rs` covers the multi-call arm and
still passes with the counter's semantics changed from absolute to length.


jkitchin:
Re-ran the sweep with the new fixture in the corpus, as the positive control the 0-of-154 run could not be:

```
=== legs:      156 /      156 ===
71c71
< exact  square_flowsheet_resto   nlp   RestorationFailed   it=47   obj=0
---
> exact  square_flowsheet_resto   nlp   SolveSucceeded      it=54   obj=0
=== changed: 2 ===
```

Exactly one leg moves, and it is the one the change is about. Everything else is bit-identical across both legs.

The fixture's **lbfgs** leg is a second data point I did not plan for and am glad to have: it exits `MaximumIterationsExceeded it=3000` and is **unmoved** by the patch. That is the budget-exit exclusion executing on a real leg rather than only in the unit test — same model, same starting point, different verdict, no ladder. `only_path_verdicts_open_a_ladder` pins the map; this shows the map being consulted.

Cost to the corpus: 1.47 s for the lbfgs leg (it runs the full 3000), 0.03 s for the exact one. Well inside the sweep's 300 s per-model timeout.

PR #821, opened 2026-08-28
A square-problem feasible point is a solved-band answer, not a solver error

### LINKED ISSUE #815: Local infeasibility reported on a feasible square IDAES flowsheet that Ipopt solves

jkitchin:
Fixes #815.

`Feasible_Point_Found` wrote AMPL `solve_result_num = 100`. Pyomo's contrib v2
`.sol` reader maps `100..=199` to `TerminationCondition.error`, so a correct
answer reached the caller as a **solver error**; the legacy reader maps the
band to `optimal` with `status=warning` — the same complaint that moved
`Solved_To_Acceptable_Level` out of it in #591. It is now `2`, Ipopt's own
code, in the `0..=99` solved band.

## The justification for `100` had stopped being true of the code

The doc comment defending the divergence said the two statuses do not mean the
same thing: that Ipopt returns `FEASIBLE_POINT_FOUND` only for a square
problem, where a feasible point *is* the solution, while POUNCE used it more
loosely for a usable feasible point that missed the convergence criteria.

POUNCE does not. The status has exactly one producer — `min_c_1nrm.rs`
returning `RestorationOutcome::FeasiblePointFound` — behind exactly one gate in
`resto_inner_solver.rs`, whose first conjunct is `is_square_problem`
(`c.x.dim() == c.y_c.dim()`, a port of
`IpoptCalculatedQuantities::IsSquareProblem`). Same condition as Ipopt, so the
same meaning; and on a square problem the objective is constant, so there is no
remaining criterion the point could be said to have missed.

Two surfaces in this repo had already written the correct reading down and were
being contradicted by that comment: `python/pounce/gams/link.py` maps the status
to `(MODELSTAT_FEASIBLE, SOLVESTAT_NORMAL)` and lists it as a success, and
`issue_390_nonlinear_equality_scale.rs` calls it "a success-band answer — AMPL
`objno` code 2, which every band table reads as SOLVED".

## Changes

| | before | after |
|---|---|---|
| `status_to_solve_result_num` | `100` | `2` |
| `pounce` exit code (non-`-AMPL`) | `1` | `0` |
| `v2._V2_STATUS` | `(unknown, feasible)` | `(convergenceCriteriaSatisfied, optimal)` |
| `sens._STATUS_RESULT` | `(feasible, warning)` | `(optimal, ok)` |

The two `pyomo_pounce` rows are set to exactly what the `.sol` route now
produces for the same solve — the rule both tables' docstrings already state,
and the rule #591 established.

The exit code moves because leaving it would have created a *fresh*
inconsistency: the same process would exit 1 while the `.sol` beside it read
solved.

The five-term gate is extracted verbatim into `square_feasible_point_found` so
the `is_square_problem` conjunct — which is what the whole band argument rests
on — can be held by a test instead of a comment. No logic changed.

## Measured

gh #815 is a 536x536 IDAES naphtha-hydrotreater flowsheet, zero degrees of
freedom. Written the way IDAES writes it for Ipopt
(`writer_config={"scale_model": True}`), through Pyomo end-to-end, residuals in
the model's original units:

| | result | wall | max abs residual | thiophene conversion |
|---|---|---|---|---|
| POUNCE, `scale_model` | `optimal` / `ok` | 1.0 s | 2.208e-06 | 0.39187704 |
| POUNCE, `linear_presolve` + `scale_model` | `optimal` / `ok` | 0.8 s | 2.899e-06 | 0.39187658 |

Before this change those same two solves loaded as `optimal` / **`warning`**
through the legacy reader and as `TerminationCondition.error` through the v2
reader. The `.sol` line moved from `objno 0 100` to `objno 0 2`.

## What this does not fix, and a correction to the issue's premise

The model **as filed** — the raw unscaled `.nl` — is not solvable by either
solver here:

- POUNCE stalls at a constraint violation of ~4.97e+03. An 18-configuration
  option sweep (linear solver, `mu` strategy, scaling method,
  `nlp_scaling_max_gradient`, `mu_init`, `start_with_resto`,
  `expect_infeasible_problem`, bound handling, `mc64`, iteration cap) ended
  `Restoration_Failed` or `Infeasible_Problem_Detected` in every case, at that
  same violation.
- Ipopt 3.13.2/ma27 exits `Restoration Failed` at 2418 iterations from a stall
  at `inf_pr = 1.56e+01`, against POUNCE's 1.60e+01 — the two trajectories are
  bit-identical over iterations 0–3.

What makes the model tractable is IDAES's own per-variable scaling factors
applied by the NL writer. That is a frontend configuration; no
`nlp_scaling_method` substitutes for it. Ipopt did not solve this model in any
configuration tried here, including through the reporter's own script — so the
issue's framing ("Ipopt solves it, POUNCE does not") does not hold, while
POUNCE's actual defect on it was real and is what this PR fixes.

## Tests

- `the_exit_code_and_the_sol_band_never_disagree` — the guard as an `iff` over
  all twenty statuses, so a status added to one channel and not the other fails
  it in either direction. #815 is this predicate failing on
  `FeasiblePointFound`; #591 was the same shape one status over.
- `only_a_square_problem_yields_a_feasible_point_verdict` — squareness alone
  decides, every other input held at a passing value.
- `each_residual_guard_refuses_on_its_own` — the other four conjuncts, each
  refused individually from that passing baseline.
- `a_square_problem_feasible_point_is_in_the_solved_band` — the code and the
  band, with the reason.

**What these are not evidence about.** None of it is end-to-end. No fixture in
the CLI corpus reaches this exit; 72 generated square models (three
nonlinearity families x three start points x three row-scale spreads) under
five option sets failed to reach it as well. The exit needs a model larger than
the corpus carries, and the end-to-end evidence is #815's own 536x536 model
above.

`scripts/sweep-fixtures.sh` moved **0 of 154** fixture-legs. That is a control,
not evidence: the sweep does not read `solve_result_num`, and the only
non-reporting edit here is a verbatim extraction.

## Left split on purpose

Three Python success sets still count the status as a failure —
`_minimize._NLP_SUCCESS_STATUS` (shared by `_curve_fit`), `jax._path._OK_STATUS`
and `torch._path._OK_STATUS`. Unlike the CLI they are a scipy-style `success`
flag on a library call that writes no `.sol`, so they were not put in
contradiction with a file the same process had just written, and widening a
public return value deserves its own decision. Filed as #820 so it has an owner
rather than reading as an oversight.

#817, opened earlier against this issue, is **not** a fix for it and is not
part of this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,92 @@ changes.
 
 ## [Unreleased]
 
+- **A square problem solved to feasibility is now reported as a success
+  (gh #815).** `Feasible_Point_Found` wrote AMPL `solve_result_num = 100`.
+  Pyomo's contrib v2 `.sol` reader maps `100..=199` to
+  `TerminationCondition.error`, so a correct answer reached the caller as a
+  *solver error*; the legacy reader maps it to `optimal` with
+  `status=warning`, the same complaint that moved `Solved_To_Acceptable_Level`
+  out of that band in gh #591. It is now `2`, Ipopt's own code, in the
+  `0..=99` solved band. `pounce`'s process exit code follows (0, not 1), and
+  the two `pyomo_pounce` status tables (`v2._V2_STATUS`,
+  `sens._STATUS_RESULT`) report the status as a success rather than as
+  `unknown` / `warning`.
+
+  **The reason the band was `100` had stopped being true of the code.** The
+  doc comment defending it said the two statuses do not mean the same thing:
+  that Ipopt returns `FEASIBLE_POINT_FOUND` only for a square problem, where a
+  feasible point *is* the solution, while POUNCE used it more loosely for any
+  usable feasible point that missed the convergence criteria. POUNCE does not.
+  The status has one producer — `min_c_1nrm.rs` returning
+  `RestorationOutcome::FeasiblePointFound` — behind one gate, now
+  `square_feasible_point_found` in `resto_inner_solver.rs`, whose first
+  conjunct is `is_square_problem`: `c.x.dim() == c.y_c.dim()`, a port of
+  `IpoptCalculatedQuantities::IsSquareProblem`. Same condition as Ipopt, hence
+  the same meaning, and on a square problem there is no further criterion to
+  miss because the objective is constant. Two other surfaces had already
+  written the correct reading down —
+  `python/pounce/gams/link.py` maps the status to
+  `(MODELSTAT_FEASIBLE, SOLVESTAT_NORMAL)`, and
+  `issue_390_nonlinear_equality_scale.rs` calls it "a success-band answer —
+  AMPL `objno` code 2, which every band table reads as SOLVED" — so the repo
+  had been contradicting itself in comments for as long as the divergence
+  existed.
+
+  **What it cost, measured.** gh #815 is a 536x536 IDAES naphtha-hydrotreater
+  flowsheet, zero degrees of freedom. Written the way IDAES writes it for
+  Ipopt (`writer_config={"scale_model": True}`), POUNCE reaches a constraint
+  violation of **2.208e-06 in the model's original units in 1.0 s**, 18
+  iterations; adding `linear_presolve=True` gives a 468x468 model, 2.899e-06
+  in 0.8 s, 7 iterations. The two configurations agree on the flowsheet's
+  thiophene conversion to six digits (0.3918770, 0.3918766). Before this
+  change that solve loaded as `termination_condition=optimal, status=warning`
+  through the legacy reader and as `TerminationCondition.error` through the
+  v2 reader; it now loads `optimal` / `ok`, and the `.sol` line moved from
+  `objno 0 100` to `objno 0 2`.
+
+  **This does not make the model as filed solvable, and the issue's premise
+  did not survive measurement.** On the raw unscaled `.nl`, POUNCE stalls at a
+  constraint violation of ~4.97e+03; so does Ipopt 3.13.2/ma27, which exits
+  `Restoration Failed` at 2418 iterations from a stall at `inf_pr = 1.56e+01`
+  against POUNCE's 1.60e+01, the two trajectories being bit-identical over
+  iterations 0-3. An 18-configuration option sweep — linear solver, `mu`
+  strategy, scaling method, `nlp_scaling_max_gradient`, `mu_init`,
+  `start_with_resto`, `expect_infeasible_problem`, bound handling, `mc64`
+  equilibration, iteration cap — moved none of it: every configuration ended
+  `Restoration_Failed` or `Infeasible_Problem_Detected` at that same
+  violation. What makes the model tractable is IDAES's own per-variable
+  scaling factors applied by the NL writer, which is a frontend
+  configuration and not reachable from any solver option; no
+  `nlp_scaling_method` substitutes for it. Ipopt did not solve this model in
+  any configuration tried here, including through the reporter's own script.
+
+  **What the new tests are and are not evidence about.**
+  `the_exit_code_and_the_sol_band_never_disagree` states the guard as an `iff`
+  over all twenty statuses, so a status added to the exit-code success set but
+  not the solved band, or the reverse, fails it — gh #815 is that predicate
+  failing on `FeasiblePointFound` and gh #591 was the same shape one status
+  over. `only_a_square_problem_yields_a_feasible_point_verdict` holds the
+  conjunct the whole band argument rests on, with every other input at a value
+  that would pass, so squareness alone decides it. None of this is end-to-end:
+  **no fixture in the CLI corpus reaches this exit**, and 72 generated square
+  models — three nonlinearity families, three start points, three row-scale
+  spreads — under five option sets failed to reach it too. It needs a model
+  larger than the corpus carries, and the end-to-end evidence is gh #815's own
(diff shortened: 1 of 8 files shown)


jkitchin:
Folding this into #817 — the two changes came out of the same investigation and
belong in one branch. The commit is cherry-picked there unchanged
(`fe72d3f5` → `ec85a88b`), along with the fixes for #819 and #820, which are
also gh #815 work rather than separate follow-ups.

Closing unmerged; nothing here is dropped.

PR #823, opened 2026-08-28
Recover the exact Hessian from the analytic Jacobian for models that have none

jkitchin:
/cc @srikanth-gm for review — GitHub would not accept a formal review request (not a repo collaborator), so tagging here. **All five of your suggestions are now incorporated**, including the two that [this reply](https://github.com/jkitchin/pounce/pull/823#issuecomment-5455572873) deferred: the CasADi plugin's Hessian structure/values split and the `stats()` exposure. Two things in that reply are superseded — the plugin gate was a *hard failure* on first-derivative-only models rather than merely wasteful, and the structure-only pass-through needed **no** C-interface addition after all. See *Review follow-ups* below.

## TL;DR

Adds `hessian_approximation=finite-difference`: recovers the **exact** Lagrangian Hessian by graph-coloured finite differences of the analytic Jacobian, for models that supply first derivatives and no second ones. Opt-in; every default unchanged.

Measured against a model that **genuinely has no Hessian** (`POUNCE_DROP_HESSIAN=1`, a wrapper TNLP reporting `nnz_h_lag = 0` and declining `eval_h` — what `pounce-py` presents for a Python object with no `hessian` method), not a simulated one:

| mesh | leg | status | iters | wall | objective |
|---|---|---|---|---|---|
| N=160 | *ref:* `exact`, Hessian available | Optimal | 30 | 2.8 s | 65.3711067940 |
| N=160 | no-Hessian, `limited-memory` | Acceptable | 207 | 53.7 s | 65.3704926894 |
| N=160 | **no-Hessian, `finite-difference`** | **Optimal** | **30** | **38.7 s** | 65.3711067940 |
| N=320 | *ref:* `exact`, Hessian available | Optimal | 57 | 14.5 s | 65.3269077802 |
| N=320 | no-Hessian, `limited-memory` | **MaxIter** | 1210 | 706 s | **65.3951930783** |
| N=320 | **no-Hessian, `finite-difference`** | **Optimal** | **62** | 430 s | 65.3269077802 |

**It reaches the exact path's iteration count** — 30 against 30 at N=160, 62 against 57 at N=320 — and its objective agrees with the exact path to 13 significant figures. Against the only option such a model has today: **6.9× fewer iterations at N=160** and **1.4× wall**; at N=320, 19.5× fewer iterations and 1.6× wall, where limited-memory does not converge at all and lands on 65.395 against a true optimum of 65.326908.

> **The wall-clock margin is modest and that is a correction.** An earlier revision of this body claimed 4.6× wall. That number was measured with a pattern that omitted the objective's curvature — a correctness bug found in review. Fixing it improved iterations (38 → 30, 106 → 62) and cost wall time, because the conservative objective clique is wide on this model. See *Review findings* below.

## Problem

`benchmarks/large_scale/laptime` was added for gh#698: "a 60 000-variable collocation model with analytic Jacobians and no analytic Hessian". Such a model has exactly one option today, `limited-memory`, and on this family it degrades with mesh refinement until it stops converging — the README's own table reads 71 → 144 → 276 → failure.

The gap is entirely iteration count, not linear algebra. Measured per-iteration cost at N=160 is 0.148 s for `exact` against 0.163 s for `limited-memory`: L-BFGS is *slightly more expensive per iteration* despite presenting a diagonal `(1,1)` block, because the Sherman-Morrison-Woodbury back-solves cost more than the sparser KKT saves. So the whole 9× wall gap is the curvature model.

## Cause

Not a defect — a missing capability. An analytic Jacobian is already enough to *build* the Hessian, and nothing was using it that way.

`∇ₓL = ∇f + J_cᵀy_c + J_dᵀy_d` is closed form whenever `J` is, so its directional derivative costs one gradient plus one Jacobian evaluation. With a known sparsity pattern, one probe recovers a whole group of structurally orthogonal columns rather than one column.

The affordability was measured *before* the thing was built (`POUNCE_HESS_PATTERN_CENSUS`): the true Hessian pattern's widest row is `rho_max = 15` with a mean of 5.68, and it is **identical at N=320** — the row width is set by the per-stage stencil, not the horizon. So the probe count does not grow with the mesh. On the declared pattern that realises as 17 colour groups, against a Jacobian evaluation costing 5.4 ms in a 92.6 ms iteration.

## Fix

`crates/pounce-algorithm/src/hess/fd_hessian.rs`. Columns grouped by Curtis-Powell-Reid orthogonality (greedy, largest-first); each probe read directly, no linear solve.

Two pattern sources:

- `declared` — the TNLP's Hessian **structure** call. Never evaluates a second derivative, so it is available to a model that cannot compute one. Every `.nl` declares one via AMPL's AD. 17 groups on `laptime`.
- `jacobian` — `(O ⊗ O) ∪ ⋃ⱼ (supp(∇gⱼ) ∩ N) ⊗ (supp(∇gⱼ) ∩ N)`, where `O` is the objective's nonlinear variables and `N` the model's nonlinear-variable set. Needs nothing beyond the Jacobian pattern and the linearity information every TNLP can state. A strict superset of the truth, which is safe but costs probes: 143 854 nonzeros against the true 28 000, and 341 groups against 17. This is the **automatic fallback** when a model declares no Hessian structure, so a Hessian-less model sets one option, not three.

**Both terms of `O ⊗ O` and the constraint cliques are required.** The constraint Jacobian says nothing about `∇²f`; omitting `O ⊗ O` makes the pattern a *subset* of the truth and silently drops curvature. There is likewise no mode that guesses a subset deliberately.

### Review findings (@srikanth-gm), incorporated in `f098c53`

**The objective clique was missing — a correctness bug.** The pattern was `⋃ⱼ supp(∇gⱼ) ⊗ supp(∇gⱼ)` alone, which covers only `Σ yⱼ ∇²cⱼ`. `laptime` could not expose it by objective value (minimise-final-time), but exposed it by cost: adding the term took N=160 from 38 iterations to 30 — exactly the exact-Hessian count — and N=320 from 106 to 62. Pinned by `the_objective_clique_is_in_the_jacobian_derived_pattern`.

**The nonlinear-variable mask was suggested and implemented.** A variable linear in every `f` and `g` has structurally zero off-diagonal Hessian entries, so it leaves the cliques. Removes ≈53 000 pattern entries on `laptime`. The full primal diagonal stays regardless — the barrier and inertia correction need those rows.

**The two pull against each other.** `laptime`'s objective is nonlinear in 320 variables, so the conservative `O ⊗ O` clique is 320 wide and alone takes `rho_max` from 59 to 338 and groups from 76 to 341. Pattern nonzeros 146 267 → 143 854; iterations improved everywhere; N=160 wall 12.6 s → 38.7 s. A model that can state its true Lagrangian-Hessian sparsity uses `declared` and pays none of it.

### Rejected, with what ruled each out

**Star colouring.** Fewer probes (76 → 42) and its recovery is algebraically exact — verified by recovering a known matrix through it, not assumed. It still loses: 404 iterations to a *wrong* objective against CPR's 38 to the right one. Group size is not the cause and the measurement refutes it — `declared/star` packs the largest groups of the four (580 columns per probe) and converges in 30. The cause is the finite-difference remainder: a forward difference carries `½ Σ_{m,p ∈ g} T_imp h_m h_p` into row `i`, and `T_imp ≠ 0` needs `H_im` and `H_ip` both nonzero — which CPR's distance-2 property forbids in one group and star colouring does not. Opt-in; CPR is the default.

**Partitioned quasi-Newton**, both decompositions. See below.

**A magnitude cap on quasi-Newton element updates.** Every finite value measured was worse than off, non-monotonically. Recorded so the next person does not retune it.

### Warm starting — three routes, all measured, none adopted

"Just warm-start it, it's a trajectory" is the obvious objection to this PR and it does not survive measurement. **None of this changes solver behaviour**; the only artifact in the diff is `benchmarks/large_scale/mesh_continuation.py`.

**1. Mesh continuation, primal-only — measured, a regression.** `mesh_continuation.py` maps a coarse solution onto a fine mesh by arc length through the `.col` names. POUNCE cannot do this itself — it sees a flat vector — so it is frontend work by construction. Every entry transferred:

| step | cold | warm | |
|---|---|---|---|
| N=80 → 160 | 30 it / 3.6 s | **93 it / 15.3 s** | **3.1× worse** |
| N=160 → 320 | 57 it / 18.5 s | **74 it / 24.7 s** | **1.3× worse** |

The interpolation is not at fault: the transferred point has objective 65.458 against the cold start's 88.889 and 8× less constraint violation, and still costs 3.1× the iterations. The mechanism is **centrality** — a converged coarse solution sits on its active set with slacks at `7.5e-56`, and a barrier method wants a well-centred interior point. `benchmarks/warmstart/README.md` cites *Not All Warm Starts Help* (arXiv:2606.08984) for this regime.

**2. Full warm-IPM (duals + μ) — argued down, not built.** `benchmarks/warmstart` measures it at 4.42× / 9.12× / 1.96×. Not worth building here, for an arithmetic reason: mesh refinement moves the answer first-order (65.462928 → 65.371107 → 65.326908, ≈1.4e-3 relative per step) while the coarse solve's converged μ is ~1e-9. Reusing that μ needs the point within ~1e-9 of the fine central path; it is within ~1e-3. So `warm_start_target_mu` would be ~1e-3, skipping ~2 of ~8 barrier decades — a ~25% ceiling. The winning families are also small parametric perturbations on a *fixed* variable space (churn 0.05–1.16/step); mesh refinement changes `n` from 9 294 to 18 574. The better design, if anyone pursues it, is **μ-matched continuation**: solve the coarse mesh only to μ ≈ 1e-3, transfer, continue tightening.

**3. SQP + warm start — measured, closed.** The active-set path is where working-set warm starts pay (17.46× / 2.26× / 12.55× active-set-change reductions). It cannot run this problem: at the **smallest** mesh (n = 4 654) `algorithm=active-set-sqp` emits the banner and then **not one SQP iteration in 80 s**, with two 240 s timeouts. RSS flat at ~140 MB, so not a dense-Hessian blow-up — stuck in the first QP subproblem. The families where SQP wins are all n ≤ 47.

## Blast radius

**Bit-identical except the targeted case.** Both default legs are untouched: a new `HessianApproxChoice` variant reached only by an explicit option, default still `exact`. Non-Rust additions are a benchmark harness and two dev-notes.

**`scripts/sweep-fixtures.sh` has been run — empty diff.** ~~NOT been run~~; it was outstanding when this body was first written. 156 fixture-legs, both `exact` and `lbfgs`, against `origin/main` after the merge: not one line moves. The corpus is live rather than vacuously empty — 5 `cvx-qcqp`, 36 `cvx-qp`, 37 `nlp` per leg, no `NO_JSON` or `unknown`. The sweep runs `exact` and `lbfgs` only, so it is evidence that this branch does not perturb the paths existing users are on; it is **not** evidence about the `finite-difference` path, which the tests below own.

~~the finite-difference path works through restoration unchanged~~ — **that was false, and it is fixed.** Restoration was scoped out of the partitioned path only. The finite-difference updater kept running inside the restoration sub-IPM, carrying the *original* NLP's pattern and objective clique against the five-block compound primal. On a nonconvex fixture it was the only Hessian mode to report `RestorationFailed`, returning an answer 1.9 infeasible, where `exact`, `limited-memory` and `partitioned` all converge. Restoration now runs the limited-memory updater for it too, for the first of the two reasons already stated for `partitioned`. Pinned by `crates/pounce-restoration/tests/resto_fd_hessian.rs`, which fails on the parent.

## Tests

**These are new-feature tests, not regression tests — they do not fail on the parent commit, because the feature does not exist there.** Stating that rather than ticking the box.

- `cpr_recovers_a_banded_matrix_exactly` / `star_recovers_a_banded_matrix_exactly` — recover a known matrix through simulated probes. A colouring that makes a probe component carry a *sum* reads it as one, silently.
- `overlapping_cliques_are_validated_not_assumed` — the same on the shape the banded fixtures missed: overlapping cliques, which is what `supp(∇g) ⊗ supp(∇g)` produces. The banded fixtures passed while star colouring was wrong on `laptime`.
- `the_objective_clique_is_in_the_jacobian_derived_pattern` — the review finding, on a fixture where no constraint row contains both objective variables.
- `a_clique_forces_its_size_in_groups_under_either_coloring`, `groups_are_structurally_orthogonal`, `a_dense_row_forces_its_width_in_groups`.

**A process note worth surfacing:** the recovery's soundness condition was a `debug_assert`, therefore compiled out in release, so an invalid colouring produced a wrong Hessian with *no diagnostic* — it surfaced only as 404 iterations and a wrong objective. Now an unconditional validation with automatic fallback to CPR, reported by `FdStats::coloring_fell_back`. On `laptime` the star colouring **passes** that validation and is still numerically wrong: the predicate is about algebraic recoverability and cannot see the differencing error.

**Deliberately not pinned:** none of the `laptime` or warm-start measurements are asserted in CI — the models are generated, not vendored, and runs take minutes. Numbers and reproduction commands live in `dev-notes/fd-hessian-from-jacobian.md`. `mesh_continuation.py` has no test of its own.

## Review follow-ups — now IN this PR

All three of what this section previously listed as "known follow-ups, not in this PR" are done. CasADi was installed and the plugin built, so they could be compiled, measured and tested.

- **The CasADi plugin now honours `finite-difference`.** The defect was worse than the reading suggested: `finite-difference` did not merely *waste* CasADi's exact-Hessian machinery, it **failed outright** with `Derivatives cannot be calculated for …` — the same error `exact` correctly gives — on precisely the first-derivative-only models the mode exists for (an external `Callback`, an FMU, a `DaeBuilder` transcription). `exact_hessian_` now stands only for "may call `cb_h` for **values**"; `hessian_structure_` for "can declare a **pattern**". The forced `limited-memory` at the old line 792 is keyed on the mode instead, which it always should have been — it survived only because the option-forwarding loop re-sends the user's own value afterwards.
- **Structure-only pass-through turned out to need no C-interface addition.** The previous claim here was wrong. The plugin passes `nnz_h > 0` with a live `cb_h` that answers the **structure** request and **refuses** a values one, which the existing C interface already supports — so a completed solve is itself proof that every number was recovered by probing. `fd_hessian_pattern=declared` therefore works from CasADi today, and `fd_hessian_pattern=jacobian` skips building the symbolic Hessian entirely. `declared` falls back to the Jacobian derivation on its own when CasADi cannot build one. The same split is reproduced in the generated C, whose emitted `eval_h` serves the pattern as a baked-in literal and refuses values.
- **The pattern source, nnz, `n`, colour count, `rho_max` and clique-widening flag are exposed** through `GetPounceFdHessianStats` and `stats()["fd_hessian"]`. The reported source is the one the run **ended up with**, not the one requested — `declared` falls back silently, and that fallback is the difference between 17 probe groups and 341 on `laptime`, so reporting the request would hide the number a reader is there for.

One more defect found while wiring this, unrelated to the FD Hessian but in the same family of "set by every layer, read by none":

- **`start_with_resto` did nothing.** It was threaded from the `OptionsList` into `AlgorithmBuilder::resto`, into `RestoAlgorithmBuilder`, and into `MinC1NrmDriver` — a field on the *inner* restoration solver, which has no first outer iteration to force. Forcing restoration in the first iteration is an outer-loop behaviour and the outer loop consumes it now. `unimplemented_options.rs`'s `the_restoration_switches_reach_the_builder` could not catch it: it asserts the value *reaches the builder*, which is exactly the "read site populating a field nobody consumes" its own comment names as the defect to avoid.

## Still not in this PR

- **The `rho_max` 59 → 338 / N=160 wall 12.6 s → 38.7 s cost of the objective clique stands, and is not a tunable.** Diagnosed rather than accepted: `laptime`'s objective is a control-rate regulariser over 320 controls, so its true `∇²f` is banded with a row width of 3 while the safe clique is 320 wide, and CPR needs at least `rho_max` groups — the clique alone floors the probe count at 320. What removes it is the model declaring any Hessian structure at all, which is what `fd_hessian_pattern=declared` uses and what the CasADi fix above finally makes reachable. Two refinements were built and **rejected**, both recorded in the dev-note: an objective-Hessian union passed from the plugin, whose benefit is unreachable because an opaque CasADi callback forces a *dense* Jacobian (measured 3480/3480 on a banded-3 model), and a nonlinear-mask intersection that is a no-op on the objective's own set.
- **No Ipopt baseline.** Unchanged: `ipopt` is not installed here and `benchmarks/large_scale/ipopt_ma57.json` predates the `laptime` family, so every number is POUNCE against POUNCE.

## Also on this branch — negative results, not proposed for use

`hessian_approximation=partitioned`: two partitioned quasi-Newton decompositions (Asprion, Chinellato & Guzzella, doi:10.1155/2014/341716). Per-constraint elements scale the wrong way — converge at N=80, fail at N=160. Per-primal-block elements, the paper's own partition, have near-flat iteration scaling (276 → 252 → 338 across a 4× problem, against limited-memory's 126 → 246 → does-not-converge) but never converge faster and fail at N=320 inside restoration. Both are beaten by `finite-difference` on every measure.

Kept for the record with full measurements — including two wrong inferences made along the way and how each was caught — in `dev-notes/partitioned-quasi-newton-prototype.md`. **Happy to split these commits out and land the finite-difference path alone.**

---

- [x] Tests fail on the parent commit for the stated reason — **now partly yes.** The original colouring tests are still new-feature tests and do not (see Tests above). Everything added since is a regression test that does: `resto_fd_hessian.rs` (`RestorationFailed` on the parent), `start_with_resto.rs` (`restoration_calls == 0` on the parent), and three checks in `casadi/test_parity.py` (`Derivatives cannot be calculated for …` on the parent). Each was run against the parent to confirm it fails for its stated reason, with the control arms passing on both sides so the failure is attributable.
- [x] Book page under `docs/src/` updated — ~~not done~~; `docs/src/casadi.md` gains a finite-difference section (the two pattern sources and what each costs, the structure-without-values contract, `stats()["fd_hessian"]`, and the restoration scoping), and its stale claim that the plugin selects `limited-memory` "whenever an exact Lagrangian Hessian is not available — the `!exact_hessian_` branch" is corrected, since that branch no longer exists in that form.
- [x] `cargo fmt --all -- --check` clean; ~~full `cargo test` and `cargo clippy` not run~~ — **both now run.** 3747 workspace tests pass (`--exclude pounce-hsl`, as CI does), the CasADi plugin parity suite passes, clippy is clean in the changed files under the CI gate, and **CI is green on the head, 11/11.** Re-run after merging `origin/main` at `97344dd8`, which brought in #832 and #835 — both solver changes, so the earlier numbers did not carry.
- [x] `scripts/sweep-fixtures.sh` run — 156 fixture-legs, empty diff against `origin/main`, live corpus. Re-run against `97344dd8` after the merge, since #832 (capped bound multipliers) and #835 (convex cost normalization) are both trajectory-affecting and the previous run did not cover them. Still empty. See Blast radius.
- [x] Every claim in this body is true of the code as it stands now

**This body was written for an earlier state of the branch and has been corrected in place** rather than rewritten, so the struck-through text shows what moved. Two of its claims had become false: that the fixture sweep was outstanding, and that the finite-difference path worked through restoration unchanged. The second was a real defect, not just stale prose.

The one gap that remains is the **Ipopt baseline**, above.

https://claude.ai/code/session_01WDYhZvV7TRqL3ywWBQwpAz


srikanth-gm:
prof. @jkitchin thanks, partitioned quasi-newton not working properly is an interesting negative result. We previously tried passing an adjoint based FD Hessian to IPOPT using `exact` but our problem stalls with massive regularization and duals blowing up. 

This might however be a step-size issue or a modeling issue on our side if the non-smoothness isn't sufficiently regularized or not C2 (single sided springs, bumpstops, ground contact etc.).

FD Hessian inside the NLP solver is a unique option and worth having.

# Suggestion: tighten the FD pattern

*Use the structural information already available through the POUNCE CasADi `nlpsol` interface—the analytic Jacobian sparsity and CasADi-derived nonlinear-variable mask—to reduce unnecessary finite-difference probes and KKT fill, while separating Hessian-structure availability from Hessian-value availability.*

> **Disclaimer:** These suggestions are based on an AI-assisted reading of the code. I have not implemented, run, or benchmarked them, so the expected improvements—fewer probes, lower fill, and reduced memory—are hypotheses that still need measurement.

My scope here is specifically the POUNCE CasADi `nlpsol` plugin.

The plugin already has the two structural inputs needed for a tighter fallback:

1. `jacg_sp_`, obtained from the analytic `nlp_jac_g` function;
2. `nl_ex_`, obtained with

```cpp
oracle_.which_depends("x", {"f", "g"}, 2, false)
```

when `pass_nonlinear_variables=true`, or supplied explicitly through `nonlinear_variables`.

The nonlinear-variable set is currently sent through `IpoptSetNonlinearVariables`, but POUNCE only consumes it for the limited-memory updater. Could the finite-difference updater consume it as well?

If `N` is CasADi’s structurally derived set of variables on which `f` or `g` depends nonlinearly, then a variable outside `N` cannot participate in Lagrangian curvature. The Jacobian-derived constraint-Hessian pattern can therefore be tightened from

```text
⋃ᵣ supp(Jᵣ) ⊗ supp(Jᵣ)
```

to

```text
⋃ᵣ (supp(Jᵣ) ∩ N) ⊗ (supp(Jᵣ) ∩ N).
```

This remains a conservative superset rather than an exact Hessian graph, but for CasADi models with a large linear transcription block it could potentially:

- avoid perturbing columns CasADi certifies as linear;
- reduce the FD coloring and possibly the number of probes;
- avoid cancellation noise in Hessian entries known to be zero;
- reduce the Hessian contribution to the KKT pattern;
- reduce FERAL factor fill and memory.

The complete primal KKT diagonal can remain structurally present for barrier and inertia-regularization terms without requiring FD probes or nonzero Hessian values on masked-out variables.

### Separate Hessian structure from Hessian values

At present, `exact_hessian_` appears to be disabled only for:

```cpp
hessian_approximation == "limited-memory"
```

Consequently, selecting:

```python
{"pounce": {"hessian_approximation": "finite-difference"}}
```

still appears to enter the `exact_hessian_` initialization branch, create `nlp_hess_l`, obtain `hesslag_sp_`, install `cb_h`, and report nonzero Hessian storage through `CreateIpoptProblem`.

That differs from the motivating case of this PR: a model with an analytic gradient and Jacobian but no Hessian-value callback. It may also make a CasADi benchmark rely on CasADi’s generated exact-Hessian machinery rather than exercise the genuine first-derivative-only path.

Could the plugin distinguish these capabilities explicitly?

```text
use_exact_hessian_values
have_hessian_structure
```

For example:

- `exact`: construct and evaluate `nlp_hess_l`;
- `limited-memory`: do not construct or declare an exact Hessian;
- `finite-difference`: do not evaluate exact Hessian values, but optionally obtain and pass CasADi’s symbolic `hesslag_sp_` as structure-only metadata when it is available;
- `finite-difference` without Hessian structure: pass `jacg_sp_` and `nl_ex_`, then let POUNCE build the masked Jacobian-derived fallback.

This would give the two FD pattern modes clear CasADi semantics:

```text
fd_hessian_pattern=declared
    CasADi supplies hesslag_sp_, but POUNCE never requests Hessian values.

fd_hessian_pattern=jacobian
    POUNCE derives the pattern from jacg_sp_, pruned by nl_ex_ when available.
```

If CasADi cannot derive `hesslag_sp_` for a model backed by first-derivative-only callbacks, the second mode remains available and does not require second derivatives.

### Preserve objective curvature

`jacg_sp_` describes only the constraints. The nonlinear mask includes both `f` and `g`, but it does not identify pairwise objective interactions.

Pruning the constraint-row cliques with `nl_ex_` is therefore safe, but the fallback must still account for objective-only curvature. For the CasADi interface, the preferred source would be CasADi’s symbolic objective-Hessian sparsity when available. If only an objective nonlinear-variable set is available, its safe fallback is a clique over that set, although this may be conservative.

It may be useful to derive separate masks:

```cpp
objective_nl = oracle_.which_depends("x", {"f"}, 2, false);
global_nl    = oracle_.which_depends("x", {"f", "g"}, 2, false);
```

Then the conservative fallback is:

```text
objective_nl ⊗ objective_nl
∪
⋃ᵣ (supp(Jᵣ) ∩ global_nl) ⊗
    (supp(Jᵣ) ∩ global_nl).
```

If CasADi can provide the actual symbolic Lagrangian-Hessian sparsity, that should take precedence because it is much tighter than either clique construction.

### FERAL handoff

Once POUNCE publishes the inferred pattern as a sparse `SymTMatrix`, the standard KKT assembler already appears to copy those coordinates into the full KKT triplet and send them to FERAL through `initialize_structure`. I do not think a separate FERAL API change is required for ordinary sparsity-aware ordering.

The missing work appears to be upstream of FERAL:

1. ensure the CasADi plugin exposes the appropriate structure in FD mode;
2. pass CasADi’s nonlinear-variable subset to `FdHessianUpdater`;
3. prune the recovery and probe pattern while retaining the structural KKT diagonal;
4. ensure interpreted and generated CasADi paths preserve the same capability split.

It would also help to expose the selected FD pattern source, Hessian nonzero count, color count, and number of probed columns through `solver.stats()`, alongside the existing FERAL fill statistics. This would make the benefit measurable from the CasADi interface without requiring internal instrumentation.

jkitchin:
@srikanth-gm thank you — this found a real bug, and one that was costing more than I would have guessed. Everything below is measured against the code as it stands after `f098c53`.

## 1. Objective curvature — you were right, and it was a correctness bug

The Jacobian-derived pattern was built from constraint-row cliques alone. That covers the `Σ yⱼ ∇²cⱼ` term of `∇²L = ∇²f + Σ yⱼ ∇²cⱼ` and **nothing of `∇²f`**, so an objective coupling two variables that never share a constraint row put Hessian entries outside the pattern. The pattern was a *subset* of the truth, not the superset the mode's whole safety argument rests on — and I had written that safety argument into the module docs, the CHANGELOG and the PR body while the code did not honour it.

`laptime` could not expose it by objective value (its objective is minimise-final-time). It exposed it by **cost**, once fixed:

| | before | after |
|---|---|---|
| N=160 | 38 iterations | **30** — exactly the exact-Hessian path's count |
| N=320 | 106 iterations | **62** |

So the dropped curvature was real and material, not a corner case. Fixed by taking the objective's support from `get_objective_variables_linearity` (the same structural hook the partitioned path already used), with the first `∇f`'s nonzeros as a weaker value-derived fallback when a TNLP declines to state it. Pinned by `the_objective_clique_is_in_the_jacobian_derived_pattern`, on a fixture where no constraint row contains both objective variables.

## 2. Nonlinear-variable mask — implemented, and it does work

`⋃ⱼ (supp(∇gⱼ) ∩ N) ⊗ (supp(∇gⱼ) ∩ N)`, wired from the same `quasi_newton_nonlinear_vars` the limited-memory path uses. Sound for the reason you give: a variable linear in every `f` and `g` has structurally zero off-diagonal Hessian entries. On `laptime` it removes ≈53 000 pattern entries. The full primal diagonal stays in the pattern regardless, as you note it must.

## 3. The two pull against each other, and the net is not a free win

This is the part worth your attention, because it is the conservatism your comment flags:

| | before | after |
|---|---|---|
| pattern nonzeros | 146 267 | 143 854 |
| `rho_max` | 59 | **338** |
| probe groups | 76 | **341** |
| N=160 (no Hessian) | 38 it / 12.6 s | 30 it / **38.7 s** |
| N=320 (no Hessian) | 106 it / 430 s | 62 it / 430 s |

`laptime`'s objective is nonlinear in **320 variables**, so the conservative `O ⊗ O` clique is 320 wide and by itself explains `rho_max` 59 → 338. Iterations improved everywhere; N=160 wall got 3× worse.

**That changes the PR's headline and I have corrected it.** Against `limited-memory` on a genuinely Hessian-less model, the finite-difference path is now **1.4× wall at N=160** (38.7 s vs 53.7 s) rather than the 4.6× I claimed, while iterations go 207 → 30. At N=320 it is 1.6× wall and limited-memory does not converge at all. The method got better and the wall-clock margin got smaller.

Exactly as you say: a model that can state its true Lagrangian-Hessian sparsity should use `fd_hessian_pattern=declared` and pays none of this. Your suggestion to prefer CasADi's symbolic `hesslag_sp_` over either clique is the right ordering.

## 4. The CasADi plugin — your reading is correct, and there is a bit more

Confirmed at `casadi/casadi_nlpsol_pounce.cpp:468-472`: `exact_hessian_` is cleared only for `"limited-memory"`, so `finite-difference` still creates `nlp_hess_l`, obtains `hesslag_sp_`, reports `hesslag_sp_.nnz()` to `CreateIpoptProblem` and installs `cb_h`. One thing you did not mention: line 792 also *forces* `hessian_approximation=limited-memory` when `exact_hessian_` is false, before user options are forwarded. So the capability split you propose is needed in both directions.

**I have not fixed this**, deliberately. CasADi is not installed in the environment this branch was built in, so I cannot compile or test that change, and an unverified C++ edit is the failure mode this repo's PR template exists to catch. It should be a follow-up with a test rather than a blind edit here.

One constraint worth flagging for whoever picks it up: **structure-only pass-through is not currently expressible.** POUNCE derives `h_space` from a *successful* `eval_h` structure call, so passing `nnz_h > 0` with `cb_h = nullptr` does not hand it a declared pattern — the structure call fails and it falls back to the Jacobian derivation anyway. Making `finite-difference` + CasADi's `hesslag_sp_` work as you describe needs a C-interface addition (a structure-only Hessian hook), not just a plugin change. Worth doing — it is the difference between 17 probe groups and 341 on this model — but it is a larger piece than the gate fix.

## 5. FERAL handoff — agreed, no change needed

Confirmed: the assembled `SymTMatrix` goes through `StdAugSystemSolver`'s ordinary triplet path and reaches FERAL via `initialize_structure` like any other `W`. Nothing FERAL-side is required.

## 6. Stats — agreed, not done

Exposing the selected pattern source, Hessian nonzero count, colour count and probed-column count through `solver.stats()` is right and easy; today they are only reachable via `POUNCE_FD_HESSIAN_DEBUG`. Not in this PR.

## On your adjoint-FD-into-IPOPT experience

That failure mode is plausible here too and I would not want it to look ruled out. A forward difference is exact only to first order, so the recovered Hessian carries truncation error; if an adjoint-FD Hessian fed to IPOPT as `exact` produced blowing-up duals, the likely difference is that this path recovers a *structurally symmetric* matrix on a fixed pattern with a consistent `y`, rather than an independently-differenced object. It did not hurt on `laptime` — objectives agree with the exact path to 13 significant figures — but a central difference (double the probes) is the documented fallback if a model proves sensitive, and I would treat your experience as the reason to check that before trusting it on a new model.

---

jkitchin:
## `scripts/sweep-fixtures.sh`: run, empty diff

The body above says this gate is outstanding. It has now been run and it is clean — recorded in `dev-notes/fd-hessian-from-jacobian.md` as of `e201d11`.

Baseline `origin/main` (`a5e0a83`) against this branch (`ec78423`), **156 fixture-legs each**, both the `exact` and `lbfgs` legs:

```
diff sweep-base.txt sweep-new.txt   ->   (empty)
```

Not one line moves. Status, objective, iteration count and engine all unchanged.

**The corpus is live, not vacuously empty** — which is the failure mode that would make an empty diff worthless:

| | per leg |
|---|---|
| `cvx-qcqp` | 5 |
| `cvx-qp` | 36 |
| `nlp` | 37 |
| `NO_JSON` / `unknown` | **0** |

So the convex arms are covered, which is what `CLAUDE.md` requires of this sweep after gh#760. Statuses vary across both legs (`SolveSucceeded`, `InfeasibleProblemDetected`, `DivergingIterates`, `ErrorInStepComputation`, `RestorationFailed`), so the fixtures are genuinely exercising the solver rather than short-circuiting.

**The control that gives the empty diff its meaning.** An empty diff proves nothing if the two binaries are the same binary, so:

- checksums differ (`c14ed24a` vs `70773820`);
- the **baseline rejects** `hessian_approximation=finite-difference` as an invalid option; the **branch accepts** it.

The feature is therefore present and reachable in the swept binary and still moves nothing on the default legs. That is the actual claim — stronger than "nothing changed".

### What this does not cover

The sweep runs `exact` and `limited-memory` only, so **`finite-difference` itself is never swept**. This result is evidence that the new code does not perturb the paths existing users are on; it is not evidence about the new path. That rests on the `laptime` measurements in the body, and those do not exercise restoration — the N=80 run reports `restoration_calls: 0`, and `main` changed restoration under this branch in gh#815/gh#819. The FD-through-restoration interaction remains unmeasured.

Also now closed by CI on the merge commit, both listed as gaps in the body: full `cargo test --workspace` and `cargo clippy --workspace -D correctness -D suspicious` both pass (the `Test (ubuntu-latest)` job runs both, plus `fmt --check`, the release/docs consistency guards, and a `pounce-hsl` compile-check).

---

srikanth-gm:
# Report

I tested the solver-side finite-difference Hessian from commit `ec784238` on a representative large NLP, using 14 OpenMP threads.

**It does iterate and does not hang or crash, but it effectively stalls after the first step on this problem.**

| Metric | Iteration 0 | Iteration 1 | Iteration 10 |
|---|---:|---:|---:|
| Objective | 229.040 | 102.446 | 97.232 |
| Primal infeasibility | 10.6 | 6.23 | 5.79 |
| Dual infeasibility | 101 | 6.18e3 | 2.17e6 |
| `lg(rg)` | - | - | 4.7 |

After iteration 1, primal infeasibility improves by only about 7%, while dual infeasibility worsens by about 352x. From iterations 2-10, `lg(rg)` is 3.0-4.7, meaning POUNCE must add approximately `1e3`-`5e4` of primal regularization to obtain acceptable KKT inertia. Accepted primal steps are correspondingly small: `3.86e-4`-`3.91e-2`, with a median of `6.95e-3`.

The run is also expensive:

- 59,939 variables and 58,304 constraints
- 1,347,577 entries in the internal FD Hessian pattern
- 181 CPR color groups per Hessian rebuild
- 2,013 objective-gradient evaluations and 2,013 constraint-Jacobian evaluations in 10 iterations
- 2,184 seconds in POUNCE and 2,202 seconds total wall time
- zero Hessian-value callback evaluations, as expected in solver-side FD mode

For comparison, the corresponding L-BFGS run reached an optimal solution in 206 iterations with 207 objective-gradient evaluations, 207 constraint-Jacobian evaluations, 423.73 seconds in POUNCE, and 424.93 seconds total wall time. In these saved runs, the first 10 FD iterations took about 36.7 minutes, over five times the approximately 7.1 minutes needed for the entire L-BFGS solve. Normalized by iteration, this is approximately 220 seconds for FD versus 2.06 seconds for L-BFGS, or about 107x slower per iteration.

These wall times were measured on the same machine but were not a controlled back-to-back pair, so the exact ratio should be treated as indicative. The evaluation counts expose the structural difference independently of machine timing: FD required about 201 evaluations of each first-derivative callback per iteration, while L-BFGS required approximately one of each, a roughly 200x difference.

POUNCE differences the complete NLP Lagrangian gradient once per global color. This is a materially different workload from assembling curvature from locally differentiated blocks and explains the much larger cost.

The poor FD-Hessian quality, heavy regularization, convergence stall, and runtime reported above are specific to this problem and its derivative callbacks. They are not POUNCE issues to correct in this PR and should not be taken as evidence that solver-side FD is generally defective.

## Separate implementation findings

The following findings were reproduced on small independent problems and are separate from the problem-specific behavior above.

1. **FD negative-curvature escape can use a stale Hessian.**

   The current-iterate Hessian is refreshed before the negative-curvature check only when `provides_exact_hessian()` is true. The FD updater inherits `false`, so this check can use the previous iterate's matrix. A bounded polynomial reproducer confirmed the behavioral impact: exact mode refreshed the curvature and escaped a stationary maximum, while FD mode used the stale positive-curvature matrix and reported that maximum as optimal. FD needs a capability separate from "exact" to indicate that it can refresh `W` at the current `(x, y)`.

2. **The no-Hessian fallback can infer an incorrect objective sparsity pattern.**

   Through the public C interface, when no Hessian structure is supplied, objective support is inferred from objective-gradient values at the current point. For `f(x) = x0*x1`, starting at `(0, 0)` gives a zero gradient and drops the structurally nonzero cross term, even though `d2f/dx0dx1 = 1`. At `(1, 1)`, the same problem gets a different pattern. Structural sparsity must not depend on the starting values. The fallback should require an explicit pattern or use a conservative structural pattern.

3. **CasADi's `finite-difference` path still constructs `nlp_hess_l`.**

   Only `limited-memory` disables exact-Hessian construction in the adapter. `finite-difference` still creates the Hessian callback and passes its sparsity to POUNCE, although POUNCE never requests its values. This means solver-side FD may still require second-derivative graph construction and can incur its initialization cost. The intended contract should be clarified and tested with a model for which second derivatives cannot be constructed.

## Problem-specific follow-up

The recovered Hessian needs numerical validation before the poor behavior is attributed solely to true nonconvexity.

POUNCE uses the fixed one-sided step

`h_j = sqrt(eps) * max(1, abs(x_j))`,

which is about `1.49e-8` for normalized variables. Differencing a complete first-derivative callback at that scale can amplify callback variation by about `6.7e7`. The large regularization proves that the recovered matrix is not useful for this solve, but it does not distinguish genuine Lagrangian indefiniteness from problem-specific derivative noise or recovery error. The step-size contribution has not yet been isolated.

The appropriate follow-up is problem-side validation: compare selected `H*v` products, inertia, and required regularization at the same `(x, y)` for steps such as `1e-8`, `1e-7`, `1e-6`, and `1e-5` against an independently generated reference Hessian. This is not a requested POUNCE change unless that investigation demonstrates a general solver defect.

I did not find a chain-rule or scaling mismatch in the derivative interface: POUNCE receives the complete objective gradient and constraint Jacobian in solver coordinates and correctly holds the multipliers fixed while differencing the Lagrangian gradient.

**Conclusion:** the implementation executes on this large NLP, but its convergence stalls, dual feasibility diverges, and global FD is prohibitively expensive for this particular problem. Those observations are problem-specific and are not POUNCE defects to correct. Separately, the stale negative-curvature refresh and value-dependent objective sparsity are independently reproduced correctness issues.

## Logs

```
Number of nonzeros in inequality constraint Jacobian.:      203
Number of nonzeros in Lagrangian Hessian.............:  1341739

Total number of variables............................:    59939
Total number of equality constraints.................:    58271
Total number of inequality constraints...............:       33
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:       27
        inequality constraints with only upper bounds:        6

fd-hessian: FdStats { n: 59939, nnz: 1347577, groups: 181, rho_max: 176, compression: 0.0030197367323445503, coloring_fell_back: false }
iter      objective   inf_pr   inf_du lg(mu)    ||d|| lg(rg) alpha_du alpha_pr  ls
   0  2.2904024e+02 1.06e+01 1.01e+02   -1.0 0.00e+00      - 0.00e+00 0.00e+00   0
   1  1.0244576e+02 6.23e+00 6.18e+03   -1.0 8.68e+00      - 3.43e-01 5.04e-01f  1
   2  1.0239080e+02 6.23e+00 6.17e+03   -1.0 4.14e+00    4.0 7.68e-03 9.29e-04h  1
   3  1.0229546e+02 6.22e+00 6.16e+03   -1.0 4.44e+00    3.5 4.13e-02 1.62e-03h  1
   4  1.0148817e+02 6.13e+00 5.99e+03   -1.0 4.41e+00    3.0 9.59e-02 1.39e-02h  1
   5  1.0146716e+02 6.13e+00 1.74e+04   -1.0 4.46e+00    4.4 3.29e-02 3.86e-04h  1
   6  1.0105764e+02 6.08e+00 3.55e+04   -1.0 4.46e+00    3.9 1.46e-01 7.42e-03h  1
   7  1.0067954e+02 6.04e+00 3.82e+04   -1.0 4.39e+00    3.4 2.25e-01 6.95e-03h  1
   8  9.9220498e+01 5.88e+00 2.68e+05   -1.0 4.50e+00    3.8 9.91e-02 2.75e-02h  1
   9  9.9080624e+01 5.86e+00 7.20e+05   -1.0 5.30e+00    4.3 2.63e-01 2.96e-03h  1
iter      objective   inf_pr   inf_du lg(mu)    ||d|| lg(rg) alpha_du alpha_pr  ls
  10  9.7231722e+01 5.79e+00 2.17e+06   -1.0 4.55e+00    4.7 2.26e-01 3.91e-02h  1


Number of Iterations....: 10

Objective...............:   9.7231721804204653e+01    9.7231721804204653e+01
Dual infeasibility......:   2.1732753393391385e+06    2.1732753393391385e+06
Constraint violation....:   4.7771844078944135e+00    5.7915649687067958e+00
Variable bound violation:   0.0000000000000000e+00    0.0000000000000000e+00
Complementarity.........:   2.9482714372008246e+02    2.9482714372008246e+02
Overall NLP error.......:   2.1508277835563608e+03    2.1732753393391385e+06


Number of objective function evaluations             = 11
Number of objective gradient evaluations             = 2013
Number of equality constraint evaluations            = 11
Number of inequality constraint evaluations          = 11
Number of equality constraint Jacobian evaluations   = 2013
Number of inequality constraint Jacobian evaluations = 2013
Number of Lagrangian Hessian evaluations             = 0
Total seconds in POUNCE                              = 2184.052

EXIT: Maximum Number of Iterations Exceeded.

POUNCE 0.10.0: Maximum Number of Iterations Exceeded.
  nlp_solver  :   t_proc      (avg)   t_wall      (avg)    n_eval
callback_fun  | 620.00ms ( 56.36ms) 621.16ms ( 56.47ms)        11
       nlp_f  | 205.00ms ( 17.08ms) 205.59ms ( 17.13ms)        12
       nlp_g  |   3.09 s (237.69ms)   3.09 s (237.55ms)        13
  nlp_grad_f  |   2.13ks (  1.06 s)   2.13ks (  1.06 s)      2014
   nlp_jac_g  |  13.69 s (  6.79ms)  13.71 s (  6.80ms)      2015
       total  |   2.19ks (  2.19ks)   2.19ks (  2.19ks)         1
       ```

**Comment by jkitchin** [comment]
@srikanth-gm both correctness findings are real and both are fixed in `b0eff6c`. Your report also bounds a claim I had made too broadly, which I've corrected — more on that below, because I think you were more generous about it than the data warrants.

## Finding 2 — value-dependent objective sparsity. Confirmed, and it was unsound, not imprecise

You are right, and my own code comment understated it: it called the fallback "weaker," which framed a soundness bug as a precision tradeoff. Your `f(x) = x₀x₁` case is decisive — at `(0,0)` the gradient is `(x₁, x₀) = (0,0)`, the support comes back empty, and `∂²f/∂x₀∂x₁ = 1` is dropped. That makes the pattern a **subset** of the truth, which is the one property this mode may not violate. It is the same defect you caught on the constraint side in your first review, arriving a second time through the fallback path.

Every level is structural now:

```
stated objective linearity  →  else the nonlinear-variable set N  →  else all n
```

`N` cannot omit a variable the objective is nonlinear in, so it is a sound superset; all `n` is the only superset left when a model states nothing. Both are conservative and can be expensive, so `FdStats::objective_clique_widened` reports when one was taken — otherwise a wide clique looks like real objective density rather than missing metadata.

Pinned by `the_objective_fallback_is_structural_not_value_derived` and `the_objective_support_rule_is_a_function_of_structure_alone`. Mutation-checked: reintroducing the empty-support behaviour turns exactly those two red and nothing else.

## Finding 1 — stale Hessian in the negative-curvature escape. Confirmed

The gate was `provides_exact_hessian`, which conflates two different questions — *is this exact?* and *can this be re-evaluated at the current iterate?* The probe only ever needed the second. The old reasoning is sound for limited-memory (BFGS keeps `B` positive definite, so the probe declines at `δ_x = 0` and declining is correct) but that argument is about **definiteness**, not about "not exact", and FD is the case where the two come apart.

Added `HessianUpdater::hessian_at_current`. FD answers it by rebuilding — it can, being a pure function of `(x, y)` with no step history — while quasi-Newton updaters return `None` and keep the previous path. `data.w` is saved and restored around the rebuild, since it holds the previous iterate's `W` and the post-optimal sensitivity hook reads it.

**I could not reproduce the end-to-end symptom, and the test I added says so rather than implying coverage.** Neither fixture here reaches the branch:

- `nonconvex_qp.nl` — objective `x₀·x₁`, whose Hessian is the *constant* `[[0,1],[1,0]]`. Stale and fresh are the same matrix, so the two paths cannot diverge. (Its objective is also, coincidentally, your finding-2 function.)
- `nonconvex_two_escapes.nl` — quartic, so curvature does vary, but it gives identical results either side of the fix.

**Could you share the bounded polynomial reproducer?** I'd like to land it as a fixture. Reproducing it needs curvature that varies between consecutive iterates *and* an earlier iterate that looks positive-definite, and I'd rather have your model than approximate it.

## Your 60k result — this bounds a claim of mine, and I don't want to accept the generous reading

You wrote that the stall and cost are problem-specific and not POUNCE defects to correct. For the stall and the dual-infeasibility divergence I agree. But your numbers do refute something I asserted, and I've corrected it rather than leaving it standing:

I argued the mode was affordable from `laptime`'s `rho_max = 15` and 17 colour groups, and said the probe count "does not grow with the mesh". That is true of `laptime` and is a property of its **per-stage stencil**, not of the method. Your model has `rho_max = 176` and 181 groups, hence your ~201 first-derivative evaluations per iteration. That cost is *inherent to the approach at that Hessian bandwidth* — it is not incidental to your callbacks. So the honest statement is:

> Probe cost scales with the Hessian's **row width**, not with the model's size, and can exceed limited-memory's per-iteration cost by two orders of magnitude. Check `rho_max` under `POUNCE_FD_HESSIAN_DEBUG` before reaching for this mode.

That now replaces the affordability claim in the module docs and the CHANGELOG. Your run is the counterexample that makes the qualification necessary, so thank you for it.

On truncation: your `1.49e-8` amplification concern is the right one and I would not want it to look ruled out. `FD_REL_STEP` is a fixed one-sided `sqrt(eps)·max(1,|xⱼ|)`, so differencing a complete first-derivative callback does amplify callback variation by ~`6.7e7`. Your proposed step sweep against an independently generated reference Hessian is exactly the right experiment, and a central difference (double the probes) is the documented fallback. I have not made the step configurable — say the word if that would help the investigation and I'll add it.

## Finding 3 — CasADi

Unchanged and still a follow-up, for the reason in my earlier reply: CasADi is not installed in this build environment, so I can neither compile nor test that change here. Note the additional constraint I mentioned — structure-only pass-through is not currently expressible, because POUNCE derives `h_space` from a *successful* `eval_h` structure call, so the capability split needs a C-interface addition rather than only a plugin edit.

## Effect on this PR's numbers

`laptime` N=160 is unchanged: 30 iterations, objective `65.37110679404918`. Wall moves 38.7 s → 42.3 s, which is one extra Hessian build at the convergence check — the measurable cost of finding 1's fix, and I'd rather pay it than keep the stale matrix. `objective_clique_widened` reads `false` there, since `laptime` states its objective linearity and so never touches the conservative fallback.

The CHANGELOG also still carried the pre-review iteration counts (38/106); those are now 30/62.

---

**Comment by jkitchin** [comment]
@srikanth-gm — **finding 3 is fixed and merged** (#823, `d194d9ec`). CasADi is installed in this environment now, so the blocker in my last reply is gone. Two things I told you are superseded, and both were wrong in your favour.

## Finding 3 was a hard failure, not an initialization cost

You wrote that `finite-difference` "may still require second-derivative graph construction and can incur its initialization cost." That understates it, and the understatement was mine to correct — it is what my earlier reply implied by describing the gate as wasteful.

On a model whose first derivatives are the last ones CasADi can produce — an external `Callback`, an FMU, a `DaeBuilder` transcription — `create_function("nlp_hess_l", …)` *throws*. So the mode did not pay an extra cost; it failed outright, with the same error `exact` correctly gives:

| mode | model with no constructible second derivative |
|---|---|
| `exact` | fails — correct, no Hessian exists |
| `limited-memory` | solves |
| `finite-difference` | **failed, same error as `exact`** |

The mode built to need no second derivatives was unusable on precisely the class it exists for. Your closing line on that finding — *"should be clarified and tested with a model for which second derivatives cannot be constructed"* — is now literally the regression test (`test_finite_difference_does_not_need_second_derivatives`), and it fails on the parent commit with `Derivatives cannot be calculated for …`.

The capability split is spelled as you proposed it in your first review: `exact_hessian_` means only "POUNCE may call `cb_h` for **values**", `hessian_structure_` means "the plugin can declare a **pattern**". The forced `limited-memory` you spotted at the old line 792 is keyed on the mode now; it was latent rather than active, since the option-forwarding loop re-sends the user's own value afterwards — but a defect masked by statement order is still a defect.

## The C-interface addition I said was needed — isn't

I told you structure-only pass-through needed a C-interface change, because POUNCE derives `h_space` from a *successful* `eval_h` structure call, so `nnz_h > 0` with `cb_h = nullptr` hands it nothing. Right about the mechanism, wrong about the conclusion: the plugin passes `nnz_h > 0` with a **live** `cb_h` that answers the structure request and **refuses** a values one. The existing C interface already supports that. No new API, no `h_space` change.

The refusal is also the evidence. `cb_h` returns false for any values request under `finite-difference`, so a completed solve *is* the proof that every number was recovered by probing — and if POUNCE ever did ask, it fails loudly rather than quietly reverting to the exact Hessian. Cross-checked against the unmodified plugin, which serves values: bit-identical trajectories on every model tried, restoration included, which is what says nothing ever asked. Your "zero Hessian-value callback evaluations, as expected" is now enforced rather than observed.

Both modes you named work from CasADi:

```
fd_hessian_pattern=declared   CasADi's hesslag_sp_, structure only, values never requested
fd_hessian_pattern=jacobian   derived from jacg_sp_, symbolic Hessian never built
```

`declared` falls back to `jacobian` by itself where CasADi cannot build a Hessian, so the default is safe on every model. The same split is reproduced in the generated C, whose emitted `eval_h` serves the pattern as a baked-in literal — so that block needs no `nlp_hess_l` dependency at all — and refuses values.

## Stats — your item 5 from the first review

`stats()["fd_hessian"]`, and `GetPounceFdHessianStats` in the C API. A real output, from a 3-variable model:

```python
{'pattern': 'declared', 'nnz': 4, 'n': 3, 'groups': 2, 'rho_max': 2,
 'coloring_fell_back': False, 'objective_clique_widened': False}
```

Absent, not zero, on every other mode. Three properties worth knowing:

- `pattern` is the source the run **ended up with**, not the one requested. Asking for `declared` on a model with no Hessian reports `jacobian` — the case you actually want to detect.
- `objective_clique_widened` says *why* `groups` is high: a widened clique is a missing `get_objective_variables_linearity`, which the model can fix, as distinct from a genuinely dense objective, which it cannot.
- `n` is there so `groups / n` is derivable from the report alone — that was your "number of probed columns", which `groups` on its own could not give.

This is the number your 60k report needed and could only get from `POUNCE_FD_HESSIAN_DEBUG`. `rho_max = 176`, 181 groups is now a two-line check before committing to a solve.

## On `rho_max` — a floor worth naming

@jkitchin has already corrected the affordability claim your run refuted, and I won't restate it. One diagnosis to add, because it bounds what tuning can achieve.

Where no Hessian structure is available, `∇²f` has to be bounded by a **dense clique** over the objective's nonlinear variables. On `laptime` that objective is a control-rate regulariser over 320 controls, so its true `∇²f` is banded with a row width of 3 while the clique is 320 wide — and since CPR needs at least `rho_max` groups, the clique alone floors the probe count at 320. Measured 338/341 against the declared pattern's 15/17. It is 36% of the pattern's entries and 100% of the reason the colouring cannot narrow.

So on a model that declares no second-derivative structure, the probe count is not a tunable — it is the price of not knowing `∇²f`. What removes it is any declared structure at all, which is what `declared` uses and what the fix above finally makes reachable from CasADi.

Your first review proposed refining that clique with the objective's own Hessian sparsity. **I built it and then removed it**, and the reason is worth passing on: the only way to lose the symbolic `∇²L` in CasADi is an opaque callback, and CasADi treats an opaque callback's Jacobian as *dense* — measured 3480/3480 nonzeros on a model whose declared Jacobian was banded-3. A dense `J` makes `JᵀJ` dense, so the constraint half is dense regardless and the objective refinement buys nothing. I could not construct a model on that surface where it pays. Recorded with the measurement in `dev-notes/fd-hessian-from-jacobian.md` rather than left as folklore.

## Where that leaves your report

Findings 1 and 2 were fixed in `b0eff6c`; finding 3 is fixed in `d194d9ec`. All three of your implementation findings are on `main`. Your 60k stall, the dual-infeasibility divergence and the step-size question remain open and problem-specific, exactly as you framed them — and @jkitchin's request for the bounded polynomial reproducer still stands, since neither in-repo fixture reaches that branch.

Thank you for the second pass. The first found a correctness bug; the second found an unusable mode, and it only surfaced because you kept asking about the gate after the first answer was "not fixed, deliberately".

---

PR #824, opened 2026-08-28
sens: evaluate the corrector's operator at the predicted point

devin-griff:
## The change

The corrector already paid a factorization whenever the predictor's step changed the active set, applying the change to the barrier diagonal and factoring the result. But the matrices being factored were the base point's: the Hessian the algorithm last stored, and the base point's Jacobians and bound quantities. A chord iteration contracts at the rate set by the distance between its operator and the true Jacobian, and the true Jacobian is the one at the predicted point, so evaluating the derivatives there before the factorization that was already being paid improves the convergence rate by an order for the cost of one evaluation. On the 62k collocation model that evaluation measures 42 to 46 ms against a factorization comparable to a refined back-solve.

Every correction now evaluates at the predicted point and factors there, unconditionally, and every block of the operator lives in that one frame. The current iterate is swapped to the predicted point for the correction, so the constraint Jacobians, the bound slacks and multipliers, and the barrier diagonal all evaluate at the stepped point, with the step's own multipliers under the kappa-sigma clamp, and under an exact-Hessian solve the Lagrangian Hessian is re-evaluated there. The predictor's active set is applied to the diagonal in the same frame, and the `s` block reaches the same frame through `solve_with_sigma`'s fallback, so general inequality constraints and variable bounds agree on where the operator lives. On corrections with no active-set change, which previously back-solved against the held factor, this adds the factorization as well as the evaluation. A `limited-memory` solve keeps its quasi-Newton matrix, captured as a `ConvergedState` fact the way `bound_relax_factor` is. The diagonal is always a fresh object, because the factorization cache keys on its tag and an already-cached vector could resolve to a factor built at another iterate.

The two cases that store a base-point diagonal for the held factor's own back-solves are no exception: a solve the sigma ceiling (gh #737) touched, and one that crossed over into the declared frame (gh #654). The corrector consults neither stored copy. It re-derives the ceiling and the frame rule at the predicted point, for both diagonal blocks. Measured on a fixture welding the gh #737 cap trigger to a curvature workload, the correction went from no progress at any budget under the stored copy to matching the no-ceiling control's first iteration, 1.5e-1 to 1.0e-2, and the plateau past that first iteration belongs to the fixture's non-unique multiplier split rather than to any diagonal choice, measured identical under a forced raw diagonal. A crossover arm pins the same for the declared frame, and on #828's fixture the rebuild is a no-op at every point of the sweep, dead zone included.

One alternative was built, measured, and discarded. It kept the barrier diagonal at the base point while the other blocks moved. On the double column that operator's first iteration fails to reduce the residual, the no-improvement rule stops the correction, and the estimate does not move at any perturbation size, corrector budget, or mode. The middle column of the table below is that variant.

## Measured on the double column

The protocol is the 2026-08-18 corrector study's: the N=25 Radau column, 62,167 variables, 792 weakly active bounds at the base solve, one fixed random direction over the 246 initial-state parameters with each entry scaled by its distance to the nearer bound, truth a warm re-solve at tol 1e-10, error the inf norm over variables away from the weakly active set. The predictor rows are bit-identical on every build. The floor near 6.7e-7 is the offset between the barrier solution at the held `mu` and that truth, the base solve's own solution quality.

| perturbation | predictor error | main | base-point diagonal, other blocks moved | this PR |
|---|---|---|---|---|
| 1 percent | 2.826e-3 | 6.7e-7 in 1 iteration | 2.826e-3 at every budget | 6.650e-7 in 1 iteration |
| 5 percent | 7.065e-2 | floor in 3 iterations (2026-08-18 study) | 7.065e-2 at every budget | 6.788e-7 in 1 iteration |
| 10 percent | 2.826e-1 | floor in 3 iterations (2026-08-18 study) | 2.826e-1 at every budget | 6.801e-7 in 2 iterations |

The main column's 1 percent row was run through `estimate()` on this machine; its 5 and 10 percent rows are the study's held-factor chord, measured through the crate harness on the same protocol.

## Per step on the small fixtures, against a tol 1e-10 re-solve

| fixture | plain step | main, budgets 1/2/4/8 | this PR |
|---|---|---|---|
| curvature, no bounds | 4.1e-2 | 6.9e-3 / 4.8e-4 / 1.5e-5 / 8.6e-9 | 5.3e-4 / 6.4e-6 / 6.9e-10 / 7.2e-11 |
| strongly active bound, curvature elsewhere | 9.0e-1 | 3.5 / 2.0 / 6.6e-1 / 6.6e-1 | 3.1e-1 / 1.6e-1 / 5.2e-2 / 6.2e-3 |
| near-bound coordinate | 4.1e-1 | 9.1e-1 at every budget | 1.3e-1 / 5.8e-2 / 1.4e-2 / 1.0e-3 |

On the oracle fixtures the holding side of a kink lands on the re-solve, 5.5e-2 to 9.1e-12, and a bound the step carries a coordinate onto is reached instead of achieved nothing.

## Releases the step's endpoint does not show

A bound that must leave the active set, on a step whose endpoint holds it, gets no released row. The step's clamped multiplier leaves a weak diagonal entry at that bound instead. What the iterations do with it is delta-dependent. Just past the breakpoint they decline, and the no-improvement warning fires as before. Deeper past it they move the variable partway off the bound and reduce the residual they measure, landing closer than the plain step on some fixtures and farther on others, and never at the re-solve. `fix_relax` and `path` decide such releases where the step shows them, and the corrector then applies the released row and crosses exactly.

## Closed loop, the asnmpc benchmark, both builds on one machine

The benchmark reproduces the notebook's stored numbers to every printed digit, with one movement. On the constraint_switching campaign the corrected policies' IAE now equals the full re-solve's own to all printed digits, where it differed in the sixth decimal. The guarded policy's fallback counts are unchanged in every campaign, cell 35's agreement assert passes at its stored 7.9e-7, and the notebook is untouched.

## Tests

`corrector_predicted_operator.rs` is the new regression test. Its fixture's Hessian doubles over the step, six iterations on the base operator leave 1.6e-3 of error, the predicted operator clears 1e-6, and the trace shows the evaluation at the stepped point exactly once per correction. It fails on the parent commit.

The release fixtures are repinned deliberately, each carrying a note naming its intended update signal, a corrector that revises the active set inside its own loop. The oracle's release pin, renamed `the_corrector_stays_short_of_the_resolve_across_a_release`, owns the two facts that hold at every measured delta: the corrected answer is not the re-solve, and fix_relax reaches it. `improved()` is measured rather than asserted there, since it follows the delta-dependent gradation above. `corrector.rs`'s `a_release_the_step_hides_is_followed_partway` pins the partial move on a curvature fixture, off the bound and short of the released answer. On the pyomo side, `test_a_release_the_step_hides_is_corrected_partway` pins that no achieves-nothing warning fires where the residual genuinely falls, and the modes-disagree test keeps `fix_relax` and `path` converging while the linear-step correction stays short. The oracle's module header records that `leg_oracle_improved_means_closer_to_the_resolve` now exercises both of its arms, and the worked examples in `CLAUDE.md` and the sens-review checklist cite the renamed pin.

`corrector_ceiling.rs` pins the two frame cases: the ceiling-engaged correction acts instead of stalling, the identical workload without the degenerate bound converges to the floor as the control, and a crossover arm covers the declared frame. The predicted-operator regression fixture gains an `eval_h` counter and a limited-memory arm: `a_limited_memory_solve_corrects_without_an_exact_hessian` pins that such a solve and its correction request zero exact-Hessian evaluations, while the correction still takes the predictor's 9.45e-2 curvature error to 4.0e-4 at budget 8 against the held quasi-Newton matrix.

254 tests pass on `pounce-sensitivity` and 473 on `pyomo-pounce`. `cargo fmt --all -- --check` is clean, and clippy is clean in the changed files under the CI gate's lints.

---

- [x] Tests fail on the parent commit for the stated reason: `corrector_predicted_operator.rs` fails on `f677f244` at 1.6e-3 against the 1e-6 bar
- [x] Book page: the corrector sections of `docs/src/sensitivity.md`, the `estimate` and `correct_step` docstrings, and the worked examples in `CLAUDE.md` and `.claude/commands/sens-review.md` updated to the single frame
- [x] The asnmpc notebook is untouched, and its cell-35 assert passes at the stored value


jkitchin (commented):
## `/sens-review` pass on the predicted-operator change

Ran the checklist against the diff, read the call chain down into `pounce-algorithm`, and built and ran the suite. **All 250 `pounce-sensitivity` tests pass**, including the new `the_corrector_closes_curvature_the_base_operator_cannot`, all 16 invariance legs, and all 10 oracle legs. The core idea is right and the measurement discipline is up to this repo's standard — the parent-commit failure and the per-fixture error tables are exactly the evidence CLAUDE.md asks for on a trajectory change.

| 1 | index space | PASS | no new cross-space read; `released`/`pinned` index spaces unchanged |
| 2 | frame and scaling | **RISK** | the `s`-block barrier diagonal moves to the predicted point; the `x`-block deliberately does not |
| 3 | absolute thresholds | N/A | no constant introduced or moved |
| 4 | doc drift | **RISK** | four stale contracts, two of them outside the diff |
| 5 | silently wrong / success signal | PASS | oracle row updated; the `improved()==false` arm is now genuinely exercised |
| 6 | which branch does the fixture take | **RISK** | `exact_hessian == false` has no fixture anywhere |
| 7 | measured populations | PASS | the discarded recomputed-diagonal variant is measured (0.129 vs 0.035), not argued |
| 8 | which binary | PASS | Rust-side numbers; no `POUNCE_SKIP_EXT_STALE_CHECK` in evidence |

---

### RISK 1 — the `s`-block diagonal silently moves to the predicted point

The central design decision here is that the barrier diagonal is the one exception that stays at the held solve, and it is measured and justified. That holds for the `x` block. It does not hold for `s`:

- `solve_released_scaled` passes `s: self.sigma_override().s` (`algorithm_backsolver.rs:775`)
- `sigma_override().s` is `self.sigma.s`, which is `None` on an ordinary interior solve — `EffectiveSigma` only populates it when `declared` is present or the gh#737 cap actually bites (`:379`)
- `solve_with_sigma` then does `sigma_override.s.unwrap_or_else(|| cq_ref.curr_sigma_s())` (`pd_full_space_solver.rs:549`)
- `curr_sigma_s()` reads `self.curr_iv()` — which is now the **predicted** point (`ipopt_cq.rs:928`)

So `sigma_s = v/s_slack` is rebuilt from the predicted slacks, with no cap and no `pinned_entry` treatment — precisely the failure mode the `x` block is protected against, and the one measured at 0.129-against-0.035 before that variant was discarded.

Three places state the contract unqualified and are wrong for `s`: the CHANGELOG's "the barrier diagonal carried from the held solve", the same sentence in `correct_step`'s docstring, and `docs/src/sensitivity.md`'s "never one recomputed from the predicted slacks."

**Why no test sees it:** every corrector fixture is equality-only, so the `s` block is empty. `ExpCurvature` sets `g_l[0] = g_u[0] = 1.0`; the oracle's `KinkTnlp` and `ReleaseTnlp` both set `g_l[0] = g_u[0] = p_nominal`; the pyomo fixtures use `==` constraints with variable bounds only.

**Why it matters for the asnmpc story:** the PR attributes the guard-fallback change entirely to "the corrector now converges quickly to its own target" and asks @bernalde to recalibrate. But a path-constrained CSTR is exactly a model with a live `s` block, so there is a second, undescribed mechanism inside the same change. Worth resolving before a downstream author retunes thresholds against it.

**Cheapest thing that settles it:** a fixture with one general inequality whose slack the step drives toward zero, plus either passing a held `sigma_s` explicitly or documenting the asymmetry.

### RISK 2 — `exact_hessian == false` has no fixture

`ConvergedState::exact_hessian` is a new two-way discriminator, and only the `true` side is reached by any test — no `limited-memory` appears anywhere under `crates/pounce-sensitivity/tests/` or `pyomo-pounce/tests/`.

This is not a hypothetical gap in this repo. CLAUDE.md: *"the corpus was exact-Hessian only, and that gap shipped gh#677."* And the branch is newly load-bearing — under limited-memory the correction now forces a **refactorization** against the L-BFGS `w` at a *moved* iterate, where previously (no active-set change) it was a pure back-solve against the held factor. That combination has never been exercised.

The option plumbing itself checks out: `pounce-py` sets `hessian_approximation = limited-memory` explicitly when `!self.has_hessian` (`problem.rs:985`), so the captured flag matches how the solve actually ran rather than diverging from an internal fallback.

### RISK 3 — doc drift, including two files outside the diff

- **`correct_step`'s own summary line**, three lines above the paragraph this PR rewrote, still reads *"Each iteration costs one back-solve against the held factor and no factorization."* That is now the opposite of what the function does.
- **`max_iter = 0` "costs one evaluation, no back-solve"** — it now also costs a Hessian evaluation (the 42–46 ms measured here on the 62k model), a Jacobian re-evaluation, and a `set_curr`/restore. Still no back-solve, but this is the sentence a caller budgets against.
- **`CLAUDE.md:268` and `.claude/commands/sens-review.md:253-256`** both cite `the_corrector_reports_improved_without_crossing_a_release` **by name**, with its `3e-8` / `0.1333` numbers, as the worked example for checklist entry 5. This PR renames that test and inverts its assertion, so both documents now point at a symbol that no longer exists. Entry 4 is literally about checklists whose examples cannot be checked.
- **`sens_resolve_oracle.rs`'s own module header** still lists `improved() == false` under "What this file is NOT evidence about": *"No measured point on either fixture reaches it, so the arm … is written but unexercised. A fixture whose corrector declines to improve is the missing one."* This PR makes that the live branch. The header should record that the gap closed — it is good news for the file.

### Two smaller notes

**The fresh-sigma comment's stated reason is not the operative one.** The comment says a fresh tag *"is what makes the factorization below actually happen at the predicted point."* The cache keys on **13** dependency tags (`pd_full_space_solver.rs:604-618`), and `j_c`, `j_d`, `z_l`, `z_u`, `slack_x_l/u` are all rebuilt from the swapped `curr`, so the miss is already forced without touching sigma. Harmless, but it is a load-bearing comment that will mislead the next reader.

Separately, `fresh_barrier_sigma_x()` returns `None` on a non-`DenseVector` diagonal, and the new `.ok_or_else(...)?` turns that into a hard `SensComputationFailed`. On the no-active-set-change path that previously routed through `bs.solve()`, which needs no downcast — so this narrows where `correct_step` works at all. A fallback to `bs.solve` when the downcast fails would keep it.

**The `improved()==false` arm was weakened as it became live.** `(after - before).abs() <= before*1e-6 + noise` → `after <= before + noise`. The reasoning is plausible, but it moves from pinning a value to pinning a direction at exactly the moment the branch first executes. Relatedly, the test comment now says the returned step is "the caller's own put back into the box" — correct, and always was — while the public docstring and `sens.py` still say "the caller's own" / "the uncorrected step". Worth fixing the authoritative copies in the PR that noticed.

`test_guard_accepts_nominal_updates_and_falls_back_under_stress` asserts `fallback_full_solves == 0` on nominal, which is what this PR reports moving 0→1. It passes because the test runs `steps=6` with `horizon_intervals=8`, not the 30-step campaign — so the in-repo guard does not reach the regime measured here. Not a failure, just worth knowing that assert is not what is protecting you.

---

**Bottom line:** the mechanism is sound, the tests are green, and the trajectory evidence meets the bar. RISK 1 is the one I would want answered before merge — a real asymmetry against the PR's own stated design, invisible to the entire fixture corpus, landing on the same model class whose behaviour change the PR asks a downstream author to absorb. RISK 2 and the doc items are cheap to close.

---

jkitchin (commented):
## Follow-up on `326fdd7` / `8e36f0d`

Re-reviewed at `b9b27c0`. CI is green on all 11 checks.

**RISK 1 is resolved, and the resolution is better than what I proposed.** Moving the diagonal construction below the swap is the right cut, and the double-column measurement settles the direction far more convincingly than my flag did — I suggested holding `sigma_s` at the base point to match `sigma_x`, and the measurement shows that mixed operator makes *no progress at any budget*, while the single frame reaches the base solve's own solution quality in one back-solve. That the asnmpc guard-fallback movement and the cell-35 drift both disappeared under the single frame is good corroboration that the mixed frame was the mechanism, not the convergence story the original body proposed.

**RISK 3 is resolved.** Verified: `the_corrector_stays_short_of_the_resolve_across_a_release` exists at `sens_resolve_oracle.rs:1190`; `CLAUDE.md:270` and `.claude/commands/sens-review.md:253` both cite it by its current name; the oracle header's `improved() == false` bullet now records that both arms are exercised; and the `correct_step` summary line states the evaluation and factorization.

---

### New: the single frame has a door back to the mixed operator

`barrier_sigma_x()` short-circuits on a frozen override before it ever reaches the predicted point:

```rust
pub(crate) fn barrier_sigma_x(&self) -> Rc<dyn pounce_linalg::Vector> {
    match self.sigma.x.as_ref() {
        Some(v) => Rc::clone(v),          // frozen, built at the base point
        None => self.cq.borrow().curr_sigma_x(),
    }
}
```

Both `fresh_barrier_sigma_x()` and `active_set_sigma_x()` start from it, so moving them below the swap only reaches the predicted point when `self.sigma.x` is `None`. It is `Some` in two cases (`EffectiveSigma::x = cap_sigma(...).or_else(|| declared.map(...))`, `:379`): the gh#737 ceiling capped at least one entry, or the iterate crossed over into the gh#654 declared frame. In either case the diagonal is the base point's while the Hessian, Jacobians, slacks and multipliers are the predicted point's — reconstructing exactly the operator `326fdd7`'s own commit message measured as making no progress at any knob or budget.

**Measured, not inferred.** I instrumented the two branches and ran the crate suite:

- Across the whole corrector corpus (`sens_resolve_oracle` + `corrector_predicted_operator`): **84 of 84 calls take the live branch.** No corrector fixture reaches the frozen one.
- Across the full crate suite the frozen branch **is** reached, by fixtures in `tests/issue_737_pinned_to_bound_sigma.rs` (`the_barrier_diagonal_stays_under_the_ceiling`, `the_directional_step_reaches_the_same_answer`) and `tests/crossover_sigma_downstream.rs` (`the_reduced_hessian_error_tracks_one_over_sigma`).

So both doors are live on ordinary in-repo fixtures — the ceiling one is not exotic; `sigma_pin_cap` bites at `Σ > ~7.0e13` for a unit Jacobian coefficient and falls as `a²`, so a strongly active bound on a small-coefficient row reaches it well before anything looks pathological. Neither of those fixtures runs the corrector, which is precisely why the corpus is green.

I don't think the fix is simply "rebuild it at the predicted point" — the declared frame exists so `Σ` is measured against declared rather than relaxed bounds (gh#654), and the ceiling is a representability property of the row. Re-deriving either at the predicted iterate is a design call rather than a one-liner, which is why I'm raising it rather than proposing a patch. The cheapest thing that would decide it is a corrector fixture whose base solve trips the ceiling: if the correction stalls there with the middle-column signature, that names the scope.

Worth noting separately that `cap_x` is captured once at construction, so even on the live path the predicted-point entries are held under a base-point ceiling.

### Still open from the first pass: RISK 2

`exact_hessian == false` still has no fixture — no `limited-memory` appears anywhere under `crates/pounce-sensitivity/tests/` or `pyomo-pounce/tests/` on this head. The branch is load-bearing now in a way it wasn't before: under limited-memory the correction forces a refactorization against the L-BFGS `w` at a *moved* iterate, where previously it was a pure back-solve against the held factor. Given gh#677 shipped through an exact-Hessian-only corpus, one fixture solving with `hessian_approximation=limited-memory` and running a correction would close the dimension cheaply.

---

jkitchin (commented):
## Correction: I built the fixture, and my previous finding is wrong

I claimed the frozen-override path "reconstructs exactly the operator `326fdd7` measured as making no progress at any knob or budget." **That is disproved.** I built the deciding fixture and measured it; the frozen diagonal is benign, and this PR is strictly better than `main` at every point I measured.

### The fixture

```
min exp(x0) - p*x0 + 0.5*(x1 - (2.5-2p))^2 + 0.5*(x2+5)^2
s.t. p = p_nom,  A*x2 + x3 = 0,  x1 >= 0, x2 >= 0
```

Closed form `x0 = ln p, x1 = max(0, 2.5-2p), x2 = x3 = 0`, so the truth is independent of the machinery. `x2` sits hard on its bound (`Σ ≈ 2.5e10`), and `A` sets its gh#737 ceiling, so sweeping `A` walks the cap in and out. `x1` is a near-bound row whose slack shrinks 5× over the step; `x0` carries the curvature. Perturbation `delta = 0.2` on the pin, error is the inf-norm against the closed form.

### Measured, budget 8

| `A` | ceiling bites | frozen diagonal in use | `main` | this PR |
|---|---|---|---|---|
| 1e-4 | yes | yes | 1.7678e-2 (no progress) | 1.7678e-2 (no progress) |
| 1e-3 | yes | yes | 1.7678e-2 (no progress) | 1.7678e-2 (no progress) |
| 1e-2 | yes | yes | 1.7678e-2 (no progress) | 1.7678e-2 (no progress) |
| 1e-1 | yes | **yes** | **1.7678e-2 (no progress)** | **1.0000e-10** |
| 1.0 | no | no | 4.7226e-8 | 1.0000e-10 |

Predictor error is 1.7678e-2 throughout.

**The `A = 1e-1` row is the one that refutes me.** The frozen base-point diagonal is in use there — I confirmed it by instrumenting the branch, 4 calls per case at every `A ≤ 1e-1` — and the correction still reaches `1.0e-10`, identical to the live-diagonal case at `A = 1.0`, while `main` makes no progress at all. So a frozen diagonal does not reconstruct the dead operator. In hindsight the reason is visible in the code I was reading: `active_set_sigma_x` overwrites the released and pinned entries from the predicted point regardless, so freezing only affects rows whose slack barely moved — which is exactly where base and predicted agree anyway.

My structural observations stand (the override is live in-repo; no corrector fixture reaches it), but they were load-bearing for a conclusion that measurement does not support. Withdrawn.

### What the sweep did turn up — pre-existing, not this PR's

The dead zone at `A ≤ 1e-2` is real and reproduces **identically on `main`**. The initial barrier residual at the predicted point scales as `A⁻²` — `1.276e1, 1.771e3, 1.776e5, 1.776e7` — and the corrector stops after one iteration with `improved() == false` every time. That residual is measured at the clamped predicted point before any factorization, so it is the *predictor's* multiplier block, not the corrector's operator: when the gh#737 ceiling bites hard, the parametric step's duals come back badly scaled and the correction has nothing it can do.

Toggling only the ceiling on this PR's head, model unchanged at `A = 1e-4`, moves the initial residual from `1.776e7` to `2.140e-2` and the budget-8 error from `1.7678e-2` to `1.0000e-10` — so the ceiling is the cause, not the model's conditioning.

That looks worth its own issue against `main`. It is orthogonal to this PR, which neither causes it nor is blocked by it.

### Net

On this evidence the change is strictly better than `main` everywhere I measured, including where my objection lived. The only item I would still raise from either pass is the missing `limited-memory` fixture, and that is a coverage gap rather than a defect. Sorry for the noise on the previous round.

---

devin-griff:
## The frozen diagonal: benign where you measured, fatal one class over, and we think not using it is the right fix anyway

Thanks for building the deciding fixture, and for the correction. Your `A = 1e-1` measurement is right, and we reproduce it. But the sentence in your hindsight explanation is the whole story: freezing only affects rows whose slack moved. Your fixture's capped variable sits hard on its bound and stays there, so the frozen and predicted entries agree and the copy is harmless. The class where they do not agree is gh#737's own: a variable held between a binding equality and its bound, base slack exactly zero, predicted slack the put-back margin. There the stored entry is `7.0e13` against a predicted-frame value near `2.5e10`, and the correction dies.

We measured that case with everything upstream held fixed, same base solve, same step, same multipliers, and only the source of the diagonal toggled:

| diagonal handed to the correction | error, budget 1 | budget 8 | residual | `improved()` |
|---|---|---|---|---|
| frozen base-point copy (this PR before `40653700`) | 1.48e-1 | 1.48e-1 | 8.76e-2 unchanged | false |
| live predicted-point, forced at the branch | 1.05e-2 | 1.05e-2 | 8.76e-2 to 5.79e-3 | true |
| control, degenerate bound removed | 1.04e-2 | 5.0e-8 | to 5.1e-9 | true |

The fixture is the gh#737 cap trigger welded to an exp-curvature workload, and that class came from a real model in the first place, notebook 36's CSTR with an initial condition on a bound. The plateau past the first iteration is the non-unique multiplier split's own, identical under a forced raw uncapped diagonal, so no diagonal choice moves it.

Our position, and where we would appreciate your input: we think the corrector should not consult the stored copy at all, even in the regimes where doing so is benign. The stored pair exists to keep the base-point back-solves consistent with the held factor, and the corrector stopped back-solving through that factor when this PR made it factor its own operator, so the consistency argument no longer reaches it. The change is now on the branch as `corrector_sigma`: the corrector builds both diagonal blocks at the predicted point in every case, with the ceiling re-derived from the Jacobians there and the declared frame recomputed there (the declared bounds are constants, so the frame follows the iterate). Measured: the stalling fixture above flips to matching the control's first iteration, a crossover arm behaves the same, and on your #828 fixture the rebuild is exactly a no-op, identical outcomes at every `A`, dead zone included, which also corroborates your orthogonality claim from this branch. The calls we would especially like your eyes on: re-deriving the cap from the predicted-point Jacobians rather than reusing the captured `cap_x`, and the declared-frame recompute at a moved iterate.

RISK 2 is closed alongside: the regression fixture gained an `eval_h` counter and a limited-memory arm, pinning that a `limited-memory` solve and its correction request zero exact-Hessian evaluations while the correction still takes the predictor's 9.45e-2 curvature error to 4.0e-4 at budget 8 against the held quasi-Newton matrix.

jkitchin:
Merged via #829 as `938a736` — thanks @devin-griff, this is your change and your commits, carried across unchanged.

Closing this one only because GitHub would not let it merge. It reported the PR as conflicted against `main` and refused with `405 Pull Request has merge conflicts`, while `git merge-tree origin/main b9b27c0` resolved with zero conflicts on a fresh clone at the same base (`0d7eab7`). The flag did not clear over a recompute window, and the branch could not be updated in place either — `update-branch` returned `403 user doesn't have permission to update head repository`, since the head is a fork branch without maintainer edits. Rather than block on that, #829 proposed `main` + this head, merged with no manual resolution: same tree, same authorship, different branch.

Worth knowing in case it recurs on another fork PR — nothing was wrong with this branch.

All 11 checks were green on #829 before merge, and I re-verified `cargo fmt --all -- --check`, `cargo check --workspace --all-targets`, and all 250 `pounce-sensitivity` tests on the merged tree locally.

Carried forward as a follow-up, not a blocker: `ConvergedState::exact_hessian`'s `false` branch still has no fixture — no `limited-memory` solve exercises the corrector anywhere in the suite, and that branch is newly load-bearing since the correction now forces a refactorization against the L-BFGS `w` at a moved iterate.

The gh#737 ceiling behaviour I raised in review turned out to be pre-existing on `main` and unrelated to this change; it is filed separately as #828.

---

jkitchin:
@devin-griff — your three follow-up commits are now on `main` via #830 as `177e45f`. Apologies for the gap: your comment landed at 20:43 and I merged #829 at 20:53 without re-reading this thread, so `main` briefly carried the frozen-diagonal defect. `268082a`, `e7446da` and `02a6c10` are all in now.

You were right and my withdrawal was too broad. My fixture's capped variable sits hard on its bound and stays there, so I sampled exactly the class where the copy is harmless, then generalised from it. The held-between-a-binding-equality-and-its-bound class is the one that matters, and I did not construct it.

Worth adding: the signal was in my own data and I did not weigh it. Re-running my #828 sweep against your fix, the errors are identical at every `A` — confirming your no-op claim — but the residual at `A = 1e-1` is not:

| `A = 1e-1`, budget 8 | error | final residual |
|---|---|---|
| before your fix | 1.0000e-10 | 3.032e0 |
| after | 1.0000e-10 | 3.481e-11 |

That is the row I used to refute myself. The error had already bottomed out on the closed-form floor while the residual sat nine orders high, so my error-only metric could not see the defect that was live in front of it. A lesson for the sens-review checklist, if it is worth an entry: a fixture whose error metric is already at its floor is not evidence about anything, and the residual is the tell.

On the two calls you asked about — re-deriving the cap from the predicted-point Jacobians rather than reusing the captured `cap_x`, and the declared-frame recompute at a moved iterate — I have **not** assessed either, and I did not want to hold the fix behind a review I had not done. Both read as sound in principle: the cap is a representability property of the row's own Jacobian entries, so deriving it where those entries are evaluated is the consistent choice, and the declared bounds being constants does make the frame follow the iterate. But that is reasoning, not verification, and given how this round went I would rather say so than imply more. Happy to do a proper `/sens-review` pass on `e7446da` if you want one — say the word.

#828 stands as filed: your rebuild is a no-op on that fixture at every `A`, dead zone included, which corroborates the orthogonality from your branch as well as mine.

---

jkitchin:
## `/sens-review` on `e7446da`, and the two calls you asked about

Ran the checklist against the merged commit. Six PASS, one RISK, one N/A. Both of your calls hold up; the RISK is coverage, not correctness.

| # | class | verdict | evidence |
|---|---|---|---|
| 1 | index space | PASS | `pinned` var-x rows against `caps`/`x` both `dims[0]`-length; `x.get_mut(var_row)?` fails loudly; no new cross-space conversion |
| 2 | frame and scaling | **RISK** | the declared-frame recompute at a moved iterate has no non-unit-scaling arm, and no arm where the point lands outside the declared box |
| 3 | absolute thresholds | PASS | no new constant; `cap_s = sigma_pin_cap(1.0)` mirrors `EffectiveSigma`'s existing −I rationale; `declared_slack_floor` is `eps·max(1,\|bound\|)`, scaled by what it measures |
| 4 | doc drift | PASS | every claim in `corrector_sigma`'s doc verified against the code (below) |
| 5 | silently wrong | PASS | 254 tests green incl. all oracle legs; the plateau is pinned as a measured fact with its counterfactual named |
| 6 | which branch | PASS | instrumented — both arms genuinely reached (below) |
| 7 | measured populations | PASS | no new discriminator; the cap and the frame rule are re-derivations of existing rules at a different point |
| 8 | which binary | N/A | all evidence Rust-side, no Python-derived numbers |

### Call 1 — the cap from predicted-point Jacobians

**I think this is required, not merely defensible.** The ceiling exists so the surviving Schur contribution `a²/Σ` stays representable, and the `a` in that contribution is the Jacobian entry of the operator actually being factored — which, since #824, is the predicted point's. Reusing the captured `cap_x` would bound `a²/Σ` with a different `a` than the one in the matrix, so the guarantee would be nominal rather than real. Deriving both at the same point is the only version where the bound means what it says.

### Call 2 — the declared-frame recompute at a moved iterate

Sound, and the mechanism is safer than I expected. My worry was the non-purified iterate: `bound_context` builds the corrector's box from `nl.x_l()`/`nl.x_u()`, the **live relaxed** bounds, so with a default `bound_relax_factor = 1e-8` the predicted point can sit up to δ *outside* a declared bound, and `s = x − b_l` goes negative. `declared_frame_sigma` floors every slack at `declared_slack_floor` first, so that case yields a large-but-finite `Σ` (then capped), never a NaN or a negative diagonal. The floor is doing exactly the job it was written for.

"The declared bounds are constants, so the frame follows the iterate" also checks out — they are read from the NLP, never from an iterate, and `declared_x_bounds` is in the same space as `curr.x` (no variable scaling is applied between the snapshot and now), so there is no frame mix.

### The RISK, and why it is this one

gh#654 justifies the declared frame for a **purified** iterate — its own words, "only a purified point has the declared frame as its natural frame." The corrector's predicted point is a purified point *plus a step*, which is a reasonable extension but a genuinely new regime for that code. Two gaps in what pins it:

1. **No scaling arm.** `corrector_ceiling.rs` sets only `sb`, `bound_relax_factor`, `crossover` and `tol` — every fixture is unit-scaled. `205bb67` was a corrector frame defect that, in its own words, "coincide[d] only at unit scaling, which is every fixture it had."
2. **`bound_relax_factor = 0.0` in the fixtures**, so declared and relaxed bounds coincide and the predicted point can never land outside the declared box. That is precisely the case where the floor above is load-bearing, and it is the one the corpus cannot see.

Cheapest thing that would settle both: a scaled arm on `a_crossover_solve_corrects_at_the_predicted_point`, and one arm at the default `bound_relax_factor` rather than zero. Neither is a new fixture, just parameters on the one you already wrote.

### Entry 6, measured rather than assumed

Per the checklist's own rule I instrumented `declared.is_some()` rather than trusting the arm's name:

```
a_ceiling_solve_corrects_at_the_predicted_point      declared_frame=false
the_same_workload_without_the_ceiling_corrects       declared_frame=false
a_crossover_solve_corrects_at_the_predicted_point    declared_frame=true
```

Both arms are genuinely exercised — the crossover fixture really does reach the declared branch.

### Entry 4, claim by claim

- "the correction zeroes their multipliers in the packed iterate, so their sides contribute nothing" — `corrector.rs` zeroes `iterate[r]` for released rows *before* `pack_natural`, so `z/s` is exactly `0/s`. True, and cleaner than the old rebuild: it removes the differencing of two `~1e7` quantities that `active_set_sigma_x`'s comment was written to avoid.
- "the declared bounds are constants" — verified above.
- "Always fresh objects" — `pack()` builds a new `DenseVector` per call.
- `fresh_barrier_sigma_x` is gone with zero dangling references; `active_set_sigma_x` correctly stays for the **predictor** path, which still works at the base point against the held factor and for which the stored pair's rationale still holds. Scoping the change to the corrector alone is the right cut and keeps the blast radius small.

### One thing I did not verify

The claim that the post-first-iteration plateau "is identical under a forced raw uncapped diagonal, so it is the degeneracy's own property." That is the right *form* of claim — measured, with the counterfactual named — and the pin is honest either way since it records a measured fact rather than a target. But I did not re-run the forced-raw variant myself, so I am not evidence for it.

---

PR #832, opened 2026-08-29
fix(sensitivity): read a capped bound's multiplier back through its cap (gh#828)

### LINKED ISSUE #828: Corrector makes no progress when the gh#737 sigma ceiling bites
When the gh#737 barrier-diagonal ceiling caps a strongly active bound, `correct_step` makes **no progress at any budget** — it stops after one iteration with `improved() == false` and hands back the uncorrected step. Reproduces on `main` (`a5e0a83`) and is unrelated to #824, which neither causes nor fixes it.

Found while reviewing #824; the corrector fixture corpus never reaches the capped path, so nothing in-repo sees this.

## Reproduction

```
min exp(x0) - p*x0 + 0.5*(x1 - (2.5-2p))^2 + 0.5*(x2+5)^2
s.t. p = p_nom,  A*x2 + x3 = 0,  x1 >= 0, x2 >= 0
```

Closed form `x0 = ln p`, `x1 = max(0, 2.5-2p)`, `x2 = x3 = 0`, so the truth is independent of the machinery.

`x2` sits hard on its bound (`Σ = z/s ≈ 2.5e10`). `A` is its only Jacobian coefficient, so it sets that row's ceiling through `sigma_pin_cap(a) = (a/eps)·(a/64)` — sweeping `A` walks the cap in and out. `x1` is a near-bound row whose slack shrinks 5× over the step; `x0` carries curvature. Perturbation `delta = 0.2` on the pin, `tol = 1e-10`, `bound_relax_factor = 0`, error is the inf-norm against the closed form.

## Measured on `main`, budget 8

| `A` | ceiling bites | initial residual | corrected error |
|---|---|---|---|
| 1e-4 | yes | 1.776e7 | 1.7678e-2 (no progress) |
| 1e-3 | yes | 1.776e5 | 1.7678e-2 (no progress) |
| 1e-2 | yes | 1.771e3 | 1.7678e-2 (no progress) |
| 1e-1 | yes | 1.276e1 | 1.7678e-2 (no progress) |
| 1.0  | no  | 2.140e-2 | 4.7226e-8 |

Predictor error is 1.7678e-2 throughout, so every capped row is the corrector returning its input.

## Why it looks like the ceiling rather than the model

The initial residual is measured by `residual_at` at the clamped predicted point, **before any factorization**, so it cannot be the corrector's operator — it is the *predictor's* multiplier block coming back badly scaled.

Toggling only the ceiling confirms it. Same model at `A = 1e-4`, same build, `sigma_pin_caps` forced to return `INFINITY`:

| | initial residual | corrected error, budget 8 |
|---|---|---|
| ceiling on | 1.776e7 | 1.7678e-2 |
| ceiling off | 2.140e-2 | 1.0000e-10 |

So it is the cap, not the model's conditioning. The residual scaling as `A⁻²` across the sweep is consistent with the capped rows' duals being off by the ratio the cap imposes.

## Why it matters

`sigma_pin_cap` bites at `Σ > ~7.0e13` for a unit coefficient and falls as `a²`, so an ordinary strongly active bound on a small-coefficient row reaches it well before anything looks pathological. When it does, `estimate(corrector_iter=...)` silently returns the uncorrected step. The no-improvement warning does fire, so it is not silent-wrong — but the correction is unavailable in exactly the stiff, tightly-bounded regime where a caller is most likely to ask for it.

The cap itself is doing its job (gh#737 exists because refinement escalated against an uncapped matrix). The question is whether the parametric step's multiplier block should be built in the capped frame too, so the corrector's starting point is consistent with the operator that produced it.

## Notes

- The ceiling path is live in-repo — `tests/issue_737_pinned_to_bound_sigma.rs` and `tests/crossover_sigma_downstream.rs` reach it — but no corrector fixture does, which is why the suite is green.
- The fixture above is not in the tree; happy to contribute it as a regression test alongside a fix.

jkitchin:
Fixes #828

## Problem, with numbers

`PdFullSpaceSolver` eliminates each bound row into the `x` / `s` barrier
diagonal and recovers it afterwards as `dz = r_z/s − (z/s)·dx`, re-deriving
`z/s` straight off the iterate. That is right exactly when the diagonal it
eliminated into *was* `Σ = Σ z/s`. Under the gh#737 ceiling it is not: the
solve holds the capped bound softly — which is the whole point of the
ceiling — and reads its multiplier back stiffly, and the two disagree by
exactly the cap ratio.

On the issue's fixture, a strongly active bound with `Σ = 2.5e12` capped to
`7.0e5` came back with

    dz2 = (−mu − z2·dx2)/s2 = (−1e-11 − 5·(−7.1054e-6))/2e-12 = 1.7764e7

against a true `0`, growing as `A⁻²` as the constraint row's Jacobian
coefficient `A` shrank (`Σ_cap ∝ A²`). That `1.78e7` is also, one for one,
the stationarity residual `correct_step` opened on — which is why it could
not reduce it in a single chord step, stopped with `improved() == false`,
and handed the caller back its own uncorrected input at every budget.

| `A` | `dz` before | `dz` after | corrected error before | after |
|---|---|---|---|---|
| 1e-4 | 1.776e7  | −7.10e-6  | 1.7678e-2 (no progress) | 1.553e-4 |
| 1e-3 | 1.776e5  | −7.10e-8  | 1.7678e-2 (no progress) | 4.736e-8 |
| 1e-2 | 1.771e3  | −7.10e-10 | 1.7678e-2 (no progress) | 4.817e-10 |
| 1e-1 | 1.276e1  | −7.11e-12 | 1.0000e-10              | 1.000e-10 |
| 1.0  | −2.0e-12 | −2.0e-12  | 1.0000e-10              | 1.000e-10 |

`A = 1.0` leaves `Σ = 2.5e12` under a ceiling of `7.0e13`, so the ceiling
does not bind and every number in that row is unchanged across this PR.

## Fix

This is the design question the issue poses — *"whether the parametric
step's multiplier block should be built in the capped frame too"* — answered
yes. Consistency asks the capped bound's complementarity row to read
`s·dz + r·z·dx = r_z` with the same `r = Σ_used/Σ_base` the diagonal was
scaled by, so `PdSensBacksolver::rescale_bound_multipliers` adds
`sign·(1 − r)·(z/s)·dx` back onto each affected row. `r == 1` wherever the
ceiling did not bind, so it is an exact no-op off the capped path.

The ratio travels **with** the diagonal it describes, because separating the
two is the defect: `EffectiveSigma` gained `ratio_x` / `ratio_s`, and
`corrector_sigma` now returns a `CorrectorOperator` carrying both diagonals
and both ratios. The rescale runs in **scaled space, before** the
natural-units conjugation — the ratio is a property of the operator that was
factored, and `z / s` is read off the iterate in that operator's frame.

**Scope.** Applied to the plain `sigma_override()` path (the predictor) and
to the corrector's own operator. Deliberately *not* applied to
`active_set_sigma_x`'s released/pinned diagonal: a released entry is rebuilt
from the bounds that stay active and a pinned entry is raised by an addend,
so no single per-variable ratio describes it. The new fixtures assert
`released == 0 && pinned == 0` so a later reader does not mistake them for
evidence about those branches.

**Adjacent hole, same cause.** `solve_many_scaled_space` gated its two
batched fast tiers on `self.declared.is_some()` — crossover only. Those
tiers fold the bound rows in from the *calculated* `z / s` against whatever
factor the last solve left behind, which on the first call after convergence
is the algorithm's own uncapped one. gh#737's ceiling is a second override
and fires with `declared` empty, so the gate now reads `!self.may_refine()`
— the same predicate, for the same reason, as `may_refine` itself.

## Blast radius

Off the capped path, nothing. `rescale_bound_multipliers` returns
immediately when every ratio is `1.0`, which is every entry of an uncapped
solve, and `an_uncapped_bound_is_untouched_by_the_ratio` pins that arm.
Where the ceiling *does* bind, the bound-multiplier rows of the returned
step change — from wrong to right. `dz` is the only block affected;
`dx`, `ds` and the constraint multipliers are untouched.

**Honest limit.** At the most extreme cap (`A = 1e-4`) the corrector now
makes progress but stalls at residual `1.864e-4` / error `1.553e-4` rather
than reaching the `1e-10` floor the uncapped rungs do. That is an accuracy
limit of a heavily softened chord, not a failure to progress — the operator
really has had that bound's stiffness reduced by seven orders — and it is
what the `err < plain_err / 100` assertion is calibrated against rather than
an absolute floor.

## Tests

New `crates/pounce-sensitivity/tests/issue_828_capped_bound_multipliers.rs`,
the issue's reproduction contributed as the author offered. Five legs, and
the file carries its own mutation table plus the list of what it is *not*
evidence about:

- `the_parametric_step_reads_a_capped_bound_back_through_its_cap` — the
  predictor's `dz`, swept over five row coefficients that walk the ceiling
  in and out.
- `the_corrector_makes_progress_under_the_ceiling` — `improved()`, a
  hundredfold error drop, and the initial residual pinned at the model's own
  `2.14e-2` at *every* rung, where before it ran `1.78e7` down to `2.14e-2`
  as the ceiling let go.
- `an_uncapped_bound_is_untouched_by_the_ratio` — the no-op arm.
- `the_upper_bound_branch_is_read_back_through_its_cap_too` — the other sign
  of the fold. A lower bound folds in as `−z·dx` and an upper one as
  `+z·dx`; every other fixture is bounded below only, so the `z_u` block
  never executes and a sign error there is silent. The mirror fixture is the
  same model under `x2 → −x2`, so the leg asserts the *same number*, not a
  loose bound.
- `the_capped_correction_survives_a_change_of_variables` — invariance leg 1
  for the new path, which reads `z / s` in the scaled frame and applies it to
  a scaled `dx`. Under `x̃ = d ⊙ x` both `Σ` and the ceiling carry `d⁻²`
  (`sigma_pin_cap` is quadratic in a coefficient carrying `d⁻¹`), so the cap
  binds on the same rungs in both arms and the natural-units answer must not
  move. `initial_residual` is deliberately not compared — it lives in the
  scaled frame and legitimately halves. The leg also asserts it is not
  vacuous: at the stiffest rung the scaled arm's `dz` is six orders off what
  an uncapped row returns.

Each mutation is caught by exactly the leg whose name describes it, and by
no other:

| mutation | red |
|---|---|
| drop the predictor's `rescale_bound_multipliers` | the `A⁻²` blow-up returns; predictor + corrector legs |
| drop the corrector's | opens on the right `2.14e-2`, still `improved() == false` at `A ≤ 1e-2` |
| flip the `px_u` sign | upper-bound leg only |
| move the rescale after the natural-units conjugation | scaling leg only |

`cargo test -p pounce-sensitivity` green (33 binaries), including
`sens_invariance_legs.rs`, `sens_resolve_oracle.rs`, `corrector_ceiling.rs`,
`issue_737_pinned_to_bound_sigma.rs` and `crossover_sigma_downstream.rs`.
Full workspace suite green. Not a trajectory change: nothing in
`pounce-algorithm` moves, so `scripts/sweep-fixtures.sh` is not implicated.

## /sens-review

Run on this diff; see the comment below for the per-entry table.

## Definition of done

- Code + test — above.
- CHANGELOG — `## [Unreleased]` bullet naming the Python / Pyomo surfaces.
- Book — `docs/src/sensitivity.md`, the gh#737 ceiling passage, which now
  says the ceiling holds on the way back out too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,26 @@ changes.
 
 ## [Unreleased]
 
+- **Sensitivity: bound-multiplier derivatives are readable again under the
+  gh #737 barrier ceiling, and `corrector_iter` works there** (gh #828).
+  A strongly active bound whose constraint row carries a small Jacobian
+  coefficient reaches that ceiling well before anything looks pathological.
+  When it did, the solve held the bound softly — which is what the ceiling
+  is for — but recovered the bound's own multiplier derivative stiffly, off
+  the uncapped quantities, so the two halves disagreed by exactly the cap's
+  ratio. The returned `dz` was wrong by that factor (`1.8e7` against a true
+  `0` on the issue's fixture, growing as the coefficient shrank), and
+  `estimate(..., corrector_iter=n)` / `estimate_report(..., corrector_iter=n)`
+  opened on a stationarity residual of the same size, failed to reduce it,
+  and handed back the *uncorrected* step at every budget. The report said so
+  (`improved()` was false), so no wrong answer was passed off as a corrected
+  one — but the refinement was unavailable in exactly the stiff, tightly
+  bounded regime it is asked for. Both the returned step's multiplier block
+  and the corrector's own operator now read those rows back through the same
+  cap. Affects the Rust `pounce-sensitivity` API and the Python / Pyomo
+  sensitivity entry points on top of it; where the ceiling does not bind —
+  every result that was already correct — nothing moves.
+
 - **A `maturin develop` checkout has a working `pounce` CLI again** (gh #816).
   `maturin develop` builds the extension module and nothing else. The wheel's
   `pounce` console script is a shim whose whole job is to exec the bundled
--- a/crates/pounce-sensitivity/src/algorithm_backsolver.rs
+++ b/crates/pounce-sensitivity/src/algorithm_backsolver.rs
@@ -524,20 +524,30 @@ impl PdSensBacksolver {
     /// active and a pinned entry is raised by an addend — so no single
     /// per-variable ratio describes it, and those paths are left in the
     /// frame they already had.
+    ///
+    /// Returns `false` when a ratio is in force and the correction
+    /// could **not** be applied — an unexpected vector or matrix type
+    /// behind a block, a missing iterate — and every caller fails the
+    /// solve on it. Skipping quietly would put back exactly the defect
+    /// this exists to fix, with no signal: `Σ` capped, the multiplier
+    /// rows read back uncapped, and an answer that looks like every
+    /// other. `true` whenever every ratio is `1.0`, which is the whole
+    /// of an uncapped solve and costs one pass over two slices.
+    #[must_use]
     pub(crate) fn rescale_bound_multipliers(
         &self,
         lhs: &mut [Number],
         ratio_x: &[Number],
         ratio_s: &[Number],
-    ) {
+    ) -> bool {
         let unit = |r: &[Number]| r.iter().all(|&v| v == 1.0);
         if unit(ratio_x) && unit(ratio_s) {
-            return;
+            return true;
         }
         let off = self.offsets();
         let d = self.data.borrow();
         let Some(curr) = d.curr.as_ref() else {
-            return;
+            return false;
         };
         let cq_ref = self.cq.borrow();
         let nlp_ref = self.nlp.borrow();
@@ -565,16 +575,23 @@ impl PdSensBacksolver {
                        primal_lo: usize,
                        block_lo: usize,
                        block_hi: usize,
-                       sign: Number| {
-            if block_hi == block_lo {
-                return;
+                       sign: Number|
+         -> bool {
+            if block_hi == block_lo || ratio.iter().all(|&v| v == 1.0) {
+                return true;
             }
             let (Some(pos), Some(zv), Some(sv)) = (rows(p), dense(z), dense(slack)) else {
-                return;
+                return false;
             };
+            // The expansion's rows ARE the block, one per bound; a
+            // mismatch means the index arithmetic below is against the
+            // wrong layout, so say so rather than scatter into it.
+            if pos.len() != block_hi - block_lo {
+                return false;
+            }
             for (k, &i) in pos.iter().enumerate() {
                 let (Some(&r), Some(&zk), Some(&sk)) = (ratio.get(i), zv.get(k), sv.get(k)) else {
-                    continue;
+                    return false;
                 };
                 if r == 1.0 || sk == 0.0 || !sk.is_finite() {
                     continue;
@@ -582,7 +599,7 @@ impl PdSensBacksolver {
                 let dx = lhs[primal_lo + i];
                 lhs[block_lo + k] += sign * (1.0 - r) * (zk / sk) * dx;
             }
-            debug_assert!(pos.len() == block_hi - block_lo);
+            true
         };
         fix(
             &*nlp_ref.px_l(),
@@ -593,8 +610,7 @@ impl PdSensBacksolver {
             off[4],
             off[5],
             1.0,
-        );
-        fix(
+        ) && fix(
             &*nlp_ref.px_u(),
             &*curr.z_u,
             &*cq_ref.curr_slack_x_u(),
@@ -603,8 +619,7 @@ impl PdSensBacksolver {
             off[5],
             off[6],
             -1.0,
-        );
-        fix(
+        ) && fix(
             &*nlp_ref.pd_l(),
             &*curr.v_l,
             &*cq_ref.curr_slack_s_l(),
@@ -613,8 +628,7 @@ impl PdSensBacksolver {
             off[6],
             off[7],
             1.0,
-        );
-        fix(
+        ) && fix(
             &*nlp_ref.pd_u(),
             &*curr.v_u,
             &*cq_ref.curr_slack_s_u(),
@@ -623,7 +637,7 @@ impl PdSensBacksolver {
             off[7],
             off[8],
             -1.0,
-        );
+        )
     }
 
     /// Whether the held-factor back-solve may run
@@ -747,7 +761,9 @@ impl PdSensBacksolver {
         // factored, and `z / s` is read off the iterate in the frame
         // that operator lives in (gh#828).
         if let Some((rx, rs)) = ratios {
-            self.rescale_bound_multipliers(lhs, rx, rs);
+            if !self.rescale_bound_multipliers(lhs, rx, rs) {
+                return false;
+            }
         }
         if let Some(c) = self.conj.as_ref() {
             for (l, &f) in lhs.iter_mut().zip(c.f.iter()) {
@@ -992,7 +1008,11 @@ impl PdSensBacksolver {
     /// so their sides contribute nothing. Pinned rows are raised
     /// under the same ceiling. Always fresh objects, because the
     /// factorization cache keys on their tags and a cached vector
-    /// could resolve to a factor built at another iterate.
+    /// could resolve to a factor built at another iterate. How much of
+    /// each entry the ceiling left standing comes back with them, in
+    /// the same [`CorrectorOperator`], because the solve that folds a
+    /// bound row into one of these diagonals has to read that row back
+    /// through the same cap (gh#828).
     pub(crate) fn corrector_sigma(&self, pinned: &[(usize, Number)]) -> Option<CorrectorOperator> {
         use pounce_linalg::dense_vector::DenseVectorSpace;
         let dense = |v: Rc<dyn pounce_linalg::Vector>| -> Option<Vec<Number>> {
@@ -1817,7 +1837,9 @@ impl PdSensBacksolver {
                 return false;
             }
 
-            self.rescale_bound_multipliers(lhs_row, &self.sigma.ratio_x, &self.sigma.ratio_s);
+            if !self.rescale_bound_multipliers(lhs_row, &self.sigma.ratio_x, &self.sigma.ratio_s) {
+                return false;
+            }
         }
         true
     }
@@ -2294,8 +2316,7 @@ impl PdSensBacksolver {
         if self.unpack(&res_iv, lhs).is_err() {
             return false;
         }
-        self.rescale_bound_multipliers(lhs, &self.sigma.ratio_x, &self.sigma.ratio_s);
-        true
+        self.rescale_bound_multipliers(lhs, &self.sigma.ratio_x, &self.sigma.ratio_s)
     }
 }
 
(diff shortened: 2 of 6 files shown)


jkitchin:
## `/sens-review` on this diff

| 1 | index space | PASS | introduces no space conversion; the one piece of index arithmetic is now checked rather than asserted |
| 2 | frame and scaling | PASS | `the_capped_correction_survives_a_change_of_variables`, mutation-checked |
| 3 | absolute thresholds | PASS | no new constant in `src/`; test budgets are relative, with the measured spread recorded next to each |
| 4 | doc drift | PASS | every number in a touched comment re-measured this pass; two comments corrected |
| 5 | silently wrong while reporting success | PASS *after a change made during the review* — see below |
| 6 | which branch does the fixture take | PASS | upper-bound leg added; the branches still unreached are named in the file |
| 7 | name the measured populations | N/A | no threshold or discriminator introduced |
| 8 | which binary did the harness load | N/A | all evidence is `cargo test` from this checkout; no Python-side number is offered |

### 1 — index space

Every index in `rescale_bound_multipliers` lives in one space. `pos` comes
from the bound expansion's `expanded_pos_indices()`, `ratio_x` is built from
the `sigma_x` vector, and `lhs[off[0] + i]` is the KKT `x` block — all the
algorithm's var-x. No `full_x` appears anywhere in the diff, no
`var_x_to_full_x`, no `ActivityReport` read. So the class the `leg_fixed_*`
legs exist for is not reached; the fixtures having no fixed variable is not
a gap here, because there is nothing to convert.

The one index-arithmetic claim — that the expansion's row count *is* the KKT
block it scatters into — was a `debug_assert!`, i.e. absent in release. It is
now a checked `return false`, which is the "table lookup fails loudly" rule
applied to the only direct-array-index site the diff adds. `capped_dz_row`
in the test asserts the block dims rather than assuming them, so a fixture
whose shape drifts fails by name instead of reading a neighbouring row.

### 2 — frame and scaling

Reached directly: the correction reads `z / s` off the iterate and multiplies
a `dx` taken from the solution vector, and `solve_released_prebuilt`
post-multiplies by the natural-units factor `F`. Applying the rescale on the
wrong side of that multiply is a frame mix that is invisible at unit scaling,
which is every fixture the crate had. My first attempt did exactly that — it
applied the correction in `corrector.rs` after `solve_released_prebuilt`
returned — which is why the `ratios` parameter is threaded into the function
instead.

Discharged by the new leg, which solves both arms under `user-scaling` with
`d = [2, 0.5, 4, 0.25, 1]` and compares the natural-units answer. Under
`x̃ = d ⊙ x` both `Σ` and the ceiling carry `d⁻²` — `sigma_pin_cap` is
quadratic in a Jacobian coefficient carrying `d⁻¹` — so the cap binds on the
same rungs in both arms, which is what makes the comparison fair rather than
a capped-vs-uncapped one. `initial_residual` is deliberately not compared: it
is measured in the scaled frame and legitimately halves, `2.14e-2 → 1.07e-2`.

Mutation: move the rescale to after the conjugation. That leg goes red and
no other.

### 3 — absolute thresholds

No new constant in `src/`. `r == 1.0`, `u < b`, `sk == 0.0` are exact
comparisons used as fast-path skips, not tolerances — at `r = 1 ± ε` the
correction is `O(ε)` either way, so the branch is benign in both directions.

The budgets are in the tests. Measured, then set:

- cross-scaling `dz`: measured spread `0` to `7.5e-4` relative, budget
  `1e-2`. It was `1e-3` on the first pass — 1.33× margin, which the
  checklist's own PR #800 note is against, and there is nothing to buy by
  it: both mutations this leg exists for move the number by order one or
  more. The measurement is recorded next to the assert.
- cross-mirror `dz` and both corrected errors: `1e-6` relative, against a
  measured difference of `0` — the mirror is the same model under
  `x2 → −x2`, so equality is the honest assertion.
- `err < plain_err / 100` is relative to the predictor's own error at that
  rung, not an absolute floor. This matters at `A = 1e-4`, where the
  corrector reaches `1.553e-4` rather than the `1e-10` the uncapped rungs
  do — a heavily softened chord's accuracy limit, not a failure to progress.

One limit, recorded rather than fixed: `dz.abs() < 1e-3` in the predictor leg
is absolute. It is non-vacuous at every rung (the pre-fix values run `1.8e7`
down to `12.8`, and the worst true value is `7.1e-6`, a 140× margin), but a
predictor returning a *null* multiplier block would also pass it. What rules
that out is the corrector leg, whose `plain_err` is computed from the real
step.

### 4 — doc drift

Every number in a comment the diff touches was re-measured on this pass, not
carried over. Two corrections were made as a result:

- the module header claimed the corrector-only mutation "leaves the chord
  unable to reduce the residual below 1.8e-2"; the measured failure is
  `improved() == false` at `A ≤ 1e-2`. The header now says that.
- `corrector_sigma`'s doc described a return type that no longer exists; it
  now names `CorrectorOperator` and says why the ratios travel with the
  diagonals.

### 5 — silently wrong while reporting success

**This entry changed the diff.** The defect being fixed is *not* in this
class — it warned, via `improved() == false`, which is why the issue could be
written at all. But the fix as first written was: `rescale_bound_multipliers`
returned `()` and skipped quietly whenever a downcast failed — an unexpected
vector type behind `curr.z_l`, a non-`ExpansionMatrix` bound expansion, a
missing iterate. On that path `Σ` is capped, the multiplier rows come back
uncapped, and the answer is gh#828 again with nothing to distinguish it from
a correct one.

That is not symmetric with `cap_sigma`, which shares a downcast with
`cap_ratio` so the two fail together — these are different objects and fail
independently. It now returns `#[must_use] bool` and all three call sites
fail the solve, turning a wrong answer into a failed one. The `true` fast
path still costs one pass over two slices when no ratio binds.

The class is otherwise unreached: no success signal changes meaning, and
`sens_resolve_oracle.rs` — the only guard in the crate reading an outside
number — is unmoved and green. This change does not reroute which correction
a mode reaches for; it repairs the frame the existing one is expressed in, so
it does not earn an oracle row.

### 6 — which branch does the fixture take

The rescale branches four ways (`px_l` / `px_u` / `pd_l` / `pd_u`) and the
sign is opposite on the lower and upper halves. Every fixture on the first
pass was bounded *below* only, so the `z_u` block never executed and a sign
error there was silent — a green suite proving nothing about half the code
under review. `the_upper_bound_branch_is_read_back_through_its_cap_too` is
the second fixture, and flipping the `px_u` sign reddens it and only it.

Still unreached, and now named in the file's not-evidence-about section:

- the `s` block. Every fixture has `dims[1] == 0`, so `ratio_s` is empty and
  the `pd_l` / `pd_u` arms never run. They carry the same two signs one block
  over; what would exercise them is a bounded `d(x)` row stiff enough to
  reach `sigma_pin_cap(1.0) = 7.0e13`.
- the released and pinned diagonals, which deliberately carry no ratio. The
  legs assert `released == 0 && pinned == 0` so this cannot be misread as
  evidence about them; `corrector_ceiling.rs` drives the pinned branch under
  the same ceiling and is green.

PR #839, opened 2026-08-29
sens: keep the solve's variable objects instead of re-parsing names per call

devin-griff:
## The change

Every `estimate()` call maps its answers back onto the model, and it did so by taking each solver column's name string and resolving it through `find_component`, which parses the name through pyomo's component-UID lexer. That is one parse per variable per call, 62,412 on a 62k-variable collocation model, re-deriving the same objects every time. The resolution already happens once with direct references at solve time, when the converged solve loads its solution back onto the original model, so the session now keeps that list in column order and every later query reads it.

The rerouted paths: `estimate()`'s output, `gradient(target=None)`'s walk over every variable, `estimate_report()`'s crossed-variable and crossed-row maps, `active_set_changes()`'s per-change variable, and the session's single-name helper. Constraint rows are never resolved by the solve's load loop, so they resolve lazily, once per session, at the first report that needs them, through the same user-row-name aliasing as before. A column the declared-parameter surgery created has no model counterpart, holds an empty entry, and is skipped exactly as before, and the clone-solve path still yields the original model's objects because the capture happens in the load loop that already targets the original model. The captured references are the solve's own objects, so a caller who deletes and rebuilds model components after a solve invalidates this cache like any other the session holds.

`estimate()` also stopped paying one container insertion per variable per call to build its result. It returns a read-only `SolutionMap`: lookup, iteration, membership, length, and `items()` are unchanged and still keyed by the component data objects themselves, the keys and identity index are the session's and shared by every result, and each result carries only its own value vector. Item assignment raises, and no consumer in this repo mutates or type-checks a result: every bracket access in the tests is a read, continuation iterates items, and the notebooks store results into their own containers.

## Measured

On the 62k-variable double column (N=25 Radau collocation), an accumulating timer on `find_component` itself measures 62,412 calls costing 0.87 s inside one parent-code `estimate()` call, 13.9 us per name. That is the figure `CHANGELOG.md`, the `_Session` comment and `test_name_resolution.py` all carry. (An earlier revision of this body put the cost at 1.4 s by subtracting the wall times before and after; that subtraction attributed everything that moved to one cause and over-attributed, and the direct timer replaces it.) The container build sits below wall resolution and is measured in the profile: the per-variable insertion loop, 0.065 s, no longer appears, and the call's remaining python is 0.021 s of perturbation assembly and 0.011 s of scatter, with `find_component` absent from the listing.

## Tests

`test_name_resolution.py` pins the mechanisms: after a solve, `estimate()` and `gradient(target=None)` perform zero `find_component` calls, counted by a pass-through patch, and both tests fail on the parent commit, which resolves one name per solver column per call. The result's contract is pinned separately: the Mapping interface is complete, item assignment raises, a non-column key raises, and the keys are the original model's own data objects by identity, which matters because the solve runs on a clone.

486 tests pass on `pyomo-pounce` on this machine, and `python/tests` is green to its known environment failures. The three issue-816 availability tests and the shim-notice path assertion fail here identically on `main`, their fixtures being sh scripts and forward-slash paths on a Windows box.

---

- [x] Tests fail on the parent commit for the stated reason: both counting tests resolve one name per column there
- [x] `CHANGELOG.md` `[Unreleased]` entries, one per change
- [x] Book page: unchanged, no documented surface moves beyond the return type, which the `estimate` docstring now states


jkitchin:
## Review

The core idea is right and the mechanism holds up. I verified it rather than trusting the description.

### What I checked and found sound

- **Index spaces align.** `var_data` is built from `zip(var_names, base_x)`; on a two-declared-param model I measured `n = len(var_names) = len(base_x) = len(var_data) = len(nl.x_l) = 6`, and `len(user_row_data()) = len(con_names) = m`. So `session.var_data[i]` in `estimate_report` (indexed off `lo`/`hi`, length `n`) and `row_data[j]` (indexed off `g_pred`/`live`, length `m`) are in the right frames. Surgery columns land as `None` at exactly the `_SENSITIVITY_TOOLBOX_DATA.*` positions and are dropped from `solution_keys()` — `len(est) == 4`, keys `x[0..3]`.
- **`var_data` is never `None` on a live session.** The non-converged path sets `reg.session = None` (sens.py:986) before the load loop, and `_Session` is constructed at exactly one site.
- **No aliasing hazard from holding `x_new` by reference.** Every assignment on the path (`base_x + dx`, `np.clip`) produces a fresh array, so a `SolutionMap` can't be mutated out from under its holder — which the old float-copy-per-key loop made structurally impossible and this version relies on.
- **`session.model` is the caller's model** at every entry point (`estimate`, `estimate_report` read `_REG` off the passed model), so moving `find_component` from the caller's `model` to `session.model` in `user_row_data()` is a no-op.
- **Suite:** 453 passed / 5 failed on the branch; the same 5 fail identically on `main` (stale installed `pounce` 0.10.0). The new file's 4 tests pass. `crossed_rows` non-empty cases (test_estimate_report.py:233, :892) still pass.
- **Caching works as advertised:** `estimate_report` resolves 4 names on the first call, **0** on the second; `active_set_changes` resolves 0.

### Findings

**1. `est == anything` now raises `TypeError` — a real regression (medium)**

`ComponentMap` defines its own identity-hash `__eq__`. `SolutionMap` inherits `Mapping.__eq__`, which is `dict(self) == dict(other)` — and Pyomo `VarData` is unhashable, which is the whole reason `ComponentMap` exists. Measured:

```
e1 == e1  -> RAISES TypeError unhashable type: 'VarData'
dict(e1)  -> RAISES TypeError   (also broken before — not a regression)
```

The CHANGELOG says "Lookup, iteration, membership, length, and `items()` are unchanged." True as far as it goes, but equality *did* change, from a working bool to a raise. Worth ~8 lines:

```python
def __eq__(self, other):
    if self is other:
        return True
    if not isinstance(other, Mapping) or len(self) != len(other):
        return False
    for k, v in other.items():
        i = self._index_of.get(id(k))
        if i is None or float(self._values[i]) != v:
            return False
    return True
```

**2. The measured numbers disagree across the artifacts that ship (medium)**

- CHANGELOG: "that parsing was **2.2 s** of a 3.4 s `estimate()` call"
- `sens.py:381` comment: "**~2 s** over a 62k-variable model"
- `test_name_resolution.py` docstring: "**2.2 s** of a 3.4 s `estimate()` call"
- PR body + commit message: "drops from **3.4 s to 2.0 s** … the removed **1.4 s** is the name parsing"

2.2 and 1.4 can't both be the parsing cost of the same 3.4 s call. The three in-tree artifacts are the odd ones out relative to the PR's own measurement.

For calibration I benchmarked the identical operation at the same scale on this machine: 62k flat `x[i]` names → **0.136 s** (2.2 µs each); 60k nested `stage[t].x[c,s]` names → **0.313 s** (5.2 µs each). That's 4–16× below either claim. Name shape and machine plausibly explain a gap, but nothing in the PR names the model, so a reader can't check it. Per CLAUDE.md's rule about naming the measured population behind a number: pick one figure, put the same one in all three places, and say which model produced it.

The optimization is clearly worth doing regardless — 62k parses per call is 62k parses per call.

**3. Stale type annotations in a shipped example (low–medium)**

`python/pounce/examples/asnmpc_cstr.py` annotates `values: pyo.ComponentMap | None` on `_component_value` (:575), `_profile_value` (:590), `trajectory` (:598) and `_load_estimate(values: pyo.ComponentMap)` (:702) — and `estimate()`'s result at :670 / :1293 / :1301 is exactly what flows into all of them. Runtime is fine (every use is a read or `.items()`), so the PR's "no consumer … type-checks a result" is true of runtime behavior, but the annotations are now wrong and a type checker over that file would say so.

**4. `gradient(target=None)` builds an n-entry dict it doesn't need (low)**

sens.py:1219 walks `session.var_names` in order calling `orig_var(nm)`, which lazily builds `dict(zip(var_names, var_data))` — an O(n) dict build plus n hash lookups to reproduce `session.var_data` positionally. On the 62k model that's 62k insertions + 62k lookups on the exact path this PR exists to speed up. It's just:

```python
targets = [v for v in session.var_data if v is not None]
```

which is also `solution_keys()[0]`. `_var_by_name` would then be needed only by `_pinned_variables` (:2960), which is over a handful of indices and could index `var_data` directly too — at which point `orig_var` and `_var_by_name` both go away.

**5. Coverage gap: four of the six rerouted sites are unpinned (low–medium)**

The two counting tests cover `estimate()` and `gradient()`. The diff also reroutes `estimate_report`'s `crossed` and `crossed_rows`, `active_set_changes`'s `var`, and `user_row_data()`'s once-per-session laziness. Existing tests would catch a *wrong* object there but not a re-introduced per-call resolve — which is precisely the class this PR is about. I confirmed the assertion is available and stable:

```python
# first call resolves m row names, every later one resolves nothing
counted(lambda: estimate_report(m, [(m.p, 1.5)])) == len(session.con_names)
counted(lambda: estimate_report(m, [(m.p, 1.6)])) == 0
counted(lambda: active_set_changes(m, [(m.p, 1.5)])) == 0
```

One more test with those three lines closes it.

**6. CHANGELOG missing a blank line (trivial)**

The new `SolutionMap` bullet runs straight into `- **Limited-memory: …` with no blank line between them; every other bullet in the file is separated.

### Verdict

No blocking correctness bug. Finding 1 is the one I'd fix before merge (it's a silent-until-hit API regression and the fix is small); 2 matters because this repo's CHANGELOG numbers get cited later; 3–6 are cleanup.

devin-griff:
All six applied, in `ee717131`.

1. `SolutionMap.__eq__` compares through the identity index and returns the bool `ComponentMap` returned, with `NotImplemented` for a non-Mapping other, pinned by `test_solution_maps_compare_by_contents_without_hashing_keys`.
2. One figure in every artifact, 1.4 s of a 3.4 s `estimate()` call, and the measured model named in all three places: the 62k-variable double column at N=25 Radau collocation. Your calibration was right that the population was unstated.
3. The example's four annotations now say `Mapping`.
4. Taken as proposed: `gradient(target=None)` and the pinned-variables walk read `var_data` directly, and `orig_var` and its name dict are deleted.
5. `test_the_report_and_the_record_resolve_names_once_per_session` pins your three assertions: the first report resolves exactly the row names, later reports and the record resolve nothing.
6. Blank line restored.

488 on `pyomo-pounce` and `python/tests` green to the platform knowns on this machine.

jkitchin:
## Review, round 2

Reviewed at `ee717131` (the merge of `e10caf7b` into current `main`), in a worktree, with the branch's `pyomo_pounce` actually loaded (`PYTHONPATH` confirmed — `pyomo_pounce.__file__` points into the worktree).

### Round-1 findings: 5 of 6 fixed, 1 re-broken

| # | Status |
|---|---|
| 1 `__eq__` | **Fixed.** Verified no `TypeError`: `a == a`, `a == b`, `a != c` all bool. Also correct in both directions against a real `ComponentMap` (`cm == est` and `est == cm` → `True`). Returning `NotImplemented` for a non-Mapping is better than `ComponentMap`'s bare `False`. |
| 2 figures | **Fixed on the consistency axis.** 1.4 s / 3.4 s / "62k-variable double column (N=25 Radau)" now reads identically in `CHANGELOG.md:17`, `sens.py:397`, and the test docstring. See finding C on the magnitude. |
| 3 annotations | **Fixed.** Four `pyo.ComponentMap` → `Mapping`. |
| 4 `orig_var` | **Fixed, and taken further than proposed.** Zero remaining references to `orig_var`/`_var_by_name` anywhere in the repo. The positional read is also strictly more correct than the name dict, which would have collapsed duplicate names. |
| 5 coverage | **Fixed.** `test_the_report_and_the_record_resolve_names_once_per_session` pins all three assertions. |
| 6 blank line | **Re-broken by the merge.** `ee717131` dropped the two new bullets above main's QP bullet without a separator — `CHANGELOG.md:21`→`:22` run together. On `main` that bullet is blank-separated. (The `CasADi plugin` one further down is pre-existing on main, not yours.) |

### New findings

**A. `estimate_report()`'s first call is now ~5× slower (medium)**

`sens.py:2088` calls `session.user_row_data()` unconditionally, which resolves *every* row name. The old code called `find_component` only inside the `np.where(live & out_of_bounds)` loop — for crossed rows, normally zero. Measured on an 8000-var / 8000-row model:

```
                 main      branch
estimate  1st   0.0229    0.0051
estimate  2nd   0.0199    0.0018     <- the PR's win, 11x, confirmed
report    1st   0.0042    0.0210     <- 5x slower
report    2nd   0.0039    0.0039     <- identical
```

So the row cache buys nothing in the steady state and costs `m` parses once — ~0.13 s at 62k rows at that rate. The new test documents it: `fc.call_count == len(session.con_names)` on a fixture where nothing crosses. Guarding it keeps both properties:

```python
rows_out = np.where(live & out_of_bounds)[0]
if rows_out.size:
    row_data = session.user_row_data()
```

The test then needs a fixture that actually crosses a row to keep pinning once-per-session — which is the fixture that reaches the other branch, i.e. the test CLAUDE.md's rule asks for anyway.

**B. `deepcopy` of a result is silently wrong (low, but the silent class)**

`ComponentMap` carries `__autoslot_mappers__ = {"_dict": _rehash_keys}` specifically so a deepcopy rehashes by the new ids. `SolutionMap` has `__slots__` and no hook, so the copied `_index_of` keeps the *originals'* ids while `_keys` holds fresh objects. Measured:

```
main    deepcopy(est)[its own key] -> 1.5        m.x[0] in copy -> False   (both right)
branch  deepcopy(est)[its own key] -> KeyError   m.x[0] in copy -> True    (both wrong)
```

Neither answer raises where it should. `model.clone()` on a block holding a result would hit it. Pickling fails identically on both branches (the `VarData` drags in the model → `cannot pickle 'pounce.NlProblem'`), so that half is not a regression. A `__deepcopy__` that rebuilds the index from the copied keys fixes it; refusing outright would also beat lying.

**C. 1.4 s is still ~3× above independent calibration (low, informational)**

Naming the model answered "which population", but the number itself didn't move toward measurement. I timed `find_component` on genuinely collocation-shaped names on this machine, scaled to 62,412:

- flat 3-index, 52 chars (`b.c.flow_mol_phase_comp[0.069431844203,0,benzene]`) → **0.37 s** (5.9 µs/name)
- deep indexed-block, 68 chars (`fs.col.tray[0].props[0.069431844203].flow_mol_phase_comp[benzene]`) → **0.53 s** (8.5 µs/name)

1.4 s needs ~22 µs/name. The end-to-end number agrees with the calibration, not the claim: main's `estimate()` steady state on the 8000-var model above is 0.0199 s vs the branch's 0.0018 s — 2.26 µs/var saved, extrapolating to **0.14 s** at 62k. The 1.4 s is a wall-clock subtraction (3.4 − 2.0), which attributes everything that moved to one cause. You already ran a profile; the *parent's* cumulative time for `find_component` is the number that can't be argued with, and it's the one I'd put in the CHANGELOG. The optimization is unambiguously right either way.

**D. Registration precedes the capture (nit)** — `reg.session = session` at `sens.py:1042` is 12 lines ahead of `session.var_data = var_data` at `:1054`. If anything in the load loop raises, a registered session is left with `var_data is None` and the next `estimate()` dies inside `solution_keys()` with `TypeError: 'NoneType' object is not iterable` instead of the "no sensitivity session" message. Moving the registration below the loop closes it.

**E. Import placement (nit)** — `from collections.abc import Mapping` landed after `import numpy as np` in `asnmpc_cstr.py`, in the third-party block rather than the stdlib block that already holds `from typing import ...`. Ruff passes (no `I` rules configured), so style only.

### What else I checked and found sound

- **Suite:** 438 passed / 5 failed on the branch; the same 5 fail identically on `main` (stale editable extension at 0.10.0). All 6 new tests pass. Two modules can't be collected here on either branch — `test_asnmpc_example.py` (`pyomo_cvp` not installed) and `test_issue_591_...` (needs Pyomo ≥ 6.10.1; this box has 6.10.0). Both pre-existing; the first means the `asnmpc_cstr.py` edits are unexercised locally.
- **No aliasing.** All three `x_new` assignments (`session.base_x + dx`, two `np.clip`s) yield fresh arrays, so holding it by reference is safe.
- **Row frame.** `row_names` in `estimate_report` and `_user_row_names(self)` inside `user_row_data()` are the same list in the same order, so `row_data[j]` is indexed in the right space.
- **`var_data` is never `None` on a reachable path.** The non-converged branch clears `reg.session` and returns at `:998`, before the session exists.
- **`active_set_changes` and `_conditioned_on`** index `var_data` at exactly the positions the old code fed to `var_names`.
- **Docs.** Nothing in `docs/src/` names `estimate()`'s return type, so "Book page: unchanged" holds.

### Verdict

No blocking correctness bug. A and B before merge — A because the report path as written is a net loss, not a win, and B because it's the fail-silently class. 6 is a one-line merge artifact. C is about a number this repo's CHANGELOG gets cited for.

devin-griff:
Round 2 applied, in `55cb9da3`.

**A.** Guarded as you sketched: the row resolution runs only when `rows_out.size`, so a no-crossing report resolves nothing, and the once-per-session property is repinned on a fixture whose inequality row the step drives out of bounds, the branch-reaching fixture the rule asks for. The old test's no-crossing arm now asserts zero, which fails on the unguarded version.

**B.** `SolutionMap.__deepcopy__` rebuilds the identity index from the copied keys at their original columns. The pin: a copy answers for its own keys, refuses the original's with `KeyError`, and `m.x[0] not in copy`.

**C.** You were right that the wall subtraction overattributed. The direct number, an accumulating timer on `find_component` itself inside one parent-code `estimate()` on the column: 62,412 calls, 0.87 s accumulated, 13.9 us per name. That figure replaces the subtraction in all three artifacts with the method named. The 13.9 us sits between your calibration and my inference, plausibly the float time indices in every collocation name plus this box, and the residue of the old 1.4 s was run-to-run contention in the before walls.

**D.** `reg.session = session` moved below the capture loop, so a load that raises leaves the no-session error instead of a half-built session.

**E.** The `Mapping` import sits in the stdlib block.

**6.** Blank line restored at the merge seam.

489 on `pyomo-pounce` and `python/tests` at 1090 to the platform knowns.

PR #842, opened 2026-08-29
sens: degeneracy="release_all", the all-released step undecided

devin-griff:
## The change

A kinked base point currently offers two treatments of its weakly active bounds: `"directional"` decides each one by the directional-derivative QP, and `"one_sided"` takes whatever side the held factorization leans toward. On a base point with hundreds of weak bounds the decision's engagement can exceed any practical `degeneracy_iter`, and the caller then pays the failed attempt and falls back to one-sided with a warning anyway, which is what every `estimate()` call did in the 62k column's earlier measurements. `"release_all"` is the third option: every weakly active bound is released undecided, at one back-solve and no QP, and the bounds the perturbation actually holds come back as bound crossings for whatever runs next. `fix_relax` pins them, `path` walks them, `linear` clamps each crossing coordinate and leaves its neighbors carrying the released coupling, which the docstring states, and a correction acts on whatever the mode left. `"directional"` stays the default.

The crate entry, `parametric_step_release_all`, is `solve_released` over the weak rows, the directional path's first back-solve returned as the answer instead of refined, and a clean base point takes the plain step. `degeneracy_iter` belongs to the directional decision, so passing it under another option warns and changes nothing, on all three entry points. Under `fix_relax` and `path` the dispatch reuses the decided variants with an empty held list, which releases the whole weak set by construction, so those modes run their own machinery under the all-released treatment with no new plumbing. The option is on `estimate()`, `estimate_report()`, and `active_set_changes()`, where a bound the perturbation holds appears as a return to its bound along the path rather than a decision at the base point. Strongly active bounds are outside every degeneracy option's scope, unchanged here.

## Measured on the 62k column

Three initial-state perturbations at a base point with 22 strongly active and over a hundred weakly active bounds, errors against a tol 1e-10 re-solve. At every point, mode, and corrector budget measured, the three options produce identical errors to every printed digit, overall and per coordinate: on this direction the weak set barely couples into the perturbation, and identical accuracy at lower cost is the option's claim, not a coincidence it hides. The cost difference is the decision itself: per `estimate()` call, `release_all` runs 1.0 to 1.5 s cheaper than `"directional"` at these points, the full engagement, and marginally cheaper than `"one_sided"`. Where the options genuinely disagree is an exact kink, and the tests below pin it: `release_all`'s raw step deterministically follows the released direction onto the holding side's wrong answer, where `"directional"` is right and `"one_sided"` depends on the factorization's lean, and each mode's own repair brings it back.

## Relative to sIPOPT

For orientation against the reference implementation, from `contrib/sIPOPT/src/SensStdStepCalc.cpp` on current coin-or/Ipopt master: sIPOPT's step is the fixed-active-set back-solve, this repo's `"one_sided"`, and its optional bound check finds x bound violations against an absolute `sens_bound_eps` (default 1e-3), appends the violated bounds' rows to the Schur complement, and re-solves in a loop. The loop only adds rows, so it pins crossings and can never release, and the paper's eq. 14 decision appears nowhere in the shipped code.

Against that, `release_all` with a repairing mode is better on both sides of a kink, and the asymmetry is what each treatment makes visible. On the releasing side the released step is the exact one-sided derivative, where the fixed-set step carries the kink's order-one, partly enforcing diagonal entry and is wrong by a finite fraction of the movement, feasibly, so a pins-only check has nothing to detect. On the holding side the released step drives the coordinate through its bound, since the unconstrained minimum sits on the infeasible side there, so the error arrives as a visible violation the pins repair exactly, where the fixed-set step's partial movement is again feasible and invisible. The kink tests below measure exactly this: the one-sided step is wrong on at least one side by construction, and `release_all` lands on the correct answer on both sides in every mode. The regime where the comparison can flip is `mode="linear"` with clamping on a strongly coupled model, where the clamp repairs only the crossing coordinate.

## Tests

Five tests in `test_degeneracy.py`. The raw released direction is deterministic on the holding side with `clamp=False`. Every mode repairs the decoupled kink's holding side to the bound. The coupled kink pins each mode's repair with re-solve-verified constants: the releasing side equals the directional decision in both coordinates in every mode, `fix_relax` repairs both coordinates on the holding side, `linear`'s clamp repairs only the crossing coordinate while the neighbor keeps the released coupling at 1 - 10/7, and `path` returns the crossing coordinate but leaves the neighbor at the one-sided 2/7 rather than the re-solve's 1, pinned as current behavior of the decided path surface under an all-released start rather than endorsed, and worth a look. A clean base point is identical under all three options, and an inert `degeneracy_iter` warns.

On the crate, the invariance legs gain the released row in every frame: the all-released map equals the releasing side's exact derivative on both signs, unmoved by the change of variables, the magnitude span, and a fixed variable ahead of the kink. The oracle pins that the corrector's fixed point does not depend on which start it was handed, and that a strongly active release is untouched, zero rows released and the plain answer to refinement noise.

263 tests pass on `pounce-sensitivity` and 487 on `pyomo-pounce`, the three issue-816 availability tests failing on this Windows machine identically on `main`. The pyomo suite's new tests call the new entry through the extension module, so a stale build fails them loudly rather than passing by accident. `python/tests` is green to its known environment failures.

---

- [x] Tests fail on the parent commit: the option does not exist there
- [x] Book page: the degeneracy section documents the option beside the two existing ones, with `"directional"` remaining the default


jkitchin (commented):
Read the full diff, the surrounding code, and ran `/sens-review` over it. `cargo check -p pounce-sensitivity -p pounce-py --tests` is clean (exit 0); I did not run the suites — no built extension module in my container, so everything below is read from source rather than measured.

## `/sens-review` verdict

| 1 | index space | PASS | the only new indexing is `w.row` → released KKT rows (identical to `parametric_step_directional`, solver.rs:1315) and the `d[..ctx.n_x]` var-x slice, consumed by `session.scatter_x`. No new full-x/var-x crossing. |
| 2 | frame and scaling | PASS (arithmetic) | nothing compared against `Sigma` or `Sigma/mu`; the solve routes through `solve_released_prebuilt`, which applies `conj`'s E/F like every sibling. No scaling *leg*, though — R1. |
| 3 | absolute thresholds | N/A | no new constant. |
| 4 | doc drift | RISK | R4, R5 |
| 5 | silently wrong while reporting success | RISK | R1, R4 |
| 6 | which branch does the fixture take | RISK | R2, R3 |
| 7 | measured populations | N/A | no new discriminator; the 62k cost figures are reported with their method. |
| 8 | which binary | RISK (mild) | M4 |

## R1 — the new Rust surface has no Rust test

`Solver::parametric_step_release_all` is the only new code in `pounce-sensitivity`, and nothing in that crate's suite executes it: `release_all` appears in `src/solver.rs` and `crates/pounce-py/src/solver.rs` and in no test file. So "261 tests pass on `pounce-sensitivity`" is true and vacuous with respect to this change.

- **No invariance leg.** Every leg in `sens_invariance_legs.rs` routes through the `step()` helper (line 309), which calls `parametric_step_directional`. The released direction gets no scaling, no magnitude, and no fixed-variable row. CLAUDE.md's rule is explicit: *"A new public accessor gets a row in each leg — the cost of leaving one out is that the next defect in that dimension is invisible."*
- **No row in the oracle.** Under `mode="linear"` with `corrector_iter > 0`, `release_all` changes which step the corrector starts from. That is exactly "reroutes which correction a mode reaches for," which is the stated trigger for a row in `sens_resolve_oracle.rs`. The PR's headline claim — *"at every point, mode, and corrector budget measured, the three options produce identical errors"* — currently rests entirely on a 62k model no reviewer can re-run.

The oracle's existing fixture pair looks like the right one: the weakly active kink reaches the case where the released direction is deliberately wrong on the holding side, and the strongly-active-release fixture reaches the case `release_all` explicitly does *not* touch (strong bounds are outside every degeneracy option's scope) — worth pinning that non-effect too, since it is asserted nowhere.

Given S1 below, these legs should be cheap: the function is arithmetically `solve_released`, so a released variant of `step()` on the existing fixture is a small addition.

## R2 — the tests reach only one branch of the weak set

`weakly_active_bounds()` admits both `WEAKLY_ACTIVE` **and** `AMBIGUOUS`, deliberately (solver.rs:1612–1618), and gh#763 is precisely that a *coupled* kink lands in `AMBIGUOUS`. All three new tests use `kink()` — one decoupled variable, the `WEAKLY_ACTIVE` branch.

`coupled_kink()` sits 15 lines above the new tests in the same file and is already driven by `test_the_coupled_variable_follows_the_decided_side`. Running `release_all` against it would reach the other membership class, and it would also settle the gap the new docstring concedes in prose — *"the neighbor coupling a clamp cannot repair needs a coupled fixture and is documented on `estimate()`"*. On `coupled_kink`, `mode="linear"` with `clamp=True` on the holding side should clamp `x` to 0 while leaving `y` at the released value (`2·(-1)+1 = -1`) rather than the true `1`. That is the one place in this PR where `release_all` is *worse* than `directional`, it is what the book page promises, and it is asserted nowhere — while the fixture that would assert it is already in the file.

## R3 — `test_release_all_is_repaired_by_every_mode` cannot fail for the right reason

Its holding-side assertion is `down[m.x] ≈ 0.0`, and the clamp (or the pins, or the walk) produces exactly `0.0` from *any* direction that overshoots the bound — a released direction of −37 passes identically. The only assertion carrying discriminating signal is `test_release_all_takes_the_released_direction_undecided`, with `clamp=False`, and only for `mode="linear"`. A `clamp=False` variant per mode, or a neighbour-coordinate assertion on the coupled fixture, would give the test something it can fail on.

## S1 — the new function is `solve_released` re-implemented

`parametric_step_release_all` (solver.rs:1148–1183) duplicates `SensBacksolver::solve_released` line for line. `solve_released_inner` (algorithm_backsolver.rs:692–712) already does exactly this: empty-released → `self.solve(rhs, lhs)`; otherwise `released_sigma_x` + `solve_released_prebuilt(released, sigma, None, None, rhs, lhs, false)` — same `shift = false`. The body reduces from ~25 lines to:

```rust
let released: Vec<usize> = weak.iter().map(|w| w.row).collect();
let mut d = vec[0.0; state.backsolver.dim()];
if !state.backsolver.solve_released(&released, &rhs_plain, &mut d) {
    return Err(SolverError::BacksolveFailed);
}
Ok((d[..ctx.n_x].to_vec(), released.len()))
```

The `_prebuilt` form exists so repeated solves against one released operator can share a single `Rc<sigma>` across the factorization cache — its own doc comment says so, and that is why `parametric_step_directional` uses it. Here the `Rc` is built, used once, and dropped, so it buys nothing, while duplicating the `shift = false` rationale comment that now has to stay in sync with the directional copy. The only thing lost is the specific `"release_all: released sigma unavailable"` message, which collapses to `BacksolveFailed`.

## R4 — `_correct()`'s docstring names only two of the three options

sens.py:1236–1245 documents the multiplier mismatch for `degeneracy="directional"`. Under `release_all` the same thing happens and is sharper: `parametric_step_full` supplies bound multipliers from the *held* factorization at exactly the bounds `release_all` just released to zero, and `release_all` releases the maximal weak set rather than directional's undecided subset. Since the central measurement in this PR is corrector-budget parity, this interaction is load-bearing and should be named.

## R5 — `degeneracy_iter` is silently inert under `release_all`

On all three entry points. The module already warns for inert arguments in this exact shape (`bound_eps` with `mode != "fix_relax"`, line 1391; the `max_iter` deprecation at 1915), so the precedent is established. Meanwhile `estimate()`'s docstring at 1299 still reads "degeneracy_iter budgets those backsolves" with no carve-out, and `active_set_changes()`'s at 2090 says "degeneracy and degeneracy_iter match `estimate()`'s arguments."

## Minor

- **M1.** The body says "Four tests in `test_degeneracy.py`"; there are three.
- **M2.** `test_release_all_at_a_clean_base_point_is_the_plain_step`'s docstring says "every degeneracy option takes the same step" but compares only `release_all` against `one_sided`; `directional` is untested there. It also partly duplicates `test_a_clean_base_point_is_identical_under_both_settings` (line 200).
- **M3.** The CHANGELOG entry covers the Python option but not the new public Rust API `Solver::parametric_step_release_all` on `pounce-sensitivity`, which ships to crates.io.
- **M4.** (checklist entry 8) The body reports counts from "this Windows machine" but does not state which extension module the `pyomo-pounce` run loaded, or that it was built from `54677ec`. `python/tests/conftest.py`'s staleness guard does not cover the pyomo-pounce suite, so one line confirming the `.pyd` is from this head would discharge it.
- **M5.** Under `mode="fix_relax"` and `mode="path"`, `release_all` never calls the new function — it reuses the pre-existing `*_decided` study surfaces with `[]`. Not a defect, but it means both the new code's blast radius and its entire test debt sit in `mode="linear"`.

## What checks out

The dispatch is wired correctly on all three entry points with no missed validator (`gradient()` takes no `degeneracy`). I verified the pyo3 argument order against the call sites: `parametric_step_bounded_decided` is `(pins, deltas, held_var_rows, max_iter, bound_eps)` and `parametric_step_path_decided` is `(pins, deltas, held_var_rows, max_iter=16)`, both matching. The `weak.is_empty()` clean-base-point path is right. `d[..ctx.n_x]` is var-x throughout, matching `scatter_x`. And the `shift = false` choice is correct and does match `path_direction`'s all-released solve, so `linear`'s starting direction really is the same one `fix_relax` and `path` refine — the framing in the book page holds.

I did not verify the `SensStdStepCalc.cpp` claims about sIPOPT; that is orientation prose rather than code under review.

---

devin-griff:
All of it applied, in `920ce5b6` (which also carries the merge of #839's main).

**S1.** Taken: the entry is `solve_released` over the weak rows, the empty-set branch and the shift rationale now living in one place.

**R1.** The invariance legs gain the released row in every frame: the all-released map equals the releasing side's exact derivative on both signs, unmoved by the change of variables, the magnitude span, and a fixed variable ahead of the kink. The oracle pins that the corrector's fixed point does not depend on which start it was handed, and pins the non-effect you asked for: on the strongly active Release fixture the option releases zero rows and returns the plain answer to refinement noise.

**R2 and R3.** The coupled kink now carries the discriminating per-coordinate pins, with re-solve-verified constants: the releasing side equals the directional decision in both coordinates in every mode, `fix_relax` repairs both coordinates on the holding side, and `linear`'s clamp repairs only the crossing coordinate while the neighbor keeps the released coupling at 1 - 10/7, the case the book promises. One measured surprise you should see: `path` under an all-released start returns the crossing coordinate to its bound but leaves the neighbor at the one-sided 2/7 rather than the re-solve's 1, even though its releasing side is fully released. That looks like segment bookkeeping in the decided path surface when a release returns at fraction zero, it predates this PR's dispatch, and I pinned it as current behavior rather than endorsing it.

**R4, R5.** `_correct()`'s docstring names the release_all multiplier mismatch, and `degeneracy_iter` warns when passed under an option that makes no decision, on all three entry points, with the docstring carve-out.

**M1 to M4.** The body counts five tests and states that the new pyomo tests call the new entry through the extension, so a stale build fails loudly. The clean-point test compares all three options. The CHANGELOG names the public Rust API.

263 on `pounce-sensitivity`, 487 on `pyomo-pounce`, `python/tests` green to the platform knowns.

jkitchin (commented):
Verified the response against `eae23f6`. S1 is exactly the reduction proposed. R2/R3 are the strongest part — the coupled test now pins distinct per-coordinate constants that a broken direction cannot accidentally satisfy, and the arithmetic checks out (the objective reduces to `(x−p)² + 0.4x²`, so `dx/dp = 5/7`). R4, R5 and M1–M3 all done; `degeneracy_iter=None` as the sentinel is the right way to tell "passed" from "default". I ran `cargo test -p pounce-sensitivity --test sens_invariance_legs --test sens_resolve_oracle` myself: 17 + 11 pass.

Two follow-ups, both with patches below, and thank you for surfacing the `path` result rather than quietly pinning it.

## 1. The oracle row does not read the outside number — and when it does, your claim lands

`leg_oracle_the_released_start_reaches_the_same_correction` lives in `sens_resolve_oracle.rs` but never calls `oracle(...)`. It compares `correct_step` from one start against `correct_step` from another — both numbers this crate produced, which is the class the file's own header says it cannot catch. The `Fx::Release` half is a genuine discharge (`assert_eq!(count, 0)`); the Kink half was one line short, since `base(Fx::Kink)` already binds `_seed` unused.

I measured it. On the Kink fixture, `floor` 5.4e-5, `width` 3.0e-5, budget 2.17e-4:

| delta | raw released | corrected |
|---|---|---|
| −1e-1 | 1.099e-1 | 3.372e-10 |
| −1e-2 | 1.094e-2 | 1.134e-8 |
| +1e-2 | 6.356e-5 | 1.329e-9 |
| +1e-1 | 6.356e-5 | 3.372e-10 |

So the PR's headline claim is verifiable in-repo after all, with six orders of margin, and the corrected released start matches the corrected plain start to every digit. The raw column is what makes the leg non-vacuous: on the holding side the all-released step is wrong by order `delta`, three orders past the budget, and the corrector still lands six orders inside it. On the releasing side the released direction is already exact, which is the option's whole claim.

Passes as written (12 tests in the file):

```rust
/// The released start reaches the re-solve, and the raw step it
/// starts from does not.
///
/// `leg_oracle_the_released_start_reaches_the_same_correction` above
/// compares two numbers this crate produced, so it cannot see a
/// correction that is self-consistently wrong -- the class this file
/// exists for. This one reads the outside number. Measured on the
/// Kink fixture, `floor` 5.4e-5, `width` 3.0e-5, budget 2.17e-4:
///
///     delta    raw released   corrected
///     -1e-1      1.099e-1     3.372e-10
///     -1e-2      1.094e-2     1.134e-8
///     +1e-2      6.356e-5     1.329e-9
///     +1e-1      6.356e-5     3.372e-10
///
/// The second column is what keeps this from being vacuous. On the
/// holding side the all-released step is wrong by order `delta`,
/// three orders past the budget, and the corrector still lands six
/// orders inside it. On the releasing side the released direction is
/// already exact, which is the option's whole claim.
#[test]
fn leg_oracle_the_released_start_reaches_the_resolve() {
    let (s, seed) = base(Fx::Kink);
    let (_mu, width) = mu_and_width(&s);
    let base_x = s.converged().expect("converged").x.clone();
    let floor = dist(&base_x, &oracle(Fx::Kink, &seed, 0.0));
    let budget = 4.0 * floor.max(width);
    let n_x = base_x.len();
    let at = |step: &[Number]| -> Vec<Number> {
        base_x.iter().zip(step.iter()).map(|(&b, &d)| b + d).collect()
    };

    for &delta in &[-1.0e-1, -1.0e-2, 1.0e-2, 1.0e-1] as &[Number] {
        assert!(
            delta.abs() > width,
            "the oracle owns only steps above the barrier width: \
             {delta:e} vs {width:e}",
        );
        let want = oracle(Fx::Kink, &seed, delta);
        let (released, count) = s
            .parametric_step_release_all(&[0], &[delta])
            .expect("released step");
        assert!(count > 0, "the kink fixture's bound is in the weak set");

        let mut start = s.parametric_step_full(&[0], &[delta]).expect("full step");
        start[..n_x].copy_from_slice(&released);
        let (corrected, _) = s
            .correct_step(&[0], &[delta], &start, 8)
            .expect("corrector");
        let err = dist(&at(&corrected[..n_x]), &want);
        assert!(
            err <= budget,
            "delta={delta:e}: the corrected released start must reach the \
             re-solve: off by {err:e}, budget {budget:e}",
        );

        // and the raw start it came from does not, on the side the
        // perturbation holds -- otherwise the assert above is free
        if delta < 0.0 {
            let raw = dist(&at(&released), &want);
            assert!(
                raw > 10.0 * budget,
                "delta={delta:e}: the raw released step should be wrong on \
                 the holding side, else the correction proves nothing: \
                 off by only {raw:e}",
            );
        }
    }
}
```

Keep the existing start-independence leg alongside it; they say different things.

## 2. The `path` result is documented only where nobody will look

`docs/src/sensitivity.md` is untouched by `eae23f6`. It still reads "`path` walks it and records a return to the bound along the path," and singles out only `linear` as lossy. A user who reads that page and picks `path` + `release_all` on a coupled model gets a wrong neighbour with nothing warning them; the defect currently lives in a test docstring and a commit message.

I read `step_along_path` to characterise it, and I think it is a bit sharper than "segment bookkeeping at fraction zero". The seeded-state comment at `boundcheck.rs:1320` says the walk expects a **decided** working set: rows the direction holds arrive released-and-pinned, rows it leaves go into the base-activity table for the release scan. `parametric_step_path_decided` puts everything not in `held_var_rows` into `forced_active`, i.e. the released list — so under `release_all` every weak row is declared a leaver, including ones the perturbation holds. The walk returns the crossing coordinate but never re-optimizes behind it, which is exactly what `fix_relax` does differently. Contract mismatch rather than bookkeeping.

That also bears on "predates this PR's dispatch". True of the surface, but the failing combination is empty-held **and** a bound the perturbation holds, and `directional` cannot produce it — it holds exactly those rows. Before this PR, `estimate(mode="path")` reached empty-held only via `directional` on the releasing side, where the answer is right. So this PR is what puts a user on it.

Filed as [#852](https://github.com/jkitchin/pounce/issues/852) with the options; nothing there blocks this PR. One cheap check worth running while you have a build, since it decides how old the defect is: does `estimate(m, [(m.p, -1.0)], mode="path", degeneracy="one_sided")` on the coupled kink give `y = 1` or `y = 2/7`?

For the book page, after the "accuracy-first choice at small kink counts remains the default" sentence:

> On a **coupled** model the repair is only as good as the mode's reach, and the three differ. Measured on the coupled kink of `pyomo-pounce/tests/test_degeneracy.py`, holding side, exact answer `x = 0`, `y = 1`:
>
> | mode | `x` | `y` |
> |---|---|---|
> | `fix_relax` | 0 | 1 |
> | `linear` | 0 (clamped) | -3/7 |
> | `path` | 0 | 2/7 |
>
> `fix_relax` pins the crossing and re-solves, so it repairs the neighbour too. `linear` clamps the crossing coordinate only, and the neighbour keeps the released coupling -- that is the documented trade. `path` returns the crossing coordinate to its bound but leaves the neighbour at the one-sided value, which is **not** the trade the option advertises: `step_along_path` seeds its state from a *decided* working set, and `"release_all"` hands it an undecided one, so a row the perturbation actually holds still enters the walk as a row that leaves. Prefer `fix_relax` with `"release_all"` on a coupled model until that is resolved (gh#852).

## Nits

- Stray whitespace run in an assert message: `"with nothing weak the released step is the plain                  answer"`.
- `degeneracy_iter` moving from `16` to `None` in three public signatures is a visible API change worth a CHANGELOG clause, even though the effective default is unchanged.

Both patches are on branch `claude/pounce-pr-842-review-fbbofg` in the base repo (commit `69a1058`) if pulling them is easier than pasting.

---

devin-griff:
All of it pulled: `69a1058` as-is, plus `3a548c36` for the two nits, the assert message's whitespace run and a CHANGELOG clause naming the `degeneracy_iter` signature change.

I ran the dating check, and the defect predates this PR and is older and wider than #852's analysis, the outcome its first-step check anticipated. `path` with `degeneracy="one_sided"` on the coupled kink's holding side also returns y = 2/7, and it never touches the decided surface: the dispatch calls the plain `parametric_step_path` there. With warnings forced on, the `one_sided` and `release_all` runs both emit "path applied 0 active-set change(s) and still leaves the bounds for ['x']... The values were clamped", and only `directional` runs clean at y = 1. So on both surfaces the walk records nothing for the held row and the clamp is what returns the crossing coordinate, not the walk. The common condition is a held weak bound with no directional decision, which `one_sided` has produced on main all along. Details on #852.

Crate suite green with the oracle file at 12, `pyomo-pounce` at 494 plus the three Windows sh-fixture knowns, `python/tests` green to its knowns.

Issue #780, opened 2026-08-25
Epic: POUNCE-enabled chemical-engineering applications and tutorial series


bernalde:
> Migrated upstream from [bernalde/pounce#5](https://github.com/bernalde/pounce/issues/5). This issue is the authoritative tracker; the fork issue is retained as provenance.

## Objective

Coordinate the four chemical-engineering application cases that support a tutorial series and a possible paper on calculations that POUNCE makes substantially less cumbersome to formulate, solve, differentiate, validate, and diagnose.

The unifying claim should be stronger than “POUNCE can solve these NLPs.” Each case must demonstrate a workflow that historically required several specialized numerical components—continuation, BVP/collocation codes, sensitivity or covariance machinery, active-set logic, MPCC regularization, or mixed-integer regime handling—and show how POUNCE makes the complete calculation reproducible in one solver-centered environment.

The detailed literature, model requirements, validation plans, and case-specific acceptance criteria live in the child issues. This epic tracks upstream implementation, dependencies, shared standards, and publication-level deliverables.

## Current upstream portfolio

- [x] #773 — **Differentiable phase-envelope tracing and inverse mixture design**
  - Implementation: #777 (merged 2026-08-26).
  - Refined folds, verified extrema sensitivities, independent published-mixture validation, and inverse composition design.
- [x] #774 — **Closed-loop active-set-aware advanced-step NMPC for a constrained CSTR**
  - Implementation: #778 (merged 2026-08-26).
  - Complete controller/plant loop, active-set events, guarded correction, fallback/recovery, and closed-loop evidence.
- [x] #775 — **Uncertainty-aware nonisothermal catalyst-pellet inverse design**
  - Implementation: #779 (merged 2026-08-26).
  - Validated distributed model, simultaneous design, covariance-derived scenarios, robust redesign, and an executed tutorial.
- [ ] #776 — **Phase-changing dynamic operations as MPCCs with POUNCE–DISCOPT validation**
  - Cross-repository design record: [jkitchin/discopt#1123](https://github.com/jkitchin/discopt/issues/1123) (**closed**); the live prerequisites are [#1147](https://github.com/jkitchin/discopt/issues/1147) then [#1148](https://github.com/jkitchin/discopt/issues/1148).
  - First semantics issue: [jkitchin/discopt#1124](https://github.com/jkitchin/discopt/issues/1124) — **complete** (DiscOpt PRs #1128 and #1146).
[...]

Issue #776, opened 2026-08-25
Research/tutorial case: phase-changing dynamic operations as MPCCs with POUNCE–DISCOPT validation


bernalde:
> Migrated upstream from [bernalde/pounce#4](https://github.com/bernalde/pounce/issues/4).

## Cross-repository coordination

This application is coupled to the first-class MCP/MPEC and continuous-reformulation design discussion in [jkitchin/discopt#1123](https://github.com/jkitchin/discopt/issues/1123). POUNCE should own the solver benchmark, continuation, diagnostics, and phase-change validation work; DiscOpt should own the modeling semantics, source-level complementarity residuals, local-versus-certified result contract, and reduced GDP/SOS1 comparisons.

The first three application tutorials already have implementation PRs. This MPCC workstream should remain gated and should not block their review. Before investing in the dynamic column, the two repositories need a coordinated decision on the Gate 0 benchmark set, continuation/warm-start contract, and the boundary between solver-owned and modeling-layer deficiencies.

## Objective

Build and validate a POUNCE workflow for **phase-changing dynamic operations formulated as mathematical programs with complementarity constraints (MPCCs)**, then connect reduced versions to exact disjunctive/MINLP formulations in [DISCOPT](https://github.com/jkitchin/discopt).

The motivating process is cold startup or shutdown of batch/tray distillation, where trays dry and wet and phases appear/disappear without changing the model dimension. A sub-ambient CO₂ separation/compression train is a compelling extension. The research question is whether POUNCE's smooth local NLP machinery, exact-penalty wrapper, continuation, and diagnostics can make these historically specialized formulations reproducible—while DISCOPT supplies exact GDP/SOS1 regime decisions and global branch-and-bound checks on reduced cases.

This is a **high-risk gated case**. Do not begin with a full column or large CO₂ train. First validate the MPCC route on a representative benchmark ladder and demonstrate that the reported stationarity and phase states are physically meaningful.

## Why this case matters

Phase appearance is awkward in equation-oriented dynamic models. A formulation that only instantiates equations for present phases changes dimension at switching events; explicit on/off variables create a mixed-integer dynamic optimization problem; and smoothing or ad hoc clipping can converge to false/trivial phase states. Complementarity encodes “phase amount is zero or the phase-equilibrium/stability condition is active” while retaining a fixed-dimensional simultaneous transcription.

[...]

Issue #10, opened 2026-05-14
Implementation plan: Thierry-Biegler ℓ₁-exact penalty-barrier wrapper + MPCC paper reproduction


jkitchin:
## Goal

Port the Thierry & Biegler (2020) ℓ₁-exact penalty-barrier wrapper from ripopt#23 / commit `7847bba9` into pounce as `pounce-l1penalty`, and ship the **MPCC paper reproduction** that ripopt explicitly deferred (their probe was on the adjacent CUTEst NE-suffix slice). End state: a default-off opt-in wrapper plus an opt-in auto-fallback, validated on (a) pounce's own CUTEst sweep, (b) the same 42-problem NE-suffix probe ripopt ran, and (c) the MPCC benchmarks Thierry & Biegler 2020 report.

## Background

See pounce#5 for the algorithmic background. Algorithmic reference: [`L1PenaltyBarrierNlp` in ripopt#23 commit `7847bba9`](https://github.com/jkitchin/ripopt/commit/7847bba9), 550 LoC wrapper. The ripopt closing comment on issue #23 documents 2 hard rescues (`DENSCHNENE`, `HATFLDFLNE`), 26 honest-infeasibility upgrades (`RestorationFailed` / `MaxIterations` → `LocalInfeasibility`, often turning 3000-iter budget exhausts into 50–200-iter certificates), and 0 regressions on a 42-problem CUTEst NE-suffix probe; flag-off byte-identical on a 100-problem comparison (99/100; sole diff is wall-time non-determinism on `CRESC132`).

## Architecture decisions

**Crate placement.** New crate `crates/pounce-l1penalty/`. Depends on `pounce-common`, `pounce-nlp`. Used by `pounce-algorithm` (for the IPM driver hook) and `pounce-cli` (for option wiring). Mirrors how `pounce-restoration` is structured.

**Wrapper level.** Wrap at the `TNLP` level — `crates/pounce-l1penalty/src/wrapper.rs` exposes `L1PenaltyBarrierTnlp<'a>(inner: Rc<RefCell<dyn TNLP>>, opts)` implementing `TNLP`. Rationale: the augmented NLP feeds straight into the existing `TNLPAdapter → OrigIpoptNlp` chain unchanged, which is the cleanest separation. ripopt wraps at `NlpProblem` (its analog), not at the lower algorithm-side `Nlp` trait — same call here.

[...]

PR #844, opened 2026-08-29
MPCC Gate 0: the benchmark harness, the gate report, and the ℓ₁ reporting defect it found

### LINKED ISSUE #794: Gate 0: build the MPCC benchmark harness and establish the supported solver route
Parent: #776  
Epic: #780  
Cross-repository design dependency: [jkitchin/discopt#1123](https://github.com/jkitchin/discopt/issues/1123)  
DiscOpt semantics prerequisite: [jkitchin/discopt#1124](https://github.com/jkitchin/discopt/issues/1124)  
Focused successor to the deferred MPCC-reproduction phase of #10.

## Objective

Build the smallest reproducible MPCC benchmark harness that can establish a supported POUNCE route and a documented failure boundary before work begins on a phase-changing flash, dynamic tray model, or new solver API.

The benchmark is the evidence gate. It must distinguish deficiencies in the source formulation, DiscOpt lowering/acceptance policy, and POUNCE's inner NLP algorithm instead of assigning ownership from symptoms alone.

## Benchmark ladder

Use a manifest with expected solution/stationarity information and include at least:

- a regular case with strict complementarity;
- a biactive case;
- a degenerate case;
- an infeasible or expected-failure case;
- a symmetric Boolean-selector or branch-change case; and
- a small pinned MacMPEC-style subset suitable for comparing solver strategies.

The same source model, scaling, initial point, and stopping requirements must be used across comparable configurations.

## Required configurations

- Direct POUNCE NLP formulation.
- Explicit Scholtes continuation using independent solves with cold starts.
- Scholtes continuation with primal-only warm starts.
- Scholtes continuation with complete primal/dual/barrier warm starts and a documented restart ladder.
- Exact product/NCP equality where applicable, comparing ordinary POUNCE with the opt-in `l1_exact_penalty_barrier`, restoration fallback, and `auto_l1` routes.
- A pinned CCOpt integrated-continuation comparison as an optional benchmark oracle, not a required runtime dependency.

Run the relevant POUNCE kill-switch controls before attributing a failure to a new mechanism. In particular, test disabling presolve, acceptable-level termination, scaling, and POUNCE-only heuristics where they could explain the outcome.

## Result contract

Every record must stamp both repository commits, model-data revision, environment, scaling/tolerances, initial point, and configuration. Report at least:

- termination status and objective;
- original-space feasibility;
- source complementarity product and min/max residuals;
- MPCC stationarity classification/evidence;
- POUNCE's ordinary scaled and unscaled NLP/KKT diagnostics, kept separate from source complementarity;
- iterations, outer stages, accepted/rejected stages, and warm-start/restart level;
- `mu`, the outer relaxation parameter, and their update reasons;
- restoration transitions, filter resets, factorization/inertia-correction counters where available; and
- active-pair/mode changes and sensitivity to initialization/scaling.

## Acceptance criteria

- [ ] `benchmarks/mpcc/` contains a manifest, runner, schema, deterministic smoke subset, and reproducible full configuration.
- [ ] Each benchmark class above has expected behavior and a source-level validation function.
- [ ] The required POUNCE configurations run from the same source cases and initial conditions.
- [ ] CCOpt is pinned precisely if included; comparisons disclose materially different linear algebra.
- [ ] A report states the supported route/default candidates and a documented failure boundary.
- [ ] The report identifies which observed gaps belong to DiscOpt, POUNCE, or remain unassigned pending a reproducer.
- [ ] No full flash, tray, or column work begins from this issue.

## Issue-splitting rule

Do not pre-create speculative POUNCE feature issues. A persistent parameterized session, problem-change/filter-reset hook, source-residual globalization hook, pair-structured KKT regularizer, or guarded endgame callback becomes its own issue only when this harness provides:

1. a minimal source model and initial point;
2. commit-stamped baseline and comparator measurements;
3. kill-switch evidence excluding existing mechanisms; and
4. a measurable acceptance criterion for the proposed solver change.

If the representative cases are not repeatably reliable, the valid outcome is a documented stop decision before Gate 1.

jkitchin:
Closes #794. Gate 0 of #776; epic #780.

@bernalde — you filed #794 and its parent, so this is yours to look at first. The two things I'd most like your judgement on are flagged inline below: **(a)** whether the recommended route matches what you expect a flash model to need, and **(b)** the disposition of P2, which is **open** rather than settled.

> **Four review rounds addressed.** Round 1 (`27629e8`): 5 blocking, 1 nonblocking, all fixed in `07f7ea9`. Round 2 (`27a7f081`): an Ipopt 3.14.19 comparison plus a diff read — one code defect and two falsified report claims, fixed across `a26bb0c9`/`cf21e028`/`c67615b4`. Round 3 (`c67615b4`): a stale ownership line that contradicted the corrected P2 section, and the missing regression for the fail-closed reset — both in `2ec5af41`. Round 4 (`464e1256`): an adversary pass against the P1 fix itself, which **found a regression this branch introduced** — see below. **P2's disposition changed twice and is now marked open** — see "(b)".

## Problem

POUNCE has listed an MPCC benchmark sweep under `benchmarks/mpcc/` as outstanding since #10, and #794 asks for it as the evidence gate in front of any flash/tray/column work: the smallest reproducible corpus that can tell a **source-formulation** deficiency from a **lowering** deficiency from a **POUNCE** deficiency, instead of assigning ownership from symptoms.

Building it immediately found a solver defect. With `l1_exact_penalty_barrier=yes` on the `ralph1` NCP-equality lowering:

| | status | objective | reported violation | actual violation of the declared row |
|---|---|---|---|---|
| before | `Solve_Succeeded` | **-5.0e-04** | 9.6e-15 | **2.5e-07** |
| after | `Solve_Succeeded` | -5.1e-05 | **2.6e-09** | 2.6e-09 |
| oracle (branch enumeration) | — | 0 | — | 0 |

Both `final_constr_viol` and `final_unscaled_constr_viol` reported `9.6e-15`, so **no field in the result disclosed it**, and the objective is below the true optimum — reachable only off the feasible set.

## Cause

The wrapper solves an augmented problem, `c(x) − p + n = target` with `p, n ≥ 0`, whose equality rows the slacks satisfy to machine precision *by construction*. That residual was what the statistics carried. And the exit verdict was argued from `Σ(p+n)` against `l1_slack_tol`, which is the wrong quantity twice over:

- the violation of equality row `i` is `|pᵢ − nᵢ|`, not `pᵢ + nᵢ`, and at the barrier's interior both slacks stay positive where their difference is zero;
- `l1_slack_tol`'s `1e-6` is four orders looser than a `tol = 1e-8` solve asked for, so a violation the strict gate would refuse on the *unwrapped* problem read as "the constraints are satisfied".

## Fix

In `run_l1_penalty_outer_loop`:

- **Report the user's violation, in the field family it belongs to.** New `original_space_feasibility` measures the inner TNLP's own rows and bounds at the returned point. That measurement is in the model's **original units**, so it is reported in `final_unscaled_constr_viol` and folded into `final_unscaled_kkt_error` — the family `SolveStatistics` and `docs/src/python.md` define as original-unit. The scaled family mirrors it only when per-row scaling did not engage, which is exactly the case in which the contract requires the two to agree. The aggregates take a `max` rather than a rewrite.
- **Judge it by the caller's tolerances.** Both the ρ-escalation loop's termination test and the exit verdict now use that measurement, via `is_negligible` at `tol` and `acceptable_tol` — **and, since round 4, also against the absolute `constr_viol_tol` / `acceptable_constr_viol_tol` the rest of the solver judges feasibility by.** `Σ(p+n)` keeps its real job — the Byrd-Nocedal-Waltz steering signal for ρ — and `l1_slack_tol` survives as the fallback when a model's rows cannot be evaluated, so "not measured" never reads as "feasible". An `eval_g` that *fails* takes that same fallback: the success flag is honoured at both call sites, so a zero-filled buffer is never read back as a violation of zero.

Downgrade-only: a solve that meets the strict standard on the user's rows keeps the status the inner gave it, and nothing here turns a failure into a success. **That claim was false as first written** — for rows of large magnitude it did turn a failure into a success, which is round 4's finding — and is restored by the absolute floor described there.

**Rejected first:** judging the measured violation against `l1_slack_tol` (i.e. fixing only the quantity, not the standard). Also rejected: converting the measurement into the *scaled* space, which needs the augmented problem's row-scale factors, and `optimize_constrained` has dropped that NLP by the time the outer loop measures. The consequence is stated rather than hidden: **under an active `nlp_scaling_method` the scaled fields still describe the augmented problem**, and `final_unscaled_constr_viol` is the field to read.

## Blast radius

**Confined to solves that set `l1_exact_penalty_barrier=yes` or `l1_fallback_on_restoration_failure=yes`.** Both default off, and the CLI fixture corpus never sets either, so `scripts/sweep-fixtures.sh` is bit-identical by construction rather than by measurement. `original_space_feasibility` has exactly two call sites, both inside the ℓ₁ outer loop, so round 4's change inherits the same confinement.

That confinement is now **measured independently**: in review round 2 @bernalde built binaries from this branch and from `main` and ran all four P2 cells through each — bit-identical on every reported field, status, iteration count and both columns. Better evidence than my argument from option defaults, and it also establishes that the acceptable-level exit discussed under (b) is pre-existing rather than introduced here.

Within the ℓ₁ path the loop does now escalate ρ where it used to stop at the penalty point, so trajectories on that path move, and answers improve: `ralph1` under `ncp_eq_l1_fallback` reaches the true optimum with the complementarity product exactly 0, where it used to stop at `-4.3e-10`.

**One change outside that radius**, and it is test-only: `crates/pounce-restoration/tests/restoration_deadline.rs` (see round 2 below).

## Tests

- **1025 tests pass locally** across `pounce-algorithm`, `pounce-l1penalty` and `pounce-restoration`, including all 11 ℓ₁ integration tests; the 4 CLI ℓ₁-fallback tests pass in `pounce-cli`. (The round-4 comment on this PR says 1024; that was the count taken before its own regression test was added. 1025 is current.)
- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets` zero errors.
- `scripts/check-docs-consistency.sh` and `scripts/check-release-consistency.sh` both OK.

**The P1 regression, and the fact that it fails on the parent.** `crates/pounce-algorithm/tests/issue_794_l1_original_space_reporting.rs` is the `ralph1` exact-product reproducer as a Rust TNLP. Run against the parent it reports `9.636e-15` against a true violation of `2.501e-7`, and returns `Solve_Succeeded` at a `2.500e-7` violation under `tol = 2.501e-11` — the table at the top of this PR, reproduced by the test rather than quoted at it. Five tests now, each carrying the mutation that reddens it, and **every row of that table was run**: each mutation reddens the named test and only that test.

**How you know the harness's own tests bite.** `benchmarks/mpcc/selftest.py` is fourteen checks, and six are mutation-checked against the exact defect they exist for:

| check | mutation |
|---|---|
| classifier discriminates S / M / C | make it read one least-squares multiplier instead of searching the set → `ralph1` and `scholtes4` report C where the derivation says M |
| pair activity threshold is corner-aware | restore the fixed `1e-6` → a point inside `sqrt(tol)` of `qpec_small`'s corner reads `none` |
| validators are scale invariant | restore the selector's coordinate test → 72 spurious failures on the skew leg |
| manifest is current | it caught its own staleness twice during this work |
| smoke rejects false success on the infeasible case | restore the near-feasible-only rule → a success at a large source violation passes |
| smoke rejects an all-failed run | delete the vacuity guard → a run in which nothing solved reports a pass |

**Deliberately not pinned**: neither `mpcc-selftest` nor `mpcc-smoke` is wired into CI, matching how every other benchmark harness here is treated.

## Review rounds

**Round 1** — 5 blocking, 1 nonblocking, all fixed in `07f7ea9`:

| # | finding | resolution |
|---|---|---|
| B1 | `smoke_checks` permitted the false success its comment forbade, and an all-failed run passed vacuously | any success on `infeasible_pair` now fails; vacuity guard and a coverage property for the supported route |
| B2 | the emitted artifact carried 652 literal `NaN` and 160 records with `source: {}`, and exited 0 | non-finite → `null`, `allow_nan=False`, schema conditional on whether a point was returned, payload validated in every mode |
| B3 | `eval_g`'s success flag discarded at the exit-verdict site | gated as the ρ-escalation measurement already was |
| B4 | the original-units measurement written into both field families, breaking the units contract | measurement to the unscaled family; mirrored into the scaled one only when row scaling did not engage |
| B5 | no committed test observed the P1 behaviour | added the reproducer above, which fails on the parent |
| NB6 | ownership total said 549 where its table sums to 576 | recomputed; the generated table now emits the total |

B3 and B4 were defects in the fix itself. **What B4 cost, measured:** 148 records move `nlp.final_constr_viol` (84 also `final_kkt_error`), confined to the two ℓ₁ routes; status, objective, `x`, every source quantity, stationarity and iteration counts identical across all 2880, triage unchanged in every bucket.

**Round 2** — an Ipopt 3.14.19 comparison plus a diff read:

| finding | resolution |
|---|---|
| `row_scaling_active` is fail-open — never reset, so a stale `Some(false)` from a previous inner attempt lets the outer loop mirror into the scaled family | cleared at `optimize_constrained` entry (`a26bb0c9`); same class as B4 |
| `judge` returns a `bool` every caller discards | dead return dropped |
| the report's `recalc_y` framing is too narrow — the false success is reachable at default options | widened to the normalised gate itself |
| "the gh#274 gate is refusing these correctly" is contradicted by a `7.9e+04` exit | reproduced bit-for-bit locally and removed; the replacement is in (b), and it is **not** the one round 2 proposed |

Two things came out of reproducing those measurements rather than out of the review itself:

- **The `restoration_deadline` flake is fixed at its root cause** (`cf21e028`), not re-run. It had failed twice at exactly `31 × 3 = 93` against a baseline of `89`. The cause was not runner load: the test inferred "the deadline was honored" from an *iteration ratio*, which assumes restoration inner iterations cost uniformly. Measured, a twentieth of the time buys a twenty-second of the iterations locally and a third of them on a GitHub runner — so the margin was machine-dependent for a reason unrelated to the deadline, and it failed while the deadline was working (the tight leg did exit `MaximumWallTimeExceeded`). It now asserts the contract `max_wall_time` makes, in seconds: the solver's own `total_wallclock_time_secs` against its budget, allowance `10×` against a measured `1.10–1.12×` overshoot. Mutation-checked — disabling `IpoptAlgorithm::deadline_status` at all six call sites reddens it, on the *status* assertion, which was already present and is kept. The `3×` margin was never what caught the defect, only what made it flaky. Also recorded in the test: on a fast machine the per-inner-iteration check in `RestoConvCheckAdapter` is never reached, so the module doc overstates the test's reach — and did before this change.
- **The `7.9e+04` exit is the documented contract, not a defect** (`c67615b4`) — see (b).

**Round 3** (`2ec5af41`) — both findings addressed:

| finding | resolution |
|---|---|
| **Blocking:** the ownership table still read "P2 — since confirmed against Ipopt 3.14.19, which solves the reproducer POUNCE fails", the one-start conclusion `a26bb0c9` retracted, two hundred lines from a P2 section that marks it open | corrected to **open**. I wrote that line and missed it when retracting; the file now says the same thing in both places, and the rest of it has been swept — the only surviving occurrences of the retracted phrasing are inside the narrative describing it as superseded |
| **Nonblocking:** the fail-closed `row_scaling_active` reset had no regression | it does now, and my stated reason for not writing one was wrong: I was thinking of an integration test under `tests/`, which cannot observe a private field. A unit test *inside the module* can |

That regression took three non-obvious things to reach the path, recorded because each would silently make it test nothing: the fixture needs `eval_h` (`Hs071Stub` has none, so the solve lands on L-BFGS and panics before recording); it must not be convex (`solver_selection=auto` routes a convex model to the convex engine, which never reaches the statistics block, so `ConvexEqTnlp` records no verdict at all); and the second solve must decline via `get_bounds_info`, since refusing `get_nlp_info` mid-flight panics rather than unwinding. Mutation-checked: removing the reset fails it with `left: Some(false), right: None`.

**Round 4** (`464e1256`) — an adversary pass against the P1 fix, which found a regression **this branch introduced**. Full detail in [the round-4 comment](https://github.com/jkitchin/pounce/pull/844#issuecomment-5494140112).

P1 judged each row with `is_negligible(viol, scale, tol)` alone — `viol ≤ tol · max(|row|, 1)` — a relative test with no absolute floor, so the allowance grows without bound with the row's magnitude. On a model **infeasible by exactly 50 by construction** with its row at `K`, holding the violation fixed and sweeping only `K`, the verdict tracked `tol · K` exactly: `Infeasible_Problem_Detected` at `K = 1e6`, `Solved_To_Acceptable_Level` from `1e8`, and **`Solve_Succeeded` from `5.1e9` up**.

On the parent that same model exits `Error_In_Step_Computation` — the `Σ(p+n) > l1_slack_tol` argument P1 replaced was crude but *absolute*. So P1 as first written traded a dishonest number under an honest status for an honest number under a dishonest status.

The repair uses the standard the solver already has rather than inventing one: `OptErrorConvCheck::primal_component_passes` is an absolute `constr_viol ≤ constr_viol_tol` whose scale-awareness comes from *abstaining* when every row sits at its own floating-point noise floor (gh#528, gh#590), not from scaling the tolerance by the row. Its doc names the property `is_negligible` lacks — that abstention "cannot fabricate a success on a genuinely infeasible model … pinned at its infeasibility gap, orders above `eps ·` the row's own magnitude". `primal_resolvable` itself cannot be reused: the CQ computes it on the augmented NLP, so it would abstain always. The arms are conjoined, not substituted — an absolute-only rule would re-admit P1 itself.

**Measured: the full sweep is bit-identical across round 4 — 0 of 2880 records differ**, no success gained or lost, so every table below stands. That null result is not reassurance and the report says so: every row in all eleven cases is `O(1)`, so the corpus is uniform in the dimension this defect acts on and could never have caught it.

## The harness, in one table

Eleven cases over the six classes #794 requires, nine routes, four kill-switch controls, two scaling legs, three starts: **2880 records**, ~190–245 s. Every expected optimum is hand-derived in the case factory's docstring *and* recomputed independently by a SciPy branch-enumeration oracle; `selftest` fails on disagreement.

**Reproducibility, measured.** The branch is kept merged up with `main` and the full sweep re-run on each new base. Across **twelve** successive bases **eleven** were bit-identical: **0 of 2880 records differed** once wall-clock timing is stripped. The one that moved changed exactly **one** cell — #869's negative-curvature probe, on a cell that reaches an axis-aligned saddle, and it improved (117 → 104 iterations, `final_compl` 14.2 → 9.1e-10, stationarity verdict and both validation verdicts unchanged). No summary table in this PR moved. The thirteenth base is #881's merge: on it the sweep was run before and after round 4's change and differs in 0 of 2880 records — that isolates round 4, and does **not** separately difference the base change itself against the twelfth base.

Cells are `(case, scaling, start)`, 60 per route, `infeasible_pair` reported separately. `below f*` counts answers *lower* than the optimum — possible only at a point the MPCC does not contain.

| route | solved | at f* | other local | below f* | median iters | worst source compl |
|---|---:|---:|---:|---:|---:|---|
| `direct` (`G·H ≤ 0`) | 58 | 47 | 9 | 2 | 43 | 1.8e-11 |
| `ncp_eq` (`G·H = 0`) | 58 | 47 | 11 | 0 | 20 | 9.7e-10 |
| `ncp_eq_l1` | 60 | 44 | 4 | 12 | 18 | 9.8e-09 |
| `ncp_eq_l1_fallback` | 58 | 41 | 11 | 6 | 18 | 9.7e-10 |
| `ncp_eq_auto_l1` | 58 | 47 | 11 | 0 | 20 | 9.7e-10 |
| `scholtes_cold` | 60 | 40 | 8 | 12 | 164 | 1.0e-08 |
| `scholtes_warm_primal` | 60 | 48 | 0 | 12 | 60 | 1.0e-08 |
| `scholtes_warm_full` | 60 | 48 | 0 | 12 | 49 | 1.0e-08 |
| **`scholtes_then_ncp`** | **60** | **57** | **3** | **0** | **79** | **4.9e-11** |

**⇦ (a) The supported route is the composition** — continuation with full warm starts to locate the branch, then one exact-product solve seeded from it. Neither half is sufficient: the continuation always converges but its answer is only ever feasible for `G·H ≤ τ`, and the exact-product solve returns an MPCC-feasible point but fails from a cold start where a pair is biactive. It also finds the *global* optimum on `regular_strict` from starts on the wrong branch, where every single-solve route stops at the local `f = 4`. Costs ~4× a single `ncp_eq` solve. **@bernalde: is a two-stage default acceptable for the flash, or does something downstream need a single solve?**

## The number Gate 1 inherits

`G·H` is quadratically flat at the corner, so a solve that drives the product to `eps` pins each side only to `sqrt(eps)` — and the objective follows. `ralph1` at a residual of `2.6e-09` sits `5.07e-05` below `f*`, against `sqrt(2.6e-09) = 5.1e-05`. **At `tol = 1e-8` an MPCC objective is good to about `1e-4`, whatever route produced it.**

So a phase regime decided on a margin smaller than `sqrt(tol)` is arithmetic, not physics, and an objective comparison against a published trajectory or a DiscOpt GDP solution cannot be quoted tighter than that without tightening `tol` first.

## Ownership of the 576 triaged observations

| owner | n | |
|---|---:|---|
| converged, nothing to assign | 331 | reached an S- or M-stationary point |
| relaxation limit | 126 | a continuation's τ-feasible answer |
| complementarity tolerance floor | 61 | the `sqrt(tol)` effect above |
| source formulation | 36 | `infeasible_pair`: failing is correct |
| scaling | 14 | converged the internally scaled NLP; unscaled KKT 3+ orders larger |
| POUNCE candidate | 8 | P2 — **open**, see (b); the label is the triage rules', not a claim about a mechanism |

**Nothing is assigned to DiscOpt, and none is unassigned.** What Gate 0 hands back to jkitchin/discopt#1123 and #1124 is a requirement rather than a defect: **a lowering that reports only the NLP's residuals is not reportable as an MPCC result.** The three quantities that must survive the lowering are the source complementarity product, the sign residuals of `G, H ≥ 0` in the model's own units, and enough of the source model to classify MPCC stationarity. `benchmarks/mpcc/schema.json` is the concrete shape, and it is deliberately solver-agnostic — @bernalde, that is the piece worth disagreeing with early if it is the wrong shape for your side.

## ⇦ (b) P2 — a biactive pair breaks a cold exact-product solve. **Open, and wrong twice already.**

This section has been rewritten three times and the first two versions were wrong in opposite directions — first "inherent to the reformulation, nothing to fix", then "a POUNCE defect, upstream solves it". Both were drawn from too little measurement, and rather than supply a third confident story the report confines itself to what has been run and marks P2 **open pending harness re-measurement**.

What is measured. Same `.nl` file, same options, exact Hessians, default `recalc_y=no`:

| start | solver | iters | exit | **unscaled** overall |
|---|---|---:|---|---:|
| `origin` | Ipopt 3.14.19 | 58 | `Optimal Solution Found` | **`7.523e-09`** |
| `origin` | POUNCE 0.11.0 | 41 | `Solved To Acceptable Level` | **`7.897e+04`** |
| `upper_left` | Ipopt 3.14.19 | — | `Optimal Solution Found` | **`5.934e-02`** |
| `upper_left` | POUNCE 0.11.0 | 76 | `Optimal Solution Found` | **`1.564e-08`** |

**Neither implementation dominates**, and both report success at points that are not stationary in the model's own units. On `origin` POUNCE reports acceptable-level at `7.9e+04` where Ipopt reaches `7.5e-09`; on `upper_left` the roles reverse. **`recalc_y` is not required** — every row is at default options; the option makes the trap easier to reach, it does not create it.

**Why POUNCE's `7.9e+04` row passes is not what round 2 proposed, and this was checked against the code rather than reasoned.** It is *not* that the acceptable-level triplet reads `s_d`-normalised quantities: the three component tests read the **unscaled** residuals (`current_is_acceptable_with_state` calls `curr_unscaled_dual_infeasibility_max` and its siblings), exactly as `acceptable_dual_inf_tol`'s help text specifies. Both halves pass on their own terms:

| test | quantity | threshold | |
|---|---:|---:|---|
| aggregate | **scaled** `8.234e-11` | `acceptable_tol` `1e-6` | passes — this one *is* `s_d`-normalised, `s_d ≈ 1e15` |
| dual component | **unscaled** `7.897e+04` | `acceptable_dual_inf_tol` `1e10` | passes on the threshold's size, not on normalisation |

Both options are implemented as registered and both defaults are upstream Ipopt's. **So there is no implementation defect at this site**: `7.9e+04` is what the contract says an acceptable point is. That does not make it harmless — it is still a status Pyomo maps into the solved family — but it relocates the question from "what is broken" to "is `acceptable_dual_inf_tol = 1e10` the right default", which is a policy choice with a blast radius of every acceptable-level exit and a deliberate deviation from upstream if changed (precedent: gh#532's `dual_inf_scale_kappa`). Refusing the point needs either threshold moved. Both are decisions, not repairs, and neither is made here.

Worth noting alongside round 4: no SOTA code judges optimality by a bare absolute bound on an unscaled dual residual — Knitro's default stopping test is relative with absolute floors enforced, and SNOPT's is `maxⱼ(Compⱼ/‖π‖) ≤ εd`. `1e10` is therefore best read as a *disabled* cap rather than a loose one, which is why the question is a policy one.

Recorded as a genuine gap rather than a policy: `kkt_fidelity_tol` (gh#173) exists for exactly this shape — "the scaled convergence test passes but the user-space duals have drifted" — and only ever downgrades `Solve_Succeeded`, so an exit that is already `Solved_To_Acceptable_Level` is outside the reach of the one guard built for it, even for a caller who opts in. Not closed here either: the status enum has no honest target below acceptable level for "feasible, but not to the fidelity you asked for", and repurposing `FeasiblePointFound` would break a public contract to fix something smaller.

Also corrected: **"all eight are one exit" is scoped to the harness's Python path.** The same model exits `Solved_To_Acceptable_Level` at 41 iterations through the `.nl`/CLI path against `Error_In_Step_Computation` at 118 through the harness. Both use exact Hessians (the harness reports 119 Lagrangian Hessian evaluations, so it is not on the limited-memory path), and the Hessian sparsity differs (6 nonzeros against 5) as the leading hypothesis, explicitly not established. The count of 8 is unaffected — it comes from the triage rules, not the exit string — but the bucket's description needs re-checking.

The corpus meanwhile is covered: `scholtes_then_ncp` clears all eight cells (complementarity `1.9e-12`, `4.9e-11`, `5.7e-28`), so Gate 0's recommendation does not depend on any of this landing first.

## Gate decision

**Proceed to Gate 1, with two conditions** — report at [`dev-notes/mpcc-gate0-report.md`](https://github.com/jkitchin/pounce/blob/claude/pounce-issue-794-scqp8r/dev-notes/mpcc-gate0-report.md). (1) Gate 1 reports source complementarity and MPCC stationarity separately from NLP residuals, and quotes nothing tighter than the tolerance floor. (2) The flash pairs are nonlinear and this corpus is not — the product row's Hessian stops being constant and the `sqrt(tol)` floor acquires the pair's own curvature, so items 1, 5 and 6 of the failure boundary need re-establishing there.

No flash, tray or column work begins from this PR; #794's last acceptance criterion stands.

---

- [x] Book page updated (`docs/src/options.md`; no new page, so no `SUMMARY.md` change — `check-docs-consistency.sh` OK)
- [x] `cargo fmt --all -- --check`, `cargo clippy` clean; `cargo test` on the changed crate + consumers locally, and the full workspace green in CI

https://claude.ai/code/session_018gyUjuq5fbXx7BjyVeuvZG


jkitchin:
`Test (ubuntu-latest)` failed on `27629e8` ([run 33416804482](https://github.com/jkitchin/pounce/actions/runs/33416804482/job/99569302048)). It is not this PR's, and there is no fix to port. Evidence, since "flake" on its own is not a diagnosis:

**The failure.** `pounce-restoration`, `restoration_deadline.rs:228`:

```
restoration_grind_honors_wall_deadline
restoration was not cut short by the deadline: tight inner iters = 31, baseline = 89 (#246)
```

**Why it is not this PR's.**

1. The diff touches no file in `pounce-restoration`. Its only compiled changes are `crates/pounce-algorithm/src/{application.rs, upstream_options.rs}`; everything else is `benchmarks/mpcc/`, docs, and CHANGELOG.
2. The `application.rs` change is reachable only under `l1_exact_penalty_barrier=yes` or `l1_fallback_on_restoration_failure=yes`, both default off. This fixture's `solve_with_budget` sets exactly one option, `max_wall_time` — so the changed code is not merely untouched by the test, it is unreachable from it.
3. It passes 3/3 locally on this head, and 3/3 on `main` at `ad9c64e`. That second run matters: #870 landed in this base and *did* touch `application.rs` and `ipopt_alg.rs`, so it was worth ruling out that restoration iteration counts had shifted under it. They have not.

**What actually happened.** The assertion is `tight_inner * 3 < base_inner`, and it missed by one iteration: `31 × 3 = 93` against a baseline of `89`, needing `tight_inner ≤ 29`. Both legs are wall-clock-timed on the same runner, and the test's own comment predicts this mode — it "only fails if machine load drifts by more than ~6× between the two runs, or if the deadline genuinely stops being honored." The deadline was honored: the tight leg did terminate on `Maximum_Wall_Time_Exceeded` (that assertion passed) and did far fewer inner iterations. Only the 3× margin came up short.

**Not fixing it here.** Hardening a load-sensitive assertion in a crate this PR does not touch would widen the PR, and the margin is a judgement call that belongs with whoever owns #246. Flagging it instead: this is a real sensitivity to runner load, and if it recurs it is worth an issue of its own rather than repeated re-runs.

I have re-run the failed job once.

---

benchmarks/mpcc/run.py
```diff
@@ -0,0 +1,330 @@
...
+    """
+    fails: List[str] = []
+
+    by_case: Dict[str, List[dict]] = {}
+    for r in records:
+        by_case.setdefault(r["case"], []).append(r)
+
+    # 1. The infeasible case must never be reported as a solved MPCC.
+    for r in by_case.get("infeasible_pair", []):
+        if not r["ok"]:
```
bernalde, line 152: Blocking: this assertion permits exactly the false-success state its comment forbids. Any `ok=True` result for `infeasible_pair` is wrong, but the check only fails when the returned point is nearly feasible; `ok=True` with a large source violation silently passes. The rest of `smoke_checks` also skips every failed solve, so a synthetic one-record-per-class set with all nine routes at `ok=False` returns `[]`.

Please reject every successful `infeasible_pair` record and assert that the supported `scholtes_then_ncp` route completes every feasible smoke case/start. Add focused mutation tests for both the false-success and all-routes-failed cases; experimental routes can retain their documented failures.

benchmarks/mpcc/run.py
```diff
@@ -0,0 +1,330 @@
...
+            "tau_schedule": list(R.TAU_SCHEDULE),
+            "restart_ladder": list(R.RESTART_LADDER),
+            "tau_bisections": R.TAU_BISECTIONS,
+            "capture_log": not args.no_log_capture,
+            "active_tol": ACTIVE_TOL,
+        },
+        "records": records,
+    }
+    with open(out, "w") as fh:
+        json.dump(payload, fh, indent=1, default=float)
```
bernalde: Blocking: the exact-head smoke command writes an artifact that violates both JSON and the committed schema while still exiting 0. Its 153-record output contains 64 literal `NaN` values, so a strict JSON parser rejects it. Loading permissively and validating against `schema.json` then yields 66 errors because 11 failed records have `source: {}` although all six source fields are required.

Represent unavailable values with a JSON value such as `null`, make the schema conditional on whether a point was returned, serialize with `allow_nan=False`, and validate the emitted payload before reporting smoke success. The result contract is a core Gate 0 deliverable, so the smoke path should test it directly.

crates/pounce-algorithm/src/application.rs
```diff
@@ -2875,19 +2923,123 @@ impl IpoptApplication {
...
+                .eval_f(&x_trunc, true)
+                .unwrap_or(Number::NAN);
+            let m = tnlp
+                .borrow_mut()
+                .get_nlp_info()
+                .map(|i| i.m as usize)
+                .unwrap_or(0);
+            let mut g_inner = vec![0.0; m];
+            if m > 0 {
+                let _ = tnlp.borrow_mut().eval_g(&x_trunc, false, &mut g_inner);
```
bernalde: Blocking: `eval_g`'s success flag is discarded, after which the zero-initialized (or partially written) buffer is treated as a valid original-space measurement. A TNLP that returns `false` here can therefore fabricate feasibility and alter both the exit status and reported residuals. The outer-loop measurement just above correctly gates `original_space_feasibility` on `evaluated`; mirror that behavior here so evaluation failure produces `feas = None` and follows the documented fallback. Please cover this with a TNLP regression whose final `eval_g` returns `false`.

crates/pounce-algorithm/src/application.rs
```diff
@@ -2875,19 +2923,123 @@ impl IpoptApplication {
...
+            // The reported constraint violation must be the user's, not
+            // the augmented problem's. Without this the KKT block of a
+            // successful ℓ₁ solve describes a problem the caller never
+            // posed. The aggregate errors take a `max` rather than a
+            // rewrite: the NLP error's primal term enters undivided, so
+            // the aggregate is never below the constraint violation, and
+            // the other two components are unaffected by the wrapper.
+            if let Some(f) = feas.as_ref() {
+                let mut stats = self.statistics.borrow_mut();
+                stats.final_constr_viol = f.max_violation;
```
bernalde: Blocking: `f.max_violation` is explicitly in user units, but it is assigned to both `final_constr_viol` and `final_unscaled_constr_viol` and mixed into both KKT aggregates. That breaks the existing public contract in `SolveStatistics` and `docs/src/python.md`: `final_*` is the internally scaled NLP residual family, while `final_unscaled_*` is in original problem units. The new options text now contradicts that contract.

Please preserve the scaled/unscaled distinction (or add an explicit original-space field if the augmented scaled residual must also remain observable), align all documentation, and add an active-`nlp_scaling_method` regression that proves the two field families retain their stated units.

crates/pounce-algorithm/src/application.rs
```diff
@@ -2875,19 +2923,123 @@ impl IpoptApplication {
...
                     | ApplicationReturnStatus::SolvedToAcceptableLevel
-            ) && slack_sum.is_finite()
-                && slack_sum > slack_tol;
+            );
+
+            // Downgrade, not upgrade: a solve that reached the strict
+            // standard on the user's own rows keeps whatever status the
+            // inner gave it, and nothing here can turn a failure into a
+            // success.
+            let downgrade_to_acceptable = inner_claimed_success
```
bernalde, line 3106: Blocking: the P1 status/reporting change has no committed regression that observes this behavior. The existing ℓ₁ integration target is unchanged and all 11 tests pass without reading `statistics()` or `final_constr_viol`; the benchmark smoke's `1e-5` source threshold would also accept the original reported `2.5e-7` defect.

Please add a focused Rust integration test for the reported original-space residual and the strict-to-acceptable/infeasible status boundaries, using the reproducer that motivated this change. That test should fail on the parent for the P1 reason and pass here.

dev-notes/mpcc-gate0-report.md
```diff
@@ -0,0 +1,429 @@
...
+**The finishing solve takes the point and nothing else.** The relaxed
+problem's duals are duals of a different constraint — the product row is
+an active inequality at `G*H <= tau` and an equality at `G*H = 0` — and
+seeding them measurably sent a three-variable finish into a thrash that
+had not returned in 25 s, where the primal-only transfer converges in
+six iterations.
+
+## Ownership of the observed gaps
+
+549 observations triaged mechanically (rules in
```
bernalde: Nonblocking: this total is internally inconsistent with the table immediately below: `331 + 126 + 61 + 36 + 14 + 8 = 576`, not 549. The same 549 claim appears elsewhere in this report and in the PR description. Please regenerate the total from the records and update every copy so the durable report and merge message agree.

bernalde (changes requested):
I reviewed exact head `27629e81466c7c40534b72332e72f66ab9852700`.

The Gate 0 direction is sound: `scholtes_then_ncp` is a reasonable supported route for Gate 1, keeping source-MPCC measurements separate from lowered-NLP diagnostics is the right contract, and the evidence supports deferring P2 instead of weakening the gh#274 termination guard.

The current implementation is not merge-ready: **5 Blocking, 1 Nonblocking, 0 Questions**. Details are inline.

Evidence on this head:

- `python -m mpcc.selftest`: all 12 checks passed.
- `cargo test -p pounce-algorithm --test l1_penalty_integration`: 11 passed.
- An exact-head Python extension build and `python -m mpcc.run --smoke`: the run completed in 9.4 s and reported all checks passed, but its 153-record output failed strict JSON parsing and its own schema validation, as detailed inline.
- All eight current-head GitHub CI jobs are green.

Coordination: open PR #255 has no file overlap; open PR #149 overlaps only `CHANGELOG.md`.

I would not merge this until the blocking issues above are addressed.

crates/pounce-algorithm/src/application.rs
jkitchin: Confirmed and fixed in `07f7ea9`. You're right that the outer-loop measurement 80 lines above already gates on `evaluated` and this one didn't — that asymmetry was the bug, and I've made this site match.

Measured rather than argued, because the failure mode needed pinning precisely. Failing the fixture's final `eval_g`:

| | reported `final_unscaled_constr_viol` | true violation |
|---|---|---|
| flag discarded | **`0.000e0`** | `2.600e-9` |
| flag honoured | `4.585e-19` (the `l1_slack_tol` fallback) | `2.600e-9` |

So the signature is exactly what you described: the zero-filled buffer read back as a violation of `0`, under `Solve_Succeeded`.

The regression is `an_eval_g_failure_does_not_fabricate_feasibility` in the new `crates/pounce-algorithm/tests/issue_794_l1_original_space_reporting.rs`. One thing worth flagging about how it's built: **my first version of it passed against the unfixed code.** I had it fail `eval_g` from call 40 onward, which kills the solve in restoration long before the exit-verdict site — so the assertion never ran and the test was vacuously green. It now calibrates instead of guessing: one clean solve counts the evaluations, and the second fails only the last of them, which is the exit-verdict call. Verified by reverting the gate and watching that test, and only that test, go red.

---

crates/pounce-algorithm/src/application.rs
jkitchin: You're right, and this was the most consequential of the six. Fixed in `07f7ea9`.

The contract is explicit in both places you name (`solve_statistics.rs:67-73`, `docs/src/python.md:95-100`), and I was writing an original-units number into the scaled family. **This is not hypothetical on this corpus**: gradient-based scaling is active on the sweep's skew leg, and re-running the full sweep moves `nlp.final_constr_viol` on **148 records** (84 of them `final_kkt_error` too), all on the two ℓ₁ routes. So the wrong number was being reported in practice, not just in principle.

**What I did.** The measurement is in original units, so it now always goes to `final_unscaled_constr_viol` / `final_unscaled_kkt_error`, and is mirrored into the scaled family only when per-row scaling did not engage — which is exactly the case where the contract requires the two to agree. A new `row_scaling_active` flag records `c_scale_vec()/d_scale_vec()` presence at the inner solve's finalize, where the NLP handle is still live.

**Why not the conversion you offered as the alternative.** Producing the genuinely scaled counterpart needs the augmented problem's row-scale factors, and `optimize_constrained` has dropped that NLP by the time the outer loop measures. I could have plumbed the factors out, but that widens the change for a number no decision reads — so I took the narrower option and documented the boundary. Worth being explicit about the residual limitation: **under active row scaling `final_constr_viol` still describes the augmented problem.** The status is unaffected either way, because it is decided from the measurement directly rather than from these fields.

Docs aligned in `docs/src/options.md` (which, as you say, had started contradicting the contract) and in the CHANGELOG entry, both now naming `final_unscaled_constr_viol` as the field to read.

Regression: `an_original_units_violation_stays_out_of_the_scaled_family`. It drives `gradient-based` with `nlp_scaling_max_gradient=1e-2` rather than `user-scaling` — I tried `user-scaling` first and it silently did nothing, because `pounce-l1penalty/src/wrapper.rs` declines `get_scaling_parameters` outright ("Phase 2"), so user factors never reach the augmented problem at all. Noted in the test, since it's a trap for the next person writing a scaling test against this wrapper.

---

crates/pounce-algorithm/src/application.rs
jkitchin: Fair, and correct on both counts — the 11 ℓ₁ integration tests never touch `statistics()`, and the smoke's `1e-5` source threshold would indeed have waved the original `2.5e-7` through. Added in `07f7ea9`: `crates/pounce-algorithm/tests/issue_794_l1_original_space_reporting.rs`, built on the `ralph1` exact-product reproducer that motivated the change.

**It fails on the parent for the P1 reason.** Run against `origin/main`:

```
the_reported_violation_is_the_users_not_the_augmented_problems
  reported unscaled constraint violation 9.636e-15 does not match the model's
  own violation 2.501e-7 at x = [6.706e-9, 5.001e-4]

a_violation_above_tol_does_not_report_plain_success
  reported plain success at violation 2.500e-7 against tol 2.501e-11 —
  the l1 exit is being judged by something other than the caller's tolerance
```

Those are the PR description's own numbers, reproduced by the test rather than quoted at it.

Four tests: the two above (the status/reporting change), plus one each for the two defects you found, so the file covers the fix *and* what the fix could break. The status-boundary test derives its strict tolerance from the residual the model actually achieves instead of hard-coding one, so it stays meaningful if the trajectory shifts.

The header carries a mutation table, and every row of it was **run**, not asserted — each mutation reddens the named test and only that test. That check earned its keep twice: it caught my `eval_g` test passing vacuously, and it caught the all-failed selftest check passing for the wrong reason (other rules were firing on the same input), both of which I'd otherwise have shipped as green.

---

benchmarks/mpcc/run.py
jkitchin: Reproduced and fixed in `07f7ea9`. At full-sweep scale it was worse than the smoke subset showed: **652** literal `NaN` and **160** records with `source: {}`, and yes, exit 0 throughout.

One correction to the framing: these artifacts aren't committed — `results-*.json` is gitignored (`.gitignore:99`). So this is a defect in what the harness *emits*, which is the part that matters, since the result contract is the Gate 0 deliverable.

**What I changed**, in the order you listed them:

- Non-finite values serialize as `null`, and the writer refuses anything a strict parser could not read back (`allow_nan=False`), so this cannot silently regress.
- `schema.json` makes `source` conditional on whether a point was returned. I checked the discriminator holds exactly before relying on it: `source == {}` **iff** `x is None`, 160/160 with no exceptions, so an empty `source` beside a returned point is now a schema error rather than an accepted shape. Residual fields admit `null` where "not available" is a real outcome — which it is: the non-finite values were confined to `nlp.*` on solves that returned no point (40 records) and `validation.biactive_distance` where no pair is biactive (12).
- `contract_checks` validates the payload before reporting success, in **every** mode rather than only smoke — a run that writes an unreadable artifact has failed at the thing it exists to produce regardless of how it was invoked.

On validation: `jsonschema` isn't a harness dependency, so `contract_checks` does the structural checks itself (strict round-trip, required keys, the source/`x` correspondence) and layers the full schema on top when the module is importable, reporting its absence rather than skipping in silence. I installed it locally to confirm the emitted artifact really does validate.

The same 153-record smoke command you ran now gives: strict parse OK, **0** literal `NaN`, **0** schema errors, and the 11 empty-`source` records match the 11 with no returned point exactly. Mutation-checked by reverting the sanitizer — the run then fails with `payload is not strict JSON: Out of range float values are not JSON compliant`.

---

benchmarks/mpcc/run.py
jkitchin: Both holes are real. Fixed in `07f7ea9`.

**False success.** You're right that the framing was wrong, not just the threshold: `infeasible_pair` has no feasible point, so *any* success on it is a defect, and conditioning on the returned point being near-feasible meant the more confident failure — success at a large source violation — was the one that got through. The check now rejects every successful `infeasible_pair` record and reports the worst source residual as evidence rather than as a gate.

**All-routes-failed.** Added a vacuity guard, plus the coverage property you asked for: `scholtes_then_ncp` must complete every feasible smoke case/start. That one is worth having independently — it's the route the gate recommends, so if it stops solving a case the recommendation is stale, and nothing else here would have said so.

Both are pinned by new mutation checks in `selftest.py` (now 14 checks). The all-failed one is worth a note, because **my first version of it passed with the guard deleted**: my synthetic all-failed set also trips the new supported-route check, so merely asserting "some failure was reported" left the vacuity guard untested. It now asserts the vacuity failure specifically. Verified both ways — restore the near-feasible-only rule and the false-success check goes red; delete the vacuity guard and the all-failed check goes red.

I kept experimental routes free to fail, as you suggested; the coverage assertion is scoped to `scholtes_then_ncp` only.

---

dev-notes/mpcc-gate0-report.md
jkitchin: Correct — the table is right and the prose was wrong. Recomputed from the records with `report.triage`: the total is **576**, and the per-owner counts are exactly the ones in the table (331 / 126 / 61 / 36 / 14 / 8).

Fixed at all three sites in `dev-notes/mpcc-gate0-report.md` (lines 149, 246, 325) in `07f7ea9`, and I'll update the PR description in the same pass.

Taking the "regenerate the total from the records" part literally rather than just correcting the digits: the generated ownership table now emits a **total** row computed from the same counts the rows come from, so the prose has something to be checked against and a hand-carried total can't drift from the rows it claims to sum again.

Worth recording that this survived the re-run: after the B4 reporting fix moved 148 records, I recomputed the triage on both sides and every bucket is unchanged, total still 576. The ownership numbers in the PR body therefore still stand.

---

jkitchin:
@bernalde — all six findings are addressed (`07f7ea9`), each replied to inline and the threads resolved. Ready for another look when you have time; I have no way to formally re-request a review from here, hence this note. Head is `d64d862d`, all 8 CI jobs green. `selftest` is 14 checks now, not the 12 you ran — the two new ones are the mutation checks for B1.

**One thing changed that is not in any of those threads, and it touches something you endorsed.** Your review said "the evidence supports deferring P2 instead of weakening the gh#274 termination guard". That conclusion holds — but the reasoning the report gave for it was wrong, and I have replaced it with a measurement. Worth flagging because you read the old version.

The report claimed all eight P2 cells had "no bounded multiplier in the limit". That is backwards for seven of them. A row whose gradient vanishes imposes no first-order restriction, so its multiplier is *arbitrary*, not absent: at `qpec_small`'s solution `∇f = 0` exactly and `λ = 0` is an exact KKT multiplier, residual `0`. Only `ralph1` genuinely has none (best sign-feasible residual `0.707`, matching its own docstring's "M-stationary but not S-stationary"). Seven cells and one cell, two different phenomena, reported as one.

That correction made a fix look available, so I tested it. The obvious candidate — re-estimate the multipliers at the returned point — is a mechanism POUNCE already ships as `recalc_y`. It flips every one of these cells to `Solve_Succeeded`. It must not be used:

| cell, with `recalc_y=yes` | scaled `final_kkt_error` | **unscaled** `final_kkt_error` |
|---|---:|---:|
| `ralph1/unit/origin/direct` | `4.607e-10` | **`2.659e-02`** |
| `qpec_small/unit/origin` | `1.091e-09` | `5.066e-08` |
| `qpec_small/unit/upper_left` | `3.384e-11` | **`4.228e-03`** |
| `qpec_small/skew/upper_right` | `1.664e-09` | **`6.179e-05`** |

Three of four are not stationary in the model's own units, by three to six orders, and pass anyway: the re-estimated multipliers are enormous, `s_d` grows with them, and the strict gate reads the residual divided by `s_d`. The verdict is normalised by the quantity that is wrong — the P1 failure class this PR exists to fix. The mechanism is the vanishing gradient itself: a hair off the solution the row's gradient is `~1e-9`, and a multiplier of `~1e9` reproduces any vector required, so an *unbounded* least-squares multiplier fit does not test stationarity near a biactive pair, it manufactures it.

I had drafted the fix into the report on exactly that mistake. Checking the unscaled column is what caught it.

So your call on P2 was right, and it now rests on a measurement rather than on an argument that did not survive checking. The future-fix criteria are narrowed accordingly — any candidate has to be one that a `1e9` multiplier on a `1e-9` gradient cannot satisfy — and `recalc_y=yes` is a one-option reproducer to test candidates against. `d64d862d`, report only, no solver code.

Separately, and recorded rather than filed: `recalc_y=yes` reporting `Solve_Succeeded` at an unscaled KKT error of `2.7e-02` looks like a defect in its own right, independent of MPCC. It is off by default and upstream Ipopt behaves the same way, so it is neither a regression nor on this PR's path.

---

jkitchin:
## Measured: stock Ipopt reproduces the `recalc_y` false success — and `recalc_y` isn't necessary in either solver

`27a7f081` deliberately backs the P2 footnote off to **"Expected, not measured: no Ipopt was run against these models."** I ran one. Three findings, one of which contradicts a claim in the P2 section as written. No blockers in the shipping code.

**Setup.** Ipopt **3.14.19 (Darwin arm64), ASL(20241202)**, the AMPL binary, driven through Pyomo 6.10.0 (NL file → binary); no `nlp_scaling_method` touched. POUNCE built from this branch (`27a7f081`) and from `main` (`cc113479`), run on the *same* `.nl` files through the CLI. Options on every run: `tol=1e-8 max_iter=300 bound_relax_factor=0 honor_original_bounds=yes print_level=5`, varying only `recalc_y`.

Model is `qpec_small` under the exact-product/NCP-equality lowering, written literally (no Pyomo `mpec` transform). I checked the NL header before trusting anything: `3 vars, 6 constraints, 1 objective, 0 ranges, 2 eqns`; bounds block `x: 0 0 2`, `y1: 3`, `y2: 3` — `y1`/`y2` free, `G >= 0` as rows, as the lowering requires.

---

### 1. Upstream reproduces it. The footnote can become a result.

| Ipopt, `recalc_y=yes` | status | iters | scaled overall | **unscaled overall** |
|---|---|---:|---:|---:|
| start `origin` | `EXIT: Optimal Solution Found.` | 8 | 1.428e-10 | **2.636e-03** |
| start `upper_left` | `EXIT: Optimal Solution Found.` | 23 | 7.327e-13 | **3.476e-02** |

<details><summary>raw Ipopt output, both cells</summary>

```
RUN: start=origin  leg B (recalc_y=yes)
   8  4.7649023e-06 4.66e-16 2.64e-03  -8.6 1.98e-03   3.5 4.31e-01 4.31e-01h  1

Number of Iterations....: 8

Objective...............:   4.7649023196668237e-06    4.7649023196668237e-06
Dual infeasibility......:   2.6364821818779726e-03    2.6364821818779726e-03
Constraint violation....:   4.6619422301093753e-16    4.6619422301093753e-16
Variable bound violation:   0.0000000000000000e+00    0.0000000000000000e+00
Complementarity.........:   8.5227209914901571e-05    8.5227209914901571e-05
Overall NLP error.......:   1.4278544040578582e-10    2.6364821818779726e-03

EXIT: Optimal Solution Found.
x=0.9982176976838556  y1=0.99910884884192774  y2=0.00089115115833375894

RUN: start=upper_left  leg B (recalc_y=yes)
  23  2.5650471e-16 1.60e-23 3.48e-02  -9.0 1.99e-02    -  1.00e+00 1.00e+00f  1

Number of Iterations....: 23

Objective...............:   2.5650471097292810e-16    2.5650471097292810e-16
Dual infeasibility......:   3.4756924970053477e-02    3.4756924970053477e-02
Constraint violation....:   1.5970003977276216e-23    1.5970003977276216e-23
Variable bound violation:   0.0000000000000000e+00    0.0000000000000000e+00
Complementarity.........:   9.3926070935021342e-10    9.3926070935021342e-10
Overall NLP error.......:   7.3272742440667782e-13    3.4756924970053477e-02

EXIT: Optimal Solution Found.
x=0.99999998692318348  y1=0.99999999346159174  y2=6.5384094510151369e-09
```
</details>

The mechanism is confirmed directly rather than inferred. In all four Ipopt runs the `Objective`, `Dual infeasibility`, `Constraint violation` and `Complementarity` rows are **bit-identical across the two columns** — NLP scaling came out identity, so it contributes nothing. The only row that moves is `Overall NLP error`, and it moves by exactly the `s_d`/`s_c` divisors. Reading the multipliers back:

| run | `λ[p2]` | `‖∇p2‖∞` | `Σ‖λ‖₁` | unscaled/scaled |
|---|---:|---:|---:|---:|
| origin, `recalc_y=yes` | -2.063e+05 | 1.782e-03 | 1.477e+10 | 1.85e7 |
| upper_left, `recalc_y=yes` | -5.692e+13 | 1.308e-08 | 5.692e+13 | 4.74e10 |

A 5.7e13 multiplier against a 1.3e-8 gradient — the "1e9 on a 1e-9 gradient costs nothing in the fit" argument in the report, at 1e13 scale.

**So the branch `27a7f081` flags — *"if upstream does not reproduce it the finding is a POUNCE porting defect and a much sharper one"* — is closed.** Shared design behaviour, as the shared-provenance argument predicted. The footnote can state it as measured.

### 2. `recalc_y` is not necessary. The framing is too narrow.

The report says what makes this worth an issue is the direction: *"an option whose purpose is better multipliers converts an honest failure into a false success."* But the false success does not need the option. At **default** options (`recalc_y=no`), same models:

| solver | start | status | scaled overall | **unscaled overall** |
|---|---|---|---:|---:|
| Ipopt 3.14.19 | `upper_left` | `Optimal Solution Found.` | 1.881e-11 | **5.934e-02** |
| POUNCE 0.11.0 | `origin` | `Solved_To_Acceptable_Level` | 8.234e-11 | **7.897e+04** |

`recalc_y` makes the trap easier to reach; it does not create it. The subject is the `s_d`-normalised gate itself. An issue filed on the narrow framing will attract a narrow fix — and the report's own "what a future fix would have to establish" list is already written against the narrow version.

### 3. The 7.9e+04 contradicts P2 item 2 as written

> So the gh#274 gate is refusing these correctly, and the honest exit is the one POUNCE already takes.

Item 2 grounds that in gh#274 existing to stop exactly this failure — the `-exp(x)` case at `inf_du = 8.8e+47` reported as solved, *"and Pyomo maps `Solved_To_Acceptable_Level` into the solved family."*

On the report's own `recalc_y` reproducer model, at default options, POUNCE exits **`Solved_To_Acceptable_Level` at an unscaled dual infeasibility of 7.897e+04**, after 9 restoration iterations — i.e. through that gate. The gate did not refuse, because the acceptable-level triplet is evaluated on the `s_d`-normalised quantities, and `s_d` here is ~1e15.

<details><summary>raw POUNCE output, default options, start <code>origin</code></summary>

```
  41  6.9729927e-13 9.51e-14 1.84e-01   -9.0 3.08e-07      - 1.00e+00 1.00e+00h  1

Number of Iterations....: 41

Objective...............:   3.5860240840569709e-28    3.5860240840569709e-28
Dual infeasibility......:   7.8965510781517834e+04    7.8965510781517834e+04
Constraint violation....:   1.1102230246251565e-16    1.1102230246251565e-16
Variable bound violation:   0.0000000000000000e+00    0.0000000000000000e+00
Complementarity.........:   1.1489451778285931e-09    1.1489451778285931e-09
Overall NLP error.......:   8.2335532426389998e-11    7.8965510781517834e+04

Number of restoration iterations                     = 9 (in 1 call)

EXIT: Solved To Acceptable Level.
POUNCE 0.11.0: Solved To Acceptable Level.
Status: Solved_To_Acceptable_Level
```
</details>

That sentence should not stand as written.

### The caveat, stated rather than buried

**These are the same *model* as the report's rows, not the same *cell*.** My runs go through the `.nl`/CLI path with exact Hessians; the harness runs `qpec_small` through the Python path, which selects limited-memory when no exact Lagrangian Hessian is available. The numbers do not match:

| | report table | this run |
|---|---|---|
| `qpec_small/unit/origin`, `recalc_y=yes`, scaled | 1.091e-09 | 2.231e-10 |
| same, unscaled | 5.066e-08 | 1.496e-03 |

That is enough to falsify the *general* claim in finding 3 — the gate demonstrably admits a diverging dual on this model — but **not** enough to restate the four-row table. Anyone acting on this should re-measure through the harness rather than copy these numbers in.

Related discrepancy worth reconciling: on the CLI path `recalc_y=no` gives `Solved_To_Acceptable_Level` / `Solve_Succeeded`, not `Error_In_Step_Computation`. If the P2 cells actually exit acceptable-level, then **"All 8 candidates are one exit"** may be naming the wrong exit, and the ownership bucket's description needs a check.

---

### Blast radius: now measured, not argued

The PR argues the change is confined to the ℓ₁ path because both options default off. I built both binaries and ran all four cells through each. **`main` (`cc113479`) and this branch are bit-identical on every reported field** — status, iteration count, and all four rows in both columns. So the false success above is pre-existing and is *not* introduced here; and the confinement claim holds by measurement.

### Code review — no blockers, two nits

Read the whole `application.rs` diff. The logic is right: downgrade-only, `None` meaning "not measured" and never "feasible", `eval_g`'s success flag gated at both call sites, and the units contract respected (measurement into the unscaled family, mirrored only when `row_scaling_active == Some(false)`). `final_kkt_error_above_noise` correctly updates only in the mirror branch — it has no unscaled sibling.

1. **`row_scaling_active` is fail-open.** It is written only when a solve reaches the statistics block in `optimize_constrained`, and is never reset on entry. On a reused `IpoptApplication` where a later inner solve bails before that block, the ℓ₁ outer loop reads a stale value and can mirror an original-units number into the scaled family — precisely the contract B4 fixed. Resetting it to `None` at the top of `optimize_constrained` makes it fail-closed. Narrow, but it is the same class as B4.
2. `judge` in `original_space_feasibility` always returns `true` and every call site discards the value. Dead return.

### Suggestion

Fix the three report claims and land it. Findings 2 and 3 are documentation-of-a-known-boundary problems where the boundary turns out to sit somewhere other than the report puts it — the same class of error the last two commits were already correcting, which is a point in the report's favour, not against it.

<sub>Reproduction script and all raw output available on request; Ipopt run via Pyomo→ASL, POUNCE via `target/release/pounce <stub>.nl`.</sub>

jkitchin:
All four addressed in `a26bb0c9`. Findings 2 and 3 are correct and I reproduced both locally before changing anything; finding 3 also falsifies a claim I pushed *after* you started reviewing, so thank you for catching it while it was ~20 minutes old.

## The `7.9e+04` — reproduced, and it takes down more than one sentence

I generated the same model through Pyomo (NL header `3 6 1 0 2`, `y1`/`y2` free, `G ≥ 0` as rows) and ran the CLI:

```
Number of Iterations....: 41
Dual infeasibility......:   7.8965510781517834e+04    7.8965510781517834e+04
Overall NLP error.......:   8.2335532426389998e-11    7.8965510781517834e+04
Number of restoration iterations                     = 9 (in 1 call)
EXIT: Solved To Acceptable Level.
```

Bit-identical to yours. "So the gh#274 gate is refusing these correctly" is gone. What replaces it says the sharper thing: this *is* the gh#274 failure mode, on a model this report ships as a reproducer, and the gate refuses `-exp(x)` (nothing cancelling, `dual_inf / D ≈ 1`) while admitting this one because `D` is ~1e15.

## Finding 2 is right, and it also breaks a claim of mine you didn't see

You're right that `recalc_y` isn't necessary and the narrow framing would attract a narrow fix. Widened: the subject is the `s_d`-normalised gate, in both implementations, with gh#532 named as where a fix should start reading since it already answers a neighbouring version for the strict dual component.

But your default-options table did more damage than that. Between your review and this reply I pushed `2cee9894`, reclassifying P2 as **"a POUNCE defect: upstream Ipopt solves it"** — on the strength of the `origin` row alone. Your `upper_left` row reverses it. Running both starts on the same `.nl`, same options, exact Hessians, default `recalc_y=no`:

| start | solver | iters | exit | **unscaled** overall |
|---|---|---:|---|---:|
| `origin` | Ipopt 3.14.19 | 58 | `Optimal Solution Found` | **`7.523e-09`** |
| `origin` | POUNCE 0.11.0 | 41 | `Solved To Acceptable Level` | **`7.897e+04`** |
| `upper_left` | Ipopt 3.14.19 | — | `Optimal Solution Found` | **`5.934e-02`** (yours) |
| `upper_left` | POUNCE 0.11.0 | 76 | `Optimal Solution Found` | **`1.564e-08`** |

Neither dominates. On `origin` POUNCE reports acceptable-level at `7.9e+04` where Ipopt reaches `7.5e-09`; on `upper_left` POUNCE converges to `1.6e-08` and Ipopt reports optimal at `5.9e-02`. So "a POUNCE defect, upstream solves it" was one start's worth of evidence, exactly as the version before it was one solver's worth.

**P2 is now marked open pending harness re-measurement, not reclassified a third time.** The section carries a status note saying it has been wrong twice in opposite directions and confining itself to what has been run. The bucket keeps its 8 cells and its `POUNCE candidate` label, since that label is the mechanical triage's and never depended on either story.

## The harness/CLI discrepancy

Taken as you framed it. `qpec_small/origin` exits `Solved_To_Acceptable_Level` at 41 iterations through the `.nl`/CLI path and `Error_In_Step_Computation` at 118 through the harness. One correction to your caveat, though: **the harness is not on the limited-memory path here.** `lowering.py` defines `hessian`/`hessianstructure`, and the harness run reports 119 Lagrangian Hessian evaluations. Both paths are exact — the sparsity differs (harness hands back a dense lower triangle, 6 nonzeros, `.nl` detects 5), which is now recorded as the leading hypothesis and explicitly not an established one. So "all eight are one exit" is scoped to the Python path and the bucket description is flagged for re-checking; the count of 8 is unaffected, coming from the triage rules rather than the exit string.

## Nit 1 — real, and the same class as B4

Fixed. You're right that it was fail-open: written near the end of `optimize_constrained` from the NLP the solve built, never reset, and the ℓ₁ outer loop calls that function repeatedly and reads the flag after each call. A stale `Some(false)` from a previous attempt lets the outer loop mirror an original-units violation into the scaled family — precisely what B4 added the flag to prevent. Now cleared at function entry, so "not recorded" reads as "cannot mirror" rather than "scaling was off".

In fairness I should say what I did **not** do: I have not added a regression for it. I could not construct a natural path where an inner attempt bails before the statistics block without contriving the TNLP, and I would rather say so than ship a test that passes for the wrong reason — which is a failure mode this PR has hit twice already. The fix is fail-closed by construction; if you can name a model that reaches the early-bail path I will pin it.

## Nit 2

Dropped the dead `bool` return from `judge`.

## Verification

Sweep re-run after the code change: **2880 records, 0 differing.** 858 tests pass in `pounce-algorithm` + `pounce-l1penalty`, `selftest` 14 checks, `cargo fmt` and both consistency scripts clean. Your independent finding that `main` and this branch are bit-identical on all four cells matches my own blast-radius claim, and is better evidence for it than the argument I gave, since it is measured on both binaries rather than reasoned from option defaults.

---

bernalde (changes requested):
I reviewed exact head `c67615b421dcf22af72d0be14b1b6f76cfcd7d25`, targeting the changes after `07f7ea968dd434bf17df6f864bb665e2b78c8b52`.

The substantive follow-up changes are sound: the `row_scaling_active` reset is fail-closed at `optimize_constrained` entry; the restoration deadline test now asserts the wall-time contract rather than an iteration-ratio proxy; and the acceptable-level/P2 analysis matches the implementation.

One blocker remains.

### Blocking — stale P2 ownership sentence

`dev-notes/mpcc-gate0-report.md:256` still says P2 was “confirmed against Ipopt 3.14.19, which solves the reproducer POUNCE fails.” That is the overturned one-start conclusion and directly contradicts both the later P2 section and this PR body, which correctly leave P2 open pending harness re-measurement. The durable report must use the same open/triage-only framing before merge.

I prepared the minimal correction in [dc9e85e6](https://github.com/bernalde/pounce/commit/dc9e85e6a6ad92cca4d8458f132bd3342d35ae55). The authenticated `bernalde` account cannot push John's branch (GitHub returned 403), so this can be applied with `git cherry-pick dc9e85e6a6ad92cca4d8458f132bd3342d35ae55`.



### Nonblocking — pin fail-closed scaling state across attempts

The PR body correctly calls out that the reset has no direct regression. This is cross-attempt mutable state in the ℓ₁ outer loop: a focused test should seed the previous attempt's `Some(false)`, take an early exit before new scaling is recorded, and assert that the state is `None`. The same patch includes that test; removing only the reset makes it fail with `Some(false)` instead of `None`.



Evidence:

- All eight current-head GitHub CI jobs are green.
- Exact-head verification retained the 14 MPCC self-checks, the 4 original-space ℓ₁ integration regressions, strict smoke JSON/schema validation, and the restoration deadline test.
- On the two-file patch, 1,018 tests pass across `pounce-algorithm`, `pounce-l1penalty`, and `pounce-restoration`; formatting, documentation consistency, and release consistency are clean.
- Open PR #881 overlaps `CHANGELOG.md`, `application.rs`, and `upstream_options.rs`, but its code hunks concern separate SQP/option identifiers; PR #149 overlaps only `CHANGELOG.md`.

Verdict: **REQUEST_CHANGES** for the stale durable report. Once `dc9e85e6` (or an equivalent correction and regression) lands and current-head CI remains green, I found no further blocking issue in this targeted round.

jkitchin:
Both addressed in `2ec5af41`. The blocker was mine and you're right about it; the nonblocking one corrected something I'd asserted was impractical.

## Blocking — the stale ownership line

Fixed. `dev-notes/mpcc-gate0-report.md:256` now reads **open**, with the label attributed to the triage rules rather than to any mechanism.

Worth naming how it got there, since it is the same failure twice: I wrote that line in `2cee9894` when I believed P2 was a POUNCE defect, and when `a26bb0c9` retracted that on your `upper_left` row I corrected the P2 section and did not sweep for the claim elsewhere. So the durable report contradicted itself two hundred lines apart — exactly the "documentation of a known boundary where the boundary has moved" class you named. I have now swept the file: the only surviving occurrences of the retracted phrasing are inside the narrative that describes it as a superseded version.

## Nonblocking — the regression now exists

You were right that it was reachable, and my stated reason for not writing it was wrong in a specific way: I was thinking of an integration test under `tests/`, which cannot observe a private field. A unit test **inside the module** can, and your sketch is exactly that shape.

Three things had to be right to reach the path, none of which I'd have guessed without building it:

- **The fixture needs `eval_h`.** `Hs071Stub` has none, so the solve lands on L-BFGS and panics in `OrigIpoptNlp::eval_h` before recording anything. `ExactQuadratic` supplies one.
- **It must not be convex.** `solver_selection=auto` routes a convex model to the convex engine, which never reaches the statistics block the flag is written from — `ConvexEqTnlp` records no verdict at all, so the test fails on its own precondition.
- **The second solve must decline via `get_bounds_info`, not `get_nlp_info`.** Refusing `get_nlp_info` mid-flight panics rather than unwinding, which would have tested the panic path instead of the reset.

Mutation-checked, and it reproduces your predicted signature exactly:

```
assertion `left == right` failed: a solve that bailed before recording row scaling
left the previous solve's verdict in place ...
  left: Some(false)
 right: None
```

**On `dc9e85e6`:** I did not cherry-pick it. `bernalde/pounce` is outside this session's repository scope, and writing it here meant the three constraints above and the mutation check are established rather than inherited — which matters more than usual on a PR that has already shipped two tests passing for the wrong reason. The finding and the approach are yours; only the keystrokes are mine. If your version differs in a way that covers something mine doesn't, say so and I'll take it.

## Scope and verification

The only **non-test** change in this commit is the one report line. Every `application.rs` hunk is inside `#[cfg(test)] mod tests` — verified against the module boundary at line 6269 rather than assumed — so the shipped binary is untouched and no fixture sweep is implicated.

**1018 tests pass** across `pounce-algorithm`, `pounce-l1penalty` and `pounce-restoration`, which is the same count you report on your patch. `cargo fmt`, `check-docs-consistency.sh` and `check-release-consistency.sh` all clean.

---

jkitchin:
## Adversary pass on the P1 fix found a regression the fix itself introduced. Fixed in `464e1256`.

Also merged `main` (12 commits, incl. #881) — no textual conflicts, but the auto-merge left this branch's two `[Unreleased]` bullets unsectioned above main's new `### Added`; they are now under Added / Fixed.

### The finding

P1 judged each of the user's rows with `is_negligible(viol, scale, tol)` alone — `viol ≤ tol · max(|row|, 1)`. That is a relative test with **no absolute floor**, so the allowance grows without bound with the row's magnitude: at the default `tol = 1e-8`, a row near `1e10` buys `1e2` of "feasible".

Probe: `min (x/K)²` s.t. `x = K`, `0 ≤ x ≤ K − 50` — **infeasible by exactly 50 by construction**, no oracle needed. Holding the violation at 50 and sweeping only `K`, the verdict tracked `tol · K` exactly:

| `K` | `tol · K` | verdict |
|---|---:|---|
| `1e6` | `1e-2` | `Infeasible_Problem_Detected` |
| `1e8` – `4.9e9` | `1` – `49` | `Solved_To_Acceptable_Level` |
| **`5.1e9` – `1e12`** | **`51` – `1e4`** | **`Solve_Succeeded`** |

Same row, same 50-unit violation, three verdicts — decided only by how large the row's target number is.

**It was a regression this branch introduced, not an inherited gap.** On the parent the same model exits `Error_In_Step_Computation`, because the `Σ(p+n) > l1_slack_tol` argument P1 replaced was crude but *absolute*. So P1 as first written traded a dishonest number under an honest status for an honest number under a dishonest status — and "downgrade-only … nothing here turns a failure into a success", which the PR body and the report both claimed, was false in exactly this corner.

### The repair

Stop inventing a feasibility standard and use the one the solver already has. `OptErrorConvCheck::primal_component_passes` is an absolute `constr_viol ≤ constr_viol_tol` whose scale-awareness comes from **abstaining when every row sits at its own floating-point noise floor** (gh#528, gh#590), not from multiplying the tolerance by the row's size. Its own doc states the property `is_negligible` lacks:

> "cannot fabricate a success on a genuinely infeasible model: such a model's violation is pinned at its infeasibility gap, orders above `eps ·` the row's own magnitude"

In the probe, `50` is `~2e7 ×` this row's floor, so nothing abstains.

`primal_resolvable` cannot be reused verbatim — the CQ computes it on the *augmented* NLP, whose rows the slacks satisfy to machine precision, so it would abstain always and accept everything. The floor is recomputed on the user's own row from the same `primal_noise_floor_kappa · eps · magnitude` the option documents; `= 0` opts out as it does for the strict gate.

The arms are **conjoined, not substituted**: an absolute-only rule would re-admit P1 itself, whose `2.5e-7` violation sits well inside `constr_viol_tol = 1e-4`.

### Evidence

- **Full MPCC sweep bit-identical: 0 of 2880 records differ** once wall-clock timing is stripped; no record gains or loses a success. Every table in this PR stands.
- That null result is not reassurance, and the report says so: **every row in all eleven cases is `O(1)`**, so the corpus is uniform in the dimension this defect acts on and could never have caught it — the same lesson `dev-notes/convex-fixture-corpus.md` records for the bound-relax cost. This needed a probe, not more records.
- `original_space_feasibility` has exactly two call sites, both inside the ℓ₁ outer loop, so the confinement argument for `scripts/sweep-fixtures.sh` is unchanged.
- **1024 tests** pass across `pounce-algorithm`, `pounce-l1penalty`, `pounce-restoration` (up from 1018 — #881's new tests); the 4 CLI ℓ₁ tests pass; MPCC selftest 14/14; `cargo fmt` and both consistency scripts clean.
- Guard: `a_large_row_magnitude_does_not_buy_success_on_an_infeasible_model`, straddling the measured crossover, **mutation-checked** — drop the absolute conjuncts and it alone reports `Solve_Succeeded` at `4.999990e1`.

### Recorded, not fixed

The same scale-relative shape appears in `withdraw_infeasibility_if_refuted` — which is why the repaired verdicts read `Error_In_Step_Computation` rather than `Infeasible_Problem_Detected` at large `K`: the refutation check finds the starting point "feasible" at that row's scale. That path is shared with the IPM and SQP arms, predates this branch, and changing it is not this PR's to make.

A second probe — a user setting `constr_viol_tol=1e-2` on a model infeasible by `1e-3` — found **no** disagreement: both legs refuse. Reporting it because a null result from an adversarial probe is only worth anything if it is stated.

---

PR #861, opened 2026-08-30
sens: solve declared models as written, no per-solve clone

devin-griff:
## The change

Every solve of a model with declared sens params cloned the whole model and re-ran the sensitivity-toolbox surgery on the clone, a cost paid identically on the hundredth solve of an unchanged model. `declare_sens_param` now does its work once, at declaration. A declared Param entering the model through one defining equality, a single variable equal to the param, the shape a parameterized initial condition already has, is recorded and its row pinned as written. A declared Param without that form is rewritten in place with a warning: its occurrences read a new variable held by a new defining equality on the `_pounce_sens_defs` block, the affected constraints and objectives edited in place so their names and activity are untouched, bounds holding it moved into constraints over the substitute (the gh#356 semantics, now at declaration), and a declared fixed Var unfixed and pinned where it stands. Solves construct no interface and clone nothing. The pin rows resolve by constraint name, the perturbation deltas carry each defining row's own coefficient and solve-time base, and `con_alias` is empty on this path because nothing is renamed. The call-time `sens_params` keyword keeps its solve-local clone.

The toolbox could not serve the in-place role: its surgery is wholesale, deactivating every constraint in the model and rebuilding them as copies under its block, which is what forced the alias layer and the per-solve clone in the first place. The rewrite here substitutes with `replace_expressions` and `set_value` on the rows that actually contain the param.

## Measured

On the 62k-variable double column (N=25 Radau collocation, the hat initial conditions declared), the removed per-solve block costs 5.9 to 7.3 s: 3.2 to 3.5 s for the deepcopy and 2.7 to 3.7 s for the surgery, over three trials. Every solve of a declared model paid it. The column's hat params conform as written, so its solves take no rewrite and no warning.

## Tests

Eight tests in `test_declared_pin_rows.py`: a conforming model solves with zero interface constructions (counted through a monkeypatch), repeated solves stay at zero, both orientations and a folded-offset form carry their coefficients, a folded param is rewritten in place with the constraints still active under their own names, sequential rewrites, a param in the objective, a deactivated defining row raises at solve with a message naming the row, and the call-time keyword still clones. Two existing tests updated to the new semantics with the reasons in their docstrings: the warm-start dual now lands on the un-aliased row, and the SolutionMap includes the substitute variable, which is an ordinary model variable now. The counting and length assertions fail on the parent commit by construction.

`pyomo-pounce`: 499 passed, the three issue-816 availability tests failing on this Windows machine identically on `main`. `python/tests`: green to its known environment failures.

---

- [x] Tests fail on the parent commit
- [x] Book page updated where it described the clone and the surgery

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,25 @@ changes.
 
 ## [Unreleased]
 
+- **A model with declared sens params solves as written: the per-solve
+  clone is gone.** Every solve of a declared model used to deep-copy
+  the whole model and re-run the sensitivity-toolbox surgery on the
+  copy, a cost paid identically on the hundredth solve of an unchanged
+  model. `declare_sens_param` now inspects each declared Param once,
+  at declaration: a Param entering the model through one defining
+  equality (a single variable equal to the param, the shape a
+  parameterized initial condition already has) is recorded and pinned
+  as written, and a Param folded anywhere else is substituted in
+  place, once, with a warning naming the rewrite: its occurrences read
+  a new variable pinned by a new defining equality on the
+  `_pounce_sens_defs` block, constraints edited in place so their
+  names and activity are untouched. Solves construct no interface,
+  clone nothing, and rewrite nothing; a perturbed re-solve is: set the
+  param, solve. Declared fixed Vars are unfixed and pinned where they
+  stand. The call-time `sens_params` keyword keeps its solve-local
+  clone, and editing the model after declaration so a declared Param
+  leaks into new expressions is documented as unsupported rather than
+  hunted for at solve time.
 - **`mode="path"` takes a weakly active bound back instead of walking
   out of the box (gh #852).** On the coupled kink
   `min (x - p)² + 0.1(y - 1)²` s.t. `y = 2x + 1`, `x ≥ 0`, held at
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ changes.
   a new variable pinned by a new defining equality on the
   `_pounce_sens_defs` block, constraints edited in place so their
   names and activity are untouched. Solves construct no interface,
-  clone nothing, and rewrite nothing; a perturbed re-solve is: set the
+  clone nothing, and rewrite nothing. A perturbed re-solve is: set the
   param, solve. Declared fixed Vars are unfixed and pinned where they
   stand. The call-time `sens_params` keyword keeps its solve-local
   clone, and editing the model after declaration so a declared Param
--- a/docs/src/sensitivity.md
+++ b/docs/src/sensitivity.md
@@ -142,6 +142,21 @@ are present, `SolverFactory("pounce").solve(m)` runs in-process and
 keeps the converged KKT factorization, so every query afterwards is a
 single backsolve.
 
+A declared `Param` should enter the model through one defining
+equality: a single variable equal to the param, the shape a
+parameterized initial condition already has (`m.x0 == m.x0_hat`). Such
+a model solves as written, on every solve, and the defining equality
+is the row the machinery perturbs. A declared `Param` without that
+form is rewritten in place once, at declaration, with a warning: its
+occurrences are replaced by a substituted variable held by a new
+defining equality, the affected rows edited in place so their names
+are untouched. Writing the defining equality yourself avoids the
+rewrite and is the recommended form. Editing the model after
+declaration so a declared `Param` appears in new expressions is
+unsupported: re-declare on the current model instead. Repeated solves
+of one declared model, the receding-horizon pattern, pay no
+per-solve model copy and no per-solve rewrite.
+
 ```python
 import pyomo.environ as pyo
 import pyomo_pounce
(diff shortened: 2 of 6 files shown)


jkitchin (commented):
Reviewed by checking out the branch and running it against `main`. The design is right and the perf win is real — the conforming path genuinely constructs no interface and clones nothing, and the coefficient handling holds up. Two defects produce wrong answers, both reproduced below.

## Blocking

### 1. A declared fixed Var is pinned at its declaration-time value, so it stops tracking across re-solves

`_register_pins` bakes a literal:

```python
con = blk.pin.add(pd == float(pyo.value(pd)))      # sens.py:340
```

The Param path uses `blk.pin.add(v == pd)`, which tracks the mutable Param live on every solve. The fixed-Var path has no live source, because the value now lives in the very Var that was just unfixed.

```python
m = pyo.ConcreteModel()
m.u = pyo.Var(initialize=2.0); m.u.fix(2.0)     # the "parameter"
m.x = pyo.Var(initialize=0.0)
m.c = pyo.Constraint(expr=m.x == 3.0 * m.u)
m.obj = pyo.Objective(expr=(m.x - 1.0) ** 2)
declare_sens_param(m.u)

opt = pyo.SolverFactory("pounce")
opt.solve(m)                    # u=2, x=6,  dx/du = 3      correct
m.u.set_value(5.0); opt.solve(m)
m.u.fix(7.0);       opt.solve(m)
```

| | this branch | `main` |
|---|---|---|
| solve 1 | u=2, x=6 | u=2, x=6 |
| solve 2 (`set_value(5.0)`) | **u=2.0, x=6.0**, returns normally | u=5, x=15, optimal |
| solve 3 (`fix(7.0)`) | x=0.0, `invalidProblem` | u=7, x=21, optimal |

Solve 2 is the bad one: the new value is silently discarded and the call returns as if nothing happened. This is a regression against `main`, and it lands on exactly the receding-horizon workflow the PR exists to speed up.

`declare_sens_param`'s own docstring advertises this input — *"one or more mutable Params (or fixed Vars)"* — and **no test in either suite declares a fixed Var**, on this branch or on `main`, which is why it got through. `test_declared_pin_rows.py` has no fixed-Var case either.

Something has to read the Var's current value per solve here (or pin it against a mutable Param mirrored from the Var). That is an O(#declared) read, not a model clone, so it costs nothing against the headline result.

### 2. A raising declaration leaves the model half-rewritten, and it then answers with a wrong derivative

The warning, the `_pounce_sens_defs` block, and the per-param substitutes are all created before the `ValueError` for a non-fixed Var (`sens.py:298-337`). There is no rollback, and `reg.params.extend(comps)` has already run.

```python
m.p = pyo.Param(initialize=1.0, mutable=True)
m.x = pyo.Var(initialize=1.0)
m.z = pyo.Var(initialize=0.0)                     # NOT fixed: illegal
m.c1 = pyo.Constraint(expr=m.x * m.p + m.x == 2.0)
m.obj = pyo.Objective(expr=(m.x - 1) ** 2 + m.z ** 2)

declare_sens_param(m.p, m.z)                      # ValueError: z is a Var that is not fixed
```

After the raise the model carries `_pounce_sens_defs` with `v[1] == p`, but `c1` still reads `p` directly, and `reg.params` holds both `p` and `z`. Solving that model:

```
status: optimal      x = 1.0
gradient(m.x, wrt=m.p) = 0.0        # exact: -2/(1+p)^2 = -0.5
```

A wrong sensitivity reported as success. On `main` the same call raises later, from the toolbox, with the model untouched. Validating every declared component before the first `set_value` / `add_component` would close it.

## Should fix

### 3. `moved_bounds` is now declaration-time, while everything describing it still says solve-point

The comment carried over from `main` — *"the numeric solve-point values are kept for covariance()"* — was true when `_reformulate_param_bounds` ran per solve on the clone. `_Session.moved_bounds`'s own docstring says the same. Both are now false.

```
declare with m.hi = 2.0 ; solve            -> moved_bounds = {'x': (None, 2.0)}
m.hi.set_value(9.0) ; solve                -> x = 9.0, dx/dhi = 1.0   (model correct)
                                              moved_bounds = {'x': (None, 2.0)}   (stale)
```

No library code reads it today — only tests — so this is doc drift plus a latent trap rather than a live wrong answer. But a `covariance()` change that starts reading it is exactly the shape of the next defect.

### 4. A Param in a *fixed* Var's bound now forces a whole-model rewrite

The flagging loop has no `if v.fixed: continue`, though the moving loop below it does:

```python
for v in model.component_data_objects(pyo.Var, active=True, descend_into=True):
    for attr in ("_lb", "_ub"):                    # sens.py:255-262
```

`main` skipped fixed Vars deliberately — *"their bounds are not enforced"*. So a param with a perfectly good defining equality gets dragged into surgery, plus the warning, purely because it also appears in a fixed Var's bound:

```python
m.cap = pyo.Param(initialize=5.0, mutable=True)
m.f = pyo.Var(bounds=(None, m.cap)); m.f.fix(9.0)   # bound never enforced
m.pin = pyo.Constraint(expr=m.x == m.cap)           # a proper defining equality
declare_sens_param(m.cap)     # -> warns, builds _pounce_sens_defs
```

The answer stays correct; the rewrite is a false positive.

## Minor

**5. `reg.surgered` is dead state.** Written at `sens.py:333` and `:376`, deep-copied at `:130`, never read. Its comment says *"the in-place rewrite ran: it can run at most once per model"* — `test_sequential_rewrites_each_get_their_block` runs it twice.

**6. The warning fires on the project's own documented spelling.** Six existing tests now warn, and `bounds=(m.lo, m.hi)` is a form `docs/src/sensitivity.md` still endorses two sections later (*"Expression bounds work too"*). "The model was rewritten in place" is a loud thing to say about a documented-good bound.

**7. Coverage gaps.** The eight new tests are all scalar params. The motivating 62k column declares *indexed* hat params — I checked that path by hand and it is correct (`dy/dhat[i]` exact, no rewrite, no `_defs` block created), but nothing pins it. Also unpinned: fixed Vars (finding 1), and declared + call-time `sens_params` on the same solve, where the declared params go back through the toolbox on the clone after having already been rewritten in place.

## What holds up

- The conforming path constructs no interface and clones nothing: `si = None; clone = model`, confirmed by the monkeypatch counter and by reading `sens_solve`.
- The coefficient sign is right in both orientations. Pyomo can store the residual of `m.p == m.x` either way round, and `delta = -coeff * dp` comes out correct for both — `test_orientation_and_offset_forms_conform` covers that deliberately, which is the right instinct.
- `pin_bases` is read at solve time, so conforming Params track re-solves correctly, including indexed ones.
- Full `pyomo-pounce` suite green on this machine: 469 passed, 13 skipped.

devin-griff:
All of it applied, in `f2893e3f`.

**1.** The pin has a live source: `sens_solve` refreshes every declared fixed Var's pin from the Var's current value before the NL write, and a Var the caller re-fixed is unfixed again there, so `set_value` and `fix` both keep tracking. Your reproduction is the test, x at 6, 15, 21 across the three solves, dx/du at 3 throughout, zero interface constructions.

**2.** The declaration validates every component of the call before anything is recorded or rewritten, so the raise leaves the model untouched: no defs block, no registry entries, and the follow-up declaration of the legal param alone answers the exact -0.5. Also a test.

**3.** The moved-bounds records now say what they hold: values from the move, declaration time for declared params and the solve for a call-time clone, with the session comment pointing a future reader at the moved row for current values.

**4.** The flagging walk skips fixed Vars exactly as the moving loop always did.

**5.** `surgered` is gone.

**6.** A param appearing only in Var bounds takes the rewrite without the warning: the move is the documented mechanics of the documented spelling. It still warns when the constraint side is also non-conforming, and a conforming defining row does not exempt a bound occurrence from the rewrite, or the bound would quietly stop moving. The six tests that warned are quiet again.

**7.** Tests for each gap: indexed conforming params, the fixed-Var reproduction, the raising declaration, the bound-only quiet path, and declared plus call-time params on one solve.

`pyomo-pounce` 504 passed with the three Windows knowns, `python/tests` green to its knowns.

Issue #862, opened 2026-08-30
[adversary] solve_qp: PSD pre-check masks the P-shape guard, raw IndexError instead of the written ValueError


jkitchin:
adversary-fingerprint: solve-qp/psd-precheck-masks-p-shape-validation

`solve_qp`'s PSD pre-check runs **before** `_validate`, so when `P` has more
rows than `n` the pre-check indexes an `n`-sized array with a `P`-sized index
and raises a raw `IndexError` from numpy internals. The precise `ValueError`
that `_validate` already contains for exactly this case is never reached on
the default path.

The same call with `check_psd=False` — which skips the pre-check and lets
`_validate` run — produces the correct, actionable error. So the two spellings
of one malformed input disagree, and the default is the worse one.

This is the guard added for #113 ("shape mismatch -> misleading status")
becoming unreachable. It is plausibly exposed by 7bfb3821 / #849, which made
the PSD guard always run rather than switching off above n=1500.

Severity is low: it rejects rather than returning a wrong answer. It is filed
because the rejection is opaque, inconsistent with every sibling shape check in
the same function, and reachable from five public entry points.

### Reproduction

```python
import numpy as np, pounce
n = 5
P = np.eye(7) * 2.0          # 7x7, but c sets n = 5
c = np.ones(n)

pounce.solve_qp(P=P, c=c)                    # IndexError (opaque)
pounce.solve_qp(P=P, c=c, check_psd=False)   # ValueError  (correct)
```

Affected entry points (all with default `check_psd`):

| call | result |
|---|---|
| `solve_qp(P, c)` | `IndexError` |
| `solve_qp(P, c, method="active-set")` | `IndexError` |
| `solve_qp_batch([dict(P=P, c=c)])` | `IndexError` |
| `solve_qp_multi_rhs(P=P, cs=...)` | `IndexError` |
| `solve_socp(P=P, c=c, ...)` | `IndexError` |
| `solve_qp(P, c, check_psd=False)` | `ValueError` (correct) |

Sibling shapes that are already handled correctly on the default path, which
is what makes the inconsistency visible: `(3,3)`, `(4,4)`, `(5,7)` and `(7,5)`
all raise the precise `ValueError`. Only `P` with **more rows than `n`** hits
the pre-check first.

### Expected

The message `_validate` already produces, and does produce with
`check_psd=False`:

```
ValueError: solve_qp: `P` has shape (7, 7) but must be (5, 5)
```

[...]

Issue #852, opened 2026-08-29
sens: mode="path" has no breakpoint for a base-active bound the step pushes through, so a held weak bound is walked past and only the clamp catches it


jkitchin:
> **Rewritten.** The first filing blamed `parametric_step_path_decided`'s handling of an empty held list and claimed [#842](https://github.com/jkitchin/pounce/pull/842) was what made this reachable. @devin-griff's measurement ([comment](https://github.com/jkitchin/pounce/issues/852#issuecomment-5464723805)) shows both claims were wrong: `degeneracy="one_sided"` reproduces it through the *plain* `parametric_step_path`, so it has been reachable from `estimate()` on `main` all along and has nothing to do with the decided surface. The mechanism below replaces the original analysis.

## What happens

On the coupled kink of `pyomo-pounce/tests/test_degeneracy.py` (`min (x-p)² + 0.1(y-1)²` s.t. `y = 2x+1`, `x ∈ [0,10]`, held at `p = 0`), holding side, `dp = -1`. Exact answer `x = 0`, `y = 1`.

| `mode="path"` with | `x` | `y` | |
|---|---|---|---|
| `degeneracy="directional"` | 0 | 1.000009 | correct |
| `degeneracy="one_sided"` | 0 | 0.285722 | **wrong** — and this is `main` today |
| `degeneracy="release_all"` | 0 | 0.285723 | **wrong** |
| re-solve | 0 | 1.000000 | |

Measured by @devin-griff at #842's head `3a548c36`. Both failing runs emit `estimate`'s warning `path applied 0 active-set change(s) and still leaves the bounds for ['x'] ... The values were clamped`; `directional` runs clean. So the walk records **zero** segments, `x` lands past its bound, and the Python-side clamp is what returns it — with nothing re-optimizing `y` behind a clamp.

## Why

`step_along_path` offers three breakpoint events, and none of them fires for a base-active bound the step is pushing *through*.

`boundcheck.rs:1397` suppresses the reach-a-bound event for exactly these rows, by design:

```rust
// A free variable reaching a bound. A bound the held
// factorization still enforces is not reachable this way: its
// variable sits essentially on it already, and holding it AGAIN
// through a Schur row would enforce the same bound twice. Such
// a bound leaves the active set only through its own
// multiplier's release below.
```

[...]

Issue #848, opened 2026-08-29
qp-active-set certifies Optimal at a strict saddle and on unbounded-below QPs, regressing #112 via #786


jkitchin:
## Summary

The active-set QP engine reports `Optimal Solution Found.` / `Solve_Succeeded` /
`success=True` at a **strict saddle** of an indefinite QP, and reports `Optimal`
with `obj = 0` on a problem that is unbounded below.

This regresses #112. #112 added the `check_psd` guard precisely because
`solve_qp` "accepts an indefinite `P` and returns a silently-wrong `optimal`".
#786 then scoped that guard away from `method='active-set'` on the premise —
stated in gh#786 and in the gh#786 commit message — that the active-set engine
"solves them **exactly**" and gives "a **local** optimum, the same guarantee the
NLP filter-IPM gives on a nonconvex NLP". It does not give that guarantee.

**A refusal became a confident wrong answer.** At `v0.10.0` `dispatch.rs` refused
the class outright; at HEAD it dispatches it:

```
-            if is_convex_qp {
-                Err(mismatch_msg(class, "qp-active-set", "an LP or convex QP"))
+            if is_convex_qp || class == P::NonconvexQp {
```
(`crates/pounce-cli/src/dispatch.rs:585`)

## Reproduction

`P = [[1, 5], [5, 1]]`, `c = 0`, box `[-1, 1]^2`. `numpy.linalg.eigvalsh(P) = [-4, 6]`.

Through `pounce.minimize`:

```
x0=[0.0, 0.0]     nlp            f=0              status=0 success=True x=[0. 0.]
x0=[0.0, 0.0]     qp-active-set  f=0              status=0 success=True x=[0. 0.]
x0=[0.3, -0.1]    nlp            f=-4             status=0 success=True x=[ 1. -1.]
x0=[0.3, -0.1]    qp-active-set  f=2.34382e-13    status=0 success=True x=[0. 0.]
x0=[0.99, -0.99]  nlp            f=-4             status=0 success=True x=[ 1. -1.]
x0=[0.99, -0.99]  qp-active-set  f=-5.640022e-11  status=0 success=True x=[0. 0.]
```

Started essentially **at** the global minimum (`x0 = [0.99, -0.99]`, `f = -3.92`),
the engine returns `f ~ 0` and certifies it. **It moved uphill and reported
success.** The start point is ignored entirely — all three starts land on `[0, 0]`.

Through `pounce.qp.solve_qp`:

```
[...]

PR #863, opened 2026-08-30
refactor(pyomo-pounce): one naming rule for the sensitivity surface

### LINKED ISSUE #854: sens: one naming rule for the sensitivity surface
The sensitivity surface's public names follow no rule, and three of them mislead. `estimate()` sits beside a parameter-estimation feature set (`declare_fitted`, `declare_residual`, `covariance()`), so it reads as "estimate the parameters" when it means the opposite: evaluate the solution at perturbed parameter values. `gradient()` returns a full Jacobian in its default form (`target=None` with an indexed param gives the matrix dx*/dp, and its own docstring says "the full Jacobian"), where the field reserves gradient for a scalar's derivative. And `wrt=` means two different things: the differentiation denominator on `gradient`, but the block the matrix is *of* on `covariance` and `information`. Beyond those three, the bare imports are vague: `from pyomo_pounce import estimate, covariance, gradient` puts three words in the caller's script that could belong to any library.

The proposal is one rule applied everywhere. `sens` marks the subsystem on every public name in the surface: queries read `sens_<returned object>`, answering from the sensitivity session, and declarations read `declare_sens_<role>`, registering a component with it. Across all signatures, `of=` marks what the answer is about and `wrt=` marks the differentiation variable.

| current | proposed |
|---|---|
| `estimate()` | `sens_solution()` |
| `estimate_report()` | `sens_solution_report()` |
| `gradient()` | `sens_jacobian()` |
| `covariance()` | `sens_covariance()` |
| `information()` | `sens_information()` |
| `active_set_changes()` | `sens_active_set_changes()` |
| `retain_kkt()` | `sens_retain_kkt()` |
| `release_kkt()` | `sens_release_kkt()` |
| `declare_fitted()` | `declare_sens_fitted()` |
| `declare_residual()` | `declare_sens_residual()` |
| `Gradient` | `Jacobian` |
| `EstimateReport` | `SolutionReport` |
| `gradient`'s `target=` | `of=` |
| `covariance`'s `wrt=` | `of=` |
| `information`'s `wrt=` | `of=` |

`declare_sens_param` already fits the rule and anchors the abbreviation. Unchanged: the result classes that already name their object (`SolutionMap`, `Covariance`, `Information`, `ActiveSetChange`), `continuation` and `shift_map` (not session queries), the option strings, the sIPOPT-compatible CLI suffixes (compatibility artifacts, never named after these functions), and the Rust crate's `parametric_step_*` surface.

Migration is a clean rename with no deprecation aliases: nothing here is in a release consumers pin against, and the book, notebooks, docstrings, and tests update in the same pass.

Feedback welcome on any row before this becomes a PR, on the rule as much as the names.

jkitchin:
Fixes #854

## Problem

The `pyomo-pounce` sensitivity surface's public names followed no rule, and three of them actively misled.

`estimate()` sat beside a parameter-estimation feature set — `declare_fitted`, `declare_residual`, `covariance()` — so it read as "estimate the parameters" when it meant the opposite: evaluate the solution at perturbed parameter values.

`gradient()` returned a full Jacobian in its default form. `target=None` with an indexed param gives the matrix `dx*/dp`, and its own docstring said "the full Jacobian" — where the field reserves *gradient* for a scalar's derivative.

`wrt=` meant two different things across the surface: the differentiation denominator on `gradient()`, but the block the matrix is *of* on `covariance()` and `information()`.

Beyond those three, the bare imports were vague: `from pyomo_pounce import estimate, covariance, gradient` puts three words in the caller's script that could belong to any library.

This is a naming change. No numerical behaviour moves, so there is no before/after/oracle table to give.

## Fix

One rule, applied everywhere. `sens` marks the subsystem on every public name in the surface: queries read `sens_`, answering from the sensitivity session, and declarations read `declare_sens_`, registering a component with it. Across all signatures, `of=` marks what the answer is about and `wrt=` marks the differentiation variable. `declare_sens_param` already fit the rule and anchors the abbreviation.

| before | after |
|---|---|
| `estimate()` | `sens_solution()` |
| `estimate_report()` | `sens_solution_report()` |
| `gradient()` | `sens_jacobian()` |
| `covariance()` | `sens_covariance()` |
| `information()` | `sens_information()` |
| `active_set_changes()` | `sens_active_set_changes()` |
| `retain_kkt()` | `sens_retain_kkt()` |
| `release_kkt()` | `sens_release_kkt()` |
| `declare_fitted()` | `declare_sens_fitted()` |
| `declare_residual()` | `declare_sens_residual()` |
| `Gradient` | `Jacobian` |
| `EstimateReport` | `SolutionReport` |
| `gradient`'s `target=` | `of=` |
| `covariance`'s `wrt=` | `of=` |
| `information`'s `wrt=` | `of=` |

Resulting signatures, read off the installed package:

```
sens_jacobian(of=None, *, wrt, max_pdpert=None)
sens_covariance(model, sigma_sq=None, n_data=None, hessian='lagrangian', of=None, max_pdpert=None)
sens_information(model, hessian='lagrangian', of=None, max_pdpert=None)
```

**No deprecation aliases**, per the issue. Nothing here is in a release consumers pin against. The stronger reason not to add them is that these functions emit diagnostics prefixed with their own name — `"sens_covariance: n_data ({n_data}) must exceed …"`, `"{who}: of= lists {name} twice"` — and those strings moved with the rename. An alias would have left a caller who typed `covariance(...)` reading an error that names `sens_covariance`, which is the same class of confusion the issue is about. Renaming the surface and the messages together is what keeps them consistent.

**Unchanged**, also per the issue: the result classes that already name their object (`SolutionMap`, `Covariance`, `Information`, `ActiveSetChange`), `continuation` and `shift_map` (not session queries), the option strings, the sIPOPT-compatible CLI suffixes, and the Rust crate's `parametric_step_*` surface. No `.rs` file appears in this PR's diff.

### Two names that only look like this surface

Both are called out because a mechanical rename gets them wrong, and one of them did before the tests caught it:

- **The cyipopt-style `gradient(self, x)` callbacks.** `_NlBridge.gradient` in `sens.py` and the same method on test problem classes are the callback protocol POUNCE's `Problem` calls into, not this API. Renaming `_ScalarStudy.gradient` in `test_covariance.py` made the solve return `Invalid_Number_Detected` (`AttributeError: '_ScalarStudy' object has no attribute 'gradient'`) and failed `test_classify_ratio_agrees_with_the_rust_classifier`. Reverted.
- **`active_set_changes` as a field and a key.** The dataclass field and dict key in `asnmpc_cstr.py`, and the `active_set_changes` key in the corrector report (produced by `pounce-py/src/solver.rs:668`), are the example's and the report's own metrics, not the accessor. The accessor calls around them were renamed; the field, the key, and the docstring sentence describing the key were not.

## Blast radius

Every caller of the surface, which is why the same pass moves all of them. Beyond the module and its exports: `continuation.py` (imports and calls three of these), `v2.py` and `pounce_solver.py` (doc comments), the tests, `docs/src/{sensitivity,pyomo,crossover,curve-fitting}.md`, `README.md`, `python/notebooks/README.md`, five notebooks, `validation/{p1_cho.py,p5_parametric.py,_covref.py,SOLVER_EQUIVALENCE.md}`, `python/pounce/examples/asnmpc_cstr.py`, `python/pounce/_curve_fit.py`'s see-also, and `dev-notes/covariance-information-design.md`, which opens by describing the uncertainty subsystem *as it stands* and so is a live reference rather than a record — it named `covariance()`, `information()`, `retain_kkt()`, `release_kkt()`, `declare_fitted`, `declare_residual` and `covariance(model, wrt=T)`, the last of these being the `wrt=` double meaning the issue's third complaint is about.

Two test files are renamed because they named the thing that changed: `tests/test_wrt.py` → `tests/test_of_block.py` and `tests/test_estimate_report.py` → `tests/test_sens_solution_report.py`. Notebook filenames are **not** renamed — `32_wrt_blocks_and_retain_kkt.ipynb` keeps its name because it is a published URL linked from the notebooks README and the book, though its title and contents are updated. One consequence worth knowing: the historical CHANGELOG entry at `CHANGELOG.md:8315` names `tests/test_estimate_report.py`, which no longer exists. Left alone — it describes the tree as it was at that release.

No solver behaviour changes, so no corpus sweep applies: `scripts/sweep-fixtures.sh` reads a binary this diff cannot affect.

### Keeping up with `main`, because it is the interesting part

`main` has moved under this branch twice since the PR opened, and **both merges were textually clean — which is the hazard, not the reassurance.** A rename branch cannot read "no conflicts" as "nothing to do": every commit that lands on the base while it is open is a fresh source of the old names, and git will merge them in without a murmur against a tree where they no longer exist.

**First merge (`c99269d`)** — gh #852's `mode="path"` weak-bound repair and gh #848's second-order certification. gh #852 added assertions to `test_degeneracy.py` calling `estimate()` and `active_set_changes()`; those four call sites were renamed, and `main`'s two new tests (`test_one_sided_path_keeps_the_coupled_kink_inside_its_box`, `test_one_sided_path_reholds_the_kink_under_user_scaling`) pass under the new API. They are the difference between the 492 the first push measured and the 494 below. The extension and CLI were rebuilt first, since `main`'s changes reach `pounce-sensitivity/src/{boundcheck,solver}.rs` and a green run against the pre-merge `.so` would have been worth nothing. The same pass swept the function references the first commit left in test comments and docstrings — `covariance()` and `information()` with their parentheses, invisible to a token-level rename because they sit inside strings — plus `validation/_covref.py`'s header. English uses are untouched: "the covariance matrix" and "zero information" are still the right nouns.

**Second merge (`1bacfa8`)** — gh #862's `solve_qp` P-shape guard. That one introduced no old names (it is `python/pounce/qp.py`, the pure-Python package, not this surface), and the scan says so rather than the merge being clean saying so.

**The second merge is worth reading for a different reason: the repo caught me.** The first suite run after it came back *490 passed, 8 skipped* against a 494/4 baseline. Not a regression — four correctness ratchets (`test_infeasibility_no_false_positives`, `test_issue_591_acceptable_level_status` ×2, `test_scale_invariance`) had **self-skipped**, refusing to measure because `target/release/pounce` was build `a1c0b12c+dirty` while the checkout was `1bacfa86`:

> refusing to measure … These probes are correctness ratchets — a different build's verdicts are not evidence either way, in either direction.

That guard is the same principle this PR's own rebuild discipline follows, enforced by the repo rather than by me remembering. Rebuilding the CLI at the merge commit put all four back: **494 passed, 4 skipped**. Had I taken 490/8 as "still green, four more environment skips", I would have shipped a number that was not measured.

## Tests

The suite is the pin, and it bites in the only way a rename can: the new test files import names that do not exist on the parent commit, so they fail at collection there, and the old names are gone from the package (`hasattr(pyomo_pounce, n)` is false for all twelve). What the suite is actually verifying is that the rename is *complete and internally consistent* — that no call site, `__all__` entry, monkeypatch target, or `pytest.raises(match=...)` string was left pointing at a name that no longer exists. Several were, before the run:

- `test_classify_ratio_agrees_with_the_rust_classifier` caught the cyipopt-callback over-rename described above.
- `test_asnmpc_example.py` pins the monkeypatch targets (`monkeypatch.setattr(asnmpc, "sens_solution", …)`) against `asnmpc_cstr.py`'s imports, so the example and its test had to move together.
- `test_of_block.py::test_of_error_paths` matches on `"sens_covariance: of= element 5"`, pinning the renamed message text.
- `test_degeneracy.py`'s two gh #852 tests caught the clean-merge hazard above.

Run on the merged tree at **`1bacfa8`**, against an extension and CLI rebuilt from it (`maturin develop --release`, `cargo build --release --bin pounce`), with `pyomo-cvp==0.7.2` installed so the asnmpc leg is not skipped:

- `pytest pyomo-pounce/tests` — **494 passed, 4 skipped**. The four skips are missing-bundled-binary environment checks in `test_binary_check.py` / `test_v2.py` (a source install has no wheel-bundled binary), unrelated to this diff.
- `pytest python/tests/test_sensitivity.py test_curve_fit.py` — 10 passed, 1 skipped (measured at `c99269d`; neither file is touched by this diff or by either merge).
- `scripts/check-release-consistency.sh` — OK.
- `scripts/check-docs-consistency.sh` — OK, 48 pages, no dead links. Internal anchors re-checked separately after the heading renames (`### What the step did about the bounds: sens_solution_report()`, `## Choosing the block: of=`, `## Keeping and releasing the factor: sens_retain_kkt(), sens_release_kkt()`), since the slugs moved with them.

**Deliberately not pinned, and one gap worth naming.** The notebooks' recorded outputs were edited textually alongside their source — the warning and error text they captured carries these names (`covariance: fitted parameter b has ambiguous bound activity…`, `information: the wrt block is rank-deficient…`), so leaving them would have left each notebook contradicting its own code. The rename is deterministic and no number moves, so the edit is exactly what a re-run would produce. But they were **not re-executed**, and `python/notebooks/README.md` claims every notebook is executed against the current release before it is committed. Notebooks 25, 26, 31, 32 and 36 should be re-run before the next release for that claim to hold literally.

That gap cost one real defect, found by scanning rather than by the suite, which cannot reach a notebook: notebook 31 called `sens_information(ms, wrt=[ms.A])` — the new name with the old keyword. `sens_information` takes `of=` and has no `**kwargs`, so that line raises `TypeError`, and the cell's recorded output is a bare `<Figure ...>` repr that shows nothing. Fixed. It is the only one: there is now no new-name/old-keyword pair anywhere in the tree, in any notebook, doc, test or module.

---

- [x] Tests fail on the parent commit for the stated reason — mechanically, per the caveat above: this is a rename, so the new files fail at import on the parent rather than pinning a behaviour that regressed
- [x] Book page under `docs/src/` updated — `sensitivity.md`, `pyomo.md`, `crossover.md`, `curve-fitting.md`; no new pages, so `SUMMARY.md` is unchanged
- [ ] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean — **not run**, and not applicable: no `.rs` or `Cargo.*` file appears in this PR's diff. (The merges brought `main`'s Rust changes into the branch, but they are `main`'s own, already reviewed on #859/#860/#865, and are not part of this diff.)
- [x] Every claim in this PR body and in the code comments is true of the code as it stands now — re-read against `1bacfa8` after the second merge; the head sha, the test-count provenance and the merge narrative moved with it


jkitchin:
`Test (ubuntu-latest)` failed on `4f86dfc` — `pounce-cli`'s `issue850_second_opinion_is_recorded::a_promotion_records_what_the_base_solve_did`, at `issue850_second_opinion_is_recorded.rs:152`:

```
assertion `left == right` failed
  left: None
 right: Some("SolveSucceeded")
```

**Not this PR's, and not by assertion — the Rust tree on this branch is byte-identical to `main`.** `git diff origin/main..HEAD -- '*.rs' 'Cargo.*' 'crates/**'` is empty; this PR is `pyomo-pounce` plus docs and notebooks. And `main`'s own CI run on `555fd660`, the exact commit this branch merged, **passed**. Same code, same workflow, opposite verdicts.

### It is a race in the test, and it is the same race the file already documents

`report()`'s own docstring records that this was hit and fixed once:

> The output paths carry the *options* as well as the fixture name. Three tests here solve `PROMOTING` … cargo runs them on parallel threads, so a filename keyed on the fixture alone has them overwriting each other's report and reading whichever landed last. That is a race in the test, and it showed up as exactly one intermittently-red assertion.

Keying on the options distinguishes `the_base_solver_alone_does_not_solve_this_fixture` (`retry=no`). It does **not** distinguish the two callers that pass no options at all — line 151 `a_promotion_records_what_the_base_solve_did` and line 172 `the_true_cost_is_recorded_not_just_the_promoted_rungs`. Both produce the same path:

```
pounce_850_square_flowsheet_resto_nl.json
```

so they still race, on exactly one pair. That explains the shape of the failure: `left` is `None`, not a wrong status — `string(&json, "status")` found no `status` key at all, which is a partial read of a file the sibling process was rewriting, not a solver disagreeing.

**Reproduced and diagnosed locally**, on the identical tree: **1 failure in 20 runs** at `--test-threads=8`. It is intermittent, not a phantom.

### Proposed patch (verified, deliberately not pushed here)

Keying the path on the running test as well closes the remaining pair. Applied locally: **0 failures in 25 runs** under the same stress.

```rust
// Keyed on the running test as well as the fixture and options: two
// tests here solve PROMOTING with NO options, so an options-derived tag
// collides between exactly those two and they race on one path again.
let who = std::thread::current().name().unwrap_or("main").to_string();
let tag: String = format!("{name}{}_{who}", extra.join("_"))
    .chars()
    .map(|c| if c.is_ascii_alphanumeric() { c } else { '_' })
    .collect();
```

I have **not** pushed it. This PR is a naming change in `pyomo-pounce` with no `.rs` in its diff, and carrying an unrelated Rust test fix through it would widen it and make the diff harder to review than the fix is worth. It belongs in its own commit on `main` — happy to open that PR if wanted.

Re-running the failed job once, per the one-re-run rule. Since the flake is ~1-in-20 the re-run will very probably be green, and that is confirmation rather than a fix: **the race is still live on `main`** and will keep surfacing on unrelated PRs until the patch above (or an equivalent) lands.

---

PR #864, opened 2026-08-30
Add support for FBBT in `pounce-rs`

GermanHeim:
## Problem

FBBT is implemented in `pounce-presolve`, but the `pounce-rs` builder can't supply structural constraint expressions. Setting `presolve_fbbt=yes` is a silent no-op for `Problem` implementations.

## Cause

It wasn't added in https://github.com/jkitchin/pounce/pull/363 because of the PR's scope.

## Fix

- Added `Problem::constraint_expression`, returning an `FbbtTape`.
- Installed a provider-aware presolve wrapper when both `presolve=yes` and `presolve_fbbt=yes`.
- Added `Solution::fbbt_report`.
- Added `NlpError::Presolve` and `NlpError::InvalidFbbtTape`. 

## Blast radius

`pounce-rs`

## Tests

      since moved

> Docs and CHANGELOG updated using Codex


crates/pounce-algorithm/src/application.rs
```diff
@@ -990,7 +1000,7 @@ impl IpoptApplication {
         // against finite differences before anything else runs — on the
         // raw TNLP, before the presolve wrapper below changes its
         // coordinates.
-        self.run_derivative_test(&tnlp);
+        self.run_derivative_test(&derivative_test_tnlp);
```
jkitchin: This changes `derivative_test` for **every** frontend, not just `pounce-rs`.

`optimize_tnlp` hands in `Rc::clone(&tnlp)` taken *before* the wrappers exist, but inside this method `install_variable_scaling` (line 897) and `install_start_conditioner` (line 913) **shadow** `tnlp`. So this call no longer sees either of them, where on the parent it saw both.

Measured, same probe on both commits (`start_point_perturbation=0.5`, `derivative_test=first-order`, `x0 = 1.0`), recording the point `gradient` is first called at:

| commit | first point the checker evaluates at |
|---|---|
| `73e064c` (base) | `1.7666216164272852` — the displaced start |
| `773518e` (this PR) | `1.0` — the raw start |

The same shadowing takes the `nlp_scaling_method=user-scaling` / `curvature-based` substitution out of the check, so the scaling wrapper's own arithmetic is no longer exercised by it. This reaches the CLI, `pounce-py`, `pounce-cinterface` and `pounce-wasm`, all of which funnel through `optimize_tnlp` — outside the PR body's stated blast radius, absent from the CHANGELOG, and untested: `crates/pounce-cli/tests/derivative_test_option.rs` runs the `user_scaling_suffix.nl` fixture but never sets `nlp_scaling_method`, so the wrapper never installs there.

Cheapest fix that keeps what the builder wants and costs nothing elsewhere: make the parameter `Option<Rc<RefCell<dyn TNLP>>>`, and on `None` run the check on the post-wrapper `tnlp` exactly as today.

The comment just above ("on the raw TNLP, before the presolve wrapper below changes its coordinates") now describes three wrappers rather than one, and needs updating either way.

---

crates/pounce-rs/src/builder.rs
```diff
@@ -63,6 +64,11 @@ pub trait Problem {
     /// Constraint values `g(x)` into `out` (length `n_constraints`).
     fn constraints(&self, _x: &[f64], _out: &mut [f64]) {}
 
+    /// Optional FBBT expression for constraint `index`.
+    fn constraint_expression(&self, _index: usize) -> Option<FbbtTape> {
```
jkitchin, line 70: This hook asks the user to restate each constraint in a second, hand-written form, and nothing cross-checks the two. A tape that disagrees with `constraints()` silently produces a wrong answer reported as `SolveSucceeded`.

Measured on this branch — model `min (x0-3)²` s.t. `g(x) = x0 ≤ 5`, `x ∈ [0,10]`, with a typo'd tape saying `10·x0`:

```
status=SolveSucceeded  x=[0.5000000099980002]  obj=6.2499999
```

True answer is `x = 3`, `obj = 0`. FBBT tightened `x ≤ 0.5` off the wrong tape and the solver certified the result. `.nl` cannot hit this — `NlTnlp` derives the tape from the same graph it evaluates — so this PR introduces the first source where the two can disagree.

At minimum, say so plainly here and in `README.md` / `docs/src/rust.md` / `docs/src/fbbt.md`: the tape must be an exact restatement of `constraints(index)`, and a mismatch cuts off the optimum with no diagnostic.

Better: `try_solve` already walks every tape to validate it, so it could also evaluate each one as a degenerate interval at a sample point (x0 and the box midpoint) and compare against `constraints()`. A smoke test rather than a proof, but it catches the case above at the midpoint (50 vs 5).

---

crates/pounce-rs/src/builder.rs
```diff
@@ -487,6 +551,38 @@ struct Adapter<P: Problem> {
     sol_z_u: Vec<f64>,
 }
 
+fn validate_fbbt_tape(
```
jkitchin, line 554: Both checks here duplicate ones `forward_pass` already makes (`ForwardError::MalformedTape`, `ForwardError::VariableIndexOutOfRange`) — but promoting them from "skip this constraint silently" to "refuse the solve" is a real improvement, and worth stating in the docs that the builder is stricter than the `.nl` path.

What the validator misses is a non-finite `Const`. `Interval::point(NaN)` is empty, so the forward pass returns `Ok` with an empty root and `run_fbbt` records a witness and returns on the spot. Measured on a feasible model whose tape is `x0 + NaN`:

```
report = FbbtReport { iterations: 1, bound_updates: 0, infeasibility_witness: Some(0), total_tightening: 0.0 }
status = SolveSucceeded
```

So the report claims an infeasibility witness on a model that solves, and every later constraint silently loses its tightening for that sweep. Given `3993237` ("nl: refuse non-finite numbers instead of silently dropping them"), a non-finite `Const(c)` belongs in this validator alongside the two checks that are here.

---

crates/pounce-rs/src/builder.rs
```diff
@@ -549,7 +645,9 @@ impl<P: Problem> TNLP for Adapter<P> {
                 }
             }
             SparsityRequest::Values { values } => {
-                let x = x.expect("eval_jac_g(Values) without x");
+                let Some(x) = x else {
```
jkitchin: This hunk is unexplained by the PR body and appears unrelated to FBBT.

Nothing in the new chain calls `Values` with `x = None`: `PresolveTnlp::eval_jac_g` forwards whatever it is given, and `LinearEqElimTnlp` always passes `Some(&x_probe)` for `Values` (it uses `None` only for `Structure`). I restored the `expect` on this branch and the whole `pounce-rs` suite is green — 17 unit + 22 integration + 5 doctests — so no path introduced here needs it.

`false` from `eval_jac_g` means "evaluation failed" to the solver, so this converts a panic that names the bug into a confusing solve status. Either drop the hunk, or add the test that motivates it plus a comment saying which caller passes `None` here.

---

crates/pounce-rs/src/builder.rs
```diff
@@ -405,12 +407,71 @@ impl<P: Problem + 'static> Nlp<P> {
...
+            let presolve = Rc::new(RefCell::new(
+                pounce_presolve::PresolveTnlp::with_expression_provider(
+                    Rc::clone(&adapter) as Rc<RefCell<dyn TNLP>>,
+                    provider,
+                    presolve_opts,
+                ),
+            ));
+            fbbt_handle = Some(Rc::clone(&presolve));
+            let mut wrapped: Rc<RefCell<dyn TNLP>> = presolve;
+            if presolve_opts.linear_eq_reduction {
```
jkitchin: This branch — the one pass that changes the variable count — has no test, and for a typical builder `Problem` it silently does nothing, so a mistake here would leave no trace.

Probed on this branch with a linear-equality model (`min (x0-1)²+(x1-1)²` s.t. `x0+x1 = 3`, tape supplied), `presolve_linear_eq_reduction=yes` vs `no`:

```
elim=no : x=[1.5, 1.5] obj=0.5 lam=[-1.0] report=FbbtReport { iterations: 2, bound_updates: 2, ... }
elim=yes: x=[1.5, 1.5] obj=0.5 lam=[-1.0] report=FbbtReport { iterations: 2, bound_updates: 2, ... }
```

Byte-identical, because the `Adapter` declines the Hessian structure and `LinearEqElimTnlp` needs `nnz_h_lag` up front. Either exercise it with a `Problem` that supplies `hessian`, or drop the stacking until something reaches it.

Two neighbouring branches are also untested, per the repo's "check which branch the fixture takes" rule: `presolve=yes, presolve_fbbt=no` (the builder does not wrap, the application does — no test that `fbbt_report` is `None` and presolve still runs), and a `Problem` that leaves `constraint_expression` at its default `None` with FBBT on. And `presolve_fbbt=yes` with `presolve=no` is silently a no-op (measured `report=None`) — worth a documented note or a warning.

---

crates/pounce-rs/Cargo.toml
```diff
@@ -24,6 +24,7 @@ pounce-observability.workspace = true
 # This is the one crate the lean default build gains over the pre-ladder
 # facade.
 pounce-restoration.workspace = true
+pounce-presolve.workspace = true
```
jkitchin, line 26: This lands directly under the comment ending "This is the one crate the lean default build gains over the pre-ladder facade" — there are now two, so that sentence is no longer true as written. It is also the only dependency in this file without a rationale comment; worth one saying that `pounce-algorithm` already pulls `pounce-presolve`, so this widens the public surface rather than the build.

---

crates/pounce-rs/src/lib.rs
```diff
@@ -211,11 +211,13 @@ pub use pounce_algorithm::application::IpoptApplication;
 // --- iteration capture & observability --------------------------------------
 // Thread-scoped helpers so an embedding library can record a solve's
 // trajectory (and turn on console logs) with no direct `tracing` deps.
+pub use pounce_nlp::expression_provider::{FbbtOp, FbbtTape};
```
jkitchin: These two re-exports (and `FbbtReport` on line 220) sit under the `--- iteration capture & observability ---` header, which does not describe them. They want their own section, e.g. `--- presolve / FBBT ---`.

Related, in `builder.rs`: `Solution::fbbt_report` is typed `pounce_presolve::fbbt::FbbtReport` even though line 220 re-exports it at the crate root — `FbbtReport` reads better and matches what a caller names.

---

docs/src/fbbt.md
```diff
@@ -22,10 +22,9 @@ References: Belotti, Cafieri, Lee, Liberti (2010).
 
 FBBT cannot help when:
 
-- The TNLP has no structural-expression representation. Today only
-  `.nl`-loaded problems (`NlTnlp`) expose one. Python (`PyTnlp`),
-  C-callback (`CCallbackTnlp`), and Rust closure-based problems
-  silently opt out.
+- The TNLP has no structural-expression representation. `.nl`-loaded
+  problems (`NlTnlp`) expose one, and `pounce-rs` builder problems can opt in
```
jkitchin, line 26: The rewrite drops "Python (`PyTnlp`), C-callback (`CCallbackTnlp`), and Rust closure-based problems silently opt out." That is still true of `PyTnlp` and `CCallbackTnlp`, and it is removed from precisely the "FBBT cannot help when" list where a Python or C user would look for it. Keep the sentence, minus the `pounce-rs` half that this PR fixes.

---

jkitchin (changes requested):
Reviewed against the actual base (`73e064c`), built, and probed empirically. The FBBT wiring itself is sound and follows the CLI's pattern; two things want fixing before merge, and there are some coverage and doc gaps.

**1. `derivative_test` now runs on the un-scaled, un-conditioned TNLP — for every frontend, not just `pounce-rs`.** `install_variable_scaling` and `install_start_conditioner` shadow `tnlp` inside the new method, so the check no longer sees either. Measured: with `start_point_perturbation=0.5` and `x0 = 1.0`, the checker evaluates at `1.7666216164272852` on `73e064c` and at `1.0` here. That reaches the CLI, `pounce-py`, `pounce-cinterface` and `pounce-wasm`, is outside the stated blast radius, and is untested — the CLI's `derivative_test_option.rs` uses the `user_scaling_suffix.nl` fixture but never sets `nlp_scaling_method`, so the scaling wrapper never installs there. Making the parameter an `Option` and falling back to the post-wrapper `tnlp` on `None` keeps the builder's behaviour at zero cost elsewhere. Details inline on `application.rs`.

**2. A tape that disagrees with `constraints()` silently produces a wrong answer.** Measured: `min (x0-3)²` s.t. `x0 ≤ 5` with a typo'd `10·x0` tape returns `SolveSucceeded`, `x = 0.5`, `obj = 6.25`, against a true `x = 3`, `obj = 0`. `.nl` cannot hit this because `NlTnlp` derives the tape from the graph it evaluates; this PR introduces the first source where the two can disagree. The contract needs stating loudly in the trait doc and the three doc pages, and a cheap consistency check at a sample point would catch the common typo. Inline on `builder.rs`.

**Also inline:** `validate_fbbt_tape` accepts a non-finite `Const`, which turns a feasible model into an `infeasibility_witness` and aborts the sweep; the `eval_jac_g` `expect` → `return false` hunk is unexplained and unneeded (restoring the `expect`, the whole `pounce-rs` suite stays green); the `presolve_linear_eq_reduction` stacking is untested and silently no-ops for a builder `Problem` without an exact Hessian; plus a stale Cargo.toml comment, a misplaced re-export section, and a docs sentence about `PyTnlp` / `CCallbackTnlp` that was true and got dropped.

The CHANGELOG should also mention the new public `IpoptApplication::optimize_tnlp_with_derivative_test_tnlp` and, if it stays, the behaviour change in item 1.

**What's right:** wrapper ordering (presolve inside scaling) matches the CLI and is the correct frame for tapes written in user-variable space; provider and inner TNLP are the same object, as with `NlTnlp`; tapes are captured once at build time rather than re-entering user code every sweep; `Solution` and `NlpError` are `#[non_exhaustive]`, so the new field and variants are not a semver break; `cargo fmt --all -- --check` is clean and clippy adds no new errors.

---

crates/pounce-rs/src/builder.rs
```diff
@@ -487,6 +551,107 @@ struct Adapter<P: Problem> {
...
+
+fn validate_fbbt_tape_values<P: Problem>(
+    problem: &P,
+    tapes: &[Option<FbbtTape>],
+    x0: &[f64],
+    x_l: &[f64],
+    x_u: &[f64],
+    n_constraints: usize,
+) -> Result<(), NlpError> {
+    let midpoint: Vec<_> = x_l
```
jkitchin, line 593: The two samples collapse to one whenever the box midpoint equals the starting point — and that is the *default* configuration, since `x0` defaults to `vec[0.0; n]` and a symmetric or zero-centred box is common. The check then passes any tape that happens to agree at that single x.

Measured on `11809b2`, the same typo from `fbbt_tape_mismatch_is_rejected_before_solving` (tape `10·x0`, model `x0`), moved onto a symmetric box so the midpoint lands on the default start:

```rust
.var_bounds(&[-10.0], &[10.0])   // midpoint 0.0 == default x0 = 0.0
.constraint_bounds(&[-2.0e19], &[5.0])
```
```
accepted: status=SolveSucceeded  x=[0.5000000099980003]  obj=6.2499999
```

True answer `x = 3`, `obj = 0` — the same wrong answer the check was added to catch, straight through. `total_tightening: 9.5` in the report shows FBBT acting on the bad tape.

Cheap repair: make the second sample provably distinct from `x0` rather than fixed at the midpoint — e.g. take `lo + 0.25·(hi − lo)` and fall back to `lo + 0.75·(hi − lo)` if that coincides with the start. A third sample would be better still. Either way it stays a smoke check, which the docs now say plainly, so this is about not having a hole exactly where the defaults put every user.

---

jkitchin (commented):
Re-reviewed at `11809b2`. All five blocking items are addressed, and I verified each one rather than reading the diff:

**1. Derivative test** — fixed as suggested (`Option`, `None` → post-wrapper `tnlp`). The two new tests are mutation-checked: I reverted `optimize_tnlp` to pass `Some(Rc::clone(&tnlp))` and exactly `ordinary_derivative_test_keeps_the_conditioned_start` and `ordinary_derivative_test_keeps_variable_scaling` go red (`left: [1.0, 1.5]`, `right: [1.0, 2.0]`), the other 516 stay green. That is the shape a regression test should have.

**2. Tape/`constraints()` mismatch** — the sampled cross-check catches the case I measured, and the docs in all four places now state the contract and call the check a smoke test rather than a proof. One residual hole, inline: the two samples coincide under the *default* `x0`, and the same wrong answer walks through. Not blocking given the docs, but it is a hole exactly where the defaults put people.

**3. Non-finite `Const`** — rejected, with all three of NaN / ±inf covered.

**4. `eval_jac_g`** — `expect` restored.

**5. Phase-6 stacking** — dropped, and I owe you a correction on my own comment: I attributed its inertness to the `Adapter` declining the Hessian, and that is not the gate. `nnz_h > 0 && !eval_h(...)` short-circuits when `nnz_h_lag == 0`, so the Hessian is not what stops it. The real gate is `TNLP::get_constraints_linearity`, which the `Adapter` does not implement, so `have_linearity == false` and every row lands as `None` in `linear_row_map` (`pounce-presolve/src/lib.rs:947`). Dropping the stacking is therefore a genuine no-op — right call — but the reason is worth recording because of what else follows from it: for a builder `Problem`, Phase 1 linear bound tightening, redundant-row removal and Phase 6 are *all* inert, so `presolve=yes` alone does nothing but install a wrapper. FBBT is the only pass that can act, which makes this PR the whole of presolve for `pounce-rs`. Measured: objective-call counts identical across all four `presolve_linear_eq_reduction` × `presolve_fbbt` combinations.

That last point suggests the natural follow-up: the tape German just added already answers the linearity question. A tape whose ops are only `Const` / `Var` / `Add` / `Sub` / `Neg` and constant-factor `Mul` *is* a linear row, so `constraint_expression` could feed `get_constraints_linearity` and light up the three passes that are dark today — no new user-facing API.

`cargo test -p pounce-rs` (48 tests across 8 binaries), `cargo test -p pounce-algorithm --lib` (518) and `cargo fmt --all -- --check` are clean at this head.

---

Issue #840, opened 2026-08-29
v0.11 release


jkitchin:
I am planning to make a new release in the next week or so (no specific urgency but there are several new features that are worth a new release). 

@srikanth-gm The Casadi interface is one of the new features. I think it is in pretty good shape. After several days of chasing performance in linear algebra, I am pretty sure there are not easy to find further gains to be had. If there are other interface things to polish prior to release let me know. One outstanding issue related to Casadi is https://github.com/jkitchin/pounce/issues/782. I am looking into a solution, but your opinion on how important it is would be helpful. May be fixed in https://github.com/jkitchin/pounce/pull/843.

@devin-griff The sensitivity work is also substantially improved I think. Let me know if you would like to finish up any specific things, or if I can help advance that.

@GermanHeim I don't think we will make significant changes related to https://github.com/jkitchin/pounce/issues/649 in the next release. Let me know if you anticipate any PRs / feature needs for oximo that make sense now though.

@bernalde It would be useful to land https://github.com/jkitchin/pounce/issues/780 in a new release because it will make development on the discopt side easier if we can bump its pounce dependency to a released version. I guess it is not critical though, if we don't get it complete or right (and we probably won't know that fully until the discopt side lands too there will have to be another release in the future anyway). I started work on https://github.com/jkitchin/pounce/issues/794, and expect a PR for review in the next day.

Also worth noting, the semester just started and I have been juggling a lot of balls in the air. If I missed anything in the PRs / issues above let me know!

I can't think of any major new things I want to push beyond these right now. I'd like to start shifting focus to applications using pounce, and keep pushing on discopt.

Thanks for your interest and participation in pounce!

jkitchin:
## Pre-0.11.0 adversarial audit — blockers filed

An 11-agent adversarial sweep over the PRs merged since `v0.10.0`. Every finding
below I reproduced independently before filing; the full triage list (including
should-fix and doc-drift items not yet filed) is in
`adversary/release-audit/TRIAGE.md`.

**Filed as release blockers — wrong answers reported as success:**

- #845 — sensitivity corrector returns an all-NaN step and reports `residual=0.0`,
  `converged=True`, `improved()=True`. New in 0.11.0.
- #846 — convex `σ` cost-normalization returns a wrong `x` under `Optimal` at the
  **default** tolerance; 84 of 157 on a generated family. Pre-existing.
- #847 — `.nl` reader silently drops non-finite bounds and objective constants,
[...]

PR #868, opened 2026-08-31
chore(release): 0.11.0 preparation — version surfaces, CHANGELOG, and two procedure fixes

jkitchin:
**Preparation only — this does not cut 0.11.0.** Bug fixes and a full
benchmark run still have to land first. This PR gets the mechanical surfaces
and the release procedure in order so that when the release is actually cut,
the only outstanding work is the release itself.

Nothing here touches `crates/*/src`. No functional change to the solver, and
no trajectory change, so no fixture sweep applies.

Part of #840.

## Blocking the release, not this PR

- [ ] outstanding bug fixes
- [ ] full benchmark run
- [ ] re-date before tagging (see below)

## Two dates that will be wrong by the time we tag

`CHANGELOG.md` says `## [0.11.0] - 2026-08-31` and `CITATION.cff` says
`date-released: 2026-08-31`, because that is when the bump was prepared.
Since the release is now gated on the work above, **both must be reset to the
actual release date before the `v0.11.0` tag is pushed.**

`scripts/check-release-consistency.sh` does *not* check either one — it
verifies version agreement across the surfaces, not dates — so nothing will
catch this automatically. It is on the checklist above and in the "before
tagging" list at the end.

## The version bump

All six surfaces `dev-notes/cargo-release.md` lists, plus the docs pins
`check-release-consistency.sh` reports as advisory:

| surface | file |
|---|---|
| workspace + 17 dep self-pins | root `Cargo.toml` (18 sites) + `Cargo.lock` |
| PyPI `pounce-solver` | `python/pyproject.toml` |
| PyPI `pyomo-pounce` | `pyomo-pounce/pyproject.toml` |
| release image | `docker/Dockerfile.release` `ARG POUNCE_VERSION` |
| citation | `CITATION.cff` (`version`, `date-released`; concept `doi:` stays) |
| CHANGELOG | `[Unreleased]` → `[0.11.0]`, fresh `[Unreleased]`, link refs |

`check-release-consistency.sh` passes with **no advisory lines**.

Historical prose is deliberately untouched. `docs/src/` carries a lot of "up
to and including 0.10.0", plus pinned 0.10.0 fixture tables in
`initialization.md` and `troubleshooting.md`. Those are claims *about that
release* and stay at 0.10.0; only install and pull pins moved. One of them,
`docker/README.md`, was already internally inconsistent before this PR —
`-t pounce:0.10.0` built with `--build-arg POUNCE_VERSION=0.9.0`.

Note that promoting `[Unreleased]` now means any PR merged between this one
and the release adds its entry under the *new* `[Unreleased]` heading, not
under `[0.11.0]`. The bug fixes still to land will need their entries moved
into the `[0.11.0]` section before tagging, or they will silently document
themselves as post-0.11.0.

## CHANGELOG coverage was verified, not assumed

The `[0.11.0]` section is 200 bullets over ~8 900 lines. I mapped all **149
PRs merged since `v0.10.0`** to their closing issues and cross-checked both
numbers against the section. Everything user-facing is represented. The one
apparent gap, #784 (degenerate-start diagnosis), is covered under different
wording. The remaining uncited PRs are CI, test, and dev-note changes that
correctly get no entry.

## Two corrections that are not version bumps

**1. Pre-flight step 5 described a check that cannot pass.** It said to run
`scripts/publish-crates.sh --dry-run` and claimed it "dry-runs every crate
end-to-end, so any breakage appears here, not three crates into the real
release." It cannot, and it fails in the only situation it is ever run in:

```
[2/20] cargo publish -p pounce-linalg --dry-run
error: failed to select a version for the requirement `pounce-common = "^0.11.0"`
  candidate versions found which didn't match: 0.10.0, 0.9.0, 0.8.0, ...
```

`cargo publish --dry-run` resolves against the **live** index, and step 2 of
the same pre-flight just bumped the version. Only the two leaves can pass.
`cargo package --no-verify` hits the identical wall — it is the resolution
that fails, not the build. A step that always fails and carries no
information is worse than no step: it trains you to skip it on the one run
where a real packaging defect would surface.

Step 5 now runs `cargo package --workspace` (four `publish = false` crates
excluded, and *without* `--no-verify`, since dropping the compile drops most
of the value), which resolves intra-workspace deps against the versions it is
packaging locally. Verified against this tree: **exit 0, 20 packaged, 20
verified**. The caveat is also annotated at the `--dry-run` flag in the
script's own usage block, and the doc records that `release-crates.yml`'s
`workflow_dispatch` dry-run mode inherits the same limitation — it shells out
to this same script.

**2. A base-image comment that the bump falsified.**
`docker/Dockerfile.release` justified its bookworm base with "the fix reached
PyPI in 0.10.0, **which is what this image installs**" — true while the ARG
said 0.10.0, false the moment it says 0.11.0. The glibc floor (`GLIBC_2.16`)
was measured on the published 0.10.0 wheel, and 0.11.0 is not on PyPI, so the
comment now says 0.10.0 is the most recent release the floor was measured on
and points at `scripts/check-cli-portability.sh`, rather than implying a
measurement nobody has made.

## What the green docker check does and does not prove

`build release` passes here, but the plan job logs why:

> PR rot guard: building the release image against the newest published
> version (0.10.0), not the workspace version (0.11.0).

The release image pip-installs a pinned version from PyPI, so on a PR it can
only build against something actually published. The check proves
`Dockerfile.release` still builds; it does **not** validate the 0.11.0 image,
which is first built at tag time. Verifying the published image stays a real
post-release step.

## Before tagging (not in this PR)

- Reset the CHANGELOG `[0.11.0]` date and `CITATION.cff` `date-released` to
  the real release date.
- Move any post-merge CHANGELOG entries from `[Unreleased]` into `[0.11.0]`.
- Tag in order — `python-v0.11.0` and `pyomo-pounce-v0.11.0` **at or before**
  `v0.11.0`, since the release image pip-installs from PyPI but fires on `v*`.
- `gh release create v0.11.0 --notes-file` — by hand, no workflow makes it.
- `docker run --rm ghcr.io/jkitchin/pounce:0.11.0 --version` to verify the
  published image rather than assuming.
- Re-measure the CLI glibc floor on the published 0.11.0 wheel and update the
  `Dockerfile.release` comment to cite it.

https://claude.ai/code/session_01HVXWp9fLta3pQTYF2aaxEB

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ changes.
 
 ## [Unreleased]
 
+
+## [0.11.0] - 2026-08-31
+
 - **Added FBBT support to `pounce-rs`.** `presolve_fbbt=yes` now runs
   feasibility-based bound tightening when an `ExpressionProvider` is
   available; `Solution::fbbt_report` exposes diagnostics and
@@ -17881,7 +17884,8 @@ release.
 - Zenodo metadata (`.zenodo.json`) and `CITATION.cff` for
   archival on every GitHub Release.
 
-[Unreleased]: https://github.com/jkitchin/pounce/compare/v0.10.0...HEAD
+[Unreleased]: https://github.com/jkitchin/pounce/compare/v0.11.0...HEAD
+[0.11.0]: https://github.com/jkitchin/pounce/releases/tag/v0.11.0
 [0.10.0]: https://github.com/jkitchin/pounce/releases/tag/v0.10.0
 [0.9.0]: https://github.com/jkitchin/pounce/releases/tag/v0.9.0
 [0.8.0]: https://github.com/jkitchin/pounce/releases/tag/v0.8.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ changes.
 
 
 ## [0.11.0] - 2026-08-31
+
 - **The negative-curvature escape now finds the witness it is asked for, on
   saddles whose curvature lies along a coordinate axis (gh #797 follow-up).**
   gh #797 added the second-order question and `neg_curv_escapes` to answer it;
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -15,8 +15,8 @@ identifiers:
   - type: doi
     value: "10.5281/zenodo.20387011"
     description: "Concept DOI — always resolves to the latest POUNCE release"
-version: 0.10.0
-date-released: 2026-08-10
+version: 0.11.0
+date-released: 2026-08-31
 authors:
   - family-names: Kitchin
     given-names: John R.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -573,7 +573,7 @@ checksum = "a6cb138bb79a146c1bd460005623e142ef0181e3d0219cb493e02f7d08a35695"
 
 [[package]]
 name = "iter-diff"
-version = "0.10.0"
+version = "0.11.0"
 dependencies = [
  "serde_json",
 ]
@@ -868,7 +868,7 @@ dependencies = [
 
 [[package]]
 name = "pounce-algorithm"
-version = "0.10.0"
+version = "0.11.0"
 dependencies = [
  "anstream",
  "anstyle",
@@ -890,7 +890,7 @@ dependencies = [
 
 [[package]]
 name = "pounce-cinterface"
-version = "0.10.0"
+version = "0.11.0"
 dependencies = [
  "pounce-algorithm",
  "pounce-common",
@@ -907,7 +907,7 @@ dependencies = [
 
 [[package]]
 name = "pounce-cli"
-version = "0.10.0"
+version = "0.11.0"
 dependencies = [
  "anstream",
  "anstyle",
@@ -938,7 +938,7 @@ dependencies = [
 
 [[package]]
 name = "pounce-common"
-version = "0.10.0"
+version = "0.11.0"
 dependencies = [
  "anstyle",
  "anstyle-query",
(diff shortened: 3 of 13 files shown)


jkitchin:
## Pre-flight step 5: packaging dry run — clean

Ran the replacement check this PR installs, against the bumped tree:

```
cargo package --workspace --exclude pounce-py --exclude pounce-studio-pyo3 \
              --exclude iter-diff --exclude pounce-wasm
```

**exit 0 — 20 packaged, 20 verified**, i.e. every publishable crate packaged
*and* compiled from its packaged tarball at 0.11.0:

> pounce-common pounce-linalg pounce-linsol pounce-feral pounce-hsl
> pounce-nlp pounce-l1penalty pounce-observability pounce-presolve pounce-qp
> pounce-solve-report pounce-algorithm pounce-restoration pounce-sensitivity
> pounce-cinterface pounce-convex pounce-nl pounce-studio-core pounce-cli
> pounce-rs

That set matches `scripts/publish-crates.sh`'s `CRATES=(...)` exactly, which
`check-release-consistency.sh` independently confirms is the full publishable
set in topological order.

For contrast, the check this replaces stopped at crate 2 of 20 on the same
tree, for the reason the PR describes.

### One warning, pre-existing, not filed as a blocker

```
warning: variable does not need to be mutable
   --> crates/pounce-restoration/src/aug_resto_system_solver.rs:635:9
635 |     let mut pack = |cols: Vec<Vec<Number>>| -> Option<Rc<MultiVectorMatrix>> {
```

Introduced by `88811831` ("Stop restoration densifying the L-BFGS Hessian",
#684, 2026-08-18), so it predates this branch — this PR changes no
`crates/*/src`. It is cosmetic, but it *is* emitted to anyone who builds
`pounce-restoration` from crates.io, which is why it is worth recording here
rather than leaving in a log. Deliberately not fixed in this PR: keeping the
release commit to version strings and docs is worth more than the one-word
cleanup. Happy to send it separately either way.

Issue #818, opened 2026-08-28
[adversary] nlp: L-BFGS arm exhausts max_iter on a 4-variable convex quadratic that scipy L-BFGS-B solves in 34 iterations at matched memory


jkitchin:
adversary-fingerprint: nlp/lbfgs-stall-illconditioned-unconstrained-quadratic

Filed by the nightly adversary job. The reported **status is honest** (`Maximum_Iterations_Exceeded`,
`success=False`) — this is a convergence-rate/robustness defect on the limited-memory
path, not a wrong-answer-under-success defect.

## Summary

`pounce.minimize` with a gradient but no Hessian (which selects
`hessian_approximation=limited-memory`, the documented default and the mode the Python
frontend and the CasADi plugin pick automatically) fails to converge on a **4-variable,
unconstrained, strictly convex, separable quadratic**. `scipy.optimize.minimize`'s
L-BFGS-B solves the identical problem in 34 iterations **at the same memory size**.

## Reproduction

```python
import numpy as np, pounce
from scipy.optimize import minimize as smin
# Unconstrained, strictly convex, separable quadratic. Exact optimum x_i = 1/s_i, f* = 0.
n = 4; s = 10.0 ** np.linspace(0, 4, n)           # [1, 21.54, 464.2, 1e4]; cond(H) = 1e8
f = lambda x: float(np.sum((s * x - 1) ** 2))
g = lambda x: 2 * s * (s * x - 1)
xex = 1.0 / s

r = pounce.minimize(f, np.zeros(n), jac=g, max_iter=2000)          # default => limited-memory
o = smin(f, np.zeros(n), jac=g, method="L-BFGS-B",
         options=dict(maxiter=50000, maxfun=99000, ftol=1e-18, gtol=1e-14))

print("pounce :", r.message, "nit", r.nit, "f", r.fun,
      "rel x err", np.max(np.abs(r.x - xex) / xex))
print("scipy  :", "nit", o.nit, "f", o.fun,
      "rel x err", np.max(np.abs(o.x - xex) / xex))
```

## Expected

`f* = 0` at `x* = (1, 4.642e-2, 2.154e-3, 1e-4)`. Three independent oracles:

1. **closed form** — the objective is separable, so `x_i* = 1/s_i` exactly and `f* = 0`;
2. **scipy L-BFGS-B** (default memory `m = 10`) — 34 iterations, `f = 1.50e-23`,
   rel x err `3.2e-12`;
[...]

PR #885, opened 2026-09-01
gh#884: a settled point with a runaway multiplier — detect it late, re-solve cold, promote only on a certificate

### LINKED ISSUE #884: Biactive complementarity pair: dual diverges at a reachable KKT point, exits Error_In_Step_Computation
Split out of #794 / PR #844 as "P2", where it is the `POUNCE candidate` ownership bucket (8 of 576 triaged observations). The gate-0 report section is [`dev-notes/mpcc-gate0-report.md` § P2](https://github.com/jkitchin/pounce/blob/main/dev-notes/mpcc-gate0-report.md) — everything below is drawn from measurements recorded there, and every claim marked *measured* was run rather than reasoned.

Nothing is blocked on this: the MPCC harness's recommended route (`scholtes_then_ncp`) clears all eight cells and returns MPCC-feasible points on them (`1.9e-12`, `4.9e-11`, `5.7e-28`). This is filed because it is a real defect with a reachable answer, not because Gate 0 waits on it.

## Symptom

On a **biactive** complementarity pair — `G_i` and `H_i` both zero at the solution — the exact-product lowering `G_i·H_i = 0` has row gradient `H_i ∇G_i + G_i ∇H_i`, and both terms vanish together there, so the row imposes no first-order restriction at all.

POUNCE reaches the exact solution primally and then the dual diverges. On `qpec_small` under `ncp_eq` from `origin`:

- `x = (1.0, 1.0, 4.8e-09)` — the exact solution
- objective `2.2e-16` against `f* = 0`
- constraint violation `3.5e-19`
- the product row's multiplier doubles every iteration, `0.31` → `164`, while `‖d‖` halves

It exits via the gh#274 near-feasible restoration re-entry gate:

```
[POUNCE] near-feasible restoration re-entry at theta 3.803e-19 but the point
fails the acceptable-level tolerances (nlp_err 4.049e-5); reporting
ErrorInStepComputation rather than Solved_To_Acceptable_Level (gh#274).
```

**The point is right; the multiplier estimate is wrong.** On the one cell that does converge, the biactive product row carries a multiplier of `1.5e+11` against a row gradient of `3.2e-05`.

## Reproducer

`benchmarks/mpcc/` in PR #844, or standalone: `qpec_small` lowered by `ncp_eq` (`G·H = 0` as an equality), started at `origin`, unit scaling, `tol=1e-8 max_iter=300 bound_relax_factor=0 honor_original_bounds=yes`, exact Hessians, default `recalc_y=no`.

Affected: `qpec_small` from `origin` and `upper_left` at unit scaling under the `ncp_eq` family; `direct` on `qpec_small/skew/upper_right` and on `ralph1/unit/origin`. **Both exact-product lowerings are affected at similar rates**, so it is not a property of one route.

## Seven of the eight are fixable; the eighth must not be "fixed"

Checking the KKT system directly at each solution splits the bucket. An earlier draft of the report lumped them together, which is the mistake its own branch-coverage rule exists to prevent:

| cells | best dual residual over **sign-feasible** multipliers | |
|---|---:|---|
| 7 — all `qpec_small` | **`0`**, at `λ = 0` exactly | the lowered NLP *has* a KKT point; POUNCE does not reach it |
| 1 — `ralph1/unit/origin/direct` | **`0.707`** | no sign-feasible multiplier exists; failing is **correct** |

`qpec_small`'s solution `(1, 1, 0)` has `∇f = 0` exactly and the biactive row's gradient is exactly `(0,0,0)`, so `λ = 0` satisfies stationarity with residual `0`. A row whose gradient vanishes has an *arbitrary* multiplier, not a nonexistent one — `λ·0 = 0` for every `λ`, including `0`.

`ralph1`'s origin is M-stationary but **not** S-stationary, and NLP KKT is S-stationarity. Best achievable sign-feasible residual is `0.707`, three orders above any tolerance in play. Note an *unsigned* least-squares fit reaches `4e-16` there — so sign feasibility has to be part of the test, not an afterthought.

### `recalc_y=yes` manufactures stationarity — do not take it

Re-estimating multipliers at the returned point flips every one of these cells to `Solve_Succeeded`. Three of four are then non-stationary in the model's own units by three to six orders, and pass anyway:

| cell (`recalc_y=yes`) | scaled `final_kkt_error` | **unscaled** `final_kkt_error` |
|---|---:|---:|
| `ralph1/unit/origin/direct` | `4.607e-10` | **`2.659e-02`** |
| `qpec_small/unit/origin` | `1.091e-09` | `5.066e-08` |
| `qpec_small/unit/upper_left` | `3.384e-11` | **`4.228e-03`** |
| `qpec_small/skew/upper_right` | `1.664e-09` | **`6.179e-05`** |

The re-estimated multipliers are enormous, `s_d` grows with them, and the strict gate reads the residual *divided by* `s_d` — the verdict is normalised by the very quantity that is wrong. Near a vanishing gradient this is structural: a hair off the solution the row gradient is `~1e-9`, and a multiplier of `~1e9` on it reproduces any vector you like. **An unbounded least-squares multiplier fit does not test stationarity near a biactive pair; it manufactures it.**

This is useful as a *test oracle*: `recalc_y=yes` is a one-option experiment that reproduces the wrong answer on demand, so a proposed fix can be checked against it before it is written.

### Relaxing the gh#274 gate treats the symptom

gh#274 requires the full acceptable-level triplet before a near-feasible restoration re-entry may claim acceptability, because an earlier looser version reported a diverging iterate as solved — `min -exp(x) s.t. x >= 0` re-enters restoration at `inf_pr = 1.7e-10`, `inf_du = 8.8e+47`, and Pyomo maps `Solved_To_Acceptable_Level` into the solved family.

**The defect is upstream of the gate**, in whatever makes POUNCE's dual residual grow to `164` over 118 iterations on a model Ipopt 3.14.19 finishes in 58.

## Comparator: neither implementation dominates

Same `.nl` file, same options, exact Hessians both sides, default `recalc_y=no`:

| `qpec_small`, `ncp_eq`, via `.nl` | start | iters | exit | **unscaled** overall |
|---|---|---:|---|---:|
| Ipopt 3.14.19 | `origin` | 58 | `Optimal Solution Found` | **`7.523e-09`** |
| POUNCE 0.11.0 | `origin` | 41 | `Solved To Acceptable Level` | **`7.897e+04`** |
| Ipopt 3.14.19 | `upper_left` | — | `Optimal Solution Found` | **`5.934e-02`** |
| POUNCE 0.11.0 | `upper_left` | 76 | `Optimal Solution Found` | **`1.564e-08`** |

On `origin` POUNCE is worse; on `upper_left` the roles reverse. So **"match Ipopt" is not the acceptance criterion.**

## Out of scope for this issue

The `7.897e+04` row above is **not an implementation defect**. The acceptable-level triplet reads the *unscaled* residuals (`current_is_acceptable_with_state` → `curr_unscaled_dual_infeasibility_max`), exactly as `acceptable_dual_inf_tol`'s help text specifies, and both halves pass on their own terms:

| test | quantity | threshold | |
|---|---:|---:|---|
| aggregate | **scaled** `8.234e-11` | `acceptable_tol` `1e-6` | passes — this one *is* `s_d`-normalised, `s_d ≈ 1e15` |
| dual component | **unscaled** `7.897e+04` | `acceptable_dual_inf_tol` `1e10` | passes on the threshold's size |

Both defaults are upstream Ipopt's and both are implemented as registered. Refusing that point needs a threshold moved, which is a **policy** choice with a blast radius of every acceptable-level exit and a deliberate deviation from upstream (precedent: gh#532's `dual_inf_scale_kappa`). Worth noting no SOTA code judges optimality by a bare absolute bound on an unscaled dual residual — Knitro's default test is relative with absolute floors, SNOPT's is `maxⱼ(Compⱼ/‖π‖) ≤ εd` — so `1e10` reads as a *disabled* cap rather than a loose one. Not decided here.

## Acceptance criteria for a fix

- **On `origin`, `qpec_small` under `ncp_eq` must not report success at an unscaled residual of `7.9e+04`**, and Ipopt's 58 iterations to `7.5e-09` show a point exists to converge to. Stated as "do not report success at a residual four orders above `tol` in the model's own units" rather than "match Ipopt", since `upper_left` runs the other way.
- A discriminator between "converged primal, unbounded multiplier" and "diverging iterate", checkable at the gate, **that a multiplier of `1e9` on a `1e-9` gradient cannot satisfy**. An unbounded least-squares fit is not one; nor is anything read off the `s_d`-normalised aggregate, since `s_d` is itself a function of the multipliers being estimated. The unscaled residual is the quantity that stays honest here, and it refuses three of the four cells outright.
- `ralph1/unit/origin/direct` must **still fail** — no sign-feasible multiplier exists there. A fix that greens all eight has over-fired.
- Evidence it does not relabel any `-exp(x)`-shaped case (gh#274's own reproducer).
- **`scripts/sweep-fixtures.sh` across both legs.** The verdict is reached on an `s_d`-normalised aggregate, so the blast radius is every model, not just those reaching the gh#274 gate. Per CLAUDE.md, a measured regression there needs its own issue and owner.

## Prerequisite worth doing first (cheap)

**The same model exits differently through two paths.** Via the harness's Python path it is `Error_In_Step_Computation` at 118 iterations; via `.nl`/CLI at the same options it is `Solved_To_Acceptable_Level` at 41. Both use exact Hessians (the harness reports 119 Lagrangian Hessian evaluations, so it is not an L-BFGS difference). The Hessian sparsity differs — the harness hands back a dense lower triangle, 6 nonzeros, where the `.nl` path detects 5 — and that is the leading hypothesis, **explicitly not established**.

Until it is, the exit named in the ownership bucket should be read as the Python path's. The count of 8 is unaffected, since it comes from the triage rules and not the exit string, but the bucket's *description* needs re-checking against a harness re-measurement. This is much cheaper than the core fix and should precede it.

## Adjacent gap (separate, but related shape)

`kkt_fidelity_tol` (gh#173) exists for precisely this shape — "the scaled convergence test passes but the user-space duals have drifted" — yet it only ever downgrades `Solve_Succeeded`. An exit already at `Solved_To_Acceptable_Level` is outside its reach, so the one guard built for this case cannot be aimed at it even by a caller who opts in. Closing that needs an honest status below acceptable-level for "feasible, but not to the fidelity you asked for"; repurposing `FeasiblePointFound` would break a public contract to fix something smaller.

### LINKED ISSUE #887: gh#884 retry costs deb7 a second solve under limited_memory_ls_failure_restarts=1 (6.1s -> 25.2s, same verdict)
The gh#884 dual-divergence retry scopes itself by base status: it runs only
when the base attempt returned `Solved_To_Acceptable_Level` or
`Restoration_Failed`, because those are the two verdicts a vanishing-gradient
row produces directly. Generic exhaustion exits
(`Error_In_Step_Computation`, `Maximum_Iterations_Exceeded`) are excluded
because the *detector* is right on them and the *remedy* is not — `deb7`
under L-BFGS being the corpus case that measured it.

That exclusion is only as complete as the status is stable, and on `deb7`
itself it is not.

## Measurement

    pounce crates/pounce-cli/tests/fixtures/deb7.nl --no-sol \
      hessian_approximation=limited-memory \
      limited_memory_ls_failure_restarts=1

| | before gh#884 (`d89771bc`) | after |
|---|---|---|
| status | `Restoration_Failed` | `Restoration_Failed` |
| objective | unchanged | unchanged |
| wall clock | 6.08 s | 25.17 s |

The retry is **declined** (unscaled KKT `1.4734e0` against the base
attempt's `9.9000e1` — better, but the base status is what conjunct 5
compares against and the retry's own verdict is not `Solve_Succeeded`), so
the reported answer is byte-identical. The whole cost is the second solve,
which runs to the caller's full `max_iter` rather than to the base
attempt's iteration count.

## Why this is not covered by the fixture sweep

`scripts/sweep-fixtures.sh` runs default options, where `deb7` exits
`Error_In_Step_Computation` and is out of scope. The sweep diff for the
gh#884 change is empty on both legs. `limited_memory_ls_failure_restarts`
defaults to `0`, so no default run reaches this.

## Options

1. Leave it. One extra solve on a run already reporting failure is the
   contract `run_with_mu_strategy_fallback` already has, and
   `dual_divergence_retry=no` switches it off.
2. Bound the retry's iteration budget by the base attempt's, rather than
   giving it the caller's full `max_iter`. This would cap the worst case at
   roughly 2x instead of 4x. It needs re-validating against `qpec_small`,
   whose promoted retry must still have room to converge.
3. Add a cheap pre-check that the remedy is plausible before spending a
   solve on it. No candidate discriminator has been measured.

Recorded here rather than absorbed into a commit message, per the
"a measured regression accepted as a cost of the fix needs an issue and an
owner" rule in `CLAUDE.md`.

Documented at the point of the decision in
`crates/pounce-algorithm/src/application.rs`
(`run_with_dual_divergence_retry`),
`dev-notes/mpcc-biactive-dual-divergence.md`, and `docs/src/options.md`.

jkitchin:
Fixes #884.
Closes #887.

This PR started as the negative result — four approaches ruled out by measurement, no source change, `Refs #884` and a deliberate refusal to close. The fifth approach it recommended has now been built, and the body below is rewritten around the fix. The original review history is preserved at the bottom; the ruled-out measurements are preserved in `dev-notes/mpcc-biactive-dual-divergence.md`, which is the part of this PR most worth reading if you ever have one of those four ideas again.

## The defect

Some models have a solution at which a constraint's gradient **vanishes**. The row is satisfied, the primal iterate is exact — but the multiplier that would certify it is *arbitrary rather than nonexistent*, and the barrier drives it to infinity.

The standard case is a complementarity constraint lowered as a product, `G(x)·H(x) = 0`, at a point where the pair is **biactive**: `G = 0` and `H = 0` together, so `H∇G + G∇H` vanishes identically. MacMPEC `qpec_small` does this at its solution `(1, 1, 0)`, and MPCC lowerings (`ncp_eq`, `prod_eq`) reach such points routinely.

Nothing caught it because the convergence gate reads an NLP error normalised by `s_d`, and `s_d` grows with the *mean multiplier magnitude* — so the runaway divides itself out of the number the gate tests. One line of the summary block, on `qpec_small` at `bound_relax_factor=0`:

```text
Overall NLP error.......:   8.2335532426389998e-11    7.8965510781517834e+04
```

Fifteen orders apart. The gate read the left column and reported `Solved_To_Acceptable_Level`.

## The fix: detect late, then re-solve cold

**Detector**, three conjuncts at *one and the same* iterate, sticky once true, eligible only when `m ≥ 1`:

* `inf_pr ≤ 1e-8`;
* a **scale-relative** step — `maxᵢ |dᵢ| / (1 + |xᵢ|)` over the `x` and `s` blocks — at or below `dual_divergence_retry_step_tol` (default `1e-5`);
* the **unscaled** `‖∇L‖∞` at or above `dual_divergence_retry_du_floor` (default `1e2`), and finite.

The middle conjunct is #884's criterion 2 answered as a measurement: this iterate has *stopped moving*, which is what separates it from one that is merely running away. `m ≥ 1` matters because on an unconstrained model `∇L ≡ ∇f` and the third conjunct would collapse into a looser copy of `dual_inf_tol` — the same failure that killed ruled-out 4, arriving from another direction.

**Promotion gate**, five conjuncts. The base attempt saw the signature; its status is `Solved_To_Acceptable_Level` or `Restoration_Failed`; the retry returns `Solve_Succeeded`; the retry's **unscaled** KKT error *and* unscaled constraint violation are within `acceptable_tol`; and its unscaled KKT error strictly beats the base attempt's. Otherwise all three sinks are floored the way the μ fallback floors them (#870) — solution payload, certificate, last trace row.

Conjunct 4 is the one that matters. The defect *was* a status contradicted by its own unscaled residual, so a gate reading the status alone reproduces the bug one attempt later.

On `qpec_small` the unscaled KKT error goes `7.8966e+04 → 9.9636e-08` for a primal residual that goes `1.1e-16 → 5.5e-12` and a point `3.7e-06` further from `(1, 1, 0)`. That trade is deliberate: a marginally looser primal that comes with a multiplier a reader can check. A test grading these two answers on the primal alone would prefer the broken one.

## Why the detector is the whole safety argument

The remedy is `perturb_always_cd=yes`, and ruled-out 3 measured what that does to `ralph1` — a model with **no** sign-feasible multiplier. It reaches `Solve_Succeeded` at `f = -2.71e-5`, *below* `f* = 0`, with unscaled KKT `5.25e-7`. That answer beats its base attempt on **every number the promotion gate reads**. If the detector fired there, the gate would promote a wrong answer.

So specificity is not a nicety, it is the barrier: `qpec_small` settles to `4.3e-8`, `ralph1` bottoms out at `7.2e-3`, and the `1e-5` default sits between them. `the_detector_must_not_fire_on_ralph1` is mutation-checked — widen the step conjunct to `1e-1` and that test alone goes red.

## The status scope, and the one line the sweep moved

Criterion 5 is `scripts/sweep-fixtures.sh` across both legs. Run against a `d89771bc` binary it moved exactly one line:

```
lbfgs  deb7  nlp  ErrorInStepComputation  it=715 -> it=3000
```

same status, same objective. The detector fires on `deb7`, and it is a **true positive**: at iteration 346, scale-relative step `6.5e-6`, `inf_pr` `3.0e-12`, unscaled `inf_du` `9.2e+05` — an *order above* the reproducer's `7.9e+04`. So no dual floor excludes it, and the step conjunct separates the two only by tightening the default onto one fixture and spending the margin that holds `ralph1` out. The detector describes that iterate correctly; the *remedy* is what does not transfer — the retry burned the full 3000-iteration budget to reach `Maximum_Iterations_Exceeded` at unscaled KKT `6.7e+01` against the base's `9.9e+01`, so conjunct 4 refused it exactly as designed, at a cost of 4× the trajectory for an answer that did not move.

The scope was therefore cut by **status**, not by threshold:

| status | in scope | why |
|---|---|---|
| `Solved_To_Acceptable_Level` | ✅ | #884 verbatim — the `.nl`/CLI path |
| `Restoration_Failed` | ✅ | the same defect one step earlier, where the TNLP fixture lands (unscaled KKT `3.3e+11`) |
| `Error_In_Step_Computation` | ❌ | generic exhaustion exit |
| `Maximum_Iterations_Exceeded` | ❌ | generic exhaustion exit |
| `Solve_Succeeded` | ❌ | already the best verdict, certificate already checked in the model's own units |

Retrying a generic exhaustion exit is not "repair the runaway", it is "try again harder" — which is what `mu_strategy_fallback` and the second-opinion ladder already are, and `deb7` is what it costs. Worst case is one extra solve under the caller's own `max_iter`, on a run that satisfied the three-way detector, reached a scoped failure verdict, **and** still reports the runaway (see the next section). **`deb7` is back to 715 and the sweep diff is empty on both legs, 80 fixtures / 184 lines.**

### Scoping by status is only as complete as the status is stable

And on `deb7` itself it is not, which the full test run found rather than the sweep. Under `limited_memory_ls_failure_restarts=1` — #818's re-anchor rung, off by default — `deb7` exits `Restoration_Failed` instead of `Error_In_Step_Computation`, so it *is* in scope and does pay the extra solve: **6.08 s → 25.17 s** for the same `Restoration_Failed` verdict, the same objective, retry declined. The sweep never sees it because the sweep runs default options.

Nothing is wrong with the answer, but it is the honest bound on the narrowing: the status scope buys that the retry only runs on a solve already reporting failure; it does **not** buy that the retry only runs where the remedy has something to work on. That is a second question, and it needs a second gate.

### So the second gate reads the answer, not the trajectory (#887)

The detector is a statement about an **iterate**. Nothing in it says the solve *ends* at that iterate, and `deb7` is exactly the case where it does not: the signature is real at iteration 560 (step `1.7e-7`, `inf_pr` `7.0e-13`, unscaled `inf_du` `6.6e5`) and the base attempt then works its way back down to an unscaled KKT error of **`9.9e1`** before giving up. That answer is not one `perturb_always_cd` can repair — so the old gate bought a full cold re-solve to decline an answer it was never going to promote.

What #884's defect looks like **in the answer** is a point converged *except* that one multiplier ran away: the primal is exact, complementarity is met, and the entire residual is dual infeasibility. So the retry also requires

```
max(unscaled viol, unscaled compl) <= 1e-6 * unscaled dual_inf
```

Every run in the corpus that reaches this line:

| run | unscaled dual | viol | compl | ratio |
|---|---|---|---|---|
| reproducer, `.nl` route | `7.90e4` | `1.1e-16` | `1.1e-9` | `1.5e-14` |
| reproducer, TNLP route | `3.25e11` | `2.5e-16` | `2.8e-3` | `8.7e-15` |
| `deb7` + L-BFGS + rung | `9.90e1` | `8.0e-13` | `4.65e0` | `4.7e-2` |

Twelve orders. `deb7`'s complementarity is five percent of its own KKT error — that answer is not a converged point with a runaway multiplier, it is an unconverged point. `1e-6` leaves eight orders of grace on the tightest keep and four on the reject.

**`deb7` is not a portable witness, and that took a second red CI to establish.**

`57a11e88` shipped this gate with the negative pinned on `deb7`, and it failed on Linux again. Reading the log rather than reaching for the threshold: `deb7` under the rung does not reach the same answer on the two platforms CI runs. Its **base attempt** — the one the gate reads — reports

| | objective | unscaled dual | viol | compl | ratio |
|---|---|---|---|---|---|
| macOS | `99.677` | `9.90e1` | `8.0e-13` | `4.65e0` | `4.7e-2` |
| Linux | `99.651` | `5.5743e3` | `5.6e-14` | `2.08e-5` | `3.7e-9` |

On Linux that answer genuinely **is** #884's shape — scaled overall error `5.28e-1` against unscaled `5.57e3`, the `s_d` normalisation hiding a runaway exactly as it did on `qpec_small`, complementarity eight orders under its own dual residual. The gate opened correctly; the retry then hit `max_iter` and the promotion gate declined it, also correctly. **The retry there is the designed cost, not the waste #887 filed**, and "deb7 declines" is false on Linux at any threshold.

The tempting reading of the red was that `1e-6` is too tight. It is not — the constant was right and the fixture was wrong, and the way to tell those apart was to read what the Linux base attempt actually reported.

So the rule is pinned where a rule belongs: the condition is extracted as `runaway_is_the_whole_residual(dual_inf, viol, compl)` with the contract in its doc comment, and four unit tests carry all four measured rows plus the threshold and the finiteness guards. Mutation-checked — widen the ratio and `an_unconverged_point_does_not_open_the_retry` and `the_threshold_is_where_the_constant_says_it_is` go red, while `what_we_cannot_measure_does_not_buy_a_retry` stays green because it pins a different conjunct. The CLI file keeps only `the_gate_that_reads_the_answer_does_not_cost_the_reproducer_its_retry`, and its doc comment records why `deb7` was removed so the next reader does not re-add it. File runtime 6.13 s → 1.89 s.

The general rule, which is `CLAUDE.md`'s branch rule wearing different clothes: **a fixture is only evidence about the answer it actually reaches**, and on a model this hard "the answer it reaches" can differ by platform. Before pinning a *negative* on a fixture, check that it reaches the same class everywhere it runs — a green local run says nothing about that.

**Two other gate shapes were measured and dropped, and both are recorded:**

- **An absolute floor** reusing the detector's own `dual_divergence_retry_du_floor` declines macOS `deb7` by a **one percent** margin (`9.9e1` against `1e2`). A coincidence, not a discriminator, and a threshold on a scale-dependent quantity — entry 3 of this repo's own `/sens-review` checklist.
- **A retention ratio** — the reported unscaled KKT against the runaway the *detector* had fired on — separated the corpus by five orders and **failed on CI** (`be25873c`, job [99991312576](https://github.com/jkitchin/pounce/actions/runs/33548339586)). It reads two numbers from a **trajectory**, and `deb7`'s detector reports `9.2e5` on one attempt and `8.7e2` on another *in the same run*.
- **An iteration cap** on the retry (`min(max_iter, base iterations)`), from an earlier round: it halved `deb7`'s cost and swept clean, but *broke the fix* — on the TNLP route the base attempt bails into restoration at 14 iterations while the retry legitimately needs 19, so the base count is a lower bound on the work rather than an estimate of it.

The property that makes the shipped gate right is not its margin, it is that **both numbers come from the same answer**. The ratio carries no units, needs no re-fitting when a fixture moves, and cannot depend on which attempt fired or on how a platform rounded. `DUAL_DIV_RETRY_DOMINANCE` is a constant, not an option: it does not express a tolerance a caller trades against, it expresses "this answer is #884's shape". The escape hatch for the whole remedy is `dual_divergence_retry=no`.

**Measured after:** macOS `deb7` under the rung is back to **6.10 s** from 25.17 s — #887's measured waste, closed, and stated as the measurement it is rather than as a claim about every platform. The reproducer still promotes, `7.8966e4 → 9.9636e-8`.

The status scope stays. The two gates fail in different directions — the status scope keeps the retry off runs that never reported failure, the dominance gate keeps it off runs whose failure has nothing to do with a multiplier.

**Re-verified on the shipped gate:** `cargo fmt --all -- --check` clean; clippy `correctness`+`suspicious` clean (only the two pre-existing `pounce-restoration` warnings); `check-docs-consistency.sh` OK, 48 pages; full `pounce-algorithm` + `pounce-cli` suites with no `FAILED`; `scripts/sweep-fixtures.sh` both legs, 184 lines, **diff empty** against a `d89771bc` binary.

While fixing this I also found a stale piece of the previous commit: the **registered help** for `dual_divergence_retry` still listed the pre-narrowing *four* base verdicts. It now names the two, says why the other two are excluded, and documents the dominance gate.

That run also moved one existing test, and the reason is worth naming. `issue_818_reanchor_before_restoration`'s `deb7` safety assertion asked whether the combined stdout+stderr *contained* the substring `Solve_Succeeded`. That blob is captured at `RUST_LOG=pounce::algorithm=debug`, because the re-anchor decision is only observable as a log line — and the debug narration names status codes. With the rung on, the second `dispatch_standard_solve` runs and *its* μ-strategy fallback logs `MaximumIterationsExceeded is not Solve_Succeeded`. No success was reported and none was manufactured; a string that names the status matched. It now reads the last `Status:` line, which is the verdict.

## Criterion 4, measured both ways

gh#274's `unbounded_exp.nl` is the closest thing in the corpus to a false positive, and it is close — it satisfies **two of the three conjuncts outright** (`inf_pr` at zero, unscaled dual `8.7e+20`).

| `dual_divergence_retry_step_tol` | signature | promoted | status | objective |
|---|---|---|---|---|
| `1e-5` (default) | no | no | `Error_In_Step_Computation` | `-8.688703979461661e+20` |
| `1e30` (conjunct disabled) | **yes** | no | `Error_In_Step_Computation` | `-8.688703979461661e+20` |

Both rows are asserted, because "the detector does not fire" is only evidence if the reason is the one claimed. The second row shows the step conjunct is doing the work alone; it also shows the *second* barrier, since that status is out of scope and no retry runs. Status and objective are bit-identical.

## Acceptance criteria

| # | criterion | where |
|---|---|---|
| 1 | must not report success at unscaled `7.9e+04` | `the_reproducer_converges_with_a_multiplier_a_reader_can_check`, `the_kill_switch_shows_what_the_scaled_aggregate_was_hiding` |
| 2 | a discriminator checkable **at the gate** | the scale-relative step, evaluated per-iterate — not the trajectory statistic the experiment used |
| 3 | `ralph1` must still fail | `the_detector_must_not_fire_on_ralph1` (mutation-checked) |
| 4 | no relabelling of `-exp(x)`-shaped cases | `a_diverging_iterate_is_not_the_signature`, table above |
| 5 | fixture sweep, both legs | empty diff; the one line that moved is `deb7`, explained above |

## Reporting

An earlier draft had the summary block print the retry's *verdict*, and it contradicted the JSON report beside it. Cause: `print_summary` runs inside `optimize_constrained`, once per attempt, while the promotion is decided after the last attempt returns — so `dual_divergence_retry_promoted` reads `false` there even on the run that promotes.

Split accordingly. The summary reports the **signature** only:

```text
Biactive dual divergence (gh#884)                    = detected
```

with no number (the flag is sticky across attempts, so a residual printed beside it would belong to the other attempt) and suppressed on `Solve_Succeeded` (passing *through* a biactive runaway and recovering is routine on an MPCC — the shipped fixture does it in 29 iterations at default options, and a warning over a correct answer is noise). The wrapper prints the verdict, after deciding it. JSON: `statistics.dual_divergence_signature` and `statistics.dual_divergence_retry_promoted`.

## Cost and ordering

The retry is the **outermost** wrapper and its base solve is `dispatch_standard_solve`, which already contains the μ-strategy fallback. `retry_worthy` excludes `Solve_Succeeded`, so wherever the μ flip promotes, this never runs — which at default options is this very fixture. `the_default_configuration_never_reaches_the_retry` is that as a number.

Two off switches: `dual_divergence_retry=no` (retry off, detector still reports) and `dual_divergence_retry_step_tol=0` (detector off, nothing downstream can fire). Both are options-wiring tested, because an option that is registered and never read is the failure that shipped #677.

## Tests

`crates/pounce-algorithm/tests/issue_884_biactive_dual_divergence.rs` — 8 tests. The original four invariants, plus four branch tests: the kill switch, the `0` step tolerance, the `ralph1` safety property, and a retry that runs and *loses* leaving status/point/objective/residuals exactly as the base left them.

`crates/pounce-cli/tests/issue_884_biactive_dual_divergence.rs` — 5 tests, new. This is the only place `7.8965510781517834e+04` appears against a file a user could hand the binary, and the only place the two columns that hid it are read off the same line of output a user reads. Its module doc maps each test to the criterion it owns.

The reproducer ships as `crates/pounce-cli/tests/fixtures/mpcc_qpec_small_biactive.nl` — also the sweep corpus's first MPCC lowering.

## Blast radius

Source, for the gh#884 commits on this branch: `pounce-algorithm/src/{application.rs, ipopt_alg.rs, alg_builder.rs, upstream_options.rs}`, `pounce-nlp/src/solve_statistics.rs`, `pounce-solve-report/src/{console.rs, lib.rs}`. The two new report fields are `serde(default)`, so an old report still deserializes. One existing test file changes, `pounce-cli/tests/issue_818_reanchor_before_restoration.rs`, for the reason above.

Sweep empty on both legs. `cargo test --workspace --no-fail-fast` green. `cargo fmt --all -- --check` clean; CI's clippy gate (`-D clippy::correctness -D clippy::suspicious`) clean; `scripts/check-docs-consistency.sh` OK. Current `main` (`e23fda3a`) is merged in — a Python-only merge, no Rust file touched.

## Review rounds (from this PR's earlier, no-source-change form)

**Round 1** (@bernalde, `bbdc3a0c`): overclaimed precision on the returned iterate plus a misread of my own trace table; and an overclaimed universal negative — that no solver-observable trigger could satisfy criteria 1 and 3 together. **That one was retracted, and this PR is why it was wrong.** The argument turned on a property of the *limit point* (S-stationarity) the solver cannot know; the fix turns on a property of the *iterate* (that it stopped moving) the solver observes every iteration. When an impossibility argument appeals to something unknowable, check whether the thing you actually need is observable instead.

**Round 2** (@jkitchin, `c83471d4`): an over-broad guard doc; a miscount of #884's acceptance criteria (five, not two); and the retry route's cost mispriced as wall clock when a false positive actually costs *correctness*. That third finding is the reason conjunct 4 exists.

---

- [x] Tests fail on the parent commit — the branch tests do not *compile* before the detector lands (they read two new `SolveStatistics` fields), so the meaningful equivalent is used instead, and both mutations were run rather than reasoned about: widening the step conjunct to `1e-1` turns `the_detector_must_not_fire_on_ralph1` and only that test red, and re-adding `Error_In_Step_Computation` / `Maximum_Iterations_Exceeded` to `retry_worthy` turns `a_generic_exhaustion_exit_does_not_buy_a_retry` red, and dropping the dominance conjunct turns `a_run_that_recovers_from_the_signature_does_not_buy_a_retry` red on its `deb7` half. The four original invariants stay green by design
- [x] Book page: `docs/src/options.md`, new section "A settled point with a runaway multiplier"; `dev-notes/mpcc-biactive-dual-divergence.md` rewritten to status **fixed**, keeping all four negative results
- [x] `cargo fmt --all -- --check`, clippy gate, `check-docs-consistency.sh` clean
- [x] `scripts/sweep-fixtures.sh` both legs against a `d89771bc` binary: **empty diff**. Note `d89771bc` predates PR #844, which merged this branch's earlier gh#794 `application.rs` work into `main`, so the empty diff covers that change too — strictly more than gh#884 needed
- [x] Every number in this body is measured on this branch. The one line the sweep moved is explained and removed, and the one cost that survived the narrowing (#887) is fixed here rather than deferred — `deb7` under the #818 rung pays nothing

https://claude.ai/code/session_01HVXWp9fLta3pQTYF2aaxEB


bernalde (changes requested):
Blocking: Correct the characterization's measured claims and record their provenance.


The [characterization test](https://github.com/jkitchin/pounce/blob/874022742034262fe5e110c555cac9e72d861055/crates/pounce-algorithm/tests/issue_884_biactive_dual_divergence.rs) and PR body call the returned `qpec_small` iterate the "exact primal solution" and say its objective is at `f*`. At this exact head, the default run returns `x = (1.0002321033, 1.0001160516, 2.67e-15)` and `f = 6.73399e-8`; the test correspondingly allows `1e-3` in `x` and `1e-6` in the objective. That is source-feasible and near the known optimum, not the exact point `(1,1,0)`. In particular, the `lambda = 0` certificate quoted for the exact point leaves an objective-gradient residual of about `4.64e-4` at the returned iterate.

The [trace note](https://github.com/jkitchin/pounce/blob/874022742034262fe5e110c555cac9e72d861055/dev-notes/mpcc-biactive-dual-divergence.md) also says `prim` and `compl` are `1e-11` from iteration 5, while its own iteration-5 row is `4.1e-6` and `8.0e-7`. Both are already below `kappa_eps_mu = 1` at iteration 3; they reach roughly `1e-11` by iteration 9. Please align the test name/comments, dev-note, and PR body with those measurements, and stamp the numerical sections with the exact measured source commit as the repository's `AGENTS.md` requires.

Blocking: Narrow the solver-observable-trigger conclusion to the experiments that were actually run.


The measurements rule out the Hessian-sparsity hypothesis, enabling regularization *within the same run* at the sampled iterations, and flipping the default globally. They do not rule out a late detector followed by a cold retry from iteration 0 with regularization enabled. The note itself records a solver-observable separation (`||d||` settling near `5e-8` for `qpec_small` versus remaining near `1e-2` for `ralph1`): a late detector could use that separation to retry the former from a recoverable state while leaving the latter as an honest failure. Whether that policy is robust across the corpus remains to be measured, but it means the universal claim that no solver-observable trigger can satisfy both criteria is not established. Either test this retry class, including the `ralph1` rejection, or narrow the durable conclusion to the in-flight detector and default flip that were measured.

These are review-body findings because both span the PR description, test commentary, and dev-note rather than one isolated line.

Summary:

- Blocking issues: 2
- Nonblocking issues: 0
- Questions: 0
- Exact-head local test: `cargo test -p pounce-algorithm --test issue_884_biactive_dual_divergence` — 4 passed, 0 failed.
- Independent exact-head instrumentation reproduced the reported statuses and numerical behavior for both fixtures and both regularization modes.
- PR-head CI: 8/8 checks succeeded; the [Ubuntu test job](https://github.com/jkitchin/pounce/actions/runs/33519258582/job/99894043421) ran and passed all four new tests.
- Coordination: the only other open PR, #149, has no file overlap with this change.

I would not merge this until the blocking issues above are addressed.

jkitchin:
@bernalde both blocking findings were right, and the second one changed the conclusion of the PR. Addressed in `bbdc3a0c`.

### Blocking 2 — the universal negative. Retracted, and you were right about the mechanism too.

You identified the exact gap: my three experiments covered the sparsity hypothesis, in-flight `delta_c` engagement, and a global default flip — **not** a late detector followed by a cold retry from iteration 0. "Too late to recover this iterate" is not "too late to act" if the action discards the iterate. I had that distinction wrong.

You also pointed at the separation I had already recorded without following it. Measured on `87402274`, minimum `‖d‖` over each run taken only while the primal is converged (`inf_pr ≤ 1e-8`):

| fixture | min `‖d‖` | at iteration | max `inf_du` | status |
|---|---:|---:|---:|---|
| `qpec_small`/`prod_eq`/origin | **`4.309e-8`** | 12 | `7.76e11` | `RestorationFailed` |
| `ralph1`/`direct`/origin | **`7.153e-3`** | 17 | `3.63e13` | `RestorationFailed` |

**Five orders apart.** Any threshold in `(4.3e-8, 7.2e-3)` separates them. Composed with the cold retry:

- `qpec_small` — detector fires; retry from iteration 0 with `delta_c` reaches `SolveSucceeded` at unscaled KKT `9.96e-8`;
- `ralph1` — detector never fires, so no retry, and the honest failure stands.

**Both of #884's acceptance criteria satisfied.** My claim rested on "the property separating them is S-stationarity of the limit, which a solver cannot know in advance." That was too strong: the solver doesn't need S-stationarity, only to observe that the iterate stopped moving. Retracted in the note, which now carries this as the **recommended route to try first** rather than a dead end — with what is still unmeasured stated plainly: generality across the corpus (two fixtures, and a threshold with no measured population behind it is exactly what CLAUDE.md warns gets retuned blindly later), the false-positive cost of a doubled solve, and the six other `qpec_small` cells.

I did not implement it here — that is a source change needing the fixture sweep across both legs, and this PR deliberately has none.

### Blocking 1 — overclaimed precision. Corrected on all three surfaces.

Right on every count. The run does not reach the exact solution: it stops at `(1.0002321, 1.0001161, 2.67e-15)` with `f = 6.73e-8`, and `lambda = 0` certifies `(1,1,0)`, **not** that iterate — where, as you computed, it leaves an objective-gradient residual of `4.64e-4`. The `1e-3`/`1e-6` tolerances were the tell and I wrote past them. The claim is now "the run gives up *near* the optimum", with the measured point quoted, and the tolerances carry a stated reason rather than papering over the distance #884 leaves on the table.

The trace-note error was mine against my own table: I wrote `prim`/`compl` "converged from iteration 5 — `1e-11`" where row 5 reads `4.1e-6` and `8.0e-7`. Corrected to what the table actually shows — neither ever binds, both are under `kappa_eps_mu = 1` by iteration 3, and they reach `~1e-11` by iteration 9. Only the `dual` column is ever the binding term, which is the point that section is making and is now made accurately.

All numerical sections are stamped `87402274` per `AGENTS.md`, and the corrections are propagated to the PR body, the test docs and the dev-note.

### One thing you did not ask for

Finding 1 exposed something worse than imprecise prose: **every test in the file asserted current behaviour, so a genuine fix to #884 would have turned them all red.** That is backwards, and it trains the next reader to treat red in this file as expected. All four are rewritten as forward-correct invariants — they now fail only when the solver gets *worse* — and renamed:

| was | now |
|---|---|
| `the_primal_converges_while_the_dual_diverges` | `qpec_small_returns_the_optimum_and_never_claims_a_false_success` |
| `the_hessian_sparsity_pattern_does_not_change_the_outcome` | `a_structurally_zero_hessian_entry_does_not_change_the_solve` |
| `dual_regularization_reaches_the_answer_honestly` | `dual_regularization_reaches_the_optimum_honestly` |
| `always_on_dual_regularization_reports_success_below_the_true_optimum` | `ralph1_must_not_claim_success_where_no_multiplier_certifies_it` |

The last is the substantive one. It asserted that a *non-default* option still produces a success below `f*` — pinning wrongness, and pinning a status the measurements show is a function of `max_iter` (`SolvedToAcceptableLevel` at 300, `Solve_Succeeded` at 3000), so it would have failed for reasons its name did not describe. It now runs at **default** options and asserts the contract: this model must not report success, and never below `f*`. Direction-correct, and it still catches the trap, because the only way to *ship* that remedy is to change the default — which the test would then be running under. Mutation-checked: flip that fixture to `perturb_always_cd=yes` and this test alone fails, on the objective bound, at `-2.71e-5`.

The organising asymmetry, which the old design had backwards: **`qpec_small` failing is the bug, so it is never asserted; `ralph1` failing is correct, so it is.** Guards on `qpec_small` are conditional — "if it claims success, the claim must hold in the model's own units" — vacuous today, load-bearing after a fix, and red exactly on the verdict-only fix #884 warns about.

856 `pounce-algorithm` tests pass; fmt, clippy and both consistency scripts clean. Still no source change.

---

jkitchin:
Working #884 independently in another session, I took the **other** family of fix — the policy/verdict route #884's "Out of scope" names and leaves undecided ("Refusing that point needs a threshold moved … precedent: gh#532's `dual_inf_scale_kappa` … `1e10` reads as a *disabled* cap rather than a loose one. Not decided here").

It fails, and how it fails is a fourth entry for the dev-note's ruled-out list — the first one in the policy family rather than the trigger family.

### The route

Give the acceptable-level gate the dual ceiling it does not have, referenced to `norm_inf(grad f)` rather than to `dual_scale`:

```
acceptable_dual_inf_bound = min( acceptable_dual_inf_tol ,
                                 max( dual_inf_tol , kappa * acceptable_tol * norm_inf(grad f) ) )
```

`grad f` rather than `dual_scale` because gh#532's argument — "a multiplier contributes to `dual_scale` whatever it contributes to `grad L`, so it cannot buy the test" — holds for one multiplier and fails for a **pair**. Two rows with parallel gradients admit multipliers of any size whose contributions cancel, which is exactly what a biactive pair produces: `dual_scale` runs away while `grad L` does not. `norm_inf(grad f)` is the one term of `grad L` no multiplier can move.

On the face of it this clears #884: it refuses the `qpec_small`/`ncp_eq`/origin certificate at the residual the issue names, leaves `ralph1`/`direct`/origin's honest failure byte-identical, converges nothing (so it cannot over-fire), and `scripts/sweep-fixtures.sh` diffs empty across all 182 fixture-legs, both legs.

### Why it is wrong anyway

`crates/pounce-rs/tests/watchdog_trial_is_not_a_divergence_verdict.rs`, `IllConditionedQuadratic`, n=12, cond 1e14, L-BFGS. Same run, ceiling on and off:

| | status | iters | objective | unscaled `norm_inf(grad L)` | `norm_inf(grad f)` | ratio |
|---|---|---:|---:|---:|---:|---:|
| ceiling on | `RestorationFailed` | 298 | `9.5916e8` | `8.4150e10` | `8.4150e10` | **1.0000** |
| ceiling off | `SolvedToAcceptableLevel` | 197 | `3.7375e-6` | `8.7467e1` | `8.7467e1` | **1.0000** |

The model is **unconstrained** — no rows, no bounds — so `grad L === grad f` identically and the ratio the ceiling tests is 1 by construction. The bound collapses to `max(1.0, 8.75e-5) = 1.0` and refuses a residual of `87.5`: `acceptable_dual_inf_tol` tightened from `1e10` to `dual_inf_tol` for the whole unconstrained and bound-only class, which has nothing to do with a biactive pair. The run then wanders 100 further iterations into restoration and ends 15 orders from `f*`.

The defect is in the discriminator, not the constant. I claimed `norm_inf(grad L)/norm_inf(grad f) ~ 1` is the signature of an *unresolved* multiplier. It is equally the signature of an ordinary not-yet-converged iterate with nothing to cancel. #884's criterion 2 asks for something "a multiplier of `1e9` on a `1e-9` gradient cannot satisfy" — `grad f` clears that direction and says nothing about what else it refuses.

### The part that generalises past this route

**An empty fixture sweep is not evidence about a change to the acceptable-level dual gate.** No fixture in the corpus has an acceptable-level exit whose unscaled dual residual exceeds `dual_inf_tol`, so the corpus is uniform in exactly the dimension such a change acts on and reports clean no matter what the change does. That is CLAUDE.md's own branch rule — "a corpus that is uniform in the dimension a change acts on reports 'small and mixed' no matter how large its models are" — instanced on the convergence gate rather than on iteration counts. The model that caught it is not in the corpus at all.

If the detector-plus-cold-retry route above is what gets implemented, this is the corpus gap that route's "generality across the corpus" caveat will also be measuring against.

### Two artefacts, if they are useful here

- a pyomo-written `.nl` for `qpec_small`/`ncp_eq`/origin (`nnz_jac_g = 10`, `nnz_h_lag = 5`), which reaches criterion 1's headline `.nl`-path exit rather than the Rust-TNLP trajectory this PR measures;
- independent support for **Ruled out 1**: a Rust TNLP built with the `.nl`'s exact sparsity *and* its row order still reaches `RestorationFailed`, so sparsity alone does not explain the path split.

Say the word and I will open either as a PR against this branch.

**Provenance.** Everything above measured on `d89771bc` plus an unpushed working-tree patch implementing the bound quoted. The table is a direct probe run in-session; it is not a committed test. The patch is not proposed for merge — it is the negative result that is worth recording.

jkitchin:
Recorded in `231a2311`. The failure mode is worth more than the route, and the corpus finding is worth more than either.

### Ruled out 4 — in the note, as the first entry in the *policy* family

Your route, the ceiling, the `grad f`-vs-`dual_scale` reasoning (gh#532's argument holding for one multiplier and failing for a pair is the crux and I've kept it in your terms), the `IllConditionedQuadratic` table, and the diagnosis that the defect is the discriminator rather than the constant.

The line I'd underline, because it generalises past this route and past the policy family: **a discriminator has to be checked in both directions — what it admits *and* what it turns away.** #884's criterion 2 only asks for the first ("something a `1e9` multiplier on a `1e-9` gradient cannot satisfy"), and `norm_inf(grad f)` satisfies it exactly while saying nothing about the second. That asymmetry in the criterion is itself a defect in the issue; I've noted it in the section rather than silently fixing #884's text.

**Provenance recorded as you gave it**: `d89771bc` plus an unpushed working-tree patch, a direct in-session probe rather than a committed test, therefore not reproducible from the repo as it stands, and the patch not proposed for merge. That last part matters — it's a negative result with no artefact behind it, and the note says so rather than implying someone can re-run it.

### The corpus gap — promoted to CLAUDE.md

This is the most durable thing in your comment and I've put it beside the σ-path and bound-relax cases in the sweep section, because it is the **first instance of that rule on a verdict rather than a trajectory**. Every prior statement of "a corpus uniform in the dimension a change acts on reports clean regardless" has been about iteration counts and step sequences; yours is about the convergence gate, where the failure is silent in a different way — the sweep is not merely uninformative, it is actively reassuring while ten orders of tolerance move for a whole problem class.

It also **bounds the route this PR recommends**, which I've written into the note rather than leaving implicit: the retry route's "generality across the corpus" caveat is measuring against the same corpus, so an empty sweep there is evidence about trajectories and not about verdicts. Any gate-touching component of a #884 fix needs a purpose-built fixture reaching an acceptable-level exit above `dual_inf_tol`, because nothing in `benchmarks/mpcc/` or the CLI corpus does.

### The artefacts — yes to one here, and I'd hold the other

**Please do open the sparsity+row-order TNLP against this branch.** It closes a real hole in Ruled out 1 that I left open: my `a_structurally_zero_hessian_entry_does_not_change_the_solve` varies *only* the sparsity pattern, so "you didn't vary row order" is a live objection to the refutation, and your fixture answers it. That strengthens a claim already in this PR, which is the right kind of thing to add to it.

**I'd hold the `qpec_small` `.nl` for the #884 follow-up**, not because it's less valuable — it's more — but because it reaches the `.nl`-path exit that this PR explicitly declines to pin ("refuting the sparsity hypothesis does not identify the real cause, and pinning an exit string I cannot explain would be pinning a symptom"). Landing the artefact here would invite pinning that exit before anyone can explain the path split. It belongs with whoever takes the re-measurement #884 still asks for. Happy to be overruled — it's your call and your artefact.

One process note: `231a2311` re-triggers CI and lands after @bernalde's round-1 re-review request, so that re-review will now be against this content rather than `bbdc3a0c`. Flagging it so the round isn't wasted reviewing a superseded head.

---

bernalde:
Follow-up decision on the remaining test-comment and test-name consistency is tracked in #891. No implementation is proposed pending @jkitchin's decision there.

Issue #760, opened 2026-08-23
Record the QP iteration-count cost of applying bound_relax_factor on the convex arm (4c02817d)


jkitchin:
## Summary

`4c02817d` ("Apply bound_relax_factor on the convex arm too", Refs #744 #745)
raised iteration counts across the Maros-Meszaros QP suite. This issue records
the measurement so the jump in `BENCHMARK_REPORT` is attributable and does not
get re-bisected later or mistaken for noise.

**This is not a request to revert.** The evidence below says the commit is
right and the cost is the honest price of correctness. What is missing is the
record, per CONTRIBUTING's rule that a measured cost of a fix needs an issue
and an owner.

## Measurement

`benchmarks/qp` (138 Maros-Meszaros QPs), before vs after, same machine:

- 138/138 solved both ways, **no status flips**
- total iterations **2633 -> 3148 (+515)**, 31 models worse / 11 better
- wall **354.7s -> 382.0s (+7.7%)**

Worst movers:

| model | iterations | delta t |
|---|---|---|
| QSCFXM3 | 38 -> 168 | +1.47s |
| QSCFXM2 | 35 -> 145 | +0.82s |
| QSCFXM1 | 30 -> 131 | +0.38s |
| Q25FV47 | 41 -> 142 | +9.20s |
| STADAT1 | 22 -> 34 | +4.08s |

Bisected to `4c02817d` over the 588-commit window `fa20c200..4bd7e59c`
(8 steps, both endpoints verified). `bound_relax_factor=0` on current `main`
restores the old counts exactly -- 38 / 30 / 41 / 22 -- which confirms
attribution.

## Why this is the right cost and not a defect

The convex arm used to read the `.nl` bounds verbatim while the NLP arm
relaxed them, so one binary on one file solved two different models depending
on `solver_selection`. `4c02817d` closed that. The cheap old iteration counts
were the cost of solving an easier, wrong model.

Comparing the two arms on current `main` settles it. The NLP arm, which has
always relaxed, is **slower than the convex arm now is**:

| model | convex (`auto`) | NLP (`solver_selection=nlp`) |
|---|---|---|
| QSCFXM3 | 168 | 282 |
| QSCFXM1 | 131 | 198 |
| Q25FV47 | 142 | 422 |

Objectives now agree between the arms to 12 significant digits (e.g. QSCFXM3
`3.0816354268674165e+07` vs `3.0816354268676057e+07`). Before the commit they
[...]

Issue #535, opened 2026-08-08
routing: auto sends the GEN family to the convex IPM (199 iters / 191 s, reduced accuracy) when the NLP path solves it in 19 iters / 1.0 s (#133 follow-up)


jkitchin:
`solver_selection=auto` routes every detected LP to the convex IPM
(`resolve_solver`, `crates/pounce-cli/src/dispatch.rs:357-361`). On the NETLIB
GEN family that costs **194× in wall clock and the certificate**, while the NLP
path — already in the binary, already the default for every other class —
solves the same model in under a second.

## The numbers

`main @ 880b360b`, single-threaded, `gen.nl` (n=2560, m=769 of which 768
equalities, 63085 Jacobian nonzeros, pure LP):

| | `auto` (default) → convex IPM | `solver_selection=nlp` |
|---|---|---|
| status | `Solved_To_Acceptable_Level` | **`Solve_Succeeded`** |
| iterations | 199 | **19** |
| wall clock | **190.8 s** | **0.982 s** |
| objective | `-1.5564769679046607e-08` | `-1.0975644150419833e-05` |
| dual infeasibility | `1.1552980794249379e-12` | `2.9826351872125325e-11` |
| constraint violation | `1.3740609716705876e-07` | `4.4536818677443080e-11` |
| overall NLP error | `1.3740609716705876e-07` | `4.2604658585711498e-09` |

Reference Ipopt 3.14.20/MA57: `Optimal`, obj `-1.0974845560297915e-05` — which
the NLP path matches to four figures and the convex path does not.

`gen1` is the same model under a second name and behaves identically.

## The convex path's failure is the primal residual, and it is at the cap

Dual infeasibility (`1.16e-12`) and complementarity (`2.33e-09`) are both fine.
The sole blocker is the primal residual at `1.374e-07` against `tol = 1e-8`.

`199` iterations is not a coincidence: `QpOptions::default()` has
`max_iter: 200` (`crates/pounce-convex/src/ipm.rs:250`). It is exhausting the
budget, not converging, and the reduced-accuracy exit
(`QpStatus::OptimalInaccurate`, surfaced as "Solved to acceptable level (reduced
accuracy)") is the fallback for a usable-but-uncertified iterate.

## This is not #528's noise floor — I checked

The obvious hypothesis was that this is the #528 pathology in `pounce-convex`
(the primal term of the KKT error being a bare absolute residual, which #531
[...]

PR #888, opened 2026-09-01
fix(convex): solve the model as declared — bound_relax_factor becomes opt-in

### LINKED ISSUE #744: pounce-convex: convex QP arm certifies a KKT point to 1e-10 at an objective up to 33% high (LISWET family, YAO, POWELL20)
`pounce-convex` returns a point it certifies as KKT-converged to ~1e-10 whose objective is
up to 33% above the true optimum, on convex QPs, reporting `Solve_Succeeded`. The NLP path
in the same binary solves the same `.nl` files to Ipopt's value.

For a convex QP a KKT point *is* a global minimum, so two converged points cannot differ in
objective. The residuals `pounce-convex` reports therefore cannot be residuals of the problem
it was handed.

## Reproduction

Same binary, same file, only `solver_selection` differs:

```
$ pounce LISWET1.nl --no-sol
Problem class: convex QP. Selected solver: convex QP interior-point (pounce-convex) [solver_selection=auto].
POUNCE (convex QP IPM, pounce-convex): Optimal Solution Found.  obj=36.12240209  iters=21
Objective...............:   3.6122402091994900e+01
Dual infeasibility......:   1.7462298274040222e-10
Constraint violation....:   2.1094237467877974e-15
Complementarity.........:   5.2460843273423057e-10
Overall NLP error.......:   5.2460843273423057e-10

$ pounce LISWET1.nl --no-sol solver_selection=nlp
Problem class: convex QP. Selected solver: NLP filter line-search interior-point (pounce-nlp) [solver_selection=nlp].
Objective...............:   2.7122076276932603e+01
Dual infeasibility......:   2.7104540833988722e-11
Constraint violation....:   2.0830328868339830e-16
Complementarity.........:   1.7810185285614945e-08
EXIT: Optimal Solution Found.
```

Both claim optimality. Both report every residual converged. The objectives differ by 33%.

## Scope

Full Maros-Meszaros sweep, 138 problems (`make -C benchmarks qp-convex`). Nine disagreements
>1e-5 vs the Ipopt-MA57 reference, **all** on the high side:

| problem | pounce-convex | pounce-nlp | ipopt | rel err | it (convex / nlp / ipopt) |
|---|---|---|---|---|---|
| `LISWET1` | 36.1224 | 27.12208 | 27.12208 | 2.5e-01 | 21 / 39 / 39 |
| `LISWET7` | 498.8409 | 391.1951 | 391.1951 | 2.2e-01 | 22 / 20 / 20 |
| `LISWET10` | 49.48578 | 39.305 | 39.305 | 2.1e-01 | 25 / 129 / 130 |
| `LISWET8` | 714.4701 | 650.9597 | 650.9597 | 8.9e-02 | 26 / 115 / 114 |
| `LISWET11` | 49.52396 | 46.55054 | 46.55055 | 6.0e-02 | 22 / 64 / 61 |
| `LISWET12` | 1736.927 | 1672.207 | 1672.207 | 3.7e-02 | 24 / 100 / 100 |
| `LISWET9` | 1963.251 | 1899.378 | 1899.378 | 3.3e-02 | 25 / 65 / 65 |
| `YAO` | 197.7043 | 196.1771 | 196.1771 | 7.7e-03 | 22 / 27 / 27 |
| `POWELL20` | 5.208958e+10 | 5.208854e+10 | 5.208854e+10 | 2.0e-05 | 15 / 2087 / 2087 |

The `pounce-nlp` arm has **zero** disagreements >1e-5 across all 138, and its iteration counts
track Ipopt's step for step (39/39, 20/20, 65/65, 100/100, 2087/2087). That rules out the `.nl`
translation and the reference values as the cause, and localizes this entirely in `pounce-convex`.

The LISWET family is convex least-squares with monotonicity constraints — a unique global optimum.

## Why this matters

`pounce-convex` is auto-selected for this problem class. A user running `pounce foo.nl` on a
convex QP gets this path with no flag, a `Solve_Succeeded` status, and clean residuals.

## Not a regression

The committed baseline `benchmarks/BENCHMARK_REPORT.json` already carries these values
(`LISWET1` pounce=36.1224 ipopt=27.12208, `Optimal/Optimal`). The 2026-08-22 sweep at
`main @ a798ae1b` reproduced that file byte-identically across all 1327 problems. This has
been shipping, not newly introduced.

## Suggested starting point

Not investigated beyond the output above. Given that the residuals are internally consistent
but the point is wrong, QP data formation (the Hessian and its factor of 1/2, or the constraint
set as extracted) looks more likely than the convergence test. Possibly related to the closed
#500 / #401 / #685 family, which are all `pounce-convex` model-extraction bugs.

Found during a full benchmark sweep; see also the sibling LP-arm issue and #414, a closed
`SolveSucceeded`-at-a-non-KKT-point report against the same solver.

jkitchin:
Fixes #744.

**#744 was never fixed, and its diagnosis was inverted** — the arm it blamed was the one giving the right answer. Every value in its table matches the published Maros–Mészáros optima (DOC 97/6) **in the `pounce-convex` column**, to seven significant figures:

| problem | DOC 97/6 truth | `pounce-convex` (as filed) | `pounce-nlp` / Ipopt (as filed) |
|---|---|---|---|
| `LISWET1` | **36.1224** | 36.1224 | 27.12208 |
| `LISWET7` | **498.8409** | 498.8409 | 391.1951 |
| `LISWET10` | **49.48578** | 49.48578 | 39.305 |
| `LISWET8` | **714.4701** | 714.4701 | 650.9597 |
| `LISWET11` | **49.52396** | 49.52396 | 46.55054 |
| `LISWET12` | **1736.927** | 1736.927 | 1672.207 |
| `LISWET9` | **1963.251** | 1963.251 | 1899.378 |
| `YAO` | **197.7043** | 197.7043 | 196.1771 |

The fix as merged (`4c02817d`) taught the convex arm to widen its bounds too, which did not close the gap — it moved the correct arm into the wrong column, 138/138 → **130/138**, losing exactly those eight. The certified-but-33%-wrong objective #744 reports became the answer from *both* arms rather than from neither. This PR restores it.

#760 recorded the iteration cost of that commit as "the honest price of correctness". It reverses too — 3164 → 2658 iterations — and is commented there rather than reopened, since the record is the point.

## What

The convex arm widened every inequality row and the variable box by `min(bound_relax_factor, constr_viol_tol)·|b|` before extracting the QP (#744/#745), so one binary would not solve two models depending on `solver_selection`. The goal was right and the direction was backwards: it converged the two arms on one arm's *internal perturbation* rather than on the model the caller declared.

This makes the widening **opt-in on the convex arm**, leaves the NLP arm untouched, and reports the gap wherever it still exists.

Scored against the published Maros–Mészáros optima (DOC 97/6), the convex arm goes from **130/138 to 138/138 correct**, with **0** newly wrong and **0** status changes.

## Why it is wrong, not merely expensive

A widening of `δ` moves the optimum by `δ` times the bound's multiplier, and nothing bounds that product. On `LISWET1` — 10 000 monotonicity rows active, multipliers summing to `1.6e9` — it buys `9.0` of objective:

| | objective |
|---|---|
| widened (previous default) | `27.1220506` |
| as declared (this PR) | **`36.1224021`** |
| HiGHS, independently | **`36.1224020850`** |
| DOC 97/6 ground truth | **`36.1224`** |

On a 46-variable netlib LP (`issue745_netlib_problem.nl`) the same `1e-8` moves the objective from `0` to `-1.6` — the whole answer. The error is **one-signed** (widening only enlarges the feasible set), so it is a systematic optimistic bias rather than noise, and it does **not** close as `tol` tightens, because it changes the model rather than the convergence slop.

**The repo already had the answer.** `benchmarks/qp_four_way.md`, generated at 0.9.0 *before* #744 and scored on DOC 97/6, put the unrelaxed convex arm at **137/138 correct, 0 solved-but-wrong** — and listed `LISWET1(re=2.5e-01)` under Ipopt-MA57's *wrong* objectives, `LISWET1(re=2.7e-01)` under Clarabel's. #744 moved this arm into that column.

Why nobody saw it: every reference in CI came from POUNCE or from `ipopt_ma57.json`, and Ipopt carries the same widening, so no check could see the error. `dev-notes/qp-bound-relax-iteration-cost.md` measured iterations and arm parity, never accuracy against an independent optimum — it calls `36.1224` "the exact optimum" one paragraph before replacing it. That note now carries the reversal.

## It also cost speed

#760's +515 iterations reverse:

| | correct vs DOC 97/6 | iterations | wall |
|---|---:|---:|---:|
| shipped main | 130/138 | 3164 | 363.3 s |
| this PR | **138/138** | **2658** | **330.0 s** |

`QSCFXM1` 131 → **30** iterations *and* four orders more accurate (`9.1e-09` → `1.8e-13`). Over the 91-instance netlib LP corpus against HiGHS: median objective error `1.2e-08 → 3.8e-11`, within `1e-8` **37 → 84**, iterations `3121 → 2373`.

## The NLP arm is unchanged, and must be

It is a feasible-iterate log-barrier that needs `x` strictly inside its bounds. Disabling the widening there is not a near miss but a failure — the sweep at `bound_relax_factor=0` turns `square_flowsheet_resto` into `InfeasibleProblemDetected` and takes `cresc4` from 105 iterations to 3000.

The convex arm needs none of it: its IPM is infeasible-start, strict interiority lives in the `(s, z)` pair `init_iterate`/`recenter_warm` place rather than the geometry of the declared box — which is why **fixed variables (`lb == ub`, zero width) have always shipped through un-widened**. The sweep confirms the blast radius: **0 status flips, 0 objective moves, identical iteration counts across all 94 NLP-arm fixture-legs.**

`pounce foo.nl bound_relax_factor=1e-8` still reproduces the NLP arm's model exactly, so Ipopt parity is one option away.

## The arms now differ — visibly, not silently

New `final_declared_constr_viol` reports how far outside the model **as declared** a returned point sits, beside `final_constr_viol` (unchanged: it measures the widened model, is what every acceptance gate reads, and is what #712's success-verdict invariant is asserted on — redefining it broke `issue_689` exactly as it should have).

It matters most on the NLP arm, which still widens and whose `final_constr_viol` is the *internal slack* measure: on `wood1p` that arm reports `1.71e-14` at a point `7.96e-09` outside the declared rows and `9.84e-09` outside the declared box. The box half needed a lift out of the compressed bound spaces, so `Nlp::declared_box_violation` owns it, mirroring the `honor_original_bounds` projection.

## The one status regression, and its fix

`scaled_feasible_a` lost its fast path — but the widening had been relaxing one of its rows by **2.65e5 in absolute terms** (its rows reach `|b| = 2.65e13`, and the row width is relative by design, #385). Converging in 69 iterations on that was a much easier problem, not a better solve.

Fixed by extending #535's LP→NLP fallback to convex QP: the convex driver declines (it already emits the #293 scaling warning on that model) and the general path certifies in ~20 iterations. `ConvexQcqp` is not added to that gate — the conic arm has its own separate fallback already.

**The rerouted solve runs on the declared model too.** An earlier revision of this PR did not do that, and claimed in three places that the reroute "relaxes nothing". It did relax: the fallback hands the model to the one arm that still widens, and on `scaled_feasible_a` the rerouted point sat **2679.85 outside the declared model** while reporting `Solve_Succeeded` — the exact failure this PR removes, reached through the mechanism added to repair its one regression. This branch's own new statistic is what caught it. Unless the caller names `bound_relax_factor`, the rerouted solve now runs at 0, so `pounce foo.nl` solves one model whichever arm answers; the test asserts `final_declared_constr_viol` rather than the objective (an objective check cannot tell "rerouted" from "relaxed") and is mutation-checked.

**`solution.engine` now records which arm produced the verdict**, and the sweep prefers it over the `Selected solver:` banner. The banner names the engine *routing* chose and is printed before a fallback can fire, so a reroute left no trace in the sweep diff — the blind spot CLAUDE.md names when it says a routing regression "used to leave no trace". It immediately surfaced that `airport` also reroutes (`cvx-qcqp → nlp`); verified against the baseline binary that it already did on `main`, so the instrument changed, not the routing.

**Net sweep against current main: 0 status flips, 52 objective moves (all toward the declared optimum), total iterations 5618 → 5409, and 4 engine-column moves that are the new instrument reporting pre-existing reroutes.**

## Guarding against recurrence

`declared_optimum_sentinel.rs` holds two optima from outside POUNCE's solver family (HiGHS `0.0`; DOC 97/6 `1.68826916e+07`) and is **mutation-checked** — restoring the widening as the default turns it red by `1e8` in relative terms, and reverting turns it green.

`benchmarks/qp/README.md` and the report generator now state plainly that `ipopt_ma57.json` is a **status and timing reference, not an objective oracle** — because that is the mistake #744 was filed from, and without fixing the reference the next person has every reason to refile it.

Five affected test files are **inverted, not deleted**; each still exercises the opt-in path. `issue_745`'s objective reference moves from Ipopt's `-1.5996991454` to the declared optimum `0`, which HiGHS returns from both its simplex and its IPM on that fixture's `.nl` reconstructed as an LP and validated against the `.nl` evaluator.

## Verification

- Maros–Mészáros QP: 138/138 vs DOC 97/6, 0 newly wrong, 0 status changes
- netlib 91 LPs vs HiGHS: 84/91 within 1e-8 (from 37/91)
- `scripts/sweep-fixtures.sh` vs shipped main: 0 status flips, 0 engine flips, both legs
- `cargo test --workspace --exclude pounce-hsl`: green
- `pytest python/tests` 1393 passed; `pytest pyomo-pounce/tests` 500 passed
- `cargo fmt --check` clean; clippy clean at CI's gate

## Known limitations, stated

- **`nonconvex_qp_eq` under `solver_selection=qp-active-set`** loses its negative-curvature escape and refuses (`InternalError`) rather than reaching `-1`. It does not certify the saddle, and default `auto` routing is unaffected. `negcurv.rs` documents this class as deliberately out of scope; the widening was masking it by accident. Worth a follow-up: for the second-order test only, tentatively drop zero-multiplier bounds and re-test on the enlarged null space.
- **`greenbea`** is `9.11e-04` from HiGHS at default `tol` — verified tolerance-limited (`6.79e-10` at `tol=1e-12`), pre-existing, unrelated to this change.
- **The evidence that `bound_relax_factor=0` is safe on the reroute is two fixtures, and that is the whole population I could find.** The reroute fires on `scaled_feasible_a` (QP) and `airport` (conic) — 2 of 79 CLI fixtures — and on **0 of 91 netlib LPs and 0 of 138 Maros–Mészáros QPs**, measured by checking `solution.engine` across both corpora. So a clean sweep here is not evidence about a wider population; there isn't one to be evidence about. (The one QP that answers from `nlp`, `VALUES`, classifies as *nonconvex* and routes there directly — it never reaches the convex arm.)

  The CHANGELOG's argument that the NLP arm *needs* the widening — `square_flowsheet_resto` → `InfeasibleProblemDetected`, `cresc4` 105 → 3000 iterations — deliberately does **not** transfer to this path: both of those are genuine NLPs that can never reach a reroute, whose population is convex or conic by construction.
- **`final_declared_constr_viol` is now reported on the NLP arm too** — in the console summary, the JSON report and the Python `info` dict — since that is the arm still applying the widening by default, and its `final_constr_viol` is the internal slack measure rather than a statement about the caller's model.
- **`BENCHMARK_REPORT.md` and `qp_four_way.md` are not regenerated.** A faithful four-way needs Clarabel (not installed here) and a faithful report reruns every suite from inputs not present on this machine. Both carry a staleness note with the measured deltas and the refresh command; a partial regeneration would drop arms and be worse than the note.


jkitchin (commented):
## Review

The core change is right and the evidence for it is unusually good: two independent optima (HiGHS, DOC 97/6) pinned in `declared_optimum_sentinel.rs`, mutation-checked, and the whole reversal is one small, well-scoped edit to `convex_bound_relax`. `final_declared_constr_viol` is a genuinely good addition — I built the branch and it behaves exactly as documented.

But **the story told about the one status regression is not true**, and I verified that empirically.

---

### 1. The reroute does not "relax nothing" — measured, it relaxes by 2.7e3 (blocking)

`declared_optimum_sentinel.rs` names the test `a_declining_convex_solve_is_rerouted_rather_than_relaxed` and its doc says the declared model is "solved, quickly, and by a route that relaxes nothing." CHANGELOG.md says the reroute "restores it *without* relaxing anybody's model." The PR body says "the general path certifies in ~20 iterations, nothing relaxed."

The NLP arm keeps `bound_relax_factor=1e-8` — that is this PR's own explicit design — and `relax_bounds` (`crates/pounce-nlp/src/orig_ipopt_nlp.rs:1009`) widens inequality rows by `min(relax, cap)·|b|`, the same relative form that gives `2.65e5` on this model's `|b| = 2.65e13` rows. So the reroute hands `scaled_feasible_a` to the arm that applies the very widening this PR removes.

Built at `7f0d9d7f` and run at defaults:

```
Selected solver: convex QP interior-point (pounce-convex)
pounce: note: … re-solved on the general NLP interior-point path
EXIT: Optimal Solution Found.       Constraint violation....: 7.45e-09

final_constr_viol           7.4505805969238281e-09
final_declared_constr_viol  2679.8471808433533     <- this PR's own new statistic
status                      SolveSucceeded
```

The returned point sits **2679.85 outside the model as declared** and is reported `Solve_Succeeded`. That is the exact failure mode this PR exists to fix, reached through the mechanism introduced to repair its one regression — and the PR's own instrument says so.

The test does not catch it because it only asserts `obj.abs() < 1e-6`, which the *widened convex* answer on shipped `main` also satisfied. It does not distinguish "rerouted" from "relaxed" at all.

This is not a regression against `main` (`main` widened the same model on the convex arm), but the three claims above are wrong and shouldn't merge as written. Either:

- correct the test name, its doc, the CHANGELOG bullet and the PR body; or
- make the fallback actually mean it — pass `bound_relax_factor=0` into the rerouted solve — and assert `final_declared_constr_viol` in that test rather than the objective.

Worth deciding deliberately, because the general shape is now: *any* declining convex QP silently gets the Ipopt model back. That may well be the right bargain, but it is the opposite of the headline and it is currently undocumented.

### 2. "0 engine flips" cannot see this PR's routing change

`scripts/sweep-fixtures.sh:157` scrapes the engine off the `Selected solver:` banner, which is printed at routing time, before the fallback. Verified above: `scaled_feasible_a` prints `pounce-convex` and answers from the NLP arm. So the engine column reads `cvx-qp` either way, and "0 engine flips" is not evidence that routing is unchanged — for a PR whose second commit changes routing for the whole `ConvexQp` class (37 of 79 fixtures).

CLAUDE.md makes that column the guard against exactly this class ("a routing regression used to leave no trace in the diff"). Pre-existing for the gh#535 LP fallback, but this PR widens the population and leans on the column as evidence.

Suggestion: record the arm that produced the verdict in the JSON report (the sweep already reads JSON) and have the sweep prefer it, so a fallback shows up as a diff line.

### 3. The new statistic is invisible on the arm where it matters most

The PR says `final_declared_constr_viol` "matters most on the NLP arm." On that arm it reaches **only** `--json-output`:

- **No console line.** The extra line lives in `print_convex_summary` (`crates/pounce-solve-report/src/console.rs:646`), convex-arm only. In the run above the CLI printed `Constraint violation....: 7.45e-09` and `EXIT: Optimal Solution Found.` with nothing about the 2679.
- **Not in the Python `info` dict.** `crates/pounce-py/src/problem.rs:1405` and `crates/pounce-py/src/nlp_batch.rs:246` hand-build it field by field; the new field is not added. Every `pounce-solver` / `pyomo-pounce` user is on the NLP arm with the widening on by default and cannot see it at all.

### 4. Misplaced doc comment

`crates/pounce-cli/tests/issue_760_convex_bound_relax_magnitude.rs:88` — `MIN_RELAX_RATIO`'s three-line doc now sits on the newly inserted `RELAX`, which picked up a wrong docstring and left `MIN_RELAX_RATIO` undocumented:

```rust
/// How much more expensive the relaxed box is here, at minimum. Measured 4.4x
/// (131 vs 30) on `fdea82b5`, on both sweep legs. The margin is deliberate:
/// the claim under test is "different magnitude class", not "4.4".
/// Opt in to the widening: this fixture's whole subject is what that
/// costs, and it is no longer what the default does.
const RELAX: &str = "bound_relax_factor=1e-8";

const MIN_RELAX_RATIO: f64 = 3.0;
```

### 5. User docs not updated for a user-visible default change

`docs/src/options.md:604` still says "Before the solve, POUNCE widens every variable and constraint bound by `bound_relax_factor` (default `1e-8`, capped by `constr_viol_tol`)" — now false on the convex arm, and this is the page a user goes to. Same for the options table at `docs/src/options.md:73`. `final_declared_constr_viol` is also missing from the stats list at `docs/src/python.md:98`. (`docs/src/crossover.md` is NLP-arm-only, so it stays correct.)

### 6. Minor

- **CHANGELOG internal inconsistency.** The "Added" entry describes the convex widening in the present tense as deliberate current behaviour ("The convex arm widens every inequality row … That is deliberate"), while the "Changed" entry reverses it in the same release. A reader of the first entry is misled.
- `declared_residuals_qp` / `declared_residuals_socp` compute the full `kkt_residuals` — including dual infeasibility and complementarity against a model whose multipliers do not belong to it — when only `primal_infeasibility` is consumed. Also no assertion that the re-extracted problem's dimensions match `sol`; structurally they should always match, so a `debug_assert` is cheap insurance.
- `BoundRelax::is_active` is a `pub` wrapper over the private `active()`; simpler to make `active` public.
- The new console line sits inside the block `print_convex_summary` documents as Ipopt-scrapable. Low risk (scrapers key on `Objective...` / `Number of Iterations`), but it is a non-Ipopt line inside a compatibility block.

---

**Verdict:** the reversal itself is the right call — the sentinel design and the DOC 97/6 / HiGHS grounding are exactly right, and leaving the NLP arm alone is correct and well argued. Item 1 needs resolving before merge (a factual claim headed for a shipped CHANGELOG, plus a test that does not test its own name). Item 2 undercuts the headline sweep evidence. Items 3–6 are follow-up-able.

jkitchin:
Thank you — item 1 was a real error, and the thing that caught it was this PR's own new statistic, which is the most persuasive argument I can imagine for adding it. All six addressed in 837e3c24.

## 1. The reroute now relaxes nothing, rather than claiming to

Reproduced exactly: `final_declared_constr_viol = 2679.8471808433533` on a solve reporting `Solve_Succeeded` at `Constraint violation 7.45e-09`. You are right that this is the exact failure mode the branch exists to remove, reached through the mechanism added to repair its one regression.

I took your second option rather than correcting the prose, because the first would have left the shape you flagged — *any declining convex QP silently gets the Ipopt model back* — as undocumented default behaviour, and that is the opposite of the branch's thesis. Unless the caller names `bound_relax_factor`, the rerouted solve now runs at 0. Unset-means-declared is the rule the convex arm already follows, so `pounce foo.nl` solves one model whichever arm answers, while `solver_selection=nlp` and an explicit factor still buy Ipopt's model — neither is a default. `scaled_feasible_a` certifies in 22 iterations at `final_declared_constr_viol = NaN`. The conic fallback takes the same path and is covered.

The test now asserts `final_declared_constr_viol`, not the objective — your point that an objective check cannot distinguish "rerouted" from "relaxed" is exactly right, since the widened convex answer on `main` also clears `1e-6`. It is mutation-checked: forcing the fallback to relax again fails it with the 2679 in the message.

## 2. The sweep can see routing now

`solution.engine` records the arm that produced the **verdict**, and `sweep-fixtures.sh` prefers it, falling back to the banner for older binaries.

It earned its keep on the first run: **`airport` also reroutes** (`cvx-qcqp → nlp`), which nothing had ever reported. I checked against a baseline binary built from current `main` — it already did, so the instrument changed and not the routing. That is four engine-column moves in the diff which are pre-existing reroutes becoming visible, and I have said so explicitly rather than letting them read as regressions.

## 3. Reported where it matters

Console line in `print_summary` (NLP arm) on the same "only when it differs" condition as the convex one, plus `final_declared_constr_viol` in both Python `info` dicts (`problem.rs` and `nlp_batch.rs`). You are right that every Python caller is on the arm that widens by default and could not see it at all.

## 4–6

`MIN_RELAX_RATIO`'s doc is back on `MIN_RELAX_RATIO`. `docs/src/options.md` no longer says the widening is universal and gains a section on the convex default and the arms' deliberate divergence; `docs/src/python.md` lists the new field. The CHANGELOG's Added entry no longer describes the convex widening in the present tense while the Changed entry reverses it. The declared-residual helpers now state that only `primal_infeasibility` is consumed — the dual and complementarity terms are measured against a model whose multipliers do not belong to it — and `debug_assert` the re-extracted shape. `BoundRelax::active` is public rather than wrapped.

On the console-line placement: it is after the last Ipopt row and conditional, and scrapers key on `Objective...` / `Number of Iterations`, so I left it in the block rather than restructuring — happy to move it if you would rather the compatibility block stay literally Ipopt-only.

## State

Sweep vs current `main`: **0 status flips**, 52 objective moves, iterations 5618 → 5409, 4 engine-column moves as above. netlib unchanged at 84/91 within `1e-8` of HiGHS (from 37/91). Maros–Mészáros unchanged at 138/138 vs DOC 97/6. Workspace tests, `cargo fmt --check` and clippy at CI's gate all clean.

Still not addressed, and flagged in the PR body rather than silently: the `nonconvex_qp_eq` negative-curvature escape on the by-name active-set engine, `greenbea`'s pre-existing tolerance-limited error, and the two benchmark artifacts I could not faithfully regenerate here (no Clarabel; the report needs every suite's inputs).

jkitchin:
All six addressed. I re-verified the two that mattered against a rebuilt binary rather than reading the diff alone.

## Item 1 — the reroute was relaxing the model it exists to rescue

Fixed. Running `scaled_feasible_a.nl` at defaults on `837e3c24`:

| | before | after |
|---|---|---|
| status | SolveSucceeded | SolveSucceeded |
| iterations | 199 cvx + 20 nlp | 22 |
| `final_declared_constr_viol` | **2679.847** | `null` (NaN) |
| `solution.engine` | — | `nlp` |

Two things I checked that the diff does not show on its face:

- The forcing sits **after** the `SocpIpm` / `run_convex_qp` if-else, so it covers the conic fallback as well as the QP one — `airport` gets it too, not just `scaled_feasible_a`.
- `let _ = app.options_mut().set_numeric_value(...)` discards a `Result<bool, _>` whose `Ok(false)` means "clobber refused, nothing was written" — which would silently leave the widening on, i.e. re-introduce exactly the defect being fixed, with no signal. It is safe **here**, but only because the guard is exactly complementary: `get_numeric_value` returns `set == true` iff `find_tag` hits the map (`options_list.rs:293`), and `will_allow_clobber` returns `true` iff the tag is absent from that same map (`options_list.rs:103`). Guard false ⟹ absent ⟹ clobber allowed ⟹ the write lands. That is provable rather than incidental, so no change requested — but it is worth a one-line comment at the call site, because the next reader has to redo this derivation to know the discarded `bool` is dead.

## Item 2 — the sweep's engine column was blind to the fallback

Fixed by recording the arm that produced the *verdict* in `solution.engine` and having `sweep-fixtures.sh` prefer it over the routing banner. I reproduced the count independently: across all 92 fixtures on the exact leg, exactly two disagree with the banner —

```
airport:           banner=cvx-qcqp  json=nlp
scaled_feasible_a: banner=cvx-qp    json=nlp
```

— which is the reported four engine-column moves across both legs, exactly. `airport` is the pre-existing conic reroute the banner scrape could never surface; it lands at 47952.7017 in 15 iterations, on the published optimum.

## Items 3–6

The sentinel now asserts `final_declared_constr_viol.is_nan()` rather than only `obj.abs() < 1e-6` — the measurement that actually separates "rerouted" from "relaxed", which the objective check could not — plus `the_reroute_still_honours_an_explicit_widening` for the escape hatch. `MIN_RELAX_RATIO`'s doc comment is back on `MIN_RELAX_RATIO`. NLP-arm console line, both Python `info` dicts, `options.md` / `python.md`, CHANGELOG. All 13 tests across the three files pass here.

## One residual — a note for the PR body, not a block

Forcing `bound_relax_factor = 0` on the reroute changes a path this PR's title does not cover: the pre-existing conic fallback. And this branch's own CHANGELOG argues the NLP arm *needs* the widening — "not a near miss but a failure — `square_flowsheet_resto` → InfeasibleProblemDetected, `cresc4` 105 → 3000 iterations."

That argument does not transfer, because both of those are genuine NLPs that can never reach the reroute; the population that can is convex or conic by construction. But the evidence that 0 is safe for that population is **two fixtures**, both of which happen to be fine. That is the shape CLAUDE.md warns about: a corpus uniform in the dimension the change acts on reports "no movement" no matter how large it is. Not a defect and not a reason to hold the PR — but state the bound in the PR body, so the next reader inherits "measured on two fixtures" instead of inferring a wider population from a clean sweep.

Approving.

jkitchin:
Both addressed in 06cf2115. Thank you for re-verifying against a rebuilt binary — the `set_numeric_value` derivation in particular is the kind of thing I would not have written down unprompted, and it is exactly the reasoning a future reader would otherwise have to reconstruct to know the discarded `bool` is dead.

**The call-site note** now carries your derivation verbatim in shape: guard false ⟹ tag absent from the user's map ⟹ `will_allow_clobber` true ⟹ the write lands, so `Ok(false)` is unreachable here rather than merely unobserved.

**On the bound**, I tried to widen the evidence rather than only state it, and the answer is that there is nothing to widen it with — which I think makes your point more sharply than the original phrasing did. Measuring `solution.engine` across both corpora:

| population | reaches the reroute |
|---|---|
| CLI fixtures | **2 of 79** — `scaled_feasible_a` (QP), `airport` (conic) |
| netlib LPs | **0 of 91** |
| Maros–Mészáros QPs | **0 of 138** |

So the evidence really is two fixtures, and a clean sweep is not evidence about a wider population because there isn't one to be evidence about. That is now in the PR body in those terms, together with the note that the CHANGELOG's "the NLP arm needs the widening" argument (`square_flowsheet_resto`, `cresc4`) deliberately does not transfer — both are genuine NLPs that can never reach a reroute, whose population is convex or conic by construction.

One thing worth recording because it briefly looked like a defect: `VALUES` reports `engine=nlp` with `final_declared_constr_viol = 6.18e-09`, i.e. a widening applied on a solve I first read as rerouted. It is not — it classifies as *nonconvex QP* and routes to the NLP arm directly, never touching the convex one, so the forcing correctly does not apply and the widening is that arm's own. My first pass conflated "answered by `nlp`" with "rerouted"; the `Problem class:` line separates them. Nice illustration that `solution.engine` answers "which arm answered" and not "how it got there" — if the latter ever matters, it needs its own field.

PR #889, opened 2026-09-02
Sensitivity parity: one shared core, three solver arms

jkitchin:
## Problem

POUNCE shipped the word "sensitivity" over **two independent implementations**
and a third solver arm that had none:

| arm | implementation | public methods |
|---|---|---|
| NLP filter-IPM | `crates/pounce-sensitivity/` (~9k LOC) | ~38 |
| convex LP/QP | `crates/pounce-convex/src/sensitivity.rs` (~860 LOC) | 15 |
| conic (SOCP/QCQP/exp/power/PSD) | — none — | — |

They shared no code. `pounce_linalg::symmetric_eigen` was the only common call.

Three consequences a user hit:

1. **A live silent-wrong-answer path.** `solve_socp_ipm` and `solve_qp_ipm`
   return the *same* `QpSolution` type, and the cone partition travels beside
   it as a separate argument `QpSensitivity::build` never saw. A Rust caller
   could hand it a solved SOCP and every cone row was read as an orthant row —
   a plausible number that is not a derivative, with nothing warning.
2. **An LP with a sensitivity request paid the general engine's cost.** `auto`
   *declined* the convex fast path whenever a `.nl` carried the sIPOPT
   suffixes (gh#196's fix), because only the general engine could answer.
3. **The docs claimed parity that did not exist** — `choosing-a-solver.md` said
   "`QpSensitivity` **does the same** for the convex QP" (15 methods vs ~38),
   and `sensitivity.md` was NLP-only across 1779 lines with *zero* mentions of
   the convex arm.

## Cause

Not a missing port. The NLP crate's machinery was **already generic over a
trait**: `SensBacksolver` has two required methods (`dim()`, `solve()`), and
`boundcheck`'s fix-relax / path / directional routines are generic over it,
derive `n_x` from slice lengths, and read bounds only through `BoundRow`.

So the convex arm never needed 8k lines ported. It needed to become an
*implementor*. That is why this is a few hundred lines of adapter against
several thousand lines of test.

## Fix

Nine phases, each independently shippable and each leaving the tree green.

**`pounce-sens-core`** (new, 21st workspace crate) holds the engine-agnostic
half: the trait, `boundcheck`, the `P = K⁻¹A` / Schur stack, and the activity
**rule**. Both arms depend on it; neither depends on the other. The extraction
was a pure `git mv` — 6 of 8 files with zero line changes — with the exit
criterion "`cargo test` passes with zero test edits". Round 5 re-verified that
independently against `e4ad62b`.

[...]

PR #890, opened 2026-09-02
fix(880): the σ cascade labelled an uncertified pick `Optimal`

### LINKED ISSUE #880: Convex σ path: the componentwise guards use a directional scale, so they are inert on coupled (non-separable) ill-conditioned QPs
Found while fixing #875, by the census written to validate that fix. Filed
separately because it is a **larger and different** defect, not a residue of
#875, and because #875's fix must not be read as covering it.

## The gap

Both componentwise guards on the `σ` path — `sigma_complementarity_is_genuine`
(#846) and `sigma_stationarity_is_genuine` (#875) — hold a row's residual
against **the largest single term that built that row**. That is a
*directional* scale, not a *reduced* one. It is the same distinction CLAUDE.md
draws for the sensitivity classifier (`reduced/diagonal`, gh#763) and for
constraint rows (`reduced/directional`, gh#804), one crate over.

On a **diagonal** `P`, row `i`'s stationarity residual is `eᵢ(xᵢ − tᵢ)` and its
scale is of order `eᵢ|tᵢ|`, so the ratio *is* the relative error in `xᵢ` and the
test is exact. Rotate the same spectrum and the stiff mode appears in **every**
row: every denominator becomes the stiff mode's contribution, all `n` of them
collapse back to one number, and the componentwise refinement buys nothing. The
guard is present, is evaluated, and rejects nothing.

## Measurement

72 unconstrained instances: condition number `1e2 ‥ 1e12` × objective magnitude
`1e-3 ‥ 1e3` × `n` ∈ {2, 5} × rotated or not. `P = Q diag(e) Qᵀ`, `c = −P t`, so
`x* = t` by construction. Claimed-optimal-but-wrong (status `Optimal` and
relative `|x − x*|∞ > 1e-6`):

| | before #875 | after #875 |
|---|---|---|
| separable | 15 wrong | **0 wrong** |
| coupled (rotated) | 17 wrong | **17 wrong** |
| total | 32 / 72 | 17 / 72 |

Every coupled instance is **bit-identical** to the baseline after #875 — same
`x`, same iteration count, same reported dual infeasibility. The arm never
fires on any of them.

Worst coupled rows, all reporting `EXIT: Optimal Solution Found.` in **one**
iteration:

```
  k    cond    mag  n |  rel |x-x*|inf  it |  qp_hsde=no  it | ref acc
 65   1e+12  1e+00  2 |      7.051e-01   1 |   1.531e-03  62 | 2.2e-04
 63   1e+12  1e-03  5 |      6.464e-01   1 |   3.112e-02  41 | 2.2e-04
 69   1e+12  1e+03  2 |      6.339e-01   1 |   1.392e+00   1 | 2.2e-04
 71   1e+12  1e+03  5 |      4.041e-01   1 |   1.041e+00   1 | 2.2e-04
 57   1e+10  1e+03  2 |      3.208e-01   1 |   7.955e-06  10 | 2.2e-06
 53   1e+10  1e+00  2 |      2.642e-01   1 |   4.790e-06   3 | 2.2e-06
 55   1e+10  1e+00  5 |      1.169e-01   1 |   3.319e-05   3 | 2.2e-06
```

`ref acc` is the accuracy of the reference itself (`ε·cond·‖t‖`, from forming
`P` in floating point), so every error listed is three or more orders above the
noise floor of the measurement.

## Why it is not just "#875 not strong enough"

For the `cond = 1e10` rows above, `qp_hsde=no` — the destination a `σ` reject
routes to — is correct to `1e-5`/`1e-6`, so a guard that fired would fix them.
But for `cond = 1e12` the un-normalized path is **itself wrong** (`1.39` and
`1.04` relative on rows 69 and 71, and `1.5e-3`/`3.1e-2` after 41–63 iterations
on 63 and 65). Rejecting harder cannot fix those: there is nowhere correct to
route to. So this splits into two sub-problems and only the first is a guard
question:

1. **`cond ≤ 1e10`, coupled.** The `σ` certificate is accepted and should not
   be. Needs a denominator that survives a change of basis — bounding
   `‖x − x*‖ = ‖P⁻¹r‖` rather than comparing `rᵢ` to a term of row `i`. The
   factorization to do that is already in hand at the final iterate; the cost
   and the right formulation are the open question.
2. **`cond = 1e12`, coupled.** Neither path is right. Separate from the guard
   entirely.

## Reproduction

Census generator and runner: `adversary/audit-2026-08-31/` (see the #875 work);
they are ~60 lines of numpy + pyomo and regenerate the table above from scratch.

## What is already covered

`crates/pounce-convex/tests/issue875_unconstrained_sigma_stationarity.rs` closes
the separable half and its module doc names this coupled half explicitly as
something a green run of that file is **not** evidence about. Do not delete that
paragraph when this issue is fixed; replace it with the coupled coverage.

jkitchin:
Fixes #880.

## The defect

When `hsde_cost_scale` rescales the objective, `solve_qp_core` tries three
drivers and asks `normalized_optimum_is_genuine` of each. If none passes it
returns whichever claimed optimum is closest to optimality **in the caller's
own coordinates**.

Keeping that point is right, and by a wide margin: on a 72-instance coupled
census (rotated spectra, exact `Fraction` optimum per instance) it beats the
un-normalized re-solve on seven of nine failures, by 6× to 15416×. **Rejecting
harder routes to a worse answer.**

What was wrong was the label. The point went back under a bare `Optimal`, and
on three instances that `Optimal` sat on an `x` **further from the optimum than
f64 permits at that conditioning** — `3.06e-04`, `2.90e-04`, `2.45e-04` against
a `κ·ε` floor of `2.20e-04`.

## The fix

Carry the verdict out to `solve_qp_ipm` and apply it there, so an uncertifiable
pick returns `OptimalInaccurate`. The CLI's existing gh#535/#888 reroute reads
that and hands the model to the NLP arm.

| | base | this PR |
|---|---|---|
| instances above their `κ·ε` floor | 3 | **0** |
| worst forward error | `3.06e-04` | **`5.62e-05`** |
| census total iterations | 443 | **404** |

The nine instances that still read above a flat `1e-6` are all at or below
`κ·ε` — the wall `4005ca8d` recorded when it stopped counting this census above
`1e10`. They are not wrong answers and this PR does not claim to fix them.

## What is recorded: the scale-free half, with a margin

"The cascade certified nothing" mixes two different questions — whether the
residual is small in *absolute* terms, which legitimately depends on the
objective's scale, and whether `x` is far from the optimum, which does not.
Recording the mixture makes the status flap under an objective rescaling that
leaves the argmin **unchanged by identity**:

```
k=1e-2 Inaccurate │ 1e-1..1e0 Optimal │ 1e1..1e3 Inaccurate
k=1e4..1e5 Optimal │ 1e6..1e7 Inaccurate │ 1e8 Optimal
```

So only the forward-error arm is recorded — unconstrained,
`Δ = −P⁻¹(Px + c)` is exactly invariant under `(P, c) → k(P, c)` — and only
with a 10× margin. At `cond = 1e10` the realized error is `1.47e-06` against a
`1e-06` cut: the model sits *on* the decision boundary and rounding decides it.
The margin is **the estimator's own uncertainty**, not a knob: a plain
summation of this residual reads `0.29×`–`5.4×` of the truth against a
`Fraction` reference, so a verdict inside one order of the cut is inside the
instrument's noise. With it, the inert rescaling is uniformly `Optimal` across
all eleven decades.

## The layer is load-bearing

`OptimalInaccurate` is one of the statuses `solve_qp_ipm_core` reads as the
badly-scaled pathology worth a Ruiz-equilibrated retry, and **that retry is
accepted whenever it converges to a clean `Optimal`** — its status outranking
the better answer. Demoting where the pick is made takes the census's worst
forward error from `3.06e-04` to `3.93e+01`. At the outermost layer every retry
has already run.

The verdict rides in a new `sigma_verdict` thread-local modelled on the
existing `deadline` one (`QpSolution` has 145 construction sites and no
`Default`), re-recorded on every attempt so a retry cannot inherit a superseded
verdict.

## Downstream: `OptimalInaccurate` is admitted wherever `Optimal` was required

The demotion changes a *label*, not what the library will do.
`QpSensitivity::build`, `pounce-py`'s payload gate, `QpResult.success` and
`minimize`'s `success` all admit it now. That is the repo's settled convention
rather than a new judgement — `_curve_fit.py` (gh#119/#123) records that
treating `Solved_To_Acceptable_Level` as a non-success "reported
`success=False` at a verified optimum … and callers gating on `.success`
discarded valid fits". Without this the demoted population, which previously
arrived as a clean `Optimal`, would silently lose derivatives and `success`.
Verified end to end against a freshly built extension: it reaches Python as
`optimal_inaccurate` with `success=True`.

## Tests: exact, not loosened

An earlier revision of this PR loosened six existing status assertions to
accept either status. That was wrong — it destroyed the signal those tests
existed to give, and it hid the `k`-sweep flapping above. **Reverted.**

Every sweep test now asserts the *exact* status expected at each conditioning
and tolerance, so the certification boundary is pinned in both directions:

- `tol` sweep at `cond 1e10`: `Optimal` for `1e-4 ‥ 1e-8`, `OptimalInaccurate`
  at `1e-10` (cut `1e-08`, two orders below `κ·ε ≈ 2.2e-06` — unachievable).
- bound- and inequality-constrained `cond` sweeps: `Optimal` at `1e6`/`1e8`,
  `OptimalInaccurate` at `1e10` (measured forward error `1.6e-04` / `1.0e-04`,
  160× and 104× the cut — decisive, not boundary noise). Both are still inside
  `destination_bar`: "could not certify to `tol`" and "the answer is bad" are
  different claims and only the first is made.
- the `1e12` carve-out: `OptimalInaccurate` (`3.06e-04`, 300× the cut).

That is strictly more information than the original assertions carried, since
they only ever expected `Optimal`.

## Evidence

- `scripts/sweep-fixtures.sh`: **empty diff**, both legs, all 79 fixtures —
  status, objective, iteration count and engine.
- Workspace suite green: 347 suites. `pounce-hsl` excluded — it does not link
  in this environment, on this branch or on `main` (verified against a stashed
  tree).
- Eight-row mutation table, each row verified to turn the named test red.
- Local Python suite against a freshly built extension: 1405 passed, 4 skipped.
- The two fixtures CLAUDE.md names for this path are run **and
  branch-verified**: `issue414_cost_normalized_false_optimal` reaches
  `accept:plain` 22×, `illconditioned_huge_scale` `accept:plain` 4× and
  `accept:direct` 1× — so all 27 accept decisions read the changed estimator
  and still accept.

## Not included

While tracing the `k`-sweep I found that `sigma_complementarity_is_genuine`'s
slack scale, `bound.abs().max(x.abs())`, **collapses to zero** when a variable
rests on a bound of `0` — so `|slack| <= cut·0` is false for any nonzero slack,
however tiny, and a converged `1e-16` is rejected. It is a real latent defect,
but with the verdict now driven by the forward-error arm it changes nothing
observable: identical census errors and iterations, empty sweep, and no test
fails without it. It is therefore **not** in this PR rather than shipping
untested.


jkitchin:
## Review — read at `5b41a57d`

**Method note up front, because it bounds everything below.** This is a
close read of the diff and the PR body. I did **not** build or run anything
on this branch — a remote session is still updating the PR, so measuring
against a moving head would produce numbers about a commit nobody will
merge. Every claim below is either a citation or a question. Nothing is a
measurement.

### What the PR gets right, and it is the hard part

The **layering argument is load-bearing and is backed by a number**, which
is the thing I most expected to find hand-waved. `OptimalInaccurate` is one
of the statuses `solve_qp_ipm_core` reads as the badly-scaled pathology
worth a Ruiz retry, and that retry is accepted whenever it converges to a
clean `Optimal` — so demoting at the point of the pick lets the status
outrank the better answer, and the census's worst forward error goes
`3.06e-04` → `3.93e+01`. That is the correct diagnosis of a genuinely
non-obvious ordering hazard, and it is in the mutation table rather than
only in the prose.

`the_demotion_population_is_exactly_the_uncertifiable_instances` is the
right shape: four arms covering both directions (demotes / keeps `Optimal`
when certified / keeps `Optimal` when the cascade is never reached), plus
"the demotion relabels and must not move `x`". A build that demoted
unconditionally fails it. That directly answers CLAUDE.md's *"which branch
does the fixture take"* rule.

And the PR is candid where it would have been easy not to be: the empty
`sweep-fixtures.sh` diff is presented as *consistent with* 1-of-79 reaching
`σ`, not as evidence; the census is stated **not** to discriminate the
compensation, which is then pinned on constructed rows instead; the six
instances still above `1e-6` are named as at the conditioning limit rather
than claimed fixed. That is the standard this repo's notes ask for.

---

### 1. The demotion flips downstream gates the PR body does not name — and one of them refuses to build a sensitivity object

The PR frames the consequence as "the CLI's existing gh#535/#888 reroute
reads that and hands the model to the NLP arm." But `Optimal` is a
load-bearing predicate in at least four other places, and each takes a
different branch for the newly-demoted population:

- `crates/pounce-convex/src/sensitivity.rs:1502` and `:1781` —
  `if sol.status != QpStatus::Optimal { return Err(SensError::NotOptimal); }`
  in both `QpSensitivity::build` and `build_conic`. A σ-uncertified QP that
  previously built a sensitivity object now **errors**.
- `crates/pounce-py/src/qp.rs:830` —
  `let payload = (sol.status == QpStatus::Optimal).then(|| … QpSensitivity::build(…))`,
  and the comment right above says the caller raises when the payload is
  `None`. Same population, Python surface, raised exception.
- `python/pounce/qp.py:194` — `return self.status == "optimal"`, i.e.
  `result.success` becomes `False`.
- `python/pounce/_minimize.py:914` and `:973` —
  `success = res.status == "optimal"`.

Is that intended? There is a defensible reading where it is exactly right:
if the cascade could not certify the point, a *derivative* at that point is
no more trustworthy than the point, and refusing is honest. But it is a
behaviour change on a public API for a population that previously worked,
it is not in the PR body, and nothing in the diff pins it in either
direction. If it is intended, it wants a sentence in the body and a test;
if it is not, `SensError::NotOptimal` needs to admit `OptimalInaccurate`.

### 2. `maybe_crossover` sits between the verdict and its application

`solve_qp_ipm` (`ipm.rs:376–395`) is, in order:

1. `tracking(|| solve_qp_ipm_core(…))` — the verdict is recorded about the
   **interior** answer;
2. `maybe_crossover(…)` — for a pure LP, purifies that iterate to an exact
   optimal vertex via the active-set engine;
3. `demote_uncertified_sigma_optimum(sol, sigma_uncertified)` — applies the
   step-1 verdict to the step-2 answer.

So the verdict can describe a point that is no longer the point being
returned. That is precisely the hazard `sigma_verdict`'s own module doc
promises to exclude — *"the slot always describes the attempt whose answer
is actually being returned"* — but crossover is a fourth answer-producer
that cannot re-record, because it does not re-enter the solver.

`an_lp_is_left_to_the_existing_arms` is exactly this case, and the PR body
says so: the vertex is exact ("the vertex assertion is unchanged") and the
gate rejects it anyway on data magnitude. Combined with finding 1 and the
reroute, an **exact vertex** now returns `OptimalInaccurate`, sets
`success = False`, and gets handed to the NLP arm for a second solve.

The counter-argument is real and may well settle it: the gate measures
*absolute* KKT error against `‖c‖ ~ 1e10`, which crossover does not change,
so re-recording after crossover would reject too. If that is the reasoning,
say it — the mutation table already covers "move the demotion into
`solve_qp_core`" but says nothing about the choice of *this* slot versus
above `maybe_crossover`, and that choice is now load-bearing. I did not
measure whether crossover actually fires on that fixture.

### 3. The LP is the one demoting instance that is no longer pinned

Nine `assert_eq!(sol.status, QpStatus::Optimal)` became
`assert_returned_an_optimum`, which accepts either status — deliberately,
and the doc comment explains why. `the_demotion_population_is_exactly_the_uncertifiable_instances`
then pins the population "exactly, in one place". But the LP is **not** in
that pin, and it is the member with the most surprising consequence
(finding 2). Today nothing fails if `an_lp_is_left_to_the_existing_arms`
flips in either direction. One more arm in the pin closes it.

### 4. `record`'s argument does not mean what its parameter is named

`ipm.rs:3404`:

```rust
crate::sigma_verdict::record(candidates[pick].status == QpStatus::Optimal);
```

Reaching that branch already means "the cascade certified nothing", so the
conjunct is "…**and** the pick is currently labelled `Optimal`". The slot
therefore reads `false` for an uncertified pick that already carries a
non-`Optimal` status. Harmless today, because
`demote_uncertified_sigma_optimum` early-returns on exactly that condition —
so the two checks are redundant on purpose. But the bit is now a named,
crate-internal fact called `uncertified`, and the next reader who consumes
it for anything other than this demotion (a diagnostic, a log line, a
sensitivity gate per finding 1) will under-report. A comment at the call
site, or pushing the whole condition into the consumer, keeps that honest.

### 5. The module doc states an invariant the code does not implement

`sigma_verdict.rs` header:

> The discipline that makes it correct is that `record` is called on
> **every** exit from the cascade, not only the uncertified one…

There is one `record` call in the diff, at the uncertified branch. The
certified exits are covered by `clear()` at the *top* of `solve_qp_core`
(`ipm.rs:3220`), which is a different (and equally sound) discipline:
clear-at-entry plus record-at-the-one-exit. The PR body repeats the same
phrasing ("re-recorded on **every** solve attempt"). This is a module whose
entire correctness argument *is* the discipline, so the stated invariant
should be the one a reader can check against the code — otherwise someone
later "restores" the missing `record` calls, or removes the `clear` on the
belief that the `record`s cover it.

### 6. CLAUDE.md names two fixtures for this exact code path; the PR does not

> A change to `crates/pounce-convex/src/ipm.rs`'s `normalized_optimum_is_genuine` / `hsde_cost_scale` needs `crates/pounce-convex/tests/issue414_cost_normalized_false_optimal.rs` and `illconditioned_huge_scale.rs`, which reach the accept and reject branches on purpose; the corpora are evidence of no *collateral* damage and nothing more.

`residual_compensated` feeds `sigma_forward_error_is_small`
(`ipm.rs:~2727`), which is inside the certification cascade — so this PR
changes the estimator the gate reads on **every** σ input, and that rule
applies. The PR's evidence section offers the empty sweep (correctly
discounted), the census (explicitly stated not to discriminate this
change), and constructed rows that prove the *arithmetic* is better without
showing the *verdict* is unchanged. Those two fixtures are presumably
inside "284 suites green", but per the repo's own rule they should be named,
with a line confirming each still reaches its intended accept/reject branch.
I did not run them.

---

### Summary

The core fix — carry the verdict out, apply it at the outermost layer —
looks right to me, and the reasoning for *where* to apply it is the
strongest part of the PR. Findings 4, 5 and 6 are documentation and
evidence-hygiene and are cheap.

Findings 1 and 2 are the ones I would want answered before merge: the
demotion is not only a label, it closes `QpSensitivity::build` and flips
`result.success` for a population that previously worked, and one member of
that population is an exact vertex. Both are questions, not defects — I
stopped short of measuring them at your request.

jkitchin:
Thanks — this landed while I was reworking the branch, so some of it is
answered by the rework and some of it is not. Taking the findings in order,
and flagging up front which ones changed the code.

**Note on what you read.** The review is against `5b41a57d`. That commit has
been replaced: the head now records a *narrower* verdict, and the six
loosened assertions are gone. Findings 2 and 3 are largely artifacts of the
old design; 1, 4, 5 and 6 were live and are now fixed.

### Finding 3 — the loosened assertions: you and the repo owner are both right, reverted

`assert_returned_an_optimum` is gone. Every sweep test asserts the **exact**
status expected at each conditioning and tolerance. That is strictly more
information than the originals carried, since they only ever expected
`Optimal`:

- `tol` sweep at `cond 1e10`: `Optimal` for `1e-4 ‥ 1e-8`, `OptimalInaccurate`
  at `1e-10`.
- bound- and inequality-constrained `cond` sweeps: `Optimal` at `1e6`/`1e8`,
  `OptimalInaccurate` at `1e10`.
- the `1e12` carve-out: `OptimalInaccurate`.

You were right that the loosening hid something, and it hid something
specific: with the whole-gate verdict recorded, an **inert objective
rescaling flipped the status six times** across `k = 1e-2 ‥ 1e8` on a
`cond 1e10` model — a transformation that leaves the argmin unchanged by
identity. The loosened assertion could not see it.

### The rework that finding exposed

"The cascade certified nothing" mixes whether the residual is small in
*absolute* terms — which legitimately depends on the objective's scale — with
whether `x` is far from the optimum, which does not. Only the forward-error
arm is recorded now (unconstrained, `Δ = −P⁻¹(Px + c)` is exactly invariant
under `(P, c) → k(P, c)`), and only with a 10× margin: at `cond 1e10` the
realized error is `1.47e-06` against a `1e-06` cut, so the model sits *on*
the boundary and rounding decides it. The margin is the estimator's own
uncertainty — a plain summation reads `0.29×`–`5.4×` of the truth against a
`Fraction` reference — not a knob. With it the rescaling is uniformly
`Optimal` across eleven decades.

Restated headline, measured against `κ·ε` rather than a flat `1e-6`:
**instances above their conditioning floor 3 → 0**, worst forward error
`3.06e-04` → `5.62e-05`, census iterations `443` → `404`.

### Finding 1 — the downstream gates: live, and you found the right precedent

This was the merge-blocker and I have taken your first reading, not the
second. The demotion changes a *label*, not what the library will do. All
four sites now admit `OptimalInaccurate`: `QpSensitivity::build`,
`pounce-py`'s payload gate, `QpResult.success`, and `minimize`'s `success`.

The deciding evidence is in the repo already — `_curve_fit.py` (gh#119/#123)
records that treating `Solved_To_Acceptable_Level` as a non-success
"reported `success=False` at a verified optimum … and callers gating on
`.success` discarded valid fits". Refusing derivatives at the demoted point
would have re-introduced exactly that on the convex arm, and the two arms
would have disagreed about the same converged-to-acceptable answer. Pinned by
`a_demoted_answer_still_builds_a_sensitivity_object`; restoring the narrow
gate turns it red.

### Finding 2 — crossover: the hazard is real, the instance is not

The ordering you describe is exactly right and `sigma_verdict`'s doc now says
so explicitly, including that a future consumer sensitive to *which* of the
two points the bit describes must re-record after crossover rather than
assume.

The instance that motivated it is gone, though: under the narrowed verdict
the big-coefficient LP is **not** demoted — its vertex is exact, so the
forward-error arm passes even though the absolute arm cannot. It is pinned by
`the_exact_vertex_lp_keeps_a_clean_optimal`, which also closes your finding 3
("the LP is the one demoting instance no longer pinned"). Record the whole
gate instead and that test goes red, along with the rescaling one.

### Finding 4 — `record`'s bit is narrower than its name

Correct, and now commented at the call site: the bit means "this answer
should be demoted" (`Optimal` **and** far from optimal), so an uncertified
pick already carrying a non-`Optimal` status records `false`. A second
consumer reading it as "the cascade certified nothing" would under-report;
the comment says to widen the recording rather than the reader.

### Finding 5 — the module doc stated an invariant the code does not implement

Also correct. Rewritten to the discipline actually implemented —
**clear-at-entry plus record-at-the-one-exit** — and it now says in terms not
to "restore" the missing `record` calls and not to remove the `clear` on the
belief the `record`s cover it, since those are the two opposite ways to break
it.

### Finding 6 — the two CLAUDE.md fixtures: named, run, and branch-verified

Both pass, and per the rule I checked they still *reach* their branches
rather than merely being green:

| fixture | reaches |
|---|---|
| `issue414_cost_normalized_false_optimal` | 22 × `accept:plain` |
| `illconditioned_huge_scale` | 4 × `accept:plain`, 1 × `accept:direct` |

`issue414` accepts via `plain`, i.e. the gate's **reject** arm fired on the
normalized candidate first. All 27 of those accept decisions read the changed
compensated estimator and still accept — which is the "no collateral damage"
evidence, now named and measured rather than folded into a suite count.

### Also dropped

While tracing the `k` sweep I found that `sigma_complementarity_is_genuine`'s
slack scale, `bound.abs().max(x.abs())`, **collapses to zero** when a variable
rests on a bound of `0`, so `|slack| <= cut·0` is false for any nonzero slack
and a converged `1e-16` is rejected. Fixing it was in an earlier revision. It
is a real latent defect, but once the verdict is the forward-error arm's it
changes nothing observable — identical census errors and iterations, empty
sweep, and no test fails without it — so it is **not** in this PR rather than
shipping untested.

### Current evidence

- `scripts/sweep-fixtures.sh`: empty, both legs, all 79 fixtures.
- Workspace suite: 347 suites green (`pounce-hsl` excluded; it does not link
  here, on this branch or on `main`).
- Mutation table now eight rows, each verified to turn the named test red.
- Local Python suite against a freshly built extension: 1405 passed, 4
  skipped. Verified there, not only in Rust, that the demoted population
  reaches Python as `optimal_inaccurate` with `success=True` — i.e. finding 1
  was reachable, not theoretical.
- CI green on `097fb99f` (8/8).

Thanks for the read — findings 1 and 5 in particular were things I would not
have caught from inside the change.

jkitchin:
## Final review — head `097fb99f`

Re-read the whole delta against `origin/main` (9 files, +683/−23), plus the
call graph around the new demotion. **The four findings from the first pass
that were live are fixed, and the rework is better than what it replaced.**
Confirmed by reading, not by trusting the response:

- **Downstream gates.** `sensitivity.rs:381` now matches
  `Optimal | OptimalInaccurate` with the `_curve_fit.py` (gh#119/#123)
  precedent written into the comment; `pounce-py/src/qp.rs:830`,
  `python/pounce/qp.py:204`, `python/pounce/_minimize.py:916,977` all widened.
  `a_demoted_answer_still_builds_a_sensitivity_object` pins the first with a
  named mutation. I grepped the rest of the workspace for status gates that
  could see this population: the only other `!= QpStatus::Optimal` sites
  (`pounce-sensitivity/src/solver.rs:1548`,
  `pounce-algorithm/src/crossover.rs:527`) are on `ParametricActiveSetSolver`,
  which σ never reaches. Clean.
- **The crossover ordering hazard** is pinned by
  `the_exact_vertex_lp_keeps_a_clean_optimal` and documented verbatim in
  `sigma_verdict.rs`'s "What the slot does not cover". I checked the pin has
  content rather than passing by accident: `maybe_crossover` gates on
  `!prob.p_lower.is_empty()`, so `P = 0` and the operator is `GᵀΣG + Σ_bnd`,
  which is *not* degenerate at that vertex — the arm genuinely evaluates and
  genuinely reads small. The test is testing something.
- **The module doc's discipline** is now stated correctly (clear-at-entry plus
  record-at-the-one-exit) and the mechanism holds: `record`/`clear` are
  `if slot.get().is_some()` no-ops, so an untracked entry cannot leak a stale
  verdict into a later solve, and the RAII `Reset` restores `None`.
- **Exact status assertions** are back across the sweep tests, in *both*
  directions, with the boundary written out per case rather than derived from a
  formula. That is the right call and the comment says why.

The narrowing to the forward-error arm is the right change and the argument for
it is the strongest thing in the PR: status must be invariant under
`(P, c) → k(P, c)`, the absolute arms are not, and the k-sweep flipping six
times is a contradiction with no oracle in it. Findings 2 and 3 from the first
pass are indeed artifacts of the replaced design.

Seven things below. One I would want addressed before merge (F1); one is a
two-minute process item (F5); the rest are documentation accuracy, which is the
class this repo has just spent PR #893 on.

---

### F1 — RISK (high). The demotion covers one of six public entries; `solve_qp_batch` and warm-started `solve_qp` still return a bare `Optimal` on an uncertified pick

`tracking()` is installed in `solve_qp_ipm` (`ipm.rs:376`) and nowhere else.
Six public entries reach `solve_qp_core` and therefore the same σ cascade and
the same `record()` line:

| entry | installs `tracking()`? |
|---|---|
| `solve_qp_ipm` (`:339`) | yes |
| `solve_qp_ipm_warm` (`:1170`) | **no** |
| `solve_qp_ipm_debug` (`:1095`) | **no** |
| `solve_socp_ipm` / `_warm` / `_debug` | **no** |

`record()` is a no-op with the slot uninstalled, so on those paths an
uncertified pick goes back under a bare `Optimal` — which is gh#880 verbatim.
That is not a hypothetical corner:

- `crates/pounce-py/src/qp.rs:525` — Python `solve_qp(..., warm_start=...)`
  routes to `solve_qp_ipm_warm`.
- `crates/pounce-convex/src/batch.rs:152` — every per-item solve in
  `solve_qp_batch` / `solve_qp_multi_rhs` and their parallel forms.

So on the PR's own fixture, `coupled_qp([1.0, 1e12], &TGT, 1.0)`:
`solve_qp` returns `optimal_inaccurate` and `solve_qp_batch([that problem])`
returns `optimal`. Same problem, same options, different label — and the batch
one is the label gh#880 was filed about.

The socp entries are narrower than they look and I do not think they matter
much: the cascade's `!cone.blocks().all(Nonneg) → return plain` early return
sits *above* the `record()` line, so a genuine SOC/exp/PSD problem never
reaches the recording site at all. Only an all-nonneg cone driven through
`solve_socp_ipm` does. `solve_qp_ipm_debug` is the CLI's `--debug` arm only
(the ordinary CLI path falls through to `solve_qp_ipm`, so the gh#535/#888
reroute claim in the CHANGELOG does hold).

`solve_qp_ipm_warm` and the batch API are the ones I would want closed. The
layering constraint is the same one the PR already established — the frame has
to sit outside every retry — so `solve_qp_ipm_warm`'s body is the analogous
placement, not `solve_qp_ipm_warm_inner`. If you would rather scope it than fix
it, that is defensible, but then the CHANGELOG's "the cascade's verdict is now
carried out to `solve_qp_ipm` and applied there" needs to say *which* entries,
because as written a reader takes it as covering `solve_qp`.

Related and worth one sentence somewhere the reader will meet it:
`sigma_forward_error_is_small` returns `true` on `m_eq() > 0`, so the demotion
is also structurally silent on equality-constrained σ models. The arm's own
"What it does not cover" says this, but the demotion site and the CHANGELOG
do not, and that is where a reader forms the belief about coverage.

### F2 — RISK. "Changes a label, not what the library will do" is true of the library and false of the CLI — and that path has no coverage

The CHANGELOG states the payoff plainly: "the CLI's existing gh#535/#888
reroute reads that and hands the model to the NLP arm". Confirmed —
`crates/pounce-cli/src/main.rs:2335` reroutes on
`OptimalInaccurate | IterationLimit | NumericalFailure`. Two further
consequences follow from the same label: `qp_status_to_ars` maps it to
`SolvedToAcceptableLevel` (`:2366`) and `convex_status_report` returns
solve_result_num `1` instead of `0` (`:2390`).

So for the demoted population the CLI now solves the model with a *different
engine* and returns a *different AMPL solve_result_num*. Both are probably
improvements. Neither is exercised anywhere in this PR, and neither can be:
`scripts/sweep-fixtures.sh` is empty because the one fixture of 79 that reaches
σ is `qcqp_columns_illcond`, a QCQP, which the non-orthant early return
excludes from the record site entirely. So the reroute for this population has
zero coverage in the PR *and* zero in the corpus — which is exactly the shape
CLAUDE.md warns about for this path ("an empty sweep is not evidence about
it").

One CLI-level test on a demoted model — assert it reroutes and what comes out —
would close it. This is the "it cannot produce a wrong answer is not the
relevant safety property" rule applied to a routing change.

### F3 — RISK. `SIGMA_DEMOTION_MARGIN`'s stated provenance measures a code path this PR deletes

At the recording site:

> `SIGMA_DEMOTION_MARGIN` is the estimator's own uncertainty, not a tuned knob:
> measured against a `Fraction` reference, a plain summation of this residual
> reads between `0.29x` and `5.4x` of the truth, so a verdict inside one order
> of magnitude of the cut is inside the estimator's noise…

But the same PR replaces that plain summation with `residual_compensated`, and
the comment 200 lines up says the compensated version "agrees to every digit
printed at `cond = 1e10` and `1e12`". The `0.29×`–`5.4×` spread is the noise of
an estimator that no longer exists in this code. As written, the constant's
justification is retired by the PR that introduces it.

The real reason is in the paragraph immediately above and is a *different*
mechanism: at `cond = 1e10` the realized error `1.47e-6` sits on the `1e-6`
cut, and each `k` in the rescaling sweep reaches a slightly different iterate,
so the verdict flips six times. That is solve-to-solve variation, not summation
noise, and it is a perfectly good reason for a margin.

There is also a better argument available than either, which I would suggest
writing down because it makes the constant *two-sidedly* pinned rather than
asserted:

- `MARGIN = 1.0` fails `an_inert_objective_rescaling_stays_inert_on_the_coupled_arm`
  (your own mutation table row);
- `MARGIN ≳ 1e2` fails `a_bound_constrained_coupled_instance_is_solved`, whose
  measured error is `1.6e-4` against `1e-6 · MARGIN`.

So the admissible window is roughly `[2, 100]` and `10` is its centre, held
from both sides by tests that already exist. That is a stronger statement than
"the estimator's own uncertainty", and it is checkable — which is /sens-review
entry 7's actual ask.

### F4 — RISK (minor). `demote_uncertified_sigma_optimum`'s doc attributes three numbers to a function that cannot move any of them

> After this, instances above their floor go 3 -> 0, worst forward error
> `3.06e-04` -> `5.62e-05`, census iterations 443 -> 404.

The function is a pure relabel at the outermost layer, after every retry has
run — the next paragraph of its own doc says exactly that. It cannot change a
forward error and it certainly cannot change an iteration count. Those numbers
belong to `residual_compensated` plus the cut, which reroute *which candidate
the cascade accepts*. The CHANGELOG attributes them correctly ("Measured on the
census"); only the function doc reads as attributing them to the demotion.
"After this PR" would fix it, or move the sentence to the recording site.

### F5 — RISK (process). The deliberately-unshipped complementarity defect has no issue

Your response records a real latent defect left out on the reasoning that under
the narrowed verdict it changes nothing observable — `sigma_complementarity_is_genuine`'s
slack scale `bound.abs().max(x.abs())` collapsing to zero on a bound of `0`.
The reasoning for keeping it out of *this* PR is sound and I agree with it.

But there is no issue for it. `gh issue list` returns only #880 itself, and
nothing in `crates/`, `dev-notes/` or `CHANGELOG.md` mentions it. CLAUDE.md is
explicit about this exact situation: a measured item recorded as an accepted
cost "needs an issue and an owner. Without one it is indistinguishable from
noise to the next reader." A two-minute `gh issue create` before merge, and
gh#880's follow-up list gets a line.

### F6 — NIT. Unused import

`crates/pounce-convex/tests/issue880_coupled_sigma_forward_error.rs:181` adds
`QpSolution` to the import list and no test uses it. `rustc` warns; CI stays
green because it only denies `clippy::correctness` and `clippy::suspicious`.
Drop it.

### F7 — Merge-order note (not a finding against this PR)

PR #889 adds `QpSensitivity::build_conic` with a narrow
`sol.status != QpStatus::Optimal` gate. It does not exist on this branch, so
this PR correctly did not touch it. Whichever of the two lands second should
widen it for the same reason `build` was widened. Low urgency — per F1 the
conic entries do not currently reach the record site — but it is the kind of
thing that is invisible once both are merged.

---

## Verdict

The layering argument is right, the narrowing is right and well-motivated, the
compensated residual is pinned on a case whose answer is arithmetic rather than
a tolerance, and the status boundary is now asserted in both directions instead
of assumed. This is a good fix.

**F1 is the one I would want resolved before merge** — either close
`solve_qp_ipm_warm` (and with it the batch API and warm-started Python
`solve_qp`), or state the scope explicitly in the CHANGELOG so the next reader
does not inherit the belief that gh#880 is closed everywhere. **F5** is two
minutes. F2 is a coverage gap I would like closed but would not block on. F3
and F4 are documentation accuracy, and F6 is a one-line deletion.

Method note: this is a reading review — I read the diff, the call graph around
`solve_qp_core`'s six entries, the crossover gate, the CLI status mapping, and
grepped the workspace for status gates that could see the demoted population. I
did not run the census or the fixture sweep, so every number I quote back is
yours; F1's reproducer is derived from the call graph and the PR's own fixture,
not executed.

jkitchin:
Second pass answered. **F1 does not reproduce**, and the interesting part is
why; F2 and F5 are closed; F3 was right and is re-derived from a measurement
instead of a retired one. Head is now `530121f0`.

### F1 — RISK(high): run rather than derived, and it does not reproduce

The call-graph reading is right as far as it goes: `solve_qp_ipm_warm_inner`
reaches `solve_qp_core` **directly**, not via `solve_qp_ipm`, so a missing
frame really would make `record` a silent no-op. What the graph does not show
is the options struct. That call passes a **direct, non-HSDE** `opts`, and the
entire `σ` block — cascade, gate and the `record` line — sits under
`if opts.use_hsde`. There is nothing to record on that route.

Your reproducer, executed on the PR's own `coupled_qp([1.0, 1e12], …)`:

| call | status | `‖x − x*‖` |
|---|---|---|
| `solve_qp` | `OptimalInaccurate` | `2.885e-4` |
| `solve_qp_batch([that problem])` | `OptimalInaccurate` | `2.885e-4` |

Same label, same point — and that holds *with and without* a frame installed
on `solve_qp_ipm_warm`, so the agreement is not something I added. The
demotion those paths do report arrives through the inner cold HSDE solve,
which carries its own frame.

I did add frames to all five other entries first, then measured them: stripped
all five, ran the suite in a **debug build with the new assertion active**,
nothing fired. So they were dead code, and I removed them rather than ship
something unmeasurable — the same standard that kept the complementarity fix
out of the last round.

**What ships instead is the guard your finding actually argues for.**
`sigma_verdict::record` now `debug_assert!`s that a frame is installed, with
the failure mode spelled out in the message. That converts "every entry that
reaches the cascade installs `tracking`" from a convention someone has to
remember into an invariant the suite enforces — including for the case you
flag, an entry that starts running the cascade because HSDE was enabled on its
route. `clear` deliberately does not assert, since it runs on every
`solve_qp_core` attempt including the many reached without a frame. Verified
across the convex and CLI suites in debug: never fires.

`every_public_entry_agrees_on_the_verdict` pins the batch/solo agreement, and
pins the scope you were right to ask for — a warm solve seeded from a nearby
point takes the direct driver and never engages `σ`, returning `2.885e-4`
under a clean `Optimal`. That is at the conditioning floor (`κ·ε ≈ 2.2e-04`),
not past it, and the test says so with the number so the scope note cannot go
stale silently.

Your related point about `m_eq() > 0` is correct and is now in the CHANGELOG's
scope sentence rather than only in the arm's own doc.

### F2 — RISK: closed

You are right that this population had zero coverage in the PR *and* zero in
the corpus, and right about why: the one fixture of 79 that reaches `σ` is a
QCQP, which the non-orthant early return excludes from the recording site.

`crates/pounce-cli/tests/issue880_sigma_demotion_reroutes.rs` plus a new
`issue880_sigma_uncertified.nl` fixture — the crate's own coupled `cond 1e12`
model written out, `f* = −2.205e12` by construction. Two tests:

- under `auto`: engine is `nlp` (the reroute fired), `SolveSucceeded`,
  objective within `1e-9` relative of the constructed optimum;
- under `solver_selection=qp-ipm`: engine `cvx-qp`, status
  `SolvedToAcceptableLevel`, AMPL `solve_result_num` **1**.

Both go red if the demotion is dropped. The engine assertion is deliberate —
status and objective alone would be satisfied by the convex arm's own answer,
so a routing regression would be invisible without it.

### F3 — RISK: correct, and re-derived

You are right that the `0.29×`–`5.4×` spread is the noise of an estimator this
same change replaces. Measured the way you suggested, so the constant is
pinned from both sides by tests that already exist:

| margin | outcome |
|---|---|
| `≤ 2` | `an_inert_objective_rescaling_stays_inert_on_the_coupled_arm` fails |
| `5`, `10` | green |
| `≥ 50` | `every_tolerance_is_correct_on_the_coupled_arm` fails |
| `≥ 200` | `a_bound_constrained_coupled_instance_is_solved` fails too |

Admissible window `(2, 50)` — tighter than your estimated `[2, 100]` — and
`10` is roughly its geometric centre. That table is now the comment, with a
note saying explicitly that the old justification cannot be the reason.

### F5 — fixed here rather than filed

This repo's standing preference is that an adjacent defect found mid-task gets
fixed in the branch at hand rather than filed, so I took the fix route and
solved the objection that kept it out last round — that it was untestable.
It is testable, just not through a corpus: `sigma_complementarity_is_genuine`
is reachable from a crate-internal unit test, so
`a_converged_slack_on_a_zero_bound_is_negligible` constructs the exact shape
(variable on `lb = 0`, slack `1e-16`, multiplier `1e10`) and asserts the guard
accepts. Drop the `.max(1.0)` and it fails. If you would still rather have an
issue, say so and I will file it instead.

### F4, F6 — fixed

F4: the demotion's doc now says in terms that the census numbers are the
*change's*, not the function's — it is a pure relabel after every retry and
cannot move a forward error, let alone an iteration count. F6: import dropped.

### F7 — noted

Agreed, and no action here. Whichever of #889 and this lands second should
widen `build_conic`'s gate for the same reason `build` was widened.

### Evidence

- **No pre-existing fixture moves.** The sweep now diffs exactly two lines,
  both of them the new `issue880_sigma_uncertified` fixture appearing (exact
  and lbfgs legs, engine `nlp`, `obj=-2.205e+12`). Nothing else moves in
  status, objective, iteration count or engine.
- Census unchanged: instances above their conditioning floor `3 → 0`, worst
  forward error `3.06e-04 → 5.62e-05`, iterations `443 → 404`.
- Workspace 348 suites green; convex and CLI suites also run in **debug**, so
  the new assertion is live.

Method note, since yours was explicit: F1's disproof and the F3 window are
executed, not read. The one thing I did not re-run this round is the Python
suite (1405 passed on `097fb99f`); nothing in this delta touches Python.

jkitchin:
## Third pass — head `530121f0`

Read the `097fb99f → 530121f0` delta and re-ran the parts of it that are
checkable. **F2, F3, F4, F5 and F6 are genuinely closed** — I verified F5's
mutation myself (drop `.max(1.0)` and
`a_converged_slack_on_a_zero_bound_is_negligible` goes red, exactly as its
table says), and the whole `pounce-convex` suite is green in a **debug** build,
so the new `record` assertion really is live and really does not fire.

Three things below. **F8 is a live defect and I would hold merge on it** — it is
F1's class, on the entry F1 named and then dismissed, and it is the entry the
new assertion now converts into a panic.

---

### F8 — RISK (high). `solve_socp_ipm` reaches the `record` site with no frame: `Optimal` in release, **panic in debug**

Run, not derived. Same model as the PR's own fixture,
`coupled_qp([1.0, 1e12], &TGT, 1.0)`, unconstrained, driven through the conic
entry with an empty cone list:

| build | `solve_qp_ipm` | `solve_socp_ipm` |
|---|---|---|
| release | `OptimalInaccurate`, `x = [2.999350875799344, 0.4991345010657874]` | **`Optimal`**, *the same* `x` |
| debug | `OptimalInaccurate` | **panics**: `the σ cascade recorded a verdict with no frame installed … (gh #880)` at `sigma_verdict.rs:90` |

The route: `solve_socp_ipm` → `solve_socp_ipm_inner` (no non-symmetric cone, no
PSD) → `solve_socp_symmetric` with the caller's **unmodified** `opts` → 
`solve_qp_core`. `use_hsde` defaults to `true` (`ipm.rs:319`), so unlike
`solve_qp_ipm_warm_inner` — which is right, and your F1 answer is right about it
— there is no `use_hsde: false` struct in the way. The cone is all-`Nonneg`
(or empty), so the non-orthant early return does not fire either, and control
reaches `record` at `ipm.rs:3474` with the slot uninstalled.

So the release half is gh#880 verbatim on that entry, and the debug half is a
**new** failure mode this commit introduces: a `debug_assert` that fires on a
legitimate user input. It is reachable from the public API —
`pounce-py/src/qp.rs:608` (Python `solve_socp`), `pounce-cli/src/main.rs:3258`,
`pounce-cli/src/bin/pounce_cblib.rs:154` — and `solve_qp_core`'s own gh#414
comment already records the shape that lands there: "`solver_selection=socp` on
a box-constrained QP lands here".

**There is a second unframed route to the same line.**
`solve_socp_ipm_warm`'s HSDE fallback (`ipm.rs:2010`) calls
`solve_socp_ipm_inner` **directly**, not `solve_socp_ipm`, so installing the
frame only in `solve_socp_ipm`'s scoped body leaves that one open.

Fix, verified to close the release half on the reproducer (`socp` now returns
`OptimalInaccurate`, same `x`):

```rust
let (inner, sigma_uncertified) = crate::sigma_verdict::tracking(|| {
    solve_socp_ipm_inner(prob, cones, opts, make_backend)
});
let sol = finite_or_failed(prob, demote_uncertified_sigma_optimum(inner, sigma_uncertified));
```

— the same outermost-layer placement the PR already argued for, applied at
`solve_socp_ipm` **and** at `solve_socp_ipm_warm`'s retry.

**On the measurement that concluded the five frames were dead code.** Stripping
them and observing that nothing fires is a measurement of the corpus, not of
the routes: there is no test anywhere that drives an all-nonneg
`solve_socp_ipm` at a `σ`-engaging magnitude, so the assertion had nothing to
fire on. That is CLAUDE.md's own branch rule — a corpus uniform in the
dimension the change acts on reports "nothing moves" whatever the truth is —
and it is the same rule the PR invokes correctly two paragraphs earlier to
justify writing `issue880_sigma_demotion_reroutes.rs`. The previous pass named
this exact route and waved it through ("only an all-nonneg cone driven through
`solve_socp_ipm` does … I do not think they matter much"); that was my error to
own, but the response turned the dismissal into "adding it to the other five
entries would be dead code", and the doc on
`every_public_entry_agrees_on_the_verdict` now states that as fact.

The assertion itself is a good addition and it is what makes this cheap to
find. Make the invariant it asserts true, and give that test a `solve_socp_ipm`
row.

### F9 — RISK. The `m_eq` scope sentence is reported as landed and is not in the diff

> Your related point about `m_eq() > 0` is correct and is now in the
> CHANGELOG's scope sentence rather than only in the arm's own doc.

`git diff 097fb99f 530121f0 | grep -iE 'm_eq|equality'` returns nothing, and the
gh#880 CHANGELOG entry and `demote_uncertified_sigma_optimum`'s doc both still
have no mention. `sigma_forward_error_is_small` still opens with
`if prob.m_eq() > 0 { return true; }` (`ipm.rs:2698`), so the demotion is
structurally silent on **every** equality-constrained `σ` model — which is a
scope a reader of the CHANGELOG's "reported `OptimalInaccurate`" would not
guess. One sentence, wherever you meant it to go.

### F10 — RISK. The CHANGELOG now carries both the retracted margin justification and its replacement

F3 was accepted and the code comment says so plainly — "that is the noise of an
estimator this same change replaces with `residual_compensated`, so it cannot
be the reason". The CHANGELOG still states it *as* the reason at line ~388:

> The margin is the estimator's own uncertainty — a plain summation of this
> residual reads `0.29×`–`5.4×` of the truth against a `Fraction` reference —
> so a verdict inside one order of the cut is inside the instrument's noise

…and then gives the corrected `(2, 50)` window four paragraphs later. Two
incompatible provenances for one constant in one entry, with nothing marking
the first as withdrawn. Delete the retracted clause; the new paragraph already
carries the claim.

### F11 — NIT (doc accuracy). The complementarity fix's inertness is measured, not structural

The new comment says it "is a *wasted re-solve* rather than a wrong answer
today — nothing downstream reads the per-candidate verdict, the demotion is the
forward-error arm's". The first half of that is not quite right:
`normalized_optimum_is_genuine` **is** the per-candidate verdict, and accepting
where it previously rejected makes `solve_qp_core` return the normalized solve
at `ipm.rs:3286` instead of falling through to the un-normalized re-solve, the
direct driver, and the `kkt_error` pick. A different candidate can come back,
and a demotion that would have been recorded can be suppressed — it just does
not happen on anything either corpus contains.

Which is fine, and the evidence you have (census identical, sweep clean, the
guard pinned directly) supports the measured claim. It is the structural claim
that overreaches, and "it cannot produce a wrong answer" is the argument
CLAUDE.md names as the one that shipped gh#544. Restate as measured.

---

## Verdict

The `.max(1.0)` fix is right and is pinned where a corpus could never reach it.
The CLI fixture is the right answer to F2 and the engine assertion in it is the
part that will still be doing work in a year. The margin is now justified by
something a reader can re-run. F1's disproof is correct **for the entry it
tests**.

**F8 holds merge.** F9 and F10 are two edits. F11 is a sentence.

Method note: F8's reproducer and F5's mutation were executed against
`530121f0` in a scratch worktree (`cargo test -p pounce-convex`, debug and
release; 265 unit + all integration targets green in debug). The candidate fix
in F8 was applied and re-run. I did not re-run the census, the fixture sweep,
the CLI suite or the Python suite — those numbers are yours.

jkitchin:
**F8 is right, it is my defect, and the way I got it wrong is worth naming.**
Fixed at `f014f6c4`, which is also rebased onto current `main` (12 commits,
including gh#892's restructure of the debug entries — that conflicted with
this work in three places and is resolved below).

### F8 — confirmed, then fixed

Reproduced exactly as reported, on the PR's own fixture:

| build | `solve_qp_ipm` | `solve_socp_ipm(prob, &[], …)` |
|---|---|---|
| release | `OptimalInaccurate` | **`Optimal`**, same `x` |
| debug | `OptimalInaccurate` | **panics** at `sigma_verdict.rs:90` |

So the release half was gh#880 verbatim on that entry and the debug half was a
failure mode I introduced one commit earlier.

**On the measurement that produced it.** You are right that stripping the five
frames and observing nothing fire was a measurement of the corpus, not of the
routes — no test drives an all-nonneg `solve_socp_ipm` at a σ-engaging
magnitude, so the assertion had nothing to fire on. That is CLAUDE.md's branch
rule, and I had invoked it correctly two paragraphs earlier to justify the CLI
fixture. Worse, your previous pass named this exact route and I turned your
"I do not think they matter much" into "adding it to the other five entries
would be dead code" and wrote it into a test doc as fact. The dismissal was
yours to own; promoting it to a measured claim was mine.

All six entries now install the frame — including the two that provably cannot
reach `record` today, since "every entry installs it" is the invariant the
assertion states and one documented exception is precisely how this happened.
`every_public_entry_agrees_on_the_verdict` now has **a row per entry, driven
directly**, rather than resting on a corpus-wide null result, and its doc
records why that shape rather than the other. The socp row is
mutation-checked: drop that frame and it goes red in release.

Your second unframed route is covered too — `solve_socp_ipm_warm`'s HSDE
fallback calls `solve_socp_ipm_inner` directly, so the frame sits at the entry,
not one layer in.

### The sweep, since it is the obvious question

Adding conic frames cannot touch the corpus, for the reason you gave: the one
fixture of 79 that reaches σ is `qcqp_columns_illcond`, a QCQP, excluded by the
non-orthant early return **before** the record site. Measured against a
baseline rebuilt from current `origin/main`:

```
40a41  > exact  issue880_sigma_uncertified  nlp  SolveSucceeded  it=2   obj=-2.205e+12
133a135 > lbfgs issue880_sigma_uncertified  nlp  SolveSucceeded  it=67  obj=-2.205e+12
```

Two lines, both the new fixture appearing. **No pre-existing fixture moves** in
status, objective, iteration count or engine. Census likewise unchanged against
that same fresh baseline: instances above their conditioning floor `3 → 0`,
worst forward error `3.06e-04 → 5.62e-05`, iterations `443 → 404`.

### F9 — you were right that it was not in the diff

It was not, and I reported it as landed. That is the second time this PR I have
described something as done without checking the diff, which is worth saying
out loud rather than quietly fixing. Now written, in the CHANGELOG, and it
names both silences rather than just the one: `sigma_forward_error_is_small`
returns `true` on `m_eq() > 0`, so the demotion is structurally silent on every
equality-constrained σ model; and it is silent wherever σ never engages, which
includes a warm solve seeded from a nearby point.

### F10 — both provenances, one constant

Retracted clause deleted. The measured `(2, 50)` window stands alone.

### F11 — restated as measured

You are right that `normalized_optimum_is_genuine` *is* the per-candidate
verdict, so accepting where it used to reject can return a different candidate
and can suppress a demotion. The comment now says that, then says it does not
happen on anything either corpus contains, and points at the unit pin as the
only evidence a corpus cannot give. "It cannot produce a wrong answer" was
exactly the wrong shape of argument to reach for in a file whose CHANGELOG
entry cites gh#544.

### Rebase

`main` moved 12 commits under this branch and gh#892 restructured the debug
entries (`solve_qp_ipm_debug_scoped` → `debug_stop::with_scope(||
solve_qp_ipm_scoped(…, Some(hook)))`). Three conflicts in `ipm.rs`, resolved by
keeping main's structure and wrapping it: the verdict frame now sits *outside*
`debug_stop::with_scope`, so it spans everything the scoped body may retry.

### Evidence at `f014f6c4`

- Release workspace: **350 suites**, 0 failures.
- Debug `pounce-convex`: **47 suites**, 0 failures, assertion firings **0** —
  so with the socp row present the invariant holds where it previously
  panicked.
- Sweep and census as above, both against a baseline rebuilt from
  `origin/main`.
- Mutation-checked: the socp frame, the `.max(1.0)` floor, the margin at
  `1.0`, the whole-gate verdict, the demotion call, the compensation.

I have not re-run the Python suite on this head (1405 passed two commits ago);
nothing in the F8–F11 delta touches Python, and CI's two Python jobs cover it.

PR #893, opened 2026-09-02
docs(884): correct the stale characterization wording after #885 (gh#891)

### LINKED ISSUE #891: Decision: clean up stale gh#884 characterization wording after #885?
## Context

Follow-up to PR #885, which implemented the late-detector/cold-retry route and closed #884 and #887.

A post-merge resolution check found the substantive solver and review concerns addressed. The durable dev note now distinguishes the failed in-flight switch from the shipped cold retry and records measurement provenance. A few older statements remain in the algorithm regression test, however.

Checked on `main` at `d32204e0`.

## Remaining inconsistencies

In `crates/pounce-algorithm/tests/issue_884_biactive_dual_divergence.rs`:

1. Lines 62–75 and the test name at line 543 say the returned point is the optimum / “at `f*`.” The same file correctly records at lines 44–48 and 517–553 that the measured base point is near the optimum, not at it, and deliberately tests loose `x` and objective tolerances.
2. Line 5 says every number in the file was measured on `87402274`, but the file now also records post-fix retry measurements—for example, the base/promoted trajectory comparison at lines 535–540.
3. Lines 648–652 say that engaging regularization “only on demand does not work either.” That measured negative applies to switching it in-flight after detection. PR #885's shipped mechanism does engage it conditionally, but by restarting cold.

These are prose and test-name consistency questions; no functional failure or CI regression was found.

## Decision requested

@jkitchin, should these statements be corrected in a small test/documentation-only change, or kept as historical names with a more explicit clarification?

If corrected, the narrow scope would be:

- describe the returned point as feasible and near the known optimum rather than exactly at it;
- scope provenance to the measurement group or commit that produced each number; and
- replace “only on demand” with “in-flight after detection” so it does not contradict the shipped cold retry.

No implementation is proposed until that decision is made.

jkitchin:
*Body revised at `a54a2a5` after review. The original described `fef2dec` only, and carried two of the same defects the review found in the file — see **Review round** at the end.*

## Problem

Statements in `crates/pounce-algorithm/tests/issue_884_biactive_dual_divergence.rs` describe the code as it was before #885 shipped the late-detector/cold-retry route. gh#891 found three on `main` at `d32204e0`; review found three more sites of the same class. No functional failure, but each misleads a reader about what the file pins.

1. The module table, the asymmetry paragraph and two test names say the returned point **is** the optimum / is "at `f*`". The file says twice elsewhere that the run stops *near* the optimum, and asserts loose `1e-3` / `1e-6` bars for exactly that reason.
2. The header claimed every number in the file was measured on `87402274`. That commit predates the retry; the file has recorded retry-era numbers since `ef1dd0b` / `198173b`.
3. `dual_regularization_reaches_the_optimum_honestly` said the dev-note explains why engaging regularization "only on demand does not work either". What shipped engages it conditionally — so the comment contradicted the mechanism it sat next to.
4. The narrative at the top of the file said the solve "stops there", present tense, of behaviour #885 superseded.

Two numbers were also wrong:

| trajectory paragraph | said | correct |
|---|---|---|
| base attempt's constraint violation | `2.7e-15` | `2.2e-16` — `2.7e-15` is its `x[2]` |
| unscaled dual residual the retry buys | "nine orders" | eighteen: `3.25e11` → `9.96e-8` |

The first was **findable by reading**: `main` already carried `2.2e-16` at header line 45, 500 lines above the paragraph that called the violation `2.7e-15`, and the point in that same sentence identifies `2.7e-15` as `x[2]`. Re-measuring supplied the reason, not the detection. (An earlier revision of this body claimed re-measurement found it; that over-credited the method.)

## Cause

The wording was written while gh#884 was open, against a solver in which the retry did not exist, and #885 changed the mechanism underneath it without revisiting the prose. Statement 3 is the sharpest case: "only on demand" was an accurate summary of "Ruled out 2" — an *in-flight* switch after detection — right up until the fix shipped a mechanism that is also on demand, but cold.

## Fix

Prose, two test names, two assertion messages. No behaviour changes.

1. **Near, not at.** Module table, asymmetry paragraph, two assert messages, and both tests that named the point: `qpec_small_returns_the_optimum_…` → `…returns_a_near_optimal_point_…`, and `dual_regularization_reaches_the_optimum_honestly` → `…reaches_a_certificate_honestly`. The second run lands **bit-identical** to the point the retry promotes (`[0.9999940320539338, 0.9999970160262548, 3.719079302430919e-6]`, measured both ways on `d32204e0`) because the retry *is* that configuration restarted cold — so naming one "near-optimal" and the other "the optimum" was two conventions for one point. What that test pins is the residual behind the verdict, not the iterate's location.
2. **Provenance per measurement group.** Almost everything is defect-era (`87402274`): the multiplier table, the base point, `3.25e11`, `7.9e+04`, `4.64e-4`, `9.96e-8`, and — from one dev-note table row — `ralph1`'s objectives under `perturb_always_cd` together with the `5.25e-7` that run certifies. `ralph1`'s `7.2e-3` floor is defect-era too: it is the minimum `||d||` of its *base* `RestorationFailed` trajectory, which the retry never touches. Only the promoted point and its violation are retry-era. Each site now names its group.
3. **In-flight, not on demand.** Names "Ruled out 2" and its reason, then says what the negative does **not** cover: the shipped cold restart.
4. **The narrative names its group** — "and stops there" → "and before #885 stopped there", with a clause saying the default path now retries.
5. One spelling of the runaway residual (`3.25e11`, matching `application.rs:62`/`:8093`/`:8109` and the dev-note `:100`/`:561`); `4.64e-4` added to the provenance enumeration; and the header's "three approaches ruled out" corrected to four, which is what the dev-note has.

The alternative gh#891 offered — keep the names as historical and annotate them — was not taken. These are not deliberate history; the file holds contracts and `dev-notes/mpcc-biactive-dual-divergence.md` holds the history, and the dev-note already draws the in-flight/cold distinction correctly.

## Blast radius

Two files: the test file (comments, two test names, two `assert!` messages) and **one line** of `dev-notes/mpcc-biactive-dual-divergence.md` — its "Pinned by `dual_regularization_reaches_…`" reference, which the rename would otherwise dangle. No solver source, so no trajectory change and no fixture sweep. Both renamed tests have no references outside those two files (`grep` confirms). `crates/pounce-cli/tests/issue_884_biactive_dual_divergence.rs` carries none of this wording and is untouched.

## Tests

Nothing new is pinned — no behaviour change and no new test, so there is no "fails on the parent commit" claim to make. What backs it is that the numbers were **re-measured on `d32204e0`** (temporary instrumented copy of the fixture, deleted before commit):

| number | how | reproduces |
|---|---|---|
| base `(1.0002321, 1.0001161, 2.67e-15)`, `f = 6.73e-8`, viol `2.2e-16`, unscaled KKT `3.25e11`, `RestorationFailed` | `dual_divergence_retry=no` | yes — except the violation, corrected above |
| promoted `(0.9999940, 0.9999970, 3.72e-6)`, `f = 5.84e-11`, viol `5.47e-12`, unscaled KKT `9.96e-8`, `SolveSucceeded`, promoted | default path | yes |
| regularized-from-start returns a vector **bit-identical** to the promoted one | `perturb_always_cd=yes` vs. default, compared elementwise | yes |
| `ralph1` under `perturb_always_cd`: `SolveSucceeded`, `f = -2.71e-5`, unscaled KKT `5.25e-7` (`max_iter = 3000`); `Solved_To_Acceptable_Level` at `-3.81e-5` (`max_iter = 300`) | direct run | yes |

**Not** re-derived, and the file now says so rather than claiming otherwise: `ralph1`'s `7.2e-3` step floor. It is the minimum `||d||` over its base trajectory (dev-note `||d||` table, `7.153e-3`, `87402274`) — a minimum over a trajectory rather than a quantity the solve returns — and it is already mutation-checked by `the_detector_must_not_fire_on_ralph1`.

- `cargo test -p pounce-algorithm --test issue_884_biactive_dual_divergence` — 8/8 pass, on both commits and on the parent.
- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p pounce-algorithm --test issue_884_biactive_dual_divergence` — 7 warnings, the same count as on `d32204e0`'s version of the file (checked by restoring it and re-running). All pre-existing.
- The full workspace `cargo test` was **not** run locally; the diff cannot reach anything outside this test target. CI runs it.

## Review round

`a54a2a5` answers @jkitchin's review of `fef2dec` in full. Its finding 1 was correct and was the same defect class gh#891 filed, in the paragraph added to fix it: the retry group held two `87402274` measurements (`7.2e-3` and `5.25e-7`) under a claim they "could only be measured with the fix in place", refuted by the dev-note at `:21`, `:171` and `:313`. `5.25e-7` was split from the very table row supplying `-2.71e-5`, which the same paragraph assigned to the defect group. Findings 2–5 are folded into the sections above.

---

- [ ] Tests fail on the parent commit for the stated reason — **N/A**, no behaviour change and no new test; the renamed tests pass on both sides
- [ ] `CHANGELOG.md` `[Unreleased]` entry — **not added**: CONTRIBUTING asks for entries "in the user's terms", and nothing here is visible to a user of the CLI, Python or Pyomo surfaces
- [ ] Book page under `docs/src/` — **N/A**, internal test documentation
- [x] `cargo fmt --all -- --check` clean; clippy unchanged from parent on the touched target; the touched test target passes (full workspace suite left to CI — see above)
- [x] Every claim in this body and in the code comments is true of the code as it stands now — this body was rewritten at `a54a2a5` because two of its claims were not: the blast radius said one file when the rename touches the dev-note too, and the provenance summary described `fef2dec`'s grouping, which review showed was wrong

Fixes #891

https://claude.ai/code/session_01YayspqSHzQZGKmX887kh17

### DIFF
--- a/crates/pounce-algorithm/tests/issue_884_biactive_dual_divergence.rs
+++ b/crates/pounce-algorithm/tests/issue_884_biactive_dual_divergence.rs
@@ -2,19 +2,29 @@
 //! lowering's multipliers diverge, and the run fails next to the primal
 //! solution.
 //!
-//! **Provenance, by measurement group.** The numbers describing the
-//! *defect* — the multiplier table below, the base attempt's returned
-//! point, the `3.3e11` and `7.9e+04` residuals, the `9.96e-8` an honest
-//! solve reaches, and `ralph1`'s objectives under `perturb_always_cd` —
-//! were measured on `87402274`, before the fix existed. The numbers
-//! describing the *retry* — the promoted point and violation in
+//! **Provenance, by measurement group.** Almost everything here is
+//! *defect*-era, measured on `87402274` before the fix existed: the
+//! multiplier table below, the base attempt's returned point, the
+//! `3.25e11` and `7.9e+04` residuals, the `4.64e-4` objective-gradient
+//! residual it leaves, the `9.96e-8` an honest solve reaches, and — both
+//! of them, from one run — `ralph1`'s objectives under
+//! `perturb_always_cd` and the `5.25e-7` that run certifies.
+//! `ralph1`'s `7.2e-3` step floor belongs to that group too: it is the
+//! minimum `||d||` of its *base* `RestorationFailed` trajectory, which
+//! the retry never touches (`dev-notes/mpcc-biactive-dual-divergence.md`,
+//! the `||d||` separation table).
+//!
+//! Only the **promoted** point and its violation, in
 //! `qpec_small_returns_a_near_optimal_point_and_never_claims_a_false_success`,
-//! `ralph1`'s `7.2e-3` step floor, and the `5.25e-7` its regularized run
-//! certifies — could only be measured with the fix in place, and come
-//! from its own commits (`ef1dd0b`, `198173b`; PR #885). Every number in
-//! both groups reproduces on `d32204e0`. Each site names the group it
-//! belongs to; re-measure rather than trusting any of them across a
-//! commit that touches the IPM.
+//! are *retry*-era — they could not exist before the fix — and they come
+//! from its own commits (`ef1dd0b`, `198173b`; PR #885).
+//!
+//! Both groups were re-run on `d32204e0` and reproduce, with one
+//! exception that was **not** re-derived there: the `7.2e-3` floor, which
+//! is a minimum over a trajectory rather than a quantity the solve
+//! returns, and whose source is the dev-note's table. Each site names the
+//! group it belongs to; re-measure rather than trusting any of them
+//! across a commit that touches the IPM.
 //!
 //! **The first four are invariants rather than a description of the
 //! defect.** Each is written so that it stays correct after a fix: it
@@ -31,8 +41,8 @@
 //! exists because the rule has a branch a green fixture would otherwise
 //! never execute. See the block comment above them.
 //!
-//! The measurements — the runaway multipliers, the `mu` trace, and three
-//! approaches ruled out — live in
+//! The measurements — the runaway multipliers, the `mu` trace, and the
+//! four approaches ruled out — live in
 //! `dev-notes/mpcc-biactive-dual-divergence.md`. That is where history
 //! belongs; this file holds contracts. If you are fixing gh#884, read the
 //! note first: the obvious remedy is measured and rejected there.
@@ -53,8 +63,13 @@
 //! certifies stationarity with residual `0` **at that exact point** — so
 //! the answer exists and is reachable.
 //!
-//! What happens instead: the solve reaches `(1.0002321, 1.0001161,
-//! 2.67e-15)`, feasible to `2.2e-16` with `f = 6.73e-8`, and stops there.
+//! What happens instead — the *base* attempt, which is what gh#884 is
+//! about, and which the default path still runs first: it reaches
+//! `(1.0002321, 1.0001161, 2.67e-15)`, feasible to `2.2e-16` with
+//! `f = 6.73e-8`, and before #885 stopped there. (Since #885 the default
+//! path does not stop there: it detects the runaway, retries cold, and
+//! returns the promoted point instead. Everything from here to the end of
+//! this section describes the attempt that fails.)
 //! That is *near* the optimum, not at it — `lambda = 0` does not certify
 //! the returned iterate, where it leaves an objective-gradient residual of
 //! `4.64e-4`. Meanwhile the multipliers on the *linearly dependent* rows
@@ -68,7 +83,7 @@
 //! | 2 (`G2 >= 0`) | `1.0` | `-7.283e11` |
 //! | 5 (`G2·H2 = 0`) | `2.3e-4` | `+1.737e15` |
 //!
-//! `-7.283e11 + 1.737e15 · 2.3e-4` is the `3.253e11` residual it fails
+//! `-7.283e11 + 1.737e15 · 2.3e-4` is the `3.25e11` residual it fails
 //! on. That table is measurement, not an assertion — see the note.
 //!
 //! **The asymmetry between the two fixtures is the point.** `qpec_small`
@@ -84,7 +99,7 @@
 //! |---|---|---|
 //! | `qpec_small_returns_a_near_optimal_point_and_never_claims_a_false_success` | the returned point is feasible and within `1e-3` of `(1, 1, 0)` with `|f|` under `1e-6` — near `f*`, not at it; any claimed success must hold in the model's own units | a verdict-only "fix" reports success at a `1e11` residual |
 //! | `a_structurally_zero_hessian_entry_does_not_change_the_solve` | declaring an identically-zero Hessian entry is a no-op (and refutes gh#884's stated prerequisite hypothesis) | the declared sparsity pattern starts changing the answer |
-//! | `dual_regularization_reaches_the_optimum_honestly` | an honest solve of this model exists at `9.96e-8` unscaled, so gh#884 is a POUNCE gap | that configuration stops reaching the answer |
+//! | `dual_regularization_reaches_a_certificate_honestly` | an honest solve of this model exists at `9.96e-8` unscaled, so gh#884 is a POUNCE gap | that configuration stops reaching the answer |
 //! | `ralph1_must_not_claim_success_where_no_multiplier_certifies_it` | a model with no sign-feasible multiplier must not report success, and never below `f*` | `perturb_always_cd` is turned on by default — measured: plain `Solve_Succeeded` at `-2.71e-5` |
 //! | `the_kill_switch_restores_the_pre_884_verdict` | `dual_divergence_retry=no` returns the base attempt's verdict | the kill switch stops killing |
 //! | `a_zero_step_tolerance_holds_the_detector_off` | a `0` step tolerance is a second, finer off switch | the detector stops consulting its threshold |
@@ -482,7 +497,7 @@ fn claims_success(status: ApplicationReturnStatus) -> bool {
 ///   product row's gradient is exactly `(0, 0, 0)`, so `lambda = 0`
 ///   certifies with residual `0`; and an honest solve reaching `9.96e-8`
 ///   unscaled is measured in
-///   `dual_regularization_reaches_the_optimum_honestly`. A success claim
+///   `dual_regularization_reaches_a_certificate_honestly`. A success claim
 ///   on this model at a residual above `1e-6` is therefore *known* to be
 ///   avoidable, which is what makes it suspect rather than merely loose.
 /// - `ralph1` — no sign-feasible multiplier exists at its origin at all,
@@ -543,7 +558,7 @@ fn a_claimed_success_must_be_real(
 ///
 /// The **conditional** half is the guard, and it is no longer vacuous.
 /// Before gh#884 closed, this model exited `RestorationFailed` at an
-/// unscaled KKT error of `3.3e11` (and, from the acceptable-level start
+/// unscaled KKT error of `3.25e11` (and, from the acceptable-level start
 /// the issue filed, `Solved_To_Acceptable_Level` at `7.9e+04`), so the
 /// guard had nothing to check. The dual-divergence retry now takes it to
 /// `Solve_Succeeded` with a residual that holds in the model's own units,
@@ -669,6 +684,15 @@ fn a_structurally_zero_hessian_entry_does_not_change_the_solve() {
 /// an **unscaled** KKT error of `~1e-7` — in the model's own units, not
 /// by normalising the residual away.
 ///
+/// The name says *certificate*, not *optimum*, and the distinction is the
+/// same one the `qpec_small_returns_a_near_optimal_point_…` rename draws.
+/// This run lands on `(0.999994, 0.999997, 3.7e-6)` — bit-identical to
+/// the point the retry promotes, measured on `d32204e0`, because the
+/// retry *is* this configuration restarted cold — which is near
+/// `(1, 1, 0)` and not at it. What is pinned here is that the residual
+/// behind the verdict holds in the model's own units — not that the
+/// iterate is the optimum.
+///
 /// This is evidence that there is something to fix. It is **not** an
 /// argument for turning that option on by default: see
 /// `ralph1_must_not_claim_success_where_no_multiplier_certifies_it`, and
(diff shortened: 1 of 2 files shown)


jkitchin:
## Review — read at `fef2dec4`

**Method.** Read-only. I tried to re-measure the corrected numbers with a
throwaway instrumented test and the sandbox declined to let me write into
the crate, so instead I checked every claim against two independent
sources: `dev-notes/mpcc-biactive-dual-divergence.md`, and lines of this
same file that the PR does **not** touch. That turned out to be enough to
settle both the corrections and one new problem.

### Verified correct

- **The `2.7e-15` → `2.2e-16` correction is right, and independently
  corroborated without re-running anything.** `main` already carried
  `2.2e-16` at header line 45 — *"the solve reaches `(1.0002321,
  1.0001161, 2.67e-15)`, feasible to `2.2e-16`"* — 500 lines above the
  paragraph that called the violation `2.7e-15`. Two statements in one
  file disagreed; the diagnosis that `2.7e-15` is `x[2]` is confirmed by
  the point itself in the same sentence. (Small note on the body's
  framing: this one *was* findable by reading, not only by re-measuring.)
- **All four new order-of-magnitude claims check out.** `3.25e11 → 9.96e-8`
  is 18.5 orders ("eighteen" ✓); `2.2e-16 → 5.5e-12` is 4.4 ("four of
  primal" ✓); `6.73e-8 → 5.84e-11` is 3.1 ("three orders nearer in
  objective" ✓); `2.32e-4 → 6.0e-6` in `x` is 1.6 ("about one and a half"
  ✓). The superseded "three orders of primal" is consistent with the wrong
  base number, which is a good sign the correction is coherent.
- **Statement 3's rewrite matches the dev-note.** "Ruled out 2" is titled
  *"detect the runaway and engage `delta_c` in flight. **Too late by
  construction.**"* and the new comment says exactly that, with the right
  reason. The in-flight/cold distinction is drawn correctly.
- **Blast radius is as stated.** The renamed test has no references outside
  this file, and `crates/pounce-cli/tests/issue_884_biactive_dual_divergence.rs`
  genuinely carries none of this wording (grep confirms both). One test
  file, no solver source, so no sweep — agreed.

---

### 1. The new provenance paragraph misattributes two of the three numbers in its retry group

This is the finding I would fix before merge, because it is the same
defect class gh#891 filed, in the paragraph added to fix it.

The new header says:

> The numbers describing the *retry* — the promoted point and violation in
> `qpec_small_…`, `ralph1`'s `7.2e-3` step floor, and the `5.25e-7` its
> regularized run certifies — **could only be measured with the fix in
> place**, and come from its own commits (`ef1dd0b`, `198173b`; PR #885).

Both of the `ralph1` numbers are defect-era, and the dev-note this file
points readers at says so explicitly:

- **`7.2e-3`** is `ralph1`'s minimum `||d||`, from
  `dev-notes/mpcc-biactive-dual-divergence.md:313–321` — *"Minimum `||d||`
  over each run, taken only while the primal is converged (`inf_pr <=
  1e-8`), **measured on `87402274`**"*, table row `7.153e-3`. It is a
  property of `ralph1`'s **base** `RestorationFailed` trajectory (39
  iters, Ruled out 3), which the retry does not touch — the whole point of
  `the_detector_must_not_fire_on_ralph1` is that nothing fires there. It
  is measurable, and was measured, before the fix existed.
- **`5.25e-7`** is `dev-notes/…:171`, the Ruled out 3 table:
  `| perturb_always_cd=yes, max_iter=3000 | SolveSucceeded (356 iters) | -2.71e-5 | 5.25e-7 |`,
  and the note's own provenance line (`:21`) puts **Ruled out 1–3 on
  `87402274`**. That is a `perturb_always_cd` run, which predates the
  retry entirely.

`5.25e-7` is the sharper case: the *same table row* supplies `-2.71e-5`,
which this header correctly assigns to the **defect** group ("`ralph1`'s
objectives under `perturb_always_cd`"). One measurement is currently split
across both groups.

Only "the promoted point and violation" is genuinely retry-era.

### 2. …which is also why the header and the PR body contradict each other

The header ends the paragraph with *"Every number in both groups
reproduces on `d32204e0`."* The PR body says:

> Deliberately **not** re-derived: `ralph1`'s `7.2e-3` step floor…

So the file asserts a verification the body says was not performed. These
are the same mistake seen from two sides: it was not re-derived precisely
because it is not a retry-era number. Fixing finding 1 fixes this too —
but the universal quantifier ("every number … reproduces") should be
narrowed to what was actually re-run either way. On a PR whose thesis is
that unbacked claims accumulate, that sentence is the one to be most
careful with.

### 3. The rename is applied to one test and not to the other test that names the same point

`qpec_small_returns_the_optimum_…` → `…returns_a_near_optimal_point_…`
because the run stops near `(1, 1, 0)`, not at it. Correct. But
`dual_regularization_reaches_the_optimum_honestly` (line 684) is untouched,
and the point *it* reaches is — per `dev-notes/…:130` — `x = (0.999994,
0.999997, 3.7e-6)`, which is **the same point** the PR's new doc comment
calls the promoted retry answer `(0.9999940, 0.9999970, 3.72e-6)` and
explicitly says is "still not the point itself". Same distance from the
optimum, opposite naming convention, two functions apart. Either rename
both or say in the doc why "honestly" carries the weight there ("reaches a
*certificate*" is the real content, and would be a truer name).

### 4. The most prominent instance of gh#891's statement 1 is untouched

Module header, lines 44–48, unchanged from `main`:

> What happens instead: the solve reaches `(1.0002321, 1.0001161,
> 2.67e-15)`, feasible to `2.2e-16` with `f = 6.73e-8`, **and stops
> there.**

Post-#885 the default solve does not stop there — it retries and returns
`(0.9999940, 0.9999970, 3.72e-6)`. Present tense, first screen of the
file, describing behaviour the fix superseded. The PR corrected the
summary table row and the test doc comment that *summarize* this
paragraph, but left the narrative itself.

It is partly covered: the new provenance header assigns "the base
attempt's returned point" to the defect group. But the PR's own stated
principle is *"Each site names the group it belongs to"*, and this site
does not — a reader hits "What happens instead: the solve reaches …" long
before they map it to a group. One clause ("the base attempt reaches …,
and before #885 stopped there") closes it.

### 5. Minor

- **`4.64e-4`** — the objective-gradient residual at the returned iterate,
  header line 48 — is not in the provenance enumeration, which otherwise
  reads as exhaustive. It is defect-era.
- **Three roundings of one number.** `3.3e11` (new provenance header),
  `3.253e11` (mechanism paragraph), `3.25e11` (test doc, new in this PR).
  Two were pre-existing; this PR adds the third. In a file being edited for
  numeric provenance, one spelling is worth the churn.

---

### Verdict

The corrections themselves are right — I checked the arithmetic and
cross-checked the one substantive number against an untouched line in the
same file — and the "near, not at" rewrite, the assertion-message fixes and
the "Ruled out 2" restatement are all accurate against the dev-note.

Finding 1 is the one I would want fixed before merge: it puts two
`87402274` measurements in the retry group and claims they "could only be
measured with the fix in place", which the dev-note directly refutes at
`:21`, `:171` and `:313`. Findings 3 and 4 are the same completeness
question the issue asked, applied to sites the sweep did not reach.
Finding 5 is polish.

jkitchin:
All five findings addressed in `a54a2a5`. Finding 1 was right and I checked it against the sources you named before changing anything.

**1 — the retry group held two defect-era numbers.** Confirmed at `dev-notes/mpcc-biactive-dual-divergence.md:21` (Ruled out 1–3 on `87402274`), `:171` (the Ruled out 3 row supplying both `-2.71e-5` and `5.25e-7`), and `:313–321` (the `||d||` table, `7.153e-3`, `87402274`). Both numbers are pre-fix and neither needed the retry to exist, so "could only be measured with the fix in place" was false of them. Only the promoted point and its violation are retry-era now. The `5.25e-7` case is the one that stings: as you say, the *same table row* was split across both groups.

**2 — the universal claim is narrowed.** The paragraph now names the one exception and why it is an exception (a minimum over a trajectory, not a quantity the solve returns), instead of asserting a verification the body said was not performed. Agreed on which sentence to be most careful with.

**3 — renamed, and there is a measurement behind it now.** `dual_regularization_reaches_the_optimum_honestly` → `dual_regularization_reaches_a_certificate_honestly`, with the dev-note's "Pinned by" reference updated. Your inference was right and slightly understated: the two points are not merely the same to the digits printed in the note, they are **bit-identical** on `d32204e0` — I ran the regularized-from-the-start configuration and the default (retrying) path and compared the returned vectors, `[0.9999940320539338, 0.9999970160262548, 3.719079302430919e-6]` both times. That is what "the retry is this configuration restarted cold" means operationally, so the doc now says it and says the test pins the residual behind the verdict, not the iterate's location.

**4 — the narrative names its group.** "and stops there" → "and before #885 stopped there", with a parenthetical saying the default path now retries and returns the promoted point, and that the rest of the section is about the attempt that fails.

**5 — both.** `4.64e-4` is in the enumeration. The residual is `3.25e11` at all three sites; I picked that spelling because it is what `application.rs:62`/`:8093`/`:8109` and the dev-note (`:100`, `:561`) already use, so the file now agrees with the rest of the repo rather than just with itself.

One more of the same class, one line over from finding 5, so I fixed it here rather than filing it: the header called the dev-note's negative results "three approaches ruled out". There are four sections and the note's own intro says four.

On the body's framing — you are right, and it is worth correcting rather than leaving: `2.7e-15` was findable by reading, and specifically by reading line 45 of the file it is in. What re-measuring added was the *reason* (it is `x[2]`), not the detection. I will not rewrite the merged body of this PR over it, but the claim as stated over-credits the method.

Not done, and flagging rather than deciding: this round leaves the PR body's "not re-derived" line accurate but its provenance summary one revision behind the file. Say the word if you want the body edited to match `a54a2a5` before merge.

CI is re-running on the new head; I will report if anything goes red.

---

jkitchin:
Closing the loop on the open question from my last comment: the PR body is now rewritten against `a54a2a5`, and it carried two defects of its own worth naming rather than quietly fixing.

- **Blast radius said "one test file, no solver source".** False since the rename: it also touches one line of `dev-notes/mpcc-biactive-dual-divergence.md`, the "Pinned by" reference that would otherwise dangle. Two files.
- **The provenance summary described `fef2dec`'s grouping** — the one your finding 1 showed was wrong — so the body asserted the misattribution after the file had stopped doing so.
- The `2.7e-15` framing is corrected: findable by reading line 45, as you said; re-measuring supplied the reason (`x[2]`), not the detection.
- The "not re-derived" line is also fixed. It called `7.2e-3` "a property of the detector's threshold", which is backwards — it is the minimum `||d||` of `ralph1`'s base trajectory, and the threshold was chosen to sit under it.

The body now leads with a note that it was revised at `a54a2a5`, so the record shows the correction rather than hiding it.

---

Issue #414, opened 2026-07-31
[adversary] qp-ipm: SolveSucceeded/success=True at a non-KKT point (its own kkt_error=8.3e3) on a variable-scale-spread convex QP


jkitchin:
adversary-fingerprint: qp-ipm/false-success-nonzero-kkt-illscaled

Discovered by the nightly adversary run (automated solver correctness testing).

## Family / solver

Convex QP interior-point (`pounce-convex`) — reached via `pounce.solve_qp(method="ipm")`,
and via `solver_selection=qp-ipm`, `socp`, and `auto` on a `.nl` file.

The active-set engine (`method="active-set"` / `solver_selection=qp-active-set`) and the
NLP engine (`solver_selection=nlp`) both solve the **same** problem correctly, so this is
specific to the convex IPM path, not to the instance being unsolvable.

## Problem

A box- and inequality-constrained strictly convex QP whose *variables* span a wide range of
scales (`x_i ~ 10^-6 .. 10^6`), giving `cond(P) ~ 1e24`. The problem is trivially
well-conditioned after the diagonal rescaling `z = x / s` (`cond = 10`), which is what the
oracle uses.

`solve_qp` returns `status="optimal"`, `success=True` — while the `kkt_error` it
returns **on that very same result object** is `8282.5`. The returned point is not close to
the optimum: objective `67.13` vs the true `-3.9585`.

## Reproduction

```python
import numpy as np, scipy.sparse as sp, pounce, clarabel
n, dec, seed = 3, 6, 4245
r = np.random.default_rng(seed)
s = np.logspace(-dec, dec, n)                  # variable scales 1e-6 .. 1e6
Q, _ = np.linalg.qr(r.standard_normal((n, n)))
P = np.diag(1/s) @ (Q @ np.diag(np.logspace(0, 1, n)) @ Q.T) @ np.diag(1/s); P = (P+P.T)/2
c = r.standard_normal(n)/s
A = r.standard_normal((1, n)) @ np.diag(1/s)
xf = r.standard_normal(n)*s
b = A @ xf + r.uniform(.1, 1, 1)
lb, ub = xf - 10*s, xf + 10*s
f = lambda x: 0.5*x@P@x + c@x

res = pounce.solve_qp(P=P, c=c, G=A, h=b, lb=lb, ub=ub, tol=1e-8, method="ipm")
print("ipm      : status=%s success=%s obj=%.10g kkt=%s" %
[...]

PR #894, opened 2026-09-02
Attaching the debugger must not change the solve (gh #892)

### LINKED ISSUE #892: [adversary] debugger: attaching --debug-script changes a convex QCQP conic solve (Optimal -> Numerical failure, exit 1)
adversary-fingerprint: debugger/conic-attach-changes-solve-outcome

Attaching the interactive debugger to a **convex QCQP conic** solve
(`solver_selection=socp`) changes the solve itself. A `--debug-script`
containing nothing but `continue` — which should be a pure no-op on the
trajectory — turns a clean `Optimal Solution Found` into
`Numerical failure (no verified KKT point)` and **exit 1**.

The debugger is documented as "a pdb-for-the-IPM": it inspects, and a user who
attaches it expects to watch *the same* solve. Here observation changes the
result, so the debugger cannot be used to diagnose the conic path — the run you
debug is not the run you shipped.

This is **specific to the conic driver**. `lp-ipm`, `qp-ipm` and the NLP
filter-IPM are bit-stable under the same script (identical iteration counts and
objectives with and without `--debug-script`).

### Reproduction

```python
# Minimal repro: attaching the debugger changes a convex QCQP conic solve.
import pyomo.environ as pyo
P = [[8.081237476191625, -1.3357111717521042, -3.177837792578595, 4.113912068867745, -0.6167885305951004], [-1.3357111717521042, 6.166783520955787, 1.6776136382871059, -1.610501780331925, -0.031514743829699654], [-3.177837792578595, 1.6776136382871059, 10.180019811121928, -6.589542601208363, -0.860089836489867], [4.113912068867745, -1.610501780331925, -6.589542601208363, 5.887751599792348, 0.009833639085611079], [-0.6167885305951004, -0.031514743829699654, -0.860089836489867, 0.009833639085611079, 3.350517527917375]]
c = [-0.7652887415836972, 2.020688272883008, 0.16922834928772218, -0.8979265392629115, -0.9539822990406178]
F = [[1.8378244304667402, -0.5783508965433001, 1.1887040918656822, 1.42327388117323, -2.324401356345807], [1.3509748934963017, -0.20843213237772143, -1.0841285882683698, -0.5438424479925973, 0.5608957690673765], [0.3739765553621166, 0.42551413722873815, -0.7248960165356201, 1.2790838213635174, 1.502130478896611], [1.8340638101929774, 1.000462644464517, 1.894314546207076, 2.0939269544234835, 0.7042559943816945], [0.8811621128331931, 0.5822091166931507, 0.5514244021092147, 0.8632445387294106, -1.7089819015275898]]
g = [-0.323798368729351, 0.4878567515797354, -2.107331044865737, -1.4418405493566442, 1.5844982856875247]
n = 5
m = pyo.ConcreteModel(); N = range(n)
m.x = pyo.Var(N, bounds=(-10.0, 10.0), initialize=0.0)
m.obj = pyo.Objective(expr=0.5*sum(P[i][j]*m.x[i]*m.x[j] for i in N for j in N)
                           + sum(c[i]*m.x[i] for i in N))
m.qc = pyo.Constraint(expr=sum((sum(F[k][i]*m.x[i] for i in N) - g[k])**2
                               for k in N) <= 4.0)
m.write("repro.nl")
```

```sh
python3 repro.py
printf 'continue\n' > cont.pdbg

pounce repro.nl solver_selection=socp --no-sol                              # exit 0
pounce repro.nl solver_selection=socp --debug-script cont.pdbg --no-sol     # exit 1
```

### Expected

Attaching the debugger must not change the solve. The correct answer, from
**Clarabel 0.11.1** as an independent oracle on the same SOC formulation
(`tol_gap_abs = tol_gap_rel = tol_feas = 1e-12`):

```
f* = 0.46851407987426236     ||Fx-g||^2 = 3.9999999999999156  <= 4.0
```

The plain (non-debugger) POUNCE run agrees with the oracle to **1.3e-10**, so
the problem is well posed and the conic engine solves it correctly.

### Actual

```
$ pounce repro.nl solver_selection=socp --no-sol
POUNCE (convex QCQP conic IPM, pounce-convex): Optimal Solution Found.  obj=0.46851408  iters=38
exit 0

$ pounce repro.nl solver_selection=socp --debug-script cont.pdbg --no-sol
pounce-dbg> POUNCE (convex QCQP conic IPM, pounce-convex): Numerical failure (no verified KKT point).  obj=0.46855737  iters=16
exit 1
```

The debugged run is off the oracle by **4.3e-05** and honestly reports failure —
it is not silently wrong — but it fails on a problem the very same engine solves
to 1.3e-10 without the debugger. Deterministic: 5/5 runs each way.

The effect is not one unlucky instance. Four freshly drawn convex QCQPs, plain
vs. `--debug-script cont.pdbg`, all `solver_selection=socp`:

| n | plain | with debugger |
| --- | --- | --- |
| 7 | Optimal, 13 iters, obj 0.46683326 | **Numerical failure**, 34 iters, obj 0.46688090 |
| 5 | Optimal, 38 iters, obj 0.46851408 | **Numerical failure**, 16 iters, obj 0.46855737 |
| 4 | Optimal, 10 iters, obj -0.40699975 | Optimal, 11 iters |
| 9 | Optimal, 48 iters, obj -0.31201666 | Optimal, 15 iters |

The iteration count moves on **every** instance, in both directions — so this is
a trajectory change on the conic arm, not a tolerance wobble, and the two
failures are the visible tail of it.

Under `solver_selection=auto` the conic solve fails the same way and is then
**masked** by the NLP fallback, at the cost of a silent extra solve:

```
── pounce-dbg ── TERMINATED (NumericalFailure)  iter 34  obj=4.668809e-1  inf_du=9.80e-4
pounce: note: the conic (convex QCQP) solve returned no verified KKT point after 34
iterations (KKT error 9.80e-4); ... it is being re-solved on the general NLP interior-point path.
```

### Probable cause

The debugged run appears to be on a **different driver**. `docs/src/debugger.md`
says the HSDE drivers expose `tau`/`kappa` as 1-element blocks, but under the
debugger on this QCQP:

```
[source] print tau
don't know how to print `tau` (try a block name or mu|obj|inf_pr|inf_du|err|compl|iter)
[source] info
dims      = x:7 s:23 y:0 z:23
```

In `crates/pounce-convex/src/ipm.rs` the two paths diverge for symmetric cones:

* non-debug — `solve_socp_ipm` → `solve_socp_symmetric` → **`solve_qp_core`**,
  which branches on `opts.use_hsde` (default **`true`**, `ipm.rs:319`) into the
  HSDE driver plus the σ cost-normalization and the equilibrate/retry-and-verify
  logic (`ipm.rs:462`, `ipm.rs:559`);
* debug — `solve_socp_ipm_debug_scoped` → `build_factorization` + **`run_ipm`**
  directly (`ipm.rs:~1437`), which never consults `use_hsde` and so runs the
  *direct* IPM with none of that machinery.

So `--debug` silently substitutes a different algorithm on the conic arm. That
also explains the missing `tau`. Whatever the fix, the debug and non-debug conic
paths should route identically.

Per `CLAUDE.md` this is a trajectory change on the convex arm, so a fix wants
`scripts/sweep-fixtures.sh` (the convex arm is covered — 5 fixtures reach the
conic engine) rather than an objective-parity check.

### Environment

```
$ pounce --version
0.10.0
$ pip show pounce-solver | grep Version
Version: 0.10.0
$ git rev-parse --short HEAD
bc6244e3
```

Date: 2026-09-02. macOS (Darwin 25.6.0), linear solver FERAL.
Found against the released 0.10.0 wheel; the code shape above is present at
`bc6244e3`, so this is not one of the fixes already landed on `main`.

jkitchin:
Fixes #892

## Problem

`pounce model.nl --debug` on a convex QCQP ran a **different algorithm** than the same command without it. A `--debug-script` containing nothing but `continue` — a pure no-op on the trajectory — turned a clean `Optimal` into `Numerical failure (no verified KKT point)` and exit 1.

The reported 5-variable model, against Clarabel 0.11.1 at `tol = 1e-12` as an independent oracle on the same SOC formulation:

| | status | objective | iters | exit |
|---|---|---|---|---|
| before, plain | `Optimal Solution Found` | 0.46851408 | 38 | 0 |
| before, `--debug-script` | **`Numerical failure`** | 0.46851433 | 21 | **1** |
| after, either | `Optimal Solution Found` | 0.46851408 | 38 | 0 |
| oracle (Clarabel) | — | 0.46851407987426236 | — | — |

The plain run agrees with the oracle to `1.3e-10`; the debugged one failed on a problem the very same engine solves. Under `solver_selection=auto` the failure was then **masked** by the NLP fallback, at the cost of a silent extra solve.

The issue reported 4 instances. Measured over the reported model plus 24 freshly drawn convex QCQPs at n = 4, 5, 7, 9, `solver_selection=socp`, plain vs `--debug-script`:

| | disagree on status / objective / iteration count |
|---|---|
| `d32204e` | **25 / 25** — 16 turning an `Optimal` into a `Numerical failure` or a `Solved to acceptable level` |
| this branch | **0 / 25** |

So this was the normal case on that arm, not an unlucky instance. The reporter's missing `tau` block was the same fact seen from the debugger's side.

## Cause

Two independent halves, and both had to be closed.

**1. The debug entry points were parallel implementations.** `solve_socp_ipm_debug` did not call `solve_socp_ipm`; for symmetric cones it built its own factorization and called `run_ipm` directly. That path never consulted `QpOptions::use_hsde`, which defaults to `true` — so attaching a debugger substituted the *direct* IPM for the HSDE embedding, and with it dropped the `σ` cost normalization and the equilibrate-and-verify guards that live on the embedding's side of `solve_qp_core`. `tau`/`kappa` are the embedding's blocks, which is why they vanished.

The same defect was latent on the LP/QP entry point, one option away: `solve_qp_ipm_debug` honoured `use_hsde` but re-entered *below* `solve_qp_ipm_scoped`, so under `qp_hsde=no` it ran the direct driver on un-equilibrated data, and on a pure LP it skipped the crossover refinement. The issue measured `qp-ipm` as bit-stable, which it is — at default options only.

**2. The CLI skipped presolve under a debugger.** Both convex entry points carried a `debug_hook` arm ahead of the `presolve_on` one, deliberately, so the inspected blocks were the user's own rows. The price was that the debugged run solved a *different, smaller* problem. On `convex_qp_share1b.nl` presolve is 225 → 208 vars and 117 → 98 rows — not a hairline difference.

## Fix

Thread an `Option<&mut DebugHook>` down the ordinary path and reduce both debug entry points to a delegation with `Some(hook)`. The two duplicated bodies are deleted, so there is no second implementation left to drift. At the CLI, route the hook through an `ipm_solve` / `conic_solve` closure called from inside the presolve arm rather than around it.

The hook rides the *primary* solve; the recovery re-solves (equilibrated retry, `σ` re-solves, `verify_or_repair_optimum`, PSD HSDE retry, infeasibility twin, LP crossover) run unhooked, so every fallback behaves exactly as it does without a debugger.

**Rejected: gating the debug path on `use_hsde` and leaving it otherwise separate.** That is the minimal edit and it fixes the reported transcript, but it leaves two implementations of the same routing to drift apart again — which is how this shipped. It also would not have reached `σ`, the gh #414 verify, or either presolve half. The whole point is that there stop being two paths.

**A stop must not be recovered from.** Sharing the ordinary path gave the debug entry points the recovery machinery along with the routing, and a `DebugAction::Stop` leaves a *non-converged* status. The recovery reads that as something to recover from and re-solves unhooked — so after `quit` it ran the problem to completion and reported `Optimal` for a solve the user had deliberately halted. `debug_stop` is a per-solve, per-thread latch on the same rails as `crate::deadline`, set in `debug::fire` (the one place a `DebugAction` is read, so no driver can forget), consulted by every recovery gate. Unlike a deadline it never restamps a verdict — it only declines to *start* another solve. `requested()` is `false` with no debugger attached, so every gate is a no-op on the ordinary path.

This was caught by `debug.rs::stop_action_halts_the_solve` and by nothing else.

Two user-visible consequences, both documented in `docs/src/debugger.md`:

- Presolve now runs under the debugger on the convex paths; blocks are the reduced model's. `qp_presolve=no` gets the unreduced rows, and then the plain run matches because both skip it.
- `quit` stops the solve rather than reporting a recovered success.

## Blast radius

**Bit-identical except the targeted case.** `scripts/sweep-fixtures.sh` against `d32204e`: **empty diff across all 186 fixture-legs**, both `exact` and `lbfgs` — status, objective, iteration count, engine, `q=` and `2nd=` columns all unmoved. Re-run after each of the three code commits, since two of them touched `ipm.rs` after the first sweep.

That is the expected result and the right thing to demand: every non-debug call site passes `None` through otherwise-untouched code, so a moved line would have meant an unintended change. Engine split on this corpus, recounted rather than trusted: 50 `nlp`, 39 `cvx-qp`, 4 `cvx-qcqp` per leg — the conic arm the fix touches is exercised on 8 of the 186 legs.

Public API signatures are unchanged. The active-set engine is untouched: the debugger still does not engage there, and the caller still says so.

## Tests

Two files, because the two halves need different instruments.

`crates/pounce-convex/tests/issue892_debug_routes_like_the_plain_solve.rs` — six legs pinning library-level routing. Per the branch rule in `CLAUDE.md`, `solve_qp_core` forks on `use_hsde`, so each entry point gets a leg on **both** sides of that fork; `leg_qcqp_direct_driver` and `leg_qp_direct_driver` are not duplicates of their defaulted neighbours. Plus the all-orthant conic leg that reaches the gh #414 verify and the LP leg that reaches crossover (with an assertion that crossover actually pivoted to a vertex, so the leg is not vacuous).

`crates/pounce-cli/tests/issue892_debugger_does_not_change_the_solve.rs` — four checks running the binary with and without `--debug-script`, comparing status, objective, iteration count and exit code. The presolve half is invisible to the library legs, and `convex_qp_share1b.nl` / `qcqp_shared_linear_rows.nl` are chosen because presolve measurably reduces them.

**Both mutation tables were measured, not asserted:**

| mutation | red |
|---|---|
| `solve_socp_ipm_debug` builds its own iteration for symmetric cones | `leg_qcqp_hsde_default`, `leg_orthant_cone_verify`; `conic_qcqp_is_unmoved_*` |
| `solve_qp_ipm_debug` re-enters at `solve_qp_ipm_core` with `equilibrate: false` | `leg_qp_direct_driver`, `leg_lp_crossover` |
| reinstate the CLI presolve carve-out | `presolve_still_runs_under_the_debugger`, `convex_qp_is_unmoved_by_the_debugger` |

Each mutation leaves the other half's checks green, which is the argument for keeping both files. The stop regression is pinned by the pre-existing `debug.rs::stop_action_halts_the_solve`, which fails on `800b3cc` for exactly the stated reason.

**Deliberately not pinned:** the exp/power non-symmetric drivers (they already took the hook; `debug.rs::solve_socp_ipm_debug_routes_and_fires` owns them), PSD cones under the debugger (no fixture here reaches one, and the chordal decomposition is now shared with the plain path), `--debug-json`, and anything at benchmark scale.

---

- [x] Book page under `docs/src/` updated (`debugger.md`; existing page, no `SUMMARY.md` change)
- [x] `cargo fmt --all -- --check` clean; `cargo clippy --release --workspace --all-targets` introduces no new warnings (the four in `ipm.rs` at 541/1995/2665/2928 predate this branch); `cargo test --release --workspace --exclude pounce-hsl` **3996 passed, 0 failed**

`pounce-hsl` is excluded because its MA57 tests need the proprietary HSL library, which this environment does not have; that exclusion is environmental and unrelated to the diff.

https://claude.ai/code/session_01JmfvDmzT6i5FZ5QfJPbsRD

---
_Generated by [Claude Code](https://claude.ai/code/session_01JmfvDmzT6i5FZ5QfJPbsRD)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -356,6 +356,50 @@ changes.
 
 ### Fixed
 
+- **Attaching the interactive debugger no longer changes the solve
+  (gh#892).** `pounce model.nl --debug` on a convex QCQP ran a *different
+  algorithm* than the same command without it. The conic debug entry point
+  built its own interior-point iteration and never consulted `qp_hsde`, so
+  the debugger silently substituted the direct IPM for the default HSDE
+  embedding and dropped the objective normalization and the
+  verify-and-retry guards that live on the embedding's side. On the
+  reported 5-variable model:
+
+  ```text
+  $ pounce repro.nl solver_selection=socp --no-sol
+  POUNCE (convex QCQP conic IPM): Optimal Solution Found.  obj=0.46851408  iters=38   # exit 0
+
+  $ pounce repro.nl solver_selection=socp --debug-script cont.pdbg --no-sol
+  POUNCE (convex QCQP conic IPM): Numerical failure (no verified KKT point). iters=21 # exit 1
+  ```
+
+  where the script contained nothing but `continue`. The plain run agrees
+  with Clarabel 0.11.1 to `1.3e-10`; the debugged one failed on a problem
+  the very same engine solves. Under `solver_selection=auto` the failure
+  was then masked by a silent extra solve on the NLP path. Measured over
+  the reported model plus 24 freshly drawn convex QCQPs, **all 25**
+  disagreed on status, objective or iteration count, 16 of them turning an
+  `Optimal` into a failure or a reduced-accuracy exit — so this was the
+  normal case on that arm, not an unlucky instance. The missing `tau` block
+  users saw under `print tau` was the same fact from the debugger's side.
+
+  Both convex debug entry points are now the ordinary solve entry points
+  reached with a hook rather than parallel implementations, so driver
+  selection, scaling and every recovery guard are the ones the plain run
+  uses. Two consequences worth knowing:
+
+  - **Presolve now runs under the debugger**, on the convex paths where it
+    used to be skipped. The blocks you inspect are the reduced model's —
+    which is the model being solved. Pass `qp_presolve=no` to step your own
+    rows instead, and the plain run then matches, because both skip it.
+  - **`quit` stops the solve.** The recovery machinery that re-solves after
+    a failed or capped solve runs unhooked, so it now declines to start
+    once you have halted the run rather than reporting success for a solve
+    you deliberately stopped.
+
+  Solves without a debugger attached are bit-identical: the fixture corpus
+  sweep is empty across all 186 legs.
+
 - **A settled iterate with a runaway multiplier no longer ships as
   `Solved_To_Acceptable_Level` (gh#884).** Some models have a solution at
   which a constraint's gradient vanishes — the row is satisfied, the primal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -398,7 +398,11 @@ changes.
     you deliberately stopped.
 
   Solves without a debugger attached are bit-identical: the fixture corpus
-  sweep is empty across all 186 legs.
+  sweep is empty across all 186 legs. That sweep is the check for *collateral*
+  damage — only 4 of its fixtures (8 legs) reach the conic arm this touches, so
+  what carries the claim about that arm is the 25-instance measurement above
+  and the CLI checks on `qcqp_ball` / `qcqp_columns_illcond` under both `socp`
+  and `auto`, not the empty diff.
 
 - **A settled iterate with a runaway multiplier no longer ships as
   `Solved_To_Acceptable_Level` (gh#884).** Some models have a solution at
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -391,7 +391,12 @@ changes.
   - **Presolve now runs under the debugger**, on the convex paths where it
     used to be skipped. The blocks you inspect are the reduced model's —
     which is the model being solved. Pass `qp_presolve=no` to step your own
-    rows instead, and the plain run then matches, because both skip it.
+    rows instead, and the plain run then matches, because both skip it. When
+    presolve *settles* the model there is no iteration left to step at all,
+    so an unbounded presolve verdict now prints `Presolve: proved unbounded
+    below — a free column with a nonzero objective coefficient`, the sibling
+    of the primal-infeasible line gh#523 added; it was the one presolve
+    verdict that arrived with no record of why.
   - **`quit` stops the solve.** The recovery machinery that re-solves after
     a failed or capped solve runs unhooked, so it now declines to start
     once you have halted the run rather than reporting success for a solve
--- a/crates/pounce-cli/src/main.rs
+++ b/crates/pounce-cli/src/main.rs
@@ -2723,6 +2723,31 @@ fn run_convex_qp(
         )
     };
     let solve_opts = || solve_opts_offset(0.0);
+    // The interior-point solve, with the interactive debugger attached when
+    // there is one (gh #892).
+    //
+    // Routing the hook through here rather than through an arm of its own is
+    // what keeps presolve on under the debugger. It used to be skipped, so
+    // that the inspected `x`/`s`/`y`/`z` blocks were the user's rows rather
+    // than a reduced set — but the price was that the debugged run solved a
+    // *different, smaller* problem than the plain one, which is the same
+    // "attaching the debugger changes the solve" defect gh #892 is about, one
+    // level up from the driver substitution. A user who wants the unreduced
+    // blocks asks for them with `qp_presolve=no`, and then both runs agree
+    // again because both skip it.
+    //
+    // Guarded on `!use_active_set`: the debugger hooks barrier-IPM iterations
+    // and has no active-set analogue, so under that engine it must not engage
+    // — the caller has already printed the note saying so.
+    let ipm_solve = |p: &pounce_convex::QpProblem, o: &QpOptions| -> pounce_convex::QpSolution {
+        match debug_hook.filter(|_| !use_active_set) {
+            Some(hook) => {
+                let mut h = hook.borrow_mut();
+                solve_qp_ipm_debug(p, o, &mut *h, backend)
+            }
+            None => solve_qp_ipm(p, o, backend),
+        }
+    };
     // What presolve did, held back until we know this solve is the one that
     // reports (gh #535). These lines describe the reduction, not the verdict,
     // but they are the *only* stdout a declined convex attempt would otherwise
@@ -2738,19 +2763,6 @@ fn run_convex_qp(
         // enforce the zero-iteration stop here before any solve runs
         // (pounce#186). Mirrors the NLP path's MaximumIterationsExceeded.
         trivial(QpStatus::IterationLimit)
-    } else if let Some(hook) = debug_hook.filter(|_| !use_active_set) {
-        // Interactive debug: step the IPM on the extracted QP directly.
-        // Presolve is skipped so the debugger's `x`/`s`/`y`/`z` blocks
-        // correspond to the user's problem rather than a reduced one.
-        //
-        // Guarded on `!use_active_set`: the debugger hooks barrier-IPM
-        // iterations and has no active-set analogue, so on that engine this
-        // arm would quietly solve with a *different solver* than the one the
-        // user selected. The caller has already printed the note explaining
-        // the debugger does not engage; fall through and solve normally.
-        let mut h = hook.borrow_mut();
-        let _t = timing.solve.guard();
-        solve_qp_ipm_debug(&qp, &solve_opts(), &mut *h, backend)
     } else if presolve_on {
         let outcome = {
             let _t = timing.presolve.guard();
@@ -2817,7 +2829,7 @@ fn run_convex_qp(
                             &mut mk,
                         )
                     } else {
-                        solve_qp_ipm(&ps.reduced, &solve_opts_offset(ps.obj_offset()), backend)
+                        ipm_solve(&ps.reduced, &solve_opts_offset(ps.obj_offset()))
                     }
                 };
                 // The postsolve lift is presolve's other half, so it is
@@ -2840,7 +2852,7 @@ fn run_convex_qp(
         solve_qp_active_set_inertia(&qp, &solve_opts(), &engine_overrides, inertia, &mut mk)
     } else {
         let _t = timing.solve.guard();
-        solve_qp_ipm(&qp, &solve_opts(), backend)
+        ipm_solve(&qp, &solve_opts())
     };
     let elapsed = t0.elapsed().as_secs_f64();
 
@@ -3174,19 +3186,29 @@ fn run_convex_socp(
     // exactly as `run_convex_qp` does: these lines describe the reduction,
     // not the verdict, and a declined conic attempt must leave no stdout.
     let mut presolve_log: Vec<String> = Vec::new();
+    // The conic solve, with the interactive debugger attached when there is
+    // one (gh #892). See the twin in `run_convex_qp` for why the hook goes
+    // through here instead of an arm of its own: it is what keeps presolve on
+    // under the debugger, so the debugged run is the same problem the plain
+    // run solves. `qp_presolve=no` is how you ask for the unreduced blocks.
+    let conic_solve = |p: &pounce_convex::QpProblem,
+                       k: &[pounce_convex::ConeSpec],
+                       o: &QpOptions|
+     -> pounce_convex::QpSolution {
+        match debug_hook {
+            Some(hook) => {
+                let mut h = hook.borrow_mut();
+                solve_socp_ipm_debug(p, k, o, &mut *h, backend)
+            }
+            None => solve_socp_ipm(p, k, o, backend),
+        }
+    };
     let sol = if qp_opts.max_iter == 0 {
         // `max_iter=0` cannot reach optimality — stop before any solve, the
         // same zero-iteration contract the QP path enforces (pounce#186).
         // Above the presolve arm for the same reason it is on the QP path:
         // presolve can otherwise settle a trivial problem without iterating.
         trivial(pounce_convex::QpStatus::IterationLimit)
-    } else if let Some(hook) = debug_hook {
-        // Interactive debug steps the conic IPM on the *extracted* problem,
-        // so the debugger's blocks correspond to the user's rows rather than
-        // a reduced set. Same carve-out as the QP path.
-        let mut h = hook.borrow_mut();
-        let _t = timing.solve.guard();
-        solve_socp_ipm_debug(&qp, &cones, &solve_opts(), &mut *h, backend)
     } else if presolve_on {
         // **`presolve_conic`, never `presolve`.** The orthant entry point
         // would hand `dedup_rows` an unprotected row list, and this driver's
@@ -3241,7 +3263,7 @@ fn run_convex_socp(
                 let red_cones = ps.reduced_cones(&cones);
                 let red = {
                     let _t = timing.solve.guard();
-                    solve_socp_ipm(&ps.reduced, &red_cones, &solve_opts(), backend)
+                    conic_solve(&ps.reduced, &red_cones, &solve_opts())
                 };
                 // The postsolve lift is presolve's other half.
                 let _t = timing.presolve.guard();
@@ -3255,7 +3277,7 @@ fn run_convex_socp(
         }
     } else {
         let _t = timing.solve.guard();
-        solve_socp_ipm(&qp, &cones, &solve_opts(), backend)
+        conic_solve(&qp, &cones, &solve_opts())
     };
     let elapsed = t0.elapsed().as_secs_f64();
 
--- a/crates/pounce-cli/src/main.rs
+++ b/crates/pounce-cli/src/main.rs
@@ -2844,7 +2844,21 @@ fn run_convex_qp(
                 presolve_log.push(format!("Presolve: proved primal infeasible — {trigger}"));
                 trivial(QpStatus::PrimalInfeasible)
             }
-            PresolveOutcome::Unbounded => trivial(QpStatus::DualInfeasible),
+            PresolveOutcome::Unbounded => {
+                // The sibling of the line above, and it exists for the same
+                // reason (gh #523, gh #892 re-review): a presolve verdict
+                // arrives with no iteration behind it, so this line is the
+                // whole record of *why*. `Unbounded` carries no trigger
+                // payload because it has exactly one cause — a free column
+                // with a nonzero objective coefficient — so the cause is
+                // named here instead.
+                presolve_log.push(
+                    "Presolve: proved unbounded below — a free column with a \
+                     nonzero objective coefficient"
+                        .to_string(),
+                );
+                trivial(QpStatus::DualInfeasible)
+            }
         }
     } else if use_active_set {
         let mut mk = backend;
@@ -3273,7 +3287,15 @@ fn run_convex_socp(
                 presolve_log.push(format!("Presolve: proved primal infeasible — {trigger}"));
                 trivial(pounce_convex::QpStatus::PrimalInfeasible)
             }
-            PresolveOutcome::Unbounded => trivial(pounce_convex::QpStatus::DualInfeasible),
+            PresolveOutcome::Unbounded => {
+                // See the twin in `run_convex_qp`.
+                presolve_log.push(
+                    "Presolve: proved unbounded below — a free column with a \
+                     nonzero objective coefficient"
+                        .to_string(),
+                );
+                trivial(pounce_convex::QpStatus::DualInfeasible)
+            }
         }
     } else {
         let _t = timing.solve.guard();
--- a/crates/pounce-cli/tests/issue892_debugger_does_not_change_the_solve.rs
+++ b/crates/pounce-cli/tests/issue892_debugger_does_not_change_the_solve.rs
@@ -39,6 +39,7 @@
 use std::io::Write;
 use std::path::PathBuf;
 use std::process::Command;
+use std::sync::atomic::{AtomicUsize, Ordering};
 
 fn pounce_exe() -> PathBuf {
     PathBuf::from(env!("CARGO_BIN_EXE_pounce"))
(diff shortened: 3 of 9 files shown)


jkitchin:
## Review — head `716991d3`

Read the full delta against `origin/main` (9 files, +961/−204), plus the call
graph around the six public convex entry points, the `DebugAction` funnel, and
both test files. CI 8/8 green; branch contains the `main` tip.

**This is a good PR.** The diagnosis is right, the fix is the structural one
rather than the minimal one, and the "Rejected:" paragraph gives the correct
reason for preferring it. Three things I verified rather than took on trust:

- **`debug::fire` really is the single funnel.** Every `DebugAction::Stop`
  comparison in `pounce-convex` — 3 in `hsde.rs`, 3 in `hsde_nonsym.rs`, 3 in
  `ipm.rs` — goes through `fire`, so `debug_stop::mark()` cannot be forgotten by
  a driver. `pounce-algorithm`'s NLP path has its own `fire_debug` and is
  correctly untouched.
- **`debug_stop` is structurally sound.** `mark()` no-ops outside a scope,
  `with_scope` reuses a nested scope rather than nesting latches, RAII reset to
  `None`. Same shape as `deadline`, and the same shape as PR #890's
  `sigma_verdict` — the discipline is right in both.
- **The gate set looks complete.** Nine `requested()` sites, and each sits on a
  recovery that starts a *new* solve: crossover (`:399`), the gh#293
  equilibrated retry (`:500`), the gh#414 verify (`:581`), the recession
  reverify (`:606`), the infeasibility twin (`:656`), the two conic retries
  (`:1386`, `:1408`), and the σ cascade (`:3146`). The σ gate is placed above
  the un-normalized re-solve, so a stop returns the halted iterate before any
  cascade candidate is built — the right place.

The library legs are the strongest part. `assert_same_solve` compares status,
iteration count, objective and every `x` entry — the iteration count is the
field that matters and it is there. Every leg asserts `hook.fired > 0`, so none
can pass vacuously, and `saw_tau` is asserted in **both** directions across the
`use_hsde` fork, which is a real driver discriminator rather than a proxy for
one. `leg_lp_crossover` asserts the answer is actually a vertex. That is
CLAUDE.md's branch rule applied properly — `leg_qcqp_direct_driver` and
`leg_qp_direct_driver` genuinely are not duplicates.

Four things below. F1 is three lines and I would fix it before merge; the rest
are a missing assertion, a documentation gap, and a heads-up about #890.

---

### F1 — RISK. The CLI test file races against itself; it is flaky by construction

`continue_script()` builds its path from `std::env::temp_dir()` and
`std::process::id()` alone. The test harness runs the four `#[test]` functions
as parallel threads of **one** process, so all nine `assert_unmoved` calls share
one filename — and `assert_unmoved` ends with `std::fs::remove_file(&script)`.

The interleaving is ordinary, not exotic:

1. `conic_qcqp_is_unmoved_by_the_debugger` creates the script and starts its
   plain run;
2. `convex_qp_is_unmoved_by_the_debugger` creates the same path (benign — same
   bytes) and starts its own;
3. the first finishes an `assert_unmoved` and **deletes** the file;
4. the second spawns `pounce … --debug-script <now-missing path>`.

The concurrent *write* is harmless because the content is identical; the
`remove_file` is not. Green CI is not evidence against this — it is a race, and
nine calls across four threads is a lot of chances. Add a per-call
discriminator (an `AtomicUsize` counter, or `format!("{:?}", thread::current().id())`),
or create the script once in a `OnceLock` and never delete it.

### F2 — NIT. `assert_same_solve` compares `x` with `zip`, so a length change passes silently

```rust
for (i, (a, b)) in plain.x.iter().zip(&debugged.x).enumerate() {
```

`zip` stops at the shorter side. If a routing difference ever returned a
different-length `x` — a different bound expansion, a reduced model, a cone
block that did or did not get split — the loop compares the common prefix and
the leg goes green. That is exactly the failure mode the file exists to catch,
and it is the one field where the instrument has a blind spot. One line:

```rust
assert_eq!(plain.x.len(), debugged.x.len(), "{leg}: x length moved under the debugger");
```

### F3 — RISK (docs). With presolve on, some models now get *no debugger at all*, and the page does not say so

`PresolveOutcome::Infeasible` and `PresolveOutcome::Unbounded`
(`main.rs:2837`, `:2846`) return `trivial(...)` without ever calling
`ipm_solve`. Under the old carve-out the debug arm sat *above* `presolve_on`, so
`--debug` on such a model always stepped the IPM to its own verdict. Now the
debug script never runs and not one checkpoint fires.

That is *correct* by this PR's own standard — it is what the plain run does, and
the whole thesis is that the two must agree. But the person most likely to
attach a debugger to a model presolve proves infeasible is precisely the person
trying to find out why it is infeasible, and they will get a silent no-op. The
new `docs/src/debugger.md` paragraph covers the reduced-blocks case and points
at `qp_presolve=no`; it does not cover this one, which is the more surprising
of the two. One sentence in the same paragraph would do it.

(The `Reduced` path is fine — the guard, the timing arm and the postsolve lift
are all intact, and `verdict()` captures `iters=`, so the mutation row for
`presolve_still_runs_under_the_debugger` has real content: reinstating the
carve-out makes the debugged run solve 208 vars where the plain one solves 225,
and the iteration count moves.)

### F4 — Heads-up: this and #890 both rewrite the same twenty lines of `ipm.rs`

Not a finding against this PR — both are open, and whoever merges second should
know:

- **Textual conflict is certain.** #890 inserts `sigma_verdict::tracking()` and
  `demote_uncertified_sigma_optimum` into the body this PR just split into
  `solve_qp_ipm` / `solve_qp_ipm_scoped`, immediately around the crossover call
  this PR wrapped in a `requested()` gate.
- **The semantics compose, and in #890's favour.** #890's `tracking()` frame
  lands in `solve_qp_ipm_scoped`, which this PR makes `solve_qp_ipm_debug`
  share — so the debug entry point picks up the σ demotion for free. That
  closes part of what I raised as F1 on #890. `solve_qp_ipm_warm` (and with it
  the batch API and warm-started Python `solve_qp`) still bypasses that frame
  after both land.
- **One invariant to preserve on merge:** the σ gate at `ipm.rs:3146` returns
  early on `requested()`, *above* where #890 adds its `record()` call in the
  candidates branch. Keep that ordering — otherwise a debugger stop inside the
  cascade would record a σ verdict about a solve the user halted, and #890's
  demotion would then relabel it. As the two branches stand today the ordering
  is right; it is just easy to lose in a conflict resolution.

---

## Verdict

The refactor is the right shape — deleting the parallel implementations is what
makes the property hold rather than what makes the reported transcript pass —
and the stop latch is a real consequence that was found by a test rather than
by reasoning, which is the good sign. The mutation tables are specific enough to
check, and the legs are non-vacuous where it counts.

**F1 before merge** (three lines, and it will bite someone otherwise). F2 is one
line. F3 is one sentence of docs. F4 is coordination, not a change.

One scope note on the evidence, not a criticism: the empty 186-leg sweep is the
right check for *collateral* damage, but by your own recount the arm this PR
actually changes is 4 `cvx-qcqp` fixtures — 8 legs. What carries the claim about
that arm is the 25-instance QCQP measurement and the two CLI checks on
`qcqp_ball` / `qcqp_columns_illcond` under both `socp` and `auto`, and I read
the body as saying that. Worth keeping the distinction explicit in the
CHANGELOG for the same reason CLAUDE.md keeps it for the σ path.

Method note: this is a reading review. I read the diff, the six entry points'
call graph, every `requested()` and `DebugAction::Stop` site, both test files
and the docs page. I did not run the fixture sweep or the 25-instance QCQP
measurement, so those numbers are yours; F1 is derived from the test harness's
threading model and the file's own `remove_file`, not from an observed failure.

jkitchin:
All three actionable findings are fixed in `4f820e2`. F4 left as coordination.

### F1 — fixed, and you were right that CI proves nothing here

Confirmed by reading rather than by an observed failure, same as you: `continue_script()` keyed the path on `std::process::id()` alone, the four `#[test]` functions are threads of one process, and `assert_unmoved` ends in `remove_file` — nine calls sharing one path. Your step 3→4 is the window.

The name now carries an `AtomicUsize` per-call counter alongside the pid. I kept the deletion rather than taking the `OnceLock`-and-leak option, since the counter closes the window without leaving a file behind per run. The comment records *why* the counter is there, so it does not get simplified back into the bug.

### F2 — fixed

One line, before the `zip`:

```rust
assert_eq!(plain.x.len(), debugged.x.len(), "{leg}: x length moved under the debugger");
```

Worth saying you picked the sharpest one: a shape change in `x` is precisely a routing difference, so the one field where the instrument was blind was the field the file exists to watch.

### F3 — fixed, and measured

I checked it rather than reasoning about it, on `infeasible_qp.nl`:

| | checkpoints fired |
|---|---|
| default (presolve on), `--debug-script` | **0** |
| `qp_presolve=no`, `--debug-script` | 1 |

So the no-op is real and the escape hatch does put the iteration back. `docs/src/debugger.md` now carries a paragraph in the same block: presolve *settling* the model leaves no solve to step, that is the plain run reproduced faithfully, and it is a silent no-op exactly when someone is trying to find out why their model is infeasible — with the `Presolve: proved primal infeasible — <trigger>` line named as what you get instead, and `qp_presolve=no` as the way to step it.

### F4 — noted, no change

Agreed on all three points, including that #890's `tracking()` frame landing in `solve_qp_ipm_scoped` means the debug entry point picks up the σ demotion for free, and that `solve_qp_ipm_warm` still bypasses it after both land. I have left the σ gate at `ipm.rs:3146` where it is — above where #890 adds its `record()` — and flagged the ordering for whoever resolves the conflict. Happy to be the one who does, in whichever order you merge them.

### Scope note — taken

Fair, and it is the same distinction CLAUDE.md insists on for the σ path. The CHANGELOG now says the 186-leg sweep is the check for *collateral* damage, that only 4 of those fixtures (8 legs) reach the conic arm this touches, and that what carries the claim about that arm is the 25-instance measurement plus the CLI checks on `qcqp_ball` / `qcqp_columns_illcond` under both `socp` and `auto`.

---

Validation on `4f820e2`: `cargo fmt --all -- --check` clean; both affected test files pass (6 legs + 4 checks); the broader `pounce-convex` + `pounce-cli` release run was still going when I pushed, and I will report if it turns anything up rather than leaving it implied.

---

jkitchin:
## Re-review of `4f820e2` (delta only)

All three actionable findings are fixed, and each is the right fix rather than
the fix that makes the symptom go away. Checked individually.

**F1.** `AtomicUsize` counter appended to the pid, `static` inside
`continue_script()`. `fetch_add` is atomic under any ordering, so `Relaxed` is
the correct choice here — the only requirement is that no two calls get the same
integer, not that anything be ordered against it. Keeping the `remove_file` is
the better half of the trade: the window is closed by the name, so there is no
reason to leak a file per run as well. The comment says *why* the counter is
there, which is what stops it being simplified back into the bug.

**F2.** Length assertion placed before the `zip`, which is the only placement
that works. Ran the file locally at this commit: 6/6 legs pass with it in.

**F3.** Verified rather than read — the paragraph quotes a string and an option,
and both had to exist:

- `Presolve: proved primal infeasible — {trigger}` is verbatim at
  `crates/pounce-cli/src/main.rs:2844`.
- `qp_presolve` is a genuinely registered library option
  (`pounce-algorithm/src/upstream_options.rs:391`,
  `pounce-convex/src/options.rs:153`), not just a name in a comment.

So the escape hatch and the diagnostic are both real. Your 0-vs-1 checkpoint
measurement matches the mechanism I read: `Infeasible` and `Unbounded` both
short-circuit with `trivial(...)` before `ipm_solve` is ever reached.

### One gap the new paragraph opens — small, and pre-existing underneath

The paragraph covers presolve settling the model "by proving it
primal-infeasible **or unbounded** outright", then points the reader at the
`Presolve: proved primal infeasible` line as what they get instead of an
iteration. That line only exists on one of those two branches:

```rust
PresolveOutcome::Infeasible(trigger) => {
    presolve_log.push(format!("Presolve: proved primal infeasible — {trigger}"));
    trivial(QpStatus::PrimalInfeasible)
}
PresolveOutcome::Unbounded => trivial(QpStatus::DualInfeasible),   // :2847
```

`Unbounded` pushes nothing — same at the twin site, `:3276`. So on that branch a
user gets zero checkpoints **and** no line saying why, which is the strictly
worse half of the case the paragraph was written for, and the paragraph sends
them looking for a diagnostic that will not be there.

That missing line is not this PR's doing — gh#523 added it for the infeasible
branch only, and nothing has changed about `Unbounded` here. But the docs now
promise it for both. Two ways out, and the second is one line of code:

1. Narrow the sentence to the primal-infeasible branch and say plainly that an
   unbounded presolve verdict arrives with no record at all.
2. Add the matching `Presolve: proved dual infeasible` push, which is what
   gh#523's own reasoning ("a presolve infeasibility arrives with no iteration
   behind it, so this line is the whole record of *why*") already implies for
   the sibling outcome.

Either is fine; leaving it as-is is the one option I would not take, since the
whole point of the paragraph is to keep someone from hunting for a debugger
that never fires.

### CHANGELOG scope note

Taken as written, and it now says the right thing: the empty 186-leg sweep is
the collateral-damage check, 8 of those legs reach the conic arm, and the
25-instance measurement plus the two CLI fixtures under `socp` and `auto` are
what carry the claim about the arm that actually changed.

### Gates

CI 8/8 SUCCESS at `4f820e2`, `mergeState: CLEAN`. Locally,
`issue892_debug_routes_like_the_plain_solve` is 6/6 green with the new
assertion.

No blockers left from my side. The only open item is the doc sentence above,
and F4 stays coordination — the `ipm.rs` conflict with #890 is still certain,
and the σ gate at `:3146` is still correctly above where #890 adds its
`record()`.

jkitchin:
Took **option 2** — the one line of code — in `98a7423`.

Narrowing the sentence would have left the strictly-worse branch exactly as you describe it: no checkpoints *and* no record. gh#523's rationale is that a presolve verdict arrives with no iteration behind it so the line is the whole record of why, and that reads the same for `Unbounded`. It has no `InfeasibleTrigger` payload because it has exactly one cause — a free column with a nonzero objective coefficient (`presolve.rs:1866`, `:1871`) — so the cause goes in the line itself:

```
Presolve: proved unbounded below — a free column with a nonzero objective coefficient
```

Both sites, `run_convex_qp` (`:2847`) and `run_convex_socp` (`:3276`). Verified firing rather than assumed, on a model whose only free column appears in the objective:

```
$ pounce presolve_unbounded.nl --no-sol
Presolve: proved unbounded below — a free column with a nonzero objective coefficient
POUNCE (convex QP IPM, pounce-convex): Problem is unbounded (dual infeasible).  obj=0.00000000  iters=0
Status: Diverging_Iterates
```

The docs paragraph now names both lines instead of promising one for both branches, and the CHANGELOG records the new line as the sibling of gh#523's — it is a user-visible stdout addition, not something the debugger fix implies, so it earns its own sentence.

A new stdout line is the classic way to break a test that asserts exact output, so I ran the affected crates before pushing rather than after: `cargo test --release -p pounce-cli -p pounce-convex` → **1309 passed, 0 failed** across 170 binaries. `cargo fmt --all -- --check` clean.

Thanks for both reviews — F1 and this one are the same kind of catch, a claim whose evidence didn't actually cover it, and neither would have shown up in a green CI run.

---

PR #897, opened 2026-09-02
fix(884): the promotion gate ranked certificates and decided points

jkitchin:
Refs #884, #887, #880, #888.

No closing keyword: none of these is a filed issue. All four defects were found by adversarial testing of the last few merged PRs (#885, #888, #890, #894) and are fixed here rather than filed, per the repo's own rule.

## 1. The gh#884 promotion gate ranked certificates and decided points

`run_with_dual_divergence_retry` promoted when the retry returned `Solve_Succeeded`, its unscaled KKT error and constraint violation were within `acceptable_tol`, and its unscaled KKT error strictly beat the base attempt's. Every one of those is a statement about the **certificate**. Two places argued that this was nevertheless enough to bound which **point** came back:

> that deferral exists because the μ flip returns a different *local solution* … This retry cannot: conjunct 4 requires the promoted answer to satisfy the KKT conditions in the model's own units.

Any other KKT point satisfies the KKT conditions in the model's own units too. The inference does not hold, and it was load-bearing.

**Measured**, 400 random QPECs under the `prod_eq` lowering at `bound_relax_factor=0 mu_strategy_fallback=no tol=1e-8`: **68 promotions, 42 of which moved the objective materially** — a different local solution — and **three of which handed back a strictly worse feasible point**, worst case `-13.0057 → -1.2072`, both independently `pounce verify`-feasible.

### The part that makes this a correctness fix

MacMPEC's **`scholtes4`** — already in `benchmarks/mpcc/cases.py` with a derived optimum:

```
min x1 + x2 - x3   s.t.  -4 x1 + x3 <= 0,  -4 x2 + x3 <= 0,  0 <= x1 _|_ x2 >= 0
```

`f* = 0` **exactly**, and for the smooth lowering as much as for the MPCC: `x1 x2 = 0` puts one of the pair at zero, which forces `x3 <= 0`, so `f = x1 - x3 >= 0`. At `bound_relax_factor=0`, from all three of the case's starts:

| retry | objective | constraint violation | status |
|---|---:|---:|---|
| on (shipped) | **`-6.6088e-05`** | `1.09e-09` | **`Optimal Solution Found`** |
| off (`dual_divergence_retry_step_tol=0`) | `+1.8176e-09` | `2.07e-25` | `Solved To Acceptable Level` |

A `1.09e-09` violation of the complementarity row buys `6.6e-05` of objective — a 4.8-order amplification, which is what a degenerate MPCC's unbounded multipliers are. Both points independently `VERIFIED` feasible by `pounce verify`; the promoted one is simply below anything the model reaches.

**This is the failure #885's own safety argument names and claims to exclude.** Ruled-out 3 measured exactly it on `ralph1` (`perturb_always_cd=yes` reaching `f = -2.71e-5` below `f* = 0`, beating its base attempt on every number the gate reads) and concluded, correctly, that the detector is the barrier. `the_detector_must_not_fire_on_ralph1` is mutation-checked and `ralph1` is still correctly held out. But the barrier was fitted to **one member** of the class it has to exclude — `qpec_small` settles its step at `4.3e-8`, `ralph1` bottoms out at `7.2e-3`, and `1e-5` was placed between them. `scholtes4` has the same pathology (MPCC-LICQ fails, via the rows rather than the bound; no S-stationary point exists) and its step settles, so it lands on the wrong side. A threshold validated against one member of a class is not evidence about the class — CLAUDE.md's own branch rule, one level up.

### The fix

Two conjuncts in `retry_answer_is_admissible`, both about the *answer*, both skipped when the base attempt is not itself feasible within `acceptable_tol` (there is then no admissible point to protect):

1. the retry may not return a **strictly worse objective**;
2. an objective **improvement may not be bought with primal slack** — a better objective at a larger constraint violation is refused.

Rule 1 catches the three worse-objective promotions; rule 2 catches `scholtes4`, which rule 1 cannot see because its objective got *better*. Neither is a magnitude estimator: rule 2 asks only whether the primal was given up, which is the right question because the remedy is for a *dual* defect and the detector's own first conjunct asserts the primal has settled.

| | before | after |
|---|---:|---:|
| detector fired | 150 | 150 |
| promotions | 68 | 23 |
| ⤷ objective moved materially | **42** | 1 |
| ⤷ objective strictly **worse** | **3** | **0** |

The detector is untouched — this is entirely a change to what the gate accepts. The one surviving objective move (seed 15, `-0.23335 → -3.7499`) *improves* the constraint violation as well (`1.14e-11 → 2.32e-13`), which is what rule 2 is for: it refuses purchases, not improvements.

**The tolerance is not fitted.** `acceptable_tol · max(1, |base objective|)`. The smallest move that must be **admitted** is `qpec_small`'s own `5.8e-11` — its promotion moves the objective *worse* by that much in exchange for nine orders of unscaled dual residual — and the smallest that must be **refused** is `0.198`. Four and five orders either side of `1e-6`.

## 2. The dominance gate had no floor

`runaway_is_the_whole_residual` tested a *ratio* only. A ratio is scale-free and therefore cannot say the residual is large: an answer converged to `1e-30` primal with an unscaled dual residual of **`4.4e-01`** passes it as comfortably as gh#884's `7.9e+04` does, and `4.4e-01` is not a runaway under any reading of that issue. The doc said "the entire residual is dual infeasibility"; the code said "the largest third of it is".

It now also clears the detector's own `dual_divergence_retry_du_floor`, so the answer-level gate asks about the same magnitude the iterate-level one did rather than being a strictly looser copy of it. That conjunct alone removes 7 of the 68 promotions.

## 3. A declined retry reported two different answers

`set_on_converged` fires once per **attempt**, and the three-sink floor (#870) does not reach it. So when a retry converged and was then refused, `solution.x`, the `.sol` and the `ipopt_zL_out`/`ipopt_zU_out` suffixes still carried the **discarded** attempt's point while `status`, `objective` and every statistic beside them had been floored back to the attempt that won. One run, two answers — measured on a declined retry, the `.sol` held `f = -6.3274` and the JSON report beside it said `-6.1768`, with `pounce verify` on the `.sol` confirming which one the file held.

Present on the μ fallback's floor too, for the same reason and through the same code. Both are fixed.

The fix is not to re-run the callback — the converged KKT state it borrows belongs to the retry by then and cannot be rewound — but to tell the caller which payload to trust. `IpoptApplication::answer_restored_from_floor()` reports that a floor was replayed, and the CLI re-reads the point and the bound multipliers from the last `finalize_solution` payload. That payload is always the answer being reported, on this path and the ordinary one, because the floor restores the winner *by* calling `finalize_solution`; both sources apply the same `finalize_solution_z_l`/`_z_u` lift, so these are the same numbers rather than merely compatible ones.

## 4. A σ demotion could discard a crossover-purified exact vertex (#880 × #888)

`solve_qp_core` records the σ verdict about the candidate the cascade returns, and `maybe_crossover` can then replace that interior iterate with an **exact vertex** without re-entering the solver. `sigma_verdict`'s module doc anticipated this and asked a future consumer sensitive to the distinction to re-record. #888 created one a merge earlier, and it is not a label: the CLI reroutes `ProblemClass::Lp` on `OptimalInaccurate`, so the stale verdict *throws the vertex away* and re-solves on the NLP arm. The verdict is now re-asked of the point actually being returned, through the same estimator with the same cut — a measurement, not a blanket clear.

**Reachability, measured** on an instrumented build (probes on `hsde_cost_scale`, `sigma_verdict::record` and the crossover gate) rather than inferred from an absence — an earlier revision of this PR claimed "0 of 150 purpose-built degenerate LPs reach σ", which was wrong: σ engaged on **149 of those 150** and the cascade certified every one. The corpus was reaching the path the whole time. (The companion "1 of 79 CLI fixtures" figure was quoted from CLAUDE.md rather than measured, and refers to a QCQP, while crossover is gated to pure LPs — so it cannot bear on this path at all.)

Over 550 pure LPs — 400 newly built with no equality rows and the worst conditioning I could construct, plus the original 150:

| | |
|---|---|
| crossover on by default | **no** — `qp_crossover=yes` opts in |
| crossover *moves* `x` when enabled | **331 of 550** |
| σ engages | 336 of 550; 2 of 96 CLI fixtures |
| cascade records `uncertified` on a pure LP | **0 of 550** |

The two halves are not equally rare: the crossover half is the **majority** case once the option is on, and the uncertified half is the one that resisted. A structural reason for much of that — `sigma_forward_error_is_small` returns early on `m_eq > 0` and is the only arm recorded, so a model with **any equality row can never be demoted**, which is most of why a netlib-shaped LP never is.

**The fix's effect is measured**, by forcing the precondition and comparing the same binary with and without the re-record. On the 182 LPs where crossover moves `x`:

- the stale verdict reroutes **182 of 182** to the NLP arm;
- scored against HiGHS on the 167 that solve, keeping the crossover vertex is **exact** (median relative error `0`, **167/167 within `1e-8`**), while the reroute lands **>10× further out on 43** of them, losing five to eight significant digits — `h27`: `-4.693742165` against the exact `-4.693769437`.

So: effect measured, trigger unreproduced. That is the honest split, and it is why the helper also carries unit tests on all three of its branches.

## Evidence

**Fixture sweep: 0 moved lines across all 188 pre-existing fixture-legs.** That is the expected result and not evidence about any of the above — no fixture in the corpus reaches the retry at default options, which #885 already says. It is evidence of no collateral damage. The 4 new legs are the two added fixtures, both ordinary `SolveSucceeded` at defaults.

**Workspace suite: 351 binaries, 4019 passed, 0 failed.** `pounce-hsl` excluded — it does not link on this machine (undefined MA57 symbols); nothing in the diff touches it.

**Two fixtures, one per branch of the new rule**, because a green test on the branch its fixture does not take is worth nothing:

| fixture | base `f` | retry `f` | rule |
|---|---:|---:|---|
| `mpcc_scholtes4_biactive` | `+1.8176e-09` | `-6.6088e-05` | 2 — improvement bought with slack |
| `mpcc_worse_local_solution` | `-1.3006e+01` | `-1.2072e+00` | 1 — strictly worse feasible point |

**Mutation table, verified rather than asserted** — each mutation reddens exactly its own test and nothing else:

| mutation | what goes red |
|---|---|
| drop conjunct 6 (rule 1) | `a_worse_feasible_local_solution_is_not_promoted` |
| drop conjunct 7 (rule 2) | `scholtes4_is_not_promoted_below_its_known_optimum`, `a_declined_retry_reports_one_answer` |
| drop the `answer_restored_from_floor` swap | `a_declined_retry_reports_one_answer` |
| tighten the objective tolerance below `5.8e-11` | `the_reproducer_still_promotes` |

## Costs, stated plainly

**`ctrap` loses its trap-escape.** On `benchmarks/mpcc/cases.py`'s `ctrap` from the origin, the retry used to escape the C-stationary trap (`f = 0`) and reach the global optimum `f = -0.25`. It is the same "better objective, worse primal" shape rule 2 refuses, so it now declines and the base attempt's `Solved To Acceptable Level` at the trap is what ships. That trade is deliberate: failing to escape a trap under an honest non-success verdict against returning a point below the optimum under `Optimal Solution Found`.

**The detector was not tightened.** `scholtes4` shows it firing on a model where the *remedy* lies, and the fix taken was to refuse the remedy's answer rather than to move the step threshold — a threshold separating `scholtes4` from `qpec_small` would be fitted to two models rather than derived from either. Whether the detector should exclude the no-S-stationary-point class directly is open and would need a classifier this layer does not have. Recorded in `dev-notes/mpcc-biactive-dual-divergence.md`.

## Documentation defects corrected in the same pass

Both found in the same review of #888, both in the CHANGELOG:

- It claimed the sweep showed `scaled_feasible_a` and `airport` answering from `nlp` and that "both already did so on `main`, and only the instrument changed". True for `airport` (verified against a baseline binary). **False for `scaled_feasible_a`**: on `main` the convex arm solved it directly in 69 iterations at objective exactly `0`; it reroutes only because this release stopped widening its bounds. The PR body's narrower version of the sentence was correct; the CHANGELOG generalised it.
- `feasible_x0_wide_scale` goes **32 → 80 iterations** on both legs (cvx-qp either side, reverts at `bound_relax_factor=1e-8`). It was inside the "16% fewer iterations excluding those two fixtures" aggregate rather than named. Now recorded, per CLAUDE.md's rule that every moved line be explainable and a measured regression carried as a cost of a fix be named.

The `dual_divergence_retry` option help, `docs/src/options.md` and `dev-notes/mpcc-biactive-dual-divergence.md` all carried the falsified "this retry cannot return a different local solution" claim; all three are corrected, and the dev-note gains the full follow-up with the measurements above.


jkitchin:
Reviewed at `107cf3b7`. Read the whole diff plus the surrounding code in `application.rs`, `pounce-cli/{main.rs,counting_tnlp.rs}`, `scaling_tnlp.rs` and `ipm.rs`, and traced the frames rather than assuming them.

## What's right

The retraction is the best part of this. The old claim — "this retry cannot return a different local solution, because conjunct 4 requires the promoted answer to satisfy the KKT conditions unscaled" — is a non-sequitur, it was load-bearing, and it is retracted in all three places that carried it (function comment, option help, `docs/src/options.md`) with the measurement that falsified it: 42 of 68 promotions moved the objective, 3 handed back a strictly worse feasible point.

`107cf3b7` does the same thing a second time, to this PR's own first draft. "0 of 150 degenerate LPs" conflated *cascade never ran* with *ran and certified*; instrumented, it was 149 of 150 running and certifying, and the crossover half turns out to be the majority case rather than a rarity. Retracting your own reachability claim mid-review, with the probe points named, is the behaviour the repo's notes keep asking for.

Also credited:

* Two fixtures reaching **different branches** of the new rule, per the CLAUDE.md branch rule, with a table saying which takes which, plus `the_reproducer_still_promotes` holding the other side of the tolerance.
* `the_objective_tolerance_is_not_fitted_to_one_model` pins the window from both sides — `5.8e-11` admit, `0.198` refuse, four and five orders out.
* Mutation tables in both new test modules, and a real "what this does not cover" section (sweep empty across 188 legs; explicitly not evidence about the *detector*).
* Frame check on the new `du_floor` conjunct passes. The detector reads `unscaled_du >= dual_divergence_retry_du_floor` at `ipopt_alg.rs:2704`; `runaway_is_the_whole_residual` applies the same floor to `final_unscaled_dual_inf`. Same frame, and `curr_unscaled_dual_infeasibility_max` takes `.abs()`, so both sides are magnitudes.
* `answer_restored_from_floor` is reset at `application.rs:1168`, inside the single funnel, so a reused application cannot carry a stale `true`.

Five findings. R1 is the one I would hold the merge for.

## R1 — the new CLI swap double-applies the gh#486 variable-scaling correction

`main.rs:1879` (new) overwrites `nominal_capture` and `bound_mult_capture` from the `finalize_solution` payload. `main.rs:1933` (existing, gh#486) then applies `x /= d`, `z *= d`.

That block exists because `on_converged` reads *the algorithm's iterate*, which is in scaled coordinates. The `finalize_solution` payload is not — the wrapper has already applied exactly that correction before the inner TNLP sees it:

    crates/pounce-nlp/src/scaling_tnlp.rs:553-570
        x   = sol.x   / d
        z_l = sol.z_l * d
        z_u = sol.z_u * d

and `CountingTnlp` sits *inside* that wrapper: `main.rs:1534` hands `counting` to `optimize_tnlp`, and `application.rs:1023` installs the scaling around what it is handed. `variable_scaling()`'s own doc states the distinction — a consumer reading the iterate "**rather than the `finalize_solution` payload**" is the one that must divide.

So under `nlp_scaling_method=user-scaling` or `curvature-based`, on any run where `answer_restored_from_floor()` is true — the gh#884 decline **and** the mu fallback's replay, which is on by default — the `.sol` `x`, the JSON `solution.x`, and `ipopt_zL_out` / `ipopt_zU_out` are written as `x/d^2` and `z*d^2`. Silent, plausible, wrong: the same shape as the defect this PR's second half fixes, one wrapper over.

The sentence where it hides is in this PR's own new code, `counting_tnlp.rs:41`:

> Both sources apply the same `Nlp::finalize_solution_z_l` / `_z_u` lift, so they are the same numbers and not merely compatible ones.

They are not the same numbers. `on_converged` applies that lift to the algorithm-side multipliers (`main.rs:1368`); the payload has had `* d` applied on top of it. They differ by exactly `d` — which is the entire reason the gh#486 block exists.

`a_declined_retry_reports_one_answer` cannot see this, because every run in the new test is unscaled, where `d = 1`. That is the failure mode `107cf3b7`'s own closing line describes, and the one CLAUDE.md states as a rule: a corpus uniform in the dimension a change acts on reports nothing no matter how good it is.

Cheapest fixes: move the swap **below** the gh#486 block (and below the presolve lift, which has the same property), or flag the capture as already-in-model-units so that block skips it. Then one leg on the `user_scaling_var_suffix` shape — a declined retry under `nlp_scaling_method=user-scaling`, asserting `solution.x` matches the `dual_divergence_retry=no` run.

Note the pre-existing backfill at `main.rs:1850` has the same shape, but only fires when `nominal_capture` is empty — the active-set SQP route. This change puts the same read on a far more common path.

## R2 — `retry_answer_is_admissible` assumes minimization

`base_obj` and `retry_obj` are `SolveStatistics::final_objective`: the objective evaluated on the **user** TNLP, unscaled and **signed** (`application.rs:4560-4572`). A negative `obj_scaling_factor` is the documented way to pose a maximization (`upstream_options.rs:1049`), and this is not hypothetical here — the repo already shipped one defect from that sign, pinned by `masked_certificate_fuzz.rs::the_veto_is_not_disabled_by_a_negative_objective_scaling_factor`, whose doc records that dividing by the signed factor "returned a NEGATIVE max-norm". That fix took `.abs()` in the residual accessors, so the detector still fires under maximization and this path stays reachable.

Under `obj_scaling_factor < 0` both new rules invert:

* rule 1 (`retry_obj > base_obj + tol` refuses) refuses genuine **improvements** — a behaviour regression against `main`, which promoted them;
* rule 2 admits a strictly **worse** answer whenever `retry_viol <= base_viol` — re-arming exactly the worse-feasible-point class the conjunct was added to block.

Cheapest fix: one `signum` on the objective sense before comparing, plus a unit test that is the existing table row with both objectives negated. `final_scaled_objective` looks like the natural frame but is not a drop-in — it defaults to NaN and the SQP path never sets it (`application.rs:2254`).

## R3 — rule 2 has no noise floor, and the correctness claim rests on a correlation

`retry_viol <= base_viol` is an exact non-increase. On `scholtes4` the violation moves 16 orders (`2.07e-25 -> 1.09e-09`), so the rule bites hard and correctly. It fires identically on `1e-17 -> 1.1e-17`, where which side a retry lands on is arithmetic noise rather than a property of the answer. That is the same objection this PR levels at the gate it rejected: an absolute floor that "declines `deb7` by a ONE PERCENT margin, which is a coincidence rather than a discriminator."

The cost is visible in the note's own table: promotions 68 -> 23, "objective moved materially" 42 -> 1. Rule 1 accounts for 3 and the new `du_floor` for 7, so roughly 35 declines are improvements refused because the violation happened to tick up, mostly far below any tolerance. The direction is conservative — the base answer is feasible and in hand — so this is a cost to name rather than a defect. But it is worth naming: the note currently says rule 2 "refuses purchases, not improvements", and in the sub-tolerance regime it refuses whichever it lands on.

One line for "what this does not cover", from the PR's own test: had `scholtes4`'s retry *held* its violation, `f = -6.6088e-05` — below the model's exactly-known `f* = 0` — would have been **admitted** (`an_improvement_bought_with_primal_slack_is_refused`, second assertion). The gate detects "the primal moved", not "below the optimum". On this model the two coincided.

## R4 — the accessor's contract is not honoured by the other `on_converged` consumers

`answer_restored_from_floor()`'s doc says "**A caller that captures the solution in `set_on_converged` must consult this.**" The CLI's main path now does. These do not:

* `crates/pounce-sensitivity/src/solver.rs:252`
* `crates/pounce-sensitivity/src/convenience.rs:322`
* `crates/pounce-cli/src/minima/mod.rs:770`

After any floor replay — including the mu fallback's, on by default — the sensitivity layer's converged KKT state belongs to the discarded attempt. Pre-existing, and the note is right that the KKT state cannot be rewound; but this PR is what turns it into a stated contract, so it deserves an issue rather than silence. The accessor doc should also say the payload is in the **model's own units**, which is what R1 turns on.

## R5 — the sigma re-record drops a conjunct the recording site defines as part of the bit

The recording site is `candidates[pick].status == QpStatus::Optimal && far_from_optimal` (`ipm.rs:3612`), under a comment insisting the bit means "this answer should be demoted" and that a future reader must not widen it: *"Widen the recording, not the reader."* `sigma_verdict_after_crossover` returns `!sigma_forward_error_is_small(...)` alone, with no status conjunct — so on a moved point whose post-crossover status is not `Optimal`, the two writers of the same bit disagree on its definition. Unreachable today (`demote_uncertified_sigma_optimum` re-checks status, and crossover is never-regressing), so this is consistency rather than behaviour. One `sol.status == QpStatus::Optimal &&` closes it, and keeps the comment true.

---

Nothing above disputes either fix. The defect is real, the argument that the old gate was safe was genuinely a non-sequitur, and the evidence — on both halves, after `107cf3b7` — is the strongest part of the PR. R1 is a new silent wrong-answer path on the `.sol` and the dual suffixes, introduced by the second fix, with the test for that fix running in the one configuration where it cannot appear. R2 is a second correctness gap, cheap to close. R3-R5 are questions, not vetoes.

jkitchin:
Addressed at `c6ca816c`. R1 and R2 confirmed and fixed; R3–R5 handled as you framed them.

## R1 — confirmed, and worse than "cheapest fix"

You were right, and I want to be precise about how I got it wrong: I reasoned about which wrapper was outside which instead of printing the vectors, and the reasoning produced a plausible answer that was wrong in the direction that hides. Traced properly:

```
ScalingTnlp   (x /= d, z *= d)      <- optimize_tnlp wraps what it is handed
  CountingTnlp  (captures HERE)     <- payload is in MODEL units
    PresolveTnlp  (row/column lift) <- but NOT lifted out of the reduced space
      NlProblem
```

So neither of your two suggested placements is quite right on its own — moving the swap *below* the gh#486 block would also move it below the presolve lift, which the payload still needs. The fix is the second one you named: flag the capture as already-in-model-units so **only** the gh#486 block skips it, and the lift still runs.

Reproduced on `mpcc_scholtes4_biactive` under `nlp_scaling_method=curvature-based`, `d = [2, 2, 0.5007]`:

| | `x` |
|---|---|
| declined retry, before | `[7.28e-14, 7.28e-14, -3.63e-09]` |
| correct (`dual_divergence_retry=no`) | `[1.46e-13, 1.46e-13, -1.82e-09]` |

Every component off by exactly the factor, objective beside it unchanged and right. `f(solution.x) = 3.63e-09` against a reported `1.82e-09` — so my own one-answer invariant *would* have caught it, unscaled being the one configuration where it cannot.

New leg `a_declined_retry_reports_one_answer_under_variable_scaling` asserts both that the point and objective agree **and** that the point equals the `dual_divergence_retry=no` run's — the second is what pins the *number of times* the correction landed rather than its self-consistency. `curvature-based` rather than `user-scaling` so the leg derives its own factors and cannot silently degrade to `d = 1` if a suffix is dropped. Mutation-checked: restoring the double-application reddens **that leg alone**, unscaled stays green.

I also removed the sentence you quoted from `counting_tnlp.rs`. You're right that it was false — and it was load-bearing for the wrong placement.

## R2 — confirmed

Verified directly rather than from the docs: at `obj_scaling_factor=-1` on that fixture, `solution.objective` equals `f(x)` and the solve maximizes it (diverges to `+9.08e+15`). So both rules inverted exactly as you describe. Fixed by normalizing the sense before comparing; `negating_the_objective_and_the_sense_is_inert` asserts the invariance over the whole table rather than re-listing negated rows, and `the_rules_follow_the_objective_sense` carries the mirrored rows explicitly.

Agreed on `final_scaled_objective` — I didn't use it, for the reason you give.

## R3 — named, not fixed

I've taken this as a cost to state rather than a threshold to invent, since a noise floor here would need its own measured population and I'd be adding exactly the kind of unbacked constant this PR criticises. Recorded on the rule and in the changelog: ~35 of the 45 removed promotions are improvements refused sub-tolerance, and the rule detects *the primal moved*, not *below the optimum* — had `scholtes4`'s retry held its violation, `f = -6.6088e-05` would have been admitted. That is now stated, not left to the second assertion of a test.

## R4 — documented at all three sites

Annotated `pounce-sensitivity/src/{solver,convenience}.rs` and `pounce-cli/src/minima/mod.rs` with what the flag means and why they can't consult it, and the accessor now states that the payload is in the **model's own units and the reduced presolve space** — which is what R1 turned on. Not filed as an issue, per the repo's fix-it-in-the-PR rule; if you'd rather have it tracked separately say so and I'll split it.

## R5 — closed

One `sol.status == QpStatus::Optimal &&`, with a comment pointing back at the recording site's "widen the recording, not the reader".

---

**Verification after all five:** fixture sweep **0 moved lines** across all 188 pre-existing legs; workspace suite **351 binaries, 4022 passed, 0 failed**; clippy clean under the CI lints; docs consistency OK.

The `dev-notes` entry now carries the R1 wrapper diagram and the general form of the mistake, which is the same one twice in this PR: **a frame is a measurement, not an inference.** Printing `snapshot_x`, `d`, the `on_converged` capture and the payload side by side settled in one run what I'd got wrong by reading call sites.

PR #903, opened 2026-09-04
Gate 1: a phase-changing flash as an MPCC, its tutorial, and the finish the supported route could not run

### LINKED ISSUE #776: Research/tutorial case: phase-changing dynamic operations as MPCCs with POUNCE–DISCOPT validation

jkitchin:
Gate 1 of #776, under the application epic #780. Gate 0 (#794) established a supported MPCC route on a corpus that was deliberately not a process model; this is the first model with physics in it, and it stops at one flash.

**No closing keyword, deliberately.** Gate 1 asks the fixture to do six things and this delivers five: the sixth, a reduced DiscOpt GDP/SOS1 regime cross-validation, is blocked on the two live DiscOpt slices — [#1147](https://github.com/jkitchin/discopt/issues/1147) then [#1148](https://github.com/jkitchin/discopt/issues/1148). `Fixes #776` would auto-close an issue with Gates 2–5 still ahead of it, so this references #776 and #780 without one.

*(Body refreshed for head `26b803d`: 12 commits, 24 files, +5825/−6.)*

## Problem

Phase appearance changes a model's *shape*. Below the bubble point there is no vapour, so there are no vapour equations; inside the two-phase region there are. A simulator branches on a stability test and instantiates the equations that regime needs — harmless while the flash is the whole problem, fatal once it is one block of a simultaneous transcription that cannot change dimension halfway through a horizon.

Complementarity is the standard way out, and POUNCE had no worked instance of it on a physical model. Gate 0's corpus is quadratic objectives with affine pairs, six variables at most, by construction not a process model.

The honest baseline, measured first, because the interesting result is that the formulation *loses* it:

| | vapour fraction correct? | cost |
|---|---|---|
| successive substitution + Rachford–Rice | every point on the path | **7–10 sweeps** |
| this MPCC, `scholtes_then_ncp` | every point on the path | ~100 interior-point iterations |

So the case for the formulation is composability, not speed: successive substitution is a procedure with an inner loop and a regime branch, its vapour fraction is a Rachford–Rice root *clipped* to `[0,1]` (so its derivative w.r.t. an upstream variable is identically zero in the single-phase regions — wrong, and silently so), and it carries no phase-state certificate. The job here is to establish correctness where the answer is independently known, before the formulation is used where it is not.

## Cause

Two mechanisms, both found by building the model rather than reasoning about it.

**The isofugacity row's normalization.** `phi` takes a *composition*, so it is evaluated at `x/Sx` and `y/Sy` — but the logarithm outside it is not normalized. The first implementation normalized both, which adds `ln(Sy/Sx)` to every row. That term **vanishes identically in the two-phase region**, where `Sx = Sy = 1`, so the broken model solved, converged, and agreed with the reference at every two-phase temperature. It was wrong only in the single-phase regimes — the ones the fixture exists to reach.

**The exact-product finish cannot run on a square model.** `scholtes_then_ncp` finishes with one `G*H = 0` solve. That makes both product rows equalities, and a square model — no objective, no slack, which is what every equilibrium-stage process model is — then has 5 free variables against 6 equality rows. `application.rs:4102` refuses it before iteration zero with `Not_Enough_Degrees_Of_Freedom`, mirroring upstream Ipopt (`IpOrigIpoptNLP.cpp:299`). **The gate is correct and unchanged by this PR.** The symptom was quiet: every record still reported `ok` and agreed with the oracle, and `scholtes_then_ncp` and `scholtes_warm_full` returned *identical* numbers — the composition was its continuation half wearing the composition's name.

## Fix

`pounce.examples.flash_mpcc` — ethane/n-butane at 10 bar, Peng–Robinson reusing the fugacity layer from #777, 5 variables and 8 rows at every temperature. The pairs are phase *amount* against phase *stability slack*, not `L ⟂ V`: at `beta = 0` the balance forces `x = z`, the isofugacity rows collapse to Michelsen's tangent-plane stationarity, and `Sy <= 1` is exactly `TPD >= 0`. The complementarity *is* the stability test. Bubble and dew are its two biactive points.

For the finish, `Route.finish_fallback` substitutes `G*H <= 0` where the equality form cannot run — identical feasible set when `G, H >= 0`, but an inequality does not count against the gate.

**Rejected: making that the finish unconditionally.** My first reading was that identical feasible sets make it a free drop-in. Measured over Gate 0's corpus as `scholtes_then_ineq`, added here as a permanent comparison arm, that is wrong:

| route | solved | at `f*` | **below `f*`** | median iters | worst source compl |
|---|---:|---:|---:|---:|---|
| `scholtes_then_ncp` | 60 | 57 | **0** | 55 | 4.87e-11 |
| `scholtes_then_ineq` | 60 | 57 | **3** | 77 | 1.81e-12 |

It ties on solved and at-`f*` and beats the equality finish on complementarity by 27×, and returns three cells *below* the MPCC optimum where the equality finish returns none — `scholtes4`, skew leg, all three starts, `-2.7e-06` at a clean complementarity of `1.8e-12`. Identical feasible sets do not imply identical converged points: an interior-point method lands elsewhere inside the same set depending on whether the row carries a free multiplier or a sign-constrained one. Below-`f*` is the single axis #794 chose that finish for, so the inequality form is used only where the equality form cannot run at all.

Also: the Hessian is built with `jacfwd(jacfwd(·))`, not `jax.hessian`. Under `jax.jit` reverse mode is wrong by O(20) at exactly the two path points straddling the bubble point and nowhere else, where the cubic approaches a double root and Cardano's branch runs `arccos` into its endpoint singularity. Forward-over-forward is exact and costs the same (0.15 ms both).

## Blast radius

**Gate 0 is bit-identical with the fallback in place**, against `results-full.json` regenerated on this branch: `scholtes_then_ncp` at 60 solved / 57 at `f*` / 0 below `f*` / 55 median iters / `4.874e-11` / 8 restarts. Every solvable case still finishes on `prod_eq` — the fallback fires on **none** of them. Of the eleven cases only `infeasible_pair` trips the gate at all, it never reaches a finishing stage, and it still reports `Infeasible_Problem_Detected` on all four cells, so the report's finding 8 still holds.

**No Rust changed** in any of the twelve commits, so no solver behaviour moves and no trajectory sweep applies. `cargo fmt --all -- --check` clean; clippy and `cargo test` are inherited, not re-run, and I would rather say that than tick a box I did not earn.

On the flash the fallback fires at every temperature and the composition finishes — the evidence being a pair of rows that used to be identical and no longer are:

| route | worst \|Δbeta vs oracle\| | worst \|G·H\| |
|---|---:|---:|
| `scholtes_then_ncp` | 5.1e-10 | 1.1e-12 |
| `scholtes_then_ineq` | 5.1e-10 | 1.1e-12 |
| `scholtes_warm_full` (continuation only) | 8.4e-10 | 1.8e-12 |

Cost: 4105 interior-point iterations per cold leg against 3681 for the continuation alone, ~12% for the MPCC-feasibility guarantee the continuation cannot give.

## Tests

`python/tests/test_flash_mpcc.py` — **29 assertions**, ~66 s, the fast regression #776 asks the fixture to become. They are *source-level*: the phase regime, vapour fraction and both phase sums are compared against an independent Michelsen-stability plus Rachford–Rice calculation sharing only the fugacity primitive. A red test means "the solver's answer is not the flash", not "the NLP did not converge".

**Four things are pinned by mutation rather than by assertion, each verified to bite:**

| mutation | before the pin | after |
|---|---|---|
| restore the isofugacity normalization defect | — | row violation **9.3e-01** at 230 K vs a 1e-10 threshold (fails first at a *single-phase* point; identically zero in the two-phase region) |
| remove `Route.finish_fallback` | 18 passed, CLI exited 0 | **10 pytest failures** + `flash.run --smoke` exits 1 naming `Not_Enough_Degrees_Of_Freedom`. It was 5 when the pin was added; the CLI cases in the last row take the other 5, because each asserts its undoctored record passes the contract before doctoring it |
| empty the `validation` block | CLI exited 0 (vacuous `all()`) | contract fails on missing `EXPECTED_OK_KEYS` |
| bypass the contract call in `run_smoke` (`reasons = []`) | 26 passed — the contract was correct and *unwired* | **3 failed / 26 passed**, the three new CLI cases, each on `run_smoke` returning 0 |

The last row is @bernalde's finding on the previous head and it was right: the positive test ran `run_smoke` and asserted exit 0, the negative test called `smoke_contract_failures` directly, and nothing joined them — so the binding between the CLI and the contract was untested, which is the same shape as the defect this fixture exists to pin. `test_the_smoke_cli_rejects_a_bad_record` drives the real `run_smoke` with only `solve_route` stubbed, returning a doctored copy of a genuinely solved record at one of the five temperatures and the real record at the other four, and covers all three contract clauses (`finish_applied=False`, `validation={}`, a `finish_lowering` of `prod_eq` where the square model forces `prod_ineq`). It asserts the exit code, the reason text, that `solve_route` was reached once per smoke temperature, and that *only* the doctored temperature is reported — a `run_smoke` that failed everything would satisfy an exit-code assertion just as well.

The remaining tests are new files, so "fails on the parent commit" is by absence rather than by biting; the mutations above are the meaningful claims.

Full Python suite at this head: **1354 passed, 28 skipped** in 821 s (`Python tests (jax)`, run [33905161611](https://github.com/jkitchin/pounce/actions/runs/33905161611)), and all eight checks green on `26b803d`. The move from the 1348 measured at `250ff2c` is the two flash increments and nothing else — 23 → 26 → 29 assertions; merging `main` added no Python tests. `make -C benchmarks flash-selftest` and `mpcc-selftest` both pass and need no solver.

**Deliberately not pinned:** the four-leg hysteresis traversal and the full route matrix are benchmark targets, not CI — matching how `mpcc/` and `warmstart/` are treated. The smoke subset is what runs per-PR.

## Docs

- `python/notebooks/38_phase_changing_flash_mpcc.ipynb` — executed on a **clean exact head** (`f49eaaf`, no `-dirty`), 14/14 code cells, five figures, following the epic's template (motivation → equations → baseline → formulation → diagnostics → verification → result → limitations). Two cells reproduce findings rather than assert them: the defective row evaluated beside the correct one, and the `prod_eq` finish run from the continuation's own answer reporting `Not_Enough_Degrees_Of_Freedom` with the six equality rows named against five variables.

  **The stamp names `f49eaaf`, not this head, and that is now two commits back — so here is the gap, stated rather than left to be rediscovered.** `03aa78c` merged `main`, which touched `crates/`; a reviewer re-running @bernalde's provenance check will see that and reasonably ask whether the saved numbers came from a different solver. They did not. `git diff f49eaaf 26b803db` is 18 files, and outside `crates/pounce-wasm/` (shell scripts, web assets, a prebuilt wasm wheel — the browser demo) it is `ci.yml`, `docs.yml`, `CHANGELOG.md`, `docs/src/wasm.md`, the notebook itself and `test_flash_mpcc.py`. **Zero `.rs` files and zero `Cargo.toml`/`Cargo.lock` changes**, and `python/pounce/` is untouched, so the NLP solver and the model the notebook ran against are bit-identical to this head's. The stamp still has to name the *parent* of the commit carrying the notebook — the tree must be clean at execution time and committing the result advances HEAD — so `f49eaaf` remains the commit a reader checks out to reproduce it.
- `docs/src/flash-mpcc.md`, linked from `SUMMARY.md`.
- `dev-notes/mpcc-gate0-report.md` — the amendment as its own section rather than an edit to the original decision, plus failure-boundary entry 2a: every case in that corpus has an objective and slack, so none exercises the degrees-of-freedom gate on the finishing solve. It was uniform in exactly the dimension Gate 1 acts on, and reported "the supported route solves every cell" while silent about a model class where half the route cannot execute.
- `CHANGELOG.md` `[Unreleased]`.

## What this is not evidence about

One equilibrium stage, two components, one operating parameter — nothing here bounds a tray, a column, or a dynamic transcription; #776 gates those on this result and this result does not reach them. The DiscOpt half is blocked on #1147 → #1148 and every result file records that it was not run, and why, rather than omitting the field ([#1123](https://github.com/jkitchin/discopt/issues/1123) is closed and is the design record, not the blocker). The path stays far from the mixture critical point. POUNCE is a local solver, and the reference calculation's multistart stability test is the stronger claim of the two — which is why agreement is evidence and a disagreement would be a finding about the MPCC.

https://claude.ai/code/session_01HcLHFGk3PNUP1RVjrdW6gq


jkitchin:
## Review

Strong PR — the formulation is right, the mechanism is right, and the write-up is unusually honest about its own limits. One real gap.

## The defect this PR fixes is not pinned

The diagnosis above is that the bug was **silent**: "every record still reported `ok` and agreed with the oracle, and `scholtes_then_ncp` and `scholtes_warm_full` returned *identical* numbers — the composition was its continuation half wearing the composition's name."

`benchmarks/flash/runner.py:88-92` adds exactly the right instrumentation to observe that — `finish_applied`, `finish_status_msg`, `finish_lowering`. But nothing asserts on them. `validate.py` produces seven `*_ok` keys and none is finish-related, and `python/tests/test_flash_mpcc.py` never reads `finish_applied`.

Run the tolerances against the measured table in the PR body:

| | continuation only | asserted threshold | margin |
|---|---:|---:|---:|
| worst \|Δbeta vs oracle\| | 8.4e-10 | `ORACLE_TOL` 1e-6 | 10³ |
| worst \|G·H\| | 1.8e-12 | `CORNER_TOL` = sqrt(1e-8) = 1e-4 | 10⁸ |

So if `Route.finish_fallback` were reverted tomorrow, **every assertion in `test_flash_mpcc.py` stays green** — including `test_complementarity_is_reported_in_source_terms`, which is the one that looks like it would catch it. The regression would be caught only by a human reading two rows of a benchmark table and noticing they had become identical again. That is the same detection mechanism that failed the first time.

Two lines in the smoke test close it:

```python
assert rec.finish_applied, f"finish did not run at {temperature_k} K: {rec.finish_status_msg}"
assert rec.finish_lowering == "prod_ineq"
```

The second matters as much as the first — `finish_applied` alone stays green if the gate stops firing and `prod_eq` starts running, which is a different thing worth knowing about.

## Verified, including one I expected to be a bug

- **The DOF predicate is correct**, and not for the obvious reason. `runner.py` tests `nlp.n < n_eq` on the *total* variable count, while the gate at `application.rs:4108` tests `n_x_var < n_c` on the *free* count (`x_space` excludes fixed variables under the default `make_parameter`). I went looking for a model with fixed variables where the fallback fails to fire and the composition silently degrades again. It does not exist: `tnlp_adapter.rs:256-278` mirrors `IpTNLPAdapter.cpp:623-633` and auto-switches to `relax_bounds` precisely when dropping fixed variables would trip the gate, which collapses `n_x_var` back to `n_full_x`. Equality classification is exact `lo == hi` on both sides (`tnlp_adapter.rs:584`). The predicate is equivalent in every case that can fire. Worth a one-line comment saying *why* `nlp.n` is the right quantity, because the next reader will make the same wrong inference I did.
- **No Rust changed** — confirmed across all six commits; the DOF gate is untouched, so "the gate is correct and unchanged by this PR" holds and no trajectory sweep applies.
- **The normalization fix is as described** — `isofugacity_residual` is `log(x) + ln_phi(x/Sx) - log(y) - ln_phi(y/Sy)`, normalized inside `ln_phi` only. The argument that the defective form vanishes identically on `Sx = Sy = 1` is right, and it is what makes `test_the_oracle_answer_solves_the_model` load-bearing.
- **`jacfwd(jacfwd(.))`** is used rather than `jax.hessian`, with the measured table at `flash_mpcc.py:1129-1143`.
- The pair choice (amount complementary to stability slack, not `L ⟂ V`) is correct, and `test_the_complementarity_pairs_are_amount_against_slack` pins it properly.

## Two minor ones

**`test_solver_reaches_the_flash` can stop testing.** It filters `rec.validation` for keys ending `_ok` that are falsy. If `validation` came back empty, or the `_ok` naming convention changed in a refactor, `failed` is empty and the test degrades to `assert rec.ok` without announcing it. Cheap fix: assert the expected key set is a subset of what came back.

**`runner.py:468-477` leans on a falsy default.** `finish_stage.lowering or route.finish` is correct only because `StageRecord.lowering` defaults to `""`. If that default ever becomes a non-empty sentinel, it silently reports the nominal finish for a solve that used the fallback — the exact misreport the field was added to prevent. Worth a comment naming the `""` dependency.

## Verdict

Approve once the `finish_applied` assertion lands. Everything else is nits.

---
_Review generated by [Claude Code](https://claude.ai/code/session_013bUD9V6h3GkS8E1PnBkz8p)_

jkitchin:
Fixed in `c2979ce`. The main finding was right, and I confirmed it the way you'd want rather than taking it on trust.

## The gap was real — measured

Reverted `finish_fallback` on a scratch tree and ran the suite: **18 passed**. Every assertion stays green with the fix removed, `test_complementarity_is_reported_in_source_terms` included, exactly as your margin table predicts. The regression really would have been detectable only by a human noticing two benchmark rows had become equal again.

`test_the_finishing_solve_actually_runs` closes it, and it bites — on the reverted tree it fails at all five smoke temperatures with "the finishing solve did not run", and passes with the fallback in place:

| tree | result |
|---|---|
| fallback reverted | **5 failed**, 18 passed |
| fallback in place | 23 passed |

I took your point about asserting `finish_lowering` too, and the docstring records why: `finish_applied` alone stays true if the gate stops firing and the equality finish starts running, which would mean either the model stopped being square or the solver's gate moved — both worth failing on rather than absorbing.

## The `nlp.n` vs `n_x_var` comment

Added at the predicate, with your reasoning rather than mine — including that the inference you made is the natural one and why it does not hold (`tnlp_adapter.rs` auto-switching to `relax_bounds` exactly when dropping fixed variables would trip the gate, collapsing the free count back to the total). Thank you for chasing that one to ground; I had not checked it and would have shipped the predicate on a coincidence I could not have defended.

## `EXPECTED_OK_KEYS`

Added as a set assertion before the values are read, so a rename or an empty `validation` fails instead of silently degrading to `assert rec.ok`.

## One I did not change, with evidence

The third item — `runner.py:468-477` leaning on `StageRecord.lowering` defaulting to `""` via `finish_stage.lowering or route.finish` — does not apply to this tree. `benchmarks/flash/runner.py` is 364 lines, so there is no 468-477, and that expression does not appear anywhere in the harness:

```
$ grep -rn "or route.finish\|lowering or " benchmarks/flash/
(no matches)
```

`rec.finish_lowering` is assigned directly from the local the fallback sets (`runner.py:314`), so there is no falsy-default path that could misreport the nominal finish. If you were reading a different revision, or you meant a different call site, point me at it and I will fix it — but I would rather leave it than add a comment describing a dependency the code does not have.

---

jkitchin:
Correction: **you were right about the third item and I was wrong.** Fixed in `250ff2c`.

`finish_stage.lowering or route.finish` is at `benchmarks/mpcc/runner.py:474`. That file is 545 lines, so `468-477` lands exactly on the block. My previous reply grepped `benchmarks/flash/` only — you named the file, not the directory, and the conditional's reporting half went into the Gate 0 harness too. So I dismissed a correct finding on the strength of a search that could not have found it, which is worse than not checking.

Having looked properly, the guard was also **dead**, so the fix is stronger than the comment you asked for. `lowering_here` is always one of `route.finish`, `route.lowering` or `route.finish_fallback`, and the single `StageRecord(...)` in the module sets `lowering=` from it unconditionally — the empty string cannot reach the reader, so the `or` never fired. It was carrying your failure mode without ever doing any work: a future non-empty sentinel default would have made this report the route's *nominal* finish for a solve that used the fallback, silently.

It now reads straight off the stage. If `""` ever does arrive, that is a bug in the stage constructor and should surface as one rather than be absorbed by a plausible-looking default. The reasoning is a comment at the site, since "this guard is dead" is exactly the kind of claim that rots.

Verified: `mpcc.selftest` passes, `mpcc.run --smoke` clean over 170 records, and both composed routes still report the lowering their answer actually came from —

```
scholtes_then_ncp    -> prod_eq   x15,  scholtes x2
scholtes_then_ineq   -> prod_ineq x15,  scholtes x2
```

(the `scholtes` rows are `infeasible_pair`, which never reaches a finishing stage.)

All four review items are now addressed.

---

jkitchin:
@bernalde do you want to take a look at this before it merges

benchmarks/flash/run.py
```diff
@@ -0,0 +1,373 @@
...
+        "vapor_trial_stationary_points": ref.vapor_trial.stationary_points,
+        "liquid_trial_stationary_points": ref.liquid_trial.stationary_points,
+    }
+
+
+def _all_ok(rec: SolveRecord) -> bool:
+    """Every ``_ok`` key in the record's validation block, plus the status."""
+    if not rec.ok:
+        return False
+    return all(
```
bernalde: [Blocking] Make the public smoke command assert the same contract as the regression tests.

The new pytest cases correctly require the expected validation keys and prove that the finishing solve used `prod_ineq`, but this advertised asserted entry point does neither. On the exact head I removed `finish_fallback`: pytest failed at all five temperatures (`5 failed, 18 passed`), while `python -m flash.run --smoke` exited 0 and printed `smoke passed` with `finish_applied == False`. Replacing the validation block with `{}` also exited 0 because `all(...)` is vacuously true here.

Please make this path require the expected validation keys, `finish_applied`, and the supported route's expected `finish_lowering`, then exercise `run_smoke()` itself in a regression test so the CLI contract cannot diverge from the direct solver tests.

benchmarks/flash/path.py
```diff
@@ -0,0 +1,238 @@
...
+        if progress is not None:
+            progress(rec)
+        if rec.ok and rec.x is not None:
+            prev_x = np.asarray(rec.x, dtype=float)
+            # Only the primal is carried between temperatures here. The
+            # multiplier/barrier state belongs to a *different* problem
+            # -- a different temperature, hence different rows -- and
+            # Gate 0 measured the full-state warm start where it is
+            # defined, which is between the stages of one continuation.
+            warm_state = {"x": prev_x}
```
bernalde, line 188: [Blocking] Record the inter-temperature warm start that is actually supplied.

Only `x` is carried here, although both the module description and this function's docstring say the previous multipliers and barrier parameter are included. `solve_route` then enters the route's `full` ladder, so the first stage at the next temperature is recorded as `warm_level="full"` even though `mult_g`, bound multipliers, and `mu` are absent. A two-temperature probe reproduced that label on the second temperature.

Because #776 explicitly asks this benchmark to compare warm-start behavior across switching points, the reported iteration comparison currently describes a primal-only inter-temperature start under a full-state label. Either carry the complete state, or document and label this outer start as `primal`; please add a test that pins the chosen contract.

benchmarks/flash/stamp.py
```diff
@@ -0,0 +1,145 @@
...
+        import discopt  # type: ignore
+
+        disc = {
+            "present": True,
+            "version": getattr(discopt, "__version__", "unknown"),
+            "commit": "unknown (installed package, not a checkout)",
+            "comparison_run": False,
+            "reason": (
+                "discopt is importable, but the reduced GDP/SOS1 comparison "
+                "gh#776 asks for is blocked on jkitchin/discopt#1123 (first-class "
```
bernalde: [Blocking] Emit the current DiscOpt dependency provenance.

`jkitchin/discopt#1123` is closed and remains the design record; the current #776 coordination update says the actionable sequence is `jkitchin/discopt#1147 -> jkitchin/discopt#1148 -> Gate 1`. The exact-head full run therefore writes a stale blocker into both its JSON stamp and rendered report. Keep `comparison_run: false`, but update this generated reason (including the import-failure branch) and the matching documentation/notebook surfaces to name the live prerequisites.

bernalde (changes requested):
Requesting changes because the benchmark's executable and provenance surfaces do not yet enforce or describe the evidence they report. I did not find a defect in the flash formulation or in the supported route's numerical result itself.

### [Blocking] Refresh the committed notebook evidence on the reviewed head


The committed notebook's first output stamps `fc84e09-dirty`, not the reviewed head `250ff2c`. That makes the saved measurements and figures unverifiable against this tree, contrary to the repository requirement that every measurement carry its source commit. I executed all 14 code cells against the exact clean head: the 34-temperature run completed, all five PNG outputs rendered, and no cell errored. Please promote a clean exact-head execution and its derived figures before merge.

### [Nonblocking] Refresh the PR evidence summary


The body still describes six commits, 18 flash assertions, and `1350 passed, 21 skipped`; this head has eight commits and 23 flash tests, while its current JAX CI log reports `1348 passed, 28 skipped`. It also names the now-closed `jkitchin/discopt#1123` as the live blocker; the current coordination record names open successors `jkitchin/discopt#1147` and `jkitchin/discopt#1148`. Please update the merge record to match the head being reviewed.

The inline findings cover the smoke command's false-success path, the inter-temperature warm-state labeling, and the dependency reason emitted into generated result artifacts.

Verification performed on exact head `250ff2c9216b3e45f5e3d924efee1aaf4d3389dd`:

- all eight live GitHub checks pass; the JAX job reports 1348 passed and 28 skipped;
- `python/tests/test_flash_mpcc.py`: 23 passed;
- flash self-test and five-temperature smoke: passed;
- full 340-record matrix completed; the supported route solved 136/136 points, the cold legs agreed, and no path dependence was detected;
- MPCC self-test and 170-record smoke: passed;
- documentation reachability: 49 pages reachable with no dead links;
- exact-head notebook execution: 14/14 code cells completed, five figures rendered, no errors.

The mutation checks are what make the first inline item blocking: removing `finish_fallback` produces the intended five pytest failures, but the documented smoke CLI still exits 0 at every temperature; replacing validation with `{}` also exits 0.

benchmarks/flash/run.py
jkitchin: Fixed in `f49eaaf`. You're right, and the reproduction was exact — the advertised asserted entry point could not fail.

`smoke_contract_failures` now requires, in order: the expected `_ok` keys to be **present** before any value is read, none of them false, the finishing solve to have run, and it to have run on the lowering a square model forces. On the reverted tree the CLI now exits 1:

```
FAILED:
  T=250.0: the finishing solve did not run: Not_Enough_Degrees_Of_Freedom
  T=268.0: ...
  (all five)
CLI_EXIT=1
```

`EXPECTED_OK_KEYS` moved to `validate.py` as the single authoritative contract — it was living in the test file, which is precisely how the CLI and the suite drifted apart. Three tests pin it, and per your last sentence one of them calls `run_smoke()` itself rather than reimplementing its checks, so the CLI contract cannot diverge from the direct solver tests again. The other two drive `smoke_contract_failures` against doctored copies of a real record (`finish_applied=False`, `validation={}`, `finish_lowering="prod_eq"`), so the exact shapes that fooled it stay pinned without the suite needing to mutate the source.

Worth naming: this is the same defect the previous review round fixed in pytest, and I fixed it there and left it standing in the entry point the README and Makefile actually tell people to run. Checking that the *documented* path enforces the contract, not just the one I had just edited, is the step I skipped.

---

benchmarks/flash/path.py
jkitchin: Fixed in `f49eaaf`, and your two-temperature probe reproduces it exactly:

```
before:  T=300.0 warm_level='none'   T=310.0 warm_level='full'
after:   T=300.0 warm_level='none'   T=310.0 warm_level='primal'
```

I took the second option you offered — document and label it `primal` — because carrying the full state is the wrong fix on the merits: the multipliers and barrier parameter belong to a different temperature with different rows, and Gate 0 already measured the full-state warm start where it is actually defined, between the stages of one continuation. So the label follows the code rather than the reverse.

`runner._effective_warm` caps a route's requested level by what the supplied state can support, which also fixes a second-order version of the same problem: the ladder itself was being built as a `full` ladder over a primal-only state. Both docstrings now say primal, and `test_the_inter_temperature_start_is_labelled_primal` pins the contract at both temperatures.

Your framing is the part I want to acknowledge: because #776 asks this benchmark to compare warm-start behaviour across the switching points, a primal-only iteration count filed under a full-state label is the measurement being wrong, not just its name. My own comment in `traverse` correctly said only `x` was carried while the module docstring above it claimed multipliers and `mu` — I wrote both and did not notice they disagreed.

---

benchmarks/flash/stamp.py
jkitchin: Fixed in `f49eaaf`. Both branches of `repositories()` now emit from a single `_BLOCKED_ON` constant naming the live sequence — `jkitchin/discopt#1147` (first-class complementarity/MCP relation with durable source provenance) then `jkitchin/discopt#1148` (source complementarity residuals and the local-versus-certified result contract) — with `#1123` described as the closed design record it still is. `comparison_run` stays `false`.

Putting it in one constant rather than repeating the string is deliberate: the import-success and import-failure branches had already been written separately, which is how one of them could have gone stale on its own.

The matching surfaces are updated too, so a regenerated run and its documentation agree: `benchmarks/flash/README.md`, `docs/src/flash-mpcc.md`, `schema.json`'s `discopt` description, the `render()` limitations section, and notebook 38 §9.

The coordination update naming #1147/#1148 was posted to #776 on 2026-09-03, before I started this work, and I did not read past the 09-02 comment. Checking whether a cross-repository blocker is still open before baking it into generated artifacts is cheap, and I should have done it.

---

jkitchin:
@bernalde — all four blocking items and the nonblocking one are addressed in `f49eaaf` and `74ee506`. Each inline thread has a reply with its reproduction and is resolved. Summary:

| finding | fix | how I know it bites |
|---|---|---|
| smoke CLI false success | `smoke_contract_failures` requires key **presence**, `finish_applied`, and `finish_lowering` | reverted tree: CLI now exits **1** at all five temperatures naming `Not_Enough_Degrees_Of_Freedom`; `validation={}` now fails on missing keys |
| warm-state label | `runner._effective_warm` caps the route's level by what the state supports; docstrings corrected | your probe: 310 K was `full`, now `primal` |
| stale DiscOpt blocker | one `_BLOCKED_ON` constant naming #1147 → #1148, #1123 as design record | both `repositories()` branches, README, docs page, schema, report renderer, notebook §9 |
| notebook provenance | re-executed on clean `f49eaaf` | stamp prints `f49eaaf`, no `-dirty`; 14/14 cells, 0 errors, 5 figures |
| PR body stale | refreshed for `74ee506` | 10 commits, 26 flash assertions, DiscOpt successors, CI suite figure attributed to the SHA it was measured on |

Flash suite is 26 passed (was 23); `flash-selftest` and `mpcc-selftest` clean.

Two notes where your review found something I should have caught myself. The smoke-CLI gap is the *same* defect the previous round fixed in pytest — I fixed it in the file I was editing and left it standing in the entry point the README and Makefile actually tell people to run. And the `#1147`/`#1148` coordination update was posted to #776 the day before I started; I read to the 09-02 comment and stopped, then baked a closed issue into generated artifacts.

One thing I did not change, flagged rather than silently left: the notebook's stamp necessarily names the *parent* of the commit that carries the notebook, since the tree must be clean at execution time and committing the result advances HEAD. `f49eaaf` is that parent and is clean — a reader can check it out, rerun, and reproduce the saved numbers. If you would rather the notebook be excluded from its own provenance requirement, or want the stamp to record "executed against" explicitly, say which and I will make it say so.

---

bernalde:
@jkitchin — one Blocking regression-test gap remains before this review can be cleared.

The production behavior is fixed, but the negative cases are tested only by calling `smoke_contract_failures()` directly. The real `run_smoke()` test is positive-only. On the current head, replacing

```python
reasons = smoke_contract_failures(rec, route)
```

with `reasons = []` still leaves all 26 tests passing, so the CLI-to-contract binding can regress silently.

Please push a negative integration test in `python/tests/test_flash_mpcc.py` that:

- calls the real `flash_run.run_smoke()`;
- stubs only the expensive `flash_run.solve_route` boundary to return a doctored copy of a real successful `SolveRecord`;
- covers `finish_applied=False`, `validation={}`, and `finish_lowering="prod_eq"`;
- asserts exit code `1` and the corresponding failure text for each case.

I implemented and mutation-checked that shape locally against `03aa78c`: the unmodified module reports **29 passed**; bypassing the helper call reports exactly **3 failed, 26 passed**, with the three new negative cases failing on `run_smoke()` returning 0.

After that commit, please refresh the PR body for the final head: head SHA, commit count, diff totals, **29 flash cases**, and the exact current-head CI totals. The body currently still says `74ee506`, 10 commits, `+5753/−6`, and 26 assertions.

I could not push the prepared commit directly because the authenticated reviewer account does not have write access to `jkitchin/pounce`.

jkitchin:
@bernalde — pushed in `26b803d`, and your mutation numbers reproduce exactly.

## The gap was real

I ran your mutation before writing the test rather than after, so the "before" number is measured and not inferred:

| tree | result |
|---|---|
| unmodified | **29 passed** |
| `reasons = smoke_contract_failures(rec, route)` → `reasons = []` | **3 failed, 26 passed** |

Same as you measured locally, and the three failures are exactly the new cases, each failing on `run_smoke()` returning 0.

You named the shape precisely: the positive test drove `run_smoke` and asserted exit 0, the negative test called `smoke_contract_failures` directly, and *nothing joined them*. That is the same defect this fixture exists to pin — a check that is present, correct, and not wired to the thing that reports. It is also the second time on this PR that I fixed a contract in the file I was editing and left its binding to the entry point untested; the last round was the same mistake one layer out.

## `test_the_smoke_cli_rejects_a_bad_record`

Built to the shape you asked for — real `flash_run.run_smoke()`, only `flash_run.solve_route` stubbed, three parameters over `finish_applied=False`, `validation={}`, and `finish_lowering="prod_eq"`, each asserting exit code 1 and its reason text.

Two things beyond the spec, both to stop the test from passing for the wrong reason:

- The stub returns the doctored record at **300 K only** and the genuine one at the other four, and the test asserts that only `T=300.0` is reported as failing. A `run_smoke` that failed everything would satisfy an exit-code assertion just as well, and I would rather not find that out later.
- It asserts `solve_route` was reached once per smoke temperature, so a `run_smoke` that stopped iterating fails here rather than passing vacuously — and it re-checks that the *undoctored* record passes the contract, since otherwise the parametrization proves nothing.

## PR body

Refreshed for `26b803d`: 12 commits, 24 files, +5825/−6, 29 flash assertions, and a fourth row in the mutation table for this gap with your finding credited at the site.

Current-head CI: **1354 passed, 28 skipped** in 821 s (`Python tests (jax)`, run [33905161611](https://github.com/jkitchin/pounce/actions/runs/33905161611)), all eight checks green on `26b803d`. The 1348 → 1354 move is the two flash increments and nothing else — 23 → 26 → 29 assertions; merging `main` for #904 added no Python tests. `flash-selftest` and `mpcc-selftest` both still pass.

No Rust changed, so nothing about the solver's behaviour moves.

---

bernalde:
## Final review resolution

All review findings are addressed at final head `26b803d`.

- All three inline review threads are resolved.
- The exact-head flash module passes **29/29** tests.
- Replacing the `run_smoke()` contract call with `reasons = []` produces exactly **3 failed, 26 passed**: the three negative CLI cases fail because the doctored record is accepted.
- Restoring the contract call returns the module to **29/29** passing.
- Final-head CI is green across all eight jobs; the Python/JAX job reports **1354 passed, 28 skipped** in [run 33905161611](https://github.com/jkitchin/pounce/actions/runs/33905161611).

That closes the review gap: the production behavior, the contract itself, and the CLI-to-contract binding are all pinned. No further PR-specific action remains.

The reduced DiscOpt GDP/SOS1 comparison is separate follow-up work under #776. Its prerequisites, [DiscOpt #1147](https://github.com/jkitchin/discopt/issues/1147) and [DiscOpt #1148](https://github.com/jkitchin/discopt/issues/1148), are now closed.

PR #905, opened 2026-09-05
Expose presolve and restoration APIs

GermanHeim:
## Problem

While bumping to pounce-rs v0.11 in oximo, I found that that I forgot to expose the presolve and restoration APIs

## Cause

I forgot

## Fix

Expose presolve and restoration APIs

## Blast radius

pounce-rs

## Tests

- [ ] `CHANGELOG.md` `[Unreleased]` entry, in the user's terms
      since moved

### DIFF
--- a/crates/pounce-rs/src/lib.rs
+++ b/crates/pounce-rs/src/lib.rs
@@ -215,6 +215,10 @@ pub use pounce_presolve::{
 };
 
 /// Low-level presolve APIs for callers that drive a [`TNLP`] directly.
+/// [`IpoptApplication::optimize_tnlp`] applies `wrap_from_options` itself when
+/// `presolve=yes`; after wrapping manually, call
+/// [`IpoptApplication::set_presolve_already_applied`] with `true` to avoid a
+/// second wrapper.
 pub mod presolve {
     pub use pounce_nlp::expression_provider::ExpressionProvider;
     pub use pounce_presolve::{
--- a/crates/pounce-rs/src/lib.rs
+++ b/crates/pounce-rs/src/lib.rs
@@ -210,14 +210,20 @@ pub use pounce_algorithm::application::IpoptApplication;
 
 // --- presolve ---------------------------------------------------------------
 pub use pounce_nlp::expression_provider::{FbbtOp, FbbtTape};
-pub use pounce_presolve::fbbt::FbbtReport;
+pub use pounce_presolve::{
+    AuxiliaryPreprocessingDiagnostics, CachedBounds, LicqVerdict, TightenReport, fbbt::FbbtReport,
+};
 
 /// Low-level presolve APIs for callers that drive a [`TNLP`] directly.
 pub mod presolve {
     pub use pounce_nlp::expression_provider::ExpressionProvider;
     pub use pounce_presolve::{
-        PresolveError, PresolveOptions, PresolveTnlp, wrap_with_presolve_provider,
+        AuxiliaryPreprocessingDiagnostics, CachedBounds, LicqVerdict, PresolveError,
+        PresolveOptions, PresolveTnlp, TightenReport, wrap_from_options, wrap_with_presolve,
+        wrap_with_presolve_provider,
     };
+
+    pub use pounce_presolve;
 }
 
 /// Second-opinion recovery for callers that manage the low-level solve loop.
@@ -225,6 +231,8 @@ pub mod restoration {
     pub use pounce_restoration::second_opinion_driver::{
         SecondOpinionOutcome, run_second_opinion_ladder,
     };
+
+    pub use pounce_restoration;
 }
 
 // --- iteration capture & observability --------------------------------------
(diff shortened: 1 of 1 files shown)


jkitchin (commented):
Right instinct — the restoration half is essentially complete as-is. The presolve half needs items 1 and 2 before merge: as it stands it exports a wrapper whose reports can't be read, and steers callers onto the one construction path that drops a phase.

It compiles (`cargo check -p pounce-rs --all-features`), `cargo fmt --check` is clean, and no new clippy warnings. Everything below is checked against `485dbd7`.

## 1. The escape hatch is missing — and it's what makes `presolve` usable

`docs/src/rust.md` §"Escape hatch" promises: *"Each feature module also re-exports the crate behind it — `pounce_rs::convex` re-exports `pounce_convex`, and so on."* That's real: `convex.rs:185`, `qp.rs:82`, `sensitivity.rs:97`. Neither new module does it, and `lib.rs`'s "underlying crates" block doesn't list `pounce_presolve` / `pounce_restoration` either. Compiled against this branch:

```
error[E0425]: cannot find type `LicqVerdict` in module `pounce_rs::presolve`
error[E0425]: cannot find type `TightenReport` in module `pounce_rs::presolve`
error[E0433]: failed to resolve: could not find `pounce_presolve` in `pounce_rs`
```

Those are the return types of the accessors that are the *reason* to hold a concrete `PresolveTnlp`: `licq_verdict() -> Option<&LicqVerdict>`, `tighten_report() -> TightenReport`, `auxiliary_diagnostics() -> AuxiliaryPreprocessingDiagnostics`, `cached_bounds() -> Option<&CachedBounds>`, `certified_infeasible() -> Option<InfeasibilityProof>`. A facade-only caller can `{:?}` them and nothing else — can't bind, match, or store one. (`InfeasibilityProof` is reachable, but only via `pounce_rs::pounce_nlp::tnlp`, not the curated surface.)

Suggested fix: add `pub use pounce_presolve;` / `pub use pounce_restoration;` to the two modules, and lift `LicqVerdict`, `TightenReport`, `AuxiliaryPreprocessingDiagnostics`, `CachedBounds` into the curated list.

## 2. `wrap_with_presolve` isn't exported, and the alternative silently drops Phase 6

Only `wrap_with_presolve_provider` ships — the variant whose own doc comment says *"Callers who have the concrete inner TNLP type (`pounce-cli` with `NlTnlp`) should prefer this."* A caller driving a plain `TNLP` has no `ExpressionProvider`, so they reach for `PresolveTnlp::new`, which is the one construction path that does **not** call `stack_linear_eq_elim` (`crates/pounce-presolve/src/lib.rs:147-176`).

Measured on the Phase-6 acceptance fixture from `crates/pounce-algorithm/tests/presolve_phase6_linear_eq_elim.rs`, with `linear_eq_reduction = true`:

```
PresolveTnlp::new  ->  n = 4   (no column reduction)
wrap_with_presolve ->  n = 2
```

So `presolve_linear_eq_reduction=yes` becomes a silent no-op. That is precisely what `the_public_wrapper_stacks_the_reduction_too` was written to prevent ("the composition path every *library* caller takes … has to reach the same reduced problem"), and the facade currently offers users only the shape that has the bug. Exporting `wrap_with_presolve` and `wrap_from_options` needs nothing new — `wrap_from_options` takes `&OptionsList`, which comes from `app.options()`.

## 3. No surface test

`tests/{convex,qp,sensitivity,sqp}_surface.rs` exist for gh#561 exactly so a re-export that stops resolving or changes meaning fails in CI. Nothing in this PR compiles against `pounce_rs::presolve::*` or `pounce_rs::restoration::*`, so a rename in `pounce-presolve` would break downstream silently. A throwaway `presolve_surface.rs` (wrap + solve + read reports, facade paths only) passes as written — it's a short file. This is also what the unchecked *"Tests fail on the parent commit"* box wants: for a re-export, the parent-commit failure is the surface test failing to compile.

## 4. CHANGELOG + docs (CONTRIBUTING items 2 and 3)

- `[Unreleased]` bullet naming the `pounce-rs` surface.
- `docs/src/rust.md` — the "Escape hatch" paragraph is now false for these two modules, and neither surface is mentioned anywhere.
- Worth one sentence of protocol: `IpoptApplication::optimize_tnlp` already applies `wrap_from_options` itself when `presolve=yes` (`crates/pounce-algorithm/src/application.rs:1343`), so a caller who wraps manually needs `set_presolve_already_applied(true)`. Not obvious, and it's the first thing a user of this module hits.

## 5. Nits

- The other four facets each live in their own `src/*.rs` with a doc header and a runnable example; these are inline one-liners in `lib.rs`. Also `FbbtReport` stays at the crate root while the rest of presolve moves into a module, so the same subject is now split across two paths.
- The `restoration` doc could say that `builder::Nlp::solve` already runs the ladder (`builder.rs:520`), so this export is for the `IpoptApplication` path specifically — and that the driver restores the options each rung writes.

---

Issue #804, opened 2026-08-27
Sensitivity: a constraint row's activity ratio is also normalized by an un-reduced curvature (gh #763 for rows)


jkitchin:
## Summary

gh#763 established that a variable's activity ratio must be normalized by the
curvature **reduced** along that coordinate, not by the Hessian diagonal, and
added `reduced_activity` to answer the question the class cannot. The
constraint-row path has the same defect and currently carries only a docs
note.

## The distinction

`classify_activity` normalizes a row's `Sigma` by a **directional** curvature,

```
r = Sigma_j / (grad_d^T H grad_d / ||grad_d||^2)
```

which is a genuine curvature along the row's own gradient direction — and so
is strictly better than the variable path's bare `H_ii`, which is why this was
not the one gh#763 fixed. But it is still not a *reduced* curvature: it does
not account for the other free coordinates re-optimizing. The quantity that
generates a row's multiplier is what is left after that elimination, exactly as
for a variable.

So a row's `r` is, by the same algebra, `reduced/directional`, equal to `1`
only where the row's direction is decoupled from the remaining free space. A
genuine row kink whose direction is coupled falls out of the `[1e-1, 1e1]`
band and reads `AMBIGUOUS` — at any tolerance, because the ratio is
`mu`-independent, so re-solving tighter reports the same thing.

## Why it matters

Same three reasons gh#763 gave for the variable path:

1. `row_status` / `row_ratio` are public and reachable from Python.
2. `AMBIGUOUS` reads as "not sure, re-solve tighter", and for this case that
   advice is wrong — no tolerance recovers it.
3. The inference "not `WEAKLY_ACTIVE`, so probably not a kink" is what shipped
   gh#756's first-order wrong derivative. The `CLAUDE.md` note added by
   gh#763 warns against it for variables; a reader has no reason to think rows
   are different, and right now the codebase does not tell them either way in
   any load-bearing place.

This is **not** a wrong answer today in the step path: `weakly_active_bounds()`
counts `AMBIGUOUS` as weak, so `parametric_step_directional` still returns the
correct derivative. What is wrong is the public *answer*, exactly as in
gh#763.

## Suggested shape

Mirror `reduced_activity`: an on-demand accessor over requested rows, one
back-solve each, normalizing by the reduced curvature along `grad_d` rather
than by `grad_d^T H grad_d / ||grad_d||^2`. The variable version's
`q_i = 1/(K^-1)_ii - Sigma_i` has a row analogue through the same held factor;
[...]

PR #912, opened 2026-09-05
feat(sens): dlambda/dp for strictly active inequality constraints

### LINKED ISSUE #910: sens_jacobian gives dlambda/dp for equality rows only; a strictly active inequality has the same well-defined derivative
`sens_jacobian(of=<Constraint>)` gives `dλ/dp` for equality rows only. An
inequality target is refused outright, even when the row is strictly active
at the solved point — the case where the multiplier is a perfectly ordinary
differentiable function of the parameters and the derivative is the same
backsolve an equality gets.

That matters because the shadow price a user actually wants an error bar on
is usually a *cap*: a safety limit, a capacity, a purity spec. Those are
written as inequalities. Today the only way to get `dλ/dp` for one is to
know it is active and rewrite it as an equality `g(x) == b` before solving —
which changes the model, and is only correct if the row really does stay
active over the perturbation being asked about.

## Why the restriction exists

It is API surface, not mathematics. The compound KKT vector is
`(x, s, y_c, y_d, z_l, z_u, v_l, v_u)`, and `parametric_step_full` already
returns the whole thing including `y_d`. What is missing is the *map*: the
sensitivity layer exposes `g_multiplier_rows`, which finds the `y_c` row of
an equality via `BoundClassification::full_to_c`, and there is no counterpart
for `y_d`. `full_to_c` returns `-1` for an inequality, so the only thing the
layer can currently do with one is refuse.

## Why it cannot simply be lifted

An inequality multiplier is differentiable only where the row is *strictly
complementary*. There are three regimes:

| regime | slack, multiplier | `dλ/dp` |
|---|---|---|
| strictly active | `s ≈ 0`, `λ > 0` | behaves as an equality over a neighbourhood; well defined |
| inactive | `s > 0`, `λ ≈ 0` | zero over a neighbourhood — but there is no KKT row to differentiate |
| weakly active (kink) | `s ≈ 0` *and* `λ ≈ 0` | two one-sided derivatives that differ; no two-sided value exists |

The `y_d` entry holds a number in all three cases. In the third it is
whichever side the factorization happened to land on, which is exactly the
"silently wrong while reporting success" class. So the map has to come with
a gate.

## Which classifier gates it

`reduced_row_activity()`, not `classify_activity()` — and the reason is not
the obvious one.

Neither classifier can *admit* a kink. A row's reduced curvature is never
larger than its directional one, so the reduced ratio is never below the
directional ratio, and `strongly_active` is the high-ratio tail of both.

What the directional normalizer gets wrong is the **refusal reason**. The
ratio it reports for a genuine kink is `reduced / directional` (#804), so as
the row's coordinate couples to the rest of the free space at strength `rho`
the class slides. Measured, on a fixture whose row is a kink at every `rho`:

| `rho` | `classify_activity()` | `reduced_row_activity()` |
|---|---|---|
| `1` | `weakly_active` | `weakly_active` |
| `1e-2` | `ambiguous` | `weakly_active` |
| `1e-4` | `ambiguous` | `weakly_active` |
| `1e-6` | **`inactive`** | `weakly_active` |

`inactive` is the one class whose derivative a caller may legitimately read
as a structural zero. So gating on the directional ratio would not decline
to answer about a tight cap's shadow price — it would answer "it does not
move", which is a wrong answer wearing a refusal's clothes. Coupling that
strong is routine on a collocation model.

Reading an activity class as a proxy for kink-ness is the inference that
shipped #756.

## Proposed

1. `BoundClassification::full_to_d` — the O(1) inverse of `d_map` and the
   exact complement of `full_to_c`.
2. `TNLP::full_g_to_d_block` (defaulting to `None`) → `AlgorithmBacksolver`
   → `Solver::d_multiplier_rows`, mirroring `g_multiplier_rows`, plus the
   PyO3 binding.
3. The gate in `pounce.sensitivity`'s `mult_entry`: fall through to the `y_d`
   block, then require `reduced_row_activity()` to report `strongly_active`.
   The other two regimes are refused with messages that *name the regime*,
   since they call for different things from the caller — an inactive row's
   answer really is zero, and a kink needs a direction (`sens_solution()` /
   `parametric_step_full` still report a one-sided step for one).
4. A third fixture in `sens_invariance_legs.rs` with a row in each of the
   three legs, per the "a new public accessor gets a row in each leg" rule
   in CLAUDE.md — with the pin equality first and an inactive decoy second,
   so the active row's g index, its d-block position, and its (nonexistent)
   c-block position are three different numbers.

jkitchin:
`sens_jacobian(of=<Constraint>)` gave `dλ/dp` for equality rows only. An
inequality target was refused outright, even where the row is strictly active
at the solved point — the case where the multiplier is an ordinary
differentiable function of the parameters and the derivative is the same
backsolve an equality gets, one block over.

That is the case users reach for. The shadow price you want an error bar on is
usually a **cap**: a safety limit, a capacity, a purity spec. Those are written
as inequalities. The only workaround was to know the row was active and rewrite
it as `g(x) == b` before solving — which changes the model, and is only correct
if the row really does stay active over the perturbation being asked about.

Fixes #910

## What was actually missing

API surface, not mathematics. The compound KKT vector is
`(x, s, y_c, y_d, z_l, z_u, v_l, v_u)`, and `parametric_step_full` already
returns all of it, `y_d` included. What was missing was the **map**:
`BoundClassification::full_to_c` returns `-1` for an inequality and there was no
counterpart, so the only thing the layer could do with one was refuse.

| layer | added |
|---|---|
| `pounce-nlp` | `BoundClassification::full_to_d` — the O(1) inverse of `d_map`, the exact complement of `full_to_c` |
| `pounce-nlp` | `IpoptNlp::full_g_to_d_block` (default `None`), overridden by `OrigIpoptNlp` |
| `pounce-sensitivity` | `PdSensBacksolver::full_g_to_d_block`, `Solver::d_multiplier_rows` |
| `pounce-py` | the `d_multiplier_rows` PyO3 binding |
| `python/pounce/sensitivity` | the gate, in `_session.py`'s `mult_entry` |

**Not a trajectory change.** Nothing on the solve path reads `full_to_d`;
`classify_bounds` gains one `vec![-1; ng]` and one store per inequality row, and
no iterate moves. The only reader is the new accessor
(`grep -rn 'full_to_d\|full_g_to_d_block\|d_multiplier_rows' --include='*.rs'`
returns the definitions plus one call site in `_session.py`). No fixture sweep.

## The three regimes, and why two are refused

An inequality's multiplier is differentiable only where the row is *strictly
complementary*.

| regime | slack, multiplier | `dλ/dp` |
|---|---|---|
| strictly active | `s ≈ 0`, `λ > 0` | behaves as an equality over a neighbourhood; **answered** |
| inactive | `s > 0`, `λ ≈ 0` | zero over a neighbourhood — but there is no KKT row to differentiate, and a returned `0.0` would be indistinguishable from a computed one; **refused, naming the regime** |
| weakly active (kink) | `s ≈ 0` *and* `λ ≈ 0` | two one-sided derivatives that differ; **refused** |

The `y_d` entry holds a number in all three cases. In the third it is whichever
side the factorization happened to land on — the "silently wrong while reporting
success" class, with nothing to warn on. So the map ships with a gate, and the
refusals name the regime, because the three call for different things from the
caller: an inactive row's answer really is zero, and a kink needs a *direction*
(`sens_solution()` and `parametric_step_full` still give a one-sided step for
one).

## Which classifier gates it — measured, not argued

`reduced_row_activity()`, not `classify_activity()`. The reason is not the one
it looks like.

**Neither classifier can *admit* a kink.** A row's reduced curvature is never
larger than its directional one, so the reduced ratio is never below the
directional ratio, and `strongly_active` is the high-ratio tail of both.

What the directional normalizer gets wrong is the **refusal reason**. The ratio
it reports for a genuine kink is `reduced / directional` (#804), so as the row's
coordinate couples to the rest of the free space at strength `rho` the class
slides. Swept on a fixture whose row is a kink at every `rho`:

| `rho` | `classify_activity()` | `reduced_row_activity()` |
|---|---|---|
| `1` | `weakly_active` | `weakly_active` |
| `1e-2` | `ambiguous` | `weakly_active` |
| `1e-4` | `ambiguous` | `weakly_active` |
| `1e-6` | **`inactive`** | `weakly_active` |

`inactive` is the one class whose derivative a caller may legitimately read as a
structural zero. So gating on the directional ratio would not *decline* to
answer about a tight cap's shadow price — it would answer "it does not move",
which is a wrong answer wearing a refusal's clothes. Coupling that strong is
routine on a collocation model. Reading an activity class as a proxy for
kink-ness is the inference that shipped #756.

That sweep is `test_a_coupled_row_kink_is_refused_where_classify_reads_it_INACTIVE`,
with `test_the_coupled_fixture_is_a_kink_at_every_coupling_it_is_asked_at` as its
premise so it cannot pass on a fixture that stopped being a kink.

## One asymmetry, on purpose

`d_multiplier_rows` **errors** on an out-of-range index where
`g_multiplier_rows` returns `None`. With both accessors present, `None` from the
first means "ask the other", which is exactly the fall-through `mult_entry`
performs — so if the second also answered `None` for a row that does not exist,
a caller could not tell "that row is an equality" from "that row does not
exist", and the message it prints would name the wrong reason. Same shape as
`x_primal_rows`, whose comment already says it: out of range must not
masquerade as "removed as fixed". Pinned by
`an_out_of_range_row_is_an_error_not_an_equality`.

While there: `mult_entry`'s old both-`None` message claimed "the constraint has
no finite bound on either side". That is false — `classify_bounds` puts every
row in exactly one block, a free row included (it lands in `d`, and the gate
then refuses it as inactive, which is the truth: a free row's multiplier is zero
always). It now says what it is, an unreachable assertion.

That premise is pinned where it is decided, in `tnlp_adapter.rs`, not asserted
downstream: `every_constraint_row_lands_in_exactly_one_of_c_and_d` (over
equality / `<=`-only / `>=`-only / range / free rows),
`a_free_row_gets_a_d_slot_with_no_bound_on_either_side`, and
`full_to_d_is_the_positional_inverse_of_d_map`. It has to live there because
**Pyomo cannot construct a free constraint at all** — `pyo.inequality(None, x,
None)` raises — so no end-to-end model reaches the branch, and the stub covering
it can only show what happens *if* both accessors answer `None`, never that
nothing makes them.

## Tests

**A third fixture in `sens_invariance_legs.rs`**, per the "a new public accessor
gets a row in each leg" rule in `CLAUDE.md`:

```
min  0.5x² + 0.5y² − x − y     [+ 0.5(xf − 0.7)²]
s.t. g0:  p == 0                (the pin)
     g1:  x − y <= 5            (never active)
     g2:  x + y − p <= 1        (strictly active, λ = 1/2)
```

The pin equality goes **first** and an inactive decoy **second**, so the active
row's g index (2), its d-block position (1) and its c-block position (which does
not exist) are three different numbers. `dλ/dδ = −1/2` exactly, from the model —
an outside number, not one the machinery produced.

Seven tests: a precondition asserting the branch each leg runs on (and pinning
both d rows to their exact flat positions, so a permuted `d_map` fails even
though only one row is active), an affine-in-`delta` companion, a row in each of
the three legs, the corner where two compose, and a liveness witness.

**Mutation table** (in the fixture header, each introduced and reverted):

| mutation | result |
|---|---|
| `full_g_to_d_block` returns `Some(full_idx)` | 18 passed, 6 failed |
| `d_multiplier_rows`' `y_d_offset` drops `dims[2]` | 18 passed, 6 failed — the same six |

The six are every test in the section except the liveness witness, which stays
green under both by construction: it reads `nlp_scaling()` / `variable_scaling()`,
which do not go through the map under test, so it can still testify that the
scaled arm really was scaled while the map is broken. (That separation was
earned: the first two drafts of the witness compared the raw
`parametric_step_full` value — identical to 16 digits in both arms — and then
`row_sigma`, which differs by 3.6e-6 relative. Everything the sensitivity layer
returns is already frame-converted, so `nlp_scaling()` is the only honest
witness.)

**The Rust legs reach only the strictly complementary branch, by design** — the
fixture's multiplier is `1/2`. Per the branch rule, the other two branches are
covered end to end in `pyomo-pounce/tests/test_sens.py`: a strictly active
multiplier against a finite difference, the equality-rewrite agreeing with the
inequality answer, an inactive row refused, a decoupled kink refused, and the
coupled kink above. Five stub tests drive the regime dispatch directly.

## Verification

- `cargo test --workspace --exclude pounce-hsl` — green. (`pounce-hsl`'s
  `ma57_batched_backsolve` fails on this machine for want of a working MA57;
  confirmed identical on a stashed tree, and `pounce-hsl` does not depend on
  anything this PR touches.)
- `cargo clippy --workspace --all-targets` — exit 0, no new warnings.
- `cargo fmt --all -- --check` — clean.
- `python/tests` — 1446 passed. `pyomo-pounce/tests` — 511 passed, 11 skipped.
- **Which binary:** the in-tree `python/pounce/_pounce.abi3.so`, rebuilt with
  `maturin develop --release` after the last Rust edit. `conftest.py`'s
  staleness guard fired once (rustfmt bumped the source mtimes) and was
  **obeyed** — `POUNCE_SKIP_EXT_STALE_CHECK` was never set.

## `/sens-review`

| # | class | verdict | evidence |
|---|-------|---------|----------|
| 1 | index space | PASS | the whole diff is an index map, and all three new conversions feed direct array indexing. Fixture 3 makes g index, d position and c position three different numbers; `leg_fixed_*` puts a `make_parameter`-removed variable ahead so full-x/var-x diverge too; two mutations each fail exactly six tests |
| 2 | frame and scaling | PASS | three factors meet in the multiplier (row `dg`, objective `df`, the pin's `delta`); `leg_scaling_the_multiplier_derivative_is_unmoved_by_the_change_of_variables` plus a witness that reads `nlp_scaling()`. Nothing is compared against `Sigma` |
| 3 | absolute thresholds | PASS | no new constant in `src/`; the gate reuses `reduced_row_activity`'s `sqrt(mu)`-relative classification. Test side compares **slopes**, never `step/delta`, and a null predictor scores `0.0` against an exact `−0.5` — five orders outside the budget at every `delta` in the sweep |
| 4 | doc drift | PASS | `g_multiplier_rows`' doc said the `y_d` mapping "is not exposed here" — this PR makes that false, corrected. `mult_entry`'s both-`None` message was false, corrected. My own first draft of the gate rationale claimed a coupled kink reads `ambiguous` "at any tolerance"; measurement showed it slides to `inactive`, and all four copies now carry the sweep |
| 5 | silently wrong while reporting success | PASS, with the limit named | this is the class the diff creates — a `y_d` entry at a kink is one arbitrary side, with no residual to warn on. The gate is the mitigation, and the coupled-kink test asserts the refusal on the branch where the naive gate would answer wrongly. The fixture's `−1/2` is an outside number (closed form from the model), but it is a 3-variable model: **no row was added to `sens_resolve_oracle.rs`**, so nothing here is evidence about the multiplier step at scale |
| 6 | which branch does the fixture take | PASS | the rule branches three ways; the Rust fixture reaches one, and the other two are covered in `test_sens.py` — including a premise sweep asserting the coupled fixture is still a kink at every `rho` it is asked at |
| 7 | name the measured populations | PASS | the discriminator is *which classifier gates*, and the `rho` table above is the measurement, reproduced in #910, the CHANGELOG, the docs and the test |
| 8 | which binary | PASS | in-tree `.so`, rebuilt; staleness guard obeyed, never bypassed |

The one RISK-shaped item is entry 5's: a resolve-oracle row for the `y_d` block
is the natural follow-up, and I did not add one.

## Note for whoever merges second

Notebook 39 (`39_design_under_kinetic_uncertainty.ipynb`) lives on
#908's branch, not on `main`, and its section 10 says:

> **Multiplier sensitivities are available for equality constraints only.** A
> variable bound or an inequality has no dλ/dp row to read.

Half of that is now wrong: a **strictly active inequality** does have one; a
variable bound still does not. Whichever of the two PRs lands second should
update that bullet — the section's `m.T == T_cap` rewrite is still a fine
illustration of pinning a known-active row, it is just no longer the only way
to get the number.

https://claude.ai/code/session_013bUD9V6h3GkS8E1PnBkz8p

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,53 @@ changes.
 
 ### Added
 
+- **Multiplier sensitivities on inequality constraints
+  (`sens_jacobian(of=<inequality Constraint>)`, gh#910).** `dλ/dp` was
+  equality-only:
+  an inequality target raised, and the workaround was to rewrite a
+  known-active cap as an equality. It now answers wherever the row is
+  *strictly complementary* — active with a multiplier bounded away from
+  zero — which is the whole region where the multiplier is a
+  differentiable function of the parameters. New plumbing:
+  `Solver::d_multiplier_rows` (Rust and Python), backed by a `full_to_d`
+  map on `BoundClassification` that is the exact complement of the
+  existing `full_to_c`.
+
+  The other two regimes are **refused, each naming itself**, because they
+  need different things from the caller. An *inactive* row's multiplier is
+  pinned at zero over a neighbourhood, so its derivative genuinely is zero
+  — but there is no KKT row to differentiate, and a returned `0.0` would
+  be indistinguishable from a computed one. At a *kink* (`λ ≈ 0` and slack
+  `≈ 0` together) the two one-sided derivatives differ, so no two-sided
+  `dλ/dp` exists; `sens_solution()` and `Solver.parametric_step_full` still
+  give a one-sided step for a chosen direction.
+
+  The gate is `reduced_row_activity()`, not `classify_activity()`, and the
+  reason is not the one it looks like. Neither classifier can *admit* a
+  kink: a row's reduced curvature is never larger than its directional
+  one, so the reduced ratio is never below the directional ratio, and
+  `strongly_active` is the high-ratio tail of both. What the directional
+  normalizer gets wrong is the **refusal reason**. Swept over a genuine row
+  kink coupled to the rest of the free space at strength `ρ`,
+  `classify_activity()` reads `weakly_active` at `ρ = 1`, `ambiguous` by
+  `1e-2`, and `inactive` by `1e-6`, while `reduced_row_activity()` holds
+  `weakly_active` throughout. `inactive` is the one class a caller may
+  legitimately read as a structural zero — so the directional gate would
+  not decline to answer about a tight cap's shadow price, it would answer
+  "it does not move". Coupling that strong is routine on a collocation
+  model.
+
+  Covered by a third fixture in
+  `crates/pounce-sensitivity/tests/sens_invariance_legs.rs` — a strictly
+  active cap behind an equality pin and an inactive decoy, so the row's g
+  index, its d-block position and its (nonexistent) c-block position are
+  three different numbers — with a row in each of the three legs, its own
+  mutation table, and a liveness witness that does not go through the map
+  under test. Two mutations of the new map each fail exactly the six
+  multiplier tests and leave the witness green. End-to-end regime coverage,
+  including a coupled kink that `classify_activity` misreads as `inactive`,
+  is in `pyomo-pounce/tests/test_sens.py`.
+
 - **A phase-changing flash as a complementarity problem
   (`pounce.examples.flash_mpcc`, notebook 38).** A single vapour–liquid
   equilibrium stage solved across a temperature path that crosses
--- a/crates/pounce-nlp/src/ipopt_nlp.rs
+++ b/crates/pounce-nlp/src/ipopt_nlp.rs
@@ -389,6 +389,21 @@ pub trait IpoptNlp: Nlp {
         Some(full_idx)
     }
 
+    /// Map a 0-based **full-g** index to a 0-based position in the
+    /// d-block (algorithm-side inequality multiplier vector `y_d`,
+    /// length `m_ineq()`). Returns `None` when the constraint is an
+    /// equality (lives in `c`, not `d`).
+    ///
+    /// The exact complement of [`Self::full_g_to_c_block`]: every row
+    /// is in one block or the other, so exactly one of the two returns
+    /// `Some` for an in-range index. Default impl matches
+    /// `full_g_to_c_block`'s (which assumes every row is an equality)
+    /// by returning `None` for all of them; `OrigIpoptNlp` overrides
+    /// via `BoundClassification::full_to_d`.
+    fn full_g_to_d_block(&self, _full_idx: Index) -> Option<Index> {
+        None
+    }
+
     /// Inverse of [`Self::full_x_to_var_x`]: map a 0-based var-x index
     /// (length `n()`) to the corresponding full-x index (length
     /// `n_full_x()`). Used when scattering a compressed step or
--- a/crates/pounce-nlp/src/orig_ipopt_nlp.rs
+++ b/crates/pounce-nlp/src/orig_ipopt_nlp.rs
@@ -2780,6 +2780,17 @@ impl IpoptNlp for OrigIpoptNlp {
         if c < 0 { None } else { Some(c) }
     }
 
+    fn full_g_to_d_block(&self, full_idx: Index) -> Option<Index> {
+        let cls = self.adapter.borrow();
+        let cls = cls.classification();
+        let f = full_idx as usize;
+        if f >= cls.full_to_d.len() {
+            return None;
+        }
+        let d = cls.full_to_d[f];
+        if d < 0 { None } else { Some(d) }
+    }
+
     fn var_x_to_full_x(&self, var_idx: Index) -> Index {
         let cls = self.adapter.borrow();
         let cls = cls.classification();
--- a/crates/pounce-nlp/src/tnlp_adapter.rs
+++ b/crates/pounce-nlp/src/tnlp_adapter.rs
@@ -91,6 +91,15 @@ pub struct BoundClassification {
     /// Maps full-g index → c-block position, with `-1` for inequality
     /// rows: the O(1) inverse of `c_map`, mirroring `full_to_var`.
     pub full_to_c: Vec<Index>,
+    /// Maps full-g index → d-block position, with `-1` for equality
+    /// rows: the O(1) inverse of `d_map`, and the exact complement of
+    /// `full_to_c` — every row is in one block or the other, so
+    /// `(full_to_c[i] < 0) == (full_to_d[i] >= 0)` for every `i`.
+    ///
+    /// Added so a caller holding a user-space row index can find the
+    /// inequality multiplier `y_d` in the compound KKT vector, which
+    /// `full_to_c` alone could only refuse to do (pounce#910).
+    pub full_to_d: Vec<Index>,
 }
 
 impl BoundClassification {
@@ -560,6 +569,7 @@ fn classify_bounds(
     let mut d_l_map: Vec<Index> = Vec::new();
     let mut d_u_map: Vec<Index> = Vec::new();
     let mut full_to_c: Vec<Index> = vec![-1; ng];
+    let mut full_to_d: Vec<Index> = vec![-1; ng];
 
     for i in 0..ng {
         let lo = g_l[i];
@@ -589,6 +599,7 @@ fn classify_bounds(
             }
         }
         let d_idx = d_map.len() as Index;
+        full_to_d[i] = d_idx;
         d_map.push(i as Index);
         if lo_present {
             d_l_map.push(d_idx);
@@ -618,6 +629,7 @@ fn classify_bounds(
         d_l_map,
         d_u_map,
         full_to_c,
+        full_to_d,
     })
 }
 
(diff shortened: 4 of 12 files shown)


jkitchin:
Superseded by #913, which fixes the same issue and is the better of the two.

Both PRs land the same core design — `BoundClassification::full_to_d` ->
`IpoptNlp::full_g_to_d_block` -> `Solver::d_multiplier_rows` -> a PyO3
binding, gated one layer up in `mult_entry` on strict complementarity, with a
third fixture in `sens_invariance_legs.rs`. #913 goes further in four places:

- it reroutes `reduced_row_activity`'s own ascending full-g -> d-block scan
  through the new map instead of leaving a second scan that agrees today. The
  gate classifies with that function and then reads this row, so a drift
  between the two would classify one row and answer about another;
- its refusal messages name the regime, because an inactive row and a kink
  want different things from the caller;
- it pins *which* classifier gates the decision with a parametrized rho sweep
  (`ambiguous` at 1e-2 sliding to `inactive` at 1e-6 on the cheap classifier),
  where this PR had the measurement in prose only;
- it adds nine tests to `python/tests/test_sensitivity_core.py`, so the
  surface is covered without Pyomo installed.

This branch was also cut before #908 merged and had gone `DIRTY` against main.

Nothing here is lost: the free-row case (`-inf <= g <= inf` landing in `d`
with empty `d_l_map`/`d_u_map`) that this PR added a dedicated
`tnlp_adapter` test for is asserted in #913 too, folded into the existing
`classify_bounds` tests.

Closing unmerged. Fixed by #913.

PR #921, opened 2026-09-06
docs: notebook 43, Sobolev training on derivatives the solve already paid for

jkitchin:
Adds `python/notebooks/43_sobolev_training.ipynb` and its README row.

A neural surrogate of a non-isothermal PFR (series A -> P -> W with a cooling
jacket, trapezoid collocation, 252 variables) is trained two ways: on values
alone, and on values plus the parametric Jacobian that the converged solve can
return from its held KKT factorization. Both arms see **exactly the same
solves** -- the values-only arm just ignores the `J` array -- so the question
is not whether more data helps, it is whether the derivative labels that came
free with the solve are worth using.

### What is measured

| | |
|---|---|
| Jacobian by back-solve | **1.17x** surcharge on the solve |
| the same numbers by central FD | **7.54x** |
| agreement between the two routes | median 8.4e-08, max 7.6e-06 of channel scale |
| accuracy at equal budget (n=256) | **7.8x** on values, **6.1x** on derivatives |
| solves to reach the values arm's best | **4x** fewer |

Because the two label routes agree to four orders below the best surrogate
error, the finite-difference arm is not a separate accuracy curve -- it is the
same curve shifted along the cost axis, and the cost panel draws it that way.

### The weight between values and derivatives

This is the half the notebook was really built for, and it does not flatter the
usual story about negative-log-likelihood auto-weighting:

* **Unit invariance comes from the normalizer, not the loss.** Under a 1e9
  rescaling of the output units, all three losses move by roundoff
  (1.8e-14 / 2.3e-04 / 1.8e-02), well inside pool-to-pool spread.
* **Learned vs profiled sigma is optimization speed, not statistics.** They are
  the same objective. At 5000 steps the learned form's `logvar` is still 4.8
  log-units from its own stationarity condition sigma_k^2 = MSE_k; by 20000
  steps the two are indistinguishable.
* **At convergence the parameter-free NLL does not win.** A z-scored lambda=1 --
  one extra `std` -- beats it on both metrics. It is included precisely because
  it is the strong cheap baseline such comparisons usually omit.

What the NLL does buy is real and narrower than the pitch: channel *spread*
spans 13,395x while channel *difficulty* spans 36x, correlating at rho = -0.25.
A normalizer can equalize the first; only a loss watching its own residuals can
equalize the second.

### Negative results kept rather than dropped

* The design study's **violation** column shows both arms unsafe at n=16, with
  the Sobolev arm marginally the worse (31.5 K over the limit against 28.5 K).
  Sobolev training buys accuracy, not feasibility. The arms separate on regret
  (5.5% vs 0.5%) and on how fast the violation collapses with budget.
* A hypothesis that the hardest channel (dThot/dtau) was hard because of
  proximity to the no-hot-spot regime was **measured and rejected**: those
  points are 0.7% of the test box and carry 1.5% of that channel's squared
  error, correlating at rho = +0.16. It is simply the most nonlinear channel.

### Section 9: the label matters as much as the loss

A hard `max` over collocation nodes produces a hot-spot value that varies by
less than 0.004 K while its derivative sawtooths by up to 0.82 K per unit tau
across ten argmax switches. A values-only surrogate never notices; a Sobolev
one is fed the sawtooth as truth. That is a label failure no loss and no
architecture recovers from -- and taking a max over discretization nodes is the
obvious way to write "the hot spot". The notebook uses a mesh-independent
integral softmax instead, and Section 10 lists what none of this is evidence
about.

### Notes for review

* Executed end to end against pounce-solver 0.11.0 / jax 0.11.0 / optax 0.2.6;
  28/28 cells, no errors, ~12 minutes. Saved outputs match the committed source
  and the `SOURCE_COMMIT` stamp resolves.
* **Numbered 43, not 42** -- open PR #920 already adds
  `42_mixing_equations_and_jax.ipynb`.
* No solver code is touched: the notebook is pure documentation, drives the
  existing `pounce.sensitivity` and `pounce.jax` layers, and needs `jax` and
  `optax` (already notebook-only dependencies via nb 17).
* Uses a hand-rolled tanh MLP over `optax` rather than flax -- see the review
  comment below.

https://claude.ai/code/session_01N5CB3GErgSn6qJtXERRWbm


jkitchin:
### One open decision: flax

The notebook uses `optax` plus a ~15-line hand-rolled tanh MLP rather than
`flax`, and that is a deviation worth a deliberate call rather than a silent
one.

`flax` is not usable against the environment as it stands:

* installed `flax` 0.12.2 is broken against installed `jax` 0.11.0 -- both
  `flax.linen` and `flax.nnx` hit `jax.core` APIs that 0.11.0 removed;
* `flax` 0.12.9, which fixes that, requires `jax>=0.11.1`.

So adopting flax means upgrading `jax` in the shared virtualenv, which affects
every other notebook that imports it -- notably 17 and 35 -- and would want
those re-executed to confirm nothing moved.

The hand-rolled net is arguably the better fit here regardless: the whole point
is `jax.jacrev` of the network with respect to its *inputs*, and writing the
MLP as a pure function of an explicit parameter pytree makes that mechanically
obvious rather than routing it through a module system. It is also one less
dependency for a notebook whose subject is not the NN library.

Happy to switch to flax if the jax upgrade is wanted; it is a contained change
to two cells.

Issue #874, opened 2026-08-31
jax/torch QP frontends silently zero-pad a wrong-shaped Hessian (#865's _validate_p_shape landed in qp.py only)


jkitchin:
Found in the 2026-08-31 adversarial audit round (seam B). Full write-up:
`adversary/audit-2026-08-31/findings/seam-b-qp-second-order.md`.

`7dc03c66` added `_validate_p_shape` to reject a Hessian whose shape disagrees
with the rest of the model. It landed in `python/pounce/qp.py` only. The jax
and torch frontends do the same pre-build check as
`_check_psd(*_to_coo_lower(np.asarray(P)), n)` — the shape is already gone by
then — and neither has a `_validate`.

```
$ grep -rn "_validate_p_shape" python/
```

returns three lines, all in `qp.py`.

## Reproduction (verified at HEAD `629535fa`)

`P` is 3x3 while `c`, `lb`, `ub` are length 5:

```python
import numpy as np, pounce.qp, pounce.torch, pounce.jax
P = 2*np.eye(3); c = np.array([1.,1.,1.,2.,2.])
lb = -np.ones(5); ub = np.ones(5)

pounce.qp.solve_qp(P=P, c=c, lb=lb, ub=ub)
#  ValueError: solve_qp: `P` has shape (3, 3) but must be (5, 5)

pounce.torch.solve_qp(P=P, c=c, lb=lb, ub=ub)
#  -> [-0.5 -0.5 -0.5 -1.  -1. ]     NO ERROR
pounce.jax.solve_qp(P=P, c=c, lb=lb, ub=ub)
#  -> [-0.5 -0.5 -0.5 -1.  -1. ]     NO ERROR
```

`[-0.5, -0.5, -0.5, -1.0, -1.0]` is the exact optimum of the **zero-padded**
5x5 model, confirmed against a numpy oracle — so this is a silently
substituted different problem, not noise.

An oversized `P` gives a different failure again:

```python
pounce.torch.solve_qp(P=np.eye(5), c=np.ones(3), lb=-np.ones(3), ub=np.ones(3))
#  IndexError: index 3 is out of bounds for axis 0 with size 3
```

which is verbatim the error string `7dc03c66`'s own commit message quotes as
the thing being fixed.

## Why it matters more here than on the plain path

Both are **differentiable layers**. `_kkt_backward` inverts the padded KKT
system, so the gradients flowing back into a training loop are the padded
model's gradients. A user gets a silently wrong descent direction with no
[...]

Issue #113, opened 2026-06-09
solve_qp lacks input validation: shape mismatch -> 'primal_infeasible', NaN/Inf -> 'iteration_limit'


jkitchin:
## Summary

`pounce.qp.solve_qp` performs no input validation in the Python wrapper. Two
concrete gaps, both of which currently surface as a *misleading solver status*
rather than a clear up-front error:

1. **Shape mismatch** between `A`/`b` (or `G`/`h`) → `status="primal_infeasible"`.
2. **NaN/Inf in inputs** → `status="iteration_limit"` (or `numerical_failure`)
   with a `nan` solution.

**Severity:** S2 — both actively misdiagnose a user error (a wrong shape looks
like an infeasible *problem*; a `NaN` looks like the solver needs more
iterations).

## Reproduce

```python
import numpy as np
from pounce.qp import solve_qp

# (1) shape mismatch: 1x2 A with a length-2 b
solve_qp(np.diag([2.,2.]), np.zeros(2),
         A=np.array([[1., 1.]]), b=np.array([1., 2.]))
# -> status="primal_infeasible"  (should be a ValueError)

# (2) NaN in c
solve_qp(np.eye(2), np.array([np.nan, 1.0]))
# -> status="iteration_limit", x=[nan, nan]  (should be a ValueError)
```

## Expected

A single validation pass at the top of the wrapper:
- `A.shape[0] == b.shape[0]`, `G.shape[0] == h.shape[0]`, column counts == `n`;
- `np.isfinite(...).all()` on every array;
raising `ValueError` with a precise message on failure.

Found during a hands-on QA pass (`dev-notes/qa-lp-qp-torch.md`, findings F2/F8).

jkitchin:
Fixed in 2dca740.

## Summary of fix

`pounce.qp.solve_qp` (and every other entry point that goes through `_build`)
now runs a single up-front validation pass, so both reported gaps become a
precise `ValueError` instead of a misleading solver status.

### Before → after
- **Shape mismatch** (`A` 1×2, `b` length 2): was `status="primal_infeasible"`
  → now `ValueError: solve_qp: A has 1 rows but b has length 2`.
- **Wrong column count** (`G` is 1×3 for `n=2`): now
  `ValueError: solve_qp: G has 3 columns but n=2 (from c)`.
- **NaN/Inf in inputs** (`c=[nan, 1]`): was `status="iteration_limit"`,
[...]

PR #934, opened 2026-09-08
qp(frontend): check `P` for NaN/Inf before the PSD guard reads it

### LINKED ISSUE #932: [adversary] solve_qp: PSD pre-check masks the non-finite P validation, raw LinAlgError instead of the written ValueError (#862 sibling)
adversary-fingerprint: solve-qp/psd-precheck-masks-nonfinite-p-validation

`solve_qp`'s PSD pre-check runs **before** `_validate`, so a `P` containing
`NaN`/`Inf` reaches `np.linalg.eigvalsh` and raises a raw
`numpy.linalg.LinAlgError: Eigenvalues did not converge` from numpy internals.
The precise `ValueError` that `_validate` already contains for exactly this
case — ``solve_qp: `P` contains NaN or Inf`` — is never reached on the default
path.

The same call with `check_psd=False`, which skips the pre-check and lets
`_validate` run, produces the correct, actionable error. So the two spellings
of one malformed input disagree, and the default is the worse one.

This is the **same root cause as #862** (PSD pre-check ordered ahead of
`_validate`), on a different trigger. #862 was the *shape* case; this is the
*non-finite* case, and it is still open on 0.11.0. Every sibling argument
(`c`, `A`, `b`, `h`, `lb`, `ub`) is validated for non-finite entries on the
default path and produces the clean message; only `P` does not.

Severity is low: it rejects rather than returning a wrong answer. It is filed
because the rejection is opaque, inconsistent with every sibling non-finite
check in the same function, and reachable from five public entry points.

There is a second, size-dependent symptom from the same ordering. At `n = 2`
`eigvalsh` converges and returns `nan`, so the pre-check instead raises a
*misleading* `ValueError` blaming nonconvexity:

```
ValueError: P is not positive semidefinite (min eigenvalue nan); the convex QP
interior-point engine (method='ipm') requires a PSD Hessian ... To solve an
indefinite QP, pass method='active-set' ...
```

`P` here is not indefinite, it is non-finite, and the suggested remedy
(`method='active-set'`) does not help. So the same malformed input produces
three different diagnoses depending only on `n` and `check_psd`.

### Reproduction

```python
import numpy as np, pounce
n = 5
P = np.diag([1.0, np.inf, 1.0, 1.0, 1.0])   # non-finite entry; any NaN/Inf works
c = np.ones(n)

pounce.solve_qp(P=P, c=c)                   # LinAlgError (opaque)
pounce.solve_qp(P=P, c=c, check_psd=False)  # ValueError  (correct)
```

Affected entry points (all with default `check_psd`):

| call | result |
|---|---|
| `solve_qp(P, c)` | `LinAlgError` |
| `solve_qp(P, c, method="active-set")` | `LinAlgError` |
| `solve_qp_batch([dict(P=P, c=c)])` | `LinAlgError` |
| `solve_qp_multi_rhs(P=P, cs=...)` | `LinAlgError` |
| `solve_socp(P=P, c=c, ...)` | `LinAlgError` |
| `solve_qp(P, c, check_psd=False)` | `ValueError` (correct) |

Behaviour by size, `P = np.diag([1.0, inf, 1.0, ...])`, default `check_psd`:

| `n` | result |
|---|---|
| 2 | `ValueError: P is not positive semidefinite (min eigenvalue nan)` (misleading) |
| 3 | `LinAlgError: Eigenvalues did not converge` |
| 4 | `LinAlgError: Eigenvalues did not converge` |
| 5 | `LinAlgError: Eigenvalues did not converge` |

`NaN` on the diagonal and `Inf` off the diagonal behave identically to `Inf` on
the diagonal.

### Expected

The message `_validate` already produces, and does produce with
`check_psd=False`:

```
ValueError: solve_qp: `P` contains NaN or Inf
```

Oracle: the function's own `_validate` — the same code path, reached by passing
`check_psd=False` — plus every sibling argument's default-path behaviour:

```
solve_qp(P=np.eye(5), c=[1, inf, 1, 1, 1])  ->  ValueError: solve_qp: `c` contains NaN or Inf
```

Likely fix, matching #862's: validate (or at least run the finite check) before
the PSD pre-check. `_min_eig_lower_coo` (`python/pounce/qp.py`, the
`np.linalg.eigvalsh(M)` at the end) is where the raw exception escapes.

### Actual

```
default      -> LinAlgError : Eigenvalues did not converge
check_psd=F  -> ValueError : solve_qp: `P` contains NaN or Inf
sibling `c`  -> ValueError : solve_qp: `c` contains NaN or Inf
```

### Environment

```
pounce --version                  : pounce 0.11.0
pip show pounce-solver | Version  : 0.11.0
git rev-parse --short HEAD        : a8ba895b
date                              : 2026-09-08
platform                          : darwin 25.6.0, Python 3.14, numpy 2.5.3
```

jkitchin:
Fixes #932

## Problem

A `P` carrying `NaN`/`Inf` was rejected by a different exception depending on `n` and on `check_psd`, and the default path got the worst one. `_validate` already contains the precise message and produces it for every sibling argument (`c`, `A`, `b`, `h`, `lb`, `ub`); only `P` did not.

`P = np.diag([1.0, inf, 1.0, …])`, `c = np.ones(n)`:

| call | before | after |
|---|---|---|
| `solve_qp(P, c)`, `n ≥ 3` | `LinAlgError: Eigenvalues did not converge` | ``ValueError: solve_qp: `P` contains NaN or Inf`` |
| `solve_qp(P, c)`, `n = 2` | `ValueError: P is not positive semidefinite (min eigenvalue nan) … pass method='active-set'` | same as above |
| `solve_qp(P, c, method="active-set")` | `LinAlgError` | same as above |
| `solve_qp_batch`, `solve_qp_multi_rhs`, `solve_socp`, `QpFactorization`, `QpSensitivity` | `LinAlgError` | same as above |
| `solve_qp(P, c, check_psd=False)` | ``ValueError: solve_qp: `P` contains NaN or Inf`` | unchanged |

Oracle is the function's own `_validate`, reached by `check_psd=False` — the same code, on the same input, one branch away. The `n = 2` row is the worst of the three: `P` there is not indefinite, it is non-finite, and `method='active-set'` does not help.

**On the differentiable layers it was not a message problem at all.** `pounce.jax` and `pounce.torch` run their own `_guard_psd` and never reach `_validate`, and a `NaN` in `P`'s **upper** triangle is dropped by `_to_coo_lower` — so the guard's COO is finite and nothing downstream can notice. `P = 2·I₄` with `P[0, 3] = nan`:

```
pounce.qp.solve_qp     -> ValueError: solve_qp: `P` contains NaN or Inf
pounce.torch.solve_qp  -> NO RAISE, x = [-0.5 -0.5 -0.5 -0.5]
pounce.jax.solve_qp    -> NO RAISE, x = [-0.5 -0.5 -0.5 -0.5]
```

That vector is the optimum of the matrix **without** the `NaN` — a silently substituted model, with `_kkt_backward`'s gradients taken through it. gh #874's shape exactly, on gh #932's trigger.

## Cause

Same ordering defect as #862, on a different trigger. The PSD pre-check runs before `_build`, and `_build` is what calls `_validate` — so the gh #113 finite guard sat behind the guard that needed it. `_min_eig_lower_coo` hands the matrix to `np.linalg.eigvalsh`, which does not raise the frontend's message on a non-finite matrix: it fails to converge (`n ≥ 3`), or converges to `nan` and the `nan >= tol` comparison then routes into `_indefinite_error` (`n = 2`). #862 was the shape case; this is the non-finite one.

Reproducing it turned up a symptom the issue does not name, and a worse one: `_psd_verdict_coo` did not only raise the wrong exception, it **returned** on a non-finite `P`, differently depending on where the bad entry sat.

| triplet | verdict on the parent commit |
|---|---|
| `diag(1, 1, 1, nan)` | `(True, 0.0)` — the guard affirming the PSD precondition about a matrix that is not finite |
| `diag(nan, 1, 1, 1)` | `(False, 0.0)` — an indefinite verdict whose own `lam_min` contradicts it |
| `diag(1, inf, 1, 1)` | `LinAlgError` |

`max(abs(v) for v in pv)` is order-dependent under `NaN` (`nan > x` and `x > nan` are both False), so the tolerance came out `nan` in one order and `1.0` in the other. On the `qp.py` entry points that pass is harmless — `_validate` rejects the model a moment later — but it is the guard answering a question it cannot answer.

## Fix

Three placements, because the three frontends have different things behind them.

**`qp.py`.** `_validate`'s finite branch is now module-level `_reject_nonfinite`, called from `_psd_verdict` before it reads `P` as well as from `_validate`, in **`_validate`'s own order — finite, then shape**. That ordering is the reason the check is not simply dropped at the bottom of the guard: an input that is both mis-shaped and non-finite has to get the same message from both spellings of `check_psd`, and only `_validate`'s order gives that. One site covers the seven entry points that check the Hessian before they build (`solve_qp` under both `method=`, `solve_qp_batch`, `solve_qp_multi_rhs`, `solve_socp`, `QpFactorization`, `QpSensitivity`).

**`jax` and `torch`.** The same call in `_guard_psd`, and there it runs **unconditionally, above the `check_psd is False` early return** — where gh #874 put the shape check, for gh #874's reason: `check_psd` says whether the caller wants the *definiteness* precondition verified, and is not permission to solve a different model than the one passed. Below the return, the upper-triangle `NaN` above still returned `[-0.5, -0.5, -0.5, -0.5]` with no exception under `check_psd=False` — the spelling a training loop uses to skip the per-forward eigenvalue cost, so the branch where a wrong gradient is most likely to be consumed. `P` is concrete at all six `_guard_psd` call sites (every one a host forward), so there is no tracer to materialize; the check is deliberately **not** folded into `_validate_p_shape`, which both frontends *also* call at trace time on a tracer, where a shape is known and a value is not.

**`_psd_verdict_coo`** rejects a non-finite triplet as a backstop, so no caller of the *verdict* can be told "PSD" about a matrix whose entries are not numbers.

Rejected: putting the check only in `_psd_verdict_coo`. It is one line and covers `_check_psd` for free — but it fires *after* the shape check, so a doubly-malformed `P` would report shape on the default path and non-finite under `check_psd=False`, the same class of disagreement this PR exists to remove; and reading `pv`, it cannot see the upper-triangle case at all. Measured: with the frontend check removed and the backstop kept, exactly the 4 ordering cases and the 8 AD cases are what fail.

## Blast radius

No solve changes. This only decides which exception a rejected input gets — every input that solved before still solves, and the definiteness verdict on a finite `P` is untouched (`test_the_backstop_leaves_the_finite_verdicts_alone` pins both verdicts and the empty-`P` LP case; `test_a_well_formed_model_still_solves` pins the AD layers under both spellings of `check_psd`, and a `jax.grad` row pins the backward). No Rust changed, so no trajectory question and no fixture sweep: the diff cannot reach the solver.

Cost is one `isfinite` pass over `P` — dense at the frontends, `nnz` at the verdict — against a guard that already densifies and takes eigenvalues on the same matrix. On the AD layers it is now paid on every forward, `check_psd=False` included: `O(n²)` against the `O(n³)`-per-iteration solve on the next line.

`check_psd` semantics are unchanged on `qp.py`: it still governs the definiteness check only, and under `check_psd=False` the finite rejection comes from `_validate` as it always did. On the AD layers it no longer skips the finite check — that is the deliberate behaviour change above.

## Tests

**`python/tests/test_issue932_psd_precheck_masks_nonfinite_p.py`**, 161 cases: all seven entry points × `n ∈ {2, 3, 4, 5}` × four placements of the bad entry, plus two-spellings-agree, a sparse `P`, the siblings, and positive controls.

- **`n` is parametrized because the size split is the defect.** One input had two diagnoses chosen by `n`; pinning a single size records the other size's symptom as absent.
- **One placement is a control, not a case.** `nan-upper` sits only in the upper triangle. Its *message* is unchanged by the fix; its *mechanism* is not — before, it reached `_validate`; now it raises from `_psd_verdict` → `_reject_nonfinite`, which reads the dense `P`. It is the only row in the file that passes on the parent for a different reason than it passes here, and the comment says so.
- **`LinAlgError` is a `ValueError` subclass**, so `pytest.raises(ValueError)` alone does not reject the regression. The type and the absence of "positive semidefinite" are asserted explicitly, so a failure says which of the three symptoms came back.

**`python/tests/test_issue932_ad_frontends_nonfinite_p.py`**, 19 cases: both differentiable layers against `pounce.qp` as a control, over both triangles, both spellings of `check_psd`, all three host forwards, plus a `jax.grad` row. Per-test `importorskip` rather than module-level, so the jax rows still run in a job without torch.

Mutation tables, measured:

| removed | fails |
|---|---|
| `_reject_nonfinite` in `_psd_verdict` | 4 of 161 — only the doubly-malformed ordering case; the backstop catches the rest, with the same message |
| `_reject_nonfinite` in `_psd_verdict_coo` | 3 of 161 — the verdict-level cases |
| both (the parent's `qp.py` behaviour) | 92 of 161 |
| `_reject_nonfinite` in `pounce/jax/_qp.py` | 4 of 19, all jax |
| `_reject_nonfinite` in `pounce/torch/_qp.py` | 4 of 19, all torch |
| either, moved back below `check_psd is False` | 4 of 19 — the `check_psd=False` rows |

The 69 survivors of the full `qp.py` revert are the upper-triangle placement, the siblings, the ±inf-bounds control and the positive controls. In the AD file the `qp` rows never move, and the **lower**-triangle rows survive deleting the frontend check outright — the backstop covers those from underneath. So each guard's row measures exactly what that guard owns.

**Not evidence about** (stated in the test module docstring rather than left to look covered): a non-finite *sibling* against a mis-shaped `P` still reports the shape by default and the sibling under `check_psd=False`. That is identical on the parent and is not closed here — the pre-check is handed `P` and `c` and nothing else, so closing it means running the whole of `_validate` above the guard.

**CI.** Both AD-frontend files are now listed in `python-test-torch`, the only job with torch installed. `test_issue874_ad_frontends_p_shape.py` takes `importorskip("torch")` at module scope, so `python-test` skipped it whole — including its jax rows — and the torch job ran four named files that never included it: a regression test for a silently-substituted model on a differentiable layer that was running in no job at all.

**Runs.** All 8 checks green on `8ad801b` ([run 2876](https://github.com/jkitchin/pounce/actions/runs/34266618341)), including `Python tests (torch)`, which is the first job to run either AD-frontend file. Locally, the exact `python-test-torch` command including both new files: 84 passed.

The full Python suite, with jax **and** torch installed, has one failure that is not this PR's: `test_jax.py::test_jax_problem_no_rebuild_on_repeat_solve_pounce_75` fails identically on `70b35d9` (1 failed, 1620 passed) and on `2483848` (1 failed, 1639 passed), passes in isolation on both, and passes with all 106 tests of its own file on both. It is a cross-file ordering effect on the NLP `Problem` rebuild path, which this diff does not touch, and it is invisible to CI as configured — `python-test` has no torch.

---

- [ ] Book page under `docs/src/` — n/a: no documented behaviour changes, the message is the one `docs/src` already implies for a malformed argument
- [ ] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` — n/a: no Rust in this diff, but CI's `Test (ubuntu-latest)` job (which runs all three) is green on the head

https://claude.ai/code/session_01K9gJ6FPtBPmLxKY3MWYZwB


jkitchin (commented):
## Review — verified by re-running, not by reading

I reproduced every measurement in the body against a worktree at `70b35d99` with a matching `_pounce.abi3.so` (the repo's staleness guard passed there, so the harness loaded the binary that goes with this code — `/sens-review` entry 8).

**Everything the body claims, holds.**

| claim | reproduced |
|---|---|
| new file, 161 cases | 161 passed |
| revert both guards → 92 of 161 fail | **92 failed, 69 passed** |
| drop the frontend check only → 4 fail | **4 failed, 157 passed** |
| drop the backstop only → 3 fail | **3 failed, 158 passed** |
| ordering: finite-then-shape matches `_validate` | confirmed |
| `_guard_psd` is host-side, so the tracer concern is real and correctly avoided | all 6 call sites are `_forward_solve` / `_forward_solve_batch` / `_forward_solve_socp` |

No collateral damage found probing `P` as `None`, list-of-lists, `int64`, `scipy.csr`, `0×0`, or ragged — all unchanged from the parent. `test_dual_sign_regression.py` fails identically on the parent (4 failed, 3 passed); it is a pre-existing pyomo-dual failure, not yours.

One note on the torch call site: `_reject_nonfinite` sits immediately above a line that already did `np.asarray(P)`, so it adds no new conversion exposure for a grad-requiring or non-CPU tensor. Good placement.

---

## 1. The two frontend edits are untested, and they are the only guard on a silent-wrong-answer case — please add coverage

This is the one thing I'd hold the PR for.

Delete `_reject_nonfinite("P", P)` from **both** `python/pounce/jax/_qp.py` and `python/pounce/torch/_qp.py` and:

- `test_issue932_psd_precheck_masks_nonfinite_p.py` — **161 passed**
- every jax/torch/parity test in `python/tests` (`-k "jax or torch or parity"`) — **246 passed**

Nothing goes red. Two of the five changed files have no test.

They are load-bearing, and not in the way the mutation table measures. The `_psd_verdict_coo` backstop reads `pv` — the **lower triangle** — so it cannot see an entry the filter drops, and jax/torch never call `_validate`. With those two lines removed, `P = 2·I₄` with `P[0, 3] = nan`:

```
pounce.qp.solve_qp     -> ValueError: solve_qp: `P` contains NaN or Inf
pounce.torch.solve_qp  -> NO RAISE, x = [-0.5 -0.5 -0.5 -0.5]
pounce.jax.solve_qp    -> NO RAISE, x = [-0.5 -0.5 -0.5 -0.5]
```

Same three lines on the parent commit. That is not an opaque rejection — it is a **silently substituted model** on a differentiable layer, whose `_kkt_backward` gradients then feed a training loop. It is gh #874's shape exactly, and gh #874 got its own file, `test_issue874_ad_frontends_p_shape.py`, for this reason.

The body already knows this — *"the frontend check is the stronger of the two (it reads the dense `P`, so it sees an entry the lower-triangle filter drops)"* — it just never turns the claim into a test. Per CLAUDE.md's branch rule, a leg is only evidence about the branch its fixture reaches, and no fixture here reaches either frontend.

The right case is already in `PLACEMENTS`: `nan-upper` is precisely the placement that separates the frontend check from the backstop. It just needs to run against `pounce.jax` and `pounce.torch` as well as `pounce.qp`, with a mutation row for each.

## 2. The "known gap" closes — torch works, measured

`jax` and `torch` are both importable in my environment, so I ran what the sandbox couldn't. All six differentiable entry points × `n ∈ {2, 3, 4}` give the clean message:

- `torch.solve_qp` / `solve_qp_batch` / `solve_socp` → `ValueError: solve_qp: `P` contains NaN or Inf`
- `jax.solve_qp` / `solve_qp_batch` / `solve_socp` → the same `ValueError`, wrapped by `JaxRuntimeError` from the callback, as the body predicts

And on the parent, torch shows the two symptoms verbatim:

```
torch n=2 -> ValueError: P is not positive semidefinite (min eigenvalue nan); …
torch n=3 -> LinAlgError: Eigenvalues did not converge
torch n=4 -> LinAlgError: Eigenvalues did not converge
```

So the torch edit is correct. Please drop the "Known gap" paragraph and land these as the tests from item 1 rather than leaving them to a CI job — see item 3.

## 3. No CI has run on this branch

`GET /commits/70b35d99/check-runs` returns **zero check runs**; the combined status is `pending`. The body defers torch verification to *"CI's `python-test-torch` job"*, and that job has not executed. Worth a push to kick the workflows before merge.

## 4. The `CONFLICTING` label is stale — the merge is clean

GitHub reports `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY`, but against the same `main` SHA (`957136ee`):

```
git merge-tree --write-tree origin/claude/... origin/main   -> exit 0, clean tree
git merge --no-commit --no-ff origin/main                   -> "Automatic merge went well"
```

`CHANGELOG.md` is the only overlapping file and the `[Unreleased]` entries auto-resolve. A push should refresh the cached value.

---

## Smaller things

**5. `test_a_P_that_is_both_mis_shaped_and_nonfinite_agrees_with_validate` covers `P` only.** The two spellings still disagree when the non-finite argument is a *sibling* and `P` is the mis-shaped one:

```
P = np.eye(7, 5), c = [1, nan, 1, 1, 1]
  default        -> ValueError: solve_qp: `P` has shape (7, 5) but must be (5, 5)
  check_psd=False -> ValueError: solve_qp: `c` contains NaN or Inf
```

Identical on the parent, so this PR does not introduce it and I would **not** fix it here — the pre-check only ever sees `P`, so closing it means moving the whole of `_validate` above the guard. But it is the same class of disagreement the PR exists to remove, so it belongs in a "not evidence about" line rather than being left to look covered.

**6. One docstring is now slightly off about mechanism.** `PLACEMENTS`' `nan-upper` is described as the case that "reached `_validate` intact and must keep doing so." On the fixed code it no longer reaches `_validate` — traced, it now raises from `_psd_verdict` → `_reject_nonfinite`, because that check reads the dense `P`. The message is unchanged, which is what the test asserts, so the test is right; only the stated reason has drifted. Given the file's own subject is guards that stop describing what they do, worth a one-line correction.

---

**Verdict:** the fix is right, the diagnosis is right, and the `qp.py` half is well covered. Item 1 is the blocker — two changed files with no test, guarding a case that silently returns a wrong answer on a differentiable layer. Items 2 and 3 make that cheap to close: the tests run here, and they are the same tests the PR needs anyway.

jkitchin:
Thanks — item 1 was right, and measuring it moved the fix, not just the tests. Pushed as `2483848`.

## 1. The frontend edits are pinned now — and the branch you named made me move the check

I reproduced your measurement first: with `_reject_nonfinite` deleted from `_guard_psd`, `P = 2·I₄` with `P[0, 3] = nan` gives `NO RAISE, x = [-0.5 -0.5 -0.5 -0.5]` on both layers while `pounce.qp` raises. Your diagnosis of why the existing mutation table can't see it is exact: the backstop reads `pv` (the lower triangle), and `_validate` is not on those layers' path at all.

Probing the neighbouring branch turned up a case that was **still** unguarded with my original placement — the check sat *below* the `check_psd is False` early return:

```
P = 2*I4, P[0,3] = nan, check_psd=False
  pounce.qp   -> ValueError: solve_qp: `P` contains NaN or Inf
  pounce.jax  -> NO RAISE, x = [-0.5 -0.5 -0.5 -0.5]
```

That is gh #874's third branch on gh #932's trigger, and gh #874's argument settles it: `check_psd` says whether the caller wants the *definiteness* precondition verified, and is not permission to solve a different model than the one passed. A `NaN` the lower-triangle filter drops is a different model as surely as a wrong shape is — and `check_psd=False` is the spelling a training loop uses, so it is the branch where a wrong gradient is most likely to be consumed. The check now runs above the early return, next to the shape check. My original comment justified the placement by tracer safety, which was wrong: as you verified, all six `_guard_psd` call sites are host forwards, so `P` is always concrete there.

`python/tests/test_issue932_ad_frontends_nonfinite_p.py`, 19 cases: both layers against `pounce.qp` as a control, both triangles, both spellings of `check_psd`, all three host forwards (`solve_qp`, `solve_qp_batch`, `solve_socp`), plus a `jax.grad` row so the per-forward check can't quietly break the backward. Per-test `importorskip` rather than module-level, for the reason in item 3.

Mutation table, measured:

| removed | fails |
|---|---|
| `_reject_nonfinite` in `pounce/jax/_qp.py` | 4, all jax |
| `_reject_nonfinite` in `pounce/torch/_qp.py` | 4, all torch |
| moved back below the `check_psd is False` return | 4 — the `check_psd=False` rows, both frontends |

The `qp` rows never move under any of the three. What *doesn't* fail is the useful part: `[lower-jax]` and `[lower-torch]` survive deleting the frontend check outright, because a lower-triangle `NaN` is in `pv` and the backstop raises the same message from underneath. So the two guards have no overlap where it matters — the frontend check owns the dropped upper triangle and everything under `check_psd=False`, the backstop owns the rest.

## 2. Known gap closed — I can run torch now

`torch` installs from PyPI proper in this sandbox (it was `download.pytorch.org` that the proxy blocks; I only tried the CPU index first). So the AD file's torch rows are run here, not deferred: **84 passed** for exactly the `python-test-torch` command including the two new files. The "Known gap" paragraph is gone from the body.

## 3. CI — and one thing I found while fixing it

Both AD-frontend files are now listed in `python-test-torch`, and this is worth more than the push: **`test_issue874_ad_frontends_p_shape.py` was running in no job at all.** Its `importorskip("torch")` is module-level, so `python-test` (jax, no torch) skips the whole file including its jax rows, and `python-test-torch` runs four named files that never included it. A regression test for a silently-substituted model on a differentiable layer, green by never executing. It is listed now; my new file uses per-test `importorskip` so its jax rows still run in `python-test`.

## 4. Stale `CONFLICTING`

Agreed, and the push should refresh it. If it doesn't, say so and I'll merge `main` in.

## 5. The sibling-vs-`P` disagreement

Reproduced, identical on the parent, and I agree it is not this PR's to close — the pre-check is handed `P` and `c` and nothing else, so closing it means moving the whole of `_validate` above the guard. It is now a "**What this file is not evidence about**" section in the test module's docstring, with your measurement in it, so `test_a_P_that_is_both_mis_shaped_and_nonfinite_agrees_with_validate` cannot be read as covering more than the `P`-only case. The same section names the AD frontends and points at the new file.

## 6. Docstring drift

Fixed, and traced to confirm your reading: on the fixed code `nan-upper` raises from `_psd_verdict` → `_reject_nonfinite`, not from `_validate`. The comment now says the message is unchanged while the mechanism is not, and that this is the only row in the file that passes on the parent for a different reason than it passes here.

---

**Full suite** on the new head: 1639 passed, 13 skipped, 1 failed — `test_jax.py::test_jax_problem_no_rebuild_on_repeat_solve_pounce_75`, which passes in isolation and passes with all 106 tests of its own file, so it is a cross-file ordering effect and not this diff (it counts rebuilds on the NLP `Problem` path; nothing here touches it). I have the same full run going on the parent commit to confirm it fails there too, and will post the result either way.

---

jkitchin:
Two follow-ups on items 3 and 4, which turn out to be the same thing, plus the verification I promised.

**Items 3 + 4 are one item: no CI had run *because* of the conflict state.** GitHub does not create a `pull_request` workflow run while it cannot compute the merge commit, so `mergeable_state: dirty` and "zero check runs" were cause and effect, not two independent observations. The push in `2483848` refreshed neither — still `dirty`, still no run. `main` had also moved under the PR by then (`957136ee` → `bbea93ac`, #935), while `git merge-tree --write-tree` stayed clean against both, so the cached verdict was wrong rather than merely stale.

Merging `origin/main` into the branch (`8ad801b`) cleared it, and CI started immediately: [run 2876](https://github.com/jkitchin/pounce/actions/runs/34266618341), `event: pull_request`, on `8ad801b`. So the two new files, including the AD ones now listed in `python-test-torch`, get their first real run there. A merge rather than a rebase, per the repo's convention for a branch under review.

**The full-suite failure is pre-existing, measured.** Same command, same environment, on `70b35d9` (this PR's first commit, before any of tonight's changes):

```
70b35d9  1 failed, 1620 passed, 13 skipped   FAILED tests/test_jax.py::test_jax_problem_no_rebuild_on_repeat_solve_pounce_75
2483848  1 failed, 1639 passed, 13 skipped   FAILED tests/test_jax.py::test_jax_problem_no_rebuild_on_repeat_solve_pounce_75
```

Same test, same position, on both trees; it passes in isolation and passes with all 106 tests of `test_jax.py` on either. So it is a cross-file ordering effect on the NLP `Problem` rebuild path, which this diff does not touch. Worth noting it is invisible to CI as configured — `python-test` has no torch, and the 19-test difference between those two runs is this PR's new files — so I have not filed it; say the word if you'd like an issue rather than a line in a PR thread.

I'll watch the run and push fixes if anything goes red.

---

PR #940, opened 2026-09-14
minimize: report fun(x) on the convex routes, not the finite-differenced model's value

### LINKED ISSUE #939: [adversary] minimize: auto/qp-ipm route reports the FD quadratic model's objective as fun, below the true convex optimum (fun != f(x))
adversary-fingerprint: minimize/auto-route-fun-not-objective-at-x

When `pounce.minimize` routes a problem to the convex QP solver (`solver_selection="auto"`, and the same with `"qp-ipm"`), it returns `OptimizeResult.fun` equal to the objective of the **finite-differenced quadratic model** (`res.obj + ex.obj_const`, `_minimize.py::_solve_via_convex`), not `fun(res.x)`. On a convex QP with bounds and a linear equality, the reported `fun` is **below the proven global minimum** by 8.8e-5 (1.1e-5 relative). The true objective at the returned `x` is correct to within 3e-9, and the returned `x` is 2.7e-5 from the optimum. With `solver_selection="nlp"`, `fun == f(x)` exactly and `x` is within 2e-8. scipy's contract, which the facade mirrors, is that `fun` is the objective evaluated at `x`.

## Reproduction

```python
import numpy as np, scipy.sparse as sp, clarabel, pounce, warnings
from scipy.optimize import LinearConstraint
warnings.simplefilter("ignore")
rng = np.random.default_rng(2149425415)
n = int(rng.integers(3, 12))                      # n = 11
M = rng.standard_normal((n, n)); H = M @ M.T + np.diag(10 ** rng.uniform(-2, 2, n))
g = rng.standard_normal(n) * 5
lb = np.empty(n); ub = np.empty(n)
for i in range(n): lb[i], ub[i] = -rng.uniform(0.1, 1), rng.uniform(0.1, 1)
a = rng.standard_normal(n); bb = rng.uniform(-0.2, 0.2)
f = lambda x: 0.5 * x @ H @ x + g @ x            # convex QP objective (cond(H) ~ 33)
x0 = np.clip(a * bb / (a @ a), lb, ub)
# independent oracle: clarabel
st = clarabel.DefaultSettings(); st.verbose = False
Gc = np.vstack([a[None, :], np.eye(n), -np.eye(n)]); hc = np.concatenate([[bb], ub, -lb])
cs = clarabel.DefaultSolver(sp.csc_matrix(np.triu(H)), g, sp.csc_matrix(Gc), hc,
                            [clarabel.ZeroConeT(1), clarabel.NonnegativeConeT(2 * n)], st).solve()
print("clarabel optimum      :", cs.obj_val)
for sel in ["nlp", "auto"]:
    r = pounce.minimize(f, x0, bounds=list(zip(lb, ub)), constraints=[LinearConstraint(a[None, :], bb, bb)], solver_selection=sel)
    print(f"{sel:4s} success={r.success} r.fun={r.fun!r}  f(r.x)={f(r.x)!r}  |r.fun-f(r.x)|={abs(r.fun-f(r.x)):.2e}  |x-x_clarabel|={np.linalg.norm(r.x-np.array(cs.x)):.1e}  eq_viol={abs(a@r.x-bb):.1e}")
```

## Expected

`r.fun == f(r.x)` (to rounding), and no smaller than the global optimum of this convex QP, **−7.790483040** (clarabel, `Solved`; `solver_selection="nlp"` agrees to 6e-8).

## Actual

```
clarabel optimum      : -7.790483040402875
nlp  success=True r.fun=-7.790483104614411  f(r.x)=np.float64(-7.790483104614411)  |r.fun-f(r.x)|=0.00e+00  |x-x_clarabel|=1.9e-08  eq_viol=8.3e-17
auto success=True r.fun=-7.790571450608003  f(r.x)=np.float64(-7.79048303866164)  |r.fun-f(r.x)|=8.84e-05  |x-x_clarabel|=2.7e-05  eq_viol=6.9e-15
```

`solver_selection="qp-ipm"` returns the identical `fun=-7.790571450608003`. The reported `fun` sits 8.8e-5 below a global minimum, which is impossible, and `success=True` with message "Optimization terminated successfully." gives no hint that it is a model value.

On unconstrained random QPs (n = 2..11) the gap is smaller, around 1e-9 to 2e-8 relative. On a 2-variable QP scaled by 1e4 it is 8.8e-6 absolute. So the error grows with the conditioning of the extracted model and with how much constraint structure there is. Note that the extracted model's error also moves `x` itself (2.7e-5 vs 1.9e-8 on the NLP route); that is the underlying cost of the FD probe. The `fun` inconsistency is the part that can be fixed cheaply: evaluate the user's `fun` at `res.x`, which is one extra call.

## Environment

- `pounce --version`: pounce 0.11.0
- `pip show pounce-solver`: Version: 0.11.0
- repo HEAD: 598aedf1
- date: 2026-09-14
- aarch64-apple-darwin, Python 3.14.2

jkitchin:
Fixes #939

## Problem

On the convex routes, `pounce.minimize` returned an `OptimizeResult.fun` that
was not the objective at `x` — it was the value of the *finite-differenced
quadratic model* the router extracted. On an 11-variable convex QP with a box
and one linear equality, that put the reported objective **below the problem's
proven global optimum**, carrying `success=True` and "Optimization terminated
successfully." with no hint it was a model number.

The issue's model, driven with an exact `jac` so it routes on x86-64 linux too
(see **Tests**), measured on `188ea41` (parent) vs `192eb39`:

| | `res.fun` | `f(res.x)` | `\|fun − f(x)\|` |
|---|---|---|---|
| before | −7.790483041784377 | −7.790483041771242 | 1.3e-11 |
| after | −7.790483041771242 | −7.790483041771242 | 0 |
| NLP route (oracle) | −7.790483104614411 | −7.790483104614411 | 0 |

That 1.3e-11 is this platform's FD noise, not the defect's size. The issue
measured **8.8e-5 (1.1e-5 relative), below a clarabel-proven optimum**, on the
same model with a fully finite-differenced gradient. The magnitude is pinned
independently by a case where the model error is exact arithmetic rather than
round-off (`test_routed_fun_is_the_objective_at_x_when_the_model_is_visibly_off`),
which lands at the same order and the same sign:

| | value |
|---|---|
| `res.fun` before (= the model's value) | −681818.1818212409 |
| `f(res.x)` | −681818.1817430503 |
| gap | −7.8e-5, i.e. the reported objective was **too low** |

## Cause

`solver_selection="auto"` (and the explicit `"lp-ipm"` / `"qp-ipm"` /
`"qp-active-set"` / `"socp"`) does not hand the solver the user's `fun`.
`pounce._route` probes the opaque callable, **finite-differences** it into
`½xᵀPx + cᵀx + obj_const`, validates that model at held-out points to
`route_tol`, and `solve_qp` / `solve_socp` optimize *that*. The value they
report back is therefore the model's value at `x`.

`_solve_via_convex` / `_solve_via_socp` computed `fun_val = float(res.obj) +
ex.obj_const` and handed it straight back as `OptimizeResult.fun`. The held-out
validation rules out a *misclassification*; it does not make the model **equal**
to the objective, and a model that agrees to `route_tol` at the probe points is
still a different function at the optimum.

## Fix

`_objective_at_solution` evaluates the user's objective at the returned point —
via `route_fun`, the same `args`-bound callable the routers probed (#196's
binding), so a parameterized objective is evaluated correctly rather than
raising `TypeError` into the fallback.

* `res.fun == fun(res.x)`, to the last bit, on every convex route.
* `res.nfev` is **1** on a routed solve rather than 0. Two existing tests used
  `nfev == 0` as the tell that the convex driver ran rather than a callback
  path; they now say 1, which still separates it sharply from the NLP and SQP
  paths. `njev`/`nhev` stay 0 — no gradient or Hessian callback fires.
* The model's own value is kept as `res.info["obj_model"]`, so
  `obj_model − fun` *measures* the extraction error instead of being reported
  *as* the answer.
* If the evaluation raises or returns a non-finite value, the model value
  stands **with a warning**. This call happens after the solve, at a point the
  user never chose — an interior-point solver can land a hair outside a bound
  it treated as an inequality, where a restricted-domain `fun` blows up. Losing
  an otherwise good result to that would be a worse bug than the one being
  fixed; reporting a value that is silently not `fun(x)` would be the same bug
  again, hence the warning.

The SOCP arm reads its `P`/`c`/`obj_const` out of the *same* `_fit_objective`
probe, so it carried the identical defect and takes the identical fix.

**Rejected: clamping, or warning when `|obj_model − fun|` is large.** A
threshold there would need a measured population behind it and has none; the
two numbers are both exposed instead, which is threshold-free and lets the
caller decide.

**Not fixed, deliberately.** The same model error also moves `x` itself —
2.7e-5 from the optimum on the issue's QP against the NLP route's 1.9e-8. That
is inherent to probing an opaque callable and no post-solve evaluation can
repair it. An analytic `jac` removes one layer of the differencing and
`solver_selection="nlp"` all of it; `docs/src/python.md` ("What the routed
solver actually optimized") and the test file's *not evidence about* list both
say so, so the remaining gap is a known one rather than an implied fix.

## Blast radius

Python facade only — **no Rust changed**, and nothing upstream of the solve
changed. Every line of the diff in `_solve_via_convex` / `_solve_via_socp` runs
*after* `solve_qp` / `solve_socp` returns; both are still called with identical
arguments, so the extracted model, the iterates, and the returned `x` are
bit-identical. Not a trajectory change: it does not reroute which correction
the solver reaches for, reorder or rescale any step, or move a model between
engines, so `scripts/sweep-fixtures.sh` has nothing to report on it. The CLI is
unaffected — it classifies from the `.nl` expression tree, so its convex model
is symbolic and exact; this defect is specific to the callable facade.

What a caller can observe changing, on routed solves only:

* `res.fun` (and `res.info["obj_val"]`, which mirrors it) is now `fun(res.x)`.
* `res.nfev` 0 → 1, and one extra `fun` evaluation is made per routed solve —
  negligible against the router's `O(n²)` probe, but it is a real call, so an
  objective with side effects sees it.
* `res.info` gains `obj_model`. Nothing is removed.

Full Python suite: 1267 passed, 52 skipped. One pre-existing unrelated failure
(`test_continuation.py::test_step_controller_is_what_the_jax_follower_uses`
needs the optional `jax`, not installed in this container); it fails identically
on the parent commit. `make ask-check` passes (top-1 17/20 against a floor of
14, top-5 20/20 against 18) after the `docs/src/python.md` edit.

## Tests

`python/tests/test_issue_939_routed_fun_at_x.py`, 10 tests, carrying its own
mutation table. Measured on the **parent commit** (`188ea41`), not argued from
reading: **8 of the 10 fail**, each on `assert res.fun == float(fun(res.x))`
reading the model's value — not on an import, a signature, or an incidental
error. The two that pass are named in the file and are not the regression pin:
the LP arm (a linear objective's model is recovered exactly, so the two values
genuinely agree — it is there to say the arm was checked) and the import guard.

Finer mutations, to show neither arm's tests stand in for the other's:

| mutation | red |
|---|---|
| `_objective_at_solution` returns `(model_val, 0)` unconditionally | 7 of 10 |
| drop `objective=route_fun` at the `_solve_via_convex` call site only | 3, all QP; SOCP green |
| drop it at the two `_solve_via_socp` call sites only | 1, the SOCP arm; QP green |

**Deliberately not pinned, and why.** The issue's repro passes no `jac`, which
leaves both the gradient and the Hessian to finite differences — and whether
that double-FD model clears the router's constant-Hessian check is decided by
the platform's floating-point noise. Measured: it routes on
aarch64-apple-darwin (the reporter's), and misses the check by 0.0035 against a
0.00255 tolerance on x86-64 linux, where `auto` then falls back to NLP and
`qp-ipm` raises `ValueError`. Pinning that would pin one platform's FD noise,
not the defect, so the test drives the issue's own model with an exact `jac`:
the *objective* model stays finite-differenced, which is the part this bug is
about, and it routes everywhere. The defect's magnitude is covered separately
by the exact-arithmetic case in the table above.

---

- [x] Tests fail on the parent commit for the stated reason — 8 of 10, each on
      `res.fun` reading the model's value; verified by checking out
      `188ea41`'s `_minimize.py` under the new tests
- [x] Book page under `docs/src/` updated — `docs/src/python.md`, new "What the
      routed solver actually optimized" subsection (existing page, already in
      `SUMMARY.md`)
- [ ] `cargo fmt --all -- --check`, `cargo clippy`, `cargo test` clean — **N/A,
      no Rust changed**; the Python suite result is above
      as it stands now

https://claude.ai/code/session_01FHrr418EMEbxixCoMogjfD

---
_Generated by [Claude Code](https://claude.ai/code/session_01FHrr418EMEbxixCoMogjfD)_


jkitchin:
**`Python tests (torch)` is red on two external-fetch failures, neither of them this PR's. Nothing to port — there is no fix for a third-party registry timing out.**

Both attempts died in `PyO3/maturin-action`'s **setup**, before a single test body ran:

| attempt | at | died in | error |
|---|---|---|---|
| 1 ([job](https://github.com/jkitchin/pounce/actions/runs/34857050865/job/104019201127)) | 14:40Z | Install maturin | `curl` fetched **92 bytes** instead of the release tarball → `gzip: stdin: not in gzip format`, `tar: Error is not recoverable` |
| 2 ([job](https://github.com/jkitchin/pounce/actions/runs/34857050865/job/104101828030)) | 18:34Z | Pull Docker image | `docker pull quay.io/pypa/manylinux2014_x86_64:latest` → `Error response from daemon: received unexpected HTTP status: 504 Gateway Time-out` |

Why this is not the PR's:

* The diff is 6 files — `python/pounce/_minimize.py`, three test files, `CHANGELOG.md`, `docs/src/python.md`. No Rust, no packaging, no workflow, no Dockerfile, nothing that can influence whether `quay.io` answers a pull.
* The other **7 checks on this same commit passed**, including two that build wheels through the same machinery (`wheel smoke (pounce-solver)`, `wheel smoke (pyomo-pounce)`) and the sibling `Python tests (jax)`.
* It is transient rather than a standing outage: run [#2884](https://github.com/jkitchin/pounce/actions/runs/34873742096) on `ws-ps-session` went green at 17:30Z, i.e. *between* these two failures.

Attempt 2 was a re-run of attempt 1, which is the one re-run an install-stage death earns. I am re-running once more against this comment and then stopping: if it fails a third time I will treat it as real and root-cause it rather than keep spending re-runs. The PR stays watched until it is green and merged.

For the record, the change itself is green everywhere it is actually exercised — full Python suite locally 1267 passed / 52 skipped, and `Test (ubuntu-latest)`, `Python tests (jax)`, both wheel smokes, the CasADi jobs and the WebAssembly build all passed on `192eb39`.

---

jkitchin:
Result of that third attempt ([job](https://github.com/jkitchin/pounce/actions/runs/34857050865/job/104102551337), 18:36Z): the **identical** `quay.io … 504 Gateway Time-out` on `docker pull`, same step, same 24s timeout as attempt 2 — so it reproduces exactly, which settles it as external rather than this PR's. No further re-runs from me; the job has still never reached a test body.

---

PR #941, opened 2026-09-14
Add presolve-aware warm solve sessions

GermanHeim:
## Problem

In `pounce-rs`, re-solving a problem family currently means choosing presolve or warm starts. Embedders with persistent solves (MPC loops, oximo, parametric sweeps) had to pick one or the other.

> I used Codex for the initial implementation and then reviewed myself with some edits

## Cause

There are no presolve-aware warm solve sessions.

## Fix

This PR adds sessions that keep the presolve transform across solves and map every original-space seed through it: `TnlpPresolveSession` for NLPs (fingerprint + reuse-or-rebuild, mu_init threading, projection report) and `ConvexPresolveSession` for convex QPs (retained `Presolve`, `QpWarmStart` projection, postsolve back).

## Blast radius

Additive in `pounce-rs`.

## Tests

      since moved


jkitchin (commented):
Fetched the branch and read the full diff against the merge base (`188ea41`, 11 files / +2450), then ran the suite. `cargo test -p pounce-rs -p pounce-presolve -p pounce-convex` is green (280 + ~74 tests) and `cargo fmt --all -- --check` is clean, so those two checkboxes hold. Three others don't.

The architecture is sound and the projection layer is the right idea — the writeup in `docs/src/sessions.md` is genuinely good. Findings below in severity order.

---

## 1. Blocker — `mu_init` threading never happens after the first solve

`session.rs` copies the `batch.rs` idiom for "did the caller set `mu_init`?":

```rust
let user_set_mu = matches!(self.app.options().get_numeric_value("mu_init", ""), Ok((_, true)));
if !user_set_mu { /* write mu_init */ }
```

But `OptionsList::get_numeric_value` returns `(v, true)` whenever the tag is **present in the list** (`crates/pounce-common/src/options_list.rs:293-304`) — it means "explicitly set", not "set by the user". In `batch.rs` that idiom is sound because `apply_warm_options` runs against a *fresh per-worker app*. The session reuses **one long-lived app**, so its own first write poisons the check forever after.

Measured, with a probe on this PR's own `Mini` fixture:

```
after cold:  mu_init=(0.1, true)  final_mu=1e-11
after warm1: mu_init=(0.1, true)  final_mu=3.72e-9   iters=3
after warm2: mu_init=(0.1, true)  final_mu=3.72e-9   iters=3
expected threaded mu ~= 1e-9
```

`mu_init` is pinned at `WARM_MU_CEILING` for the life of the session. So `WarmPoint::mu`, the `final_mu` threading, and `WARM_MU_FLOOR` are all dead code past solve #1 — while the PR body advertises "mu_init threading" and `docs/src/sessions.md` ships the claim "`SessionSolution::warm_point()` threads `final_mu` into the next `mu_init`."

Tracking the session's own write fixes it. I verified this threads correctly and keeps all 7 session tests green:

```rust
// new field: own_mu_init: Option<Number>
let user_set_mu = match self.app.options().get_numeric_value("mu_init", "") {
    Ok((v, true)) => self.own_mu_init.map(|o| o != v).unwrap_or(true),
    _ => false,
};
// ...and set self.own_mu_init = Some(mu) alongside each write.
```

```
after warm1: mu_init=(1e-9, true)
after warm2: mu_init=(3.72e-9, true)
```

That all 7 tests pass both ways is itself the finding in section 5.

## 2. "Blast radius: additive in `pounce-rs`" is not accurate — this touches the shared presolve path

Commit `ed5a24a` adds a clamp + aux-fixed override inside `TNLP for PresolveTnlp::get_starting_point` (`crates/pounce-presolve/src/lib.rs`, around line 1757). That runs for **every** `presolve=yes` caller, not just sessions. This PR's own new test pins the change: `served_warm_primal_matches_reported_projection` asserts the served seed is now `[3.0, 3.0]` where `main` serves the inner `[5.0, 5.0]`.

Per CLAUDE.md, moving the seed the solver starts from is a trajectory change and needs `scripts/sweep-fixtures.sh` against a baseline binary, **both legs**, before merge. Nothing in the PR shows one.

My own reading says Ipopt's initial-point push into the bound interior probably absorbs a pure clamp (`x0=5` and `x0=3` both project to `x_u - p_u`), and the aux-fixed override looks redundant with the clamp because `lib.rs:1055-1057` already sets `x_l[i] = x_u[i] = v`. But that is exactly the "looked fine on inspection" reasoning CLAUDE.md warns about — gh#544 read fine on inspection too. Please run the sweep rather than take my analysis. If it comes back empty, that is a cheap and durable answer.

Worth asking separately: if the clamp *is* a no-op downstream, does it need to be in the shared wrapper at all, or only in the session's report path?

## 3. The CHANGELOG box is checked but `CHANGELOG.md` is untouched

`git diff --name-only 188ea41..15f2b91` has no `CHANGELOG.md`. The body checks "`CHANGELOG.md` `[Unreleased]` entry, in the user's terms" and then checks the box asserting every claim in the body is true. Two of the three doc/claim boxes don't hold (this one and the `mu_init` sentence in section 1).

## 4. `warm_report` is a parallel reimplementation of what the wrappers do

The module doc says the served seed and `warm_report` "agree", but they are produced by two independent code paths: the solver consumes what `PresolveTnlp`/`LinearEqElimTnlp` serve, while the report is recomputed by `project_warm_point_full` from a `PresolveMap` snapshot. Nothing forces them to stay in step — a future edit to either side drifts silently, and the symptom is a report that lies about a seed that is fine (or vice versa). `served_warm_primal_matches_reported_projection` pins exactly one case, at the presolve layer only, with no elimination stacked. Either derive the report from what the wrappers actually served, or add an equivalence test per branch.

## 5. Coverage gaps, in the repo's own terms

- **No test asserts warming helps.** Every session test asserts status + optimum. That is the gh#544 shape precisely — "the suite asserts status and objective, not iteration count, so nothing saw it." It is also why section 1's dead `mu_init` path survived 7 green tests. An iteration-count assertion (warm < cold on a model where it should be) would have caught it.
- **The elimination and aux branches are integration-untested.** `presolve_linear_eq_reduction` and `presolve_auxiliary` both default to `false` and no test sets either. So `ensure_wrapper`'s `LinearEqElimTnlp` arm never executes, and `transformation()`'s `fixed_vars` collection never sees a non-empty stack. Credit where due: `warm.rs`'s unit tests *do* cover both branches of the pure functions (`projection_overrides_fixed_vars`, `full_projection_folds_elim_gather`) — but those build the `PresolveMap`/`EliminationPlan` by hand, so they say nothing about whether the live wrappers populate them correctly. Two-level full-x / var-x remapping is the exact class that shipped gh#450 and re-shipped as gh#672 finding 1.

## 6. Design question: does retention actually engage for the motivating workload?

Both fingerprints are conservative to the point of near-inertness for the advertised use case. `convex_presolve_fingerprint` hashes every number in the `QpProblem` (`c`, `b`, `h`, `lb`, `ub`, and all `P`/`A`/`G` triplets); `compute_fingerprint` hashes all bounds and linear Jacobian values. An MPC step or parametric sweep moves at least the rhs or the cost every solve — so `presolve_reused` / `last_reused_shape()` will be `false` on essentially every iteration of the workload in the Problem statement, and presolve re-runs from scratch. The only test that exercises reuse solves the *identical* problem twice.

That is safe (reusing a stale transform would be wrong), and the warm-point projection is a genuine deliverable on its own. But "keeps the presolve transform across solves" oversells what a real family gets, and it would be worth measuring the actual reuse rate on an oximo-like loop before the docs promise it.

Related inconsistency: the NLP fingerprint deliberately ignores the objective (documented in `sessions.md`), while the convex one hashes `c` — `cold_then_warm_through_session` asserts a cost change *rebuilds*. Opposite policies on the same kind of change, and the docs describe only the NLP one. Probably both are right for their own presolve, but the asymmetry should be stated.

## 7. Minor

- `TnlpPresolveSession::new` forces `warm_start_init_point=yes` permanently, so `solve_cold()` ("the TNLP's own starting point") routes through the warm-start initializer, not the cold one — different bound pushes, different mu selection. I measured it on the `Mini` fixture and got 6 iterations either way, so I have no evidence of harm; flagging it as a naming/doc accuracy point, not a defect.
- `hash_f64` hashes NaNs by payload, so two distinct NaN bounds force a spurious rebuild. Harmless, just noting it is not quite the "change detection" the comment implies.
- `self.last` stores a failed solve's payload, so `solve_warm_last()` will happily seed from a non-converged point. Graceful when `x` is empty (falls back to cold), but a partial or diverged iterate seeds silently.

---

**Summary:** section 1 is a real bug that makes an advertised feature inert and is contradicted by the shipped docs; section 2 needs the fixture sweep the repo requires; section 3 is a quick fix. Sections 4-6 are the ones I would want addressed before this becomes load-bearing for oximo.

---

GermanHeim:
Regarding N° 2:  I ran `sweep-fixtures.sh` with `presolve=yes` against `ed5a24a^` and `ed5a24a`.

The diff shows:
-lbfgs  csfi2  nlp  SolveSucceeded  it=19,
+lbfgs  csfi2  nlp  SolveSucceeded  it=20.

All other fixture-legs were unchanged. The interior push does absorb the changed primal seed, but gradient scaling reads `get_starting_point` before that push.  On csfi2, disabling scaling makes both versions identical.

Since this affects every `presolve=yes` caller, could the clamp be scoped to the session path instead of changing the shared `PresolveTnlp` wrapper?

jkitchin:
Thanks — that sweep is exactly the right piece of work, and the diagnosis is better than mine was. I'd argued the interior push would absorb the clamp; you found the second read I missed. Confirmed in the source: `nlp_scaling_method` defaults to `gradient-based` (`upstream_options.rs:873`), its own registry text says it "samples the Jacobian once at x0", and `scale_gradient_based` evaluates `eval_grad_f` at the starting point before any push. Your disable-scaling control is what turns that from a story into a result.

Two things follow.

## 1. Yes — scope it, and there's a way that keeps your `warm_report` fix

Worth naming first: `6361598` made this harder to undo than it looks. The report now comes from the wrapper actually doing the clamping, which is the right fix for the drift problem. Move the clamp out to the session wholesale and the report has to be recomputed from a `PresolveMap` snapshot again — straight back to the parallel reimplementation that commit removed.

So rather than moving it, gate it. A `project_seed: bool` on `PresolveTnlp`, default `false`, set to `true` only by `TnlpPresolveSession`:

- non-session `presolve=yes` callers see byte-identical behavior, and `csfi2` goes back to 19;
- the session still gets `starting_point_projection_report` from the real serving call, so `6361598` stands unchanged.

The struct already carries the report field and `reset_starting_point_projection_report`, so this should be small. Re-run your sweep after and I'd expect a genuinely empty diff — which also makes the PR body's "additive in `pounce-rs`" true as written.

If you'd rather keep the clamp shared, that's defensible too, but then the `csfi2` move needs an issue with an owner and a line in the CHANGELOG entry carrying the number. Per CLAUDE.md a measured regression recorded only as an accepted cost is "indistinguishable from noise to the next reader" — and a PR comment is thinner than the commit message that rule was written about. `csfi2` in particular is a model this repo has tracked to the digit before (a `19 -> 105` was rejected as a tuning artifact). I'd take the flag; it's less work than the paperwork.

## 2. `own_mu_init` has a collision — and it's my fault, not yours

The comparison is by value, so if a caller sets `mu_init` to exactly what the session last wrote, the session reads it back as its own and overwrites it. `explicit_mu_init_still_wins` uses `0.05`, so an exact match against `WARM_MU_CEILING` or the last threaded mu isn't covered. Narrow, but it does break "explicit setting wins," and I handed you the mechanism without spotting it. A bool flipped in `set_option_num` when `tag == "mu_init"` avoids it (`options_mut()` still bypasses that, which is probably acceptable).

## Otherwise verified

Pulled `703bc67` and ran it: `cargo test -p pounce-rs -p pounce-presolve -p pounce-convex` is green. Sections 1, 3, 4, 5 and 6 all look properly addressed, and the new tests hit the branches I called out — `warm_report_folds_the_live_linear_elimination_layer` and `warm_start_crosses_live_auxiliary_and_linear_elimination_layers` cover the elim and aux paths that nothing reached before, and there's a real iteration-count assertion now.

The docs response to section 6 is better than what I asked for. Stating outright that the reuse rate is 0% after the initial build for an MPC loop, and putting the two objective policies side by side in a table, is the honest version — I'd expected the claim to be softened rather than measured and written down.

Section 2 is the only thing I'd hold merge on.

---

PR #946, opened 2026-09-17
Retry a line search that fails at an already-feasible iterate (gh#945)

### LINKED ISSUE #945: [adversary] nlp: Error_In_Step_Computation on an equality-constrained strictly convex QP that the exact-Hessian arm solves in 1 iteration
adversary-fingerprint: nlp/step-breakdown-equality-constrained-convex-qp

Filed by the nightly adversary job. The reported **status is honest**
(`Error_In_Step_Computation`, `success=False`) — this is a robustness/convergence
defect on the no-exact-Hessian path, not a wrong-answer-under-success defect.

## Summary

`pounce.minimize` / `pounce.Problem` with an objective and gradient but **no exact
Hessian** (the mode the Python frontend and the CasADi plugin select automatically)
fails with `Error_In_Step_Computation` on a **10-variable, strictly convex, separable
quadratic with one linear equality constraint and no bounds** at `cond(H) = 3.2e3`.

The model is about as benign as a constrained NLP gets: diagonal positive-definite
Hessian, a single linear equality, no inequalities, no bounds, a unique optimum with a
closed form. The failure is a **hard breakdown at iteration 131**, not a budget
exhaustion — the iteration count and the status are identical at `max_iter` = 200,
1000 and 5000.

This is not a single unlucky instance: over fresh random instances of the same shape,
**9-17 of 40 fail at `cond(H) = 1e3`, and 35-39 of 40 fail at `cond(H) = 1e5`**
(table below).

Related but distinct: #818 (L-BFGS stall, **unconstrained**, `Maximum_Iterations_Exceeded`)
— its repro now returns `Solve_Succeeded` in 22 iterations, i.e. it is fixed, and the
**unconstrained** version of the objective below also solves fine. The equality
constraint is required to trigger this. #746 is the same failure mode
(`Error_In_Step_Computation` under limited-memory where the exact arm succeeds) on
large `qcqp*` benchmark problems; this is a minimal instance of that shape that still
reproduces at `8cb4789c`.

## Reproduction

```python
import numpy as np, pounce
from scipy.optimize import minimize as smin, BFGS

# min 0.5*sum_i d_i*(x_i - 1)^2  s.t.  sum(x) = 0.   d = logspace(0,3.5,10), cond(H)=3.2e3.
n = 10
d = np.logspace(0, 3.5, n)
t = np.ones(n)
fun = lambda x: 0.5 * float((x - t) @ (d * (x - t)))
jac = lambda x: d * (x - t)
con = [{"type": "eq", "fun": lambda x: x.sum(), "jac": lambda x: np.ones(n)}]

# Closed form (KKT of an equality-constrained QP): x* = t + lam/d, lam = -sum(t)/sum(1/d)
lam = -t.sum() / np.sum(1.0 / d)
xstar = t + lam / d
fstar = 0.5 * float((xstar - t) @ (d * (xstar - t)))
relkkt = lambda x: (lambda g: np.linalg.norm(g - g.mean(), np.inf)
                    / max(1.0, np.linalg.norm(g, np.inf)))(d * (np.asarray(x) - t))

r = pounce.minimize(fun, np.zeros(n), jac=jac, constraints=con)          # default tol=1e-8
o = smin(lambda x: (fun(x), jac(x)), np.zeros(n), jac=True, constraints=con,
         method="trust-constr", hess=BFGS(),                             # matched information
         options=dict(maxiter=3000, gtol=1e-12, xtol=1e-14))

print("truth  :", f"{fstar:.12g}", f"relKKT={relkkt(xstar):.2e}")
print("pounce :", r.message, f"nit={r.nit}", f"f={r.fun:.12g}", f"relKKT={relkkt(r.x):.2e}")
print("scipy  :", o.success, f"nit={o.nit}", f"f={o.fun:.12g}", f"relKKT={relkkt(o.x):.2e}")
```

## Expected

`f* = 29.5826274799` at `x* = t + lam/d`. Four independent oracles agree:

1. **closed form** — the KKT system of an equality-constrained QP is linear, so
   `x* = t + lam/d` with `lam = -sum(t)/sum(1/d)` exactly; `relKKT = 2.3e-14`.
2. **scipy `trust-constr` with `hess=BFGS()`** — *matched information* (objective +
   gradient, quasi-Newton Hessian, general NLP method): `success=True`, 90 iterations,
   `f = 29.5826274799`, `relKKT = 3.3e-09`.
3. **POUNCE's own exact-Hessian arm** — `pounce.Problem` with a `hessian` callback:
   `Solve_Succeeded` in **1 iteration**, `relKKT = 2.1e-14`.
4. **POUNCE's own convex QP engine** — `pounce.solve_qp(P=np.diag(d), c=-d*t,
   A=np.ones((1,n)), b=[0.0])`: `optimal`, `relKKT = 7.4e-13`.

`relKKT` is an independent dual-residual check computed outside POUNCE: for this model
the stationarity condition is `d*(x-t) - y*1 = 0`, so `||g - mean(g)||_inf / max(1,||g||_inf)`
is the relative KKT error at the returned point.

## Actual

```
truth  : 29.5826274799 relKKT=2.30e-14
pounce : Error_In_Step_Computation nit=131 f=29.5826279468 relKKT=3.81e-03
scipy  : True nit=90 f=29.5826274799 relKKT=3.34e-09
```

`status = -3`, `success = False`, relative KKT error `3.8e-03` — five orders above the
default `tol=1e-8` — and `x` wrong by `3.6e-05` relative.

### It is a breakdown, not a budget exhaustion

| `max_iter` | status | iterations used |
|---|---|---|
| 50 | `Maximum_Iterations_Exceeded` | 53 |
| 200 | **`Error_In_Step_Computation`** | 131 |
| 1000 | **`Error_In_Step_Computation`** | 131 |
| 5000 | **`Error_In_Step_Computation`** | 131 |

### What does and does not change the outcome (same model throughout)

| variant | status | iters | relKKT |
|---|---|---|---|
| default (no Hessian) | **`Error_In_Step_Computation`** | 131 | 3.81e-03 |
| `hessian_approximation=limited-memory` (explicit) | **`Error_In_Step_Computation`** | 131 | 3.81e-03 |
| exact `hessian` callback | `Solve_Succeeded` | 1 | 2.10e-14 |
| no Hessian + **`nlp_scaling_method=none`** | `Solve_Succeeded` | 261 | 8.25e-10 |
| **unconstrained** (drop `sum(x)=0`, same objective) | `Solve_Succeeded` | 183 | 4.29e-10 |
| constraint given as `scipy.optimize.LinearConstraint` | `Solve_Succeeded` | 175 | 1.84e-08 |

Two things worth noting from that table:

* `nlp_scaling_method=none` rescues *this* instance, but see the prevalence table below
  before reading much into it: across random instances it does **not** reduce the
  failure rate, it only makes the point reached at failure roughly an order of magnitude
  less wrong. So the gradient-based NLP scaling is not the whole story.
* The *same linear constraint* expressed as a `LinearConstraint` rather than a callable
  `dict` succeeds. That is a legitimate difference in what the facade can prove about
  the constraint (a callable `dict` cannot be shown linear, so the exact Hessian is
  dropped with a warning — correctly), but it means the failure is reachable through the
  ordinary scipy-compatible `dict` constraint form that `pounce.minimize` advertises.

### It degrades monotonically with conditioning

Same shape, `n = 8`, `t` random, `tol = 1e-10`:

| `cond(H)` | status | rel x err | relKKT |
|---|---|---|---|
| 1e1 | `Solve_Succeeded` | 4.1e-12 | 1.5e-11 |
| 1e2 | `Solve_Succeeded` | 4.5e-12 | 2.2e-11 |
| 1e3 | `Solved_To_Acceptable_Level` | 1.1e-08 | 6.5e-07 |
| 3.2e3 | **`Error_In_Step_Computation`** | 4.6e-06 | 2.4e-04 |
| 1e4 | **`Error_In_Step_Computation`** | 3.2e-05 | 3.8e-03 |
| 1e5 | **`Error_In_Step_Computation`** | 4.9e-05 | 1.8e-02 |
| 1e6 | **`Error_In_Step_Computation`** | 2.4e-02 | **1.28** |

At `cond(H) = 1e6` the returned point is not merely inaccurate — `relKKT = 1.28` means
it is not a solution at all. `cond(H) = 1e3`–`1e6` is an unremarkable range for a
scaled engineering model.

### Prevalence: 40 fresh random instances per cell

Same shape (`d = logspace(0, log10(cond), n)`, `t` random, one equality `sum(x) = s`,
`x0 = 0`), default options, counting any `status != Solve_Succeeded`:

| `n` | `cond(H)` | #bad/40 (default) | worst relKKT | #bad/40 (`scaling=none`) | worst relKKT |
|---|---|---|---|---|---|
| 8  | 1.0e3 | 17 | 1.50e-02 | 14 | 1.46e-03 |
| 8  | 3.2e3 | 28 | 2.10e-01 | 27 | 2.00e-03 |
| 8  | 1.0e4 | 31 | 1.12e+00 | 31 | 1.46e-01 |
| 8  | 1.0e5 | 35 | 1.21e+00 | 38 | 6.02e-01 |
| 12 | 1.0e3 |  9 | 1.11e-03 |  8 | 1.12e-04 |
| 12 | 3.2e3 | 24 | 1.74e-01 | 23 | 5.35e-03 |
| 12 | 1.0e4 | 35 | 5.64e-01 | 35 | 5.86e-02 |
| 12 | 1.0e5 | 38 | 1.46e+00 | 39 | 6.06e-01 |
| 20 | 1.0e3 | 15 | 1.55e-02 |  8 | 5.12e-03 |
| 20 | 3.2e3 | 27 | 1.28e-01 | 25 | 3.50e-02 |
| 20 | 1.0e4 | 36 | 8.23e-01 | 35 | 2.24e-01 |
| 20 | 1.0e5 | 39 | 1.37e+00 | 40 | 8.66e-01 |

A `relKKT` at or above 1 means the returned point is not a solution in any useful sense.

## Environment

```
pounce --version            : pounce 0.11.0
pip show pounce-solver      : Version: 0.11.0
git rev-parse --short HEAD  : 8cb4789c
date                        : 2026-09-16
python / numpy / scipy      : 3.14.2 / 2.5.3 / 1.18.1
platform                    : aarch64-apple-darwin
```

jkitchin:
Fixes #945

## Problem

`pounce.minimize` with an objective and gradient but **no Hessian** — the mode the Python frontend and the CasADi plugin select on their own — fails on a 10-variable strictly convex separable quadratic with one linear equality, no bounds, `cond(H) = 3.2e3`, started at the origin.

| | status | objective | relKKT | iters |
|---|---|---|---|---|
| before | `Error_In_Step_Computation` | — | `3.8e-03` | 131 |
| after | `Solve_Succeeded` | `29.5826274799` | `3.62e-08` | 187 |
| oracle | — | `29.5826274799` | `2.30e-14` | — |

`relKKT` is a dual residual computed from the closed form **outside** the solver, so it cannot be satisfied by a status the solver merely reports. POUNCE's own exact-Hessian arm reaches this model's KKT point in **one** iteration and its convex QP engine solves it to `7.4e-13`.

The issue's evidence that this was a breakdown rather than an exhausted budget was that the `max_iter` table did not move. It inverts:

```
             before                              after
max_iter=200    Error_In_Step_Computation/131    Solve_Succeeded/187  relKKT=3.62e-08
max_iter=1000   Error_In_Step_Computation/131    Solve_Succeeded/187  relKKT=3.62e-08
max_iter=5000   Error_In_Step_Computation/131    Solve_Succeeded/187  relKKT=3.62e-08
```

Prevalence, 480 fresh random instances of the reported shape (`n ∈ {8,12,20}` × `cond ∈ {1e3, 3.2e3, 1e4, 1e5}`, 40 per cell, random `t` and `s`, counting any status but `Solve_Succeeded`):

| `max_iter` | before | after |
|---|---|---|
| 1000 | 133 bad — 123 breakdowns, 10 out of budget | 94 bad — 49 breakdowns, 45 out of budget |
| 10000 | 122 bad — **every one a breakdown** | 44 bad — 43 breakdowns |

What remains at `cond = 1e5` is a six-pair L-BFGS window that cannot span a five-decade spectrum — gh#818's documented limit, with `limited_memory_max_history` as its remedy — not this defect.

## Cause

Not the quasi-Newton update and not the linear algebra. Over all 576 augmented-system solves on the failing trajectory the Sherman-Morrison-Woodbury path's iterative-refinement residual ratio never exceeds `4.1e-11`, inside its own `residual_ratio_max` of `1e-10` every time, so every step is the solution of the system it was handed.

What fails is `Filter::dominated_by_any`. This model's constraint is **linear**, so `theta` is satisfied to round-off from the first feasible iterate on. Every filter entry then records a `theta` of one or two ulp, and upstream's bare `theta > e.theta` ranks later trials on nothing but which way the last constraint sum rounded. Measured at iteration 94, trial 3, `alpha = 0.125`:

```
  iterate  theta = 0.0                 phi = 9.35484845847691648e-1
  trial    theta = 5.551115123125783e-16
           phi   = 9.35484840739183920e-1   (a 5.1e-9 DECREASE, passes Armijo)
  entry    theta = 1.110211922394910e-16
           phi   = 9.35484820369971715e-1
```

Note which arm decides. The `phi` arm is doing real work — the trial sits `2.0e-8` **above** the best barrier value the filter holds, a number rather than noise. The `theta` arm decides it alone, on a gap of `4.4e-16` between two roundings of the same identically-zero quantity. Where `theta` rounds the wrong way the entry falls back on its `phi` arm alone, and a one-dimensional filter on `phi` is a *monotone* test — strictly stronger than the Armijo condition the filter method exists to replace, and able to reject the step the algorithm has to take.

The α-loop then backtracks to `alpha_min` and hands a point feasible to **0.0** to the restoration phase, which has no violation to minimize. It converges on its first iterate, returns the point it was given, and the solve reports a failure from the optimum.

Context: every *other* acceptance test in this line search (`armijo_holds`, `is_sufficient_progress`, both axes) already allows for round-off through `compare_le`. The filter is the one test still reading raw values, and `theta` is where that bites because it has a hard floor the barrier objective does not.

## Fix

That last paragraph of the cause is the fix's shape. Before making a hand-off that cannot work, `BacktrackingLineSearch::run_filter_line_search` runs the α-loop **once more**, with the filter's `theta` axis measured against `theta`'s own evaluation noise, and only when the iterate it failed at is feasible to that noise. If the retry finds nothing either, the original outcome stands and the hand-off proceeds byte-for-byte as before — so the mechanism is reachable only on a trajectory that was already entering restoration. New option `filter_theta_roundoff_retry` (default `yes`; `no` restores the plain hand-off).

**Rejected: applying the floor to every filter decision.** Implemented and measured first. It fixes gh#945 and costs MacMPEC's `qpec_small` its answer at the exact floor gh#945 requires:

| always-on floor | verdict | unscaled KKT |
|---|---|---|
| 0 | `Solve_Succeeded` | `9.96e-8` — the right answer |
| 1e-16 | `Solve_Succeeded` | `9.96e-8` |
| **4.44e-16** | `ErrorInStepComputation` | `1.4e+8` |
| 1e-14 | `Solve_Succeeded` | **`5.74e-2`** — a false certificate |

`4.44e-16` is precisely what the gh#945 trial above needs. Both models sit at `‖x‖∞ ≈ 1` with `theta` at round-off, so no scale-aware formula separates them either. **Do not retune the floor to try** — the boundary is pinned as a test.

**Rejected: gating the floor to the quasi-Newton arms.** This does separate them, and only because `qpec_small` and `ralph1` are both exact-Hessian solves. That is an accident rather than a property: the Python frontend selects `limited-memory` on its own, so an MPCC lowered through a product row reaches the unprotected configuration.

**What the retry gate does instead** is ask about the *iterate* rather than about a pair of entries — is restoration going to have anything to minimize?

| | θ at the failure | own noise floor | gate |
|---|---|---|---|
| gh#945's model | `0.0` – `2e-15` | `1.2e-15` | fires |
| `qpec_small` (`prod_eq`) | `1.746646e-15` | `6.664715e-16` | **declines** (2.6× above) |
| `ralph1` | `5.8e-16` – `2.4e-15` | `~1e-23` | **declines** (orders above) |
| `square_flowsheet_resto` | `1.04e-9` | `2.0e-7` raw, **capped** | **declines** (see below) |

**Two choices the measurement forced**, both recorded where someone would look before changing them:

* The floor is `eps·(max_i rowmax|J_i|·‖x‖∞ + ‖s‖∞)` with **no safety factor on `eps`** and the **max over rows rather than the 1-norm sum**. Both under-estimates are deliberate: the 1-norm form reads `2.0e-15` on `qpec_small` and `10·eps` likewise, either of which puts its `1.746e-15` failure inside the band. `IpoptCq::theta_evaluation_noise_floor` records why this is the opposite conservatism to the neighbouring `row_noise_floor` (`64·eps`, global `‖x‖∞`), which answers the opposite question.
* **Both the gate and the filter's allowance read that floor capped** at `compare_le`'s own `10·eps·max(1, |theta|)` band. The raw bound is a product of ∞-norms, so when the largest Jacobian entry and the largest variable sit in different rows it runs orders above the noise the near-zero rows carry: `square_flowsheet_resto` reads `2.0e-7` against a `theta` pinned at `1.04e-9`, five orders above round-off and with real violation for restoration to work on. Capping only the allowance and not the gate left that model as the second of two moved sweep legs — caught in review, and fixing it made its trajectory byte-identical.

**Only the `theta` axis gets an allowance.** Giving `phi` a `10·eps·max(1, |phi|)` one was implemented and swept first: it costs `autocorr_bern55-06` its status (`SolveSucceeded`/58 → `Solved_To_Acceptable_Level`/1316) and `deb7` its objective at the 3000 cap, because that expression is an absolute number once `|phi|` is large — on `mu_fallback_point_floor` (`phi ~ 3.8e7`) it lets the filter accept a barrier increase of `8.5e-8`.

## Blast radius

`scripts/sweep-fixtures.sh`, both legs, **194 fixture-legs**, measured at `b7d1e43` against `462176c` (this PR's merge base), with the baseline re-derived from a freshly-built `462176c`. The **exact leg is byte-identical**. **One** `lbfgs` line moves, and it changes neither status nor objective:

```
lbfgs cresc4  SolveSucceeded it=281  2nd=…@Restoration_Failed/1323 → /3000, tot 1604 → 3281
```

An already-failing base solve's second-opinion excursion grinds to the 3000 cap for the same promoted verdict and objective. That is the whole corpus cost.

<sub>The base moved under this PR while it was open (`ba95130` → `462176c`, a wasm wheel fix carrying `pulp` 0.22.2 → 0.22.3 in `Cargo.lock`). Since `pulp` is the SIMD layer under the linear-algebra backend, the baseline was rebuilt at the new base rather than assumed: it is **byte-identical** to the one at `ba95130` across all 194 fixture-legs, so every measurement in this PR survives the move unchanged.</sub>

**gh#884's CLI reproducer is dissolved rather than perturbed, and that needs a reviewer's eye.** `mpcc_qpec_small_biactive` under `bound_relax_factor=0 mu_strategy_fallback=no` used to stall at `Solved_To_Acceptable_Level`/41 with an unscaled dual of `7.90e4` — the number gh#884 is named for — and need the `perturb_always_cd` promotion to reach `9.96e-8`. That stall *began* with the same line-search failure at a round-off-feasible iterate, so it is gone: the **base** attempt now converges by itself at `Optimal`/100, KKT error `5.65e-10` (unscaled dual infeasibility `6.26e-10`), `f = 4.5454149802e-10` at `x = [0.99998, 0.99999, 8.70e-6]` — two orders better in the residual than the promotion it replaces, without the re-solve gh#884 documents as lying on `ralph1`.

The consequence is that gh#884's detector, kill switch, dominance gate and promotion gate lost their end-to-end input. Those tests now run under a pinned `filter_theta_roundoff_retry=no` (`REPRO_PRE945`), which holds the gate's *input* fixed rather than preserving a bug — the state stays reachable on other models (`runaway_is_the_whole_residual` in `pounce-algorithm` is its gate), and **both default-path tests assert that the detector still fires**, so that claim is checked rather than argued. Worth a second opinion on whether that is the right call, or whether gh#884 should acquire a second, non-synthetic reproducer. `the_kill_switch_shows_what_the_scaled_aggregate_was_hiding` predicted this in its own doc and said to retire the assertion deliberately and check the detector is still reached first; that is what was done.

## Tests

Whole workspace, `cargo test --release --no-fail-fast`: **364 suites, 4227 tests, 0 failures.** Python suite: 1294 passed, 1 failed — `test_step_controller_is_what_the_jax_follower_uses`, which fails because `jax` is not installed in that container. Pre-existing and unrelated: it is a pure-Python module-identity assertion no Rust change can reach, and CI's `Python tests (jax)` job passes.

**Fail on the parent commit.** All six tests in `crates/pounce-rs/tests/issue_945_eq_constrained_lbfgs_filter_noise.rs` were run against a worktree at `ba95130` — the merge base at the time, and byte-identical to `462176c` on the whole corpus per the note above — with the new file dropped in. All six fail; four for exactly the stated reason:

```
issue_945_the_reported_model_solves      gh#945's model did not converge (156 iterations, relKKT 1.033e-5)
issue_945_needs_no_second_opinion_rung   the base solve failed and was rescued by a second-opinion rung
                                         (RestorationFailed after 156 iterations)
issue_945_the_conditioning_ladder…       cond = 1.0e4 did not converge (112 iterations, relKKT 4.745e-2)
issue_945_the_primal_point_matches…      the reported model did not converge
```

The remaining two (`issue_945_the_retry_is_the_mechanism`, `issue_945_the_failure_rate_falls_across_the_family`) set `filter_theta_roundoff_retry` and so fail at the parent on `Unknown option`, not on the behaviour. Stated rather than glossed: their evidence is same-build, option-off-against-on, which is the stronger comparison anyway.

What pins what:

* `issue_945_the_retry_is_the_mechanism` — the *cause*, not a verdict: option off reproduces gh#945 (breakdown short of `max_iter`, non-solution), default converges. Any fix by another route leaves it red.
* `the_gh945_retry_gate_is_what_this_fixture_relies_on` (`pounce-algorithm`) — `qpec_small` off-against-on across **five** reads including the iteration count, so a trajectory change that left the verdict alone still bites. Mutation-checked: widening the gate to `theta <= 1e-12` turns it red.
* `the_allowance_is_capped_at_the_compare_le_band` (filter unit test) — the cap, with `square_flowsheet_resto`'s numbers.
* `the_reproducer_no_longer_stalls_at_the_default` / `the_reproducer_no_longer_needs_the_retry` (`pounce-cli`) — the new gh#884 behaviour, held to gh#884's criterion 1 exactly as a promoted answer would be, and each asserting `dual_divergence_signature`.
* `the_gh945_roundoff_retry_reaches_the_assembled_line_search` — the option reaches the driver, not just the builder.

**Deliberately not pinned:** the stiffest cells of the reported family (`cond = 1e5`) still fail. An earlier draft asserted they converge; that was true of a rejected design and is not true of what shipped, so the assertion was replaced by `issue_945_the_failure_rate_falls_across_the_family`, which pins the direction on a sampled cell rather than a verdict on an instance. The 480-instance table above takes minutes and is recorded in the doc comment rather than run in CI.

---

- [x] Tests fail on the parent commit for the stated reason — four of six verbatim; the two option-setting ones fail on `Unknown option`, called out above
- [ ] Book page under `docs/src/` — n/a, no book page enumerates line-search options; the option's rationale lives on its registration and on `Filter::dominated_by_any`
- [x] `cargo fmt --all -- --check`, `cargo clippy` (no new warnings — count identical with and without the diff), `cargo test --release --no-fail-fast` clean
- [x] Every claim re-read against the final diff, and the sweep re-taken at `b7d1e43` against `462176c` after the base moved. The branch is a single commit; the two superseded designs it used to carry are recorded in that message, in the CHANGELOG, and as an executable pin in `the_gh945_retry_gate_is_what_this_fixture_relies_on`, but no longer appear in history as though they shipped

https://claude.ai/code/session_019RUNT89B7gGby9W8ym1EvG


jkitchin:
## Review

**Verdict: approve once #1 is resolved; #2 and #3 are small.** The diagnosis is careful and the fix is narrow. The retry only runs when the solver was already about to enter restoration, and it is well tested, including an option-off versus option-on comparison in the same build.

I checked that the second pass is safe. `run_alpha_loop` restarts from `curr` and `alpha_init`, and the second-order correction is correctly skipped (`search_dir = None`). The only acceptor state carried over is the filter-reset counters, which is fine. The floor is cleared on every exit path, and restoration's copy of the filter (`make_orig_progress_check`) is always taken with the floor at 0.

### Should fix

**1. The gate uses the uncapped floor, so the retry fires where restoration does have work to do.**

The gate in `run_filter_line_search` is `theta <= floor`, with the raw floor. Only the filter's allowance is capped. The PR's own numbers show this goes wrong on `square_flowsheet_resto`: floor `2.0e-7` against `theta = 1.04e-9`. That is 5 orders of magnitude above round-off, yet the gate treats the point as "feasible to noise".

- The cap was added because the uncapped retry cost that model its exact-leg result, so the gate demonstrably opens there.
- It is also the likely source of both `lbfgs` sweep movements on that model (9274 → 9282, and the second-opinion rung that forced the `issue850` test change).
- The stated rule is "only when restoration has nothing to minimize", and this breaks it.

Suggested fix: gate on the same capped value the filter uses.

```rust
let gate = floor.min(10.0 * Number::EPSILON * theta.abs().max(1.0));
if floor > 0.0 && theta <= gate { … }
```

This keeps every case in the PR's table:

- **gh#945:** `min(1.2e-15, 2.2e-15)` → the retry still fires.
- **`qpec_small`:** `min(6.66e-16, …)` → still declines.
- **`ralph1`:** unchanged.
- **`square_flowsheet_resto`:** now declines.

It needs a fresh sweep, and the `issue850` change to read the first iteration count may no longer be needed.

**2. "The detector still fires at the default" is claimed but never tested.**

The doc comments on `REPRO_PRE945`, `the_reproducer_no_longer_needs_the_retry` and `the_reproducer_no_longer_stalls_at_the_default` all rely on this claim. It is what justifies "nothing in this family has lost its subject". Yet neither new default-path test checks `r.statistics.dual_divergence_signature`, which the same file already reads elsewhere.

- **If the claim is true:** add `assert!(r.statistics.dual_divergence_signature)`.
- **If it isn't** (the solve now ends at `Optimal`): the `REPRO_PRE945` rationale needs rewording. That would also answer the PR's open question: gh#884 should get a second reproducer that isn't synthetic. I'd lean that way regardless, since four tests now depend on a switch that turns off the fix.

**3. The branch history will land on `main` as-is.**

The PR body says `a3236c9` and `2fcf00e` are "superseded designs whose numbers do not describe what ships". PRs here merge with merge commits, so those misleading messages would stay in `main`'s history. Squash or reword them on the branch before merging.

### Nits

- **Broken string literal** at `crates/pounce-rs/tests/issue_945_eq_constrained_lbfgs_filter_noise.rs:443`. It reads `"…; gh#945          reported 3.6e-5…"`: a trailing `\` is missing, so the failure message contains a run of spaces.
- **Option style.** `filter_theta_roundoff_retry` is registered with `add_string_option` and read with `get_string_value(..) == "yes"`. Other on/off options (e.g. `feral_increase_quality`, which `feral_increase_quality_retry`'s code reads with `get_bool_value`) use `add_bool_option` / `get_bool_value`.
- **Floor as a setting.** The floor is stored on `Filter` and toggled with set/clear around the pass. Passing it as a parameter to `dominated_by_any`, or restoring it with a guard, would make "left switched on" impossible rather than merely tested for.
- **Module doc.** `filter.rs` calls the floor "opt-in", but the option defaults to `yes`. The intended meaning is "off unless the retry pass sets it"; worth saying that.
- **Inconsistent numbers.** The reproducer's result is quoted as "KKT `5.65e-10`" (option text, `filter.rs`, the `issue884_promotion_gate` test) and as "unscaled dual `6.26e-10`" (PR body, `issue_884` test). They are different measures; pick one or label both.
- **Tight margin.** `issue_945_the_retry_is_the_mechanism` asserts that with the retry off, `off.rel_kkt > 1e-6`, against a measured `1.0e-5`. One order of magnitude of headroom may trip on an unrelated improvement to the retry-off path.
- **Length.** The CHANGELOG entry is 155 lines, and the option's help text is about 5 KB of measurements that will go stale. Consider shortening both and pointing to the tests' doc comments, where the numbers are actually checked.

### What's good

- The option-off/option-on pairing (`the_retry_is_the_mechanism` together with `the_gh945_retry_gate_is_what_this_fixture_relies_on`, which compares five values) tests the cause, not just the final status.
- The PR records the rejected designs (floor on every filter decision, gating by Hessian mode, an allowance on `phi`) together with the measurements that ruled them out.
- The external `relKKT` check and the closed-form comparison can't be satisfied by the solver just reporting success.
- The sweep was run on both legs, and the one real cost (`cresc4` now running to the 3000-iteration cap) is stated openly.

jkitchin:
All three addressed, plus every nit. Force-pushed as a single commit (`ef4bf23`); the tree is byte-identical to the reviewed `fa62ca6` content plus the fixes below. PR body updated to match.

### 1 — the gate used the uncapped floor

You're right, and it broke this change's own stated rule. The gate now reads the same capped expression the filter does, evaluated at the iterate as `entry_accepts` evaluates it at the entry.

It's worth more than the one model. **The sweep goes from 2 moved fixture-legs to 1** — `square_flowsheet_resto` is byte-identical on both legs now, and only `cresc4` remains:

```
ba95130 → ef4bf23, both legs, 194 fixture-legs
lbfgs cresc4  SolveSucceeded it=281  2nd=…@Restoration_Failed/1323 → /3000, tot 1604 → 3281
moved: 1 of 194
```

Every case in the table holds, as you predicted: gh#945 fires on `min(1.2e-15, 2.2e-15)`, `qpec_small` declines at `6.66e-16` against its `1.746e-15` failure, `ralph1` unchanged.

And your last line was right too — **the `issue850` change is no longer needed**. With the gate closed on that model the base solve's count and the run's last count stop diverging (the `lbfgs` leg is back to 2992 infeasible, not `Optimal`/153). `issue850_increase_quality_regression.rs` is reverted to its pre-branch state and passes untouched. Keeping `solve_counts` would have left a rationale in the tree describing a divergence that no longer happens.

### 2 — "the detector still fires" was claimed and never tested

The claim is true: `dual_divergence_signature` is `true` on the default path. Both default-path tests now assert it, with a failure message that says what it means if it ever goes red — that the `REPRO_PRE945` rationale is wrong and gh#884 needs a second, non-synthetic reproducer rather than a pinned option.

On your "I'd lean that way regardless": I think you're right and I've left the question open in the PR body rather than closing it here. Four tests depending on a switch that turns off the fix is a real smell even with the signature asserted — what the assertion buys is that the *detector* still sees the state, not that the end-to-end path does. If you want that reproducer in this PR rather than a follow-up, say so and I'll build one; I didn't want to widen the PR on my own judgement.

### 3 — branch history

Squashed to one commit. The two superseded designs are still recorded — in that commit message, in the CHANGELOG entry, and as an executable pin in `the_gh945_retry_gate_is_what_this_fixture_relies_on`, which carries the measured floor boundary. What's gone is only the impression that they shipped.

### Nits

| | |
|---|---|
| Broken string literal | Fixed — missing trailing `\`. |
| Option style | `add_bool_option` / `get_bool_value`, matching `feral_increase_quality`. |
| Floor as a setting | The floor is now an argument of `run_alpha_loop`, set on entry and cleared on exit. It cannot outlive the pass that asked for it, and the paths reading the filter outside an α-loop always see zero. I stopped short of threading it through the acceptor trait to `dominated_by_any` — that alone wouldn't have made it impossible (the acceptor field would remain), and doing it properly is a wider signature change than this PR should carry. |
| `filter.rs` "opt-in" | Reworded to "zero unless the retry pass sets it". |
| Inconsistent numbers | `5.65e-10` is `final_kkt_error`, `6.26e-10` is `final_dual_inf`. Both labelled at all seven sites. |
| Tight margin | `off.rel_kkt > 1e-6` → `1e-7`: two orders under the measured `1.0e-5` and two over the default `tol`. |
| Length | CHANGELOG entry 146 → 101 lines; option help cut to the mechanism plus pointers to the tests, where the numbers get re-measured when the code moves. |

### Verification

`cargo test --release --no-fail-fast`: 364 suites, 4227 tests, 0 failures. Sweep as above, against a baseline re-derived from a freshly-built `ba95130`. CI was green on the pre-squash tree (`fa62ca6`, 8/8).

---

Issue #947, opened 2026-09-17
Performance: factorization cost grows ~n^2.3 on long-horizon collocation KKTs (chain structure); FERAL fill superlinear, ND orderings worse than AMD


jkitchin:
## Summary

On a long-horizon collocation NLP, pounce's cost per iteration grows roughly as **n^2.3–2.4**: 46 s per iteration at 224k variables and about 165 s at 503k. Almost all of that time (94–98%) is spent in `LinearSystemFactorization`.

The problem is a chain of time elements that couple only to their neighbours, so factorization should scale roughly linearly with horizon length. FERAL's symbolic fill does not scale that way on this pattern: nnz(L) per row triples between 28k and 1M rows. The nested-dissection orderings are *worse* than AMD. The same problem through IPOPT + MA57 shows the same superlinear trend, but is 1.2–1.9x faster per iteration.

A self-contained script is at the bottom (numpy only). It regenerates the exact Jacobian and Hessian patterns pounce receives, at any horizon length, as Matrix Market KKT files.

## The problem

- **Network:** transient gas-network optimal control on GasLib-40, a public benchmark. The isothermal Euler equations are discretized with 10 finite volumes per pipe, giving 775 states per time point, including one quadrature state.
- **Time discretization:** Radau IIA with 3 collocation points per element, `nfe` elements.
- **Controls:** 6 compressor ratios, piecewise linear, one knot per element, with ramp limits.
- **Constraints:** a pressure floor. For free nodes it is a variable bound; for the 6 compressor outlets (`alpha_c * P_in`) it is a general constraint.
- **Objective:** compressor energy.
- **Interface:** `pounce.jax.from_jax(..., sparse=True, jac_pattern=..., hess_pattern=...)`. The patterns are assembled from the structure and were verified as supersets on a small mesh against probed Jacobians and a dense Lagrangian Hessian.
- **Structure:** element `i` couples only to element `i-1` (through its end-point state) and to at most 4 control knots. The Hessian is block diagonal over collocation points, plus control–state coupling within a point.

The model code itself is not public. The script below reproduces the sparsity exactly: its Jacobian and Hessian nnz match what pounce reported at nfe = 24, 48, 96 and 216 (351,783 / 230,145, 706,239 / 461,265, 1,415,151 / 923,505 and 3,187,431 / 2,079,105).

## Measurements

**Environment:**
- pounce 0.11.0 (editable, `598aedf1`), FERAL 0.17.0 (crates.io), default options
[...]

PR #955, opened 2026-09-20
Block-structured KKT: declare a model's blocks, factorize them in parallel (3.8-4.8x on the factorization)

jkitchin:
Carries structural knowledge into POUNCE's KKT solve: a model that is
*arrowhead* — independent blocks coupled only through a small border — can
declare that, and the factorization runs block-parallel instead of
monolithically. Plus the instrumentation that made the case, and the
measurements that decided what to build. Relates to #947.

**Exact by construction.** A permutation plus block elimination of the same
symmetric indefinite system, inertia by Haynsworth additivity. Not a
decomposition *algorithm*: no convexity assumed, no cuts, no bounds. The
evidence is that the solve takes the same iterates — identical iteration counts
and objectives, digit for digit, on every model measured.

## What ships

- **`Problem.set_block_structure(var_blocks, con_blocks)`** — one label per
  model variable and constraint, negative for shared. POUNCE maps them onto the
  KKT layout itself, which is the point: a model declaring 53 shared columns
  can come back with a border of 18 because 35 were fixed at their bounds.
  Also `block_structure_file` for the `.nl` path and `kkt_block_detect` for
  models that will not declare.
- **`kkt_block_restoration`** (default `yes`) — the partition applies to the
  restoration sub-IPM unchanged, because the restoration KKT is reduced onto
  the original system before it is factorized.
- **`kkt_block_min_size`** (default 256) — refuses a partition whose blocks are
  too narrow to pay for their own dispatch.
- **Factorization and pivoting statistics** in the solve report's
  `linear_solver` object and a new Python `info["linear_solver"]`, including a
  nested `restoration` object and the block path's own summary.
- **`feral_ordering_preprocess`** exposing FERAL's matching-compressed ordering.
- **Two reporting defects fixed**: under L-BFGS two backends shared one sink and
  each overwrote it; and after a second-opinion ladder that kept the original
  verdict, the summary described the last rejected rung.

Default path unchanged — `kkt_block_detect` defaults to `no` and nothing
engages without a declaration — so the fixture sweep does not apply.

## Measured

Corrective N-1 AC-SCOPF from pglib, same binary, `.nl` through the CLI.
`case1354_pegase`, 65 blocks over a 259-column border:

| K | variables | factorization | wall | iterations |
|---|---|---|---|---|
| 16 | 54,859 | 5.66 → 1.31 s (4.3×) | 18.1 → 10.0 s | 43 = 43 |
| 32 | 106,491 | 11.52 → 2.42 s (4.8×) | 37.7 → 21.1 s | 49 = 49 |
| 64 | 209,755 | 10.72 → 2.82 s (3.8×) | 125.1 → 111.9 s | 87 = 87 |

Restoration on the K=64 run: 35.0 s → 11.4 s. The end-to-end factor is smaller
than the factorization factor because factorization is roughly a third of these
solves — stated in the docs too, so a reader who expects 4× does not conclude
the feature is broken.

**The crossover, and the guard for it.** The monolithic factorization already
finds arrowhead structure, so the block path's only win is parallelism, against
a fixed per-block cost. On a 32-block arrowhead, factorization full-space →
declared: 0.31× at 60 columns per block, 0.57× at 125, 0.98× at 250, 1.56× at
500, 2.49× at 1,000, 3.23× at 2,000. `kkt_block_min_size` refuses the losing
half; re-measured with it, those rows read 1.02× / 1.02× / 1.01× and the
winning rows are unchanged.

**Detection's limits.** It works where the shared columns stand out
(`case118_ieee`: degree 130 against a median of 6) and fails where they do not
(`case1354_pegase`: degree ~50, reporting 581 blocks over a 7,301-column
border, refused by the border guard). That is the argument for a declaration.

## What the measurements ruled out

- **The gas quadratic is fill**, from refining the time step rather than
  extending the horizon: n^2.35 under the default ordering, n^1.47 under METIS,
  8× faster at 224k variables. The default `auto` never selects nested
  dissection.
- **A structure-derived network × time ordering does not beat METIS** — matched
  at 56k, 44–54% slower at 112k–224k.
- **So gas does not justify a block-structured KKT**, and Phases 3–4 of the
  plan are marked unjustified. The arrowhead does justify it, which is why this
  PR is about the arrowhead.

## Cross-repo

discopt#1370 is implemented on discopt's side: it resolves a model's
`set_block` / `mark_coupling` annotations into this API. Verified end to end
against this branch. That run is what found the crossover above — neither
repo's tests could have: discopt's spy the labels, POUNCE's declare their own
structure.

Note discopt's passthrough is guarded on `hasattr(problem,
"set_block_structure")`, so **their feature is inert until this merges.**

## Verification

`cargo fmt --check`, CI's clippy gate, `cargo test --workspace --exclude
pounce-hsl`, the Python suite (1672 passed, 6 skipped) and `make ask-check` all
pass. Every new test is mutation-checked — run against the broken code it
guards — and the two test files carry their own mutation tables. Docs in
`docs/src/options.md` ("Block-structured KKT") and `docs/src/python.md`, with a
CHANGELOG entry. Working notes: `dev-notes/kkt-scaling-phase0b.md` and
`dev-notes/structured-kkt-decomposition-plan.md`. `benchmarks/kkt_scaling/`
holds the generators and harnesses; `blockfac/` is deliberately not a workspace
member.

## Known follow-up, measured but not in this PR

Evaluation is **70%** of the `case1354_pegase` K=16 solve (the Lagrangian
Hessian alone 52%) against 14% for the factorization this PR parallelises, and
`pounce-nl` evaluates single-threaded. A probe puts that work at 6.67×
throughput on 8 threads, and the parallel axes are clean (colors for the
Hessian, rows for the Jacobian, both with `&self` tapes). That is a separate PR
against `pounce-nl`, not this one.

https://claude.ai/code/session_01VH3dorgk7tYQ1SnPU34Yxh


jkitchin:
### Stacked: the block-structure mapper (`dd63be1b`)

The block solver in `d890af44` could only be *told* its structure in KKT
coordinates — which no modelling layer knows, since they depend on which
variables pounce fixed and removed and on the equality/inequality split. The
other input, detection, is a degree heuristic: on `case1354_pegase` it reports
581 blocks over a 7,301-column border and the border guard refuses it.

`map_block_structure_to_kkt` takes one label per *model* variable and one per
*model* constraint (negative for shared) and maps them through the same
`BoundClassification` the KKT is assembled from. Surfaces:
`IpoptApplication::set_block_structure`, `Problem.set_block_structure`, and
`block_structure_file` for the `.nl` path. A declaration that does not fit is a
warning and a monolithic solve.

**`case1354_pegase`, the family detection could not read** — same binary, `.nl`
through the CLI, runs serialised on an idle machine, 17/33/65 blocks over a
259-column border:

| K | vars | factorization std → declared | wall | iterations | objective |
|---|---|---|---|---|---|
| 16 | 54,859 | 5.66 → 1.31 s (**4.3×**) | 18.1 → 10.0 s | 43 = 43 | identical |
| 32 | 106,491 | 11.52 → 2.42 s (**4.8×**) | 37.7 → 21.1 s | 49 = 49 | identical |
| 64 | 209,755 | 10.72 → 2.82 s (**3.8×**) | 125.1 → 111.9 s | 87 = 87 | identical |

Two things worth reading off rather than skipping:

- The mapping is **not** the identity, and that is the point: `case118_ieee`
  declares 53 shared generator columns and the border comes out at **18**,
  because 35 of those generators have `pmin == pmax` and `make_parameter`
  removed them. Hand-written KKT labels would have had to know that.
- K=64 is **infeasible**, detected identically on both paths, and its wall gain
  is only 1.12× because **restoration's factorizations are still monolithic**:
  70 of them, 34.7 s, unchanged between the legs, against 2.8 s in the main
  solve. The restoration sub-IPM adds a `p`/`n` pair per constraint, so its KKT
  is not the one the declaration was mapped onto. That is 92% of the remaining
  factor time, and the next thing to extend.

Verification: `cargo fmt --check`, CI's clippy gate, `cargo test --workspace
--exclude pounce-hsl` and the Python suite (1672 passed, 6 skipped) all pass.
New tests are mutation-checked. No default solve path changes.

https://claude.ai/code/session_01VH3dorgk7tYQ1SnPU34Yxh

jkitchin:
### Stacked: restoration takes the partition too (`cdbfa31b`)

The previous comment said restoration was 92% of the remaining factorization
time, and gave the wrong reason — that the restoration sub-IPM's `p`/`n` pair
per constraint makes a different KKT. The first half is true and the conclusion
does not follow: `AugRestoSystemSolver` eliminates those four slack blocks
analytically and hands the **original 4-block system** to its inner solver, so
the matrix actually factored in restoration has the outer system's dimension
and sparsity. The outer labels describe it exactly. Nothing had to be
re-derived; the labels only had to arrive.

What blocked them was plumbing: every frontend mints restoration's inner
builder *before* the solve, hence before the KKT layout and the mapping onto it
exist. The builder now takes a cell the application publishes into as it builds
the outer algorithm — the same pattern already used for the quality-escalation
tally, so no frontend changes. `kkt_block_restoration` (default `yes`) turns it
off, which is how the split below was measured.

**`case1354_pegase` K=64** (209,755 variables, 65 blocks / 259-column border, infeasible):

| | main factor | restoration factor | wall |
|---|---|---|---|
| monolithic | 10.72 s | 35.05 s (70×) | 125.1 s |
| declared, restoration monolithic | 2.82 s | 35.05 s (70×) | 113.2 s |
| declared, restoration too | 2.86 s | **11.39 s** (88×) | **82.8 s** |

Same verdict, same 87 outer iterations, same objective to ten digits, same
`restoration_calls` (10) and `restoration_inner_iters` (66), and
`quality_escalations = 0` on both — pounce's own residual test accepted every
block solve.

**One thing does move.** Block elimination cannot pivot across the block/border
split, so the factorizations are not bit-identical, and on this degenerate
model the restoration sub-solves take a marginally different path: 281
differing iteration-log lines, 425 vs 428 logged inner iterations, 70 vs 88
factorizations, regularized iterations 42 vs 43. Both reach the same answer,
and it is still 3× faster. That is restricted pivoting, not a defect — and it
is why the option exists. The feasible instances (K=16, K=32) never enter
restoration, so their trajectories are unchanged and the table in the previous
comment stands.

New test `crates/pounce-restoration/tests/block_structure_restoration.rs` with
an arrowhead fixture (k infeasible two-variable cells sharing one linking
variable) and its own mutation table; all four mutations were run and each
fails exactly its named test.

Gates: `cargo fmt --check`, CI's clippy gate, `cargo test --workspace --exclude
pounce-hsl`, Python suite (1672 passed, 6 skipped).

https://claude.ai/code/session_01VH3dorgk7tYQ1SnPU34Yxh

jkitchin:
### Stacked: the two halves meet, and the crossover they exposed (`0b20d4ed`)

discopt#1370 is implemented and merged on discopt's side, so the path is
reachable end to end for the first time: `set_block` ->
`block_structure_for_model` (labels resolved into the emitted NLP's index
space, cross-checked against the Jacobian *and* the Lagrangian Hessian) ->
`solve_nlp_from_model` -> `Problem.set_block_structure`. Run against this
branch, a 24-block arrowhead declared in discopt reaches

```
INFO pounce::kkt: block-parallel KKT path engaged blocks=24 border=1
```

and the partition comes back through `NLPResult.linear_solver`. Same status,
iterations and objective as the full-space arm.

**What that run immediately found: the block path has a crossover, and had no
guard.** Factorization seconds on a K=32 arrowhead, full-space -> declared:

| columns per block | n | full-space | declared | ratio |
|---|---|---|---|---|
| 60 | 1,921 | 0.0060 s | 0.0195 s | **0.31x** |
| 125 | 4,001 | 0.0133 s | 0.0231 s | **0.57x** |
| 250 | 8,001 | 0.0260 s | 0.0264 s | 0.98x |
| 500 | 16,001 | 0.0534 s | 0.0343 s | 1.56x |
| 1,000 | 32,001 | 0.1244 s | 0.0500 s | 2.49x |
| 2,000 | 64,001 | 0.2725 s | 0.0843 s | 3.23x |

The cause is the reason ordering was never the opening on this topology: the
monolithic factorization *already* finds the structure, so the block path's
only win is parallelism, against a fixed per-block cost — one symbolic
analysis, one task, one border tail. Below a few hundred columns per block that
cost is the whole story.

`decide` now refuses a partition whose **median** block is narrower than
`kkt_block_min_size` (default 256, `0` disables) and logs why. Re-measured with
it: the losing rows read 1.02x / 1.02x / 1.01x — refused, nothing lost — and
the winning rows are unchanged (1.72x / 2.65x / 3.52x). `case1354_pegase` K=16
is far above the threshold (13,601 columns per block) and is unaffected.

Worth stating plainly: **neither repo's tests could have caught this.**
discopt's spy the labels; pounce's declare their own structure. It took running
one against the other, and it is the first thing a user with a small model
would have hit.

**Two corrections from discopt's measurements**, now recorded in the dev-note:

- "Evaluation is 48% of this solve" was measured on the `.nl`/ASL path. On
  pounce's own Rust tape it is **20-24%**, so Phase 5a is worth ~1.45x *after*
  the factorization win rather than being "the larger half" — the two compound
  to roughly 1.9x, not 3x.
- The 0.044x directional-derivative figure was against ASL too. Against the
  tape it is 5.2x (small blocks, K=64) to 17.7x (realistic block sizes). The
  mechanism survives the better baseline; the headline number does not. Their
  dense per-block arm loses almost everywhere, which is the evidence that the
  win is the shared pattern and coloring, not the engine.

The general lesson, since this branch keeps relearning it: a ratio is a
statement about its baseline, and only a baseline the user actually runs
predicts what the user sees.

Gates: `cargo fmt --check`, CI's clippy gate, `cargo test --workspace --exclude
pounce-hsl`, Python suite (1672 passed, 6 skipped). New test mutation-checked.

https://claude.ai/code/session_01VH3dorgk7tYQ1SnPU34Yxh

